PR #319, opened 2014-12-09
(EZ-15) Add lein-ezbake configuration.

waynr:


waynr:
@dankreek 
I agree with your `config` comment and think that would be good feedback for puppetlabs/ezbake#155. It'd probably be best to go with a sane default but make it configurable using the `lein-ezbake` key in the project config.

Also, it might be best to hold off on merging this until we actually have a released `lein-ezbake` since the patch currently depends on a `SNAPSHOT` version.

config/conf.d/global.conf
```diff
@@ -0,0 +1,9 @@
+global: {
+    # Path to logback logging configuration file; for more
+    # info, see http://logback.qos.ch/manual/configuration.html
+    logging-config: /etc/puppetserver/logback.xml
+}
+
+product: {
+    name: puppetserver
+}
```
cprice404: So, you're creating a new directory called `config` at the root, to put the ezbake stuff into.  Questions about this...

1. Do these files overlap with the ones in resources/ext/ezbake?  I think this basically boils down to the difference between composite ezbake builds and single-project ezbake builds, right?  If so, can we put a bit more thought into directory naming conventions to try to make the distinction more obvious?
2. Can we maybe call the top-level dir 'ezbake' to make it more obvious?
3. Should that path be configurable in the ezbake plugin settings?  (but still default to something explicit, like 'ezbake'?)

cprice404:
well.  i missed all of the other comments on this PR, they didn't show up in the file view for some reason... my bad.  so it seems like we're all on the same page on that stuff.

cprice404:
I've done a basic review on this code, and am pretty much +1.  I'm experiencing some really weird bug when I try to run it locally, but it doesn't seem to be happening in other environments, so I don't want to block on that.  Let's start getting the wheels turning and I'll pass along an update if I track down my local issue.

Considerations:

1. We need to look at the other outstanding PRs for ezbake and decide what order to do things in.  If there are other PRs that should land before this one (or should be targeted at 'stable'?), let's get those sorted out ASAP and get this rebased.  Or, if we decide we're going to merge this one first, let's make sure all of the authors of the other PRs are on board.
2. We need to make sure that everyone who is using ezbake for builds related to PE3.7.2 knows that this is about to land in master, and that we need to do all of the PE 3.7.2 builds off of stable.  I assume that there may be CI jobs that need to be updated accordingly.
3. If there is not already a jira ticket capturing the config directory path tweaks that were suggested in all of the previous comments on this PR, let's create one and link it in.

@waynr can you please take point on those actions, and pull in anyone else you need to?

@pcarlisle heads up that this is coming in case you guys weren't already aware of the ezbake master/stable situation.

waynr:
@cprice404 
I can take point on those items, thanks for your feedback!

Have you tried running ezbake-as-an-app off its master branch to see if the same `me.raynes.fs.list-dir` happens when you run that? Also, is it possible something in your user profile is interacting with leiningen? I'm wondering if some strange combination of hooks could lead to the problem we saw yesterday.

config/conf.d/global.conf
waynr: Just FYI this directory is configurable in `lein-ezbake` now, I just haven't changed it in this PR yet. So are you thinking `ezbake/config` would be good?

cprice404: ezbake/config would be fine for now.

cprice404:
@waynr I'd be willing to merge this this week if you can get that last change in re: the ezbake directory.  (lower priority than 3.7.2 though, obv)

waynr:
`:uberjar` profile needs tk-jetty-9 dependency

waynr:
@cprice404 I have updated the uberjar profile so this PR should be ready to go. If puppetlabs/ezbake#176 gets in before this and I have an opportunity to release lein-ezbake 0.1.1 I will update this PR to point to that new version.

cprice404:
@waynr we had some problems with acceptance tests after I merged Ruth's PR, because we need a change to the webrouting config file in ezbake.  I didn't want to deal with trying to do that on an old branch of ezbake, so I just merged your PR.  Now I think we're totally broken until you update the Jenkins jobs to use the plugin, but as soon as you do that, I can fix the smoke test failure that has to do with the webrouting config.

PR #335, opened 2014-12-17
(SERVER-85) Sync CA CRL to Host CRL at startup

camlow325:
This commit changes the master to, at each startup, copy the file
pointed at by the Puppet `cacrl` setting, if it exists, over to the
file pointed at by the Puppet `hostcrl` setting.

This commit also changes the `override-webserver-setting!` logic to
use the file at the `hostcrl` setting rather than the one at the `cacrl`
setting to determine which file the webserver uses to validate
certificates for connecting clients.  This was done based on the fact
that a "standalone" master -- one hosted without a corresponding active
CA in the same container -- would have used the `hostcrl` setting under
the Ruby master.  With the logic in place to copy the `cacrl` to the
`hostcrl` when a real CA is in place, the latest CRL, when the master
and CA are running in the same container, should now be in place at
service startup.

### DIFF
--- a/test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
+++ b/test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
@@ -448,6 +448,35 @@
               "No exception thrown even though no file existed for copying")
           (fs/copy copy cacert))))))
 
+(deftest retrieve-ca-crl!-test
+  (testing "CRL file copied when it doesn't already exist"
+    (let [tmp-confdir (fs/copy-dir confdir (ks/temp-dir))
+          settings    (testutils/master-settings tmp-confdir)
+          ca-settings (testutils/ca-settings (str tmp-confdir "/ssl/ca"))
+          cacrl       (:cacrl ca-settings)
+          hostcrl     (:hostcrl settings)
+          cacrl-text (slurp cacrl)]
+
+      (testing "Copied cacrl to hostcrl when hostcrl not present"
+        (retrieve-ca-crl! cacrl hostcrl)
+        (is (= (slurp hostcrl) cacrl-text)
+            (str "Unexpected content for hostcrl: " hostcrl)))
+
+      (testing "Copied cacrl to hostcrl when different hostcrl present"
+        (spit (:hostcrl settings) "12345")
+        (retrieve-ca-crl! cacrl hostcrl)
+        (is (= (slurp hostcrl) cacrl-text)
+            (str "Unexpected content for hostcrl: " hostcrl)))
+
+      (testing (str "Doesn't throw exception or create dummy file if no "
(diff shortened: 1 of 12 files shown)


test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -448,6 +448,35 @@
               "No exception thrown even though no file existed for copying")
           (fs/copy copy cacert))))))
 
+(deftest retrieve-ca-crl!-test
+  (testing "CRL file copied when it doesn't already exist"
+    (let [tmp-confdir (fs/copy-dir confdir (ks/temp-dir))
```
jeffmccune: My understanding is that a single temporary directory will be created and used for the following examples.  Will this mean the results will be indeterminate if the examples are run concurrently?

I need to learn more about how tests execute in Clojure and if it's common to run them concurrently or if we plan to.

jeffmccune:
+1 with my only concern being concurrent execution of the tests.

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
nwolfe: This is a common pattern we have with the CA tests - there's usually a tmp dir "sandbox" that the subsequent `testing` blocks operate within.  If each `testing` block was executed concurrently then yes, we would definitely have problems, but we don't run tests that way.  I would think that parallelization would happen at the `deftest` level rather than the lower `testing` level though, so we shouldn't have that problem.

Or maybe I'm wrong and you can parallelize the individual `testing` blocks?  If we decided to do that we'd have some other tests to redo anyways.  Either way, we're planning on doing some heavy refactoring with the persistence of the CA data so this filesystem test code won't live much longer.

jeffmccune: K, disregard the concern then.  It seems like we would want to allow concurrent testing at some point, but this pattern of initializing shared state at the top of a `deftest` isn't what's holding us back.

For what it's worth, I took a stab at using `[com.holychao/parallel-test "0.3.0"]` and I ended up having to mark the majority of `deftest` variables as :static, and I didn't even finish.  We'd need to refactor most of our tests to execute them concurrently so this probably needs to be addressed at a different level.

nwolfe:
@camlow325 Just a reminder that we'll need to update the CA proxy implementation in pe-p-s-e: https://github.com/puppetlabs/pe-puppet-server-extensions/blob/master/src/clj/puppetlabs/enterprise/services/reverse_proxy/reverse_proxy_ca_service.clj

nwolfe:
:+1:

camlow325:
@nwolfe Thanks for pointing that out.  I created PE-7395 to cover the pe-p-s-e followup work.

PR #343, opened 2014-12-20
(SERVER-253) Documentation for subcommands

nwolfe:
Considering that most of these commands are standard Ruby ones I didn't feel the need to go into much detail. This ended up being mostly about showing examples of each one and calling out the `--` when running from source.

Happy to go change any/all of this; I just wanted to get something up before the break!

### DIFF
--- a/documentation/subcommands.markdown
+++ b/documentation/subcommands.markdown
@@ -1,0 +1,111 @@
+---
+layout: default
+title: "Puppet Server: Subcommands"
+canonical: "/puppetserver/latest/subcommands.html"
+---
+
+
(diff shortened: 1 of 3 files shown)


documentation/subcommands.markdown
```diff
@@ -0,0 +1,111 @@
+---
+layout: default
+title: "Puppet Server: Subcommands"
+canonical: "/puppetserver/latest/subcommands.html"
+---
```
nfagerlund: Y'all got this right I think! Sorry for the delay. :)

PR #345, opened 2014-12-23
(SERVER-269) Passthrough binary payload properly for master requests

camlow325:
In the previous commit, the body for any HTTP requests made to the
Puppet master service with a Content-Type of <none> or
"application/octet-stream" was inappropriately decoded into a UTF-8
string.

With the changes in this commit, the body for any HTTP requests made to
the Puppet master service with a Content-Type of <none> or
"application/octet-stream" is now passed through as raw bytes to the
JRuby request layer, where it is converted into an ASCII-8BIT Ruby
String.

Puppet agent requests to the "file_bucket_file" endpoint include a
Content-Type of "text/plain".  While these requests should technically,
per the HTTP spec, be transformed to ISO-8859-1 or US-ASCII instead of
UTF-8, the actual payload is intended to be treated as "binary".  This
commit treats requests to the "file_bucket_file" endpoint as though they
were including an "application/octet-stream" Content-Type, allowing the
raw bytes in the request body to be sent down to the JRuby request layer.

### DIFF
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -41,7 +41,9 @@
     ;; the actual payload is "binary".  Coercing this to
     ;; "application/octet-stream" for now as this is synonymous with "binary".
     ;; This should be removed when/if Puppet agents start using an appropriate
-    ;; Content-Type to describe the input payload.
+    ;; Content-Type to describe the input payload - see PUP-3812 for the core
+    ;; Puppet work and SERVER-294 for the related Puppet Server work that
+    ;; would be done.
     (compojure/PUT "/file_bucket_file/*" request
                    (request-handler (assoc request
                                            :content-type
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -33,8 +33,20 @@
                    (request-handler request))
     (compojure/GET "/file_bucket_file/*" request
                    (request-handler request))
+
+    ;; TODO: file_bucket_file request PUTs from Puppet agents currently use a
+    ;; Content-Type of 'text/plain', which, per HTTP specification, would imply
+    ;; a default character encoding of ISO-8859-1 or US-ASCII be used to decode
+    ;; the data.  This would be incorrect to do in this case, however, because
+    ;; the actual payload is "binary".  Coercing this to
+    ;; "application/octet-stream" for now as this is synonymous with "binary".
+    ;; This should be removed when/if Puppet agents start using an appropriate
+    ;; Content-Type to describe the input payload.
     (compojure/PUT "/file_bucket_file/*" request
-                   (request-handler request))
+                   (request-handler (assoc request
+                                           :content-type
+                                           "application/octet-stream")))
+
     (compojure/HEAD "/file_bucket_file/*" request
                    (request-handler request))
     (compojure/GET "/catalog/*" request
--- a/src/clj/puppetlabs/services/request_handler/request_handler_core.clj
+++ b/src/clj/puppetlabs/services/request_handler/request_handler_core.clj
@@ -2,7 +2,8 @@
   (:import (java.security.cert X509Certificate)
            (java.util HashMap)
            (java.io StringReader)
-           (com.puppetlabs.puppetserver JRubyPuppetResponse))
+           (com.puppetlabs.puppetserver JRubyPuppetResponse)
+           (org.apache.commons.io IOUtils))
   (:require [clojure.tools.logging :as log]
             [clojure.string :as string]
             [clojure.walk :as walk]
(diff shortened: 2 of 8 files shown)


src/clj/puppetlabs/services/request_handler/request_handler_core.clj
```diff
@@ -2,7 +2,8 @@
   (:import (java.security.cert X509Certificate)
            (java.util HashMap)
            (java.io StringReader)
-           (com.puppetlabs.puppetserver JRubyPuppetResponse))
+           (com.puppetlabs.puppetserver JRubyPuppetResponse)
+           (org.apache.commons.io IOUtils))
```
jeffmccune: I don't think IOUtils is being used anywhere, suggest removing the import.

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
```diff
@@ -73,30 +101,25 @@
...
+  (let [body-for-jruby (body-for-jruby request)]
+    (->
+      request
+      ; Leave the slurped content under an alternate key so that it
+      ; is available to be proxied on to the JRubyPuppet request.
+      (assoc :body-for-jruby body-for-jruby)
+      ; Body content has been slurped already so wrap it in a new reader
+      ; so that a copy of it can be obtained by ring middleware functions,
+      ; if needed.
+      (assoc :body (if (string? body-for-jruby) (StringReader. body-for-jruby)))
```
jeffmccune: Would it be more idiomatic clojure to use `clojure.tools.reader.reader-types/string-reader` instead of the Java implementation?

src/ruby/puppet-server-lib/puppet/server/network/http/handler.rb
```diff
@@ -32,7 +34,12 @@ def path(request)
   end
 
   def body(request)
-    request["body"]
+    body = request["body"]
+    if body.java_kind_of?(InputStream)
+      body.to_io.read()
```
jeffmccune: Are we concerned about `IO#read` copying the entire file content in memory?  As this pull request is concerned with file bucket files, do we have a sense of the behavior when large files are being stored?

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
camlow325: Good call.  I see a few other unused ones in that namespace and will remove those too.

src/ruby/puppet-server-lib/puppet/server/network/http/handler.rb
jeffmccune: Correction, IO#read does not do an in-memory copy if I understand http://ruby-doc.org/stdlib-1.9.3/libdoc/stringio/rdoc/StringIO.html#method-i-read

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
camlow325: It would be better to avoid constructing the Java object, but I don't think the `string-reader` function that you mentioned will work.  The object is ultimately consumed by a ring middleware function, `ring.middleware.params/assoc-form-params`, which uses Clojure's `slurp`.  `slurp` tries to coerce the argument into a Java `InputStream` and fails with an `IllegalArgumentException`:

```
IllegalArgumentException Cannot open <#<StringReader clojure.tools.reader.reader_types.StringReader@53e7ec65>> as an InputStream.  clojure.java.io/fn--8616 (io.clj:173)
```

Would be nice if `slurp` knew how to consume a `clojure.tools.reader.reader_types.StringReader` properly, but looks like we're stuck here...

src/ruby/puppet-server-lib/puppet/server/network/http/handler.rb
jeffmccune: > I'm not sure there's anything which can be done to improve the memory utilization at this point.

Me either, was just wondering...  It does make sense that the behavior is the same as it always has been, so we're not making anything worse.  If it is actually a problem then the fix would have to be in Puppet anyway, not Puppet Server.

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
jeffmccune: Yep, disregard my comment.  Looks like `string-reader` defines another `StringReader` type which isn't compatible with `slurp`.

Here's the `deftype`: https://github.com/clojure/tools.reader/blob/ececc05cd8d505cc5f4ff9cee408e202b6cbf402/src/main/clojure/clojure/tools/reader/reader_types.clj#L46-L56

jeffmccune:
+1 from me.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -33,8 +33,20 @@
                    (request-handler request))
     (compojure/GET "/file_bucket_file/*" request
                    (request-handler request))
+
+    ;; TODO: file_bucket_file request PUTs from Puppet agents currently use a
```
camlow325: I added a commit, 7325909, which includes the ticket numbers for a couple of JIRA tickets I filed for the follow-on work - PUP-3812 for the core Puppet work and SERVER-294 to adapt Puppet Server to the core Puppet work if PUP-3812 is ever addressed.

nwolfe:
:+1: - tested on VM!

nwolfe:
Still waiting for @cprice404 to review

src/ruby/puppet-server-lib/puppet/server/network/http/handler.rb
cprice404: Yeah, this topic has come up a few times before and it's definitely something we should be keeping an eye on, as it would be an obvious and potentially significant improvement... but I don't think we can tackle it right now.

PR #346, opened 2014-12-28
(FOR REVIEW ONLY) Pure clojure puppet file_metadata endpoint

masterzen:
Hi,

Please find here an attempt to build a pure clojure implementation of a Puppet terminus (here the file_metadata endpoint). This is very experimental material that *must not be merged-in*, I'm pushing it here only for other people to review and comment the code. This code hasn't been tested with a real puppet client, I would expect it to break horribly in this case.

Warning: this is my very first clojure code, it might not be very idiomatic nor good in plenty of ways. Do not hesitate to comment on the style so that I can learn :)

So what does this PR contains (and not):
- [x] a trapperkeeper service that parses fileserver.conf and allows other services to find a mount from a path
- [x] implementation of the file_metadata endpoint matching what a puppet master returns for file, links and directories
- [ ] support for special mount "plugins"
- [ ] support for special mount "modules"
- [ ] support for file_metadatas
- [ ] fileserver.conf ACL support
- [ ] auth.conf ACL support

I plan to update this PR from time to time as I fix issues pointed by reviews and implement the missing features listed above.

Thanks!


src/clj/puppetlabs/services/file_serving/config/puppet_fileserver_config_core.clj
```diff
@@ -0,0 +1,93 @@
+(ns ^{:doc
+      "Tools to parse Puppet fileserver.conf."}
+    puppetlabs.services.file-serving.config.puppet-fileserver-config-core
+  (:require
+    [puppetlabs.kitchensink.core :as ks]
+    [schema.core :as schema]
+    [puppetlabs.puppetserver.ringutils :as ring])
+  (:use [clojure.java.io :only (reader)]
```
fpringvaldsen: Since Clojure 1.4.0, you've been able to pass a `:refer` option to `:require`. I believe it's now considered best practice to do this instead of `:use`.

KevinCorcoran: Yup, we tend to avoid `:use` at all costs, always use `:require mynamespace :refer [function1 function2]`, and sometimes also `:refer :all` for cases like testing or DSL usage.

src/clj/puppetlabs/services/file_serving/file_metadata/file_metadata_core.clj
```diff
@@ -0,0 +1,200 @@
...
+            [puppetlabs.services.file-serving.utils.posix-utils :as posix-utils]
+            )
+  (:use [clojure.java.io :only (input-stream)]
+        [ring.middleware.params]
+        ))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; 'handler' functions for HTTP endpoints
+
+(defn- checksum-of-file
```
KevinCorcoran: Might be preferable to use something like https://github.com/tebeka/clj-digest instead of rolling our own here.

rlinehan:
@masterzen Sorry it's taken so long for us to get to reviewing this! I'm going to go ahead and do some review and then close the PR, since it's a "for review only" PR and it's been open for a while. After that we'll move on to reviewing trapperkeeper-authorization. Does that sound okay to you?

src/clj/puppetlabs/services/file_serving/config/puppet_fileserver_config_core.clj
```diff
@@ -0,0 +1,93 @@
...
+        [clojure.string :only (split trim split)]))
+
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; internal helpers
+
+(defn- parse-one-line
+  [s]
+  (if (= (first s) \[)
+    (-> s (subs 1 (.indexOf s "]")) trim keyword)
```
rlinehan: Style nitpick: I think in general, it's better for readability to format threading macros so that each form is on it's own line:

```clojure
    (-> s
        (subs 1 (.indexOf s "]"))
        trim
        keyword)
```

src/clj/puppetlabs/services/file_serving/file_metadata/file_metadata_core.clj
```diff
@@ -0,0 +1,200 @@
...
+  [time]
+  (.format date-format (Date. time)))
+
+(defn ruby->date
+  [date]
+  (.parse date-format date))
+
+(defn- checksum-of-dir
+  [posix-attr]
+  "Compute the ctime checksum of a given directory"
```
rlinehan: Docstrings should go before arguments.

src/clj/puppetlabs/services/file_serving/config/puppet_fileserver_config_core.clj
```diff
@@ -0,0 +1,93 @@
...
+    (->> (line-seq r)
+         (map strip-comment)
+         (remove (fn [s] (every? #(Character/isWhitespace %) s)))
+         (map parse-one-line)
+         mapify)))
+
+(schema/defn find-mount :- MountPath
+  "find a given mount from a path."
+  [mounts :- Mounts
+   path :- schema/Str]
```
rlinehan: Schemas don't actually get validated unless you add `^:always-validate` to the definition (or manually use `schema/validate`). If you only want them validated during testing, you can forgo that annotation and add `(use-fixtures :once schema.test/validate-schemas)` to the test file where this function is tested. See https://github.com/Prismatic/schema#schemas-in-practice for more info.

test/unit/puppetlabs/services/file_serving/config/puppet_fileserver_config_core_test.clj
```diff
@@ -0,0 +1,21 @@
+(ns puppetlabs.services.file-serving.config.puppet-fileserver-config-core-test
+  (:require [clojure.test :refer :all]
+            [puppetlabs.services.file-serving.config.puppet-fileserver-config-core :refer :all]))
+
```
rlinehan: Probably should add

```clojure
(use-fixtures :once schema.test/validate-schemas)
```
here, to ensure that, at least during tests, the schemas are being validated.

src/clj/puppetlabs/services/file_serving/config/puppet_fileserver_config_core.clj
nwolfe: I believe this `find-mount` function *would* get validated if it was called from another function that was marked as `^:always-validate` though.

We haven't been too rigid about this in Puppet Server.  We mark the top-level functions as `^:always-validate` but don't necessarily mark the lower-level implementation functions as such.

rlinehan: Ah, good to know. Thanks @nwolfe for the clarification!

src/clj/puppetlabs/services/file_serving/config/puppet_fileserver_config_core.clj
```diff
@@ -0,0 +1,93 @@
...
+               key)
+        (recur (rest xs)
+               (assoc m x {})
+               x))
+      m)))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; Schemas
+
+(def Mount {:path schema/Str :acl [(schema/pair (schema/either (schema/eq :allow) (schema/eq :deny)) "acl" schema/Str "value")]})
```
rlinehan: I don't think this schema actually matches what it's used for, at least based on the test in https://github.com/masterzen/puppet-server/blob/feature/file_metadata/test/unit/puppetlabs/services/file_serving/config/puppet_fileserver_config_core_test.clj#L15-L17

From that  it looks like`find-mount` gets as its first argument

```clojure
 {:files2 {:acl [:allow "127.0.0.1/32" :deny "forbidden-certname.domain.com"]
           :path "/etc/puppet/files2"}
  :files1 {:acl [:allow "127.0.0.1/32" :allow "192.168.10.0/24" :allow "this-certname-only.domain.com"]
           :path "/etc/puppet/files1"}}
```
which should match the `Mounts` schema.

The part that is failing here is the value of `:acl`, for example:

```clojure
[:allow "127.0.0.1/32" :deny "forbidden-certname.domain.com"]
```

I think this schema actually requires that this be in the form 

```clojure
[[:allow "127.0.0.1/32"] [:deny "forbidden-certname.domain.com"]]
```

since `(schema/pair ...)` matches a sequencel, and in the `Mount` schema it is used within a vector.

src/clj/puppetlabs/services/file_serving/file_metadata/file_metadata_core.clj
```diff
@@ -0,0 +1,200 @@
...
+
+(def base-metadata
+  {
+    :data          {
+                     }
+    :document_type "FileMetadata"
+    :metadata      {
+                     :api_version 1
+                     }
+   })
```
rlinehan: Style/formatting nitpick: in most Clojure code I've seen/written, braces, brackets, and parentheses are never on their own line (so basically, follow https://github.com/bbatsov/clojure-style-guide#gather-trailing-parens for braces and brackets as well).

So, this would look like:
``` clojure
(def base-metadata
  {:data          {}
   :document_type "FileMetadata"
   :metadata      {:api_version 1}})
```

src/clj/puppetlabs/services/file_serving/file_metadata/file_metadata_core.clj
```diff
@@ -0,0 +1,200 @@
...
+                                                           [:representation
+                                                            :media-type]))
+                                  (assoc-in context
+                                            [:representation :media-type]
+                                            "text/pson")
+                                  context)]
+    (-> (cheshire/generate-string x)
+        (representation/as-response context-with-media-type)
+        (assoc :status 200)
+        (representation/ring-response))))
```
rlinehan: These PSON helper methods should be moved out into their own namespace, or maybe into `liberator-utils`, so they're not just copied from the CA code.

src/clj/puppetlabs/services/file_serving/file_metadata/file_metadata_core.clj
```diff
@@ -0,0 +1,200 @@
...
+      ))
+
+(defn- get-link-metadata
+  [file-path posix-attrs links]
+  (-> (assoc-in base-metadata [:data :checksum] (checksum-of-file file-path))
+      (get-base-metadata file-path posix-attrs)
+      (assoc-in [:data :relative_path] nil)
+      (assoc-in [:data :destination] (.getCanonicalPath (.toFile (posix-utils/resolve-link file-path))))
+      (assoc-in [:data :type] "link")
+      ))
```
rlinehan: Rather than doing `assoc-in` for each of the keys in `:data`, you could use `ks/deep-merge` or do something like.

```clojure
(update-in base-metadata :data #(merge % {:relative_path nil
                                          :destination :nil
                                          :type "file"}))
```

src/clj/puppetlabs/services/file_serving/config/puppet_fileserver_config_core.clj
```diff
@@ -0,0 +1,93 @@
...
+        (let [directive (-> s (subs 0 n) trim keyword)
+              argument (-> s (subs (inc n)) trim)]
+          (if (= :path directive)
+            {:path argument}
+            {:acl [directive argument]}))))))
+
+(defn- strip-comment
+  [s]
+  (let [n (.indexOf s (int \#))]
+    (if (and (not (neg? n)))
```
rlinehan: I think this could be `(if-not (neg? n))`

src/clj/puppetlabs/services/file_serving/file_metadata/file_metadata_core.clj
```diff
@@ -0,0 +1,200 @@
...
+(schema/defn get-metadata
+  [mountpath :- config/MountPath
+   links :- Link
+   source_permissions :- SourcePermission]
+  (let [[mount path] mountpath
+        file (clojure.java.io/file (:path mount) path)
+        file-path (.toPath file)
+        posix-attrs (posix-utils/unix-attributes file-path links)]
+
+    (assoc-in (cond
```
rlinehan: IMO it's not very readable having a `cond` within an `assoc-in`. I would favor breaking this out to a separate function, or, since `:links` is added no matter what type, adding `:links` to `get-base-metadata`.

src/clj/puppetlabs/services/file_serving/utils/posix_utils.clj
```diff
@@ -0,0 +1,69 @@
...
+(def GROUP_READ_MODE 0040)
+(def GROUP_WRITE_MODE 0020)
+(def GROUP_EXEC_MODE 0010)
+(def OTHERS_READ_MODE 0004)
+(def OTHERS_WRITE_MODE 0002)
+(def OTHERS_EXEC_MODE 0001)
+
+(defn- bit->mode
+  [acc bit]
+  (cond
```
rlinehan: You should be able to use `condp` here:

```clojure
  (condp = bit
    PosixFilePermission/OWNER_READ (bit-or acc OWNER_READ_MODE)
    ...
```

masterzen:
Hi @rlinehan 
I'm currently traveling abroad with little net access, so I won't be able to comment on your comments. But what I've read so far is exactly what I wanted, so many thanks. You'll certainly notice that my clojure got a bit better when you'll review trapperkeeper - authorization :-)

Thanks for the thorough review, I'll try to push a better version, including more content when I come back. 
Brice

cprice404:
@masterzen I promise that we didn't plan things out so that we'd finally have time to review all of your stuff as soon as you went on vacation!  :)

rlinehan:
@masterzen yay, I'm glad this is useful! I'm going to do a little more review of what's currently here (didn't quite finish up yesterday), and then I'll look forward to new stuff whenever you're able to get around to it.

src/clj/puppetlabs/services/file_serving/file_metadata/file_metadata_core.clj
```diff
@@ -0,0 +1,200 @@
...
+
+(schema/defn routes
+  [fileserving-config-service]
+  (compojure/context "/:environment" [environment]
+                     (compojure/routes
+                       (GET "/file_metadata/*" request
+                            (let [path (get-in request [:params :*])
+                                  links (get-in request [:params :links] "manage")
+                                  permissions (get-in request [:params :use_permissions] "use")]
+                              (file-metadata path fileserving-config-service links permissions))))))
```
rlinehan: With some of Compojure's magic, I think that you could destructure `request` as `{params :params`}, and then do e.g.
`links (get params :links "manage")`.

test/integration/puppetlabs/services/file_metadata/file_metadata_int_test.clj
```diff
@@ -0,0 +1,85 @@
...
+      deref
+      service-id))
+
+(defn wait-for-jrubies
+  [app num-jrubies]
+  (let [pool-context (-> (service-context app :JRubyPuppetService)
+                         :pool-context)]
+    (while (< (count (jruby-core/pool->vec pool-context))
+              num-jrubies)
+      (Thread/sleep 100))))
```
rlinehan: This should be pulled in from puppet_environments_int_test (or moved to a helper namespace), rather than copy-pasted.

test/integration/puppetlabs/services/file_metadata/file_metadata_int_test.clj
```diff
@@ -0,0 +1,85 @@
...
+(deftest ^:integration file-metadata-integration-test
+  (testing "get existing file metadata"
+    (fs/copy-dir (str test-resources-dir "/test") bootstrap/master-conf-dir)
+    (write-fileserver-conf (str "[test]\npath " bootstrap/master-conf-dir "/test\nallow *\n"))
+    (bootstrap/with-puppetserver-running app {:jruby-puppet
+                                              {:max-active-instances 1}}
+                                         ;; if we start making requests before we know that all of the
+                                         ;; jruby instances are ready, we won't be able to predict which
+                                         ;; instance is handling our request, so we need to wait for them.
+                                         (wait-for-jrubies app 1)
```
rlinehan: Since this is a pure clojure implementation of the file metadata endpoint, it doesn't seem like it should depend on the presence of any jruby instances, and thus this isn't needed?

peterhuene:
@masterzen Any update on the feedback to this PR?

masterzen:
@peterhuene sorry no, I haven't had the time to take all this into account :(
Maybe we can just  close this PR (since it won't ever been merged)?
Then I'll open a new one when I'll have the time to apply all the comments (and at the same time finish the code). I will definitely have some time by end of May.

rlinehan:
@masterzen Cool, I think we'll close this for now, and then yeah - feel free to reopen a new one whenever you've got some stuff for us to take a look at! :)

PR #349, opened 2015-01-01
(docs) Edit external SSL termination page for easier skimming

nfagerlund:


fpringvaldsen:
:+1:

documentation/external_ssl_termination.markdown
```diff
@@ -5,21 +5,61 @@ canonical: "/puppetserver/latest/external_ssl_termination.html"
...
+By default, reading this data from headers is disabled. To allow Puppet Server to recognize it, open `config.d/master.conf` and add  `allow-header-cert-info: true` to the `webserver` config block. See [Puppet Server Configuration](./configuration.markdown) for more information on the `master.conf` file.
 
 > **WARNING**: Setting `allow-header-cert-info` to 'true' puts Puppet Server in an incredibly vulnerable state. Take extra caution to ensure it is **absolutely not reachable** by an untrusted network.
+>
+> With `allow-header-cert-info` set to 'true', core Ruby Puppet application code will use only the client HTTP header values---not an SSL-layer client certificate---to determine the client subject name, authentication status, and trusted facts. This is true even if the web server is hosting an HTTPS connection. This applies to validation of the client via rules in the [auth.conf](https://docs.puppetlabs.com/guides/rest_auth_conf.html) file and any [trusted facts](https://docs.puppetlabs.com/puppet/latest/reference/lang_facts_and_builtin_vars.html#trusted-facts) extracted from certificate extensions.
```
nwolfe: You've got this trusted facts URL already defined as `[trusted]` below - might as well reuse it here

nwolfe:
:+1: other than minor comment

documentation/external_ssl_termination.markdown
```diff
@@ -5,21 +5,61 @@ canonical: "/puppetserver/latest/external_ssl_termination.html"
...
+## Allow Client Cert Data From HTTP Headers
+
+When using external SSL termination, Puppet Server expects to receive client certificate information via some HTTP headers.
 
-You'll need to reboot Puppet Server for the changes to take effect. The `ssl_client_header` and `ssl_client_verify_header` options in the `puppet.conf` file will then be enabled and work exactly as documented in Ruby Puppet. Please see the [Puppet Documentation](https://docs.puppetlabs.com/references/3.7.latest/configuration.html#sslclientheader) for more detail on these.
+By default, reading this data from headers is disabled. To allow Puppet Server to recognize it, open `config.d/master.conf` and add  `allow-header-cert-info: true` to the `webserver` config block. See [Puppet Server Configuration](./configuration.markdown) for more information on the `master.conf` file.
```
camlow325: It may be worth noting that the `master.conf` file isn't actually created at installation time for either for FOSS or PE for any current releases so the user would need to create the entire file at this point.  The config block section under which the `allow-header-cert-info` setting resides is named "master", not "webserver".  If no "master" config block and/or `allow-header-cert-info` setting is present in the configuration, Puppet Server will use a default value of "false" for `allow-header-cert-info`.

documentation/external_ssl_termination.markdown
nfagerlund: Ah, good call. I'll amend.

documentation/external_ssl_termination.markdown
```diff
@@ -5,21 +5,61 @@ canonical: "/puppetserver/latest/external_ssl_termination.html"
...
+
+## Configure SSL Terminating Proxy to Set HTTP Headers
+
+The device that terminates SSL for Puppet Server must extract information from the client's certificate and insert that information into three HTTP headers. See the documentation for your SSL terminator for details.
+
+The headers you'll need to set are `X-Client-Verify`, `X-Client-DN`, and `X-Client-Cert`.
+
+### `X-Client-Verify`
+
+Mandatory. Must be one of `NONE`, `SUCCESS`, `GENEROUS` or `FAILED:reason`. Puppet Server uses this to authorize requests.
```
camlow325: Do you know what the specific behaviors of `NONE`, `GENEROUS`, and `FAILED:reason` are and why/when someone would want to use them?  I hadn't seen those before and the documentation at https://docs.puppetlabs.com/references/3.7.latest/configuration.html#sslclientverifyheader only refers to the failures of 'SUCCESS' and *anything else* as "failure".

documentation/external_ssl_termination.markdown
```diff
@@ -5,21 +5,61 @@ canonical: "/puppetserver/latest/external_ssl_termination.html"
...
+
+You can change this header name with [the `ssl_client_verify_header` setting.](https://docs.puppetlabs.com/references/latest/configuration.html#sslclientverifyheader)
+
+This setting (and its twin, `ssl_client_header`) is a bit odd: its value should be the result of transforming the desired HTTP header name into a CGI-style environment variable name. That is, to change the HTTP header to `X-SSL-Client-Verify`, you would set the setting to `HTTP_X_SSL_CLIENT_VERIFY`. (Add `HTTP_` to the front, change hyphens to underscores, and uppercase everything.)
```
camlow325: Nice improvement.  Thanks!  :smiley:

documentation/external_ssl_termination.markdown
```diff
@@ -5,21 +5,61 @@ canonical: "/puppetserver/latest/external_ssl_termination.html"
...
+
+(Puppet Server will eventually UN-munge the CGI variable name to get a valid HTTP header name, and use that name to interact directly with an HTTP request. This is a legacy quirk to ensure that the setting works the same for both Puppet Server and a Rack Puppet master; note that Rack actually does use CGI environment variables.)
+
+### `X-Client-DN`
+
+Mandatory. Must be the [Subject DN][] of the agent's certificate, if a certificate was presented. Puppet Server uses this to authorize requests.
+
+You can change this header name with [the `ssl_client_header` setting.](https://docs.puppetlabs.com/references/latest/configuration.html#sslclientheader) See the note above for more info about this setting's expected values.
```
camlow325: We filed a bug around this recently, SERVER-213, which talks about Puppet Server's inability to decode the `X-Client-DN` when the client DN is in "OpenSSL" format - using slash-delimiters - as opposed to RFC-2253 - using comma-delimiters.  Might make sense to document that here so that others are aware of the outstanding problem?

documentation/external_ssl_termination.markdown
nfagerlund: Good catch and good catch. I'll add a note to the master.conf part of configuration.markdown, while I'm at it.

documentation/external_ssl_termination.markdown
```diff
@@ -5,21 +5,61 @@ canonical: "/puppetserver/latest/external_ssl_termination.html"
...
 
-With `allow-header-cert-info` set to 'true', core Ruby Puppet application code will use only the client HTTP header values---not an SSL-layer client certificate---to determine the client subject name, authentication status, and certificate. This is true even if the web server is hosting an HTTPS connection. This applies to validation of the client via rules in the [auth.conf](https://docs.puppetlabs.com/guides/rest_auth_conf.html) file and any [trusted facts](https://docs.puppetlabs.com/puppet/latest/reference/lang_facts_and_builtin_vars.html#trusted-facts) extracted from certificate extensions.
+The name of this header is not configurable.
 
-If the `client-auth` setting in the `webserver` config block is set to `need` or `want`, the Jetty web server will validate the client certificate against a certificate authority store. Only the SSL-layer client certificate---not a certificate in an  `X-Client-Cert` header---will be validated against the certificate authority store.
 
+[pem format]: https://docs.puppetlabs.com/background/ssl/cert_anatomy.html#pem-file
+[trusted]: https://docs.puppetlabs.com/puppet/latest/reference/lang_facts_and_builtin_vars.html#trusted-facts
```
camlow325: A couple of users have found that Apache mod_proxy will send the client certificate through with a "custom" encoding -- translating line feeds to spaces -- rather than the URL encode approach which is called for here.  We don't really have a suggested workaround for would-be Apache mod_proxy users of this functionality other than not to use it, but it may make sense to mention the issue and note the link to the related ticket - https://tickets.puppetlabs.com/browse/SERVER-217.

documentation/external_ssl_termination.markdown
nfagerlund: Agreed. Will do.

camlow325:
:+1: with some minor suggested additions around other known issues, etc.  This should be a ton more usable, thanks!

documentation/external_ssl_termination.markdown
nfagerlund: Word.

documentation/external_ssl_termination.markdown
nfagerlund: Uh... I think I knew this at one point!? It's not in the Apache docs, though. I found this: https://www.apachelounge.com/viewtopic.php?p=23573 -- which says `GENEROUS` is if a certificate was presented but somehow it's not `SUCCESS` or `FAILED`...???????

Um, I'm assuming we still fail anything other than SUCCESS.

nfagerlund: Well, I may as well bounce most of that if we're not really using it; I know those are the values that will come out of Apache's mod_ssl, but it sounds like the consumer end is particular about wanting SUCCESS.

nfagerlund: See also https://issues.apache.org/bugzilla/show_bug.cgi?id=45054#c5, ok done googling for this now

nfagerlund:
That should catch all of 'em!

documentation/external_ssl_termination.markdown
camlow325: Yeah, in the Puppet Server code we only look for SUCCESS vs. *something else*.  See:

https://github.com/puppetlabs/puppet-server/blob/master/src/clj/puppetlabs/services/request_handler/request_handler_core.clj#L111.

In the Puppet Rack handler (e.g., when running the Ruby master under Passenger), I believe the same was true:

https://github.com/puppetlabs/puppet/blob/master/lib/puppet/network/http/rack/rest.rb#L121.

Yeah, I'd be :+1: on bouncing the other specific non-SUCCESS values if we don't have a specific definition for how they would be used.

camlow325:
:+1: @dankreek, did you want to take a look at this before we merge?

documentation/configuration.markdown
```diff
@@ -106,7 +106,15 @@ puppet-admin: {
...
 ### `master.conf`
 
-This file contains settings for the Puppet master functionality of Puppet Server. If `allow-header-cert-info` is set to 'true', this allows the `ssl_client_header` and `ssl_client_verify_header` options in puppet.conf to work. By default, this is set to 'false'.
+This file contains settings for Puppet master features, such as node identification and authorization.
+
+In a default installation, this file doesn't exist. You'll need to create it if you want to set non-default values for these settings.
+
+* `allow-header-cert-info` determines whether Puppet Server should use authorization info from the `X-Client_Verify`, `X-Client-CN`, and `X-Client-Cert` HTTP headers. Defaults to `false`.
+
+    This setting is used to enable [external SSL termination.](./external_ssl_termination.markdown) If enabled, Puppet Server will ignore the actual certificate presented to the Jetty webserver, and will rely completely on header data to authorize requests. This is very dangerous unless you've secured your network to prevent any untrusted access to Puppet Server.
```
dankreek: Should we somehow mention that when using HTTPS the certificate will be ignored, but generally HTTP is used between the external terminator and Puppet Server and there is no cert to speak of.

documentation/external_ssl_termination.markdown
```diff
@@ -5,21 +5,65 @@ canonical: "/puppetserver/latest/external_ssl_termination.html"
...
+
+You can change this header name with [the `ssl_client_verify_header` setting.](https://docs.puppetlabs.com/references/latest/configuration.html#sslclientverifyheader)
+
+This setting (and its twin, `ssl_client_header`) is a bit odd: its value should be the result of transforming the desired HTTP header name into a CGI-style environment variable name. That is, to change the HTTP header to `X-SSL-Client-Verify`, you would set the setting to `HTTP_X_SSL_CLIENT_VERIFY`. (Add `HTTP_` to the front, change hyphens to underscores, and uppercase everything.)
+
+(Puppet Server will eventually UN-munge the CGI variable name to get a valid HTTP header name, and use that name to interact directly with an HTTP request. This is a legacy quirk to ensure that the setting works the same for both Puppet Server and a Rack Puppet master; note that Rack actually does use CGI environment variables.)
```
dankreek: Awesome! Thanks for documenting this, @nfagerlund

dankreek:
Just had the one little question, but everything looks great! 

Thanks again, @nfagerlund!

documentation/configuration.markdown
nfagerlund: Eh... nah. I think it's implicit here, and we go into enough extra detail on the guide to setting up external SSL.

nfagerlund: oh but oops just spotted a typo in there. Hold on.

nfagerlund:
OK fixed typo.

PR #350, opened 2015-01-03
(MAINT) Add sqa-utils to Gemfile.

waynr:


Gemfile
```diff
@@ -4,5 +4,8 @@ group :test do
   gem 'rake'
   gem 'rspec'
   gem 'beaker', '~>1.17.0'
+  if ENV['GEM_SOURCE']
```
cprice404: Where's this get set?  and/or what's up with the name of it?

Gemfile.lock
```diff
@@ -1,5 +1,5 @@
 GEM
-  remote: https://rubygems.org/
+  remote: http://rubygems.delivery.puppetlabs.net/
```
cprice404: Wouldn't this prevent anyone who was not on our VPN from running any of our rake tasks?

waynr: Erh, yeah I can undo this part of the commit.

Gemfile
waynr: `GEM_SOURCE` is the environment variable we use to point bundler to `rubygems.delivery.puppetlabs.net` in CI.

cprice404: Seems like kind of an overloaded use of the variable here, then.  Would be nice if we had something more clear, like "if PL_INTERNAL" or something.  GEM_SOURCE could be set by users outside of PL, right?  If we are going to condition based on that var it seems like maybe we should check the value too, and see if it's pointing to our internal gem server.

I don't care too much, won't put up a huge fight, but this doesn't seem like quite the right thing to condition on.

Gemfile
```diff
@@ -4,5 +4,8 @@ group :test do
   gem 'rake'
   gem 'rspec'
   gem 'beaker', '~>1.17.0'
+  if ENV['GEM_SOURCE'] =~ /rubygems\.delivery\.puppetlabs\.net/
```
waynr: @cprice404 
What do you think of this? I orignally just tested `GEM_SOURCE` being set since that is how other projects I've observed do it. I'm hesitant to add an otherwise meaningless variable but I think this addresses both concerns.

cprice404: I like that better :)  Not sure the regex is necessary, and if I was going to use one, I'd make it more strict than this (using a leading ^ and trailing $), but it addresses the main concern that I had.

cprice404: > Not sure the regex is necessary,

As opposed to just a string equality check, I mean.

waynr: I didn't go with string equality because I thought it would be vaguely possible that in the future for some reason our internal rubygem mirror requires `https`, the regex matches both https://rubygems.delivery.puppetlabs.net and http://rubygems.delivery.puppetlabs.net

cprice404: Then you could do:

`%r{^https?://rubygems\.delivery\.puppetlabs\.net$}`

waynr: That's true, but I think Bundler would choke before it reaches this point if the protocol portion of the URI was missing or if there were extra characters at the end.

cprice404: That doesn't mean that it's a bad idea to make your code as explicit as possible, so that the next person who reads it knows exactly what it's trying to do :)

PR #351, opened 2015-01-05
(MAINT) Fix intermittent gem install failures.

waynr:
* Change list of installed gems to reduce complexity of install. Beaker changes
  too often to use it here without specifying a concrete version, which in turn
  would require more complex handing of this list of gems throughout the rest of
  the test.
* Clears the list of rubygems sources before adding the internal mirror to
  ensure that gem installation problems are not masked by network failures.


nwolfe:
:+1:

camlow325:
:+1:

PR #352, opened 2015-01-05
(SERVER-268) Improve Acceptance README

fpringvaldsen:
Improve the README in the Acceptance folder to reflect that
setting the BEAKER_TYPE and BEAKER_LOADPATH environment variables
are required when running the tests with a Rakefile under
Workflow a.

I also went ahead and updated the PACKAGE_BUILD_VERSION to reflect a more recent snapshot build of puppet server, and updated the BEAKER_CONFIG variable to point to an actual config file in the acceptance folder, as the one previously being pointed to no longer exists.

This only addresses Rakefile Workflow a, as that's how I was running the acceptance tests when I ran into the issues with the README. Other parts of the README are likely outdated as well, so this will probably require some further investigation.

### DIFF
--- a/acceptance/README.md
+++ b/acceptance/README.md
@@ -201,9 +201,11 @@ This is the expected workflow for running vcloud tests directly from one's
 laptop.
 
     bundle install --path vendor/bundle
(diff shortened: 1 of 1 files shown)


nwolfe:
:+1:

PR #353, opened 2015-01-06
(SERVER-247) Add jira-ruby to Gemfile

jeffmccune:
Without this patch the `pl:tickets` rake task fails because the jira-ruby
dependency is missing.  This patch addresses the problem by adding jira-ruby to
the project Gemfile.  The dependency will be available following `bundle
install`.

The library is in the development group to allow the dependency to be filtered
using `bundle install --without development`.

This patch also moves rake out of the test group because the rake dependency is
always required when running rake tasks for development and testing.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,11 +1,12 @@
 source ENV['GEM_SOURCE'] || "https://rubygems.org"
 
+gem 'rake', :group => [:development, :test]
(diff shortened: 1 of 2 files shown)


fpringvaldsen:
:+1:

nwolfe:
:+1:

PR #354, opened 2015-01-07
(SERVER-297) Consolidate subcommand environment handling

jeffmccune:
Without this patch each of the `gem`, `ruby`, and `irb` subcommands handle
environment variables in different ways.  This is a problem because the project
as a whole strives to manage environment variables such as GEM_HOME and
GEM_PATH as robustly as possible.  It is also problematic to manage additional
environment variables such as JARS_NO_REQUIRE because the variable must be
added to each subcommand's implementation.

This patch addresses the problem by moving the gem/cli-gem-environment function
into subcommand/environment.  The function signature remains the same.  The
behavior changes slightly to explicitly manage all of the environment variables
puppetserver cares about from a robustness perspective.  These are GEM_HOME,
GEM_PATH, RUBYLIB, RUBYOPT, and RUBY_OPTS.

This change also provides a single location to manage additional variables such
as JARS_NO_REQUIRE for use with all subcommands.


acceptance/suites/tests/010-puppetserver-cli/subcommand/irb.rb
```diff
@@ -0,0 +1,40 @@
...
+  on(master, cmd) do
+    assert(/^OK$/.match(stdout), "#{var} is not being preserved")
+    assert_no_match(/BAD/, stdout)
+  end
+end
+
+step "Check that puppet is loadable"
+cmd = "echo 'puts Puppet.version' | #{cli} irb -- -rpuppet -f"
+on(master, cmd) do
+  assert_no_match(/LoadError/, stdout)
```
dankreek: Would it make sense to assert for a regex that matched a Puppet version string?

acceptance/suites/tests/010-puppetserver-cli/subcommand/ruby.rb
```diff
@@ -0,0 +1,60 @@
...
+on(master, "#{gem_install} hiera-eyaml")
+
+step "Check that ruby subcommand can load eyaml from hiera-eyaml"
+on(master, "#{cli} ruby -- -rrubygems -reyaml -e 'puts %{OK}'") do
+  assert(/^OK$/.match(stdout), "ruby could not load eyaml after gem install")
+end
+
+%w(GEM_PATH RUBYLIB RUBYOPT RUBY_OPTS).each do |var|
+  step "Check that #{var} is cleared"
+  on(master, "#{var}=DOOH #{cli} ruby -- -e 'puts ENV[%{#{var}}] || %{OK}'") do
```
dankreek: I believe that's spelled "D'OH"

acceptance/suites/tests/010-puppetserver-cli/subcommand/ruby.rb
```diff
@@ -0,0 +1,60 @@
...
+cmd = "GEM_HOME=UHOH #{cli} ruby -- -ryaml -e 'puts ENV.to_hash.to_yaml'"
+on(master, cmd) do
+  assert_match(/GEM_HOME/, stdout)
+  assert_no_match(/UHOH/, stdout)
+end
+
+step "Check that puppet is loadable"
+cmd = "#{cli} ruby -- -rpuppet -e 'puts Puppet.version'"
+on(master, cmd) do
+  assert_no_match(/LoadError/, stdout)
```
dankreek: same comment as above

acceptance/suites/tests/010-puppetserver-cli/subcommand/irb.rb
jeffmccune: I thought about that, but couldn't quickly come up with one that I was confident would always match.  Do you think `/\..*\./` is sufficient?

jeffmccune: Or maybe just puts something known good that also exercises Puppet, like: `puts %{GOOD: #{Puppet.version}}` then assert on `GOOD:`.

dankreek:
It looks like 33cf447 isn't actually related to SERVER-252. We should probably either create a new ticket or just label the commit as (MAINT).

acceptance/suites/tests/010-puppetserver-cli/subcommand/ruby.rb
jeffmccune: `D\'OH`

dankreek: Hahaa (don't actually submit a patch, it's fine as it is)

jeffmccune:
> It looks like 33cf447 isn't actually related to SERVER-252. We should probably either create a new ticket or just label the commit as (MAINT).

Fixed.

Also, added a positive assertion for irb and ruby exercising puppet.

src/clj/puppetlabs/puppetserver/cli/subcommand.clj
```diff
@@ -1,4 +1,5 @@
 (ns puppetlabs.puppetserver.cli.subcommand
+  (:import (org.jruby.embed ScriptingContainer))
```
nwolfe: This isn't used

dankreek: (this is a violation of Wolfe's 2nd principle)

dankreek:
Other than @nwolfe's comment about the unused import, this LGTM :+1:

src/clj/puppetlabs/puppetserver/cli/subcommand.clj
jeffmccune: What's the first principle?

jeffmccune:
> Other than @nwolfe's comment about the unused import, this LGTM :+1: 

Cool, unused import has been removed.

src/clj/puppetlabs/puppetserver/cli/subcommand.clj
nwolfe: There should be no log output during a test run :smile:

acceptance/suites/tests/010-puppetserver-cli/subcommand/irb.rb
```diff
@@ -0,0 +1,41 @@
...
+if master.is_pe?
+  cli = "pe-puppetserver"
+else
+  cli = "puppetserver"
+end
+# --------------------------------------------------------------------------
+
+%w(GEM_PATH RUBYLIB RUBYOPT RUBY_OPTS).each do |var|
+  step "Check that #{var} is cleared"
+  cmd = "echo 'puts ENV[%{#{var}] || %{OK}' | #{var}=BAD #{cli} irb -- -f"
```
camlow325: Is this missing an extra *}* before the *]* in *ENV[%{#{var}]*?

acceptance/suites/tests/010-puppetserver-cli/subcommand/irb.rb
```diff
@@ -0,0 +1,41 @@
...
+
+test_name "Puppetserver 'irb' subcommand tests."
+
+# --------------------------------------------------------------------------
+# This behavior needs to be baked into Beaker but is not yet. SERVER-79 tracks
+# the need to fix this once Beaker has been modified to make the paths to these
+# commands available.
+#
+if master.is_pe?
+  cli = "pe-puppetserver"
```
camlow325: Is this right for PE?  I just installed a recent PE drop and saw that "/opt/puppet/bin/puppetserver" and not "/opt/puppet/bin/pe-puppetserver" was laid down as part of the pe-puppetserver package.

KevinCorcoran: @camlow325 - I think you have it correct.

src/clj/puppetlabs/puppetserver/cli/subcommand.clj
```diff
@@ -15,6 +15,26 @@
...
+  subcommands.  Environment variables puppetserver needs are enforced while
+  all other variables are preserved.  This function manages GEM_HOME,
+  GEM_PATH, RUBYOPT, RUBY_OPTS, and RUBYLIB.
+
+  If no initial-env is provided then System/getenv is used.
+
+  The underlying principle is that a process should preserve as much of the
+  environment as possible in an effort to get out of the end users way.  This
+  includes PATH which should be fully managed elsewhere, e.g. the service
+  management framework."
```
camlow325: I think if we're going to go with this approach, we should consider doing it for the production code and not just the subcommands.  That way running the subcommands will get you more consistent results with what you'd get if you ran the full production stack.

PATH is an interesting one.  If we presume that the PATH variable is setup such that some system Ruby scripts suitable for running under MRI would be picked up but some JRuby equivalents would be more appropriate, I've been wondering if we'd need to account for that with a special new TK config setting like we do for GEM_HOME to make sure the environments don't mix.  I'm okay with just preserving it in the context of the subcommands for now but we may want to think through this more later. 

Would like to get @cprice404's opinion on this whole approach.

acceptance/suites/tests/010-puppetserver-cli/subcommand/ruby.rb
```diff
@@ -0,0 +1,61 @@
...
+require 'puppetserver/acceptance/gem_utils'
+
+test_name "Puppetserver 'ruby' subcommand tests."
+
+# --------------------------------------------------------------------------
+# This behavior needs to be baked into Beaker but is not yet. SERVER-79 tracks
+# the need to fix this once Beaker has been modified to make the paths to these
+# commands available.
+#
+if master.is_pe?
```
camlow325: Same comment here as for the `irb` subcommand tests.

acceptance/suites/tests/010-puppetserver-cli/subcommand/irb.rb
camlow325: Did it work for you to use "irb -- -f"?  I haven't needed to use the extra "--" when running via the "puppetserver" script, at least manually from a Unix command shell (not Beaker).  "irb -f" worked for me.

acceptance/suites/tests/010-puppetserver-cli/subcommand/ruby.rb
```diff
@@ -0,0 +1,61 @@
...
+gem_install = "#{cli} gem install "
+if additional_gem_source
+  gem_install += " --clear-sources --source #{additional_gem_source}"
+end
+
+step "Install hiera-eyaml gem."
+on(master, "#{gem_install} hiera-eyaml")
+
+step "Check that ruby subcommand can load eyaml from hiera-eyaml"
+on(master, "#{cli} ruby -- -rrubygems -reyaml -e 'puts %{OK}'") do
```
camlow325: Did this work for you?  When running from the command line, I had to omit the "--" after ruby and use "-rhiera/backend/eyaml" instead of "-eyaml" in order to avoid a "LoadError".  Maybe would also be good to `assert_no_match` for an "error" here, too?

acceptance/suites/tests/010-puppetserver-cli/subcommand/ruby.rb
```diff
@@ -0,0 +1,61 @@
...
+
+%w(FOO_DEBUG).each do |var|
+  step "Check that #{var} is preserved"
+  on(master, "#{var}=OK #{cli} ruby -- -e 'puts ENV[%{#{var}}] || %{BAD}'") do
+    assert(/^OK$/.match(stdout), "#{var} is not being preserved")
+    assert_no_match(/BAD/, stdout)
+  end
+end
+
+step "Check that GEM_HOME is managed"
```
camlow325: Should we have this one in the `irb` spec tests as well?

camlow325:
:chart_with_upwards_trend: This is looking really good.  Just left a few comments on problems I ran into trying to duplicate the result of the tests with local commands issued on one of my VMs.  Also, interested in hashing out the whole discussion on clearing/overriding/preserving of different environment variables.  You have some great points there.  Wanted to get some other opinions to see where others are on that.

cprice404:
out of curiosity--might we want to target this at stable?

camlow325:
Seems reasonable to me if @erikPrime is onboard.

cprice404:
@erikPrime what say you?  This PR is basically just tests, so it'd just improve our coverage for 1.0.1/3.7.2.

camlow325:
@cprice404 There's more than just tests here as the PR stands, though - specifically, the bit on how environment variables are passed through to the JRuby container via the subcommands.

cprice404:
well, crap.  is there a way to decouple those two changesets, then?  if not, maybe this needs to stay on master, I dunno.

nwolfe:
Yeah, maybe we should just pull that commit out of this PR and reconsider our environment support both for the CLI and the production scripting containers as a separate topic?

camlow325:
I'd be good with separating them, too, although I think the tests would need to be rewritten a bit to work for the code as currently implemented.

jeffmccune:
The tests are tied to the change in behavior in that I didn't want to spend time and effort testing something that isn't behaving well.  My vote is to keep these two commits together and just merge them into master or into stable together.

I'm not sure about the value of adding tests to stable when the behavior of the stuff being tested is or will be significantly different in master.  Adding only tests to stable is basically testing that stable misbehaves as expected.

jeffmccune:
@camlow325 @nwolfe and I chatted IRL and decided to open another pull request against stable that includes only the tests, no changes in behavior.  I'll leave this pull request open and rebase it to accommodate the new PR.

jeffmccune:
OK, rebased onto a fake master that assumes #362 is merged into stable, then stable merged up into master.

This will likely need a re-base once #362 is merged in and up.

This PR doesn't have anything to do with SERVER-252 at this point, it's really just about the behavior change around how we handle environment variables.

acceptance/suites/tests/010-puppetserver-cli/subcommand/irb.rb
```diff
@@ -0,0 +1,48 @@
...
+step "Check that GEM_HOME is managed"
+cmd = "echo 'puts ENV.to_hash.to_yaml' | GEM_HOME=UHOH #{cli} irb -- -ryaml -f"
+on(master, cmd) do
+  assert_match(/GEM_HOME/, stdout)
+  assert_no_match(/UHOH/, stdout)
+end
+
+%w(GEM_PATH RUBYLIB RUBYOPT RUBY_OPTS).each do |var|
+  step "Check that #{var} is cleared"
+  cmd = "echo 'puts ENV[%{#{var}] || %{OK}' | #{var}=BAD #{cli} irb -- -f"
```
camlow325: Is this missing a '}' before the end of the ENV reference - `ENV[%{#{var}}]`?

jeffmccune: Yes it is.  I'll fixup in #362

jeffmccune:
This pull request is a bit of a mess and should probably be ignored until #362 is merged into stable and up into master.  At that point cec3381 should be rebased onto master.

camlow325:
Yep, that makes sense.  I think it also would be good to cover the environment rework stuff under a new JIRA ticket and expand the scope to include reworking the production `ScriptingContainer` code to be consistent with that of the subcommand tools.  Sound okay?  If so, @jeffmccune, could you file that ticket?

jeffmccune:
@camlow325 Filed as [SERVER-297](https://jira.puppetlabs.com/browse/SERVER-297)

KevinCorcoran:
@jeffmccune - This can't be merged.

cprice404:
@KevinCorcoran we need to regroup on goals / implementation path before we bother rebasing this.

cprice404:
@jeffmccune is there a compelling reason not to close this until we've had our round-up meeting?  Trying to clear out our PR backlog.

PR #355, opened 2015-01-07
(SERVER-220) Include Jetty dependency in uberjar

nwolfe:
This should fix the `lein uberjar` command to produce a working Puppet
Server uberjar that includes the Jetty dependency, without interfering
with the ezbake packaging process.


jeffmccune:
:+1:

camlow325:
:+1:

PR #360, opened 2015-01-09
DOC-1488: Edit subcommand.markdown

jbondpdx:
Edited the new subcommand page for Puppet Server.

### DIFF
--- a/documentation/subcommands.markdown
+++ b/documentation/subcommands.markdown
@@ -28,7 +28,7 @@ When running from source, the format is:
 lein <subcommand> -c /path/to/puppet-server.conf [--] [<args>]
 ~~~
 
-Note that if you are running from source, you need to separate flag arguments (such as `--version` or -`e`) with `--`, as shown above. Otherwise, those arguments will be applied to Leiningen instead of to Puppet Server. This isn't necessary when running from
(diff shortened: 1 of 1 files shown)


nwolfe:
:+1:

documentation/subcommands.markdown
```diff
@@ -19,98 +18,95 @@ The following subcommands are provided:
...
+~~~sh
 lein <subcommand> -c /path/to/puppet-server.conf [--] [<args>]
-```
+~~~
 
-Note that if one of the `<args>` begins with a `-` (like `--version` or `-e`)
-then the intermediate `--` above is required for the argument to be applied to
-the subcommand and not leiningen. This is not necessary when running from
-packages (i.e. `puppetserver <subcommand>`).
+Note that if you are running from source, you need to separate flag arguments (such as `--version` or -`e`) with `--`, as shown above. Otherwise, those arguments will be applied to Leiningen instead of to Puppet Server. This isn't necessary when running from
```
camlow325: Cosmetic - seems like the "-" before `e` should be inside of the accent characters - like `-e`.

camlow325:
:+1: with just the one cosmetic comment.

documentation/subcommands.markdown
jbondpdx: Good catch, thank you! Fixed.

PR #362, opened 2015-01-12
(SERVER-252) Add acceptance tests for irb and ruby subcommands

jeffmccune:
Without this patch there are no acceptance level tests for the ruby and irb
subcommands as they relate to the `gem` subcommand.  This is a problem because
we need to ensure these commands behave as expected, particularly when the
`gem` subcommand is used to install third party libraries.

This patch addresses the problem by installing the `hiera-eyaml` gem and then
checking that the `ruby` and `irb` subcommand can `require 'eyaml'`.

This patch also adds basic testing of the behavior of GEM_HOME and FOO_DEBUG
with regard to the irb and ruby subcommands in addition to ensuring `require
'puppet'` works in both subcommands.

### DIFF
--- a/acceptance/suites/tests/010-puppetserver-cli/subcommand/gem.rb
+++ b/acceptance/suites/tests/010-puppetserver-cli/subcommand/gem.rb
@@ -8,11 +8,7 @@ test_name "Puppetserver 'gem' subcommand tests."
 # the need to fix this once Beaker has been modified to make the paths to these
 # commands available.
 #
-if master.is_pe?
-  cli = "pe-puppetserver"
-else
-  cli = "puppetserver"
-end
+cli = "puppetserver"
 # --------------------------------------------------------------------------
 
 # define gems to test
@@ -84,4 +80,21 @@ on(master, gem_env) do
   assert_no_match(/ERROR:  While executing gem/, stdout, "gem env blew up!")
   assert_match(/SHELL PATH:/, stdout, "PATH expected but not present")
   assert_match(/INSTALLATION DIRECTORY:/, stdout, "GEM_HOME not present")
+end
+
+step "Install hiera-eyaml gem."
+on(master, "#{gem_install} hiera-eyaml --no-ri --no-rdoc")
+
+step "Check that ruby subcommand can load hiera/backend/eyaml from hiera-eyaml"
+on(master, "#{cli} ruby -rrubygems -rhiera/backend/eyaml -e 'puts %{OK}'") do
+  assert_match(/^OK$/, stdout, "ruby could not load eyaml after gem install")
+  assert_no_match(/Error/i, stdout, "error loading hiera-eyaml library")
+end
+
+step "Check that irb subcommand can load hiera/backend/eyaml from hiera-eyaml"
+cmd = "echo 'puts Hiera::Backend::Eyaml::DESCRIPTION'" \
+      "| #{cli} irb -rrubygems -rhiera/backend/eyaml"
+on(master, cmd) do
+  assert_match(/OpenSSL/i, stdout, "hiera-eyaml description does not match")
+  assert_no_match(/Error/i, stdout, "error loading hiera-eyaml library")
 end
--- a/acceptance/suites/tests/010-puppetserver-cli/subcommand/irb.rb
+++ b/acceptance/suites/tests/010-puppetserver-cli/subcommand/irb.rb
@@ -1,0 +1,32 @@
+require 'test/unit/assertions'
+
+test_name "Puppetserver 'irb' subcommand tests."
+
+# --------------------------------------------------------------------------
+# This behavior needs to be baked into Beaker but is not yet. SERVER-79 tracks
+# the need to fix this once Beaker has been modified to make the paths to these
+# commands available.
+#
+cli = "puppetserver"
+# --------------------------------------------------------------------------
+
+step "Check that GEM_HOME is managed"
+cmd = "echo 'puts ENV.to_hash.to_yaml' | GEM_HOME=UHOH #{cli} irb -ryaml -f"
+on(master, cmd) do
+  assert_match(/GEM_HOME/, stdout)
+  assert_no_match(/UHOH/, stdout)
+end
+
+step "Check that FOO_DEBUG is preserved"
+cmd = "echo 'puts ENV[%{FOO_DEBUG}] || %{BAD}' | FOO_DEBUG=OK #{cli} irb -f"
+on(master, cmd) do
+  assert_match(/^OK$/, stdout, "FOO_DEBUG is not being preserved")
+  assert_no_match(/BAD/, stdout)
+end
+
+step "Check that puppet is loadable"
+cmd = "echo 'puts %{GOOD: #{Puppet.version}}' | #{cli} irb -rpuppet -f"
+on(master, cmd) do
+  assert_match(/GOOD:/, stdout)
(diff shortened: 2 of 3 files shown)


acceptance/suites/tests/010-puppetserver-cli/subcommand/irb.rb
```diff
@@ -0,0 +1,39 @@
...
+require 'puppetserver/acceptance/gem_utils'
+
+test_name "Puppetserver 'irb' subcommand tests."
+
+# --------------------------------------------------------------------------
+# This behavior needs to be baked into Beaker but is not yet. SERVER-79 tracks
+# the need to fix this once Beaker has been modified to make the paths to these
+# commands available.
+#
+if master.is_pe?
```
jeffmccune: Comment from #354 

> Is this right for PE? I just installed a recent PE drop and saw that "/opt/puppet/bin/puppetserver" and not "/opt/puppet/bin/pe-puppetserver" was laid down as part of the pe-puppetserver package.

> @camlow325 you have it right.

jeffmccune:
Addressed @camlow325's comment about `pe-puppetserver` not being a thing.

acceptance/suites/tests/010-puppetserver-cli/subcommand/gem.rb
```diff
@@ -85,3 +81,20 @@
   assert_match(/SHELL PATH:/, stdout, "PATH expected but not present")
   assert_match(/INSTALLATION DIRECTORY:/, stdout, "GEM_HOME not present")
 end
+
+step "Install hiera-eyaml gem."
+on(master, "#{gem_install} hiera-eyaml --no-ri --no-rdoc")
+
+step "Check that ruby subcommand can load hiera/backend/eyaml from hiera-eyaml"
+on(master, "#{cli} ruby -- -rrubygems -rhiera/backend/eyaml -e 'puts %{OK}'") do
```
camlow325: Is the "--" needed here?  It works from the command line when I run without it but fails if I run with it.

acceptance/suites/tests/010-puppetserver-cli/subcommand/gem.rb
```diff
@@ -85,3 +81,20 @@
...
+  assert_match(/^OK$/, stdout, "ruby could not load eyaml after gem install")
+  assert_no_match(/Error/i, stdout, "error loading hiera-eyaml library")
+end
+
+step "Check that irb subcommand can load hiera/backend/eyaml from hiera-eyaml"
+cmd = "echo 'puts Hiera::Backend::Eyaml::DESCRIPTION'" \
+      "| #{cli} irb -- -rrubygems -rhiera/backend/eyaml"
```
camlow325: Same question here on the "--".

acceptance/suites/tests/010-puppetserver-cli/subcommand/gem.rb
```diff
@@ -85,3 +81,20 @@
...
+step "Install hiera-eyaml gem."
+on(master, "#{gem_install} hiera-eyaml --no-ri --no-rdoc")
+
+step "Check that ruby subcommand can load hiera/backend/eyaml from hiera-eyaml"
+on(master, "#{cli} ruby -- -rrubygems -rhiera/backend/eyaml -e 'puts %{OK}'") do
+  assert_match(/^OK$/, stdout, "ruby could not load eyaml after gem install")
+  assert_no_match(/Error/i, stdout, "error loading hiera-eyaml library")
+end
+
+step "Check that irb subcommand can load hiera/backend/eyaml from hiera-eyaml"
```
camlow325: Since these tests are about the "gem" subcommand, I'd be good with just a follow-up "ruby" or "irb" to verify the "gem" install.  Not sure we'd need both.  That said, though, I'm good with adding them both.

acceptance/suites/tests/010-puppetserver-cli/subcommand/gem.rb
jeffmccune: It's not, no, `--` is only required when dealing with the `lein` commands, not `puppetserver`.  I'll remove them.

acceptance/suites/tests/010-puppetserver-cli/subcommand/irb.rb
```diff
@@ -0,0 +1,35 @@
...
+# --------------------------------------------------------------------------
+
+step "Check that GEM_HOME is managed"
+cmd = "echo 'puts ENV.to_hash.to_yaml' | GEM_HOME=UHOH #{cli} irb -- -ryaml -f"
+on(master, cmd) do
+  assert_match(/GEM_HOME/, stdout)
+  assert_no_match(/UHOH/, stdout)
+end
+
+%w(FOO_DEBUG).each do |var|
```
camlow325: Suppose an array isn't needed here if there's only one?

acceptance/suites/tests/010-puppetserver-cli/subcommand/gem.rb
jeffmccune: Yeah, I'd normally be fine with a single follow up too.  I included both because we've run into a situation where `gem` behaved differently from the others.  If they're too slow or problematic for some reason we could nuke one of them.

acceptance/suites/tests/010-puppetserver-cli/subcommand/irb.rb
jeffmccune: Yep, I've refactored the array out.

camlow325:
:+1:

acceptance/suites/tests/010-puppetserver-cli/subcommand/irb.rb
```diff
@@ -0,0 +1,33 @@
+require 'test/unit/assertions'
+require 'puppetserver/acceptance/gem_utils'
```
nwolfe: Don't think we need this do we?

acceptance/suites/tests/010-puppetserver-cli/subcommand/ruby.rb
```diff
@@ -0,0 +1,32 @@
+require 'test/unit/assertions'
+require 'puppetserver/acceptance/gem_utils'
```
nwolfe: Same here

nwolfe:
:+1: other than `gem_utils`

acceptance/suites/tests/010-puppetserver-cli/subcommand/irb.rb
jeffmccune: Nope, that file only defines `get_gem_list`. Good catch, removing it.

jeffmccune:
`gem_utils` has been fixed

nwolfe:
:+1:

cprice404:
Looks like the related PR has been merged into stable, but not merged up to master yet.  I took a quick swing at the merge-up and there are conflicts, so I'm punting it for now.  Happy to try again if I get back around to this before anyone else does.

jeffmccune:
@cprice404 This PR was "special" in that we weren't sure what the merge up process was so we decided to cherry-pick the commits for SERVER-262, which affected the subcommand behavior from stable to master.  This is likely the cause of your merge conflicts, at which point I'm happy to help.  It should just be a matter of making sure JARS_NO_REQUIRE is handled as expected in stable (not present) and master (set to "false") and the acceptance tests in each branch reflect that expected behavior.

PR #363, opened 2015-01-14
(SERVER-252) Fix bad irb and ruby subcommand tests

jeffmccune:
Without this patch the tests are failing because of bad logic in the tests
themselves.  This patch addresses the problem by correcting the logic in the
tests, specifically /BAD/ was matching in the irb tests because irb echos input
to standard output.  /^BAD$/ will not match the echoed input.  The ruby test
was failing because string interpolation was happening in the beaker process
when it was intended to be passed through beaker to the ruby subcommand being
tested.  This patch addresses the problem by avoiding string interpolation
entirely and using concatenation instead.

### DIFF
--- a/acceptance/suites/tests/010-puppetserver-cli/subcommand/ruby.rb
+++ b/acceptance/suites/tests/010-puppetserver-cli/subcommand/ruby.rb
@@ -20,11 +20,11 @@ end
 step "Check that FOO_DEBUG is preserved"
 on(master, "FOO_DEBUG=OK #{cli} ruby -e 'puts ENV[%{FOO_DEBUG}] || %{BAD}'") do
   assert_match(/^OK$/, stdout, "FOO_DEBUG is not being preserved")
-  assert_no_match(/BAD/, stdout)
+  assert_no_match(/BAD/, stdout, "FOO_DEBUG is being unset, it should not be")
 end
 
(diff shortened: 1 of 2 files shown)


acceptance/suites/tests/010-puppetserver-cli/subcommand/ruby.rb
```diff
@@ -20,11 +20,11 @@
 step "Check that FOO_DEBUG is preserved"
 on(master, "FOO_DEBUG=OK #{cli} ruby -e 'puts ENV[%{FOO_DEBUG}] || %{BAD}'") do
   assert_match(/^OK$/, stdout, "FOO_DEBUG is not being preserved")
-  assert_no_match(/BAD/, stdout)
+  assert_no_match(/BAD/, stdout, "FOO_DEBUG is being unset, it should not be")
```
waynr: Should this also be changed to `/^BAD$/`?

jeffmccune: It's not necessary for the ruby subcommand because ruby doesn't echo the input back to stdout like the irb subcommand does.

waynr:
:+1:

camlow325:
:+1:

PR #364, opened 2015-01-14
(SERVER-252) Fix bad irb acceptance test... again.

jeffmccune:
Same issues as before, string interpolation was happening in beaker when it
should be passed through.  Switched from #{} to concatenation with +.


camlow325:
:+1:

PR #367, opened 2015-01-22
(MAINT) Upgrade beaker and sqa-utils gems.

waynr:
I cannot run beaker locally on my laptop without upgrading to 2.2. I am currently working to test this out on an experimental pipeline.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,8 +5,8 @@ gem 'jira-ruby', :group => :development
 
 group :test do
   gem 'rspec'
-  gem 'beaker', '~>1.17.0'
+  gem 'beaker', '~> 2.2'
   if ENV['GEM_SOURCE'] =~ /rubygems\.delivery\.puppetlabs\.net/
-    gem 'sqa-utils'
+    gem 'sqa-utils', '~> 0.11'
   end
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    CFPropertyList (2.2.8)
+    CFPropertyList (2.3.0)
     activesupport (4.1.9)
       i18n (~> 0.6, >= 0.6.9)
       json (~> 1.7, >= 1.7.7)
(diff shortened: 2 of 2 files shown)


Gemfile.lock
```diff
@@ -1,7 +1,7 @@
 GEM
-  remote: https://rubygems.org/
+  remote: http://rubygems.delivery.puppetlabs.net/
```
elyscape: Be advised that this does not appear to be accessible to the internet at large.

waynr:
@elyscape Thanks for pointing that out, it is now fixed.

Gemfile
```diff
@@ -5,8 +5,8 @@ gem 'jira-ruby', :group => :development
 
 group :test do
   gem 'rspec'
-  gem 'beaker', '~>1.17.0'
+  gem 'beaker', '~> 2.2'
   if ENV['GEM_SOURCE'] =~ /rubygems\.delivery\.puppetlabs\.net/
-    gem 'sqa-utils'
+    gem 'sqa-utils', '> 0.11'
```
rlinehan: Do we need to pin this to > 0.11, or can we just `bundle update` to update the Gemfile.lock?

rlinehan:
Aside from my comment on whether we need to pin sqa-utils, I'm +1 on this. Couldn't run acceptance tests locally without beaker 2.2 - I was getting a weird stacktrace otherwise.

Gemfile
waynr: Hmm, I think I intended for this to be `~> 0.11` which I believe would prevent major version updates while still giving us minor and patch version updates.

rlinehan: Okay, I guess that makes more sense if we don't want to have automatic major version updates.

waynr:
@rlinehan 
Okay I updated the version constraing on `sqa-utils` so as far as I'm concerned it's ready to go.

rlinehan:
:+1:

cprice404:
@rlinehan does your other PR subsume this one?

rlinehan:
@cprice404 yeah, my pr has this same stuff in it because I just wanted to push it up before lunch, but I'm happy merging this and then removing my commit from my PR.

cprice404:
I'd say let's just close this and merge yours.

PR #368, opened 2015-01-22
(SERVER-149) Update Puppet Server for Puppet 4 url changes

rlinehan:
For Puppet 4 the master's routes have changed. The environment is a query
param, there is a version string (v3) for all routes, and there is a url
prefix. This makes the necessary updates to Puppet Server to work with
these changes.

The CA routes have been split out completely from the master routes.
They are now at a different prefix and version than the master routes in
Puppet's code. This means that in Puppet Server they can have their own
prefix (by default `/puppet-ca`) - they no longer need to be mounted at `/`.

In addition, the environment has been moved from the first part of the path to
a query parameter. The environment is never used in the CA routes, so this
query parameter is now ignored completely.

This also adds a check that the prefix specified in Puppet's code through
the `master_url_prefix` setting matches the prefix specified for the master
service in the webrouting config. If these don't match, a user might get very
confusing and undecipherable errors. Instead, we raise an error if a
discrepancy is detected, forcing it to be reconciled before Puppet Server can
be started.

### DIFF
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -8,13 +8,6 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Routing
 
-(defn v2_0-routes
-  "Creates the compojure routes to handle the master's '/v2.0' routes."
-  [request-handler]
-  (compojure/routes
-    (compojure/GET "/environments" request
-                   (request-handler request))))
-
 (defn v3-routes
   "Creates the compojure routes to handle the master's '/v3' routes, which
    includes '/environments' and the non-CA indirected routes. The CA-related
@@ -58,8 +51,6 @@
   "Creates all of the compojure routes for the master."
   [request-handler]
   (compojure/routes
-    (compojure/context "/v2.0" request
-                       (v2_0-routes request-handler))
     (compojure/context "/v3" request
                        (v3-routes request-handler))
     (route/not-found "Not Found")))
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -14,11 +14,10 @@
     (compojure/GET "/environments" request
                    (request-handler request))))
 
-(defn legacy-routes
-  "Creates the compojure routes to handle the master's 'legacy' routes
-   - ie, any route without a version in its path (eg, /v2.0/whatever) - but
-   excluding the CA-related endpoints, which are handled separately by the
-   CA service."
+(defn v3-routes
+  "Creates the compojure routes to handle the master's '/v3' routes, which
+   includes '/environments' and the non-CA indirected routes. The CA-related
+   endpoints are handled separately by the CA service."
   [request-handler]
   (compojure/routes
     (compojure/GET "/node/*" request
@@ -60,8 +59,8 @@
   (compojure/routes
     (compojure/context "/v2.0" request
                        (v2_0-routes request-handler))
-    (compojure/context "/:environment" [environment]
-                       (legacy-routes request-handler))))
+    (compojure/context "/v3" request
+                       (v3-routes request-handler))))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Lifecycle Helper Functions
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -1,6 +1,7 @@
 (ns puppetlabs.services.master.master-core
   (:import (java.io FileInputStream))
   (:require [compojure.core :as compojure]
+            [compojure.route :as route]
             [me.raynes.fs :as fs]
             [puppetlabs.puppetserver.ringutils :as ringutils]))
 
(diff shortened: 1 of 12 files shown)


rlinehan:
This can't be merged until the changes for PUP-3526 (https://github.com/puppetlabs/puppet/pull/3464) have been merged into Puppet. Currently, the Puppet submodule is pointing at the most recent commit on that branch. Once that is merged, this PR will need to be updated to have the submodule pointing at the correct commit.

src/clj/puppetlabs/services/config/puppet_server_config_core.clj
```diff
@@ -92,6 +93,18 @@
                     "conflict with the values to be "
                     "read from puppet: " key-conflicts))))))
 
+(defn check-master-url-prefix
+  "Ensure that Puppet Server's route for the puppet master matches what is in
+  puppet.conf."
+  [webrouter-config puppet-config]
+  (let [puppet-prefix (:master-url-prefix puppet-config)
+        webrouter-prefix (:puppetlabs.services.master.master-service/master-service webrouter-config)]
```
rlinehan: This seems like a rather ugly way to be getting this, but it didn't seem like from here we have a good way to get a reference to the master service, and without that I couldn't figure out how to use any of the helpers such as `get-route`. Would love any input on a better way to do this.

rlinehan:
Travis build failed because submodule is only on my fork of Puppet.

src/clj/puppetlabs/services/config/puppet_server_config_core.clj
cprice404: I think you could use this:

https://github.com/puppetlabs/trapperkeeper/blob/master/src/puppetlabs/trapperkeeper/services.clj#L25

if you did it directly in the `init` definition.  It looks like we don't currently have a protocol for the master service, so you'd have to create one (can be empty), and call it `PuppetMasterService` or whatever.  Then, in the `init` for PuppetServerConfigService you could do something like this:

`(get-service this :PuppetMasterService)`

and pass that to the webrouter `get-route` function.

That ^^ is pretty gross too, but probably a little safer than what you have here, because theoretically the webrouter config could be nested a little more deeply than how we currently have it configured, and if that happened then the approach of trying to pluck the value directly out of the config would either not work or would require a bunch of extra conditional logic.

src/clj/puppetlabs/services/config/puppet_server_config_core.clj
```diff
@@ -92,6 +93,18 @@
...
+    (when-not (= puppet-prefix webrouter-prefix)
+      (throw (Exception.
```
nwolfe: We should pick a more specific Exception to use here, or we should consider using slingshot instead.

Basically our pattern has been to throw an exception if you intend for it to bring down the program and give a good message to the user.  On the other hand, if you intend for the "exception" to just be caught further up the stack and handled there, use slingshot.

cprice404: We should probably choose a more specific exception type.  We could use `IllegalArgumentException` or `IllegalStateException`, or, we could use this as an opportunity to introduce our own exception base class like `PuppetServerException` if folks want to start trying to head in that direction.  Previously I thought we (PL) had been trying to avoid introducing new Java exception types, but I feel like I've been seeing them start to pop up in some of the other projects related to console services, so maybe it'd be OK.

src/ruby/puppet-server-lib/puppet/server/puppet_config.rb
```diff
@@ -81,9 +81,6 @@ def self.reset_environment_context
     end
     Puppet.push_context({:current_environment => configured_environment},
                         "Update current environment from puppet master's configuration")
-
-    require 'puppet/util/instrumentation'
-    Puppet::Util::Instrumentation.init
```
cprice404: yay!

test/integration/puppetlabs/services/jruby/puppet_environments_int_test.clj
```diff
@@ -83,7 +83,7 @@
 
 (defn catalog-contains?
   [catalog resource-type resource-title]
-  (let [resources (get-in catalog ["data" "resources"])]
+  (let [resources (get catalog "resources")]
```
cprice404: Why did this change?

src/clj/puppetlabs/services/config/puppet_server_config_core.clj
```diff
@@ -92,6 +93,18 @@
                     "conflict with the values to be "
                     "read from puppet: " key-conflicts))))))
 
+(defn check-master-url-prefix
```
nwolfe: No big deal, but our pattern for these kinds of functions has generally been to name them like `validate-master-url-prefix!` to signify that they will basically bring the program down, and that they don't return anything but are used for the side effect (in this case, throwing an exception).

cprice404:
:see_no_evil: just a few nitpicks as mentioned above.

test/integration/puppetlabs/services/jruby/puppet_environments_int_test.clj
rlinehan: Ah, I meant to separate that into it's own commit. What's being returned is no longer

```
{ ...
  "data": { "resources": [...], ... }
}
```
it's now just `{ ... "resources": [...], ... }`

This seems to have been changed as part of removing the PSON document type: https://github.com/puppetlabs/puppet/commit/7007b6f96ef953dffbf14851b37fdbd388fdc2d7

cprice404: gotcha.  sounds good.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -14,7 +15,7 @@
     (compojure/GET "/environments" request
                    (request-handler request))))
 
-(defn legacy-routes
+(defn v3-routes
   "Creates the compojure routes to handle the master's 'legacy' routes
```
fpringvaldsen: Are these still legacy routes?

rlinehan: Whoops, good catch! I'll update the docstring.

fpringvaldsen:
Looks good! :+1: barring my one comment and the feedback from @cprice404 and @nwolfe.

rlinehan:
PUP-3526 has been merged in to Puppet, and the Puppet submodule has been updated to point at that. This should be ready to go.

PR #369, opened 2015-01-23
(SERVER-101) Changed jvm-ca references to jvm-ssl-utils

jpinsonault:
* Updated dependencies to reflect renaming puppetlabs/certificate-authority project to puppetlabs/ssl-utils
* Changed references to old jvm-ca project to jvm-ssl-utils in clojure/ruby files

### DIFF
--- a/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
+++ b/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
@@ -1,7 +1,6 @@
 (ns puppetlabs.services.ca.certificate-authority-core
   (:import  [java.io InputStream])
   (:require [puppetlabs.puppetserver.certificate-authority :as ca]
-            [puppetlabs.certificate-authority.core :as ca-utils]
             [puppetlabs.puppetserver.ringutils :as ringutils]
             [puppetlabs.puppetserver.liberator-utils :as liberator-utils]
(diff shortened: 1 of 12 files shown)


src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -1,7 +1,7 @@
 (ns puppetlabs.services.ca.certificate-authority-core
   (:import  [java.io InputStream])
   (:require [puppetlabs.puppetserver.certificate-authority :as ca]
-            [puppetlabs.certificate-authority.core :as ca-utils]
+            [puppetlabs.ssl-utils.core :as utils]
```
nwolfe: So, this would break the usages of `ca-utils` since those weren't renamed to `ssl-utils`...but there aren't any!! We can just delete this :require altogether

nwolfe:
We missed the usages in https://github.com/puppetlabs/puppet-server/blob/master/src/ruby/puppet-server-lib/puppet/server/config.rb#L13

jpinsonault:
@nwolfe Good catch, updated

dankreek:
:+1: All tests passed on my machine, looks good.

fpringvaldsen:
:+1: All tests passed for me as well.

PR #371, opened 2015-01-27
(FOR REVIEW) (SERVER-300) Craft special 404 error message

fpringvaldsen:
Craft a special 404 error message that occurs when a request is
made to a route that does not contain the prefix for the Master
or CA services. This 404 error messages explains that the URLs
have been changed in Puppet 4 and that the agent should be
upgraded.

I'm marking this as FOR REVIEW because I'm fairly certain that the 404 error message I created is definitely going to need some work, and there's going to need to be some discussion on what exactly it should say. I wanted to get something up so people could look, discuss, and provide feedback.

### DIFF
--- a/dev/puppet-server.conf.sample
+++ b/dev/puppet-server.conf.sample
@@ -30,8 +30,8 @@ webserver: {
 # configure the mount points for the web apps
 web-router-service: {
     "puppetlabs.services.ca.certificate-authority-service/certificate-authority-service": "/puppet-ca"
-    "puppetlabs.services.master.master-service/master-service": "/puppet"
-
+    "puppetlabs.services.master.master-service/master-service": {master-routes: "/puppet"
+                                                                 404:           "/" }
     # This controls the mount point for the puppet admin API.
     "puppetlabs.services.puppet-admin.puppet-admin-service/puppet-admin-service": "/puppet-admin-api"
 }
--- a/dev/puppet-server.conf.sample
+++ b/dev/puppet-server.conf.sample
@@ -30,8 +30,11 @@ webserver: {
 # configure the mount points for the web apps
 web-router-service: {
     "puppetlabs.services.ca.certificate-authority-service/certificate-authority-service": "/puppet-ca"
-    "puppetlabs.services.master.master-service/master-service": {master-routes: "/puppet"
-                                                                 404:           "/" }
+    "puppetlabs.services.master.master-service/master-service": {master-routes:       "/puppet"
+
+                                                                 # The following controls 404 errors and should not
+                                                                 # be modified
+                                                                 invalid-in-puppet-4: "/" }
     # This controls the mount point for the puppet admin API.
     "puppetlabs.services.puppet-admin.puppet-admin-service/puppet-admin-service": "/puppet-admin-api"
 }
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -3,7 +3,18 @@
   (:require [compojure.core :as compojure]
             [compojure.route :as route]
             [me.raynes.fs :as fs]
-            [puppetlabs.puppetserver.ringutils :as ringutils]))
+            [puppetlabs.puppetserver.ringutils :as ringutils]
+            [puppetlabs.services.version.version-check-core :as version-check]
+            [puppetlabs.services.config.puppet-server-config-core :as config]))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Constants
+
+(def puppet-API-versions
+  "v3")
+
+(def puppet-ca-API-versions
+  "v1")
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Routing
@@ -114,6 +125,22 @@
                       "/etc/default/puppetserver on Debian systems.")))))))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Helper Functions
+
+(defn construct-404-error-message
+  [jruby-service]
+  (str "Error: Invalid URL - Puppet Server expects requests that conform to the "
+       "/puppet and /puppet-ca APIs.\n\n"
+       "Note that Puppet 3 agents aren't compatible with this version; if you're "
+       "running Puppet 3, you must either upgrade your agents to match the server "
+       "or point them to a server running Puppet 3.\n\n"
+       "Server Info:\n"
+       "  Puppet Server version: " (version-check/get-version-string "puppet-server") "\n"
+       "  Puppet version: " (:puppet-version (config/get-puppet-config jruby-service)) "\n"
+       "  Supported /puppet API versions: " puppet-API-versions
+       "  Supported /puppet-ca API versions: " puppet-ca-API-versions))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Public
 
 (defn build-ring-handler
@@ -124,3 +151,12 @@
       ringutils/wrap-exception-handling
       ringutils/wrap-request-logging
       ringutils/wrap-response-logging))
+
+(defn construct-invalid-request-handler
+  "Constructs a ring handler to handle an incorrectly formatted request and indicate to the user
+   they need to update to Puppet 4"
+  [jruby-service]
+  (fn
+    [_]
+    {:status 404
+     :body   (construct-404-error-message jruby-service)}))
--- a/src/clj/puppetlabs/services/master/master_service.clj
+++ b/src/clj/puppetlabs/services/master/master_service.clj
@@ -3,13 +3,15 @@
             [compojure.core :as compojure]
             [puppetlabs.trapperkeeper.core :refer [defservice]]
             [puppetlabs.services.master.master-core :as core]
-            [puppetlabs.puppetserver.certificate-authority :as ca]))
+            [puppetlabs.puppetserver.certificate-authority :as ca]
+            [puppetlabs.trapperkeeper.services :as tk-services]))
 
 (defservice master-service
   [[:WebroutingService add-ring-handler get-route]
    [:PuppetServerConfigService get-config]
-   [:RequestHandlerService handle-request handle-invalid-request]
-   [:CaService initialize-master-ssl! retrieve-ca-cert! retrieve-ca-crl!]]
+   [:RequestHandlerService handle-request]
+   [:CaService initialize-master-ssl! retrieve-ca-cert! retrieve-ca-crl!]
+   [:JRubyPuppetService]]
   (init
    [this context]
    (core/validate-memory-requirements!)
@@ -18,18 +20,19 @@
          certname    (get-in config [:puppet-server :certname])
          localcacert (get-in config [:puppet-server :localcacert])
          hostcrl     (get-in config [:puppet-server :hostcrl])
-         settings    (ca/config->master-settings config)]
+         settings    (ca/config->master-settings config)
+         jruby-service (tk-services/get-service this :JRubyPuppetService)]
 
      (retrieve-ca-cert! localcacert)
      (retrieve-ca-crl! hostcrl)
      (initialize-master-ssl! settings certname)
 
-     (log/info "Master Service adding a ring handler")
+     (log/info "Master Service adding ring handlers")
      (add-ring-handler
        this
       (compojure/context path [] (core/build-ring-handler handle-request))
       {:route-id :master-routes})
-     (add-ring-handler this handle-invalid-request {:route-id :404}))
+     (add-ring-handler this (core/construct-invalid-request-handler jruby-service) {:route-id :invalid-in-puppet-4}))
    context)
   (start
     [this context]
--- a/src/clj/puppetlabs/services/master/master_service.clj
+++ b/src/clj/puppetlabs/services/master/master_service.clj
@@ -8,12 +8,12 @@
 (defservice master-service
   [[:WebroutingService add-ring-handler get-route]
    [:PuppetServerConfigService get-config]
-   [:RequestHandlerService handle-request]
+   [:RequestHandlerService handle-request handle-invalid-request]
    [:CaService initialize-master-ssl! retrieve-ca-cert! retrieve-ca-crl!]]
   (init
    [this context]
    (core/validate-memory-requirements!)
-   (let [path        (get-route this)
+   (let [path        (get-route this :master-routes)
          config      (get-config)
          certname    (get-in config [:puppet-server :certname])
          localcacert (get-in config [:puppet-server :localcacert])
@@ -27,7 +27,9 @@
      (log/info "Master Service adding a ring handler")
      (add-ring-handler
        this
-      (compojure/context path [] (core/build-ring-handler handle-request))))
+      (compojure/context path [] (core/build-ring-handler handle-request))
+      {:route-id :master-routes})
+     (add-ring-handler this handle-invalid-request {:route-id :404}))
    context)
   (start
     [this context]
--- a/src/clj/puppetlabs/services/protocols/request_handler.clj
+++ b/src/clj/puppetlabs/services/protocols/request_handler.clj
@@ -3,4 +3,7 @@
 (defprotocol RequestHandlerService
   (handle-request
     [this request]
-    "Handles a request from an agent."))
+    "Handles a request from an agent.")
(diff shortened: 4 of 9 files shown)


src/java/com/puppetlabs/puppetserver/ProcessWrapper.java
```diff
@@ -48,7 +48,7 @@ public int getExitCode() {
             this.pw = pw;
         }
 
-        @Override
+        //@Override
```
KevinCorcoran: Why comment this out?

fpringvaldsen: Ah, that was an accident, sorry about that, IntelliJ was being weird and this was part of my attempt to fix it. I'll uncomment that.

fpringvaldsen: @KevinCorcoran That should be gone now.

src/clj/puppetlabs/services/protocols/request_handler.clj
```diff
@@ -3,4 +3,7 @@
 (defprotocol RequestHandlerService
   (handle-request
     [this request]
-    "Handles a request from an agent."))
+    "Handles a request from an agent.")
+  (handle-invalid-request
```
cprice404: I view the `request_handler` service as strictly being responsible for requests that require JRuby interaction.  (We should probably have chosen a name that makes that more obvious.)

Could we put the 404 stuff directly in the `master` service?

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
```diff
@@ -12,6 +12,14 @@
             [slingshot.slingshot :as sling]))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Constants
+
+(def error-message-404
+  (str "Error: This master is running Puppet 4, and the URL format has changed. "
+       "This request was likely made with a Puppet 3 Agent, as it used an invalid "
+       "URL format. Please upgrade your agents to Puppet 4 and try again."))
```
cprice404: @nfagerlund @jbondpdx we could use some input / wordsmithing from you all here.

cprice404:
Looks fine to me, other than potential wordsmithing from docs team, plus moving the handler out of the jruby `request_handler` service if possible.

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
nfagerlund: Huh. Question that just occurred to me: Are we handling 404ed URLs that we *know* correspond to old Puppet versions any differently than we'll handle URLs pointing to Jupiter? Because I'd rather not show this message to someone running, say, Puppet 6.

fpringvaldsen:
Just opened https://github.com/puppetlabs/puppet/pull/3532, which is the same change made to the ruby master.

src/clj/puppetlabs/services/protocols/request_handler.clj
fpringvaldsen: Yup, that shouldn't be difficult.

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
cprice404: Handling "known" URLs would involve all of the same problems that we're trying to get away from, about having to camp on the entire "/" URL namespace... because the first segment of the URL from a puppet 3 agent is going to be the environment name, which could be literally anything.

So, the approach we agreed on as 'best effort' for now is to just add a catch-all 404 handler that has an error message that is suitable for cases where the issue *is* related to a Puppet 3 agent, but also to cases where the issue is *not* related to a Puppet 3 agent.  So we may want to soffent up some of the wording, e.g. change "likely" to "possibly", add in something like "If your agent is puppet 3, try upgrading and running the agent again.", etc.

fpringvaldsen: @nfagerlund As it stands right now with the way I've implemented this, any route that does not have the new configurable master or CA prefix will fall through to this handler and see the 404 error message. I can change that if necessary, however.

cprice404: p.s. we can also pull this error message out of Puppet Server in a Y release sometime down the road if we see fit, which would hopefully get us to a point where it wasn't still in the code by the time there is such a thing as Puppet 6 :)

nfagerlund: Ah cool, thanks. That changes how I should write my suggested message... which I promise I'm working on in between meetings. :)

cprice404: np.  I think we have about a week or so to come up with the final verbiage.

We may go ahead and merge this PR before we finalize the verbiage, just to keep the machinery rolling, and we can open a follow-up PR to clean it up if that's the case.

nfagerlund: So, the biggest thing I'd like to get in here is a block of server info, listing the software and API versions involved. That's probably the nicest and most useful 404 API response I can imagine. (BTW, do we have an API endpoint that lists this info?) 

Anyway, I suggest something like this:

    Error: Invalid URL - Puppet Server expects requests that conform to the
    /puppet and /puppet-ca APIs.
    
    Note that Puppet 3 agents aren't compatible with this version; if you're
    running Puppet 3, you must either upgrade your agents to match the server
    or point them to a server running Puppet 3.

    Server info:
      Puppet Server version <VERSION>
      Puppet version <VERSION>
      Supported /puppet API versions:
        v3
      Supported /puppet-ca API versions:
        v1

jeffmccune: > So, the biggest thing I'd like to get in here is a block of server info, listing the software and API versions involved. That's probably the nicest and most useful 404 API response I can imagine.

Yes, this information is helpful so the end user knows where they stand presently and can plan a course of action to resolve the problem, for example, "I need to find the upgrade documentation for version X to version Y."

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
```diff
@@ -276,4 +284,9 @@
...
       (-> (ring-response/response message)
           (ring-response/status 400)
-          (ring-response/content-type "text/plain")))))
\ No newline at end of file
+          (ring-response/content-type "text/plain")))))
+
+(defn handle-invalid-request
+  [_]
+  {:status 404
+   :body error-message-404})
```
jeffmccune: Whitespace error: no newline at end of file.

dev/puppet-server.conf.sample
```diff
@@ -30,8 +30,8 @@ webserver: {
 # configure the mount points for the web apps
 web-router-service: {
     "puppetlabs.services.ca.certificate-authority-service/certificate-authority-service": "/puppet-ca"
-    "puppetlabs.services.master.master-service/master-service": "/puppet"
-
+    "puppetlabs.services.master.master-service/master-service": {master-routes: "/puppet"
+                                                                 404:           "/" }
```
jeffmccune: Minor, but I advise avoiding identifiers that start with a number and/or are entirely numeric.  `[a-z][a-z0-9]*` is usually best, as in the convention for naming a function.  I suggest `http404` or `status404` or `not-found` or something like that as an alternative identifier.

jeffmccune: On second thought, are all of these "invalid requests" or "404's" falling under the category of "old puppet agents making previous valid, but now invalid requests?"  If so, I'd make the route identifier named something to indicate that because it's a little confusing to have a route that seems like it's for anything that can't be found (404) but has behavior specific to helping the user upgrade their agent version.

jeffmccune:
The commit message does a good job of explaining what the code changes, but not why the change is being made.  Commit messages should explain why the change is being made (it could be a single sentence).  That is to say, why is a special error message being crafted versus using a normal one?  Without the special error message, does the user experience a failure, or is this more of a deprecation warning / suggestion?

test/integration/puppetlabs/puppetserver/error_handling_int_test.clj
```diff
@@ -56,4 +56,17 @@
...
+                           bootstrap/request-options)]
+            (is (= 404 (:status response)))
+            (is (= (:body response) request-handler/error-message-404))))
+
+        (testing "Puppet 3 CA URLs are no longer valid"
+          (let [response (http-client/get
+                           "https://localhost:8140/production/certificate_status/localhost"
+                           bootstrap/request-options)]
+            (is (= 404 (:status response)))
+            (is (= (:body response) request-handler/error-message-404))))))))
```
jeffmccune: This test validates the behavior of the system returns a specific symbol, which is OK, but it would be a great improvement if the test validated the behavior the end user would expect.  In this situation, the end user should expect to see a helpful error message with server info, software and API versions, etc...

It'd be great to assert the string contains that information so if we change error-message-404 to something "bad" the tests will catch it for us.

fpringvaldsen:
@jeffmccune I've updated my commit message, and I'm about to get to work on addressing the rest of the feedback left by you and @nfagerlund.

Jhesiah:
what can I do to to clear internal storage?/low on space?

Jhesiah:
Frustrating ready to factory reset phone

Jhesiah:
I do appreciate the boundary breaking family of githib I I've been working on a specific utility which conserves water, I need a team to present and market my utility.

Jhesiah:
I do appreciate the boundary breaking family of githib I I've been working on a specific utility which conserves water, I need a team to present and market my utility.

dev/puppet-server.conf.sample
fpringvaldsen: That's a good idea, I'll change the name. This response will only be returned for requests that are missing the configurable `puppet` and `puppet-ca` prefixes.

fpringvaldsen:
@cprice404 @jeffmccune @nfagerlund I've addressed all of your feedback, and also done a bit of refactoring.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -114,6 +125,22 @@
...
+  (str "Error: Invalid URL - Puppet Server expects requests that conform to the "
+       "/puppet and /puppet-ca APIs.\n\n"
+       "Note that Puppet 3 agents aren't compatible with this version; if you're "
+       "running Puppet 3, you must either upgrade your agents to match the server "
+       "or point them to a server running Puppet 3.\n\n"
+       "Server Info:\n"
+       "  Puppet Server version: " (version-check/get-version-string "puppet-server") "\n"
+       "  Puppet version: " (:puppet-version (config/get-puppet-config jruby-service)) "\n"
+       "  Supported /puppet API versions: " puppet-API-versions
+       "  Supported /puppet-ca API versions: " puppet-ca-API-versions))
```
nfagerlund: :+1: :+1: :+1:

src/clj/puppetlabs/services/master/master_service.clj
```diff
@@ -27,7 +30,9 @@
      (log/info "Master Service adding a ring handler")
```
jeffmccune: We're adding multiple ring handlers now.

dev/puppet-server.conf.sample
```diff
@@ -30,8 +30,11 @@ webserver: {
 # configure the mount points for the web apps
 web-router-service: {
     "puppetlabs.services.ca.certificate-authority-service/certificate-authority-service": "/puppet-ca"
-    "puppetlabs.services.master.master-service/master-service": "/puppet"
+    "puppetlabs.services.master.master-service/master-service": {master-routes:       "/puppet"
 
+                                                                 # The following controls 404 errors and should not
+                                                                 # be modified
+                                                                 invalid-in-puppet-4: "/" }
```
jeffmccune: Similar to comments Ruth made in the MRI pull request, will this route catch all invalid requests, or only the subset of invalid requests that happen to be previously valid?  If it's going to catch all invalid requests, it should be changed to only display the "upgrade" message when the request is from an agent that previously was valid.

test/integration/puppetlabs/puppetserver/error_handling_int_test.clj
```diff
@@ -57,3 +61,36 @@
...
+            (let [response (http-client/get
+                             "https://localhost:8140/production/catalog/localhost"
+                             bootstrap/request-options)]
+              (is (= 404 (:status response)))
+              (is (re-find puppet-server-version-regex (:body response)))
+              (is (re-find puppet-version-regex (:body response)))
+              (is (re-find puppet-api-regex (:body response)))
+              (is (re-find puppet-ca-api-regex (:body response)))))
+
+          (testing "Puppet 3 CA URLs are no longer valid"
```
jeffmccune: These are great, but we need one more test where a request is made that has never been valid.  For example, when I request `/foo/bar/baz/this/has/never/been/vaid`, I get a response that does _not_ inform the user to upgrade the agent.

dev/puppet-server.conf.sample
fpringvaldsen: @jeffmccune This should only catch requests that don't have the `/puppet` or `/puppet-ca` prefixes. Any other invalid request (say, `/puppet/v3/unknown`) will return the normal 404 error message.

I think that's a valuable thing to test, however, so I'll add tests to ensure that that is true.

test/integration/puppetlabs/puppetserver/error_handling_int_test.clj
fpringvaldsen: @jeffmccune I just added some new tests to ensure that an invalid but prefixed URL does not return the upgrade message. I think that should cover this concern, but if not let me know and I'll add another one!

dev/puppet-server.conf.sample
jeffmccune: Great, thanks for the clarification.

jeffmccune: Oh wait.  What does the user see if they request `/some/resource/that/never/existed/ever` (And a previous Puppet agent never tried to request)?

If they see the "upgrade" error message, then that's an issue we should discuss further, because that request isn't being made by an agent that needs to be upgraded.

fpringvaldsen: @jeffmccune I believe @nfagerlund raised that concern earlier, and @cprice404 addressed it above. Basically, only handling known URLs would cause us to have a number of problems that we're currently trying to get away from having around camping out on the "/" URL namespace since the environment could be anything, so we've decided that the catch-all 404 handler is the right approach for now.

cprice404: Yes, this question has been discussed at great length in many forums, and we're not going to try to restrict the message to only URLs that would have been valid in old agents.

jeffmccune: Could you refer me to the comment and Chris addressing it?  I'm having a heck of a time finding it.

To be clear, I'm not trying to prescribe how we're doing it.  I'd expect the solution would involve checking something other than the URL itself, because the URI alone doesn't give a lot of indication that the request was previously valid, but is now invalid.

I'd expect the solution to involve checking the request headers, or some other meta-data in addition to the URI to figure out if the request is a from an agent that needs to be updated.

cprice404: The message says 'if you're running Puppet 3'.  That's what we agreed on when we discussed this in architecture a while back.  We can add in some fancier logic in another PR if we have bandwidth and a good idea on how to improve it.

fpringvaldsen: @jeffmccune If you're still looking for the conversation, it's above under the third comment hidden by an outdated diff.

jeffmccune: > @jeffmccune If you're still looking for the conversation, it's above under the third comment hidden by an outdated diff.

Thanks, I see it now.  Chris said:

> So, the approach we agreed on as 'best effort' for now is to just add a catch-all 404 handler that has an error message that is suitable for cases where the issue is related to a Puppet 3 agent, but also to cases where the issue is not related to a Puppet 3 agent.

cprice404:
+1 as long as @jeffmccune 's concerns have been addressed.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -114,6 +125,22 @@
...
+(defn construct-404-error-message
+  [jruby-service]
+  (str "Error: Invalid URL - Puppet Server expects requests that conform to the "
+       "/puppet and /puppet-ca APIs.\n\n"
+       "Note that Puppet 3 agents aren't compatible with this version; if you're "
+       "running Puppet 3, you must either upgrade your agents to match the server "
+       "or point them to a server running Puppet 3.\n\n"
+       "Server Info:\n"
+       "  Puppet Server version: " (version-check/get-version-string "puppet-server") "\n"
+       "  Puppet version: " (:puppet-version (config/get-puppet-config jruby-service)) "\n"
```
cprice404: @fpringvaldsen sorry that I didn't spot this sooner, but the call into JRuby is pretty expensive; it'd be much better to do this once at startup and then pass the value into this function rather than hitting JRuby on every 404.  Would you mind opening a new ticket to capture that?

fpringvaldsen: @cprice404 I've created (SERVER-334)[https://tickets.puppetlabs.com/browse/SERVER-334] to track this. It'll be a really easy fix so I went ahead and pulled it into the current sprint and will knock it out today.

PR #372, opened 2015-01-27
(SERVER-155) Eliminate duplicated code

fpringvaldsen:
Eliminate the `reset_environment_context` and
`configure-indirector-routes` methods, which were copied directly
from Puppet, in favor of the new methods in
Puppet::ApplicationSupport.

Both the clojure and ruby tests passed. I tried running the acceptance tests but was unable to do so on my machine due to a known issue with beaker and zsh.


KevinCorcoran:
:+1: this looks correct to me.

rlinehan:
:+1:

PR #374, opened 2015-01-29
(MAINT) Revert SERVER-262

dankreek:


camlow325:
:+1: - Not quite sure if some of the acceptance test reverts which would put us back to installing beaker for the gem test will work but we should be able to see that quickly if not.

dankreek:
Updated the acceptance test per @waynr's advice

waynr:
:+1:, I'll be on my toes to watch for failures in the smoke tests

jeffmccune:
Please see https://tickets.puppetlabs.com/browse/SERVER-522 for additional information.

PR #375, opened 2015-01-30
(maint) Remove unused require from puppet-server-lib

briancain:
Prior to this commit, there was an old require that referenced a library
from puppet that no longer exists. This caused puppet-server to fail on
startup. This commit fixes that by removing the require reference.

### DIFF
--- a/src/ruby/puppet-server-lib/puppet/server/master.rb
+++ b/src/ruby/puppet-server-lib/puppet/server/master.rb
@@ -1,7 +1,6 @@
 require 'puppet/server'
 
 require 'puppet/network/http'
-require 'puppet/network/http/api/master/v2'
 require 'puppet/network/http/api/master/v3'
 
 require 'puppet/server/config'


KevinCorcoran:
Haha, oh man, the Travis build failed due to a test failure, in a test that I'm pretty sure I wrote a long time ago ...
```clj
FAIL in (config-key-conflicts) (puppet_server_config_service_test.clj:91)
Providing config values that should be read from Puppet results in an error that mentions all offending config keys.
expected: (thrown-with-msg? Exception #".*configuration.*conflict.*cacrl.*cacert" (tk-testutils/with-app-with-config app service-and-deps (assoc required-config :cacrl "bogus" :cacert "meow")))
  actual: #<Exception java.lang.Exception: The following configuration keys conflict with the values to be read from puppet: #{:cacert :cacrl}>
```
The issue is that the conflicting keys are a set and were stringified in an order that was different than what the test was expecting.  Not cool, past self.

This is obviously a bug in the test and shouldn't hold up this PR getting merged.

KevinCorcoran:
I betcha the reason we haven't hit that until now is because of https://github.com/puppetlabs/puppet-server/pull/373

briancain:
Strange! I also noticed in the Travis CI output that bundler failed. I didn't know if that was related either....

elyscape:
The bundler failure looks like a temporary connectivity issue:
> Gem::RemoteFetcher::FetchError: Errno::ECONNREFUSED: Connection refused - Connection refused (https://rubygems.org/gems/metaclass-0.0.4.gem)

rlinehan:
This is hopefully fixed by https://github.com/puppetlabs/puppet-server/commit/37c87b3862d041c030b11ccad8cd349ee0155d99, will kick travis.

briancain:
Woo, green! :)

PR #377, opened 2015-02-02
(MAINT) be more tolerant of ssl exceptions in tests

cprice404:
Because the HTTP client library we are using is asynchronous, there
is a race condition that happens when an SSL handshake fails; you
may get a "connection closed" exception from Jetty before you get
the "SSLHandshakeException" that would eventually be thrown by
the HTTP client library.

When writing tests that involve intentional SSL failures, the
tests need to be tolerant of the different types of exceptions that
can occur.  This commit changes some tests that were not working
that way to be a bit more tolerant of the different possible
exception types, while still trying to be as specific as possible
and validate that the exception actually represents the
expected SSL failure.

### DIFF
--- a/test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
+++ b/test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
@@ -41,6 +41,14 @@
   (-> ring-app
       (auth/wrap-basic-authentication authenticated?)))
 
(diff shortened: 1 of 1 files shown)


jeffmccune:
:+1:

jeffmccune:
Tests pass locally and on Travis, merging with only one reviewer since this is a small MAINT patch.

PR #378, opened 2015-02-02
Fix packaging/acceptance tests

rlinehan:
This PR consists of a combination of fixes and hacks to get Server 2.0 packages that will actually install and be able to start the service, and also to get smoke tests passing. Probably once the AIO agent packages are done we'll need to change/remove many of these things (especially changes around creating the `puppet` user).


acceptance/lib/helper.rb
```diff
@@ -203,11 +204,17 @@ def puppet_apply_as_puppet_user
     EOS
 
     user = master.puppet('master')['user']
+    on master, "groupadd -r puppet"
+    on master, "useradd -r -g puppet -d /opt/puppetlabs/agent/cache #{user} "
+    on master, "chown puppet:puppet /opt/puppetlabs/agent/cache/"
+    on master, "mkdir /opt/puppetlabs/agent/cache/reports"
+    on master, "chown puppet:puppet /opt/puppetlabs/agent/cache/reports"
+    on master, "chmod 0755 /opt/puppetlabs/agent/cache/reports"
```
cprice404: @rlinehan @haus we are expecting many/most of these to be removed when there are upstream fixes, correct?

@rlinehan would you mind just adding a comment line indicating that?  bonus points for creating a jira to track it and including the ticket number.

cprice404:
+1 from me, other than potentially adding a comment or two indicating which things we expect to be able to rip out later.

haus:
@rlinehan tickets to watch for these are... RE-3719 and RE-3878

acceptance/config/beaker/jenkins/el-6-x86_64-mdca.cfg
```diff
@@ -8,6 +8,13 @@ HOSTS:
     platform: el-6-x86_64
     template: Delivery/Quality Assurance/Templates/vCloud/centos-6-x86_64
     hypervisor: vcloud
+    puppetbin: /opt/puppetlabs/agent/bin/puppet
+    puppetbindir: /opt/puppetlabs/agent/bin
+    puppetvardir: /opt/puppetlabs/agent/cache
+    puppetpath: /etc/puppetlabs/agent
+    facterbindir: /opt/puppetlabs/agent/bin
+    hierabindir: /opt/puppetlabs/agent/bin
+
```
waynr: The best place for these variables to be set would be https://github.com/puppetlabs/puppet-server/blob/master/acceptance/config/beaker/options.rb

This A) reduces the need for duplication of these settings in other beaker config files and B) ensures that these values are available for CI consumption.

The reason for B) is that the CI jobs dynamically generate Beaker host.cfg using the `genconfig2` utility provided by `sqa-utils`.

acceptance/lib/helper.rb
rlinehan: Yeah, I can add comments here and in any other places where there's things we'll eventually want to change/remove and make tickets.

acceptance/lib/helper.rb
```diff
@@ -193,6 +193,7 @@ def puppet_apply_as_puppet_user
     # installed.
     manifest_path = master.tmpfile("puppetserver_manifest.pp")
     herp_path = master.tmpfile("herp")
+    bin_dir = "/opt/puppetlabs/agent/bin/"
```
waynr: This should be available in the options['puppetbindir'] according to https://github.com/puppetlabs/puppet-server/pull/378/files#diff-3c9fc7388a83ed36154a42ca88d77a1cR12, shouldn't it?

waynr: Also, if you look at `beaker`'s 2.3.0 tag:
https://github.com/puppetlabs/beaker/blob/beaker2.3.0/lib/beaker/host/unix.rb#L61

This value is already available on the host defaults.

waynr: (this is why I was suggesting that working off of beaker's 2.3.0 tag might help with the current effort of getting puppet-server runnning)

waynr: D'oh! I the same value is also in beaker 2.2.0:
https://github.com/puppetlabs/beaker/blob/beaker2.2.0/lib/beaker/host/unix.rb#L61

I just know that 2.3.0 includes additonal AIO magic.

acceptance/config/beaker/jenkins/el-6-x86_64-mdca.cfg
waynr: Okay apologies for my confusing barrage of comments, but given the availability of these same values in Beaker 2.2.0 and later, I would suggest using those instead of setting project-specific values anywhere in the puppet-server repo.

acceptance/lib/helper.rb
waynr: One final comment...it should be `host['puppetbindir']` rather than `options['puppetbindir']`

resources/ext/ezbake.conf
```diff
@@ -20,22 +19,42 @@ ezbake: {
                preinst: ["install --group={{user}} --owner={{user}} -d /var/lib/puppet/jruby-gems",
                          "mkdir -p /etc/puppet/manifests"],
```
rlinehan: @haus just noticed this... `/etc/puppet/manifests` is probably no longer the right directory, right?

haus: yea, is not.

acceptance/lib/helper.rb
rlinehan: I think those values are only available from beaker if `type => aio` is set. Not sure that we want to do that because I don't know what other issues that might cause?

resources/ext/ezbake.conf
haus: not sure what this is going to shake out to be based on code/config separation. /etc/puppetlabs/agent/manifests is probably a sane choice in the meantime.

acceptance/lib/helper.rb
waynr: It seems like `aio` is the style of agent we are running on this branch of puppet-server based on the fact that you have to implement these hacks to get it working...

I could be off-base though, @haus and @justinstoller probably know more than I on this subject. It seems to me that setting `type => aio` will probably be necessary before all is said and done.

haus: I have no idea what `type => aio` implies in this world.

acceptance/lib/helper.rb
```diff
@@ -28,7 +28,7 @@ def self.initialize_config(options)
 
     puppet_build_version = get_option_value(options[:puppet_build_version],
                          nil, "Puppet Development Build Version",
-                         "PUPPET_BUILD_VERSION", "3.7.3")
+                         "PUPPET_BUILD_VERSION", "4.0.0")
```
rlinehan: this probably isn't the right build version, since I don't think puppet-agent is going to be 4.0.0, but I'm not sure what the right default is.

resources/ext/ezbake.conf
```diff
@@ -7,8 +7,7 @@
 ezbake: {
    pe: {}
    foss: {
-      redhat: { dependencies: ["puppet >= 3.7.3"
-                               "puppet < 4.0.0"
+      redhat: { dependencies: ["puppet-agent >= 0.2.0"
```
rlinehan: Do we actually need to specify a version here? Or will just `puppet-agent` work?

haus: puppet-agent should just work. we may want to add a ceiling here at some point to insulate against breaking changes in the agent.

acceptance/lib/helper.rb
waynr: In the past we have tried to keep this version string syncronized with the version specified in `ezbake.conf` so in this case it would probably be `0.2.0`

PR #379, opened 2015-02-02
(SERVER-206) Close client on scripting container termination

fpringvaldsen:
Add in a new `terminate` method to the master that closes the HTTP client and call it before terminating the JRuby scripting container.

I'm definitely looking for some feedback on this one. In particular, I'm not sure about testing. I was having trouble figuring out how to test this. I did add a test to ensure that the master termination function works as intended, but wasn't sure how to write a test to ensure a client is closed before we destroy its scripting container, so any feedback or suggestions on that front are very much appreciated.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
@@ -83,7 +83,7 @@
       (let [id (inc i)
             instance (jruby-core/borrow-from-pool (:pool old-pool))]
         (try
-          (.runScriptlet (:scripting-container instance) "Puppet::Server::Master.terminate")
+          (.terminate (:jruby-puppet instance))
           (.terminate (:scripting-container instance))
           (log/infof "Cleaned up old JRuby instance %s of %s, creating replacement."
                      id count)
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
@@ -83,6 +83,7 @@
       (let [id (inc i)
             instance (jruby-core/borrow-from-pool (:pool old-pool))]
         (try
+          (.runScriptlet (:scripting-container instance) "Puppet::Server::Master.terminate")
           (.terminate (:scripting-container instance))
           (log/infof "Cleaned up old JRuby instance %s of %s, creating replacement."
                      id count)
--- a/src/ruby/puppet-server-lib/puppet/server/http_client.rb
+++ b/src/ruby/puppet-server-lib/puppet/server/http_client.rb
@@ -12,9 +12,6 @@ SyncHttpClient = com.puppetlabs.http.client.Sync
 
 class Puppet::Server::HttpClient
 
-  attr_reader :client
-  @@client = nil
-
   OPTION_DEFAULTS = {
       :use_ssl => true,
       :verify => nil,
@@ -37,8 +34,8 @@ class Puppet::Server::HttpClient
 
     @server = server
     @port = port
-    @use_ssl = options[:use_ssl]
-    @protocol = @use_ssl ? "https" : "http"
+    self.class.use_ssl = options[:use_ssl]
+    @protocol = self.class.use_ssl ? "https" : "http"
   end
 
   def post(url, body, headers, options = {})
@@ -58,13 +55,11 @@ class Puppet::Server::HttpClient
     # Ensure multiple requests are not made on the same connection
     headers["Connection"] = "close"
 
-    create_client_if_nil
-
     request_options = RequestOptions.new(build_url(url))
     request_options.set_headers(headers)
     request_options.set_as(ResponseBodyType::TEXT)
     request_options.set_body(body)
-    response = @@client.post(request_options)
+    response = self.class.client_post(request_options)
     ruby_response(response)
   end
 
@@ -72,28 +67,26 @@ class Puppet::Server::HttpClient
     # Ensure multiple requests are not made on the same connection
     headers["Connection"] = "close"
 
-    create_client_if_nil
-
     request_options = RequestOptions.new(build_url(url))
     request_options.set_headers(headers)
     request_options.set_as(ResponseBodyType::TEXT)
-    response = @@client.get(request_options)
+    response = self.class.client_get(request_options)
     ruby_response(response)
   end
 
   def self.terminate
-    unless @@client.nil?
-      @@client.close
+    unless self.client.nil?
+      self.client.close
     end
   end
 
   private
 
-  def configure_ssl(request_options)
-    return unless @use_ssl
+  def self.configure_ssl(request_options)
+    return unless self.use_ssl
     request_options.set_ssl_context(Puppet::Server::Config.ssl_context)
 
-    settings = self.class.settings
+    settings = self.settings
     if settings.has_key?("ssl_protocols")
       request_options.set_ssl_protocols(settings["ssl_protocols"])
     end
@@ -132,11 +125,29 @@ class Puppet::Server::HttpClient
     result
   end
 
-  def create_client_if_nil
-    if @@client.nil?
-      client_options = ClientOptions.new
-      configure_ssl(client_options)
-      @@client = SyncHttpClient.createClient(client_options)
-    end
+  def self.create_client
+    client_options = ClientOptions.new
+    self.configure_ssl(client_options)
+    SyncHttpClient.createClient(client_options)
+  end
+
+  def self.client
+    @client ||= create_client
+  end
+
+  def self.client_post(request_options)
+    self.client.post(request_options)
+  end
+
+  def self.client_get(request_options)
+    self.client.get(request_options)
+  end
+
+  def self.use_ssl=(use_ssl)
+    @use_ssl = use_ssl
+  end
+
+  def self.use_ssl
+    @use_ssl
   end
 end
--- a/src/ruby/puppet-server-lib/puppet/server/http_client.rb
+++ b/src/ruby/puppet-server-lib/puppet/server/http_client.rb
@@ -34,8 +34,8 @@ class Puppet::Server::HttpClient
 
     @server = server
     @port = port
-    self.class.use_ssl = options[:use_ssl]
-    @protocol = self.class.use_ssl ? "https" : "http"
+    @use_ssl = options[:use_ssl]
+    @protocol = @use_ssl ? "https" : "http"
   end
 
   def post(url, body, headers, options = {})
@@ -83,10 +83,10 @@ class Puppet::Server::HttpClient
   private
 
   def self.configure_ssl(request_options)
-    return unless self.use_ssl
     request_options.set_ssl_context(Puppet::Server::Config.ssl_context)
 
     settings = self.settings
+
     if settings.has_key?("ssl_protocols")
       request_options.set_ssl_protocols(settings["ssl_protocols"])
     end
@@ -143,11 +143,4 @@ class Puppet::Server::HttpClient
     self.client.get(request_options)
   end
 
-  def self.use_ssl=(use_ssl)
-    @use_ssl = use_ssl
-  end
-
-  def self.use_ssl
-    @use_ssl
-  end
 end
--- a/src/ruby/puppet-server-lib/puppet/server/http_client.rb
+++ b/src/ruby/puppet-server-lib/puppet/server/http_client.rb
@@ -11,7 +11,9 @@ java_import com.puppetlabs.http.client.ResponseBodyType
 SyncHttpClient = com.puppetlabs.http.client.Sync
 
 class Puppet::Server::HttpClient
+
   attr_reader :client
+  @@client = nil
 
   OPTION_DEFAULTS = {
       :use_ssl => true,
@@ -62,7 +64,7 @@ class Puppet::Server::HttpClient
     request_options.set_headers(headers)
     request_options.set_as(ResponseBodyType::TEXT)
     request_options.set_body(body)
-    response = @client.post(request_options)
+    response = @@client.post(request_options)
     ruby_response(response)
   end
 
@@ -75,8 +77,14 @@ class Puppet::Server::HttpClient
     request_options = RequestOptions.new(build_url(url))
     request_options.set_headers(headers)
     request_options.set_as(ResponseBodyType::TEXT)
-    response = @client.get(request_options)
+    response = @@client.get(request_options)
     ruby_response(response)
+  end
+
+  def self.terminate
+    unless @@client.nil?
+      @@client.close
+    end
   end
 
   private
@@ -125,10 +133,10 @@ class Puppet::Server::HttpClient
   end
 
   def create_client_if_nil
-    if @client.nil?
+    if @@client.nil?
       client_options = ClientOptions.new
       configure_ssl(client_options)
-      @client = SyncHttpClient.createClient(client_options)
+      @@client = SyncHttpClient.createClient(client_options)
     end
   end
 end
--- a/test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
+++ b/test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
@@ -82,6 +82,11 @@
                                     port use-ssl?)))
      sc)))
 
+(defn terminate-scripting-container
+  [scripting-container]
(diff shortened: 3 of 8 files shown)


cprice404:
@fpringvaldsen it looks like these failed on travis... I'll kick travis so it can try again but it looks like potentially a real issue.

src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
```diff
@@ -83,6 +83,7 @@
       (let [id (inc i)
             instance (jruby-core/borrow-from-pool (:pool old-pool))]
         (try
+          (.runScriptlet (:scripting-container instance) "Puppet::Server::Master.terminate")
```
cprice404: This is a nitpick, but I think that it might be better to add the 'terminate' method to the interface defined in `JRubyPuppet.java`, and then you can call that method directly on the `:jruby-puppet` object.  I'd prefer to try to minimize places where we're interacting directly with the `ScriptingContainer` as it is a less controlled code path.

src/ruby/puppet-server-lib/puppet/server/http_client.rb
```diff
@@ -11,7 +11,9 @@
 SyncHttpClient = com.puppetlabs.http.client.Sync
 
 class Puppet::Server::HttpClient
+
   attr_reader :client
+  @@client = nil
```
cprice404: I think we can get away without having to use this `@@` variable here; see subsequent comments.  I've had pretty bad experiences with those variables in the past.

src/ruby/puppet-server-lib/puppet/server/http_client.rb
```diff
@@ -62,7 +64,7 @@ def post(url, body, headers, options = {})
     request_options.set_headers(headers)
     request_options.set_as(ResponseBodyType::TEXT)
     request_options.set_body(body)
-    response = @client.post(request_options)
+    response = @@client.post(request_options)
```
cprice404: This could be changed to call a class method, e.g. `self.client_post(request_options)`.

The implementation of that class method might look something like this:

```
def self.client_post(request_options)
   self.client.post(request_options)
end

def self.client()
   @client ||= self.create_client
end
```

and then `self.create_client` would contain the logic that actually constructs the client instance.

src/ruby/puppet-server-lib/puppet/server/http_client.rb
```diff
@@ -75,10 +77,16 @@ def get(url, headers)
     request_options = RequestOptions.new(build_url(url))
     request_options.set_headers(headers)
     request_options.set_as(ResponseBodyType::TEXT)
-    response = @client.get(request_options)
+    response = @@client.get(request_options)
```
cprice404: same here; this would probably be better as `self.client_get`, or just `self.client.get`.  The latter is probably fine and then you could skip the step I suggested above about defining a `self.client_post` method, and just use `self.client`.

src/ruby/puppet-server-lib/puppet/server/http_client.rb
```diff
@@ -75,10 +77,16 @@ def get(url, headers)
...
     ruby_response(response)
   end
 
+  def self.terminate
+    unless @@client.nil?
+      @@client.close
```
cprice404: Same here, both of these can just be `self.client`.

src/ruby/puppet-server-lib/puppet/server/http_client.rb
```diff
@@ -125,10 +133,10 @@ def ruby_response(response)
   end
 
   def create_client_if_nil
-    if @client.nil?
+    if @@client.nil?
```
cprice404: If you use the `||=` in the `self.client` method above, then this nil check can go away.  The `||=` pattern is much more idiomatic in Rubyland.

src/ruby/puppet-server-lib/puppet/server/http_client.rb
```diff
@@ -125,10 +133,10 @@ def ruby_response(response)
...
       client_options = ClientOptions.new
       configure_ssl(client_options)
-      @client = SyncHttpClient.createClient(client_options)
+      @@client = SyncHttpClient.createClient(client_options)
```
cprice404: This whole method should become a class method; i.e. `def self.create_client`.  Then you don't need the double-@ here.

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -207,7 +225,10 @@
             (is (true? false) "Expected HTTP connection to HTTPS port to fail")
             (catch EvalFailedException e
               (is (instance? HttpClientException (.. e getCause)))
-              (is (ssl-connection-exception? (.. e getCause getCause)))))))
+              (is (ssl-connection-exception? (.. e getCause getCause)))))
+          (.runScriptlet sc "Puppet::Server::HttpClient.terminate")
+          (.terminate sc)))
```
cprice404: Would probably be best to DRY up this pattern by creating a utility function that does the two terminate calls and then re-using that across all of the tests.

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -215,25 +236,43 @@
...
+          (.terminate sc)))))
+  (testing "master's terminate function closes the client"
+    (logutils/with-test-logging
+      (jetty9/with-test-webserver ring-app-connection-closed port
+        (let [sc (create-scripting-container port)]
+          (.runScriptlet sc "response = c.get('/', {})")
+          (is (= "200" (.runScriptlet sc "response.code")))
+          (.runScriptlet sc "require 'puppet/server/master'; Puppet::Server::Master.terminate")
+          (is (thrown? EvalFailedException
+                       (.runScriptlet sc "response = c.get('/', {})")))
```
cprice404: Does the exception you're catching here give you any more info you can use in your assertion?  e.g. does it have a nested exception of a more specific type, or if you call `getMessage` on it can you then match that string in your assertion?

test/unit/puppetlabs/services/jruby/jruby_testutils.clj
```diff
@@ -86,7 +86,9 @@
                    {:pool                 pool
                     :id                   1
                     :jruby-puppet         (create-mock-jruby-instance)
-                    :scripting-container  (ScriptingContainer. LocalContextScope/SINGLETHREAD)
+                    :scripting-container (doto (ScriptingContainer. LocalContextScope/SINGLETHREAD)
+                                               (jruby-core/prep-scripting-container ruby-load-path gem-home)
+                                               (.runScriptlet "require 'puppet/server/master'"))
```
cprice404: I don't think we want to do this here; @dankreek originally put this in because loading the puppet code into the JRuby containers makes the tests take about 10x as long.  If you need this for a specific test, it's probably best to just put the logic directly in that test namespace rather than in the utility function.

cprice404:
This all looks good to me.  Just the nitpicks above about making the Ruby a little more idiomatic, DRYing up a few things, and not introducint all of the puppet loading into the jruby-testutils.

fpringvaldsen:
@cprice404 I've addressed your comments.

cprice404:
+1, would like to get @camlow325 's eyes on this as well, though.

src/ruby/puppet-server-lib/puppet/server/http_client.rb
```diff
@@ -35,8 +34,8 @@ def initialize(server, port, options = {})
 
     @server = server
     @port = port
-    @use_ssl = options[:use_ssl]
-    @protocol = @use_ssl ? "https" : "http"
+    self.class.use_ssl = options[:use_ssl]
```
camlow325: What would happen if the first request made through the client were for http (use_ssl = false) but at least one subsequent one were for https (use_ssl = true)?  Could we end up initializing a client without configuring an SSLContext and try to use that for the https calls?

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -151,28 +160,35 @@
              (is (true? false) "Expected HTTP connection to HTTPS port to fail")
              (catch EvalFailedException e
                (is (instance? HttpClientException (.. e getCause)))
-               (is (ssl-connection-exception? (.. e getCause getCause)))))))))
+               (is (ssl-connection-exception? (.. e getCause getCause))))))
+         (.runScriptlet sc "Puppet::Server::HttpClient.terminate")
```
camlow325: Could this one use "terminate-scripting-container"?

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -151,28 +160,35 @@
...
+         (.terminate sc))))
 
   (testing "Can connect via TLSv1.1 by default"
     (with-webserver-with-protocols ["TLSv1.1"] nil
        (let [sc (create-scripting-container-with-ssl-client 10080)]
          (.runScriptlet sc "response = c.get('/', {})")
          (is (= "200" (.runScriptlet sc "response.code")))
-         (is (= "hi" (.runScriptlet sc "response.body"))))))
+         (is (= "hi" (.runScriptlet sc "response.body")))
+         (.runScriptlet sc "Puppet::Server::HttpClient.terminate")
```
camlow325: Same question here.-

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -151,28 +160,35 @@
...
+         (.terminate sc))))
 
   (testing "Can connect via TLSv1 by default"
     (with-webserver-with-protocols ["TLSv1"] nil
        (let [sc (create-scripting-container-with-ssl-client 10080)]
          (.runScriptlet sc "response = c.get('/', {})")
          (is (= "200" (.runScriptlet sc "response.code")))
-         (is (= "hi" (.runScriptlet sc "response.body"))))))
+         (is (= "hi" (.runScriptlet sc "response.body")))
+         (.runScriptlet sc "Puppet::Server::HttpClient.terminate")
```
camlow325: Same question here.

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -215,25 +235,44 @@
...
       (jetty9/with-test-webserver ring-app port
-        (let [scripting-container (create-scripting-container port)
-              client1 (.runScriptlet scripting-container "c.get('/', {}); c.client")
-              client2 (.runScriptlet scripting-container "c.post('/', 'foo', {}); c.client")]
-          (is (= client1 client2)))))))
+        (let [sc (create-scripting-container port)
+              client1 (.runScriptlet sc "c.get('/', {}); c.class.client")
+              client2 (.runScriptlet sc "c.post('/', 'foo', {}); c.class.client")]
+          (is (= client1 client2))
+          (terminate-scripting-container sc))))))
```
camlow325: Since we missed this the first time, I think it would be good to create a second test here where we initialize another `Puppet::Server::HttpClient` instance and validate that its ".class.client" is identical to the ".class.client" from the original `Puppet::Server::HttpClient` instance.

src/ruby/puppet-server-lib/puppet/server/http_client.rb
fpringvaldsen: @camlow325 Hmm, that's a good point. Since we're having a static underlying client object, that client can't change across instances of the `HttpClient` class, so my guess would be the client would try to make an https call without configuring an SSLContext. 

I'm wondering how we could handle this. Maybe we could regenerate the client whenever the value of `use_ssl` is changed?

camlow325: Would it make sense to just create the underlying Apache client with the SSL context always and just have "use_ssl" be used at the instance-level -- and not at the class-level at all -- to decide how to set the protocol on each client instance?  I think the SSL context would just be unused for plain-text "http" calls, which wouldn't seem to hurt anything.

camlow325: Also, I think we should add some sort of test for "http" and "https" calls off the same client to the unit tests below to prove that this will work.  :)

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -215,25 +232,44 @@
...
+          (terminate-scripting-container sc)))))
+  (testing "client's terminate function closes the client"
+    (logutils/with-test-logging
+      (jetty9/with-test-webserver ring-app-connection-closed port
+        (let [sc (create-scripting-container port)]
+          (.runScriptlet sc "response = c.get('/', {})")
+          (is (= "200" (.runScriptlet sc "response.code")))
+          (.runScriptlet sc "c.class.terminate")
+          (try
+            (.runScriptlet sc "response = c.get('/', {})")
```
camlow325: I like this test.  :smile:

camlow325:
:+1: other than the "use_ssl" question and making "http" and "https" requests through the same client instance.

src/ruby/puppet-server-lib/puppet/server/http_client.rb
fpringvaldsen: @camlow325 If we do that, then the user will always have to specify a `hostcert`, a `hostprivkey`, and a `localcacert`, even if they're not making an https request. That's currently breaking our tests, which can easily be fixed, but will this break other projects creating instances of our client?

camlow325: The current logic derives the `hostcert`, `hostprivkey`, and `localcacert` from the global Puppet settings, right -- i.e., not passed in on a per request basis to a `Puppet::Server::HttpClient` instance?  I believe those settings will always have a non-empty value, with a default imposed by core Ruby Puppet if the user hasn't overridden it in their puppet.conf file.

I had been thinking that clients creating `Puppet::Server::HttpClient` instances just specify whether or not SSL needs to be used and not what certs and keys to use.  I'm not suggesting that this change for this PR.

Maybe the only concerning case would be for the case that Puppet Server were being configured to run via HTTP -- the external SSL termination case.  If it is possible for Puppet to come to ready with the values for `hostcert`, `hostprivkey`, and `localcacert` pointing to files that don't actually exist then trying to configure the SSL context for HTTP client calls would be problematic.  If that were the case, maybe the client logic could be made more robust here by checking that all of the files exist before trying to configure the SSL context and, if they do not, outputting a log warning and skipping the SSL context configuration.  If a client were to later try to make a request with a "use_ssl" of true, Puppet Server could fail the request with an error indicating that the context could not be configured because certs/keys were missing.  I'm thinking this would have been an issue with this code prior to all of the work done to optimize the number of clients being created, though.  If this isn't a valid use case, adding in this extra robustness wouldn't seem to be necessary.

fpringvaldsen: @camlow325 Ah, okay. I mostly wasn't sure about whether or not those settings would always be set. I'll go ahead and fix the tests then, and handle that use case if @cprice404 says it's valid.

fpringvaldsen:
@camlow325 @cprice404 Feedback addressed! The underlying Java client is now always configured with an SSL context. I've also added some more tests.

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -142,98 +151,144 @@
...
+          (let [sc (create-scripting-container 10080 {:use-ssl true})]
+            (.runScriptlet sc "response = c.get('/', {})")
+            (is (= "200" (.runScriptlet sc "response.code")))
+            (is (= "hi" (.runScriptlet sc "response.body")))
+            (.runScriptlet sc (str (format "c = Puppet::Server::HttpClient.new('localhost', %d, {:use_ssl => %s});"
+                                           port false)
+                                   "response = c.get('/', {})"))
+            (is (= "200" (.runScriptlet sc "response.code")))
+            (is (= "bye" (.runScriptlet sc "response.body")))
+            (terminate-scripting-container sc)))))))
```
camlow325: I think this is a good test although what led to the creation of this test was the concern that if an HTTP request is made before the HTTPS one that the HTTPS one would fail because the SSLContext wasn't configured when the client was first created.  I think the currently written test covers HTTPS before HTTP.  Might be good to add a second HTTP before HTTPS test to ensure both cases are covered.

fpringvaldsen:
@camlow325 I've added the extra test.

camlow325:
:+1:

PR #380, opened 2015-02-02
(SERVER-332) Allow all TLS algorithms in the dev lein profile

jeffmccune:
Without this patch the unit tests fail on JVM versions more recent than
1.7.0_u75 because SSLv3 is disabled at the JRE layer by default.  This is a
problem because the end user may explicitly configure the JRE to allow SSLv3
algorithms, even though we disable them by default at the application layer,
and we need to test the behavior of the system in this scenario.  This patch
addresses the problem by allowing all SSLv3 algorithms by overriding the
java.security jdk.tls.disabledAlgorithms property.

### DIFF
--- a/documentation/configuration.markdown
+++ b/documentation/configuration.markdown
@@ -195,3 +195,34 @@ a multi-master environment or using an external CA, you might want to disable
 the CA service on some nodes.
 
 
+## Java Secure Socket Extension (JSSE) configuration
+
+A result of the [POODLE](https://blogs.oracle.com/security/entry/information_about_ssl_poodle_vulnerability)
+is that SSLv3 is disabled by default at the JRE layer and in Puppet Server.
+It is possible to enable Puppet Server to negotiate with SSLv3 clients,
+however the recommended course of action is to upgrade clients to negotiate
+using secure protocols.
+
+SSLv3 has been disabled by default in javase 7u75 (1.7.0_u75).  See [7u75
+Update Release Notes](http://www.oracle.com/technetwork/java/javase/7u75-relnotes-2389086.html)
+for more information.
+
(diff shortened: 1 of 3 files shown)


dev/java.security
```diff
@@ -0,0 +1,44 @@
...
+# Note: This property is currently used by Oracle's JSSE implementation.
+# It is not guaranteed to be examined and used by other implementations.
+#
+# Example:
+#   jdk.tls.disabledAlgorithms=MD5, SSLv3, DSA, RSA keySize < 2048
+#
+# SERVER-332 Disable no algorithms so that unit tests are able to exercise the
+# behavior of the system when the end user explicitly configures deprecated
+# algorithms like SSLv3.
+jdk.tls.disabledAlgorithms=
```
cprice404: If we need this properties file, I think it might be better to put it in "dev-resources".  I think of that directory as containing files that are needed by tests and won't ever be changed by users, whereas I think of the "dev" directory as a place where users go to look for sample configs that they might modify, etc.

jeffmccune:
@camlow325 mentioned it would be good to have new documentation about adjusting JAVA_OPTS if this is required in production.

camlow325:
:+1:

cprice404:
That travis failure was weird but I'm +1 on this.

documentation/configuration.markdown
```diff
@@ -195,3 +195,34 @@ a multi-master environment or using an external CA, you might want to disable
 the CA service on some nodes.
 
 
+## Java Secure Socket Extension (JSSE) configuration
+
```
jeffmccune: @nfagerlund FYI, we've added a note to the puppet server configuration documentation about how to re-enable SSLv3, if required.

Re-enabling SSLv3 is not advised, at all, due to POODLE.

PR #381, opened 2015-02-03
(SERVER-326) Add /v3/environments route

fpringvaldsen:
Add the /environments route to the v3 routes, as it previously
existed in the v2 routes.

### DIFF
--- a/test/unit/puppetlabs/services/master/master_core_test.clj
+++ b/test/unit/puppetlabs/services/master/master_core_test.clj
@@ -11,6 +11,7 @@
         app         (build-ring-handler handler)
         request     (fn r ([path] (r :get path))
                           ([method path] (app (mock/request method path))))]
+    (is (= 200 (:status (request "/v3/environments"))))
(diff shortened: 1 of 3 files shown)


test/unit/puppetlabs/services/master/master_core_test.clj
```diff
@@ -11,6 +11,7 @@
         app         (build-ring-handler handler)
         request     (fn r ([path] (r :get path))
                           ([method path] (app (mock/request method path))))]
+    (is (= 200 (:status (request "/v3/environments"))))
```
nwolfe: Looks like we could just add `"environments"` to the sequence below this

fpringvaldsen: @nwolfe I thought about that, but then we'd have to special snowflake it since all the other routes have requests made to `/v3/path/foo', so I thought it would be better to just not add it to that sequence.

rlinehan:
:+1: This looks good to me and fixes one of the failing puppet acceptance tests.

PR #382, opened 2015-02-03
(SERVER-332) Allow all TLS algorithms in the dev lein profile

camlow325:
Without this patch the unit tests fail on JVM versions more recent than
1.7.0_u75 because SSLv3 is disabled at the JRE layer by default.  This
is a problem because the end user may explicitly configure the JRE to
allow SSLv3 algorithms, even though we disable them by default at the
application layer, and we need to test the behavior of the system in
this scenario.  This patch addresses the problem by allowing all SSLv3
algorithms by overriding the java.security jdk.tls.disabledAlgorithms
property.

This patch employs an override file rather than passing
-Djdk.tls.disabledAlgorithms= as a java option because passing the java
option doesn't work.  Similarly, passing
`-Ddeployment.security.SSLv3=true` as a java option as mentioned at [1]
does not have the desired effect of enabling SSLv3.

The properties file is located in `dev-resources` because the directory
contains files needed by the tests and shouldn't ever be changed by end
users.

[1]: http://www.oracle.com/technetwork/java/javase/7u75-relnotes-2389086.html

### DIFF
--- a/dev-resources/java.security
+++ b/dev-resources/java.security
@@ -1,0 +1,44 @@
+#
+# This is the "override security properties file" which is used by default
+# in the lein dev profile.  End users may override java security properties in
+# a similar manner in the production code.
+#
+# This file augments and overrides $JAVA_HOME/jre/lib/security/java.security
+# when the java process is provided the option,
+# -Djava.security.properties=./dev/java.security
+#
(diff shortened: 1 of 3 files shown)


jeffmccune:
This has already been reviewed in #380 so merging into stable and up into master.

PR #383, opened 2015-02-03
(SERVER-334) Generate upgrade message once

fpringvaldsen:
Generate the 404 upgrade message once instead of on every request
handled, since the information displayed should not change and
grabbing the Puppet Version from JRuby is an expensive operation.


cprice404:
+1

cprice404:
@jeffmccune since you reviewed the original PR would you mind taking a quick look at this and hitting the button if it looks OK to you?

jeffmccune:
+1

PR #386, opened 2015-02-03
(maint) Update packaging & acceptance tests to create puppet user

rlinehan:
Previously, the puppet package created the puppet user and group. However, the
puppet agent does not require this user, since it runs as root. The puppet
user is only required for the master. The puppet-agent AIO packages will not
be creating the puppet user or group, so the puppetserver packages must now do
this and must also set proper permissions on certain directories. This commit
adds the necessary steps to the postinstall.

It also removes the `puppet_apply_as_puppet_user` method from the acceptance
helpers, which 1) is no longer required, and 2) since the puppet user is not
created before installing the puppetserver package, we couldn't complete this
step.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -189,43 +189,6 @@ EOF
 
   end
 
-  def puppet_apply_as_puppet_user
-    # When puppet apply is run for the first time, certain directories are
-    # created as the user owning the process. In order to avoid creating these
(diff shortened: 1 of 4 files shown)


waynr:
:+1: This generally looks good to me, I know there were previously concerns on the team about the `puppet_apply_as_puppet_user` functionality. As long as all the necessary changes are being made in the postinst steps then it should be safe to remove.

PR #388, opened 2015-02-04
(SERVER-157) use the keylength settings from puppet.conf

masterzen:
Caution: this is a PR hacked at vey high speed during the Ghent Puppet Contributor Summit after several conference days, so it might have many issues :)

This new feature makes the puppet-server respect the puppet.conf keylength parameter when creating new certificates.

Unlike what suggested in the original ticket, there's no specific error message about invalid keylength as the jvm current error message looked sufficient. 

Please review and feel free to comment!
Brice

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -30,6 +30,7 @@
    :hostcrl        schema/Str
    :hostprivkey    schema/Str
    :hostpubkey     schema/Str
+   :keylength      schema/Int
    :localcacert    schema/Str
    :requestdir     schema/Str
    :csr-attributes schema/Str})
@@ -57,6 +58,7 @@
    :ca-ttl                   schema/Int
    :cert-inventory           schema/Str
    :csrdir                   schema/Str
+   :keylength                schema/Int
    :ruby-load-path           [schema/Str]
    :signeddir                schema/Str
    :serial                   schema/Str})
@@ -176,6 +178,7 @@
           :autosign
           :ca-name
           :ca-ttl
+          :keylength
           :ruby-load-path))
 
 (schema/defn settings->ssldir-paths
@@ -183,7 +186,7 @@
    paths. These paths are necessary during initialization for determining what
    needs to be created and where."
   [master-settings :- MasterSettings]
-  (dissoc master-settings :dns-alt-names :csr-attributes))
+  (dissoc master-settings :dns-alt-names :csr-attributes :keylength))
 
 (defn path-to-cert
   "Return a path to the `subject`s certificate file under the `signeddir`."
@@ -362,15 +365,14 @@
 (schema/defn generate-ssl-files!
   "Given the CA settings, generate and write to disk all of the necessary
   SSL files for the CA. Any existing files will be replaced."
-  [ca-settings :- CaSettings
-   keylength :- schema/Int]
+  [ca-settings :- CaSettings]
   (log/debug (str "Initializing SSL for the CA; settings:\n"
                   (ks/pprint-to-string ca-settings)))
   (create-parent-directories! (vals (settings->cadir-paths ca-settings)))
   (-> ca-settings :csrdir fs/file ks/mkdirs!)
   (-> ca-settings :signeddir fs/file ks/mkdirs!)
   (initialize-serial-file! (:serial ca-settings))
-  (let [keypair     (utils/generate-key-pair keylength)
+  (let [keypair     (utils/generate-key-pair (:keylength ca-settings))
         public-key  (utils/get-public-key keypair)
         private-key (utils/get-private-key keypair)
         x500-name   (utils/cn (:ca-name ca-settings))
@@ -540,8 +542,7 @@
    the master. Any existing files will be replaced."
   [settings :- MasterSettings
    certname :- schema/Str
-   ca-settings :- CaSettings
-   keylength :- schema/Int]
+   ca-settings :- CaSettings]
   (log/debug (str "Initializing SSL for the Master; settings:\n"
                   (ks/pprint-to-string settings)))
   (create-parent-directories! (vals (settings->ssldir-paths settings)))
@@ -551,7 +552,7 @@
         ca-public-key  (.getPublicKey ca-cert)
         ca-private-key (utils/pem->private-key (:cakey ca-settings))
         next-serial    (next-serial-number! (:serial ca-settings))
-        keypair        (utils/generate-key-pair keylength)
+        keypair        (utils/generate-key-pair (:keylength settings))
         public-key     (utils/get-public-key keypair)
         extensions     (create-master-extensions certname
                                                  public-key
@@ -579,12 +580,9 @@
   "Given configuration settings, certname, and CA settings, ensure all
    necessary SSL files exist on disk by regenerating all of them if any
    are found to be missing."
-  ([settings certname ca-settings]
-     (initialize-master-ssl! settings certname ca-settings utils/default-key-length))
-  ([settings :- MasterSettings
+  [settings :- MasterSettings
     certname :- schema/Str
-    ca-settings :- CaSettings
-    keylength :- schema/Int]
+    ca-settings :- CaSettings]
      (let [required-files (-> (settings->ssldir-paths settings)
                               (select-keys required-master-files)
                               (vals))]
@@ -593,8 +591,8 @@
          (let [{found   true
                 missing false} (group-by fs/exists? required-files)]
            (if (= required-files missing)
-             (generate-master-ssl-files! settings certname ca-settings keylength)
-             (throw (partial-state-error "master" found missing))))))))
+             (generate-master-ssl-files! settings certname ca-settings)
+             (throw (partial-state-error "master" found missing)))))))
 
 (schema/defn ^:always-validate retrieve-ca-cert!
   "Ensure a local copy of the CA cert is available on disk.  cacert is the base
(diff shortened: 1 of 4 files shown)


KevinCorcoran:
@masterzen  - thanks for this!  (I finally just now saw it.)  Will try to review sooner than your last PR.  :smiling_imp: 

FWIW, @masterzen hacked this up at my suggestion, as an exercise in working on this codebase.  I don't think he actually needs this feature.

fpringvaldsen:
:see_no_evil:

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -349,40 +357,42 @@
...
+                        public-key
+                        ca-exts)
+          cacrl       (utils/generate-crl (.getIssuerX500Principal cacert)
+                                          private-key
+                                          public-key)]
+      (write-cert-to-inventory! cacert (:cert-inventory ca-settings))
+      (utils/key->pem! public-key (:capub ca-settings))
+      (utils/key->pem! private-key (:cakey ca-settings))
+      (utils/cert->pem! cacert (:cacert ca-settings))
+      (utils/crl->pem! cacrl (:cacrl ca-settings)))))
```
cprice404: Were these whitespace changes intentional?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -140,6 +141,12 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Internal
 
+(schema/defn keylength :- schema/Int
+  "Returns either the specified keylength from the settings or the default keylength
+  if none were present in the settings"
+  [settings :- CaSettings]
+  (get settings :keylength utils/default-key-length))
```
cprice404: Someday I'd like to consolidate all of the definitions of the default values for various settings in a single spot, but I don't think we have that for this service yet... so, this seems good for now.

cprice404:
@masterzen sorry it keeps taking us so long to review your PRs; we've been buried in work related to Puppet 4.  Hopefully once we get over that hump we will be able to start being much more responsive.  Thanks for the PR!

This looks pretty good to me, I think we should wait until @nwolfe frees up and has time to review, but I'm good with it if he is.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
masterzen: @cprice404 there's one extra indent because the function now has multiple parameter sets. But it's possible my whitespace settings are not the same as the original file one. Is there a convention?

cprice404: Oh, duh.  :)  That makes sense, sorry I overlooked the function sig change.

masterzen:
@cprice404, there's no emergency, take your time :)

nwolfe:
@masterzen This is actually really awesome! Nice job. I think with a little more work we could get rid of some code and really clean things up.

I'd like to see us get rid of the keylength parameter altogether and have it be part of the settings parameters. The reason it's the only setting that isn't part of that map already is because it wasn't a setting that came from Puppet, so it wasn't able to follow the same plumbing. Obviously we were mistaken - it *is* a setting from Puppet, and we don't have to treat it any differently than all the others!

I'd propose the following:

In the production code...
1. Add `:keylength` to both the `CaSettings` and `MasterSettings`
2. Get rid of all `keylength` function parameters (in the initialize/generate functions) and instead get it out of the appropriate settings map. CA functions should get it from `ca-settings`, and master from `master-settings`.  Note that this means we will no longer need to use `utils/default-key-length` anywhere.

In the test code...
3. Add the default `:keylength 512` to `master-settings` in `ca_testutils`
4. Clean up tests by removing the `512` parameters nearly everywhere to the various initialize/generate functions. Tests that are actually testing this keylength should rely on the default value of `512` or assoc a new value into the settings.

Great job! I really didn't like the special-case handling we've got for the keylength setting, whatever we end up doing will be an improvement.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -59,7 +59,8 @@
    :csrdir                   schema/Str
    :ruby-load-path           [schema/Str]
    :signeddir                schema/Str
-   :serial                   schema/Str})
+   :serial                   schema/Str
+   :keylength                schema/Int})
```
nwolfe: During development I found it very useful to have these settings in alphabetical order.  We might as well stick to it here and put it between `:csrdir` and `:ruby-load-path`.

Same goes for the `MasterSettings` if/when we add it there too.

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -394,6 +394,20 @@
         (doseq [f files] (is (= "testable string" (slurp f))
                              "File was replaced"))))))
 
+(deftest initialize!-test-with-keylength-in-settings
+  (let [settings (assoc (testutils/ca-settings (ks/temp-dir)) :keylength 768)]
```
nwolfe: We have a `ca-sandbox!` function that combines the `(ca-settings (temp-dir ...))` - might use that instead.

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -540,6 +554,45 @@
                  (initialize-master-ssl! settings "master" ca-settings 512)))
             (fs/copy copy path)))))))
 
+(deftest initialize-master-ssl!-test-with-keylength-settings
+  (let [tmp-confdir (fs/copy-dir confdir (ks/temp-dir))
+        settings (-> (testutils/master-settings tmp-confdir "master")
+                     (assoc :dns-alt-names "onefish,twofish"))
+        ca-settings (assoc (testutils/ca-settings (str tmp-confdir "/ssl/ca")) :keylength 768)]
```
nwolfe: Nice job with these tests. I think we should use all of the default values from test util settings, and only specify the ones that we're testing. In this case, we're not testing the `:dns-alt-names` or the master certname, so I'd try to get rid of the `(assoc :dns-alt-names ...)` and the `"master"` parameter to `master-settings`.

We're going to be doing heavy refactoring of these tests at some point and keeping them as simple and straightforward as possible will go a long way.

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -540,6 +554,45 @@
...
+
+    (testing "should throw an error message with too short keylength"
+      (is (thrown-with-msg?
+            java.security.InvalidParameterException
+            #".*RSA keys must be at least 512 bits long.*"
+            (initialize-master-ssl! settings "master" (assoc ca-settings :keylength 128)))))
+
+    (testing "should throw an error message with too large keylength"
+      (is (thrown-with-msg?
+            java.security.InvalidParameterException
```
nwolfe: We've had minor issues in the past with using fully-qualified Java classes without `:import`-ing them in the `ns` at the top. I would probably add an import and then just refer to it as `InvalidParameterException`, but it's not a big deal if we don't.

masterzen:
Thanks @nwolfe for the thorough review. I won't have time to work on this for the next couple of weeks, but I'll try to push a new PR taking into account your suggestions and comments when I come back. 

Feel free to ping me if I forget :)

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -540,6 +554,45 @@
...
+                     (assoc :dns-alt-names "onefish,twofish"))
+        ca-settings (testutils/ca-settings (str tmp-confdir "/ssl/ca"))]
+
+    (retrieve-ca-cert! (:cacert ca-settings) (:localcacert settings))
+
+    (testing "should throw an error message with too short keylength"
+      (is (thrown-with-msg?
+            java.security.InvalidParameterException
+            #".*RSA keys must be at least 512 bits long.*"
+            (initialize-master-ssl! settings "master" (assoc ca-settings :keylength 128)))))
```
nwolfe: Regarding my general comment about putting keylength into the settings maps...if we did that (and it was in *both* CA and master settings) then we would move the assoc over to the `settings` since that's the one that the master would be getting it from in `initialize-master-ssl!`.

nwolfe:
Alright, done reviewing. Again, this will be a great improvement to the code! Really looking forward to getting rid of the special-case support for 1 of like 30 settings :smile: 

> Thanks @nwolfe for the thorough review. I won't have time to work on this for the next couple of weeks, but I'll try to push a new PR taking into account your suggestions and comments when I come back.

Sure, not a problem. It's worth the wait! :sunglasses:

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
masterzen: It's not possible to use `ca-sandbox!` here because we are changing the keylength, so we need to change the public/privage key pair. This is done only if those files don't exist at first, hence we need to recreate the CA from scratch.

masterzen:
Hi,

I've pushed a new revision that should take into account all good comments @nwolfe could make.
Please review at your convenience,
Thanks!
Brice

nwolfe:
@masterzen Nicely done. Thanks for doing all this!

:+1: :+1:

PR #389, opened 2015-02-04
(MAINT) Bump Ruby puppet version

dankreek:
This is needed to maintain acceptance test parity with the new version
of MRI Puppet.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@ source ENV['GEM_SOURCE'] || "https://rubygems.org"
 group :test do
   gem 'rake'
   gem 'rspec'
-  gem 'beaker', '~>1.17.0'
+  gem 'beaker', '~>1.20.0'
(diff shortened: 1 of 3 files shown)


rlinehan:
This looks like the SHA where the PUPPETVERSION was bumped to 3.7.4, so assuming that's what we want, :+1:.

camlow325:
I think Puppet 3.7.4 is used with a newer version of Facter, too, so we probably should bump our submodule for that also.  I think the right version is 2.4.0 - GIT SHA 8ff4fa5e0251aaa1b3fea1e2e308a34b2ffe6027.

dankreek:
@camlow325 Bumped facter up to 2.4.0

camlow325:
I think you'll also need to update the "returns the correct puppet version number" test in .../spec/puppet-server-lib/puppet/jvm/master_spec.rb from '3.7.3' to '3.7.4'.

dankreek:
I dang, I only ran 2 different tests suites before submitting this

camlow325:
:+1:

PR #393, opened 2015-02-05
(SERVER-338) Use lein-ezbake 0.1.2 to resolve netstat issue

jeffmccune:
Without this patch the dependency on the JVM packages are expressed in the
project specific ezbake configuration.  This is a problem because it duplicates
the same dependency expressed statically in the [ezbake package
specifications][requires].  The duplication is confusing and more difficult to
coordinate and maintain over time.

The principle for dependencies is that those required to satisfy the behavior
of ezbake itself, or the system level service management scripts, should be
expressed statically in ezbake since those dependencies will apply to all
projects using ezbake and the resulting scripts.  Application level
dependencies should be specified in the project specific ezbake.conf file such
that individual applications are not interdependent on common dependencies.

[requires]: https://github.com/puppetlabs/ezbake/blob/418e4815edf159c753e148395b3e725f8d202c17/resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/ezbake.spec.erb#L5

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -95,7 +95,7 @@
              :ezbake {:dependencies ^:replace [[puppetlabs/puppet-server ~ps-version]
                                                [puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty-version]
                                                [org.clojure/tools.nrepl "0.2.3"]]
-                      :plugins [[puppetlabs/lein-ezbake "0.1.0"]]
+                      :plugins [[puppetlabs/lein-ezbake "0.1.2"]]
                       :name "puppetserver"}
              :uberjar {:aot [puppetlabs.trapperkeeper.main]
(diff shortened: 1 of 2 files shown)


jeffmccune:
Relates to #ezbake/178

jeffmccune:
This also passed acceptance tests:

```
Cleanup: cleaning up after successful run
Handing 'og1hrysqosn3k1k.delivery.puppetlabs.net' back to pooling API for VM destruction
Spent 0.01 seconds cleaning up
Beaker completed successfully, thanks.
PACKAGE_BUILD_VERSION='2.0.0.SNAPSHOT.2015.02.04T1711' bundle exec beaker      8.91s user 2.18s system 1% cpu 12:23.03 total
```

rlinehan:
:+1: runs successfully on el 7

PR #395, opened 2015-02-11
(SERVER-294) delete Content-Type hack

KevinCorcoran:
Delete a terrible hack which munged the Content-Type header of incoming
/file_bucket_file requests to always be 'application/octet-stream'.
Historically, the agent has sent a value of 'text/plain' for this
header, but that is fixed as of PUP-3812.

I think that the primary risk of a regression here would be to accidentally break handling of non-UTF8 data (as we originally did in SERVER-269) - however, that should be sufficiently covered by the integration test in `puppetlabs.services.jruby.request-handler-test/file-bucket-test`.  So, I think this is good on test coverage.

### DIFF
--- a/test/integration/puppetlabs/services/jruby/request_handler_test.clj
+++ b/test/integration/puppetlabs/services/jruby/request_handler_test.clj
@@ -66,7 +66,7 @@
                options                 (merge ssl-request-options
                                               {:body (ByteArrayInputStream.
                                                        raw-byte-arr)
-                                               :headers {"accept" "application/octet-stream"
+                                               :headers {"accept" "binary"
                                                          "content-type" "application/octet-stream"}})
                response (http-client/put (str "https://localhost:8140/"
                                               "puppet/v3/file_bucket_file/md5/"
--- a/test/integration/puppetlabs/services/jruby/request_handler_test.clj
+++ b/test/integration/puppetlabs/services/jruby/request_handler_test.clj
@@ -63,18 +63,11 @@
                                                        (subs expected-md5 0 8))
                                           expected-md5
                                           "contents"])
-               ;; This accept and content-type header match what the Puppet
-               ;; agent sends to the master for a file-bucket PUT. Ideally,
-               ;; these should both be 'application/octet-stream', but for now
-               ;; we have to live with 'binary' content-type because
-               ;; everything is terrible.
                options                 (merge ssl-request-options
                                               {:body (ByteArrayInputStream.
                                                        raw-byte-arr)
-                                               :headers {"accept"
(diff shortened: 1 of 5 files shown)


test/integration/puppetlabs/services/jruby/request_handler_test.clj
```diff
@@ -63,18 +63,11 @@
...
-               ;; we have to live with 'binary' content-type because
-               ;; everything is terrible.
                options                 (merge ssl-request-options
                                               {:body (ByteArrayInputStream.
                                                        raw-byte-arr)
-                                               :headers {"accept"
-                                                           "binary"
-                                                         "content-type"
-                                                           "application/octet-stream"}})
+                                               :headers {"accept" "application/octet-stream"
```
camlow325: I think this still will have to be *binary* because core Ruby Puppet uses these short-hand representations rather than standard HTTP MIME types for the 'accept' header.  See https://tickets.puppetlabs.com/browse/PUP-3940.

KevinCorcoran: Oops, you're right.  For some reason I initially thought the accept header had been fixed as well.

KevinCorcoran: I just added 30563af97ad7a5eee393bf1f5e82610d2873e4be to address this.

camlow325:
:+1: assuming Travis passes.

fpringvaldsen:
It looks like this is causing some of the file bucket integration tests to fail.

KevinCorcoran:
Yup, this is broken ATM, I've got a fix in progress ...

KevinCorcoran:
Here's what needs to happen to get this PR in:
* https://github.com/puppetlabs/clj-http-client/pull/34 needs to land
* need to ship a new version of clj-http-client with that change
* Then, I'll update this PR to include one more change - bumping the dependency on clj-http-client to the new version

That should fix the test failure (I've tested this locally against a SNAPSHOT) and then this can be merged.

KevinCorcoran:
I added a commit which bumps this up to clj-http-client 0.4.2.

The Travis build failed with `Error: Attempt to borrow a JRuby instance from the pool timed out` on one of the platforms, but passed on the other.  I'm going to assume that's just related to Travis being really slow.

All of that said, I think this can be merged now.

rlinehan:
:+1: as far as I understand the ticket, this looks good to me.

camlow325:
Needs rebase?

KevinCorcoran:
@camlow325 This can now be merged.

PR #397, opened 2015-02-12
(SERVER-354) fix initialization of Puppet logging

KevinCorcoran:
Recently, Puppet was changed to allow the log level to be configured via
settings.  This broke the way in which we were previously initializing
it.  This commit fixes that initialization and adds a ruby spec test to
ensure that logging is initialized correctly.

In addition to the spec test, I also manually verified this change - [here's the log output](https://gist.githubusercontent.com/KevinCorcoran/4b1338a6e0ab1edd3c81/raw/5cc253523ae6ab8bac73c9c9dfcd1f892e4de29f/gistfile1.txt) of starting Puppet Server and doing a no-op agent run against it previously, and [here's the logging](https://gist.githubusercontent.com/KevinCorcoran/704169f20c06efaf86de/raw/e19594eb81ec943f9828b7262c2eaf8f8e993b9a/gistfile1.txt) for the same process but including this patch.

### DIFF
--- a/src/ruby/puppet-server-lib/puppet/server/puppet_config.rb
+++ b/src/ruby/puppet-server-lib/puppet/server/puppet_config.rb
@@ -22,7 +22,7 @@ class Puppet::Server::PuppetConfig
     Puppet[:trace] = true
 
     # Crank Puppet's log level all the way up and just control it via logback.
-    Puppet::Util::Log.level = :debug
+    Puppet[:log_level] = "debug"
 
     master_run_mode = Puppet::Util::RunMode[:master]
     app_defaults = Puppet::Settings.app_defaults_for_run_mode(master_run_mode).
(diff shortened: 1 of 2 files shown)


src/ruby/puppet-server-lib/puppet/server/puppet_config.rb
```diff
@@ -22,7 +22,7 @@ def self.initialize_puppet(puppet_config)
     Puppet[:trace] = true
 
     # Crank Puppet's log level all the way up and just control it via logback.
-    Puppet::Util::Log.level = :debug
+    Puppet[:log_level] = "debug"
```
joshcooper: I think this worked before because `log_level` wasn't a setting -- so you could set the global `level` here, and puppet would never overwrite it.

I think it stopped working, because as soon as we initialize settings below, it reads the default `log_level` setting of `notice` and resets the level from `debug` to `notice`. I *think* you could move the line you had previously after settings initialization, but then I think you'd lose debug messages that occur *while* settings are initialized.

So :+1:

KevinCorcoran: @joshcooper great, thanks for reviewing!

PR #398, opened 2015-02-12
(SERVER-360) Fix common_package_name_in_different_providers test

jeffmccune:
Without this patch the ruby puppet
`common_package_name_in_different_providers.rb` acceptance test is failing
because the createrepo command is not available on the system under test.  This
is causing the puppet-server acceptance tests to fail.

This patch resolves the problem by satisfying the dependency in the pre-suite.
This course of action is being taken over satisfying the dependency directly in
the common_package_name_in_different_providers.rb test because the state of the
puppet code-base is volatile with the AIO work.

This change will move the puppet-server tests closer to being green while also
leaving a clear indication of where the dependency installation should be moved
once the puppet code base is less volatile.

### DIFF
--- a/acceptance/suites/pre_suite/foss/50_install_puppet_acceptance_deps.rb
+++ b/acceptance/suites/pre_suite/foss/50_install_puppet_acceptance_deps.rb
@@ -1,0 +1,24 @@
+# NOTE: Ideally, the acceptance tests itself will manage its own dependencies,
+# however, the puppet agent codebase is in flux with AIO changes right now so we
+# are satisfying specific dependencies in the pre-suite of the puppet-server
+# codebase.  If you add to this file, please comment regarding what the
+# dependency is for (the specific puppet acceptance tests) so that we can easily
+# move these to the ideal location in the future.
+
+tests = options[:tests]
+
+step "Install System Dependencies for Puppet Acceptance Tests."
+hosts.each do |host|
+  variant = host['platform'].to_array.first
+  case variant
+    when /^(debian|ubuntu)$/
+      # Nothing to do
+    when /^(redhat|el|centos)$/
+      if tests.any? { |s| s =~ /common_package_name_in_different_providers/ }
+        step "Install deps for common_package_name_in_different_providers (REMOVE ONCE https://github.com/puppetlabs/puppet/pull/3601 is merged and present in ruby/puppet/)"
+        install_package host, 'createrepo'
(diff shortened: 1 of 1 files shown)


rlinehan:
This is test is also failing when the puppet-agent tests are running (I believe), so I'm not sure this is the right place to fix it? Seems like it should perhaps be fixed in the test itself, so that all the things running that test get fixed, rather than just the puppet-server tests.

Also, there's a PUP ticket for this: PUP-3960.

jeffmccune:
Thanks @rlinehan 

I'm definitely planning to fix this in puppet as well, though I did so in the acceptance test itself since I was ignorant of PUP-3960.

The plan of action is for me to fix it in both places and not block getting our tests green on the merge of https://github.com/puppetlabs/puppet/pull/3601

KevinCorcoran:
@jeffmccune - Should we close this PR?  My understanding of the situation (and I very well may be wrong) is that the plan is to resolve this issue elsewhere.

jeffmccune:
Yes, this has already been solved in ezbake by expressing a package level dependency on createrepo.

jeffmccune:
@KevinCorcoran The plan is to solve it in the puppet acceptance tests, but until we've verified those acceptance tests have been fixed there is value in this PR since it gets our board closer to green.

cprice404:
@jeffmccune this is obsolete now, right?

cprice404:
@jeffmccune can I close this?

jeffmccune:
No.  Acceptance tests are still failing on this.  There seems to have been some previous confusion, but I did edit my comment from a week ago to say this hasn't yet been merged on the puppet side.

jeffmccune:
It'd be great if someone reviewed and merged this.

cprice404:
But isn't this just a temporary hack?  And the real fix needs to go into Puppet?  So why do we want to merge it and then just revert a few days later?

cprice404:
I might just be misunderstanding the current game plan.

jeffmccune:
The test needs to pass somehow.  I'll just fix it in whatever branch I'm working on for the rest of the acceptance test failures.

cprice404:
@jeffmccune but if the correct fix is in puppet, and we upgrade to the version of puppet that contains the fix... won't that make the tests pass?  I'm not trying to be obstinate, just don't want to create more stuff that we have to unwind later than necessary...

PR #402, opened 2015-02-19
(MAINT) Cherry pick commit to migrate to ssl-utils

dankreek:
Cherry-picked Joe's commit to update from the old certificate-authority lib to the new ssl-utils lib.

Conflicts:
	src/clj/puppetlabs/puppetserver/ringutils.clj

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -15,7 +15,7 @@
   :dependencies [[org.clojure/clojure "1.5.1"]
                  [puppetlabs/trapperkeeper ~tk-version]
                  [puppetlabs/kitchensink ~ks-version]
-                 [puppetlabs/ssl-utils "0.7.0"]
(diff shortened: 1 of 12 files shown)


nwolfe:
Do we want the newer release, 0.8.0?

Also this will introduce another version of this library into the dependency tree since we have other dependencies that are pulling in older versions.  Do we want to update those too? Or are we okay with having multiple copies pulled in?

dankreek:
Oh yeah, I'll add a bump to 0.8.0, but that is a very good question. This is the "stable" branch so moving too many things around at once may not be a good idea. The new code is in a different namespace than the old code, so that definitely helps us differentiate.

fpringvaldsen:
:+1:

PR #404, opened 2015-02-19
[FOR REVIEW] (SERVER-369) Get EL-7 CI passing

jeffmccune:
This patch is basically a search and replace over the project files, updating
puppet-server to match the new AIO puppet-agent vardir.

Also changes: /opt/puppetlabs/{agent,puppet}

Without this patch the puppet-server codebase is referring to an out of date
path for the location of the puppet-agent AIO package.  This patch addresses
the problem by updating the location to the current value of
/opt/puppetlabs/puppet


.gitignore
```diff
@@ -16,3 +16,10 @@ profiles.clj
 dev/user.clj
 .lein-repl-history
 ext/packaging
+Gemfile.lock
+Gemfile.local
+.DS_Store
+.idea/
+/junit/
+/tmp/
+puppet_server.iml
```
camlow325: I think we've generally tried to keep IDE-specific lines out of the .gitignore - stuff like .idea and puppet_server.iml.  Instead, we've been assuming developers would put the appropriate IDE-specific references in their global .gitignores.

Also, are /junit/ and /tmp/ needed?  I can't remember what .DS_Store is either...

nwolfe: .DS_Store is an OSX thing - another thing to put in your global ~/.gitignore.

Gemfile.lock
```diff
@@ -1,203 +0,0 @@
-GEM
```
camlow325: I can't remember why we had this in the project to begin with.  Seems like we ended up having to remove it after cloning in CI anyway.  Maybe it was to facilitate the JRuby spec tests in the project but I seem to be able to run those successfully without this in place.  No objections, just curious.

nwolfe: @camlow325 We decided to break Puppet Labs tradition and commit this file alongside Gemfile per the recommended Bundler practice (see http://bundler.io/).

acceptance/config/beaker/jenkins/el-7-x86_64-mdca.cfg
```diff
@@ -8,6 +8,14 @@ HOSTS:
     platform: el-7-x86_64
     template: Delivery/Quality Assurance/Templates/vCloud/centos-7-x86_64
     hypervisor: vcloud
+    puppetcodedir: /etc/puppetlabs/code
+    distmoduledir: /etc/puppetlabs/code/modules
+    hieraconf:     /etc/puppetlabs/code/hiera.yaml
+    puppetconfdir: /etc/puppetlabs/puppet
+    puppetserver-confdir: /etc/puppetserver/conf.d/
```
camlow325: Is this right?  https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md appears to show `/etc/puppetlabs/puppetserver/conf.d`.

documentation/configuration.markdown
```diff
@@ -75,9 +79,9 @@ This file contains the settings for Puppet Server itself.
 # configuration for the JRuby interpreters
 
 jruby-puppet: {
-    gem-home: /var/lib/puppet/jruby-gems
+    gem-home: /opt/puppetlabs/puppet/cache/jruby-gems
```
camlow325: Not sure if this one is final, but I couldn't find this listed at https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md.  Would be good to sync this up with that doc....

resources/ext/config/conf.d/puppetserver.conf
```diff
@@ -2,13 +2,13 @@
 jruby-puppet: {
     # This setting determines where JRuby will look for gems.  It is also
     # used by the `puppetserver gem` command line tool.
-    gem-home: /var/lib/puppet/jruby-gems
+    gem-home: /opt/puppetlabs/puppet/cache/jruby-gems
 
     # (optional) path to puppet conf dir; if not specified, will use the puppet default
     master-conf-dir: /etc/puppetlabs/agent
```
camlow325: Is this right?  Should this be `/etc/puppetlabs/puppet` like https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md shows for puppet-agent confdir?  Not sure if we decided to have a different confdir for the server than what the agent would use?

acceptance/config/beaker/local/el6/1host.cfg
```diff
@@ -14,7 +14,7 @@ HOSTS:
     group: puppet
     puppetbin: /usr/bin/puppet
     puppetbindir: /usr/bin
-    puppetvardir: /var/lib/puppet
+    puppetvardir: /opt/puppetlabs/puppet/cache
     hieradatadir: /etc/puppet/hieradata
```
camlow325: Should these hiera and distmoduledir settings be under `/etc/puppetlabs/code` instead of `/etc/puppet`?

acceptance/config/beaker/local/el6/1host.cfg
```diff
@@ -14,7 +14,7 @@ HOSTS:
     group: puppet
     puppetbin: /usr/bin/puppet
```
camlow325: Should the puppetbin* settings point under `/opt/puppetlabs/puppet/bin`?

camlow325: Same question for these on the corresponding vcenter cfg files.

acceptance/config/beaker/local/el6/1host.cfg
camlow325: Same question for these on the corresponding vcenter cfg files.

acceptance/config/beaker/local/el6/1host.cfg
```diff
@@ -14,7 +14,7 @@ HOSTS:
...
     puppetbindir: /usr/bin
-    puppetvardir: /var/lib/puppet
+    puppetvardir: /opt/puppetlabs/puppet/cache
     hieradatadir: /etc/puppet/hieradata
     hieraconf: /etc/puppet/hiera.yaml
     distmoduledir: /etc/puppet/modules
```
camlow325: Should the sitemoduledir be under `/opt/puppetlabs/puppet/modules`?

camlow325: Same question on the corresponding vcenter cfg files.

acceptance/config/beaker/local/el6/1host.cfg
```diff
@@ -14,7 +14,7 @@ HOSTS:
     group: puppet
```
camlow325: Should puppetpath point to `/etc/puppetlabs/puppet`?

camlow325: Same question on the corresponding vcenter cfg files.

acceptance/lib/helper.rb
```diff
@@ -160,7 +160,7 @@ def install_puppet_server (host, make_env={})
   # setting should be able to be moved into a regular conf file.  See
   # SERVER-331.
   def get_rubylibdir host, config_key
-    "/opt/puppetlabs/agent/lib/ruby/vendor_ruby"
+    "/opt/puppetlabs/puppet/lib/ruby/vendor_ruby"
   end
 
   def configure_puppet_server
```
camlow325: Are the `/etc/puppetserver` references in here still accurate or should they be `/etc/puppetlabs/puppetserver` instead?

camlow325:
I saw references to `/etc/puppetserver/conf.d` in .../suites/pre_suite/optional/setup_puppetserver.rb.  Should those be changed to `/etc/puppetlabs/puppetserver/conf.d` now?

camlow325:
Do we still care to have the vbox cfg files updated with the new paths?  Not sure if we're still actively maintaining those...

camlow325:
Done reviewing for now.  Mostly questions on different paths, many of which I could be totally wrong about.

acceptance/config/beaker/local/el6/1host.cfg
rlinehan: Can we just remove this from these files? It seems like if everything is set correctly, we shouldn't need to have `puppetpath` etc. set in the hostconfig, because they should be set through beaker's other option-setting capabilities.

.gitignore
jeffmccune: Got it.  I think I had a global git ignore that I lost when I swapped laptops recently.  I'll remove the non-project specific ignores.

The /tmp/ and /junit/ directories are project-specific in that they're created and fill up after running beaker repetitively.

Gemfile.lock
jeffmccune: @nwolfe @camlow325 The Puppet Labs tradition deliberately goes against the Bundler recommendation because their recommendation is based on the assumption that the "application" is a web app being developed and deployed in a distributed fashion.  Their goal is to keep everyone running the exact same version.

We have a need to run different versions in different situations, particularly when we're trying to track changes in active development in multiple dependencies, e.g. puppet, facter, beaker, etc...

Hence the deliberate breaking away from the recommendation.

camlow325: @jeffmccune Thanks for the explanation.  I'm good keeping with PL tradition.

.gitignore
camlow325: Ok, I'm good with keeping /tmp/ and /junit/ then.

acceptance/config/beaker/jenkins/el-7-x86_64-mdca.cfg
jeffmccune: Dunno yet, I'll have to run it through CI and post back if it's working or not.  Since the goal is to get things green, I'm going to be following Josh Cooper's lead and approach, creating a `XXX_hacks.rb` file that has things we know to be "wrong" but are "right" in that they get CI to a known good state as quickly as possible.

FWIW, I pulled this value from https://github.com/puppetlabs/puppet/commit/70feeef3810850c99199f9e702048848c0f02bf3#diff-972ce103aac1e41c0a3dca0de9bc6a7bR13 via https://tickets.puppetlabs.com/browse/PUP-3632

acceptance/config/beaker/local/el6/1host.cfg
jeffmccune: @rlinehan Ideally, yes we could remove these, but beaker hasn't been updated yet and this is how they're handling the configuration values in `puppet`.

jeffmccune:
Reading the comments, it's become pretty clear this PR was pre-mature in that it's technically focused on the exact `vardir` changes specified in the previous ticket, but still doesn't get us that much closer to green because lots of other stuff is obviously wrong.

I'm going to close this pull-request for the time being and keep working on getting things green.  Once we have a green run for EL-7 I'll either re-open this pull request or open a new one, depending on how valuable it will be to preserve the discussion here.

documentation/configuration.markdown
jeffmccune: The AIO puppet-agent doesn't have a GEM_HOME like puppet-server does, so this value comes from the previous behavior of puppet-server placing GEM_HOME into `$vardir/gem-home`.

resources/ext/config/conf.d/puppetserver.conf
jeffmccune: Thanks, this one was an oversight and should be /etc/puppetlabs/puppet

documentation/configuration.markdown
camlow325: Ok.  The only point I wanted to make here is that I think it would be good to have the canonical "file_paths.md" also updated to whatever we set for this in puppet-server since that covers puppet-server paths, too.  Maybe that's already being taken care of independently?

PR #406, opened 2015-02-20
(MAINT) Revert the bump

dankreek:


nwolfe:
@dankreek We're going to actually need to rever the commit, not just downgrade the version, right?

PR #409, opened 2015-02-20
(SERVER-303) Remove version-check service

fpringvaldsen:
Remove the dujour version-check service and replace it with the
dujour-version-check library.

I've flagged this as for review as I needed to add a new function, `get-version-string`, to the `dujour-version-check` library for this to work. That change is present in https://github.com/puppetlabs/dujour-version-check/pull/7. Once that is merged, we can do a new release of the `dujour-version-check` library and update this from a snapshot dependency.


src/clj/puppetlabs/services/master/master_service.clj
```diff
@@ -4,25 +4,31 @@
...
+         certname          (get-in config [:puppet-server :certname])
+         localcacert       (get-in config [:puppet-server :localcacert])
+         hostcrl           (get-in config [:puppet-server :hostcrl])
+         settings          (ca/config->master-settings config)
+         jruby-service     (tk-services/get-service this :JRubyPuppetService)
+         upgrade-error     (core/construct-404-error-message jruby-service)
+         product-name      (get-in-config [:product :name])
+         update-server-url (get-in-config [:product :update-server-url])]
+
+     (version-check/check-for-updates! {:product-name product-name} update-server-url)
```
camlow325: Do we have a test which validates that the master service makes this call with the right parameters?  If not, it would seem good to create one.  Maybe could stub out the call to `check-for-updates!` with a `with-redefs` so that the full plumbing doesn't have to be exercised for this local unit test?

camlow325:
:+1: with a question on an additional unit test, assuming the dujour-version-check dependency is moved off of a SNAPSHOT, and Travis tests subsequently pass.

src/clj/puppetlabs/services/master/master_service.clj
fpringvaldsen: We do not, all we had before were tests of the various functions in the `version-check-core` namespace, which is now covered by the tests in the `dujour-version-check` library.

I'll work on adding something to ensure that the `check-for-updates!` function is successfully being called with the right parameters.

fpringvaldsen: @camlow325 I've added a test as requested.

test/unit/puppetlabs/services/master/master_service_test.clj
```diff
@@ -70,3 +69,56 @@
...
+(defn dummy-retrieve-cert-function
+  [_]
+  (println "Do nothing"))
+
+(deftest version-check-test
+  (testing "master calls into the dujour version check library using the correct values"
+    (with-redefs
+
+      [version-check/check-for-updates! version-check-test-fn
+       ; Redefine the CA functions to ensure that the master does not create
```
fpringvaldsen: I did this since it stops the master service from creating SSL files, which sped up the test a bit. I also figured it would reduce the amount of things the master is doing on initialization since all we care about in this test is whether or not it calls into the `check-for-updates!` function. Further, this eliminates the need to manage a test directory in the test.

I'm happy to remove this though, and manage a test directory in the above test.

test/unit/puppetlabs/services/master/master_service_test.clj
```diff
@@ -70,3 +69,50 @@
...
+; test if the master for some reason stops calling into the version check
+; library
+(def version-check-success
+  (atom false))
+
+(defn version-check-test-fn
+  [request-values update-server-url]
+  (is (= {:group-id "puppets" :artifact-id "yoda"} (:product-name request-values)))
+  (is (= "http://notarealurl/" update-server-url))
+  (swap! version-check-success not))
```
camlow325: This is purely a style comment, so feel free to ignore.  I'd be inclined to have this just set the atom with the contents of the arguments passed into the function - maybe as a map with a key/val pair for each argument - and move these `is` assertions into the body of the deftest.  Where possible, I like to have assertions inside the body of the `deftest` so that what it is testing is more prominent.

camlow325: This test otherwise looks good to me.

test/unit/puppetlabs/services/master/master_service_test.clj
camlow325: I'm good with mocking these here since this functionality isn't related to the test.

fpringvaldsen:
@camlow325 I've moved the assertions into the test itself, and am now using the atom to store the values of the parameters passed to `version-check-test-fn`.

cprice404:
Philosophical ranting aside... +1

fpringvaldsen:
@cprice404 @camlow325 I've moved the atom and function into a `let` block in the version check test. I've also bumped the dependency on `dujour-version-check` so this should be ready to go now.

camlow325:
:+1: Looks like Travis failed.  Not sure if that's because it was down for maintenance?

cprice404:
kicked travis

fpringvaldsen:
@cprice404 @camlow325 I released `dujour-version-check` to clojars and moved the dependency out of the `dev` profile and into the main dependencies list. Hopefully travis will pass now.

fpringvaldsen:
Okay I think I've finally fixed the Travis failures, we'll see how this new commit does.

fpringvaldsen:
@cprice404 @camlow325 It looks like Travis is finally fixed. I've made one new commit, https://github.com/puppetlabs/puppet-server/commit/fe08e87856a748c2a4d7738eb7f236cd5a91a0b6, to fix an issue with the `version-check-test`.

cprice404:
@jeffmccune I presume that you wouldn't be aversely affected by us going ahead and merging this?

src/clj/puppetlabs/services/master/master_service.clj
```diff
@@ -4,25 +4,31 @@
...
+            [puppetlabs.dujour.version-check :as version-check]))
 
 (defservice master-service
   [[:WebroutingService add-ring-handler get-route]
    [:PuppetServerConfigService get-config]
    [:RequestHandlerService handle-request]
    [:CaService initialize-master-ssl! retrieve-ca-cert! retrieve-ca-crl!]
-   [:JRubyPuppetService]]
+   [:JRubyPuppetService]
+   [:ConfigService get-in-config]]
```
nwolfe: This already depends on the `PuppetServerConfigService`, can we just get the config from that?  The entire `config` is pulled in here: https://github.com/puppetlabs/puppet-server/pull/409/files#diff-f69c98ef06ea1897ed3b68e38fb638d3R21

(Honest question - I don't remember if you actually can use the `PuppetServerConfigService` for this)

KevinCorcoran: Yes, it should use the `PuppetServerConfigService` and not the `ConfigService` directly.

test/unit/puppetlabs/services/master/master_service_test.clj
```diff
@@ -70,3 +69,52 @@
...
+                                                                        :update-server-url update-server-url)))
+          test-dir (doto "target/master-service-test" fs/mkdir)]
+      (try
+        (with-redefs
+          [version-check/check-for-updates! version-check-test-fn
+           ; Redefine the CA functions to ensure that the master does not create
+           ; any SSL files
+           ca/retrieve-ca-cert! (fn [_ _])
+           ca/retrieve-ca-crl! (fn [_ _])
+           ca/initialize-master-ssl! (fn [_ _ _])]
```
nwolfe: Could you use the disabled CA service instead of the actual CA service + with-redefs?

fpringvaldsen:
@nwolfe @KevinCorcoran I've addressed your feedback.

One further question: Do we want this to go into master for the 2.0 release? Or do we want this to go into 1.x/2.x, and thus target it at stable?

src/clj/puppetlabs/services/master/master_service.clj
```diff
@@ -4,25 +4,30 @@
...
-         upgrade-error (core/construct-404-error-message jruby-service)]
+   (let [path              (get-route this :master-routes)
+         config            (get-config)
+         certname          (get-in config [:puppet-server :certname])
+         localcacert       (get-in config [:puppet-server :localcacert])
+         hostcrl           (get-in config [:puppet-server :hostcrl])
+         settings          (ca/config->master-settings config)
+         jruby-service     (tk-services/get-service this :JRubyPuppetService)
+         upgrade-error     (core/construct-404-error-message jruby-service)
+         product-name      (get-in-config [:product :name])
```
nwolfe: Do we need the `get-in-config`? We're already pulling in the whole `config` and using the standard `(get-in config [:k1 :k2])` in this let block

test/unit/puppetlabs/services/master/master_service_test.clj
```diff
@@ -12,10 +11,12 @@
...
     [puppetlabs.services.jruby.jruby-testutils :as jruby-testutils]
     [puppetlabs.services.puppet-profiler.puppet-profiler-service :as profiler]
-    [puppetlabs.services.version.version-check-service :as version-check-service]
     [puppetlabs.trapperkeeper.testutils.logging :as logutils]
     [puppetlabs.services.ca.certificate-authority-service :refer [certificate-authority-service]]
-    [me.raynes.fs :as fs]))
+    [puppetlabs.services.ca.certificate-authority-disabled-service :refer [certificate-authority-disabled-service]]
+    [puppetlabs.dujour.version-check :as version-check]
+    [me.raynes.fs :as fs]
+    [puppetlabs.puppetserver.certificate-authority :as ca]))
```
nwolfe: The `ca` namespace isn't used anymore now that we've got the disabled CA service, right?

fpringvaldsen: It's used in the other test, which tests that the CA service correctly creates the required files.

nwolfe: Didn't you add it but are not using it anymore? The other tests use the `certificate-authority-service`, not the `ca` namespace.

fpringvaldsen: Oh, you're right, I misunderstood, sorry about that. I'll remove it

fpringvaldsen:
@nwolfe I've made the changes based on your feedback.

nwolfe:
:+1: nice!

PR #410, opened 2015-02-23
(SERVER-386) Depend on puppet-agent AIO dev version

jeffmccune:
Without this patch the puppet-agent dependency is out of date.  This is a
problem because we need to track puppet-agent as closely as possible to
minimize the likelihood of issues stemming from divergence from the underlying
Puppet behavior.

This updated version should have no behavior changes relative to the behavior
of the puppet-agent version in 33f5cf9.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -26,13 +26,13 @@ module PuppetServerExtensions
                          nil, "Puppet Version", "PUPPET_VERSION", nil) ||
                          get_puppet_version
 
-    # SERVER-339 - puppet-agent commit 49f7b1b967e44020587ab087f87e2f5e3ff2b1f6
-    # corresponds to packaged development version located at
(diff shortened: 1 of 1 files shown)


rlinehan:
Rather than continuing to add a new commit (+ PR) every time we need to bump this version, would it be possible to set the `PUPPET_BUILD_VERSION` environment variable in the jenkins job, at least as a temporary measure, until we get a stable version?

camlow325:
:+1: pending your thoughts on @rlinehan's suggestion.

cprice404:
+1 on @rlinehan 's suggestion, but I'm also happy to just merge this for now in order to keep momentum going.  @jeffmccune ?

jeffmccune:
This is no longer relevant since pull/411 included the version bump.

I totally agree with @rlinehan this should be a job parameter, not committed into the code.

PR #411, opened 2015-02-23
EL-7 AIO Changes

jeffmccune:
This set of changes results in EL-7 smoke tests passing again, which is a step
in the direction of getting CI green again.

CURRENT PLAN: All changes to the paths will be included in this pull request.  These changes are mostly captured in the tickets blocked by SERVER-369, but additional changes may be necessary that are not captured in those blocked tickets.


acceptance/lib/helper.rb
```diff
@@ -26,13 +26,13 @@ def self.initialize_config(options)
...
-    # corresponds to packaged development version located at
-    # http://builds.delivery.puppetlabs.net/puppet-agent/49f7b1b967e44020587ab087f87e2f5e3ff2b1f6/
+    # SERVER-339, SERVER-386 - puppet-agent version corresponds to packaged
+    # development version located at http://builds.delivery.puppetlabs
+    # .net/puppet-agent/
     # TODO: This build version needs to be updated to a released version
     puppet_build_version = get_option_value(options[:puppet_build_version],
                          nil, "Puppet Development Build Version",
-                         "PUPPET_BUILD_VERSION", "49f7b1b967e44020587ab087f87e2f5e3ff2b1f6")
+                         "PUPPET_BUILD_VERSION", "bc00b6e7106846aae0af5b915976a8d192e53803")
```
dankreek: Not a direct comment on this PR, but this is kind of gross. Maybe down the road we can get this value from the .gitmodules file?

resources/ext/config/conf.d/puppetserver.conf
```diff
@@ -1,14 +1,20 @@
+# Where the puppet-agent dependency places puppet, facter, etc...
+# Puppet server expects to load Puppet from this location
+os-settings: {
```
camlow325: I think this takes us part of the way toward what SERVER-331 was calling for, but not all of the way.  I think we eventually want to just fold the `ruby-load-path` setting back into the `jruby-puppet` section and do away with the `os-settings` section entirely.  @cprice404 - does that sound right?

I'm okay with doing this folding in a follow-up PR for SERVER-331 for the sake of getting this PR through.

resources/ext/ezbake.conf
```diff
@@ -2,71 +2,21 @@ ezbake: {
...
-                          "echo \\\"rundir = /var/run/puppetlabs\\\"           >> /etc/puppetlabs/agent/puppet.conf",
-                          "chown puppet:puppet /etc/puppetlabs/agent/",
-                          "chmod 775 /etc/puppetlabs/agent",
-                          "install --directory /etc/puppetlabs/agent/ssl",
-                          "chown -R puppet:puppet /etc/puppetlabs/agent/ssl",
-                          "chmod -R 770 /etc/puppetlabs/agent/ssl",
-                          "chown -R puppet:puppet /var/log/puppetlabs/",
-                          "chmod -R 770 /var/log/puppetlabs"]
+               postinst: [
+                 "/opt/puppetlabs/puppet/bin/puppet config set --section master vardir  /opt/puppetlabs/server/data/puppetserver",
```
camlow325: Do we need to do these puppet config set lines or will the `master-conf-dir` and `master-var-dir` values from the `jruby-puppet` section effectively enforce this anyway?  I was thinking that those `master-*` settings override whatever may have been in the master's corresponding "puppet.conf".

resources/ext/config/conf.d/puppetserver.conf
```diff
@@ -1,14 +1,20 @@
...
-    gem-home: /var/lib/puppet/jruby-gems
+    gem-home: /opt/puppetlabs/server/data/puppetserver/jruby-gems
 
     # (optional) path to puppet conf dir; if not specified, will use the puppet default
-    master-conf-dir: /etc/puppetlabs/agent
+    master-conf-dir: /etc/puppetlabs/puppet
 
     # (optional) path to puppet var dir; if not specified, will use the puppet default
-    master-var-dir: /opt/puppetlabs/agent/cache
+    master-var-dir: /opt/puppetlabs/server/data/puppetserver
```
camlow325: Yay, separate vardirs!

resources/ext/ezbake.conf
```diff
@@ -2,71 +2,21 @@ ezbake: {
...
+                 "/opt/puppetlabs/puppet/bin/puppet config set --section master codedir /etc/puppetlabs/code",
+                 "usermod --home /opt/puppetlabs/server/data/puppetserver puppet",
+                 "install --directory --owner=puppet --group=puppet --mode=775 /opt/puppetlabs/server/data"
```
camlow325: Should this go all the way down to `/opt/puppetlabs/server/data/puppetserver`?

camlow325: Also, should there be a step here to create the subordinate "jruby-gems" directory?  Wasn't sure if anything later on would automatically create this directory if it didn't exist, and I believe the ezbake config has been specifying that this directory be created for quite a while.

resources/ext/ezbake.conf
```diff
@@ -2,71 +2,21 @@ ezbake: {
...
-                          "echo \\\"logdir = /var/log/puppetlabs\\\"           >> /etc/puppetlabs/agent/puppet.conf",
-                          "echo \\\"rundir = /var/run/puppetlabs\\\"           >> /etc/puppetlabs/agent/puppet.conf",
-                          "chown puppet:puppet /etc/puppetlabs/agent/",
-                          "chmod 775 /etc/puppetlabs/agent",
-                          "install --directory /etc/puppetlabs/agent/ssl",
-                          "chown -R puppet:puppet /etc/puppetlabs/agent/ssl",
-                          "chmod -R 770 /etc/puppetlabs/agent/ssl",
-                          "chown -R puppet:puppet /var/log/puppetlabs/",
-                          "chmod -R 770 /var/log/puppetlabs"]
+               postinst: [
```
dankreek: I've got a fair idea, but I'm not entirely clear on how these 4 lines replace all the stuff that used to be there.

.gitignore
```diff
@@ -16,3 +18,12 @@ profiles.clj
 dev/user.clj
 .lein-repl-history
 ext/packaging
+
+# Bundler local state files
+Gemfile.lock
+Gemfile.local
+/.bundle/
```
dankreek: Are the leading slashes really necessary here? I'm guessing they aren't causing any harm, but they deviate from the previous rev of the file.

resources/ext/config/conf.d/puppetserver.conf
cprice404: Yes, SERVER-331 involves getting rid of this section entirely, and yes it's fine to deal with that in a follow-up PR (I think that makes more sense).

acceptance/lib/helper.rb
jeffmccune: This isn't actually the puppet commit, it's the puppet-agent commit, which, awesomely, is a totally separate project from puppet.

resources/ext/ezbake.conf
```diff
@@ -2,71 +2,21 @@ ezbake: {
...
-                          "echo \\\"vardir = /opt/puppetlabs/agent/cache\\\"   >> /etc/puppetlabs/agent/puppet.conf",
-                          "echo \\\"logdir = /var/log/puppetlabs\\\"           >> /etc/puppetlabs/agent/puppet.conf",
-                          "echo \\\"rundir = /var/run/puppetlabs\\\"           >> /etc/puppetlabs/agent/puppet.conf",
-                          "chown puppet:puppet /etc/puppetlabs/agent/",
-                          "chmod 775 /etc/puppetlabs/agent",
-                          "install --directory /etc/puppetlabs/agent/ssl",
-                          "chown -R puppet:puppet /etc/puppetlabs/agent/ssl",
-                          "chmod -R 770 /etc/puppetlabs/agent/ssl",
-                          "chown -R puppet:puppet /var/log/puppetlabs/",
-                          "chmod -R 770 /var/log/puppetlabs"]
```
camlow325: Are the ssl and log directories no longer a concern?  I think the ssldir is going to be at `/etc/puppetlabs/puppet/ssl`, but is it possible that the agent packaging would have pre-created as owned by root and not accessible by puppet?  Same question for the log directories.  I'm hoping that's no longer the case but just wasn't sure.

acceptance/lib/helper.rb
jeffmccune: And it's not kind of gross, it's repulsively disgusting.

acceptance/config/beaker/jenkins/el-7-x86_64-mdca.cfg
```diff
@@ -8,6 +8,14 @@ HOSTS:
     platform: el-7-x86_64
     template: Delivery/Quality Assurance/Templates/vCloud/centos-7-x86_64
     hypervisor: vcloud
+    puppetcodedir: /etc/puppetlabs/code
+    distmoduledir: /etc/puppetlabs/code/modules
+    hieraconf:     /etc/puppetlabs/code/hiera.yaml
+    puppetconfdir: /etc/puppetlabs/puppet
+    puppetserver-confdir: /etc/puppetserver/conf.d/
```
camlow325: Is this right?  Should it be `/etc/puppetlabs/puppetserver/conf.d`?  Maybe the leading `/etc/puppetlabs` part of that isn't known by the current ezbake master branch so this change would be premature?

resources/ext/ezbake.conf
```diff
@@ -2,71 +2,21 @@ ezbake: {
...
+               postinst: [
```
camlow325: Same questions here as for the "redhat" section above.

resources/ext/ezbake.conf
jeffmccune: These `puppet config set` lines are necessary because they're modifying Puppet's configuration file, not puppet-server's configuration.  The way I've been thinking about it is that Puppet's configuration must agree with and match Puppet Server's configuration, so the `puppet config set` is what takes care of this on the puppet-agent side.

FWIW, this is necessary because the puppet-agent pacakge makes assumptions that it is an agent, not a master, so we need to re-configure the puppet-agent package for use in the context of a master service.  See [SERVER-352 Consider removing puppet-server's dependency on puppet-agent](https://tickets.puppetlabs.com/browse/SERVER-352)

resources/ext/config/conf.d/puppetserver.conf
jeffmccune: Huzzah!

resources/ext/ezbake.conf
jeffmccune: > Should this go all the way down to /opt/puppetlabs/server/data/puppetserver?

Yes, but I plan to take care of that in a subsequent ticket since it wasn't technically required to get smoke tests passing.

> Also, should there be a step here to create the subordinate "jruby-gems" directory?

Maybe?  Our smoke tests are passing without it because the `gem` subcommand should automatically create the parent directory upon an `install` action provided the parent directory is writable, which it is.

resources/ext/ezbake.conf
camlow325: Ok, I don't have a big problem with it.  Was just curious if it was needed.  Since the settings from the puppet-server configuration should override the ones from puppet.conf at runtime in the context of a puppetserver master, I wasn't sure if this change would have any effect.  I guess it would if you were to try to run the puppet master under MRI Ruby directly, so I could see where this would be useful.

resources/ext/ezbake.conf
jeffmccune: They don't replace, they're the bare-minimum behavior required to get the EL-7 smoke tests passing.  I expect more lines to be added back in as we get the rest of the acceptance tests passing again.

.gitignore
jeffmccune: They're not necessary, but they're more specific in that the directory should really only be excluded from the root of the project, not sub-directories.

resources/ext/ezbake.conf
camlow325: Ok, that seems reasonable.  @cprice404 - you good with that?

resources/ext/ezbake.conf
haus: The agent package is creating this today, but that I'll have a PR in a few moments to remove it from the agent packaging.

acceptance/config/beaker/jenkins/el-7-x86_64-mdca.cfg
joshcooper: This is likely copy-pasta from puppet when we were testing against the nightly puppetserver that was using the old location. I imagine this could be removed.

resources/ext/ezbake.conf
cprice404: yup

dankreek:
:+1: This all looks good to me, barring some of @camlow325's comments

nwolfe:
Needs a rebase now that stable has been merged into master.

project.clj
```diff
@@ -95,7 +95,7 @@
              :ezbake {:dependencies ^:replace [[puppetlabs/puppet-server ~ps-version]
                                                [puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty-version]
                                                [org.clojure/tools.nrepl "0.2.3"]]
-                      :plugins [[puppetlabs/lein-ezbake "0.2.3"]]
+                      :plugins [[puppetlabs/lein-ezbake "0.3.0-SNAPSHOT"]]
```
camlow325: Will running from this SNAPSHOT work - at least temporarily - in CI?  On the CI node that runs ezbake will lein-ezbake have already been built from latest master and installed into a local .m2/repository so that the dependency can be satisfied?

jeffmccune: No it won't work until we also publish 0.3.0-Snapshot into nexus. It's more here in this PR for early review and less to merge right this moment. 



—
Jeff

resources/ext/config/conf.d/puppetserver.conf
joshcooper: Will `puppet master --configprint vardir` display the value above or fall back to the agent's vardir?

cprice404: The ruby cli tools have no knowledge of the puppetserver configs at all.  I was thinking that in P5 we could remove the master subcommand altogether and introduce additional puppetserver subcommands that would provide the necessary functionality.

cprice404: (Though I'm not certain that's a sufficient answer in terms of splitting the vardirs today... I thought that we'd all decided not to try to tackle the split vardir and the moving of the user creation until we got everything green...)

camlow325: > Will `puppet master --configprint vardir` display the value above or fall back to the agent's vardir?

I think as this PR currently stands, you'd still get "/opt/puppetlabs/server/data/puppetserver" after the initial package install because of the following line which would be added to the postinstall in "resources/ext/ezbake.conf":

```
/opt/puppetlabs/puppet/bin/puppet config set --section master vardir  /opt/puppetlabs/server/data/puppetserver
```

Although I'm starting to think maybe that's something we want to avoid doing anyway.  Like if the user had changed the puppet.conf to something else and were doing an upgrade installation, would this automatically stomp whatever the user had set?  And to @cprice404's point, the effective value that the master uses for vardir under puppetserver is sourced from the puppetserver.conf anyway, so the corresponding value in "puppet.conf" seems like it would be irrelevant under puppetserver.

I'd be :+1: on not putting more effort into ensuring that the vardir split is sound until we have a better understanding of where we stand overall around the Puppet 4 AIO + Puppet Server 2 changes in CI - unless there's something @jeffmccune or @joshcooper might know of there that would block our ability to initially get everything to green...

joshcooper: It is actually critical that `puppet master --configprint vardir` return the "correct" value, because beaker uses that to start/stop puppetserver during acceptance tests: https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/helpers.rb#L563

So I think it's ok to split the vardirs in this PR, but the postinstall step that executes `/opt/puppetlabs/puppet/bin/puppet config set --section master vardir  /opt/puppetlabs/server/data/puppetserver` must remain also.

In the future, if you decide to remove the postinstall step, please don't add another beaker configuration option, e.g. `puppetserver-vardir`, because that tightly couples beaker and puppet-server. I think it would be preferable if there was a way beaker could interrogate the host to determine this, along the lines of `puppetserver config print vardir`. That could also be used to retrieve `confdir` and then we can deprecate `puppetserver-confdir` from beaker and node definition files.

resources/ext/ezbake.conf
joshcooper: See note https://github.com/puppetlabs/puppet-server/pull/411/files#r25358485 above.

resources/ext/config/conf.d/puppetserver.conf
cprice404: I generally agree with your comments, @joshcooper, but that postinst step will only guarantee that `puppet master --configprint vardir` will work *if* the files aren't modified later.  So it should be fine for beaker/acceptance tests unless there are tests that manipulate the vardir setting, just something for us to be aware of.

And yes, I'm thinking that in P5 we should remove the 'master' command altogether and add tools for interrogating the necessary settings from Puppet Server directly.  I don't think we have any intention of adding any new beaker settings for that, totally agree with your statement.

joshcooper: Makes sense. Generally, if an acceptance test needs to change something in var/confdir, it should create a temp directory and use `with_puppet_running_on` pointing to the temp location. That said I recently found an acceptance test that deleted the puppet user and recreated it!? So yeah, hopefully we don't have any acceptance test modifying puppet.conf to point to an alternate vardir in an ad-hoc manner (not going through beaker's methods for that)..

camlow325:
:+1: These changes were validated by a successful smoke test suite run from the puppet-server aio branch.  See https://jenkins.puppetlabs.com/job/experimental_puppet-server_integration-system_smoke-aio/3/.  Acceptance tests failures still exist downstream but those will continue to be triaged while we get these changes into master to hopefully enable nightly promotion and allow for more parallel activity on the remaining aio work to be done.

PR #412, opened 2015-02-23
(SERVER-356) Fix foreground output

nwolfe:
This commit re-adds the logappender ref to the logback.xml file, which
was somehow lost during a rebase or merge. With this change, the
foreground subcommand should now produce output.

### DIFF
--- a/ezbake/config/logback.xml
+++ b/ezbake/config/logback.xml
@@ -18,6 +18,8 @@
 
     <root level="info">
         <!--<appender-ref ref="STDOUT"/>-->
+        <!-- ${logappender} logs to console when running the foreground command -->
+        <appender-ref ref="${logappender}"/>
         <appender-ref ref="F1"/>
     </root>
 </configuration>


nwolfe:
This is the fix, but I'm toying around with writing an acceptance test for this behavior.

ezbake/config/logback.xml
```diff
@@ -18,6 +18,7 @@
 
     <root level="info">
         <!--<appender-ref ref="STDOUT"/>-->
+        <appender-ref ref="${logappender:-DUMMY}"/>
```
jeffmccune: Wow.  Is that BASH style syntax for logappender taking on the value DUMMY if it's not already set?

nwolfe: Ha yep. I guess this sort of pattern was copied from PuppetDB's subcommands. I actually think that we might be able to move this default value into the foreground command itself rather than requiring every logback.xml file to specify it this way.

I'm going to try this out locally, and will update with my findings.

nwolfe: FYI the original work was done here: https://github.com/puppetlabs/ezbake/pull/148

nwolfe: Hmm....I'm not sure that we need to provide a default value here. I seem to be able to run everything as expected without the `:-DUMMY` part.

fpringvaldsen: Since this works without the `-DUMMY`, and logback doesn't complain about it, I think we should remove it. I think the code will be cleaner, more readable, and less confusing.

I also think it would be worth adding a comment explaining that this is used to log to the console when running the `foreground` command.

nwolfe:
Updated with the removal of the `:-DUMMY` value, which doesn't appear to be necessary. I've verified both `foreground` and `foreground --debug` work as expected with this change, and I don't see Logback complaining about anything either (such as a nil appender).

cprice404:
@nwolfe any reason to not go ahead and merge this?  We can add acceptance tests in a follow-up PR?

Also I think we should time-box the current attempt to write acceptance tests and if it takes longer than, say, 1 day, we should probably ticket it and come back around to it?

nwolfe:
@cprice404 No reason not to merge.  I haven't spent any time on an acceptance test yet, but after our discussion last night about possible implementations I'm not sure I consider the return on investment to be worth it right now.

I can file a ticket for maybe writing one though.

cprice404:
+1 on merging, +1 on ticketing an acceptance test.

fpringvaldsen:
:+1:

nwolfe:
:+1:

nwolfe:
Ticket for writing an acceptance test: https://tickets.puppetlabs.com/browse/SERVER-390

PR #413, opened 2015-02-23
(SERVER-319) Apache-style request logging

nwolfe:
This commit adds a new Apache-style puppetserver-access.log file to
capture HTTP traffic.

### DIFF
--- a/dev/request-logging-dev.xml
+++ b/dev/request-logging-dev.xml
@@ -1,0 +1,9 @@
+<configuration scan="true" debug="false">
+  <appender name="FILE" class="ch.qos.logback.core.FileAppender">
+    <file>./target/logs/puppetserver-access.log</file>
+    <encoder>
+        <pattern>%h %l %u %user %date "%r" %s %b %h %a %localPort %D</pattern>
+    </encoder>
+  </appender>
+  <appender-ref ref="FILE" />
+</configuration>
--- a/dev/puppet-server.conf.sample
+++ b/dev/puppet-server.conf.sample
@@ -17,6 +17,7 @@ os-settings: {
 }
 
 webserver: {
+    access-log-config: ./dev/request-logging-dev.xml
     client-auth: want
     # ssl-host controls what networks the server will accept connections from.
     # The default value below is 'localhost', so will only accept connections from
--- a/ezbake/config/conf.d/webserver.conf
+++ b/ezbake/config/conf.d/webserver.conf
@@ -1,4 +1,5 @@
 webserver: {
+    access-log-config = /etc/puppetserver/request-logging.xml
     client-auth = want
     ssl-host = 0.0.0.0
     ssl-port = 8140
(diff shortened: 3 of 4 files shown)


fpringvaldsen:
I've just tested on CentOS 7, and it looks like the configuration is pointing to the access logging file, but that `request-logging.xml` is not appearing in the `/etc/puppetserver/` directory.

nwolfe:
@fpringvaldsen It's in /var/log/puppetserver with the other log files

fpringvaldsen:
@nwolfe I'm talking about the `request-logging.xml` configuration file itself. The `webserver.conf` file is setting the access log configuration to `/etc/puppetserver/request-logging.xml`, but that configuration file does not exist, so access logging is not happening.

nwolfe:
@fpringvaldsen I've double-checked this work on a couple VMs and things are working as expected. I think you might need to make sure the .jar is deleted from your ~/.m2 and do a fresh `lein install` before building the package locally.

cprice404:
+1, one question though: would it be worth plumbing this stuff into the example files for devs as well?  Seems easy-ish and probably valuable?

nwolfe:
@cprice404 Yeah good idea, doing that now...

fpringvaldsen:
@nwolfe Confirmed that it works after doing a `lein install`, sorry about that! Thanks for your help. I have now functionally reviewed and confirmed that this is working so I'm :+1: barring putting this into the example files for devs.

dev/request-logging-dev.xml
```diff
@@ -0,0 +1,8 @@
+<configuration scan="true" debug="false">
+  <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+    <encoder>
+        <pattern>%h %l %u %user %date "%r" %s %b %h %a %localPort %D</pattern>
+    </encoder>
+  </appender>
+  <appender-ref ref="STDOUT" />
+</configuration>
```
nwolfe: @cprice404 Let me know if this is what you were thinking

nwolfe:
Updated with dev/source support

dev/request-logging-dev.xml
cprice404: Pretty much, except I hadn't been thinking we would dump that stuff to STDOUT.  I think I was thinking maybe we'd use a file appender, but hadn't thought much about where the file would go.  Maybe just in like './target/logs/puppetserver-access.log' or something?

nwolfe: Yeah, having the output mixed in with the other STDOUT looked pretty bad.  Will move it to ./target

dev/request-logging-dev.xml
```diff
@@ -0,0 +1,9 @@
+<configuration scan="true" debug="false">
+  <appender name="FILE" class="ch.qos.logback.core.FileAppender">
+    <file>./target/logs/puppetserver-access.log</file>
+    <encoder>
+        <pattern>%h %l %u %user %date "%r" %s %b %h %a %localPort %D</pattern>
+    </encoder>
+  </appender>
+  <appender-ref ref="FILE" />
+</configuration>
```
nwolfe: Updated to write to a file instead of STDOUT.

PR #414, opened 2015-02-23
(MAINT) Merge SERVER-363 changes from stable to master

cprice404:
This commit takes the changes that Nate made in the stable
branch for SERVER-363 (to add lein ezbake support to the
stable branch) and merges them up to the master branch.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,7 @@ gem 'rake', :group => [:development, :test]
 gem 'jira-ruby', :group => :development
 
 group :test do
+  gem 'rake'
   gem 'rspec'
   gem 'beaker', '~> 2.2'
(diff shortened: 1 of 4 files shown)


nwolfe:
Building packages and testing locally...

nwolfe:
I verified that I can build packages successfully, but wasn't able to get very far after that due to the AIO paths and all sorts of brokenness that we're in right now.

cprice404:
OK then it sounds like it's probably best to let @jeffmccune decide when this should land?

camlow325:
Yeah, it seems like most of this is going to be undone by @jeffmccune's next PR coming to master -- Gemfile.lock and os-settings references being deleted.  +1 on letting @jeffmccune decide when this should come in.

PR #415, opened 2015-02-25
(SERVER-389) Increase JRuby pool default-borrow-timeout

nwolfe:
This commit simply increases the timeout to 1200000 ms (20 minutes).


camlow325:
:+1: Guess we should wait and see what Travis does?

PR #416, opened 2015-02-25
(FOR REVIEW) (SERVER-325) flush jrubies after max requests

cprice404:
NOTE: not ready for merge; need to clean up some of the last few commits and add a few more tests.

---------------------

This PR does the following:

* Change our JRuby queuing strategy from FIFO to LIFO.  (This improves memory usage by a great deal when the system is not under high load, and also makes it possible for us to flush individual JRuby instances after a maximum number of requests in a sane way; in a FIFO model, all JRuby instances would hit the max number of requests at roughly the same time.)
* Refactors a lot of the JRuby code to make it possible to flush instances with more granularity than the entire pool
* Introduce a setting 'max-requests-per-instance'; after any JRuby instance handles more requests than this, it will be flushed and replaced in the pool by a fresh one

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
@@ -1,0 +1,176 @@
+(ns puppetlabs.services.jruby.jruby-puppet-internal
+  (:require [schema.core :as schema]
+            [puppetlabs.services.jruby.jruby-puppet-schemas :as jruby-schemas]
+            [puppetlabs.services.jruby.puppet-environments :as puppet-env]
+            [me.raynes.fs :as fs])
+  (:import (com.puppetlabs.puppetserver PuppetProfiler JRubyPuppet)
+           (puppetlabs.services.jruby.jruby_puppet_schemas JRubyPuppetInstance PoisonPill)
+           (java.util HashMap)
+           (org.jruby CompatVersion RubyInstanceConfig$CompileMode)
+           (org.jruby.embed ScriptingContainer LocalContextScope)
+           (java.util.concurrent LinkedBlockingDeque TimeUnit)))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Definitions
+
+(def jruby-puppet-env
+  "The environment variables that should be passed to the Puppet JRuby interpreters.
+  We don't want them to read any ruby environment variables, like $GEM_HOME or
+  $RUBY_LIB or anything like that, so pass it an empty environment map - except -
+  Puppet needs HOME and PATH for facter resolution, so leave those."
+  (select-keys (System/getenv) ["HOME" "PATH"]))
+
+(def ruby-code-dir
+  "The name of the directory containing the ruby code in this project.
+  This directory lives under src/ruby/"
+  "puppet-server-lib")
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Private
+
+(defn instantiate-free-pool
+  "Instantiate a new queue object to use as the pool of free JRubyPuppet's."
+  [size]
+  {:post [(instance? jruby-schemas/pool-queue-type %)]}
+  (LinkedBlockingDeque. size))
+
+(schema/defn ^:always-validate
+             create-pool-from-config :- jruby-schemas/PoolState
+  "Create a new PoolData based on the config input."
+  [{size :max-active-instances} :- jruby-schemas/JRubyPuppetConfig]
+  {:pool (instantiate-free-pool size)
+   :size size})
+
+(defn prep-scripting-container
+  [scripting-container ruby-load-path gem-home]
+  (doto scripting-container
+    (.setLoadPaths (cons ruby-code-dir
+                         (map fs/absolute-path ruby-load-path)))
+    (.setCompatVersion (CompatVersion/RUBY1_9))
+    (.setCompileMode RubyInstanceConfig$CompileMode/OFF)
+    (.setEnvironment (merge {"GEM_HOME" gem-home} jruby-puppet-env))))
+
+(defn empty-scripting-container
+  "Creates a clean instance of `org.jruby.embed.ScriptingContainer` with no code loaded."
+  [ruby-load-path gem-home]
+  {:pre [(sequential? ruby-load-path)
+         (every? string? ruby-load-path)
+         (string? gem-home)]
+   :post [(instance? ScriptingContainer %)]}
+  (-> (ScriptingContainer. LocalContextScope/SINGLETHREAD)
+      (prep-scripting-container ruby-load-path gem-home)))
+
+(defn create-scripting-container
+  "Creates an instance of `org.jruby.embed.ScriptingContainer` and loads up the
+  puppet and facter code inside it."
+  [ruby-load-path gem-home]
+  {:pre [(sequential? ruby-load-path)
+         (every? string? ruby-load-path)
+         (string? gem-home)]
+   :post [(instance? ScriptingContainer %)]}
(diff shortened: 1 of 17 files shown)


camlow325:
Needs rebase?

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -0,0 +1,241 @@
...
+(schema/defn borrowed-jruby-instance :- jruby-schemas/JRubyPuppetBorrowResult
+  [borrow-fn :- (schema/pred ifn?)
+   pool :- jruby-schemas/pool-queue-type
+   pool-context :- jruby-schemas/PoolContext
+   flush-instance-fn :- (schema/pred ifn?)
+   instance :- JRubyPuppetInstance]
+  (let [new-state (swap! (:state instance)
+                         update-in [:request-count] inc)
+        max-requests (get-in pool-context [:config :max-requests-per-instance])]
+    (if (and (pos? max-requests)
```
camlow325: I'm not sure if this is a big deal, but did you consider the option of having the check for max-requests done as part of the return to pool instead of in the borrow?  That way, you wouldn't end up having a request that encounters the max-requests lead a second borrow - or maybe a third, fourth, ... in the more rare case where several instances need to be flushed around the same time.

camlow325: Also, would that help with the full `flush-pool!` implementation?  If the `borrow-from-pool!*` didn't need to possibly flush an instance that was just borrowed, it would seem less likely we would end up possibly wasting one or more cases of creating and immediately flushing a new instance from `flush-pool!`.

camlow325: Although if the flush of an instance might ever need to be done on a max-stale time instead of just a max-requests basis, I suppose doing the check for flush during the borrow would make sense.  Not sure if that would ever be the case?

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -0,0 +1,142 @@
...
+                   state :- JRubyInstanceStateContainer
+                   jruby-puppet :- JRubyPuppet
+                   scripting-container :- ScriptingContainer
+                   environment-registry :- (schema/both
+                                             EnvironmentRegistry
+                                             (schema/pred
+                                               #(satisfies? puppet-env/EnvironmentStateContainer %)))]
+                  Object
+                  (toString [this] (format "%s@%s {:id %s :state (Atom: %s)}"
+                                           "puppetlabs.services.jruby.jruby_puppet_core.JRubyPuppetInstance"
```
camlow325: Do you want to use the namespace in which this is defined here rather than "jruby_puppet_core"?

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -0,0 +1,142 @@
...
+                   id :- schema/Int
+                   state :- JRubyInstanceStateContainer
+                   jruby-puppet :- JRubyPuppet
+                   scripting-container :- ScriptingContainer
+                   environment-registry :- (schema/both
+                                             EnvironmentRegistry
+                                             (schema/pred
+                                               #(satisfies? puppet-env/EnvironmentStateContainer %)))]
+                  Object
+                  (toString [this] (format "%s@%s {:id %s :state (Atom: %s)}"
```
camlow325: This looks like it will be really helpful.  :+1:

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
camlow325: Hmmm, but in the scenario where a `borrow-from-pool!*` called from `flush-pool!` would trigger a flush of the instance and the second `borrow` were to be attempted, would `flush-pool!` end up trying to do more borrows than the total instances allocated for the old pool - in which case we could block forever on one of the borrows?

test/unit/puppetlabs/services/jruby/jruby_pool_test.clj
```diff
@@ -15,76 +16,132 @@
...
       (is (thrown-with-msg? IllegalStateException
           err-msg
-          (borrow-from-pool pool)))
+          (jruby-core/borrow-from-pool pool-context)))
       (is (thrown-with-msg? IllegalStateException
           err-msg
-          (borrow-from-pool-with-timeout pool 120))))))
+          (jruby-core/borrow-from-pool-with-timeout pool-context 120))))))
 
-(deftest test-default-pool-size
```
camlow325: Is this test not relevant anymore?

test/unit/puppetlabs/services/jruby/jruby_pool_test.clj
```diff
@@ -15,76 +16,132 @@
...
 
+(deftest flush-jruby-after-max-requests
+  (testing "JRuby instance is not flushed if it has not exceeded max requests"
+    (let [pool-context  (create-pool-context 2)
+          instance      (jruby-core/borrow-from-pool pool-context)
+          id            (:id instance)]
+      (jruby-core/return-to-pool instance)
+      (let [instance (jruby-core/borrow-from-pool pool-context)]
+        (is (= id (:id instance))))))
+  (testing "JRuby instance is flushed after exceeding max requests"
```
camlow325: One other variant of this that I think would be good to have is to see that a JRuby instance which is deeper down in the pool can be flushed when it hits its max requests.  For example, do one borrow and, before returning that instance to the pool, do another borrow which is at its max-requests and confirm that the instance from that second borrow is appropriately flushed.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -48,22 +48,22 @@
...
-                          "begin; InstanceID; false; rescue NameError; true; end")))))))))
+        (await (:pool-agent pool-context))
+        (is (every? true?
+                    (jruby-testutils/reduce-over-jrubies!
+                      pool-context
+                      4
+                      (constantly
+                        "begin; InstanceID; false; rescue NameError; true; end"))))))))
+
+;; TODO: test flush instance while pool flush is in progress
```
camlow325: I agree this would be a good one to have, and that it feels like an "integration" kind of test.

KevinCorcoran:
This seems like a good thing.  At a high level, the implementation details (LIFO vs. FIFO, tracking request counts, etc.) all sounds sane me.

cprice404:
@KevinCorcoran thx; this'll get put on hold for a bit as it's targeted at 1.1 and not 1.0.4, but as soon as I rebase it and address @camlow325 's comments I was planning on pinging you for a more in-depth review.

KevinCorcoran:
sure, sounds good

cprice404:
Closing this until after our 1.0.x release; also need to rebase and try Jeremy's suggestion of switching to flush-on-return instead of flush-on-borrow.  Will re-open when that's ready.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
cprice404: I have the change over to "return on flush" implemented in my local branch, this will be included in my next PR.

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
cprice404: Fixed in my next PR.

test/unit/puppetlabs/services/jruby/jruby_pool_test.clj
cprice404: This test is resuscitated in my next PR.

test/unit/puppetlabs/services/jruby/jruby_pool_test.clj
cprice404: will look into this

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
cprice404: and this

PR #417, opened 2015-02-26
(maint) make 'lein test' include integration tests

KevinCorcoran:
The unit tests can still be run independently via `lein test :unit`.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -101,10 +101,8 @@
                        :dependencies [[puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty-version]]}
              :ci {:plugins [[lein-pprint "1.1.1"]]}}
 
-  :test-selectors {:default (complement :integration)
-                   :integration :integration
-                   :unit (complement :integration)
-                   :all (constantly true)}
+  :test-selectors {:integration :integration
+                   :unit (complement :integration)}
 
   :aliases {"gem" ["trampoline" "run" "-m" "puppetlabs.puppetserver.cli.gem"]
             "ruby" ["trampoline" "run" "-m" "puppetlabs.puppetserver.cli.ruby"]


project.clj
```diff
@@ -101,7 +101,7 @@
                        :dependencies [[puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty-version]]}
              :ci {:plugins [[lein-pprint "1.1.1"]]}}
 
-  :test-selectors {:default (complement :integration)
+  :test-selectors {:default (constantly :all)
```
nwolfe: Can we just use `:all` as the function?

nwolfe:
Yay, this will be better

project.clj
KevinCorcoran: Nope.  When I make that change, this is what happens:
```
$ lein test

lein test user

Ran 0 tests containing 0 assertions.
0 failures, 0 errors.
```
Looks like the value in this map needs to be a function, not just a keyword.

nwolfe: Ah, dang (also keywords *are* functions!)

nwolfe: Ohhh...judging from the `:integration :integration` selector, I bet when we try `:default :all` it's looking for tests annotated literally as "^:all", which would explain why 0 tests were ran.

project.clj
```diff
@@ -101,7 +101,7 @@
...
                    :integration :integration
                    :unit (complement :integration)
                    :all (constantly true)}
```
nwolfe: Since we're reverting back to the default behaviors of leiningen, we can just get rid of the `:default` and `:all` selectors altogether, leaving us with just:
```
  :test-selectors {:integration :integration
                   :unit (complement :integration)}
```

I verified this behaves exactly as specifying all four of them in all of the cases: `lein test`, `lein test :all`, `lein test :unit`, and `lein test :integration`.

KevinCorcoran: well, cool!

project.clj
KevinCorcoran: L I L B

```
(instance? clojure.lang.IFn :foo)
=> true
```

KevinCorcoran:
@nwolfe - Updated this as per your suggestion (thanks, BTW!) and manually verified that `lein test`, `lein test :unit`, and `lein test :integration` work as they should.

nwolfe:
:+1:

PR #419, opened 2015-02-28
(SERVER-331) Fold os-settings back into jruby-puppet

nwolfe:
This commit moves the :ruby-load-path setting out of the os-settings
section and into the jruby-puppet section. The os-settings section is no
longer necessary, so it was removed altogether.


jeffmccune:
:+1: LGTM

camlow325:
:+1:

PR #420, opened 2015-03-04
(SERVER-410) Always cache features in puppet server

jeffmccune:
Without this patch puppet server defers to puppet on the value of the
always_cache_features setting.  This is a problem because it does not
make sense for features to be reloaded frequently on the puppet server,
this behavior is only useful for puppet agent processes that need to
update the set of installed features mid-run.

This patch addresses the problem by explicitly setting
Puppet[:always_cache_features] to true in Puppet::Server::PuppetConfig.

This approach is employed to keep all of the static jruby-puppet related
settings in one place, internal to the JRuby contexts.  This approach is
employed in order to avoid expanding the surface area between Clojure
and JRuby.

### DIFF
--- a/spec/puppet-server-lib/puppet/jvm/puppet_config_spec.rb
+++ b/spec/puppet-server-lib/puppet/jvm/puppet_config_spec.rb
@@ -20,5 +20,12 @@ describe 'Puppet::Server::PuppetConfig' do
         expect(subject).to eq('debug')
       end
     end
+
+    describe '(SERVER-410) Puppet[:always_cache_features]' do
+      subject { Puppet[:always_cache_features] }
+      it 'is true for increased performance in puppet-server' do
+        expect(subject).to eq(true)
+      end
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


cprice404:
@jeffmccune I left a couple of 'food-for-thought' comments on the jira ticket.  One other thing we should consider is whether or not this should maybe targeted at stable and merged up?

jeffmccune:
Closing to re-implement in Ruby.  See comments on the ticket.

jeffmccune:
Work in progress, please don't merge.

jeffmccune:
Ready for merge again, moved the implementation into the PuppetConfig ruby class.

cprice404:
:shipit:

jeffmccune:
going to rebase onto stable and merge-up as Chris mentioned.

jeffmccune:
@camlow325 This is the one that may or may not need to go into Stable.

camlow325:
:+1: on the changes.

Since the PR involves refactoring the log-level tests, which don't exist on stable since the log-level implementation is different on stable, it may be better to just let this come into master as-is and do a separate PR for stable that just has the logic to set always_cache_features to true in the Ruby code and with just the always_cache_features -- and not the log-level -- tests in tow.  

I don't think the separate PR for stable is as critical since always_cache_features is set to true in the master section of default puppet.conf files, but it would be nice to just force set it to true in the Ruby code in case users happened to leave it out of their puppet.conf files.

camlow325:
I'm going to go ahead and merge this.  Not sure it's worth the time to pull it back into stable.

PR #422, opened 2015-03-06
(SERVER-404) Create empty /var/run/puppet directory

dankreek:
Create an empty `/var/run/puppet` directory before starting up puppet server

### DIFF
--- a/resources/ext/ezbake.conf
+++ b/resources/ext/ezbake.conf
@@ -22,7 +22,7 @@ ezbake: {
                install: ["echo \\\"os-settings: {\\\"                         > $DESTDIR/$projconfdir/conf.d/os-settings.conf",
                          "echo \\\"    ruby-load-path: [$rubylibdir]\\\"      >> $DESTDIR/$projconfdir/conf.d/os-settings.conf",
                          "echo \\\"}\\\"                                      >> $DESTDIR/$projconfdir/conf.d/os-settings.conf" ],
-               pre-start-action: ["install --group={{user}} --owner={{user}} -d /var/run/puppet"]
+               pre-start-action: ["/usr/bin/install --group={{user}} --owner={{user}} -d /var/run/puppet"]
              }
 
       debian: { dependencies: ["puppet-common (>= 3.7.3-1puppetlabs1)"
@@ -37,7 +37,7 @@ ezbake: {
                install: ["echo \\\"os-settings: {\\\"                       > $DESTDIR/$projconfdir/conf.d/os-settings.conf"
                          "echo \\\"    ruby-load-path: [$rubylibdir]\\\"    >> $DESTDIR/$projconfdir/conf.d/os-settings.conf",
                          "echo \\\"}\\\"                                    >> $DESTDIR/$projconfdir/conf.d/os-settings.conf"],
-               pre-start-action: ["install --group={{user}} --owner={{user}} -d /var/run/puppet"]
+               pre-start-action: ["/usr/bin/install --group={{user}} --owner={{user}} -d /var/run/puppet"]
              }
    }
 }
--- a/resources/ext/ezbake.conf
+++ b/resources/ext/ezbake.conf
@@ -21,7 +21,8 @@ ezbake: {
                          "mkdir -p /etc/puppet/manifests"],
                install: ["echo \\\"os-settings: {\\\"                         > $DESTDIR/$projconfdir/conf.d/os-settings.conf",
                          "echo \\\"    ruby-load-path: [$rubylibdir]\\\"      >> $DESTDIR/$projconfdir/conf.d/os-settings.conf",
-                         "echo \\\"}\\\"                                      >> $DESTDIR/$projconfdir/conf.d/os-settings.conf" ]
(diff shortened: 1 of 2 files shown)


dankreek:
Note that packaging won't be buildable until the release of EZ-bake 0.2.4

elyscape:
This doesn't actually empty out `/var/run/puppet`.  Or was "Create and empty" a typo for "Create an empty"?

I realize this sounds sarcastic but it's meant genuinely.

dankreek:
It was a typo. Thanks.

resources/ext/ezbake.conf
```diff
@@ -21,7 +21,8 @@ ezbake: {
                          "mkdir -p /etc/puppet/manifests"],
                install: ["echo \\\"os-settings: {\\\"                         > $DESTDIR/$projconfdir/conf.d/os-settings.conf",
                          "echo \\\"    ruby-load-path: [$rubylibdir]\\\"      >> $DESTDIR/$projconfdir/conf.d/os-settings.conf",
-                         "echo \\\"}\\\"                                      >> $DESTDIR/$projconfdir/conf.d/os-settings.conf" ]
+                         "echo \\\"}\\\"                                      >> $DESTDIR/$projconfdir/conf.d/os-settings.conf" ],
+               pre-start-action: ["install --group={{user}} --owner={{user}} -d /var/run/puppet"]
```
jeffmccune: Will this action execute every time the service starts, or only on package installation?  If it doesn't execute every time the service starts it might be better to move out of packaging and into the service management behavior, e.g. the init or unit file script.

resources/ext/ezbake.conf
```diff
@@ -35,7 +36,8 @@ ezbake: {
                          "mkdir -p /etc/puppet/manifests"],
                install: ["echo \\\"os-settings: {\\\"                       > $DESTDIR/$projconfdir/conf.d/os-settings.conf"
                          "echo \\\"    ruby-load-path: [$rubylibdir]\\\"    >> $DESTDIR/$projconfdir/conf.d/os-settings.conf",
-                         "echo \\\"}\\\"                                    >> $DESTDIR/$projconfdir/conf.d/os-settings.conf"]
+                         "echo \\\"}\\\"                                    >> $DESTDIR/$projconfdir/conf.d/os-settings.conf"],
+               pre-start-action: ["install --group={{user}} --owner={{user}} -d /var/run/puppet"]
```
jeffmccune: Same question as above regarding when this will execute.

resources/ext/ezbake.conf
camlow325: I think it has to execute each time the service starts, right - so that we can ensure it happens on the service startup that happens after a reboot?

dankreek: That is correct, the `pre-start-action`s get executed as root before starting the service.

jeffmccune: Yes, it does have to.  It's just sometimes not super clear when actions are carried out in EZBake.

In this case, pre-start-action gets mapped to `EZBake::Config[:redhat][:pre_start_action]` and the actions are carried out in the init scripts and unit files themselves, not in the packages.  Looks like this is what motivated [EZ-31](https://tickets.puppetlabs.com/browse/EZ-31) in the first place.

https://github.com/puppetlabs/ezbake/blob/e3f9ab7fafe6770733519ec6f435d5ef01d2eca8/resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/init.erb#L60-L62

camlow325: With this pre-start-action, I get a systemd config for the puppetserver.service which has:

~~~~
ExecStartPre=install --group=puppet --owner=puppet -d /var/run/puppet
~~~~

I tested out an ezbake package on CentOS 7.  If I delete the `/var/run/puppet` directory before starting the service, the directory doesn't end up getting created and so puppetserver blows up.  If I modify the puppetserver.service to use this instead ...

~~~~
ExecStartPre=/usr/bin/install --group=puppet --owner=puppet -d /var/run/puppet
~~~~

... then the directory is created fine the next time the service starts and the service comes to ready properly.

This leads me to believe that we need to use an absolute path here -- `/usr/bin/install` instead of just "install".  "install" lives under `/usr/bin` by default in the distros I tried -- CentOS 6, CentOS 7, and Debian 7.

@haus - does this path make sense to you?

haus: @camlow325 yes, Exec* in systemd need to be fully qualified. /usr/bin/install *should* be safe in the linuxes we support with ezbake.

dankreek: Ah shoot, I actually observed that behavior when the '-' was in front of the command, but removed it and somehow assumed the full path wasn't then necessary.

dankreek:
@camlow325 added the full path to the install command

camlow325:
:+1: I'm good with this.  @haus, look okay to you?

haus:
@dankreek will you merge up or will there be a new PR for master?

dankreek:
Well, the master branch may actually require a different directory to be created, or maybe none at all? This work has been captured in SERVER-414

haus:
@dankreek sweet

camlow325:
:+1:

PR #424, opened 2015-03-07
Merge up from latest on stable

cprice404:


ezbake/config/conf.d/webserver.conf
```diff
@@ -1,5 +1,6 @@
 webserver: {
+    access-log-config = /etc/puppetserver/request-logging.xml
```
camlow325: @jeffmccune With the pathing changes, should this target `/etc/puppetlabs/puppetserver/request-logging.xml`?

resources/ext/config/request-logging.xml
```diff
@@ -0,0 +1,9 @@
+<configuration debug="false">
+  <appender name="FILE" class="ch.qos.logback.core.FileAppender">
+    <file>/var/log/puppetserver/puppetserver-access.log</file>
```
camlow325: @jeffmccune `/var/log/puppetlabs/puppetserver/puppetserver.log`?

resources/ext/ezbake.conf
```diff
@@ -1,7 +1,22 @@
...
    foss: {
       redhat: { dependencies: ["puppet-agent"],
+               # This is terrible, but we need write access to puppet's
+               # var/conf dirs, so we need to add ourselves to the group.
+               # Then we need to chmod some dirs until the Puppet packaging
+               # is changed to allow group write; Haus said that this
+               # has probably the way to do it for now.  There might be a better
+               # way to get rid of the hard-coded paths here, but I don't
+               # know it.
+               preinst: ["install --group={{user}} --owner={{user}} -d /var/lib/puppet/jruby-gems",
```
camlow325: @jeffmccune Seems like these don't make sense to include here anymore?  I think we decided that the gems directory would be created by `puppetserver gem` later on anyway when needed so we don't need to do this in packaging.  The manifests directory location here wouldn't be right either and I thought the creation of the production environment was going to move over to the puppet-agent package now?

resources/ext/ezbake.conf
```diff
@@ -14,6 +29,10 @@ ezbake: {
              }
 
       debian: { dependencies: ["puppet-agent"],
+               # see redhat comments on why this is terrible
```
camlow325: Same comments here as for the redhat foss section - not sure we want any of this.  The rundir is different for Puppet v4 and I think this will get reworked for SERVER-414 anyway.

resources/ext/config/request-logging.xml
nwolfe: Jeremy's suggestion is correct

(https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md#puppetserver)

ezbake/config/conf.d/webserver.conf
nwolfe: Jeremy's suggestion is correct

(https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md#puppetserver)

jeffmccune: Yes!

resources/ext/config/request-logging.xml
jeffmccune: Yes!

resources/ext/ezbake.conf
jeffmccune: Yes, these should be removed as they were packaging hacks that have been removed as part of the AIO work.

resources/ext/ezbake.conf
jeffmccune: We don't want any of it, these were hacks that should be removed and then fixed properly, e.g. in SERVER-358 and other related tickets.

README_BRANCHING.md
```diff
@@ -0,0 +1,41 @@
+This document attempts to capture the details of our branching strategy
+for Puppet Server.  This information is up-to-date as of 2015-03-06.
+
+## THE MOST IMPORTANT THING TO KNOW
+
+tl;dr: we are putting a freeze on merging up changes from `stable` to `master`
+until Puppet Server 2.0 ships.
```
jeffmccune: One thing I've been wondering and I think is good to get down in writing.

Why are we explicitly preventing merge-ups instead of resolving the conflicts we want to avoid?  That is to say, if we don't want to pull up a new dependency, why not do the merge-up as normal, undo the dependency change in the master branch, then in the future update the dependency (or not) as it makes sense for master?

This would allow us to continue to merge-up frequently, as normal, without pulling up the undesirable changes we don't want to pull up.

cprice404: That sounds like a lot more work to me; we're not just talking about dependencies, we're talking about major code changes in the Puppet Server code base.

If we can tag 2.0 soon-ish, then this freeze should be short-lived and hopefully not a huge issue.  Failing that, I think the best thing to do would be to create a 2.1.x branch and start merging up to that from both 1.1 and 2.0.

cprice404:
Added a commit that should address PR feedback.

cprice404:
Also merged up the latest verbiage from README_BRANCHING

README_BRANCHING.md
jeffmccune: > That sounds like a lot more work to me; we're not just talking about dependencies, we're talking about major code changes in the Puppet Server code base.

Cool, just was something I was curious about as the plan has been developing.  I wasn't sure that we were talking about major changes in behavior beyond the dependency changes.

camlow325:
:+1:

PR #426, opened 2015-03-09
(SERVER-383) Add more trusted fact OIDs

dankreek:


camlow325:
The new OIDs all look right to me.  Would be good to have some test coverage for the short name to OID conversion.  Maybe we could add all of the new ones to the "additional extensions are created for a master" test case in "certificate_authority_test.clj" since it already covers most of the other Puppet extensions anyway.

camlow325:
:+1: other than adding a bit of test coverage for the new OIDs.

dankreek:
@camlow325 Added a test for the trusted facts mapping

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -868,7 +868,7 @@
                             :value    {:dns-name ["subject"
                                                   "onefish"
                                                   "twofish"]}}
-                           ;; These extensions come form the csr_attributes.yaml file
+                           ;; These extensions come form csr_attributes.yaml
```
camlow325: 'Spose we could s/form/from while we're in here...

camlow325:
:+1: assuming Travis passes.  @nwolfe - you good with this?

nwolfe:
We also have a spec test for these OIDs here: https://github.com/puppetlabs/puppet-server/blob/master/spec/puppet-server-lib/puppet/jvm/certificate_spec.rb#L27

Not sure whether to add to it or not, I'll defer to your judgement.

:+1: otherwise

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
dankreek: D'OH

camlow325:
Yeah, I'd forgotten about the spec tests.  Good find, @nwolfe.

Looks like that test is exercising the code in https://github.com/puppetlabs/puppet-server/blob/cf94a2a3ad1d409c1c2605a63a081977c75b6d9d/src/ruby/puppet-server-lib/puppet/server/certificate.rb#L40-L87.

I don't think we have any Puppet Server code in this path which refers to these explicit OIDs.  I think is all done through the PUPPET_OIDs array behind it, which lives in core Ruby Puppet code:

https://github.com/puppetlabs/puppet/blob/4dae4b348ab6942d07c178fd160726e2c7895eaa/lib/puppet/ssl/oids.rb#L28-L53

I'm not sure that we need to expand a Puppet Server repo test to validate that core Ruby Puppet code is translating all OIDs correctly.  That said, this should change very infrequently and wouldn't seem to be an onerous addition to this test.

I'll also defer to @dankreek's judgement as to whether or not to add the extra coverage.

dankreek:
Oh yeah, good call, @nwolfe. Extending the test won't explicitly test the Puppet server code base, but the PR that went into the Ruby code base didn't actually have any further testing with it. I'll go ahead and extend that test.

dankreek:
D'OH! Turns out that test actually comes from a previously generated certificate and not the yaml file. I'm honestly fine with this as it is. Ya'll want to merge?

dankreek:
@camlow325 @nwolfe ok, moar tests added.

nwolfe:
:+1:

cprice404:
we just waiting for travis on this one, then?

nwolfe:
Yeah...so you know, sometime next week...

PR #428, opened 2015-03-10
(SERVER-380) upgrade to jetty 9.2

cprice404:


test/unit/puppetlabs/services/config/puppet_server_config_service_test.clj
```diff
@@ -26,6 +26,13 @@
 (def test-resources-dir
   "./dev-resources/puppetlabs/services/config/puppet_server_config_service_test")
 
+(def required-config
```
camlow325: Should we also delete the other `required-config` from above?

camlow325:
:+1: other than the removal of the extra `required-config` def.

fpringvaldsen:
I'm also :+1: barring the removal of the extra `required-config`.

cprice404:
Fixed

nwolfe:
:+1:

PR #429, opened 2015-03-10
(SERVER-273) upgrade to jruby 1.7.19

cprice404:
This upgrades us to JRuby 1.7.19, excludes JRuby's embedded
Bouncy Castle jar from the uberjar, and sets the JARS_NO_REQUIRE
flag so that JRuby will defer to our Bouncy Castle.


project.clj
```diff
@@ -18,14 +18,15 @@
                  [puppetlabs/kitchensink ~ks-version]
                  [puppetlabs/certificate-authority "0.6.0"]
                  [puppetlabs/http-client "0.4.0"]
-                 [org.jruby/jruby-core "1.7.15" :exclusions [com.github.jnr/jffi com.github.jnr/jnr-x86asm]]
+                 [org.jruby/jruby-core "1.7.19"
+                  :exclusions [com.github.jnr/jffi com.github.jnr/jnr-x86asm]]
+                 [com.github.jnr/jffi "1.2.8"]
```
camlow325: From running mvn deps, it looked to me that jruby-core 1.7.19 was still using jffi 1.2.7 and not jffi 1.2.8.  Do we need jffi 1.2.8 specifically, or would it be safer to just stick with jffi 1.2.7 since the jruby packages presumably were tested upstream with it?

cprice404: JRuby specifies a range for this, doesn't it?

cprice404: The whole reason I had to add these in the first place was because lein complains about them specifying a dependency as a range, e.g. "[1.2.1,1.3.0)" or something, which would lead me to believe that they were getting 1.2.8 in their builds, but I haven't done anything via maven.

camlow325: https://github.com/jruby/jruby/blob/1.7.19/pom.xml#L75 shows a jffi.version of "1.2.7".  With the 1.7.19 tag selected in a local checkout of jruby src, "mvn dependency:tree -Dverbose" includes the following in the output:

~~~~
...
[INFO] org.jruby:jruby-core:jar:1.7.19
...
[INFO] +- com.github.jnr:jffi:jar:1.2.7:compile
[INFO] +- com.github.jnr:jffi:jar:native:1.2.7:compile
~~~~

A "lein deps \:tree" and "lein test" both worked okay for me against your branch with a project.clj modified to use:

~~~~
[com.github.jnr/jffi "1.2.7"]
[com.github.jnr/jffi "1.2.7" :classifier "native"]
~~~~

I don't have any evidence that 1.2.8 is harmful to use with jruby 1.7.19.  I was just thinking if it looks like that's what jruby 1.7.19 is using by default that it might be better to just stick with that.

fpringvaldsen:
:+1: barring the resolution of the discussion about the `jffi` dependency.

project.clj
cprice404: I pinged the devs in #jruby to see if they have any info.  I agree with you that we should probably use 1.2.7 if that's what is being used in their tests, assuming we look at the 1.2.7->1.2.8 changelog and don't see any kind of huge bugfix in there that we think we need.

I'm still confused about the range thing though.  I would definitely expect the lein project snippet that you pasted above to work fine.  But if you comment those out and comment out the 'exclusion' of jffi from the core dependency, and then run lein deps tree, you should hopefully see the version range error that i'm seeing.  And I'd kind of like to understand how we're getting that range error if you think that the maven dep is pinned to 1.2.7.

camlow325:
:+1: once the jffi dependency is bumped back down to 1.2.7, which is apparently what the \#jruby devs recommend for JRuby 1.7.19.  I also did some testing with the JRuby flush pool endpoint and saw that the memory leak described in SERVER-273 no longer appears with a local build of the PR from this branch.

cprice404:
Added a commit to go back to 1.2.7

ruby/puppet
```diff
@@ -1 +1 @@
-Subproject commit 40e76e807f314e9435546f57013d66dd67e0d786
+Subproject commit b792d8437f925a640f74bc1642fb6a129b276e72
```
camlow325: Was this change intentional?  Seems like we still want to have the stable branch pinned to Puppet 3.7.4 for now?

cprice404: nope, not intentional, thanks for catching.

cprice404:
squashed in a revert to that accidental submodule commit

camlow325:
:+1:

PR #430, opened 2015-03-10
(SERVER-435) Pin beaker to 7d0cccf

jeffmccune:
Without this patch puppetserver is using an older version of beaker that does
not have all of the AIO path updates as of this point in time.  This patch
updates beaker to version 7d0cccf which is the exact version Puppet is using in
their master branch as per PUP-4056.


camlow325:
:+1: Matches the version in the PUP gemfile on current head of master as of mid-day 3/10/2015 - https://github.com/puppetlabs/puppet/blob/d4c7d36f9883e91ff02f714e62ccac3be178aedd/acceptance/Gemfile#L15.

PR #427, opened 2015-03-09
(SERVER-409) Plumb directories into jruby

nwolfe:
This commit adds plumbing for codedir, rundir, logdir in the same way we
were plumbing confdir and vardir through to JRuby.

This commit also bumps up the ruby Puppet submodule.

----

:+1: Unit tests pass
:+1: Can run the Server from source

:-1: Server is broken in packages, but not sure it's due to these changes, still investigating
**UPDATE**: Broken due to bindir and ssldir issues that are expected and not related to this PR

PR #431, opened 2015-03-10
(MAINT) Move Puppet submodule to head of master

nwolfe:
This commit moves the Ruby Puppet submodule to the current head of
master.

----

The intent of #427 was that after merging we would be on the head of Ruby Puppet, but it was a few commits behind. This brings us up to the latest.

:+1: Unit tests pass
:+1: Running from source works


camlow325:
:+1:

PR #432, opened 2015-03-10
(SERVER-419) Add puppetserver-confdir to beaker options

jeffmccune:
Without this patch the following error manifests itself in the acceptance
tests:

    #<RuntimeError: Error: /puppetserver.conf does not exist on bnck3n2w2tclnq8.delivery.puppetlabs.net>
    /workspace/serious/src/beaker/lib/beaker/dsl/helpers.rb:778
    /workspace/serious/src/beaker/lib/beaker/dsl/helpers.rb:584
    /workspace/serious/src/puppet-server/acceptance/lib/helper.rb:102
    /workspace/serious/src/puppet-server/acceptance/suites/pre_suite/foss/90_validate_sign_cert.rb:5
    /workspace/serious/src/beaker/lib/beaker/test_case.rb:120
    /workspace/serious/src/beaker/lib/beaker/test_case.rb:120
    /opt/crossfader/versions/ruby/2.1.4/lib/ruby/2.1.0/benchmark.rb:294
    /workspace/serious/src/beaker/lib/beaker/test_case.rb:117
    /workspace/serious/src/beaker/lib/beaker/test_suite.rb:286
    /workspace/serious/src/beaker/lib/beaker/test_suite.rb:283
    /workspace/serious/src/beaker/lib/beaker/test_suite.rb:283
    /workspace/serious/src/beaker/lib/beaker/test_suite.rb:325
    /workspace/serious/src/beaker/lib/beaker/cli.rb:149
    /workspace/serious/src/beaker/lib/beaker/cli.rb:89
    /workspace/serious/src/beaker/bin/beaker:6
    /workspace/serious/src/puppet-server/.bundle/gems/ruby/2.1.0/bin/beaker:23
    /workspace/serious/src/puppet-server/.bundle/gems/ruby/2.1.0/bin/beaker:23
    Warning: acceptance/suites/pre_suite/foss/90_validate_sign_cert.rb errored in 2.88 seconds

This is a problem because the failure prevents the tests from succeeding on
EL-7.  This patch addresses the problem by adding the correct configuration to
the Beaker host configuration by modifying the beaker options file.  This is
equivalent to passing in command line arguments but doesn't require modifying
Jenkins job configurations, only the `options.rb` file located in version
control.


acceptance/config/beaker/options.rb
```diff
@@ -8,6 +8,7 @@
   "puppetservice" => 'puppetserver',
   "use-service" => true,
   "master-start-curl-retries" => 60,
-  "puppetpath" => "/etc/puppetlabs/agent"
+  "puppetpath" => "/etc/puppetlabs/puppet",
+  "puppetserver-confdir" => '/etc/puppetlabs/puppetserver/conf.d'
```
nwolfe: Is there a plan for getting this value into Beaker itself? I see this is already the value for PE: https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host/unix.rb#L22 (the foss/aio defaults are different though)

jeffmccune: No, we specifically do _not_ want this to go into beaker itself.  The principle @joshcooper has proposed that I think we should be following is that beaker does not have any built-in knowledge about the layout of puppetserver or puppet-agent.

That is to say, beaker should inspect path values dynamically using mechanisms like `puppet config print`.

We don't have anything like `puppet config print` for puppetserver printing out hte value of our configuration directory, so we're providing that information as something like a command line argument, but not a command line argument because a true argument would require changes to the jenkins jobs.

The options.rb file is the best place we have today to make this sort of change and manage it going forward.

nwolfe: Oh right ... duh.  Guess I was thrown off by finding this value in beaker itself already.

Thanks for the explanation.

nwolfe:
:+1:

acceptance/config/beaker/options.rb
jeffmccune: The reason beaker should not have this information is because it's become clear that it's costly in terms of time and coordination overhead.  Better and more efficient to have each project provide the information to beaker, or have beaker dynamically inspect the information.  This will allow us to make these sorts of changes in the future without having to coordinate the changes to beaker across puppet, facter, hiera, puppetserver, etc...

camlow325: Looks like https://github.com/puppetlabs/puppet/blob/d8f95998fc4e3358ed94ea61d3406b7fa0174e98/acceptance/config/aio/options.rb#L5 has the `puppetserver-confdir` change, although it has it like this:

~~~~
:'puppetserver-confdir' => '/etc/puppetlabs/puppetserver/conf.d',
~~~~

Not sure the syntax -- leading colon -- makes a difference?

Is the `puppetpath` still needed?  Wasn't sure since I didn't see it in the .../aio/options.rb file, but maybe that file's not the best one to reference for this?

jeffmccune: That's an unfortunate and interesting Rubyism.  A Ruby symbol cannot have a hyphen in it and one way to get around that is to convert the whole thing into a string, then symbolize the string.

I suspect it works with or without the colon because beaker just ends up calling `.to_s` on whatever it gets, so `:"puppetserver-confdir".to_s` and `'puppetserver-confdir'.to_s` have the same behavior in the end.

Personally I find `'puppetserver-confdir'` a little less confusing, and it works, so...

jeffmccune: FWIW:

```ruby
irb(main):001:0> :foo-bar
NameError: undefined local variable or method `bar' for main:Object
        from (irb):1
        from /opt/crossfader/versions/ruby/2.1.4/bin/irb:11:in `<main>'
irb(main):002:0> :"foo-bar"
=> :"foo-bar"
irb(main):003:0> :"foo-bar".to_s
=> "foo-bar"
```

joshcooper: puppet (ruby) no longer uses `puppetpath` and that option was removed from the AIO host objects in beaker. Ideally you should be able to remove `puppetpath` as well (now that you've moved the pin to dc53fa36a1813454b9b55243a3f291dae122d614), but perhaps there are still puppetserver acceptance tests that rely on that seting?

joshcooper: About the `puppetserver-confdir` being in beaker or not, we're in a unique situation that we want beaker to be able to introspect the host for that value, but we don't have the puppetserver equivalent of `puppet master --configprint ...` to actually do that. 

However, the setting is essentially global (as in beaker doesn't support multiple masters), so it's relatively easy for a project (like puppet) to add that to `options.rb`. And since this change only affects `aio` type acceptance tests, we can make this change in beaker without breaking existing consumers, e.g. puppet#stable.

acceptance/suites/pre_suite/foss/01_aio_hacks.rb
```diff
@@ -1,12 +0,0 @@
...
-
-# See: https://github.com/puppetlabs/puppet-specifications/blob/af825090dba94c56d332b0e61d73395eb6eb0535/file_paths.md#puppet-server
-# End users are expected to execute puppetserver by having /opt/puppetlabs/bin in their PATH
-step "AIO: Add /opt/puppetlabs/bin to PATH (REMOVE ONCE https://tickets.puppetlabs.com/browse/QENG-1891 is resolved and employed)" do
-  hosts.each do |host|
-    # /opt/puppetlabs/bin is present to get puppetserver
-    # /opt/puppetlabs/puppet/bin is present to get facter
-    on host, 'sed -i s,PATH=,PATH=/opt/puppetlabs/bin:/opt/puppetlabs/puppet/bin:, ~/.ssh/environment'
-  end
-end
```
joshcooper: If this file is not empty, do you want to delete it and update the config that specifies which setup steps to run?

jeffmccune: It should have been deleted?

jeffmccune: @joshcooper Yeah, this file has been deleted entirely, and our CI jobs just specify the directory of {{acceptance/suites/pre_suite/foss}} so deleting the file has the effect of removing it from the set of steps that are run.

acceptance/config/beaker/options.rb
jeffmccune: > Ideally you should be able to remove puppetpath as well

Removed puppetpath in 088126b

camlow325:
:+1:

PR #433, opened 2015-03-10
(server-358) Update permissions on ssldir for the puppet user

shrug:
This removes the creation/permission handling for the app
data directory, which is added to the packaging in
https://github.com/puppetlabs/ezbake/pull/198, and
adds back some code to update the permissions on the
ssldir to work with the puppet user and group.

### DIFF
--- a/resources/ext/ezbake.conf
+++ b/resources/ext/ezbake.conf
@@ -21,8 +21,9 @@ ezbake: {
                  "/opt/puppetlabs/puppet/bin/puppet config set --section master rundir  /var/run/puppetlabs/puppetserver",
                  "/opt/puppetlabs/puppet/bin/puppet config set --section master pidfile /var/run/puppetlabs/puppetserver/puppetserver.pid",
                  "/opt/puppetlabs/puppet/bin/puppet config set --section master codedir /etc/puppetlabs/code",
-                 "usermod --home /opt/puppetlabs/server/data/puppetserver puppet",
-                 "install --directory --owner=puppet --group=puppet --mode=775 /opt/puppetlabs/server/data"
+                 "install --directory /etc/puppetlabs/puppet/ssl",
+                 "chown -R puppet:puppet /etc/puppetlabs/puppet/ssl",
+                 "find /etc/puppetlabs/puppet/ssl -type d -print0 | xargs -0 chmod 770"
                ]
              }
 
@@ -34,8 +35,9 @@ ezbake: {
                  "/opt/puppetlabs/puppet/bin/puppet config set --section master rundir  /var/run/puppetlabs/puppetserver",
                  "/opt/puppetlabs/puppet/bin/puppet config set --section master pidfile /var/run/puppetlabs/puppetserver/puppetserver.pid",
                  "/opt/puppetlabs/puppet/bin/puppet config set --section master codedir /etc/puppetlabs/code",
-                 "usermod --home /opt/puppetlabs/server/data/puppetserver puppet",
-                 "install --directory --owner=puppet --group=puppet --mode=775 /opt/puppetlabs/server/data"
+                 "install --directory /etc/puppetlabs/puppet/ssl",
+                 "chown -R puppet:puppet /etc/puppetlabs/puppet/ssl",
+                 "find /etc/puppetlabs/puppet/ssl -type d -print0 | xargs -0 chmod 770"
                ]
              }
    }


resources/ext/ezbake.conf
```diff
@@ -21,8 +21,9 @@ ezbake: {
                  "/opt/puppetlabs/puppet/bin/puppet config set --section master rundir  /var/run/puppetlabs/puppetserver",
                  "/opt/puppetlabs/puppet/bin/puppet config set --section master pidfile /var/run/puppetlabs/puppetserver/puppetserver.pid",
                  "/opt/puppetlabs/puppet/bin/puppet config set --section master codedir /etc/puppetlabs/code",
-                 "usermod --home /opt/puppetlabs/server/data/puppetserver puppet",
-                 "install --directory --owner=puppet --group=puppet --mode=775 /opt/puppetlabs/server/data"
+                 "install --directory /etc/puppetlabs/puppet/ssl",
+                 "chown -R puppet:puppet /etc/puppetlabs/agent/ssl",
```
haus: `agent` => `puppet`

resources/ext/ezbake.conf
```diff
@@ -34,8 +35,9 @@ ezbake: {
...
```
haus: same as above... `/etc/puppetlabs/agent/ssl` => `/etc/puppetlabs/puppet/ssl`

haus:
:+1:

cprice404:
+1 but @jeffmccune should probably sign off before we merge it.

resources/ext/ezbake.conf
```diff
@@ -21,8 +21,9 @@ ezbake: {
                  "/opt/puppetlabs/puppet/bin/puppet config set --section master rundir  /var/run/puppetlabs/puppetserver",
                  "/opt/puppetlabs/puppet/bin/puppet config set --section master pidfile /var/run/puppetlabs/puppetserver/puppetserver.pid",
                  "/opt/puppetlabs/puppet/bin/puppet config set --section master codedir /etc/puppetlabs/code",
-                 "usermod --home /opt/puppetlabs/server/data/puppetserver puppet",
-                 "install --directory --owner=puppet --group=puppet --mode=775 /opt/puppetlabs/server/data"
+                 "install --directory /etc/puppetlabs/puppet/ssl",
+                 "chown -R puppet:puppet /etc/puppetlabs/puppet/ssl",
+                 "chmod -R 770 /etc/puppetlabs/puppet/ssl",
```
jeffmccune: I'm concerned about this line.  What happens if someone has already installed `puppet-agent`, then some time later decides to install `puppetserver`?  Will the permissions of all of the existing files be changed such that we might end up with executable PEM files and such?

resources/ext/ezbake.conf
```diff
@@ -34,8 +35,9 @@ ezbake: {
...
```
jeffmccune: Same comment as above, I'm concerned about the recursive nature of this.

resources/ext/ezbake.conf
shrug: Completely correct. Updated to only update permissions on the directories.

resources/ext/ezbake.conf
```diff
@@ -21,8 +21,9 @@ ezbake: {
                  "/opt/puppetlabs/puppet/bin/puppet config set --section master rundir  /var/run/puppetlabs/puppetserver",
                  "/opt/puppetlabs/puppet/bin/puppet config set --section master pidfile /var/run/puppetlabs/puppetserver/puppetserver.pid",
                  "/opt/puppetlabs/puppet/bin/puppet config set --section master codedir /etc/puppetlabs/code",
-                 "usermod --home /opt/puppetlabs/server/data/puppetserver puppet",
-                 "install --directory --owner=puppet --group=puppet --mode=775 /opt/puppetlabs/server/data"
+                 "install --directory /etc/puppetlabs/puppet/ssl",
+                 "chown -R puppet:puppet /etc/puppetlabs/puppet/ssl",
+                 "find /etc/puppetlabs/puppet/ssl -type d -exec chmod 770 {} \;",
```
jeffmccune: I'm getting the following error trying to run `lein with-profile ezbake ezbake stage`, which exercises these changes to the configuration file:

```
Error encountered performing task 'ezbake' with profile(s): 'ezbake'
com.typesafe.config.ConfigException$Parse: Reader: 26: in value for key 'postinst': backslash followed by ';', this is not a valid escape sequence (quoted strings use JSON escaping, so use double-backslash \\ for literal backslash)
```

My suggestion is to try this, which should be a bit more robust with shell escape issues:

```bash
find  /etc/puppetlabs/puppet/ssl -type d -print0 | xargs -0 chmod 770
```

Also, I think the trailing comma is illegal since this is JSON.

resources/ext/ezbake.conf
```diff
@@ -34,8 +35,9 @@ ezbake: {
...
```
jeffmccune: Same comments as above

resources/ext/ezbake.conf
```diff
@@ -21,8 +21,9 @@ ezbake: {
                  "/opt/puppetlabs/puppet/bin/puppet config set --section master rundir  /var/run/puppetlabs/puppetserver",
                  "/opt/puppetlabs/puppet/bin/puppet config set --section master pidfile /var/run/puppetlabs/puppetserver/puppetserver.pid",
                  "/opt/puppetlabs/puppet/bin/puppet config set --section master codedir /etc/puppetlabs/code",
-                 "usermod --home /opt/puppetlabs/server/data/puppetserver puppet",
-                 "install --directory --owner=puppet --group=puppet --mode=775 /opt/puppetlabs/server/data"
+                 "install --directory /etc/puppetlabs/puppet/ssl",
+                 "chown -R puppet:puppet /etc/puppetlabs/puppet/ssl",
+                 "find /etc/puppetlabs/puppet/ssl -type d -print0 | xargs chmod 770"
```
nwolfe: This is missing the `-0` between `xargs` and `chmod`, like so:
```sh
find  /etc/puppetlabs/puppet/ssl -type d -print0 | xargs -0 chmod 770
```

But...I'm wondering if this line is necessary at all? I tested these changes out on Cent 7 and things were working after the `chown`. I haven't tried other OSes yet though...

Here's what I did:
1. Install puppet-agent dc53fa36a1813454b9b55243a3f291dae122d614
2. puppet agent apply to create the /etc/puppetlabs/puppet/ssl directory (and its subdirectories)
3. Install puppet-server package from head of master
4. Start puppetserver service and see a Permission Denied exception when writing `ca.pem` to  `/etc/puppetlabs/puppet/ssl/certs`
5. `chown -R puppet:puppet /etc/puppetlabs/puppet/ssl`
6. Restart service successfully, do agent runs, etc.

Maybe I'm not accounting for some possibility though?

resources/ext/ezbake.conf
```diff
@@ -34,8 +35,9 @@ ezbake: {
...
```
nwolfe: Same comment here regarding the missing `-0`

resources/ext/ezbake.conf
shrug: Updated. There's not enough coffee today apparently, sorry for the churn.

It's probably not strictly necessary, more of a safety net. The case where someone has created a CA previously, this just sets everything to a known state.

nwolfe:
Needs a rebase

jeffmccune:
I'll rebase it and get it merged.

PR #434, opened 2015-03-10
(docs) Fix glitchy link and list item spacing in install page

nfagerlund:
The docs site MD parser thought we were trying to italicise, for some reason.


nfagerlund:
PRing at stable because that's what the docs site generator pulls from.

PR #435, opened 2015-03-11
(SERVER-438) Update version to 2.0.0-rc1

jeffmccune:
If this commit passes through CI then the merge commit that brings this change
in should be tagged as 2.0.0-rc1


jeffmccune:
We may want to hold off on merging this until we sort out [SERVER-439](https://tickets.puppetlabs.com/browse/SERVER-439).

PR #438, opened 2015-03-11
(maint) Check service start right out of the package

jeffmccune:
Without this patch the acceptance tests in the master branch are passing with
all green, successful, positive results.  This is a problem because those
results are misleading, immediately out of the package puppetserver fails to
start due to permission errors.

The problem is masked by the use of `puppet` in beaker shortly after installing
the puppetserver package.  The use of `puppet resource`, `puppet apply` or
`puppet agent` as root has the side-effect of changing ownership and
permissions of /var/run/puppetlabs, /var/log/puppetlabs, and
/etc/puppetlabs/puppet, all of which have the follow-on effect of allowing the
puppetserver service to start because the puppet user it runs as is then able
to create the directories it needs.

This patch addresses the problem by trying to start the puppetserver service
immediately after installing the package, just a like an user would.  The test
fails, like it should.

### DIFF
--- a/acceptance/suites/pre_suite/foss/71_smoke_test_puppetserver.rb
+++ b/acceptance/suites/pre_suite/foss/71_smoke_test_puppetserver.rb
@@ -1,0 +1,23 @@
+# This step needs to execute as quickly as possible after the
+# install_puppet_server master step located in 70_install_puppet.rb
+#
+# This is because beaker uses `puppet resource` to manage the puppetserver
+# service which has the side-effect of changing ownership and permissions of key
+# paths such as /var/run/puppetlabs and /etc/puppetlabs/puppet/ssl
+#
+# This side-effect masks legitimate issues we need to test, such as "the
+# puppetserver fails to start out of the package"
+
+step "(SERVER-414) Make sure puppet-server can start without puppet resource, "\
+  "apply, or agent affecting the known good state of the SUT in a way that "\
+  "causes the tests to pass with false positive successful results."
+
+variant = master['platform'].to_array.first
+case variant
+  when /^(redhat|el|centos)$/
+    on(master, "service puppetserver start")
+    on(master, "service puppetserver status")
+    on(master, "service puppetserver stop")
+  else
+    step "(SERVER-414) Skipped for platform variant #{variant}"
+end


acceptance/suites/pre_suite/foss/71_smoke_test_puppetserver.rb
```diff
@@ -0,0 +1,23 @@
...
+# This side-effect masks legitimate issues we need to test, such as "the
+# puppetserver fails to start out of the package"
+
+step "(SERVER-414) Make sure puppet-server can start without puppet resource, "\
+  "apply, or agent affecting the known good state of the SUT in a way that "\
+  "causes the tests to pass with false positive successful results."
+
+variant = master['platform'].to_array.first
+case variant
+  when /^(redhat|el|centos)$/
```
camlow325: Do we expect this won't work for other variants?

jeffmccune: The exact approach won't work for other variants, no.  Not ever platform implements `service foo {start,stop,status}`.

Beaker got around this by using `puppet resource service puppetserver ensure=...`, but doing that is the root cause of this issue in the first place.

For each variant we need to figure out the service management framework interface, e.g. launchctl, systemctl, /etc/init.d/foo, svcadm, etc...

acceptance/suites/pre_suite/foss/71_smoke_test_puppetserver.rb
```diff
@@ -0,0 +1,23 @@
...
+    on(master, "service puppetserver start")
```
camlow325: So if any of the "on"s encounter a "bad" exit code from the command, the test will fail, right?  *Sorry, my beaker-fu is very weak.*

acceptance/suites/pre_suite/foss/71_smoke_test_puppetserver.rb
camlow325: k, that makes sense.

acceptance/suites/pre_suite/foss/71_smoke_test_puppetserver.rb
jeffmccune: Correct, I could be more explicit with that, the implicit nature of `on` trips me up too.

camlow325: k, I'm good with that.

camlow325:
:+1: Thanks much for thinking to add this!

PR #439, opened 2015-03-12
(SERVER-414) Manage rundir during service startup

jeffmccune:
Without this patch the puppetserver puppet rundir does not exist, which is a
problem because the parent directory of rundir is owned by root and
puppetserver runs as puppet which is unable to create the rundir.  This
prevents the service from starting up out of the package.

This patch addresses the problem by creating the rundir directory in a manner
owned by the same account puppetserver runs as.  This eliminates the need for
puppetserver to create the directory and allows puppetserver to create files in
the rundir directory.

NOTE: This does not cover the case of `puppetserver foreground` which will
still fail if executed before the service is started using the service
management framework.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -96,7 +96,7 @@
              :ezbake {:dependencies ^:replace [[puppetlabs/puppet-server ~ps-version]
                                                [puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty-version]
                                                [org.clojure/tools.nrepl "0.2.3"]]
-                      :plugins [[puppetlabs/lein-ezbake "0.3.0-SNAPSHOT"]]
+                      :plugins [[puppetlabs/lein-ezbake "0.3.0"]]
                       :name "puppetserver"}
              :uberjar {:aot [puppetlabs.trapperkeeper.main]
                        :dependencies [[puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty-version]]}
--- a/resources/ext/ezbake.conf
+++ b/resources/ext/ezbake.conf
@@ -23,6 +23,10 @@ ezbake: {
                  "/opt/puppetlabs/puppet/bin/puppet config set --section master codedir /etc/puppetlabs/code",
                  "usermod --home /opt/puppetlabs/server/data/puppetserver puppet",
                  "install --directory --owner=puppet --group=puppet --mode=775 /opt/puppetlabs/server/data"
+               ],
+               # Note, pre-start-actions need to have fully qualified paths
+               pre-start-action: [
+                 "/usr/bin/install --directory --owner={{user}} --group={{user}} --mode=775 /var/run/puppetlabs/puppetserver"
                ]
              }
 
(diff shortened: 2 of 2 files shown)


jeffmccune:
@camlow325 Based on the notes I have from this morning, are you thinking this pull request should also include changes to puppetserver master that move away from ezbake 0.3.0-SNAPSHOT and move to a non-snapshot version of ezbake?

If so, it seems like the 0.3.0 release of ezbake should be another ticket.  I've been trolling through jira and confluence but I can't figure out how we normally handle these ezbake releases, the closest I found was the back-port of EZ-31 to stable, but I still don't see where the release happened.

waynr:
There is no "normal" process for releasing ezbake, primarily due to the lack of smoke tests to perform basic validation steps against projects using it. The best way to check which PRs are included in any particular version of ezbake to day as far as I know is to check the commit history: 
https://github.com/puppetlabs/ezbake/commits/stable

It looks like EZ-31 is in at least version 0.2.4 of ezbake. Looks like it is referred to in the changelog:
https://github.com/puppetlabs/ezbake/commit/0ac22b57f7001278712dfa9b2b57aed14dd0ea2f

Here is the jenkins job that was used to roll out the release:
http://jenkins-enterprise.delivery.puppetlabs.net/view/ezbake/view/all/job/enterprise_ezbake_release-clj_stable/1/

This jenkins job contains the results of work from QENG-1938 where we agreed on the general approach of running explicit release tasks in stead of "lein release"

camlow325:
:+1: on the ezbake.conf changes.

@jeffmccune - yeah, I was hoping we'd be able to do an 0.3.0 release of ezbake that had the pre-start-action support and have this PR bump puppet-server's lein-ezbake dependency to 0.3.0.

I think @dankreek did the last ezbake release on the stable branch, 0.2.4, so he or @waynr would certainly have better details than I would.  Ahead of that, I think @dankreek added a CHANGELOG entry for the new release in a MAINT PR (no associated JIRA ticket) - https://github.com/puppetlabs/ezbake/pull/196/files.  And presumably the pre-release unit test job was kicked to start the release process - http://jenkins-enterprise.delivery.puppetlabs.net/view/ezbake/view/all/job/enterprise_ezbake_unit-clj_release_stable/1/?  

I'd be good with having you create a CHANGELOG for the 0.3.0 release on ezbake master, put that in a MAINT PR (JIRA ticket optional) and then do the 0.3.0 release.  @waynr - should that part just involve running http://jenkins-enterprise.delivery.puppetlabs.net/view/ezbake/view/all/job/enterprise_ezbake_unit-clj_release_master/?

After that goes though, I think it would be good to update the lein-ezbake version dependency in this PR, and then we can get it merged.

camlow325:
@jeffmccune In the process of doing the lein-ezbake release for master, I also think it would be good to merge up any relevant changes from stable.  May not be any, but just in case...

waynr:
@camlow325 the job you linked to is the correct job to trigger for the `0.3.0` release.

Just FYI I am currently working on a smoketest pipeline for ezbake that will do cool things.

jeffmccune:
@camlow325 OK, we've got ezbake 0.3.0 now (with a bonus 0.2.5 by my error) and I added a commit here to pin to 0.3.0.

camlow325:
:+1:

PR #440, opened 2015-03-12
(SERVER-345) Fixup usages of cacert / localcacert in master

dankreek:
There is a fairly length description of this work in the ticket, but each commit has a brief explanation of what went on there.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -588,7 +588,7 @@
   localcacert is where that the CA cert file should be copied to."
   ([cacert :- schema/Str
     localcacert :- schema/Str]
-   (if (fs/exists? cacert)
+   (if (and (fs/exists? cacert) (not (fs/exists? localcacert)))
      (do
        (ks/mkdirs! (fs/parent localcacert))
        (fs/copy cacert localcacert))
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -130,11 +130,11 @@
 
 (def required-master-files
   "The set of SSL files that are required on the master."
-  [:hostprivkey :hostcert])
+  #{:hostprivkey :hostcert})
 
 (def required-ca-files
   "The set of SSL related files that are required on the CA."
-  [:cacert :cacrl :cakey :cert-inventory :serial])
+  #{:cacert :cacrl :cakey :cert-inventory :serial})
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Internal
--- a/documentation/puppet_conf_setting_diffs.markdown
+++ b/documentation/puppet_conf_setting_diffs.markdown
@@ -27,16 +27,11 @@ in the `bootstrap.cfg` file. See [Service Bootstrapping](./configuration.markdow
 
 ### [`cacert`](https://docs.puppetlabs.com/references/latest/configuration.html#cacert)
 
-If you define `ssl-cert`, `ssl-key`, `ssl-ca-cert`, and/or `ssl-crl-path` in
-[webserver.conf](./configuration.markdown#webserverconf), Puppet Server uses the file at `ssl-ca-cert` as the CA cert store for authenticating clients via SSL.
-
-If at least one of the `ssl-` settings in webserver.conf
-is set but `ssl-ca-cert` is not set, Puppet Server gives an error
-and shuts down at startup. If none of the `ssl-` settings in webserver.conf is
-set, Puppet Server uses the CA file defined for the `cacert` setting in puppet.conf
-for SSL authentication.
-
-If you enable Puppet Server's certificate authority service, it uses the `cacert` setting in puppet.conf to determine the location of the CA certificate for such tasks as generating the CA certificate or using the CA to sign client certificates. This is true regardless of the configuration of the `ssl-` settings in webserver.conf.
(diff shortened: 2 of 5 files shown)


src/clj/puppetlabs/services/master/master_service.clj
```diff
@@ -19,8 +19,8 @@
          localcacert (get-in config [:puppet-server :localcacert])
          settings    (ca/config->master-settings config)]
 
-     (retrieve-ca-cert! localcacert)
      (initialize-master-ssl! settings certname)
+     (retrieve-ca-cert! localcacert)
```
dankreek: So I wan't to make super double sure that flipping the order that these two functions are executed in is an ok thing to do. Once I added localcacert to the list of "SSL files that need to be present or missing" I needed to ensure that the master initialized its SSL files before (potentially) copying the ca cert.

nwolfe: This doesn't make sense to me how this can work now...haven't had a chance to play with the code locally yet though. 

> I needed to ensure that the master initialized its SSL files before (potentially) copying the ca cert.

How come? The partial-state check happens during master initialization, so I would think that swapping these two lines would always result in a partial-state exception since the `:localcacert` won't have been written yet.

camlow325: Yeah, @dankreek, I think you're right that it would have to go in this order now if `localcacert` is added to the `required-master-files`.  In the reverse order, the `cacert` would be copied to `localcacert` by the `retrieve-ca-cert!` function before `initialize-master-ssl!` is called.  `initialize-master-ssl!` would then throw a partial state error because the `localcacert` file already exists but the `hostprivkey` and `hostcert` do not.

It hadn't dawned on me to try to add `localcacert` to `required-master-files` as part of this PR but I suppose that makes sense.  I don't see any problems with reversing the order since `initialize-master-ssl!` shouldn't need the `localcacert` to exist.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -130,11 +130,11 @@
 
 (def required-master-files
   "The set of SSL files that are required on the master."
-  [:hostprivkey :hostcert])
+  #{:hostprivkey :hostcert :localcacert})
 
 (def required-ca-files
   "The set of SSL related files that are required on the CA."
-  [:cacert :cacrl :cakey :cert-inventory :serial])
+  #{:cacert :cacrl :cakey :cert-inventory :serial})
```
camlow325: I had been expecting we would be adding `:localcacert` here instead of to `required-master-files`, so that the CA certificate artifacts don't get created if the `:localcacert` already exists.  This might be the case for someone seeking to do an external CA setup for their master who forgot to replace the real certificate-authority service with the "disabled" certficate-authority service before restarting the master.

src/clj/puppetlabs/services/master/master_service.clj
nwolfe: > initialize-master-ssl! would then throw a partial state error because the localcacert file already exists but the hostprivkey and hostcert do not.

Ahh, right, that scenario makes sense now. And yes, that would necessitate this change.

Unfortunately, this doesn't work for the inverse scenario: the hostprivkey and hostcert exist, but the localcacert does not. This is a valid and possible state, and in this case the master won't be able to start due to the partial-state error.

After discussing this IRL with @camlow325 I think it's best to not change anything about the required files for now considering this wasn't strictly necessary to begin with.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
nwolfe: The CA doesn't ever use the `:localcacert` so that would be strange on the surface to have it a required CA file.  Deeper arguments could be made too, but given my previous comment they're moot if we're not going to add `:localcacert` as a required file (to either the master or CA).

camlow325:
Roll up comment regarding the possibility of adding `localcacert` to either the `required-master-files` or `required-ca-files`...

@nwolfe's point on how this could be problematic for the `required-master-files` for the case that the master cert and private key is generated outside of the puppetserver process is totally valid.

Regarding the possibility of adding `localcacert` to `required-ca-files` -- as the JIRA ticket description had called for -- I don't have a good use case in mind for why it might be valid for puppet-server to have the "real" certificate-authority service enabled, none of the CA's own files existing, but for the `localcacert` file to exist at startup.  Bailing out of puppetserver startup with an error might still be a reasonable thing to do for this case.  That said, though, it's probably safer to not rework this for now.  The worst that could happen, I think, is that for a user ultimately seeking to have the CA service disabled on the master but not realizing how to do so, the CA service would still be running and with its standard set of files having been generated.

So I agree with the idea of pulling out `localcacert` from `required-master-files` and restoring the call order for `initialize-master-ssl!` and `retrieve-ca-cert!`.

camlow325:
:+1: assuming Travis passes (if it ever runs).

nwolfe:
:+1: 

Waiting a little more for Travis...

camlow325:
I forgot to mention that we had some verbiage in the "puppet_conf_setting_diffs.markdown" doc which referred to the `cacert` to `ssl-ca-cert` override:

https://github.com/puppetlabs/puppet-server/blob/puppet-server-1.0.3/documentation/puppet_conf_setting_diffs.markdown#cacert

Since we're moving to doing the override from `localcacert` instead of `cacert` with this PR, I think we should move this section of the documentation to a new `localcacert` section, replacing the mention of `cacert` with `localcacert`.

cprice404:
waiting for docs update?

camlow325:
Yeah, not in yet AFAIK.

dankreek:
@cprice404 @camlow325 @nwolfe docs are updated.

nwolfe:
:+1:

camlow325:
:+1:

PR #444, opened 2015-03-12
(MAINT) Move lein-voom plugin into profile

nwolfe:
This commit moves the lein-voom plugin to a new :voom profile instead of
a top-level project plugin. The plugin causes several warnings any time
lein is invoked and this is pretty noisy.


cprice404:
+1

nwolfe:
@erikPrime Just FYI in case you're running voom commands from the CLI...you'll now need to just activate the voom profile like so:
```bash
lein with-profile +voom <command>
```

camlow325:
:+1:

PR #445, opened 2015-03-13
(SERVER-448) Reduce the default value for # JRubies

cprice404:
There was a glaring flaw in the way that we had previously been
calculating a default value for the number of JRuby instances.
The formula was 'num-cpus + 2', which meant that on a really large
server with e.g. 32 cores, you'd get 34 JRuby instances.  That
many JRuby instances will require a very large heap size, but we
default to a 2G heap size, so users were hitting OutOfMemory errors.

This commit changes the formula.  We now use 'num-cpus - 1', with
a minimum value of 1 and a maximum value of 4.  We also log a
warning message if the user hasn't provided an explicit value for
this setting, and suggest that they do so.

### DIFF
--- a/dev/puppet-server.conf.sample
+++ b/dev/puppet-server.conf.sample
@@ -51,7 +51,7 @@ jruby-puppet: {
     # (optional) path to puppet var dir; if not specified, will use the puppet default
     #master-var-dir: /var/lib/puppet
 
-    # (optional) maximum number of JRuby instances to allow; defaults to <num-cpus>+2
+    # (optional) maximum number of JRuby instances to allow
     max-active-instances: 1
 }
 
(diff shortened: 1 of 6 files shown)


camlow325:
:+1: other than 1 Wolfe Principle ™ violation that I encountered in running "lein test" against this branch:

~~~~
lein test puppetlabs.services.jruby.jruby-pool-test
2015-03-12 17:21:39,354 WARN  [p.s.j.jruby-puppet-core] No configuration value found for jruby-puppet max-active-instances; using default value of 4.  Please consider setting this value explicitly in the jruby-puppet section of your Puppet Server config files.
~~~~

cprice404:
Added a commit for
Wolfe Principle

camlow325:
:+1: whew, order is restored.

nwolfe:
There are still a couple occurrences of the old "cpus+2" text laying around in samples/docstrings:
* https://github.com/cprice404/puppet-server/blob/bug/stable/SERVER-448-reduce-default-jrubies/dev/puppet-server.conf.sample#L54
* https://github.com/cprice404/puppet-server/blob/bug/stable/SERVER-448-reduce-default-jrubies/resources/ext/config/conf.d/puppetserver.conf#L13

cprice404:
Thanks @nwolfe .  Fixed those and found a couple others as well.

camlow325:
Hmm, needs rebase now?

cprice404:
:rage:

cprice404:
rebased

nwolfe:
:+1: passes locally, tired of waiting for Travis

PR #446, opened 2015-03-13
(MAINT) Update branching README to reflect plans re: 1.0.4

cprice404:


KevinCorcoran:
this is the same as #447  ?

cprice404:
one is stable and one is master, since we can't do merge-ups right now.

camlow325:
:+1: once we decide if "1.0.4" is what we intend to call the next release...

rlinehan:
Rather than hold off on merging until we know for sure whether this release is going to be "1.0.4" or not (because who knows how long that will take or how long it will remain true for) can we just say "1.0.x" and merge it?

cprice404:
Updated to 1.0.8 per discussion with @ahpook .

PR #447, opened 2015-03-13
(MAINT) Update branching README to reflect plans re: 1.0.4

cprice404:


camlow325:
:+1: once we decide if "1.0.4" is what we intend to call the next release from stable...

cprice404:
Updated to 1.0.8 per discussion with @ahpook .

PR #448, opened 2015-03-13
(MAINT) Fix dujour issues

fpringvaldsen:
Fix an issue wherein the group-id "puppetlabs" was not being used,
which caused getting the version string for puppet-server to fail.
Make dujour configuration optional. Remove dujour configuration
from the default config file.

### DIFF
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -78,7 +78,7 @@
        "Server Info:\n"
        "  Puppet Server version: " (version-check/get-version-string (:artifact-id product-name) (:group-id product-name)) "\n"
        "  Puppet version: " (:puppet-version (config/get-puppet-config jruby-service)) "\n"
-       "  Supported /puppet API versions: " puppet-API-versions
+       "  Supported /puppet API versions: " puppet-API-versions "\n"
        "  Supported /puppet-ca API versions: " puppet-ca-API-versions))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -69,14 +69,14 @@
     (route/not-found "Not Found")))
 
 (defn construct-404-error-message
-  [jruby-service]
+  [jruby-service product-name]
   (str "Error: Invalid URL - Puppet Server expects requests that conform to the "
        "/puppet and /puppet-ca APIs.\n\n"
        "Note that Puppet 3 agents aren't compatible with this version; if you're "
        "running Puppet 3, you must either upgrade your agents to match the server "
        "or point them to a server running Puppet 3.\n\n"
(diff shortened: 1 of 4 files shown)


src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -69,14 +69,14 @@
...
   (str "Error: Invalid URL - Puppet Server expects requests that conform to the "
        "/puppet and /puppet-ca APIs.\n\n"
        "Note that Puppet 3 agents aren't compatible with this version; if you're "
        "running Puppet 3, you must either upgrade your agents to match the server "
        "or point them to a server running Puppet 3.\n\n"
        "Server Info:\n"
-       "  Puppet Server version: " (version-check/get-version-string "puppet-server") "\n"
+       "  Puppet Server version: " (version-check/get-version-string (:artifact-id product-name) (:group-id product-name)) "\n"
        "  Puppet version: " (:puppet-version (config/get-puppet-config jruby-service)) "\n"
        "  Supported /puppet API versions: " puppet-API-versions
```
camlow325: Might be nice to add a "\n" at the end of this line so that it doesn't run into the next one?

src/clj/puppetlabs/services/master/master_service.clj
```diff
@@ -23,9 +23,9 @@
          hostcrl           (get-in config [:puppet-server :hostcrl])
          settings          (ca/config->master-settings config)
          jruby-service     (tk-services/get-service this :JRubyPuppetService)
-         upgrade-error     (core/construct-404-error-message jruby-service)
-         product-name      (get-in config [:product :name])
-         update-server-url (get-in config [:product :update-server-url])]
+         product-name      (or (get-in config [:product :name]) {:group-id "puppetlabs" :artifact-id "puppet-server"})
+         upgrade-error     (core/construct-404-error-message jruby-service product-name)
+         update-server-url (or (get-in config [:product :update-server-url]) "http://updates.puppetlabs.com")]
```
camlow325: Do we want/need to code the "http://updates.puppetlabs.com" default in here?  If it is passed through as nil, doesn't the dujour-version-check library use "http://updates.puppetlabs.com" by default?  Wasn't sure we wanted to have two levels of defaults - in puppet-server and the underlying library - for the case that it doesn't appear in the config file.

camlow325:
:see_no_evil: with just the questions about possibility adding a newline to the 404 error message and whether we need to duplicate the "http://updates.puppetlabs.com" default in puppet-server.

@jeffmccune - you probably want to take a look at this assuming we're going to try to get this in for the rc build.

cprice404:
+1 but I agree with @camlow325 's comment about not having the URL hard-coded in two places.

fpringvaldsen:
@camlow325 @cprice404 I've addressed feedback. Also, ran the ca-files-test without specifying the product information and ensured that dujour isn't logging a connect exception when trying to connect to the default URL, so I think this should be good to go once travis finishes.

camlow325:
:+1: when Travis finishes.

jeffmccune:
If this gets merged into master, please update https://tickets.puppetlabs.com/browse/SERVER-457 to that effect.  I know it's a MAINT commit, but I ended up creating a ticket since this is relevant to the RC.

camlow325:
Tired of waiting for Travis and tests pass locally for me, so merging.

PR #451, opened 2015-03-13
(SERVER-450) Add proper semver parsing to unit tests

jeffmccune:
Without this patch the unit tests are not properly parsing semantic version
strings according to the specification.  This is a problem for RC versions of
dependencies with versions such as "4.0.0-rc1"  This patch addresses the
problem by using a third party library written by our very own Deepak designed
to parse and manipulate semantic version data.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -49,8 +49,7 @@
                  [slingshot "0.10.3"]
                  [ring/ring-codec "1.0.0" :exclusions [commons-codec]]
                  [cheshire "5.3.1"]
-                 [trptcolin/versioneer "0.1.0"]
-                 [grimradical/clj-semver "0.2.0" :exclusions [org.clojure/clojure]]]
+                 [trptcolin/versioneer "0.1.0"]]
 
   :main puppetlabs.trapperkeeper.main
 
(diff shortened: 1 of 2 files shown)


camlow325:
:+1: with a couple of comments:

1) Since `clj-semver` would only be used for test code and not production code yet, I think it would be good to move the `clj-semver` dependency in 'project.clj' under the `dev` profile rather than having it at the top-level.

2) To avoid the try/catch wrapper in `valid-semver-number?`, it might be good to see if we could have @grimradical publish a new version of `clj-semver` to clojars from `clj-semver`'s current master branch.  Looks like this commit - https://github.com/grimradical/clj-semver/commit/9690467728b7c9266897804c3e11704b185ffa69 - if delivered in a release would alleviate the need for the try/catch wrapper in puppet-server.

I wouldn't hold up this PR for the new `clj-semver` if that would stretch beyond the weekend.  We could follow up on that part later, if needed.

jeffmccune:
I've pushed a commit to move `clj-semver` to the dev profile and emailed @grimradical about a released version containing `valid-semver-number?`.

rlinehan:
:+1: LGTM

PR #453, opened 2015-03-16
(SERVER-437) Upgrade to jvm-ssl-utils 0.8.0

nwolfe:
This commit upgrades the jvm-certificate-authority dependency
to jvm-ssl-utils. Most/all changes are just renames.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -16,7 +16,7 @@
   :dependencies [[org.clojure/clojure "1.6.0"]
                  [puppetlabs/trapperkeeper ~tk-version]
                  [puppetlabs/kitchensink ~ks-version]
-                 [puppetlabs/certificate-authority "0.6.0"]
+                 [puppetlabs/ssl-utils "0.8.0"]
                  [puppetlabs/http-client "0.4.0"]
                  [org.jruby/jruby-core "1.7.19"
                   :exclusions [com.github.jnr/jffi com.github.jnr/jnr-x86asm]]
--- a/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
+++ b/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
@@ -1,7 +1,6 @@
 (ns puppetlabs.services.ca.certificate-authority-core
   (:import  [java.io InputStream])
   (:require [puppetlabs.puppetserver.certificate-authority :as ca]
-            [puppetlabs.certificate-authority.core :as ca-utils]
             [puppetlabs.puppetserver.ringutils :as ringutils]
             [puppetlabs.puppetserver.liberator-utils :as utils]
             [slingshot.slingshot :as sling]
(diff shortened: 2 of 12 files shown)


project.clj
```diff
@@ -16,7 +16,7 @@
   :dependencies [[org.clojure/clojure "1.6.0"]
                  [puppetlabs/trapperkeeper ~tk-version]
                  [puppetlabs/kitchensink ~ks-version]
-                 [puppetlabs/certificate-authority "0.6.0"]
+                 [puppetlabs/ssl-utils "0.8.0"]
                  [puppetlabs/http-client "0.4.0"]
```
nwolfe: Note `http-client 0.4.0` is still bringing in `certificate-authority 0.6.0`, but @camlow325 is going to be upgrading this dependency to a newer version that is also on `ssl-utils 0.8.0` soon.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -1,7 +1,6 @@
 (ns puppetlabs.services.ca.certificate-authority-core
   (:import  [java.io InputStream])
   (:require [puppetlabs.puppetserver.certificate-authority :as ca]
-            [puppetlabs.certificate-authority.core :as ca-utils]
```
dankreek: ~[](http://i.imgur.com/tP8gx9u.jpg)

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -2,7 +2,7 @@
   (:import (java.io StringReader ByteArrayInputStream ByteArrayOutputStream))
   (:require [puppetlabs.puppetserver.certificate-authority :refer :all]
             [puppetlabs.trapperkeeper.testutils.logging :as logutils]
-            [puppetlabs.certificate-authority.core :as utils]
+            [puppetlabs.ssl-utils.core :as utils]
```
dankreek: Should we stick to the `:as ssl-utils` convention here?

dankreek:
All-in-all, LGTM, I just had the one question about using `:as ssl-utils` in `certificate_authority_test.clj`

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
nwolfe: We didn't in master because we're assuming the utils were for SSL since this is the CA namespace...and I tried to make the same changes in both branches in case it might otherwise make merging more difficult

dankreek: Ahh ok, that makes sense. And this was as fairly IntelliJ/Cursive-centric comment.

dankreek:
:+1: I'm for it!

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
nwolfe: Yeah, I was following the existing pattern here, but I do think that it would be better for us to start being consistent with ns aliases regardless of what ns we're using them in.

project.clj
camlow325: `http-client 0.4.1` replaced the certificate-authority dependency with a dependency on `ssl-utils 0.7.0`.  I have a PR up for clj-http-client that would bump this up to `ssl-utils 0.8.0` - https://github.com/puppetlabs/clj-http-client/pull/36.

camlow325:
:+1: Tests all pass locally for me.  Tired of waiting for :turtle: Travis, so merging...

PR #455, opened 2015-03-16
(MAINT) Restore jira-ruby gem dependency

camlow325:
The Gemfile on puppet-server's master branch has 'jira-ruby' as a
development gem.  This commit adds that gem for the stable branch as
well.

This commit also makes 'rake' a development and test gem and sources a
Gemfile local if one is defined for the user.  These changes were done
just to replicate the behavior we now have on the master branch.


fpringvaldsen:
:+1:

nwolfe:
:+1:

PR #456, opened 2015-03-17
(SERVER-379) Initial draft of a tuning guide

cprice404:


documentation/tuning_guide.markdown
```diff
@@ -0,0 +1,156 @@
...
+the best performance in your environment.
+
+## Puppet Server and JRuby
+
+Before you begin tuning your configuration, it's helpful to have a little bit
+of context on how Puppet Server uses JRuby to handle incoming HTTP requests from
+your Puppet agents.
+
+When Puppet Server starts up, it creates a pool of JRuby interpreters to use
+as workers when we need to execute some of the Puppet Ruby code.  You can think
```
KevinCorcoran: I think it's preferable to avoid the "we" working here ... something better would be "When Puppet Server starts up, it creates a pool of JRuby interpreters to use as workers when __it needs__ to execute some of the Puppet Ruby code."

@nfagerlund you're welcome :)

documentation/tuning_guide.markdown
```diff
@@ -0,0 +1,156 @@
...
+of these almost as invidual Ruby "virtual machines" that are controlled by
```
KevinCorcoran: typo in "invidual"

documentation/tuning_guide.markdown
```diff
@@ -0,0 +1,156 @@
...
+Puppet Server; it's not entirely dissimilar to the way that Passenger spawns
+several MRI Ruby processes to hand off work to.
```
KevinCorcoran: "MRI Ruby" is redundant

documentation/tuning_guide.markdown
```diff
@@ -0,0 +1,156 @@
...
+executed in order to handle the request, Puppet Server "borrows" a JRuby instance
+from the pool, uses it to do the work, and then "returns" it to the pool.  If
+there are no JRuby instances available in the pool at the time when a request
+comes in (presumably because all of the JRuby instances are already in use handling
+other requests), Puppet Server will block the request until one becomes available.
+
+(In the future, this approach will allow us to do some really powerful things
+such as creating multiple pools of JRubies and isolating each of your Puppet
+environments to a single pool, to ensure that there is no pollution from one
+Puppet environment to the next.)
```
KevinCorcoran: I'd probably omit these parentheses.

documentation/tuning_guide.markdown
```diff
@@ -0,0 +1,156 @@
...
+
+This brings us to the two most important settings that you can use to tune your
+Puppet Server.
+
+### Number of JRubies
+
+The most important setting that you can use to improve the throughput of your
+Puppet Server installation is the [`max-active-instances`](./configuration.html#puppetserver_conf)
```
KevinCorcoran: this link looks broken to me, but perhaps that's just an artifact of how ours docs are laid-out?

documentation/tuning_guide.markdown
```diff
@@ -0,0 +1,156 @@
...
+`max-active-instances` to a value higher than the number of CPUs.)
+
+At this point you may be wondering "what's the downside to just setting
+`max-active-instances` to a really high value?"  The answer to this question, in
+a nutshell, is "memory usage".  This brings us to the other extremely important
+Puppet Server.
+
+### JVM Heap Size
+
+The JVM's "max heap size" controls the maximum amount of (heap*[[1]](#footnotes)
```
KevinCorcoran: errant asterisk in "heap*" ?

documentation/tuning_guide.markdown
```diff
@@ -0,0 +1,156 @@
...
+smoothly, you will eventually encounter an `OutOfMemoryError`, and the process
+will shut down.  (For Puppet Server, we also use a JVM argument called
+`-XX:HeapDumpOnOutOfMemoryError` to cause the JVM to dump an `.hprof` file to
+disk, which is basically a memory snapshot from the point in time where the
+error occurred; this can be loaded into various profiling tools to get a better
+understanding of where the memory was being used.)
+
+(Note that there is another setting, "min heap size", that is controlled via
+the -Xms setting; [Oracle recommends](http://www.oracle.com/technetwork/java/gc-tuning-5-138395.html#0.0.0.%20Total%20Heap|outline)
+setting this value to the same value that you use for -Xmx.)
```
KevinCorcoran: omit parens here and/or the end of the previous paragraph as well.

documentation/tuning_guide.markdown
```diff
@@ -0,0 +1,156 @@
...
+
+The most important factor when determining the max heap size for Puppet Server
+is the value of `max-active-instances`.  Each JRuby instance needs to load up
+a copy of the Puppet Ruby code, and then needs some amount of memory overhead
+for all of the garbage that gets generated during a Puppet catalog compilation.
+Also, the memory requirements will vary based on how many Puppet modules you
+have in your module path, how much hiera data you have, etc.  At this time we're
+estimating that a reasonable ballpark figure is about 512MB of RAM per JRuby
+instance, give or take a bit based on the size of your Puppet codebase.
```
KevinCorcoran: I might add to or change the wording in this paragraph to more strongly indicate that memory usage per JRuby can vary wildly and, while 512MB may be a reasonable starting spot for most installations, if a user cares about tuning that parameter they can expect to end up with something very different.

KevinCorcoran:
Looks pretty good to me!  I imagine @nfagerlund should have a look at this?

documentation/tuning_guide.markdown
```diff
@@ -0,0 +1,156 @@
...
+From a practical perspective, this setting basically controls how many Puppet
+agent runs Puppet Server can handle concurrently.  The minimum value you can
+get away with here is `1`, and if your installation is small enough such that
+you're unlikely to ever have more than one Puppet agent checking in with the
+server at exactly the same time, this will be totally sufficient.
+
+However, if you specify a value of `1` for this setting, and then you have two
+Puppet agent runs hitting the server at the same time, what will happen is that
+the requests being made by the second agent will be effectively blocked until
+the server has finished handling all of the requests from the first agent.
```
camlow325: Not sure if it's critical to cover this here, but what might happen in this scenario is that the requests might experience periodic delays when trying to get access to a JRuby instance rather than one agent owning the JRuby instance for the duration of the requests it would make for one agent run.  For example, the second agent's catalog request might be blocked temporarily while the first agent's catalog request is still in process.  While the second agent's catalog request is in process later, the first agent's follow-up file resource request might be blocked.  Exclusive access to a JRuby instance is only maintained for the duration of a single HTTP request between the agent and master as opposed to being for the duration of all requests on a single socket connection between the agent and master.

documentation/tuning_guide.markdown
```diff
@@ -0,0 +1,156 @@
...
+to actually do any work if all of the CPU cores are already busy using JRuby
+instances to handle incoming agent requests.
+
+(There are exceptions to this rule; if you have report processors that make
+a network connection as part of the processing of a report, and if there is a chance
+that the network operation is slow and will block on I/O for some period of time,
+then it may make sense to have more JRuby instances than the number of cores,
+because the JVM will be smart enough to suspend the thread that is handling those
+kinds of requests and use the CPUs for other work, assuming there are still JRuby
+instances available in the pool.  In a case like this you might want to set
```
camlow325: Not sure if this would be obvious to the reader, but if a thread holding a JRuby instance were suspended, the JRuby instance would not implicitly be returned to the pool while suspended and reacquired when the thread is resumed later.  The JRuby instance would only be freed up around the time that the response from the request is being delivered by the master.

documentation/tuning_guide.markdown
```diff
@@ -0,0 +1,156 @@
...
+a copy of the Puppet Ruby code, and then needs some amount of memory overhead
+for all of the garbage that gets generated during a Puppet catalog compilation.
+Also, the memory requirements will vary based on how many Puppet modules you
+have in your module path, how much hiera data you have, etc.  At this time we're
+estimating that a reasonable ballpark figure is about 512MB of RAM per JRuby
+instance, give or take a bit based on the size of your Puppet codebase.
+
+You'll also want to allocate a little extra heap to be used by the rest of the
+things going on in Puppet Server; the web server, etc.  So, a good rule of thumb
+might be 512MB + (max-active-instance * 512MB).
```
camlow325: s/instance/instances

camlow325:
:+1: with some minor comments, which would be fine by me to ignore if it would be too tedious to cover.

documentation/tuning_guide.markdown
cprice404: I intentionally avoiding getting into that level of detail because I thought it would just make things more confusing.  I'm happy to revisit that if @jbondpdx or @nfagerlund suggests that it would be worthwhile.

cprice404:
I'm going to fix the obvious typos that were pointed out, and probably try to address @KevinCorcoran 's suggestion about being more explicit about the variability of that 512MB number, but I think that beyond that it'd probably be best to wait for @jbondpdx and/or @nfagerlund to weigh in before changing anything else.

@jbondpdx @nfagerlund would it be easier for you all to review the PR and provide feedback?  Or would it maybe make more sense to just go ahead and get this merged after I address those issues I mentioned above and then let you guys file a fix-up PR over the top of this one?

documentation/tuning_guide.markdown
camlow325: k, I'm okay with that.

documentation/tuning_guide.markdown
cprice404: Yeah I have no idea if that's still the right way to do the links but that's how it used to work.  Hopefully @jbondpdx can let me know if that's not right.

documentation/tuning_guide.markdown
cprice404: That was supposed to indicate a footnote.

cprice404:
Added a commit to address most of the feedback.

@jbondpdx @nfagerlund I'm kind of leaning towards going ahead and getting this merged in and then letting you all come back and take a pass at it as your schedule permits.  Any objections?

documentation/tuning_guide.markdown
KevinCorcoran: I don't think I've ever seen an asterisk used before the "[1]" like that before, and it looks weird to me.  Is that a convention I'm unaware of?  :bike:

cprice404: I'm sure the whole way i'm formatting the footnotes thing is wrong/terrible, and maybe it shouldn't even be a footnote at all... this seems like the kind of thing where we should just let the docs team tell us what to do rather than guessing though.

camlow325:
@jbondpdx @nfagerlund - you good with us merging this?  Hoping to get this wrapped up ahead of our planned release this week?

jbondpdx:
I'm sorry for my lack of response on these tags! I kept meaning to get back to them, you know, "right after this thing." Last week was pretty hectic. I'm ok with this getting merged and taking a pass at it later if @nfagerlund is.

nfagerlund:
Yeah, docs team is kind of out of capacity right now. :( I too am cool with merge-and-review-later. I've asked Mike to see if he can find some editing/review time from someone else.

camlow325:
:+1: merging since @jbondpdx and @nfagerlund are good with it for now.

PR #457, opened 2015-03-17
(SERVER-454) Remove exception handler in config service test

jeffmccune:
Without this patch the config service test uses an exception hander to figure
out if a semver identifier is valid or not.  This is a problem because the
semver function may not have raised the exception being handled.  This patch
addresses the problem by updating the clj-semver library to 0.3.0 which
implements a semver check function that returns boolean true or false.


camlow325:
:+1:

dankreek:
Kim loves it, and so do I! :+1:

PR #458, opened 2015-03-20
(MAINT) Bump lein-ezbake plugin dependency to 0.2.6

camlow325:
This commit bumps puppet-server's lein-ezbake plugin dependency to
0.2.6.


nwolfe:
:+1:

PR #459, opened 2015-03-21
(SERVER-499) Get smoke tests passing for windows agents

jeffmccune:
Without this set of patches windows agents error out in the pre-suite.  This is
a problem because we need to test windows2008r2-64a against redhat7 and
ubuntu1404.

These patches install the AIO agent onto windows hosts.  It's unclear if the
full acceptance suite passes, but I'd like to get this up for review.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,7 +17,7 @@ group :test do
   gem 'rspec'
   # gem 'beaker', *location_for(ENV['BEAKER_LOCATION'] || '~> 2.2')
   # See: SERVER-435 for the reasoning behind this specific version
-  gem 'beaker', *location_for(ENV['BEAKER_LOCATION'] || 'git://github.com/puppetlabs/beaker#7d0cccf22cc7289113de862fe4a4494872f574bb')
+  gem 'beaker', *location_for(ENV['BEAKER_LOCATION'] || 'git://github.com/puppetlabs/beaker#beaker2.7.1')
   if ENV['GEM_SOURCE'] =~ /rubygems\.delivery\.puppetlabs\.net/
     gem 'sqa-utils', '~> 0.11'
   end
--- a/Gemfile
+++ b/Gemfile
@@ -15,9 +15,7 @@ gem 'jira-ruby', :group => :development
 
 group :test do
   gem 'rspec'
-  # gem 'beaker', *location_for(ENV['BEAKER_LOCATION'] || '~> 2.2')
-  # See: SERVER-435 for the reasoning behind this specific version
-  gem 'beaker', *location_for(ENV['BEAKER_LOCATION'] || 'git://github.com/puppetlabs/beaker#beaker2.7.1')
+  gem 'beaker', *location_for(ENV['BEAKER_LOCATION'] || '2.7.1')
   if ENV['GEM_SOURCE'] =~ /rubygems\.delivery\.puppetlabs\.net/
     gem 'sqa-utils', '~> 0.11'
   end
--- a/acceptance/suites/pre_suite/foss/10_install_release_repos.rb
+++ b/acceptance/suites/pre_suite/foss/10_install_release_repos.rb
@@ -1,5 +1,8 @@
 step "Setup Puppet Labs Release repositories." do
   hosts.each do |host|
-    install_puppetlabs_release_repo host
+    platform = host.platform
+    if not /windows/.match(platform)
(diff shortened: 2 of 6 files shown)


jeffmccune:
This is ready for merge.  [puppet-server (aio) 20:  Smoke Acceptance Test Suite build #53](https://jenkins.puppetlabs.com/view/Puppet%20Server/view/aio/job/experimental_puppet-server_integration-system_smoke-aio/53/) passed on everything except debian8 and fedora20 layouts.

Gemfile
```diff
@@ -17,7 +17,7 @@ group :test do
   gem 'rspec'
   # gem 'beaker', *location_for(ENV['BEAKER_LOCATION'] || '~> 2.2')
   # See: SERVER-435 for the reasoning behind this specific version
-  gem 'beaker', *location_for(ENV['BEAKER_LOCATION'] || 'git://github.com/puppetlabs/beaker#7d0cccf22cc7289113de862fe4a4494872f574bb')
+  gem 'beaker', *location_for(ENV['BEAKER_LOCATION'] || 'git://github.com/puppetlabs/beaker#beaker2.7.1')
```
rlinehan: Why can't we use beaker from a released version? Do we just want to keep the git stuff in here in case of future need?

acceptance/suites/pre_suite/foss/10_install_release_repos.rb
```diff
@@ -1,5 +1,8 @@
 step "Setup Puppet Labs Release repositories." do
   hosts.each do |host|
-    install_puppetlabs_release_repo host
+    platform = host.platform
```
rlinehan: Would it be possible to use `confine :to, :platform => 'windows'`? I don't actually know how that works/where it can be used, but this does seem like the place for it, if possible.

acceptance/suites/pre_suite/foss/30_install_dev_repos.rb
```diff
@@ -14,7 +14,10 @@
 if puppet_build_version
   step "Setup Puppet Labs Dev Repositories." do
     hosts.each do |host|
-      install_puppetlabs_dev_repo host, 'puppet-agent', puppet_build_version
+      platform = host.platform
+      if not /windows/.match(platform)
```
rlinehan: Same question as above regarding `confine :to, :platform => 'windows'`?

Gemfile
jeffmccune: In theory, this should be identical to using a release version of beaker by pinning to the exact version, but I'm guessing you're asking about using rubygems to install the dependency rather than bundler + git to install the dependency.

I've updated the change set to use the released gem, please let me know if that doesn't fully address the concern.

acceptance/suites/pre_suite/foss/10_install_release_repos.rb
jeffmccune: I'll take a look.  I didn't know about `confine :to, :platform => 'windows'` either, until very recently.

jeffmccune: Reading the documentation, looks like this is possible to do with `confine_block`.  `confine` statefully operates on the entire test case instance so I don't think it would be suitable here because we're doing things with different confinements in the same file.

jeffmccune:
@rlinehan Thanks for your review and comments.  I rebased this PR on master and included a comment that I think addresses all of the comments so far.

acceptance/suites/pre_suite/foss/70_install_puppet.rb
```diff
@@ -1,8 +1,49 @@
...
+
 step "Install MRI Puppet Agents."
   hosts.each do |host|
+    platform = host.platform
+
     puppet_version = test_config[:puppet_version]
 
-    if puppet_version
+    if /windows/.match(platform)
+      arch = host[:ruby_arch] || 'x86'
```
dankreek: Defaults to x86? Is that normal in Windows-land?

KevinCorcoran: it's like bizarro world

dankreek: Only more bizarre?

dankreek:
:+1: LGTM. Let's merge:

nwolfe:
:+1:

PR #460, opened 2015-03-23
(SERVER-377) gem doesn't work behind proxy

GeoffWilliams:
This is workaround for SERVER-377 (gem doesn't work behind proxy)

The PR checks for the presence of the `http_proxy` variable.  If set, it is split up into its component parts using `awk`  If no variable is found, we attempt to parse the `http_proxy_*` entries in the puppet.conf file.

Care has been taken to protect against:
* Only username being set (no password)
* No username or password (http_proxy)
* Don't attempt to use proxy if no proxy host is found
* Remove whitespace from entries parsed from puppet.conf
* Processing of commented out lines in puppet.conf
* Passwords cannot contain spaces as its too hard to tokenize

If after performing this parse, there is no value found for the proxy server host, the system will not attempt to use a proxy.  A message is printed indicating where the proxy was found (config file or variable) if the proxy is to be used.  The main method of debugging this script would be to put `set -x` at the top of it.

I was able to test the final version of the script which I've attached to SERVER-377 but I couldn't figure out how to compile this `.erb` file into the final script so have just copied in the tested code and created a PR from it.

### DIFF
--- a/resources/ext/cli/gem.erb
+++ b/resources/ext/cli/gem.erb
@@ -1,5 +1,48 @@
 #!/usr/bin/env bash
 
-"${JAVA_BIN}" -cp "${INSTALL_DIR}/<%= EZBake::Config[:uberjar_name] %>" \
+function parse_config_file() {
+    awk 'BEGIN{FS="="} /^[[:space:]]*'$1'/ {gsub(/ /, "",$2) ; print $2}' < "$(puppet config print confdir)/puppet.conf"
+}
+
+function parse_env() {
+    echo $http_proxy| awk "$1"
+}
+
+if [ -n "$http_proxy" ] ; then
+    HTTP_PROXY_HOST=$(parse_env 'BEGIN{FS=".*//(.+@)?"} {split($2,a,":"); print a[1]}')
+    HTTP_PROXY_PORT=$(parse_env 'BEGIN{FS=".*//(.+@)?"} {split($2,a,":"); print a[2]}')
+    HTTP_PROXY_USER=$(parse_env 'BEGIN{FS="//|@"} /@/ {split($2,a,":"); print a[1]}')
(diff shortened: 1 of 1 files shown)


acidprime:
@GeoffWilliams I think the awk is not really needed here, as this is specific to puppet master , I think something akin to `puppet master --configprint http_proxy_host` would work no? Rinse and repeat for each setting.

acidprime:
My guess is there is likely a less impact change to make this work but I have not tried to take this apart yet.

GeoffWilliams:
@acidprime FACEPALM!

GeoffWilliams:
@beergeek suggests that we search each section of the config file in order until something is found:
_I would search in order of_:
* cmdline (environment)
* user
* master
* main

It's possible to use the config face to read different sections of the config file, like this:
```
/opt/puppet/bin/puppet config print --section user http_proxy_host
```

Am working on a revised PR to do this

GeoffWilliams:
@acidprime I've fixed the script to use the config face and scan for proxy settings in the `user` section which is where @beergeek suggests these settings should live.  This will do a hierarchical parse of `user`, `main` and `defaults.rb` (contains port), however this will ignore anything present in the `master` section.

It's not possible to pass usernames and passwords to a proxy server without writing some java (see ticket).  I previously assumed that the settings for username and password chris mentioned worked but after setting up a squid server to test on I proved this wasn't the case, so I've updated this code to simply fail and say authenticated proxies aren't supported.

The _real_ fix needs to be done in clojure/java.  The file `./src/clj/puppetlabs/puppetserver/cli/gem.clj` seems to be the entry point.  I know how to fix this in Java but not Clojure so am not going to even attempt this.  Also I can't see the _real_ gem program anywhere so I don't know if it would even use the `java.net.*` classes or has its own magic.

Sorry to not have a better fix for you all.  For completeness I've reworked things anyway but hopefully a _proper_ fix will be in the next release(?)

GeoffWilliams:
closed in favor of https://github.com/puppetlabs/puppet-server/pull/463

PR #462, opened 2015-03-23
(SERVER-449) Expose HTTP client timeout settings

dankreek:


dev/puppet-server.conf.sample
```diff
@@ -65,6 +65,16 @@ http-client: {
...
+
+    # The amount of time, in milliseconds, that  an outbound HTTP connection
+    # will wait for no data to be available before closing the socket. If not
+    # defined, defaults to 20 minutes.
+    #socket-timeout: 1200000
+
+    # The mount of time, in milliseconds, that an output HTTP connection will
+    # wait to connect to connect before giving up. It not defined, defaults to
+    # 20 minutes.
+    #connect-timeout: 1200000
```
camlow325: I'd prefer that we use a default of 2 minutes here since it aligns better with Puppet's client timeout.

I think connection timeouts should generally be much shorter than socket timeouts.  Socket timeouts can relatively be more damaging.  In the socket timeout case, the server might go and and perform the request it was asked to do even if the client gave up on getting the response back after some period of time.  In the connect timeout case, however, the server wouldn't have even gotten to the point of starting to act on the client's request so there's nothing to abort on the server side.

dev/puppet-server.conf.sample
```diff
@@ -65,6 +65,16 @@ http-client: {
     #                TLS_RSA_WITH_AES_256_CBC_SHA,
     #                TLS_RSA_WITH_AES_128_CBC_SHA256,
     #                TLS_RSA_WITH_AES_128_CBC_SHA]
+
+    # The amount of time, in milliseconds, that  an outbound HTTP connection
```
camlow325: Cosmetic - extra space between "that" and "an".

dev/puppet-server.conf.sample
```diff
@@ -65,6 +65,16 @@ http-client: {
...
+    # will wait for no data to be available before closing the socket. If not
+    # defined, defaults to 20 minutes.
+    #socket-timeout: 1200000
+
+    # The mount of time, in milliseconds, that an output HTTP connection will
+    # wait to connect to connect before giving up. It not defined, defaults to
```
camlow325: Extra "to connect" here.

documentation/configuration.markdown
```diff
@@ -93,6 +93,16 @@ http-client: {
     #                TLS_RSA_WITH_AES_256_CBC_SHA,
     #                TLS_RSA_WITH_AES_128_CBC_SHA256,
     #                TLS_RSA_WITH_AES_128_CBC_SHA]
+    
+    # The amount of time, in milliseconds, that  an outbound HTTP connection
```
camlow325: Extra space again.

documentation/configuration.markdown
```diff
@@ -93,6 +93,16 @@ http-client: {
...
+    # will wait for no data to be available before closing the socket. If not
+    # defined, defaults to 20 minutes.
+    #socket-timeout: 1200000
+
+    # The mount of time, in milliseconds, that an output HTTP connection will
+    # wait to connect to connect before giving up. It not defined, defaults to
```
camlow325: Extra "to connect" again.

dev/puppet-server.conf.sample
```diff
@@ -65,6 +65,16 @@ http-client: {
     #                TLS_RSA_WITH_AES_256_CBC_SHA,
     #                TLS_RSA_WITH_AES_128_CBC_SHA256,
     #                TLS_RSA_WITH_AES_128_CBC_SHA]
+
+    # The amount of time, in milliseconds, that  an outbound HTTP connection
+    # will wait for no data to be available before closing the socket. If not
+    # defined, defaults to 20 minutes.
+    #socket-timeout: 1200000
+
+    # The mount of time, in milliseconds, that an output HTTP connection will
```
camlow325: s/mount/amount

documentation/configuration.markdown
```diff
@@ -93,6 +93,16 @@ http-client: {
     #                TLS_RSA_WITH_AES_256_CBC_SHA,
     #                TLS_RSA_WITH_AES_128_CBC_SHA256,
     #                TLS_RSA_WITH_AES_128_CBC_SHA]
+    
+    # The amount of time, in milliseconds, that  an outbound HTTP connection
+    # will wait for no data to be available before closing the socket. If not
+    # defined, defaults to 20 minutes.
+    #socket-timeout: 1200000
+
+    # The mount of time, in milliseconds, that an output HTTP connection will
```
camlow325: s/mount/amount

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
```diff
@@ -1,5 +1,6 @@
 (ns puppetlabs.services.jruby.jruby-puppet-service
   (:require [clojure.tools.logging :as log]
+            [clojure.set :as set]
```
nwolfe: "This import appears to be unused"

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -74,14 +74,16 @@
 
     * :borrow-timeout - The timeout when borrowing instances from the JRuby Pool
```
camlow325: Would be good to add a description of the `http-client-connect-timeout` and `http-client-socket-timeout` keys here for consistency.

camlow325:
Wherever the `http-client-connect-timeout` and `http-client-socket-timeout` are referenced in docs, I'd include a comment that a value of 0 is interpreted as "infinite" (or unlimited) and that a negative value is interpreted as "undefined (system default)".  This is how we described these in clj-http-client.

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
```diff
@@ -15,6 +16,16 @@
...
   1200000)
 
+(def default-http-connect-timeout
+  "The default number of milliseconds that the client will wait for a connection
+  to be established. Currently set to 20 minutes."
+  (* 2 60 1000))
+
+(def default-http-socket-timeout
+  "The default number of milliseconds that the client will allow for no data to
+  be available on the socket. Currently set to 20 minutes."
```
nwolfe: Docstrings still say 20 minutes instead of 2 now

camlow325:
I think we've been going with a units qualifier on settings like these when we've been adding them to Trapperkeeper-related projects lately, e.g., `http-client-connect-timeout-milliseconds`.

@cprice404 - do you think we should do the same here?

I guess we didn't do this with the {{borrow-timeout}} one when we added it recently though...

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
```diff
@@ -15,6 +16,16 @@
    milliseconds. Current value is 1200000ms, or 20 minutes."
   1200000)
 
+(def default-http-connect-timeout
+  "The default number of milliseconds that the client will wait for a connection
+  to be established. Currently set to 20 minutes."
```
nwolfe: Docstring still says 20 minutes

cprice404:
@camlow325 I generally like having the units in the setting name but we haven't been that consistent about it, so, your call.

camlow325:
:+1: I'd feel a little more comfortable with the Ruby spec tests if they validated that the expected `SimpleRequestOptions` were actually being passed into the `clj-http-client` client layer but I understand this would add some complexity.  I did do some functional-layer validation to see that with a puppetserver configured to talk to a "puppetdb" report processor that was either not listening on the socket or was stalling for a long time in responding that the expected connect and response timeouts were being honored.

I'm good with keeping the settings named as they are now if @cprice404 doesn't have a strong preference to change them.  Would like to get this PR in soon and not sure that the units distinction is critical.

dev/puppet-server.conf.sample
```diff
@@ -65,6 +65,18 @@ http-client: {
     #                TLS_RSA_WITH_AES_256_CBC_SHA,
     #                TLS_RSA_WITH_AES_128_CBC_SHA256,
     #                TLS_RSA_WITH_AES_128_CBC_SHA]
+
+    # The amount of time, in milliseconds, that an outbound HTTP connection
+    # will wait for no data to be available before closing the socket. If not
```
jeffmccune: s/no //

spec/puppet-server-lib/puppet/jvm/http_client_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    it "timeout settings are properly set" do
+      Puppet::Server::HttpClient.initialize_settings(
+          {"http_connect_timeout" => connect_timeout,
+           "http_socket_timeout"  => socket_timeout})
+
+      client = Puppet::Server::HttpClient.new(nil, 0, {})
+      request_options = SimpleRequestOptions.new("http://i.love.ruby")
+      client.send(:configure_timeouts, request_options)
+      request_options.get_socket_timeout_milliseconds.should == socket_timeout
+      request_options.get_connect_timeout_milliseconds.should == connect_timeout
```
nwolfe: Might rewrite these in RSpec 3 format since we have a ticket about getting rid of the deprecation warnings.
```ruby
expect(request_options.get_socket_timeout_milliseconds).to equal(socket_timeout)
```
...or whatever it is

dev/puppet-server.conf.sample
```diff
@@ -65,6 +65,18 @@ http-client: {
...
+    # defined, defaults to 20 minutes. If 0, the timeout is infinite and if
+    # negative, the value is undefined (system default).
```
jeffmccune: If you end up back in here, maybe an improvement might be, "is undefined in the application, the behavior is governed by the system default behavior"

It's a little confusing to me as it currently stands.

It's also worth asking...  Are we talking about a socket in the connected state, or are we talking about the timeout associated with establishing a TCP connection before it enters the connected state?

If we're talking about the former, a connected socket should ideally persist indefinitely even though it's idle with no data flowing.

spec/puppet-server-lib/puppet/jvm/http_client_spec.rb
dankreek: Oh duh... I'm the one who made that ticket.

dev/puppet-server.conf.sample
```diff
@@ -65,6 +65,18 @@ http-client: {
...
+    #socket-timeout: 1200000
+
+    # The amount of time, in milliseconds, that an output HTTP connection will
```
jeffmccune: "HTTP request" may be more conventional than "output HTTP connection" ?

dev/puppet-server.conf.sample
nwolfe: I see this same text in multiple places

dev/puppet-server.conf.sample
jeffmccune: It also occurs to me...  If we're talking about a TCP connection that's in the connected state, then the conventional phrasing of an "idle connection" might be more clear to the reader.

dev/puppet-server.conf.sample
dankreek: Wow.. I don't remember writing that. But it is indeed horrible.

jeffmccune:
@camlow325 mentioned...

> I'm good with keeping the settings named as they are now if @cprice404 doesn't have a strong preference to change them.

I think I'd like to change the name of `:socket-timeout`, but for reasons in addition to @cprice404 

My reasoning is that there are a lot of timeout values associated with socket connections, and I'm not sure which one we're talking about.  It seems like we're talking about an idle connection timeout, so maybe we should rename this to something like `:socket-idle-connection-timeout-ms` or something to that effect?

dev/puppet-server.conf.sample
camlow325: > "is undefined in the application, the behavior is governed by the system default behavior"

Yeah, that sounds like an improvement and accurate.

> Are we talking about a socket in the connected state, or are we talking about the timeout associated with establishing a TCP connection before it enters the connected state?

The `socket-timeout` would apply to a socket in the connected state.  If the value were set to 0 and, assuming the underlying network layers were cooperating, the socket would persist indefinitely.  The "20 minute" value is a defensive measure to guard against the case that the server otherwise might leave the socket open indefinitely as well.  We saw this to be the case with a thundering herd testing scenario where Puppet Server was connected to a PuppetDB which was essentially hanging the connections indefinitely because it was unable to write data to the disk.  20 minutes ideally should be way longer than a reasonable client request should take.

dev/puppet-server.conf.sample
jeffmccune: You should see some of my comments from the Puppet codebase.  "Let me see if I understand my grip on reality..." comes to mind.

camlow325:
> It seems like we're talking about an idle connection timeout, so maybe we should rename this to something like :socket-idle-connection-timeout-ms or something to that effect?

I'm still open to a name change if it helps clarify.  `socket-idle-connection-timeout-ms` feels a bit wordy to me, though, and I would then worry about the word *connection* in the middle leading people to conflate it with the separate `connect-timeout` setting.

Another option would be to just rename `socket-timeout` to `idle-timeout-millliseconds`.  We used that name for the equivalent server setting in the webserver - see https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/1.2.0/doc/jetty-config.md#idle-timeout-milliseconds.

If we're going to do that anyway, might as well rename `connect-timeout` to `connect-timeout-milliseconds`.

jeffmccune:
:+1: `:idle-timeout-missiseconds`

cprice404:
I'm +1 on changing it to include a unit but I feel like maybe we've used `millis` instead of `milliseconds` on other settings in other repos?  Not 100% sure.

camlow325:
All the ones I'm aware of that are expressed in milliseconds either have no units in the name or have "milliseconds".  The ones that have "milliseconds" are "idle-timeout-milliseconds" in tk-jetty9 webserver configuration and "connect-timeout-milliseconds" and "socket-timeout-milliseconds" in clj-http-client.  Didn't find any offhand in puppet-server that are expressed in milliseconds and already have the units in the name.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -72,16 +72,28 @@
     * :http-client-cipher-suites - A list of legal SSL cipher suites that may
         be used when https client requests are made.
 
+    * :http-client-connect-timeout - The amount of time, in milliseconds, that
+        an outbound HTTP connection will wait for no data to be available before
+        closing the socket. Defaults to 2 minutes. If 0, the timeout is infinite
+        and if negative, the value is undefined (system default).
+
+    * :http-client-socket-timeout - The amount of time, in milliseconds, that an
+        output HTTP connection will wait to connect before giving up. Defaults
```
jeffmccune: Suggestion: The amount of time, in milliseconds, that an HTTP client will wait to receive data in the response before giving up.

Also, suggest changing this to `:http-client-idle-timeout-milliseconds` to match up with `:idle-timeout-milliseconds`

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -72,16 +72,28 @@
...
+        to 20 minutes if not set. If 0, the timeout is infinite and if negative,
+        the value is undefined (system default).
```
jeffmccune: Same comment as above regarding system default.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -72,16 +72,28 @@
     * :http-client-cipher-suites - A list of legal SSL cipher suites that may
         be used when https client requests are made.
 
+    * :http-client-connect-timeout - The amount of time, in milliseconds, that
+        an outbound HTTP connection will wait for no data to be available before
+        closing the socket. Defaults to 2 minutes. If 0, the timeout is infinite
```
jeffmccune: This description of `:http-client-connect-timeout` sounds like it's describing the behavior of the timeout associated with a connected socket that's idling, but I think that's the behavior of `:http-client-socket-timeout` unless I'm mistaken.

jeffmccune: (I think these descriptions just got flipped around)

jeffmccune:
I'm :+1: except for my comments on naming and descriptions.

dankreek:
@jeffmccune @camlow325 I've updated the docs and those config setting names. Ya'll good?

dev/puppet-server.conf.sample
```diff
@@ -65,6 +65,18 @@ http-client: {
     #                TLS_RSA_WITH_AES_256_CBC_SHA,
     #                TLS_RSA_WITH_AES_128_CBC_SHA256,
     #                TLS_RSA_WITH_AES_128_CBC_SHA]
+
+    # The amount of time, in milliseconds, that an outbound HTTP connection
+    # will wait for data to be available before closing the socket. If not
+    # defined, defaults to 20 minutes. If 0, the timeout is infinite and if
+    # negative, the value is undefined (system default).
```
camlow325: Probably would be good to have the same text here as below for `connect-timeout-milliseconds`:

> if negative, the value is undefined in the application and governed by the system default behavior.

documentation/configuration.markdown
```diff
@@ -93,6 +93,18 @@ http-client: {
     #                TLS_RSA_WITH_AES_256_CBC_SHA,
     #                TLS_RSA_WITH_AES_128_CBC_SHA256,
     #                TLS_RSA_WITH_AES_128_CBC_SHA]
+    
+    # The amount of time, in milliseconds, that an outbound HTTP connection
+    # will wait for data to be available before closing the socket. If not
+    # defined, defaults to 20 minutes. If 0, the timeout is infinite and if
+    # negative, the value is undefined (system default).
+    #idle-timeout-milliseconds: 1200000
```
camlow325: Same comment here as for the other `idle-timeout-milliseconds` reference in "puppet-server.conf.sample".

camlow325:
:+1: again with just the couple of cosmetic comments on the "the value is defined (system default)" references, assuming Travis passes.

dankreek:
Allright.. tied a bow on it.

jeffmccune:
:+1:

PR #463, opened 2015-03-24
SERVER-377 support for (authenticated) proxies in puppetserver gem command

GeoffWilliams:
Support proxy access for the `puppetserver gem` command, including proxies requiring users to authenticate with username and password.

The PR works by building an `http_proxy` environment variable from the puppet config file if one isn't already exported from the bash instance we're running in.

The clojure code has been modified in `gem.clj` to pass the `http_proxy` variable to the underlying jruby environment which allows natural proxy support (including basic authentication)

### DIFF
--- a/resources/ext/cli/gem.erb
+++ b/resources/ext/cli/gem.erb
@@ -1,5 +1,4 @@
 #!/usr/bin/env bash
-
 PATH=/opt/puppet/bin:$PATH
 
 function parse_config_file() {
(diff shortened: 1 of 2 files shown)


GeoffWilliams:
Proxy settings should go in the `[user]` section of `puppet.conf`.  They will _work_ if you put them in `[main]` but during testing I observed spurious requests being made to my proxy (presumably from other parts of Puppet) if these were allowed to remain.

GeoffWilliams:
Tests seem to be failing with "no output received" but I don't understand why? -- there isn't a `http_proxy`/`https_proxy` variable set on the build server is there?

Couldn't figure out how to run the tests locally, get a bunch of errors like this:
```
Caused by: org.jruby.exceptions.RaiseException: (LoadError) no such file to load -- puppet
 at org.jruby.RubyKernel.require (org/jruby/RubyKernel.java:1071)
```
Which are completely different to the dieing build on the server

rlinehan:
@GeoffWilliams this looks like it may have been a transient Travis error; I've kicked it to re-run.

The error you're getting sounds like it might be because you haven't pulled down the submodules? Try running `git submodule init && git submodule update``, and then re-running the tests using `lein test`. In addition to the lein (clojure) tests, we also have rspec tests, which can be run by doing `rspec spec`.

We're going to hold off on doing much review until we've determined what approach we want to take to solve this problem, which probably depends on a discussion about how we handle environment variables in these subcommands (SERVER-297). I think we're going to try to figure that out in the next week or so; we'll leave this open until we've got a clearer idea what the right path forward is.

GeoffWilliams:
that explains it - thanks! build is now passing :)

jeffmccune:
@GeoffWilliams Thanks for taking the time to submit this pull request.  We decided to go another route to solve the HTTP proxy issue based on the conversations around SERVER-297 and SERVER-377.  I'm going to close this for the time being since this pull request partially addresses the http proxy problem.

If you'd still like this particular change made ahead of our plans to fix SERVER-297 and SERVER-377, please just let one of us on the server team know and we'll see what we can do to get things improved.

PR #464, opened 2015-03-24
adding acceptance test to check puppetserver service after reboot

erikPrime:
This adds an acceptance test that validates puppetserver is running after reboot, and a script to make launching an acceptance test easier.

### DIFF
--- a/acceptance/suites/tests/555-puppetserver-service/reboot.rb
+++ b/acceptance/suites/tests/555-puppetserver-service/reboot.rb
@@ -1,0 +1,63 @@
+if (options[:type] == 'pe') then
+  pssn='pe-puppetserver'
+else
+  pssn='puppetserver'
+end
+
+test_name "Validates the ability of the #{pssn} service to reliably start after a reboot."
+
+max=10
+sleeptime=40
+retryCounter=0
+step "\n-----------\nValidate #{pssn} is up and running, then reboot, repeat #{max} times\n-----------"
+
+for i in 1..max 
+  step "\n-----\nLoop #{i}\n-----"
+   
+  step "check uptime"
+  origUptime=on(master, "awk '{print $1}' /proc/uptime").stdout.chomp.to_i
+
+  step "reboot"
+  #rebooting with on(master,'reboot') causes an IOError: stream end.  This seems like a sane workaround.
+  #QENG-1914 for more...
+  system "ssh -i #{options[:keyfile]} -o StrictHostKeyChecking=no root@#{master.reachable_name} \'reboot\'"
+
+  step "sleeping for 15 seconds and waiting for reboot."
+  sleep 15
+
+  unless port_open_within?(master,22,120)
+    raise Beaker::DSL::FailTest, 'Could not SSH into puppet master within 2 minutes after reboot'
+  end
+
+  step "look at uptime"
+  secondUptime=on(master, "awk '{print $1}' /proc/uptime").stdout.chomp.to_i
+
+  step "Checking for uptime sanity"
+  assert_operator(origUptime, :>, secondUptime, "Expect original uptime #{secondUptime} to be greater than original uptime #{origUptime} ")
+  
+  step "sleeping for #{sleeptime} seconds and waiting for #{pssn} to start."
+  sleep sleeptime
+
+  step "Check to see that #{pssn} is up and running."
+  ec = on(master,"service #{pssn} status", :acceptable_exit_codes => 0..99).exit_code
+
+  retryCounter = 0
+  while (ec != 0 and retryCounter < 20) do
+    sleep 1
+    retryCounter += 1
+    ec = on(master,"service #{pssn} status", :acceptable_exit_codes => 0..99).exit_code
+  end
+
+  step "Post loop check"
+  on(master,"service #{pssn} status")
+
+  step "Modifing sleeptime"
+  puts "  sleeptime is currently #{sleeptime}  retryCounter is currently #{retryCounter}"
+  sleeptime = sleeptime + retryCounter
+  puts "  Now sleeptime is #{sleeptime}"
+
+end
+
+puts "We should modify sleeptime to #{sleeptime}"
+
+
--- a/acceptance/suites/tests/555-puppetserver-service/reboot.rb
+++ b/acceptance/suites/tests/555-puppetserver-service/reboot.rb
@@ -1,15 +1,18 @@
 if (options[:type] == 'pe') then
-  pssn='pe-puppetserver'
+  puppetserverServiceName='pe-puppetserver'
 else
-  pssn='puppetserver'
+  puppetserverServiceName='puppetserver'
 end
 
-test_name "Validates the ability of the #{pssn} service to reliably start after a reboot."
+test_name "Validates the ability of the #{puppetserverServiceName} service to reliably start after a reboot."
 
 max=3
 sleeptime=45
 retryCounter=0
-step "\n-----------\nValidate #{pssn} is up and running, then reboot, repeat #{max} times\n-----------"
+step "Ensure #{puppetserverServiceName} is enabled."
+on(master, "puppet resource service #{puppetserverServiceName} enable=true")
+
+step "\n-----------\nValidate #{puppetserverServiceName} is up and running, then reboot, repeat #{max} times\n-----------"
 
 for i in 1..max 
   step "\n-----\nLoop #{i}\n-----"
@@ -26,28 +29,28 @@ for i in 1..max
   sleep 15
 
   unless port_open_within?(master,22,120)
-    raise Beaker::DSL::FailTest, 'Could not SSH into puppet master within 2 minutes after reboot'
+    raise Beaker::DSL::FailTest, 'Port 22 did not open on the puppet master host within 2 minutes after reboot'
   end
 
   step "look at uptime to ensure that we actually rebooted"
   secondUptime=on(master, "awk '{print $1}' /proc/uptime").stdout.chomp.to_i
   assert_operator(origUptime, :>, secondUptime, "Expect original uptime #{secondUptime} to be greater than original uptime #{origUptime} ")
   
-  step "sleeping for #{sleeptime} seconds and waiting for #{pssn} to start."
+  step "sleeping for #{sleeptime} seconds and waiting for #{puppetserverServiceName} to start."
   sleep sleeptime
 
-  step "Check to see that #{pssn} is up and running."
-  ec = on(master,"service #{pssn} status", :acceptable_exit_codes => 0..99).exit_code
+  step "Check to see that #{puppetserverServiceName} is up and running."
+  ec = on(master,"service #{puppetserverServiceName} status", :acceptable_exit_codes => 0..99).exit_code
 
   retryCounter = 0
-  while (ec != 0 and retryCounter < 20) do
-    sleep 1
+  while (ec != 0 and retryCounter < 30) do
+    sleep 2
     retryCounter += 1
-    ec = on(master,"service #{pssn} status", :acceptable_exit_codes => 0..99).exit_code
+    ec = on(master,"service #{puppetserverServiceName} status", :acceptable_exit_codes => 0..99).exit_code
   end
 
   step "Post loop check"
-  on(master,"service #{pssn} status")
+  on(master,"service #{puppetserverServiceName} status")
 
 end
 
--- a/acceptance/suites/tests/555-puppetserver-service/reboot.rb
+++ b/acceptance/suites/tests/555-puppetserver-service/reboot.rb
@@ -8,7 +8,7 @@ test_name "Validates the ability of the #{puppetserverServiceName} service to re
 
 max=3
 sleeptime=45
-retryCounter=0
+
 step "Ensure #{puppetserverServiceName} is enabled."
 on(master, "puppet resource service #{puppetserverServiceName} enable=true")
 
@@ -44,7 +44,7 @@ for i in 1..max
 
   retryCounter = 0
   while (ec != 0 and retryCounter < 30) do
-    sleep 2
+    sleep 1
     retryCounter += 1
     ec = on(master,"service #{puppetserverServiceName} status", :acceptable_exit_codes => 0..99).exit_code
   end
--- a/acceptance/suites/tests/555-puppetserver-service/reboot.rb
+++ b/acceptance/suites/tests/555-puppetserver-service/reboot.rb
@@ -1,3 +1,12 @@
+
+
+# rebooting with on(master, 'reboot') causes Beaker::..::host.connection et all to throw IOError: closed stream.
+# using ruby's system call and executing an ssh command on the beaker host's shell works for now.
+# when BKR-70 is resolved, method should be set to beaker, or this script should be modified
+# and the system ssh method should be deleted.
+#method=beaker
+rebootMethod='system'
+
 if (options[:type] == 'pe') then
   puppetserverServiceName='pe-puppetserver'
 else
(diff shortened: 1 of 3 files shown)


acceptance/suites/tests/555-puppetserver-service/reboot.rb
```diff
@@ -0,0 +1,63 @@
...
+end
+
+test_name "Validates the ability of the #{pssn} service to reliably start after a reboot."
+
+max=10
+sleeptime=40
+retryCounter=0
+step "\n-----------\nValidate #{pssn} is up and running, then reboot, repeat #{max} times\n-----------"
+
+for i in 1..max
```
dankreek: I'm having a little trouble figured out why this loop happens. Is it to autotune™ the sleep time?

dankreek: er I mean, help us figure out the best sleep time

acceptance/suites/tests/555-puppetserver-service/reboot.rb
```diff
@@ -0,0 +1,63 @@
...
+  on(master,"service #{pssn} status")
+
+  step "Modifing sleeptime"
+  puts "  sleeptime is currently #{sleeptime}  retryCounter is currently #{retryCounter}"
+  sleeptime = sleeptime + retryCounter
+  puts "  Now sleeptime is #{sleeptime}"
+
+end
+
+puts "We should modify sleeptime to #{sleeptime}"
```
dankreek: And now I know what that loop is for... so this is just a utility script to tune the acceptance test?

acceptance/suites/tests/555-puppetserver-service/reboot.rb
erikPrime: Yes, just to help me figure out the best sleep time...

erikPrime: The primary purpose of the loop is to reboot the system under test <max> times, and to validate that the puppetserver service is up and running after the reboot.

dankreek: Disclaimer: I am not a QA 

Ahh ok, that makes sense to me now. But hmm, do you think running this loop 10 times is a bit excessive? It seems like if it comes up once after a reboot then it should be working.

dankreek:
:+1: This looks good to me.

acceptance/suites/tests/555-puppetserver-service/reboot.rb
```diff
@@ -0,0 +1,55 @@
...
+for i in 1..max 
+  step "\n-----\nLoop #{i}\n-----"
+   
+  step "check uptime"
+  origUptime=on(master, "awk '{print $1}' /proc/uptime").stdout.chomp.to_i
+
+  step "reboot"
+  #rebooting with on(master,'reboot') causes an IOError: stream end.  This seems like a sane workaround.
+  #QENG-1914 for more...
+  system "ssh -i #{options[:keyfile]} -o StrictHostKeyChecking=no root@#{master.reachable_name} \'reboot\'"
```
jeffmccune: I'm concerned that using a different SSH client than beaker normally uses will cause the test to depend on the platform specific behavior relative to where it runs.  For example, if I run this beaker test on my laptop I'm subject to the Mac OS X behavior of `ssh`, whereas when it runs in CI it will run based on the GNU Linux behavior of SSH.

Would it be possible to catch the exception thrown when logging in and rebooting the box using beaker's normal transport?

acceptance/suites/tests/555-puppetserver-service/reboot.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  step "sleeping for 15 seconds and waiting for reboot."
+  sleep 15
+
+  unless port_open_within?(master,22,120)
+    raise Beaker::DSL::FailTest, 'Could not SSH into puppet master within 2 minutes after reboot'
```
jeffmccune: This doens't actually try and SSH in, correct?  If it only tests to make sure the port is open then the message should reflect that.

This is important because there exists a window in time where the socket is open and listening, but the SSH server is not up and running yet so attempting to SSH in will fail while testing if the socket is open will succeed.

acceptance/suites/tests/555-puppetserver-service/reboot.rb
erikPrime: I've tried a few things with pry around catching errors thrown when we do
on(master, 'reboot')
with no success.

I've tried the above ssh command manually on mac and linux and not found a problem.

acceptance/suites/tests/555-puppetserver-service/reboot.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  step "look at uptime to ensure that we actually rebooted"
+  secondUptime=on(master, "awk '{print $1}' /proc/uptime").stdout.chomp.to_i
+  assert_operator(origUptime, :>, secondUptime, "Expect original uptime #{secondUptime} to be greater than original uptime #{origUptime} ")
+  
+  step "sleeping for #{sleeptime} seconds and waiting for #{pssn} to start."
+  sleep sleeptime
+
+  step "Check to see that #{pssn} is up and running."
+  ec = on(master,"service #{pssn} status", :acceptable_exit_codes => 0..99).exit_code
```
jeffmccune: I don't think these exit codes are acceptable, the status method is to return `0` if the service is successfully running and non-zero if the service is not running.

Also, is `service <name> status` implemented on all platforms?  I'm concerned there is platform specific behavior in the method of checking the status of the service (e.g. the use of systemctl on systemd systems).

acceptance/test_run_scripts/555-puppetserver-service/reboot-foss-rhel7.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
+SCRIPT_PATH=$(pwd)
+BASENAME_CMD="basename ${SCRIPT_PATH}"
+SCRIPT_BASE_PATH=`eval ${BASENAME_CMD}`
+if [ $SCRIPT_BASE_PATH = "555-puppetserver-service" ]; then
+  cd ../../..
+fi
```
jeffmccune: What's this trying to do?  I gather change directories to somewhere more appropriate, but I can't figure out the reason why.

acceptance/test_run_scripts/555-puppetserver-service/reboot-foss-rhel7.sh
```diff
@@ -0,0 +1,87 @@
...
+
+export PACKAGE_BUILD_VERSION=1.0.4.SNAPSHOT.2015.03.24T0146
```
jeffmccune: Won't this pin the behavior of this script to a specific build version?  I ask because this seems like something intended to be generally applicable to different build versions.

acceptance/suites/tests/555-puppetserver-service/reboot.rb
```diff
@@ -0,0 +1,55 @@
+if (options[:type] == 'pe') then
+  pssn='pe-puppetserver'
```
jeffmccune: I'm not a huge fan of this variable name, but please don't change it unless you're already back in the file making other changes.

My reasoning is that `pssn` doesn't mean anything to me without having to go back through and fully understand the context around where the variable was bound.  Only then does pssn make sense as the puppetserver service name.

A better name would avoid making the reader figure out the acronym, so maybe something like `service_name` as a suggestion.

acceptance/suites/tests/555-puppetserver-service/reboot.rb
```diff
@@ -0,0 +1,55 @@
...
+else
+  pssn='puppetserver'
+end
+
+test_name "Validates the ability of the #{pssn} service to reliably start after a reboot."
+
+max=3
+sleeptime=45
+retryCounter=0
```
jeffmccune: What's the expected impact on the overall runtime of our tests with this change?  I'm wondering what the best / expected / worst case situation is.  Best case seems like this test will take less than 60 seconds to complete (guessing), average looks to be about the same, but in the worst case I'm having a hard time understanding the impact on total test time.

Looks to be about 5 minutes, which is pushing the limits of what I'm comfortable with, but I'm not sure there's anything to be done.

acceptance/test_run_scripts/555-puppetserver-service/reboot-foss-rhel7.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
```
jeffmccune: Is this script used by the tests, or is more of an example of how to run the test?

If it's more of an example, probably best to omit it from the change set, or include it as a documentation example somewhere else, ideally the same place we have the use of beaker documented as it relates to puppetserver.  This may only exist in Confluence AFAIK.

acceptance/suites/tests/555-puppetserver-service/reboot.rb
erikPrime: As we gain more confidence that we aren't seeing problems with our service running after reboot, we can turn max (max reboots) down from 3 to 1.  I started at 25 and have been turning it down as I run it more times and don't see problems.

jeffmccune: Sure, but until then, do we have a sense of the overall wall-clock impact in the best / average (expected) / worst case?  If so, what are those cases?

Might be worth a comment in the code to this effect.  "This test isn't expected to take longer than X, but if it does the plan is to ..."

acceptance/suites/tests/555-puppetserver-service/reboot.rb
jeffmccune: I hear you, but I do think it's important to avoid introducing another transport mechanism, so I'd like to try and exhaust all of our options with the existing Net::SSH transport.

Have you tried forking the reboot command off as a background job, potentially delayed a moment or two, then letting Net::SSH terminate normally?

I'm thinking of something along the lines of SSHing in and using some combination of a background job and `nohup` to spin the shutdown / reboot command off into the background, then cleanly terminate the SSH connection before the reboot process does it for us.

jeffmccune: FWIW, this type of problem looks really difficult to solve with a single command line because shell escaping is getting a bit insane, but this approach seems like it could work.  It's basically a matter of copying a script to the SUT, then kicking off that script as a background process.  The script will delay a moment or two, giving beaker time to gracefully log out, then the script will bounce the box.

Here's the script I played around with:

```sh
! /bin/bash
#
# Delayed reboot that can be triggered from an SSH connection without
# unmercifully killing that very connection.

echo "Kicked off at `date`" >> /tmp/reboot.log
sleep 2
echo "Rebooted at `date`" >> /tmp/reboot.log
# shutdown -r now
```

In the real world the echo statements could be removed, so the script does very little.  It does give us a convenient thing to pass into a bash / nohup / redirection / background job mess of a command:

```
ssh jeff@localhost bash -c "'nohup /tmp/delayed_reboot > /dev/null 2>&1 &'"
```

Note, the single quotes wrapped in double quotes are important so bash, on the other end of the wire, gets the full command expression as a single argument.  This may not be an issue when calling from beaker which isn't subject to shell shell expansion.

jeffmccune:
@erikPrime I like the example script to run beaker, I've got a bunch of these as well.  With that said, helper scripts have usually lived in `ext/` for the projects I've worked on at Puppet Labs.  Would it be possible to move the help script out of `acceptance/` and into `ext/beaker/` or something similar?

The idea is to put it in a place where it has a better chance of getting discovered and used, which should also help it stay fresh and up to date.

acceptance/test_run_scripts/555-puppetserver-service/reboot-foss-rhel7.sh
jeffmccune: Nevermind this comment, example scripts are fine to pin in this way.

acceptance/suites/tests/555-puppetserver-service/reboot.rb
jeffmccune: The `0.99` makes a lot more sense now that I understand the script is trying to first detect that the service has started, then check to see later that it is actually 0.

This seems like a perfect place to try and figure out a solution to the 3-state problem of first waiting until the service is ready to start, then try and start the service, then wait to see if the service successfully started.

Would it be possible to check something else, e.g. the log file for a "starting up!" message for the first transition, then block again on the "ready to service requests!" event before the final service status check?

We've punted on this issue previously by employing polling loops against the TCP port, but this is definitely another situation where it'd be great to try and clearly define the three states of not-yet-started, started-but-not-yet-ready, and fully-operational-deathstar, with affordances for blocking until a transition to the next state transition happens.

acceptance/suites/tests/555-puppetserver-service/reboot.rb
erikPrime: One of the reasons why "max" started out set so high in this script is because I spent a lot of time testing methods similar to the above: nohup, background, sleep 2 && reboot, init 6, cron.  I didn't keep specific data, but I've previously rejected these methods as being intermittent, even before we try to move to multiple OSes.

The system -> ssh command works on my mac, and is already working in CI, at the end of the PE pipeline in qatests: http://jenkins-enterprise.delivery.puppetlabs.net/view/qatests/view/all/job/enterprise_qatests-puppet_integration-system_pe_021_puppet_service_3.8.x/LAYOUT=64mcd,PLATFORM=redhat6,UPGRADE_FROM=NONE,label=beaker/2/console

I would prefer to use the same patterns we always use with beaker, such as on(master, "somecommand"), but I spent a large amount of time trying to make that work in this case already.

Beyond reasons of good standard practice and concerns about the ssh command working in MacOS and in CI, is there a compelling reason why we should continue to struggle with the typical beaker method in this case that I'm missing?  Is there a way we can configure Ruby's Net::SSH so that it will gracefully accept that SSHD is shutting down?

jeffmccune: > Beyond reasons of good standard practice and concerns about the ssh command working in MacOS and in CI, is there a compelling reason why we should continue to struggle with the typical beaker method in this case that I'm missing?

Yes, my primary concern is trying to make this robust for the long-term.  For example, even though it works today in all of the places we typically run beaker, we're introducing a lot of variation into the system that needs to be maintained over a long time horizon.  The variation I see is this.

Today, we have one transport method, Net::SSH and specifically the beaker methods that wrap around Net::SSH.  If, in the future, we find a reason to change the behavior of this transport then we have only one place to fix it and then validate the change across all of the places we run beaker.

If we introduce another transport mechanism that calls directly down to the system and there's a change we need to make in the future, we now have many more things we need to make behave.  We'll have to update beaker's wrappers around Net::SSH and in addition make sure these direct calls to the system ssh command work on all systems we run beaker from.

It's not so much about the risk of this breaking in the future, it's about the increase in effort it will take to change the behavior in the future.  To this end, the fact that it works today doesn't really address my concern, because the concern is more, "how much effort will it take to maintain this behavior over the long haul?"

> Is there a way we can configure Ruby's Net::SSH so that it will gracefully accept that SSHD is shutting down?

I think there's a way to gracefully handle that SSHD terminates, but I don't think it involves reconfiguring Net::SSH.

I'm going to try and provide an example of catching the exception that's thrown, which is probably the most direct and un-surprising way to handle this sort of situation.

I'll follow up here with what I find out.

cprice404: The Knights of Standards and Practices

jeffmccune: For posterity here's an example of what we came up with together in chat:

```ruby
begin
  on(master, cmd)
rescue IOError => detail
  # re-raise if we caught an unexpected exception
  raise if not /closed stream/.match(detail.message)
end
```

jeffmccune:
I'd like to get the changes to beaker in [PR 776](https://github.com/puppetlabs/beaker/pull/776) merged in and then update this PR to change the pin on beaker to a version that includes that change.

jeffmccune:
@erikPrime What's the status of this PR?  Is it something we should keep open or close for the time being?  I know we discussed this briefly IRL but I'm just reviewing the PR queue and trying to capture the current state of the world in the PR itself.

erikPrime:
Status:  Once I have time to make sure the reboot script in the qatests repository is working with the new version of beaker, that script will need to be copied over the top of this PR.  I was going to close and delete this branch, but there are setting in the shell script I don't want to loose.

cprice404:
Can we close the PR for now then?  That won't result in the branch being deleted, and we can just open a new PR when the changes are ready?

PR #468, opened 2015-03-25
DOCS-465 release notes for Puppet Server 1.0.8

jbondpdx:
1.0.8 release notes and minor tuning guide edits.

### DIFF
--- a/documentation/release_notes.markdown
+++ b/documentation/release_notes.markdown
@@ -10,9 +10,9 @@ In addition to several bug fixes, this release adds new HTTP client timeout sett
 
 ###New Features
 
-#### Added new http-client timeout settings
+####Added new http-client timeout settings
 
-We've exposed two new HTTP client timeout settings, for idle and connection timeouts.
+We've exposed two new HTTP client timeout settings: `idle-timeout-milliseconds` and `connect-timeout-milliseconds`. These new settings can be configured in the http-client section of the [puppetserver.conf file](./configuration.markdown#puppetserverconf).
 
 * [SERVER-449](https://tickets.puppetlabs.com/browse/SERVER-449) - Expose http-client timeouts from Puppet Server http_connect_timeout and http_read_timeout.
 
@@ -21,6 +21,12 @@ We've exposed two new HTTP client timeout settings, for idle and connection time
 This version of Puppet Server has a special-purpose logfile to capture only the HTTP traffic. This should work out of the box, but you can [configure the location and the format](./configuration.markdown#http-traffic) of the logfile.
 
 * [SERVER-319](https://tickets.puppetlabs.com/browse/SERVER-319)
+
+####Added new JRuby default borrow timeout setting
+
+Previously, the JRuby pool borrow timeout was indefinite and wasn't configurable. As of SERVER 1.0.8, there is a new `borrow-timeout` setting in the http-client section of the [puppetserver.conf file](./configuration.markdown#puppetserverconf). If you don't specify a value for that setting, Puppet Server will use 20 minutes as a default. This allows enough time for realistic expensive catalog compilations while avoiding indefinite hanging.
+
+* [SERVER-408](https://tickets.puppetlabs.com/browse/SERVER-408) - Expose configurable `borrow-timeout` to allow JRuby pool borrows to timeout
 
 #### Added Puppet Server JRuby tuning guide
 
@@ -36,11 +42,6 @@ Previously, Puppet Server failed to start after a reboot on some systems (notabl
 
 * [SERVER-404](https://tickets.puppetlabs.com/browse/SERVER-404) - Properly create /var/run/puppetserver dir in FOSS packaging
 
-####JRuby default borrow timeouts are significantly extended
-
-The default JRuby pool borrow timeout was extended from 60 seconds to 20 minutes. This allows enough time for realistic expensive catalog compilations while avoiding indefinite hanging.
-
-* [SERVER-408](https://tickets.puppetlabs.com/browse/SERVER-408) - backport jruby borrow timeout changes to stable branch
 
 #### Startup scripts now use 'runuser'.
 
@@ -56,7 +57,7 @@ Running the `puppetserver foreground` subcommand produced no output. It should n
 
 #### CA handling fixed
 
-We've fixed an issue where Puppet Server was mishandling some CAs. Specifically, if you brought up a Puppet CA on a master where you wanted to use an external Puppet CA, the local CA superseded the certificate from the external CA. This issue has been fixed.
+Previously, Puppet Server was mishandling some CAs. Specifically, if you brought up a Puppet CA on a master where you wanted to use an external Puppet CA, but you hadn't already configured the disabled CA service in the `bootstrap.cfg` file, the local CA superseded the certificate from the external CA. This issue has now been fixed.
 
 * [SERVER-345](https://tickets.puppetlabs.com/browse/SERVER-345) - Fixup usages of cacert / localcacert in master
 
@@ -69,7 +70,9 @@ The default maximum number of JRuby instances has been capped at 4. This is a sa
 
 ## Puppet Server 1.0.3 -- 1.0.7
 
-Puppet Server 1.0.6 is a Puppet Enterprise-only release, which is included in PE 3.7.2. The only change from Puppet Server 1.0.2 is that [SERVER-262](https://tickets.puppetlabs.com/browse/SERVER-262) was reverted; this change is also included in Puppet Server 1.0.8. Other version numbers between 1.0.3 and 1.0.7 were never released.
+Puppet Server versions 1.0.3 -- 1.0.7 were never released. 
+
+However, Puppet Enterprise 3.7.2 included a version of Puppet Server that was labeled as version 1.0.6. The only change from Puppet Server 1.0.2 was that the fix for [SERVER-262](https://tickets.puppetlabs.com/browse/SERVER-262) was reverted in [SERVER-522](https://tickets.puppetlabs.com/browse/SERVER-522). This change is also included in the release of Puppet Server 1.0.8.
 
 ## Puppet Server 1.0.2
 
--- a/documentation/release_notes.markdown
+++ b/documentation/release_notes.markdown
@@ -4,6 +4,71 @@ title: "Puppet Server: Release Notes"
 canonical: "/puppetserver/latest/release_notes.html"
 ---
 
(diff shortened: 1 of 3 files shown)


documentation/release_notes.markdown
```diff
@@ -4,6 +4,73 @@ title: "Puppet Server: Release Notes"
...
+
+In addition to several bug fixes, this release adds new HTTP client timeout settings, a special logfile to capture only HTTP traffic, and a JRuby tuning guide to help you get the best performance from Puppet Server.
+
+###New Features
+
+#### Added new http-client timeout settings
+
+We've exposed two new HTTP client timeout settings, for idle and connection timeouts.
+
+* [SERVER-449](https://tickets.puppetlabs.com/browse/SERVER-449) - Expose http-client timeouts from Puppet Server http_connect_timeout and http_read_timeout.
```
camlow325: Not sure if this is critical to cover here, but the names of the configuration settings are `idle-timeout-milliseconds` and `connect-timeout-milliseconds` and can be configured under the `http-client` section of the puppetserver.conf file.  See https://github.com/puppetlabs/puppet-server/blob/86e2b6b79a9ad88c796579945d3d802cdc250359/documentation/configuration.markdown#puppetserverconf.

documentation/release_notes.markdown
```diff
@@ -4,6 +4,73 @@ title: "Puppet Server: Release Notes"
...
+
+####Fixed an issue where Puppet Server couldn't start after reboot
+
+Previously, Puppet Server failed to start after a reboot on some systems (notably RHEL 7 and Ubuntu 14.4). This was because the `/var/run/` directory, needed by Puppet Server, was being destroyed on reboot. This issue has been fixed.
+
+* [SERVER-404](https://tickets.puppetlabs.com/browse/SERVER-404) - Properly create /var/run/puppetserver dir in FOSS packaging
+
+####JRuby default borrow timeouts are significantly extended
+
+The default JRuby pool borrow timeout was extended from 60 seconds to 20 minutes. This allows enough time for realistic expensive catalog compilations while avoiding indefinite hanging.
```
camlow325: This wasn't really clear from the ticket description but actually in SERVER 1.0.2 and earlier, the JRuby pool borrow timeout was infinite/indefinite and was not user-configurable.  As of SERVER 1.0.8, however, there is a new setting in the puppetserver.conf file, under the `http-client` section, called `borrow-timeout`.  If the user hasn't specified a value for that setting, Puppet Server will use 20 minutes as a default.

We had actually done the first pass at this work on the Puppet Server master branch for early 2.0 development and committed a default of 60 seconds.  We later bumped that up to 20 minutes, though - before the first 2.0 RC was released.  We haven't done any releases of Puppet Server where there was a 60 second default, so probably not worth mentioning that here.

I updated the description in SERVER-408 to hopefully make this a little more clear.

documentation/release_notes.markdown
```diff
@@ -4,6 +4,73 @@ title: "Puppet Server: Release Notes"
...
+
+####`puppetserver foreground` now produces output
+
+Running the `puppetserver foreground` subcommand produced no output. It should now provide its usual output again.
+
+* [SERVER-356](https://tickets.puppetlabs.com/browse/SERVER-356) - puppetserver foreground produces no output
+
+#### CA handling fixed
+
+We've fixed an issue where Puppet Server was mishandling some CAs. Specifically, if you brought up a Puppet CA on a master where you wanted to use an external Puppet CA, the local CA superseded the certificate from the external CA. This issue has been fixed.
```
camlow325: For a little clarification, the local external CA certificate would have only been superseded by a Puppet CA certificate if the user had not already configured the disabled CA service in the bootstrap.cfg file.  Not sure if it's critical to add that qualification here - will defer to your judgement.

documentation/release_notes.markdown
```diff
@@ -4,6 +4,73 @@ title: "Puppet Server: Release Notes"
...
+#### Default maximum JRuby instances capped at 4
+
+The default maximum number of JRuby instances has been capped at 4. This is a safer maximum for use with the default 2GB JVM memory.
+
+* [SERVER-448](https://tickets.puppetlabs.com/browse/SERVER-448) - Change default max-active-instances to not exceed 4 JRubies
+
+
+## Puppet Server 1.0.3 -- 1.0.7
+
+Puppet Server 1.0.6 is a Puppet Enterprise-only release, which is included in PE 3.7.2. The only change from Puppet Server 1.0.2 is that [SERVER-262](https://tickets.puppetlabs.com/browse/SERVER-262) was reverted; this change is also included in Puppet Server 1.0.8. Other version numbers between 1.0.3 and 1.0.7 were never released.
```
camlow325: I'm not sure what the best way to describe this to users but here's an attempt...

PE 3.7.2 included a Puppet Enterprise version of Puppet Server which was labeled as version 1.0.6.  PE Puppet Server 1.0.6 included Open-Source Puppet Server version 1.0.3.  A label was applied to the puppetserver github project for Open-Source Puppet Server version 1.0.3 but Open-Source Puppet Server version 1.0.3 was never released outside of Puppet Enterprise in package form.  The only difference between Open-Source Puppet Server versions 1.0.2 and 1.0.3 is that the fix for SERVER-262 was reverted for the 1.0.3 release, which SERVER-522 now covers.

camlow325:
I think it would be good in the release notes to mention that we upgraded Puppet Server's dependencies on JRuby to 1.7.19 (SERVER-273) and on the Jetty web server to 9.2.10 (SERVER-380 and SERVER-513).

There aren't any major functional changes that we're anticipating as part of those updates.  A few maybe worth mentioning, however, are:

* JRuby 1.7.19 resolves a memory leak around the use of the "jruby-pool" DELETE admin API.

* Between the 9.1.0 version that Puppet Server 1.0.2 was using and the 9.2.10 version that Puppet Server 1.0.8 will use, the Jetty team has implemented a number of fixes related to problems with intermittent response handling failures and associated threading deadlocks.

* The upgrade to the newer Jetty version also coincides with changes to a couple of the default values that Puppet Server uses for the webserver.  Via a change to the `shutdown-timeout-seconds` setting, the webserver can now create up to 200 threads -- rather than up to 100 threads -- by default.  Via a change to the `shutdown-timeout-seconds` setting, the webserver will now allow for up to only 30 seconds for a graceful shutdown rather than up to 60 seconds by default.  These are documented in SERVER-513.

documentation/tuning_guide.markdown
```diff
@@ -47,113 +47,103 @@ setting.  The value of this setting is used by Puppet Server to determine how
...
-`max-active-instances` to a value higher than the number of CPUs.)
-
-At this point you may be wondering "what's the downside to just setting
-`max-active-instances` to a really high value?"  The answer to this question, in
-a nutshell, is "memory usage".  This brings us to the other extremely important
+then it might make sense to have more JRuby instances than the number of cores. The JVM is smart enough to suspend the thread that is handling those kinds of requests and use the CPUs for other work, assuming there are still JRuby instances available in the pool. In a case like this you might want to set `max-active-instances` to a value higher than the number of CPUs.)
+
+At this point you may be wondering, "What's the downside to just setting
+`max-active-instances` to a really high value?" The answer to this question, in
+a nutshell, is "memory usage". This brings us to the other extremely important
```
camlow325: The last sentence in the paragraph reads a little strange to me here.  Maybe should say something like, "This brings us to the other extremely important *setting to consider* for Puppet Server?"

camlow325:
Looks good to me, with just a few comments / clarifications on some of the items in the notes that you may want to consider adding / revising.

documentation/release_notes.markdown
camlow325: Thinking about this a little more, I wonder if it might be better to more clearly call out that PE Puppet Server and OSS Puppet Server are two different products - with an ongoing lifetime for each.  For example:

The Puppet Server in Puppet Enterprise contains an Open-Source Puppet Server plus some Enterprise-only functionality, e.g., advanced server metrics.  PE Puppet Server version numbers may not always align directly with the versions of the corresponding Open-Source Puppet Server packages that they use.  Below is a table which shows the mapping among PE version, PE Puppet Server version, and Open-Source Puppet Server version.  Note that for the Open-Source Puppet Server versions marked by an asterisk (*), the packages were included in PE but not released in standalone form:

PE Version | PE Puppet Server  | OSS Puppet Server
--------------- | --------------------------| --------------------------
3.7.0           | 0.3.2                       | 0.3.0
3.7.1           | 0.4.1                       | 0.4.1 (*)
3.7.2           | 1.0.6                       | 1.0.3 (*)

documentation/release_notes.markdown
jbondpdx: Awesome, thanks. I'm revising this and moving it to new features, then.

documentation/release_notes.markdown
jbondpdx: I think I won't add the gory details (which they can find in the ticket if they really want them), but I will change it to indicate that it didn't ALWAYS do it.

documentation/release_notes.markdown
camlow325: @jbondpdx Not as of PE 3.8, no.  @cprice404, @nfagerlund, and I chatted about the possibility of enhancing the "puppetserver --version" CLI command in PE to output both the PE and OSS Puppet Server versions.  I created a ticket, PE-8821, to cover that work.  That won't land until a later PE release, though.  Until then, however, the only way I think a user would be able to see that directly would be in any documentation we'd create.

Longer-term, I would think we should be more upfront with users about the distinction between the PE and OSS Puppet Server versions so that users would have that context when browsing our JIRA tickets and documentation.  Otherwise, the distinction between the versions of each could continue to be a source of confusion.  But I'd have to defer to you and @nfagerlund on the best approach for doing that.

documentation/release_notes.markdown
camlow325: Cool, sounds good.

camlow325:
@jbondpdx Latest edits look good to me.  Were you planning on making any further changes around how the PE Puppet Server vs. Puppet Server versions are described or any of the other additional items I mentioned that could be documented - new JRuby and Jetty version changes?  If not, I could go ahead and merge this...

camlow325:
I think this ticket was done as part of https://tickets.puppetlabs.com/browse/SERVER-465 not DOCS-465, since the latter doesn't seem to be a real JIRA ticket.  SERVER-465 is marked "done" so presuming that @jbondpdx was done with her edits on this and will merge.

PR #469, opened 2015-03-25
(BKR-147) add Gemfile setting for BEAKER_VERSION for puppet...

anodelman:
puppetdb, etc

- support for BEAKER_VERSION and BEAKER_RSPEC_VERSION in gemfile

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ gem 'jira-ruby', :group => :development
 
 group :test do
   gem 'rspec'
-  gem 'beaker', *location_for(ENV['BEAKER_LOCATION'] || '~> 2.9.0')
+  gem 'beaker', *location_for(ENV['BEAKER_VERSION'] || '~> 2.9.0')
   if ENV['GEM_SOURCE'] =~ /rubygems\.delivery\.puppetlabs\.net/
     gem 'sqa-utils', '~> 0.11'
   end


Gemfile
```diff
@@ -13,11 +13,16 @@ end
 gem 'rake', :group => [:development, :test]
 gem 'jira-ruby', :group => :development
 
+beaker_version = ENV['BEAKER_LOCATION'] || ENV['BEAKER_VERSION']
```
jeffmccune: We recently added `BEAKER_LOCATION`.  Why is `BEAKER_VERSION` preferred?  Is it a convention in Jenkins Job parameters or something?

There's no real reason to use `BEAKER_LOCATON` as far as I know, the `_LOCATION` simply comes from the existing convention in the puppet Gemfile.

Ideally there would be only one environment variable, not two.  Which one should it be?

Gemfile
```diff
@@ -13,11 +13,16 @@ end
...
 group :test do
   gem 'rspec'
   # gem 'beaker', *location_for(ENV['BEAKER_LOCATION'] || '~> 2.2')
   # See: SERVER-435 for the reasoning behind this specific version
-  gem 'beaker', *location_for(ENV['BEAKER_LOCATION'] || 'git://github.com/puppetlabs/beaker#7d0cccf22cc7289113de862fe4a4494872f574bb')
+  if beaker_version
+    gem 'beaker', *location_for(beaker_version)
+  else
+    gem 'beaker', 'git://github.com/puppetlabs/beaker#7d0cccf22cc7289113de862fe4a4494872f574bb')
+  end
```
jeffmccune: Could this be reduced down to `gem 'beaker', *location_for(ENV['BEAKER_VERSION'] || ...` ?

Gemfile
anodelman: We are standardizing on BEAKER_VERSION and BEAKER_RSPEC_VERSION - it's already integrated into a number of other projects.

Gemfile
anodelman: Just set up this way for readability.

jeffmccune:
@anodelman Thanks for updating this PR.

We've since changed the pinned version of beaker so this needs a rebase.

In the process, could we just kill `BEAKER_LOCATION` entirely and simply this down to:

```ruby
beaker_version = ENV['BEAKER_VERSION'] || 'git@github.com:puppetlabs/beaker.git#8909e35a97b8c30b11b77b784a47c8d175cd4ec8'
gem 'beaker', *location_for(beaker_version)
```

anodelman:
Updated patch to support the standardized BEAKER_VERSION, instead of BEAKER_LOCATION.

camlow325:
Apologize I haven't kept up with the discussion on this one, but, @anodelman, do you believe this is still needed?

As github doesn't think we can automatically merge this, I'd also guess that there's a merge conflict between this PR and what's on our master branch right now...

anodelman:
Rebased.  Still desired to bring naming in line with the rest of beaker dependent projects.

camlow325:
Thanks, @anodelman.

I'm ignorantly :+1: on this.  @jeffmccune - any concerns?

jeffmccune:
:+1:

PR #470, opened 2015-03-25
(maint) Update to ezbake 0.3.1

jeffmccune:
Without this patch puppetserver depends on ezbake 0.3.0 which is a
problem because that version of ezbake does not build working packages
for debian flavor platforms.  This patch addresses the problem by
updating to ezbake 0.3.1 which includes support for AIO with debian7,
ubuntu1404, and ubuntu1204


nwolfe:
:+1:

PR #471, opened 2015-03-26
(SERVER-518) Update group-id and artifact-id for version checks

camlow325:
This commit removes the hyphen from "puppet-server" in the project.clj
defproject.  This is needed in order to preserve backward compatibility
with dujour checkins of Puppet Server, which used "puppetserver" as the
artifact-id.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -10,7 +10,7 @@
     :password :env/nexus_jenkins_password
     :sign-releases false })
 
-(defproject puppetlabs/puppet-server ps-version
+(defproject puppetlabs/puppetserver ps-version
   :description "Puppet Server"
 
   :dependencies [[org.clojure/clojure "1.6.0"]
--- a/project.clj
+++ b/project.clj
@@ -83,7 +83,7 @@
                    ; SERVER-332, enable SSLv3 for unit tests that exercise SSLv3
                    :jvm-opts      ["-Djava.security.properties=./dev-resources/java.security"]}
 
-             :ezbake {:dependencies ^:replace [[puppetlabs/puppet-server ~ps-version]
+             :ezbake {:dependencies ^:replace [[puppetlabs/puppetserver ~ps-version]
                                                [puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty-version]
                                                [org.clojure/tools.nrepl "0.2.3"]]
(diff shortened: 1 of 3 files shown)


cprice404:
+1

camlow325:
There ended up needing to be more done to fixup the version check than what I had in my initial PR.

In addition to the artifact-id changing as a result of the lein ezbake consolidation, the group-id also changed.  It had previously been "puppetlabs.packages" but would now be just "puppetlabs", since the latter is what is in the Puppet Server project.clj file.  In 2879a59, I updated the location where this name is hardcoded in Puppet Server code and verified that with the change that a version check was working both when running Puppet Server from source and when running Puppet Server as a service from the jar file.

@cprice404 told me that he believes changing the group-id from "puppetlabs.packages" to "puppetlabs" is a reasonable thing to do and shouldn't create any downstream problems, so I went with his recommendation.

project.clj
```diff
@@ -10,7 +10,7 @@
     :password :env/nexus_jenkins_password
     :sign-releases false })
 
-(defproject puppetlabs/puppet-server ps-version
+(defproject puppetlabs/puppetserver ps-version
```
nwolfe: Need to change the `:ezbake` profile dep too

src/ruby/puppet-server-lib/puppet/server/socket_error.rb
```diff
@@ -0,0 +1,4 @@
+require 'puppet/server'
+
+class Puppet::Server::SocketError < SocketError
+end
```
nwolfe: What's this?

camlow325:
Per PR feedback, changed the ezbake profile reference from "puppet-server" to "puppetserver".  That's in commit d76f685.  Confirmed that with a clean local .m2/repository directory that I could do an ezbake build with the change and install the resulting package on a CentOS 7 VM.  Confirmed that the version check with that installed package worked properly at service startup and that the service successfully started.

src/ruby/puppet-server-lib/puppet/server/socket_error.rb
camlow325: That was me being stupid.  :hankey:  Purged this from the last commit.

jeffmccune:
:+1:

nwolfe:
:+1:

PR #473, opened 2015-03-26
(SERVER-499) Fix inverted confinement for 10_install_release_repos

jeffmccune:
Without this patch the confinement added in 2f8d33c is a problem because it
changed behavior to install the release repository only on windows, but the
desired behavior is to avoid this step on windows.

This patch addresses the problem by inverting the confinement.


jeffmccune:
Merging this in without review since this implements the behavior everyone who already reviewed #459 thought they were getting, but didn't because I tripped up over the words `not`, `confine`, and `except`.

Clearly the original code should have used `unless`.  That would have cleared things up.

PR #474, opened 2015-03-26
(SERVER-520) Remove hyphen from puppet-server in project.clj

camlow325:
This commit removes the hyphen from "puppet-server" in the project.clj
defproject. This is needed in order to preserve backward compatibility
with dujour checkins of Puppet Server, which used "puppetserver" as the
artifact-id.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -10,7 +10,7 @@
     :password :env/nexus_jenkins_password
     :sign-releases false })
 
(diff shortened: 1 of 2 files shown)


dankreek:
:+1: but *sniff*, I'm going to miss the hyphen.

jeffmccune:
:+1: so long hyphy

dankreek:
The Hyphy Movement

PR #475, opened 2015-03-26
(SERVER-517) Raise Ruby errors from HTTP client

nwolfe:
This commit changes the HTTP client so it catches underlying Java
exceptions and raises them as a Ruby Puppet::Server::HttpClientError
instead.

This change allows consumers of the HTTP client to catch errors using
only MRI Ruby and not require the use of JRuby.

### DIFF
--- a/spec/puppet-server-lib/puppet/jvm/http_client_spec.rb
+++ b/spec/puppet-server-lib/puppet/jvm/http_client_spec.rb
@@ -21,4 +21,19 @@ describe Puppet::Server::HttpClient do
       expect(request_options.get_connect_timeout_milliseconds).to equal(connect_timeout)
     end
   end
+
+  context "requests" do
+    it "throw Ruby SocketErrors instead of Java exceptions" do
+      client = Puppet::Server::HttpClient.new("localhost", 0, {})
+      [lambda { client.get('/', nil) },
+       lambda { client.post('/', nil, nil) }
+      ].each do |request_block|
+        expect(&request_block).to raise_error { |error|
+          expect(error).to be_a(SocketError)
+          expect(error).to be_a(Puppet::Server::HttpClientError)
+          expect(error.message).to eq('com.puppetlabs.http.client.HttpClientException: Error executing http request')
+        }
+      end
+    end
+  end
 end
--- a/spec/puppet-server-lib/puppet/jvm/http_client_spec.rb
+++ b/spec/puppet-server-lib/puppet/jvm/http_client_spec.rb
@@ -4,35 +4,59 @@ require 'puppet/server/http_client'
 require 'java'
 java_import com.puppetlabs.http.client.SimpleRequestOptions
 
-describe Puppet::Server::HttpClient do
-  context "options" do
-    let(:connect_timeout) { 42 }
-    let(:idle_timeout) { 24 }
+describe 'Puppet::Server::HttpClient' do
+  let :client do
+    Puppet::Server::HttpClient.new("localhost", 0, {})
+  end
 
-    it "timeout settings are properly set" do
+  context "when settings are initialized with specific values" do
+    before :all do
+      @settings = Puppet::Server::HttpClient.settings
       Puppet::Server::HttpClient.initialize_settings(
-          {"http_connect_timeout_milliseconds" => connect_timeout,
-           "http_idle_timeout_milliseconds"  => idle_timeout})
+        {"http_connect_timeout_milliseconds" => 42,
+         "http_idle_timeout_milliseconds"  => 24})
+    end
 
-      client = Puppet::Server::HttpClient.new(nil, 0, {})
+    after :all do
+      Puppet::Server::HttpClient.initialize_settings(@settings)
+    end
+
+    subject do
       request_options = SimpleRequestOptions.new("http://i.love.ruby")
       client.send(:configure_timeouts, request_options)
-      expect(request_options.get_socket_timeout_milliseconds).to equal(idle_timeout)
-      expect(request_options.get_connect_timeout_milliseconds).to equal(connect_timeout)
+      request_options
+    end
+
+    it 'then get_socket_timeout_milliseconds is 24' do
+      expect(subject.get_socket_timeout_milliseconds).to eq(24)
+    end
+
+    it 'then get_connect_timeout_milliseconds is 42' do
+      expect(subject.get_connect_timeout_milliseconds).to eq(42)
     end
   end
 
-  context "requests" do
-    it "throw Ruby SocketErrors instead of Java exceptions" do
-      client = Puppet::Server::HttpClient.new("localhost", 0, {})
-      [lambda { client.get('/', nil) },
-       lambda { client.post('/', nil, nil) }
-      ].each do |request_block|
-        expect(&request_block).to raise_error { |error|
-          expect(error).to be_a(SocketError)
-          expect(error).to be_a(Puppet::Server::HttpClientError)
-          expect(error.message).to eq('com.puppetlabs.http.client.HttpClientException: Error executing http request')
-        }
+  context "when making a request that triggers a Java exception" do
+    let :requests do
+      {
+        get: lambda { client.get('/', nil) },
+        post: lambda { client.post('/', nil, nil) }
+      }
+    end
+
+    [:get, :post].each do |request|
+      describe "#{request} request" do
+        subject { requests[request].call }
+
+        it 'raises a SocketError' do
+          expect { subject }.to raise_error SocketError
+        end
+        it 'raises a Puppet::Server::HttpClientError' do
+          expect { subject }.to raise_error Puppet::Server::HttpClientError
+        end
+        it 'raises an Error with a specific message' do
+          expect { subject }.to raise_error "Error executing http request"
+        end
       end
     end
   end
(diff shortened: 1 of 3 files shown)


src/ruby/puppet-server-lib/puppet/server/http_client.rb
```diff
@@ -50,8 +59,8 @@ def post(url, body, headers, options = {})
       end
 
       # http://en.wikipedia.org/wiki/Basic_access_authentication#Client_side
-      headers["Authorization"] =
-          "Basic #{Base64.strict_encode64 "#{credentials[:user]}:#{credentials[:password]}"}"
+      encoded = Base64.strict_encode64("#{credentials[:user]}:#{credentials[:password]}")
+      headers["Authorization"] = "Basic #{encoded}"
```
nwolfe: This particular nesting of strings and hashes was breaking syntax highlighting. Also pretty difficult to read. I just pulled it out into a local variable.

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -126,17 +126,27 @@
...
-             (catch EvalFailedException e
-               (is (instance? HttpClientException (.. e getCause)))
-               (is (instance? ConnectionClosedException (.. e getCause getCause)))))))))
+      (let [sc (create-scripting-container 10080)]
+        (logutils/with-test-logging
+          (try
+            (.runScriptlet sc (raise-caught-http-error "c.get('/', {})"))
+            (is false "Expected HTTP connection to HTTPS port to fail")
+           (catch EvalFailedException e
+             (is (instance? ConnectionClosedException (.getCause e)))))))))
```
nwolfe: Digging in to the nested exception was easier to do in Ruby from within the scripting container.

The previous tests were validating the root `HttpClientException` in Clojure, but now they're validating the Ruby `HttpClientError` implicitly with the `rescue` block. The underlying Java exception is still validated in Clojure though.

spec/puppet-server-lib/puppet/jvm/http_client_spec.rb
```diff
@@ -21,4 +21,19 @@
...
+    it "throw Ruby SocketErrors instead of Java exceptions" do
+      client = Puppet::Server::HttpClient.new("localhost", 0, {})
+      [lambda { client.get('/', nil) },
+       lambda { client.post('/', nil, nil) }
+      ].each do |request_block|
+        expect(&request_block).to raise_error { |error|
+          expect(error).to be_a(SocketError)
+          expect(error).to be_a(Puppet::Server::HttpClientError)
+          expect(error.message).to eq('com.puppetlabs.http.client.HttpClientException: Error executing http request')
+        }
```
nwolfe: Pardon this test...but it violates The Wolfe Principle by spitting out `Connection refused` exception log messages.
I tried for 5 minutes to figure out a way to silence them, but it's an interop cluster****.

Suggestions welcome.

spec/puppet-server-lib/puppet/jvm/http_client_spec.rb
```diff
@@ -21,4 +21,19 @@
       expect(request_options.get_connect_timeout_milliseconds).to equal(connect_timeout)
     end
   end
+
+  context "requests" do
```
jeffmccune: More conventional to use "describe" here since I get the impression you're describing requests.

I typically use context when the spec is expressed in the "given ... when ... then" format.

They're aliases of each other so this purely semantic.

spec/puppet-server-lib/puppet/jvm/http_client_spec.rb
```diff
@@ -21,4 +21,19 @@
...
+    it "throw Ruby SocketErrors instead of Java exceptions" do
+      client = Puppet::Server::HttpClient.new("localhost", 0, {})
+      [lambda { client.get('/', nil) },
+       lambda { client.post('/', nil, nil) }
+      ].each do |request_block|
+        expect(&request_block).to raise_error { |error|
+          expect(error).to be_a(SocketError)
+          expect(error).to be_a(Puppet::Server::HttpClientError)
+          expect(error.message).to eq('com.puppetlabs.http.client.HttpClientException: Error executing http request')
```
jeffmccune: Pretty unconventional to have so many expectations in a single `it` block.  Is there a way you can see breaking these out into separate it blocks?

spec/puppet-server-lib/puppet/jvm/http_client_spec.rb
```diff
@@ -21,4 +21,19 @@
       expect(request_options.get_connect_timeout_milliseconds).to equal(connect_timeout)
     end
   end
+
+  context "requests" do
+    it "throw Ruby SocketErrors instead of Java exceptions" do
```
camlow325: Maybe a better description would be "raises a Ruby SocketError for a connection failure"?

spec/puppet-server-lib/puppet/jvm/http_client_spec.rb
camlow325: Yeah, I was assuming we might need to do something like `with-log-output-atom` - https://github.com/puppetlabs/trapperkeeper/blob/trapperkeeper-1.1.0/test/puppetlabs/trapperkeeper/testutils/logging.clj#L88-L106 - in trapperkeeper's testutils does to stop current loggers, run the test functionality you don't want to have processed by the console appender, and restart the original loggers when done.  Maybe as a Ruby helper method that these spec tests could call as desired?  Definitely not worth holding this PR up, though.

Someday we'll tell future generations about the day @nwolfe intentionally violated his own principle.  :smile:

src/ruby/puppet-server-lib/puppet/server/http_client.rb
```diff
@@ -70,12 +79,24 @@ def get(url, headers)
...
+    response = client_get(request_options)
     ruby_response(response)
   end
 
   private
 
+  def client_get(request_options)
+    SyncHttpClient.get(request_options)
+  rescue Java::ComPuppetlabsHttpClient::HttpClientException => e
+    raise Puppet::Server::HttpClientError.new(e.inspect, e)
```
camlow325: I wonder if it might be better to just pass up `e.message` instead of `e.inspect` here.  For consumers who would try to read `message` on the exception we're raising here and get a payload that looks like the serialized body you get from `inspect` would feel a little strange to me.

src/ruby/puppet-server-lib/puppet/server/http_client.rb
```diff
@@ -70,12 +79,24 @@ def get(url, headers)
...
+  end
+
+  def client_post(request_options)
+    SyncHttpClient.post(request_options)
+  rescue Java::ComPuppetlabsHttpClient::HttpClientException => e
+    raise Puppet::Server::HttpClientError.new(e.inspect, e)
```
camlow325: Same question here on using `e.inspect` vs. `e.message`.

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
camlow325: I actually kinda like it better this way anyway because it allows us to indirectly exercise the exception handling in Ruby, which is more interesting from a production code perspective than the Clojure code path.

camlow325:
:+1: with just a few cosmetic comments.

For good measure, I did some functional testing of this against a full puppetserver stack with a puppetdb on the back end.  In one case, the puppetdb was not actually listening and so the requests failed, as expected, with reasonable looking connection errors written to the master log.  In another case, the puppetdb was listening and the socket succeeded / request was handled correctly, as expected.  All looks good to me.

I'm supportive of some refactoring to the tests for better rspec purity, along the lines of @jeffmccune's suggestions.  I'd really like to get this PR in today, though, so I'm hoping this can be done quickly.

spec/puppet-server-lib/puppet/jvm/http_client_spec.rb
nwolfe: Ah yeah, I didn't give it any thought but `describe` makes more sense. Will change.

spec/puppet-server-lib/puppet/jvm/http_client_spec.rb
nwolfe: That is better. Will change.

src/ruby/puppet-server-lib/puppet/server/http_client.rb
nwolfe: I went with `inspect` because it's supposed to be the most verbose object->string method, and that's what the classifier code was using to print out this error. Happy to go with something else if we know it's better or more conventional.

spec/puppet-server-lib/puppet/jvm/http_client_spec.rb
nwolfe: Happy to refactor this in any way! @jeff let's pair for 15 minutes

src/ruby/puppet-server-lib/puppet/server/http_client.rb
camlow325: The way I'd been thinking about it is that is that the first argument you give to the `HttpClientError` constructor becomes what is returned back from the `message` method on the object which is created so it might make most sense to just extract the `message` from the HttpClientException and pass that as-is into the `HttpClientError` that we're re-raising.  It just felt a little weird to me to take the output from `inspect` on one object and have that become `message` on a different object.

Practically speaking, I think the difference in doing that is only that the name of the Java Exception class will end up not getting appended to the `message` in the `HttpClientError` Ruby class.

@jeffmccune - do you have any recommendations / preferences here?

jeffmccune: I agree, I think it should be message rather than inspect.

nwolfe:
@camlow325 Updated.

I'm so sorry team. I know what I'm doing this weekend :disappointed:

camlow325:
:+1: and I understand that @jeffmccune is :+1: as well from the paired code review he did on this, so merging...

PR #483, opened 2015-03-31
DOCS: fixed broken bullet list in configuration.markdown

jbondpdx:
One little line, causing the doc to render wrong.


nwolfe:
:+1:

PR #484, opened 2015-04-01
(SERVER-325) reorganize code for max requests

cprice404:
This PR is a pre-cursor to the real implementation details for SERVER-325.  The goal of that ticket is to introduce a new setting that allows users to configure JRuby instances to be flushed after being used for a certain number of requests.

There was a lot of code clean-up / reorganization necessary in order to make it possible to implement that functionality.  The main issue was that there were some circular dependencies between some of the namespaces that make up the JRuby pool functionality, and we needed to break those namespaces up into smaller units in order to be able to wire everything together without circular dependencies.

The commits related to that code reorganization are very susceptible to merge conflicts if any other work is being done related to the JRuby scripting containers.  Therefore, in the interest of minimizing conflicts, I'm submitting this PR to go ahead and get all of the re-organization out of the way.

This should hopefully be in a mergeable state as-is, and if we can get it in, that should allow us to unblock other JRuby-related changes while I finish up the flushing implementation.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
@@ -60,6 +60,20 @@
         (throw (IllegalStateException. "There was a problem adding a JRubyPuppet instance to the pool." e))))))
 
 (schema/defn ^:always-validate
+  flush-instance!
+  "Flush a single JRuby instance.  Create a new replacement instance
+  and insert it into the specified pool."
+  [{:keys [scripting-container id]} :- jruby-core/JRubyPuppetInstanceOrRetry
+   new-pool :- jruby-core/pool-queue-type
+   new-id   :- schema/Int
+   config   :- jruby-core/JRubyPuppetConfig
+   profiler :- (schema/maybe PuppetProfiler)]
+  (.terminate scripting-container)
+  (log/infof "Cleaned up old JRuby instance with id %s, creating replacement."
+             id)
+  (jruby-core/create-pool-instance! new-pool new-id config profiler))
+
+(schema/defn ^:always-validate
   flush-pool!
   "Flush of the current JRuby pool.  NOTE: this function should never
   be called except by the pool-agent."
@@ -81,12 +95,9 @@
     (log/info "Swapped JRuby pools, beginning cleanup of old pool.")
     (doseq [i (range count)]
       (try
-        (let [id (inc i)
-              instance (jruby-core/borrow-from-pool (:pool old-pool))]
-          (.terminate (:scripting-container instance))
-          (log/infof "Cleaned up old JRuby instance %s of %s, creating replacement."
-                     id count)
-          (jruby-core/create-pool-instance! new-pool id config profiler)
+        (let [id        (inc i)
+              instance  (jruby-core/borrow-from-pool (:pool old-pool))]
+          (flush-instance! instance new-pool id config profiler)
           (log/infof "Finished creating JRubyPuppet instance %d of %d"
                      id count))
         (catch Exception e
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
@@ -1,5 +1,5 @@
 (ns puppetlabs.services.jruby.jruby-puppet-agents
-  (:import (clojure.lang IFn Agent)
+  (:import (clojure.lang IFn)
            (com.puppetlabs.puppetserver PuppetProfiler)
            (puppetlabs.services.jruby.jruby_puppet_schemas PoisonPill RetryPoisonPill))
   (:require [schema.core :as schema]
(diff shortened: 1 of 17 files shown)


cprice404:
This PR probably looks a little weird on its own, as it doesn't really change the behavior very much compared to the current system and it doesn't actually introduce the support for the max-requests-per-instance setting, but I discussed with @camlow325 this morning and we agreed that it would make sense to try to get this in soon so that we're not having to tip-toe around the JRuby code so much over the next few weeks.

nwolfe:
Not a big deal but commit https://github.com/cprice404/puppet-server/commit/c829af05327a231546c49ac96d55aec2490b0109 is incorrectly marked `(SERVER-235)` instead of `(SERVER-325)`

cprice404:
@nwolfe thanks, rebased with a fix to that commit message.

src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
```diff
@@ -1,34 +1,21 @@
 (ns puppetlabs.services.jruby.jruby-puppet-agents
-  (:import (clojure.lang IFn Agent)
-           (puppetlabs.services.jruby.jruby_puppet_core PoisonPill RetryPoisonPill)
-           (com.puppetlabs.puppetserver PuppetProfiler))
+  (:import (clojure.lang IFn)
+           (com.puppetlabs.puppetserver PuppetProfiler)
+           (puppetlabs.services.jruby.jruby_puppet_schemas PoisonPill RetryPoisonPill))
```
nwolfe: Interesting - how come the `:import` with a clojure namespace?  Can't we just reference the pills with `jruby-schema/PoisonPill`?

Not a big deal, just curious since I don't think I've seen this style before.

cprice404: `defrecord` ultimately results in creating a Java class behind the scenes, so this is how you have to reference them as far as I understand.  (Cursive just did this for me though, so I'm slightly less than 100% certain about that.)

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -48,22 +48,22 @@
...
-                          "begin; InstanceID; false; rescue NameError; true; end")))))))))
+        (await (:pool-agent pool-context))
+        (is (every? true?
+                    (jruby-testutils/reduce-over-jrubies!
+                      pool-context
+                      4
+                      (constantly
+                        "begin; InstanceID; false; rescue NameError; true; end"))))))))
+
+;; TODO: test flush instance while pool flush is in progress
```
KevinCorcoran: TODO

KevinCorcoran:
I reviewed this but skimmed it pretty quickly.  I'll do a more thorough review when the more controversial changes land.

:+1:

test/unit/puppetlabs/services/jruby/jruby_testutils.clj
```diff
@@ -52,15 +54,14 @@
...
   will be specified."
   ([]
-   {:ruby-load-path  ruby-load-path
-    :gem-home        gem-home
-    :master-conf-dir conf-dir})
+   (jruby-core/initialize-config
+     {:jruby-puppet {:gem-home        gem-home
+                     :master-conf-dir conf-dir
+                     :master-var-dir  var-dir}
+      :os-settings  {:ruby-load-path ruby-load-path}}))
```
nwolfe: Is it weird that `:os-settings` is now in two places in the overall config? The `jruby-puppet-tk-config` function above this adds it to the top-level config too.

Not a big deal, just thinking out loud

cprice404: So, this is confusing, but there are actually two different kinds of config objects in play here.

The one you commented on is basically the raw structure that you'd find in the config files (on `stable`, anyway; in `master` the `os-settings` map should be gone entirely).

The one before that is a different thing.  During the initialization of the jruby service there is some validation / transformation code:

https://github.com/cprice404/puppet-server/blob/feature/stable/SERVER-325-reorganize-code-for-max-requests/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj#L84-L103

whose job it is to take the raw config from the files and build up a map that complies with the `JRubyPuppetConfig` schema, which is what we pass around from that point forward:

https://github.com/cprice404/puppet-server/blob/feature/stable/SERVER-325-reorganize-code-for-max-requests/src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj#L30-L64

It's confusing that we use both in the tests, but I think it's probably warranted because they're both necessary if you want to have tests at both levels of the system.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -0,0 +1,205 @@
...
+
+          ((some-fn nil? jruby-schemas/retry-poison-pill?) instance)
+          instance
+
+          :else
+          (throw (IllegalStateException.
+                   (str "Borrowed unrecognized object from pool!: " instance))))))
+
+(schema/defn ^:always-validate
+  borrow-from-pool :- jruby-schemas/JRubyPuppetInstanceOrRetry
```
nwolfe: Should this be `borrow-from-pool!` since it calls `borrow-from-pool!*`?

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -0,0 +1,205 @@
...
+  timeout. If the timeout runs out then nil will be returned, indicating that
+  there were no instances available."
+  [pool-context :- jruby-schemas/PoolContext
+   timeout :- schema/Int]
+  {:pre  [(>= timeout 0)]}
+  (borrow-from-pool!* (partial borrow-with-timeout-fn timeout)
+                      (get-pool pool-context) pool-context))
+
+(schema/defn ^:always-validate
+  return-to-pool
```
nwolfe: Should this be `return-to-pool!` since it doesn't return anything but instead changes the argument? Not a big deal though, just like all my other comments :smile:

nwolfe:
@cprice404 Looks like the commit message for https://github.com/cprice404/puppet-server/commit/4a12a0911c3aedaf3e96a2ff021fec2a9d65a289 is backwards, should be LIFO->FIFO like the previous commit explains.

I only mention it since you went through the trouble of writing up good commit messages, which is awesome since I was curious about the LIFO/FIFO change.

nwolfe:
:see_no_evil: :+1:

cprice404:
thx @nwolfe , will hit those last few things you mentioned now.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
KevinCorcoran: Actually, I was thinking `borrow-from-pool!*` should just be `borrow-from-pool*`.  It returns a value, right?

cprice404: i don't really care what they are called but i was in the middle of changing them right now, so, we need to pick one :)

cprice404: or punt this to a later PR or something

KevinCorcoran: It's not a big deal, punting is fine.

My opinion is that the `!` is most valuable when it is used only on functions that do not return a value and only exist for their side effects.

cprice404: i'm going to punt for now then.  the `return` function conforms to your definition but the `borrow` one does not.

cprice404:
I confused myself for a second, but the change we're making is from FIFO->LIFO.  Updated the first commit message to reflect that.

cprice404:
and, punting on the `!` stuff for now, so I think I've addressed all of the feedback so far.

KevinCorcoran:
I left a couple of comments on commits that seem to have gotten lost during a force-push, but they were very minor so no big deal.  Looks like we're all good on this.

PR #485, opened 2015-04-03
(SERVER-535) Fix several bugs in j-p-service-test

cprice404:
Many (most?) of the configuration maps that were being used
in the jruby-puppet-service-tests were invalid, but this wasn't
surfacing in the test runs because of the way the mock
JRuby instances were being created.  This commit improves
the validation of the input to the mock fixture, and fixes all
of the test configurations.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
@@ -51,7 +51,11 @@
 
     * :http-client-cipher-suites - A list of legal SSL cipher suites that may
         be used when https client requests are made."
-  {:ruby-load-path              (schema/both [schema/Str] (schema/pred vector?))
+  ;; NOTE: there is a bug in the version of schema we're using, which causes
+  ;; the order of things that you put into a `both` to be very important.
+  ;; The `vector?` pred here MUST come before the `[schema/Str]`.  For more info
+  ;; see https://github.com/Prismatic/schema/issues/68
+  {:ruby-load-path              (schema/both (schema/pred vector?) [schema/Str])
    :gem-home                    schema/Str
    :master-conf-dir             (schema/maybe schema/Str)
    :master-var-dir              (schema/maybe schema/Str)
--- a/test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
+++ b/test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
@@ -125,7 +125,7 @@
               pool-context (:pool-context context)]
           (let [jrubies (jruby-testutils/drain-pool pool-context pool-size)]
             (is (= 1 (count jrubies)))
-            (is (every? #(jruby-schemas/jruby-puppet-instance? %) jrubies)))
+            (is (every? jruby-schemas/jruby-puppet-instance? jrubies)))
           (let [test-start-in-millis (System/currentTimeMillis)]
             (is (nil? (jruby-protocol/borrow-instance service)))
             (is (>= (- (System/currentTimeMillis) test-start-in-millis) timeout))
--- a/test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
+++ b/test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
@@ -1,6 +1,5 @@
 (ns puppetlabs.services.jruby.jruby-puppet-service-test
-  (:import (java.util.concurrent ExecutionException)
-           (com.puppetlabs.puppetserver JRubyPuppet))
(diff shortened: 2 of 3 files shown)


test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
```diff
@@ -117,7 +123,9 @@
         (let [service (app/get-service app :JRubyPuppetService)
               context (services/service-context service)
               pool-context (:pool-context context)]
-          (jruby-testutils/drain-pool pool-context pool-size)
+          (let [jrubies (jruby-testutils/drain-pool pool-context pool-size)]
+            (is (= 1 (count jrubies)))
+            (is (every? #(jruby-schemas/jruby-puppet-instance? %) jrubies)))
```
nwolfe: Don't need the anonymous function wrapping a function, right?

test/unit/puppetlabs/services/jruby/jruby_testutils.clj
```diff
@@ -49,10 +54,14 @@
    :jruby-puppet pool-config
    :certificate-authority {:certificate-status {:client-whitelist []}}})
 
-(defn jruby-puppet-config
+(schema/defn ^:always-validate
+  jruby-puppet-config :- jruby-schemas/JRubyPuppetConfig
   "Create a JRubyPuppet pool config. If `pool-size` is provided then the
```
nwolfe: Since we're in here and updating the docstring...can we get rid of this old `pool-size` mention? This function no longer takes that as a parameter.

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -51,7 +51,11 @@
 
     * :http-client-cipher-suites - A list of legal SSL cipher suites that may
         be used when https client requests are made."
-  {:ruby-load-path              (schema/both [schema/Str] (schema/pred vector?))
+  ;; NOTE: there is a bug in the version of schema we're using, which causes
+  ;; the order of things that you put into a `both` to be very important.
+  ;; The `vector?` pred here MUST come before the `[schema/Str]`.  For more info
+  ;; see https://github.com/Prismatic/schema/issues/68
```
nwolfe: Should we create a ticket for upgrading to the latest schema?  We're several versions behind, and I think we have other projects that are already on newer versions.

cprice404: I was chatting with Dan in clojurians yesterday, and he mentioned that when he tried upgrading to a newer version, it broke his recursive schemas.  We don't have any recursive schemas in Puppet Server yet, but we will as soon as we switch over to the new web routing library, so I was thinking I might try updating when I get back to working on that project.

If you want to create a ticket that'd be fine, but I'd add a comment capturing what Dan said.

cprice404:
@nwolfe: added a commit addressing your two things.

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
```diff
@@ -14,18 +13,24 @@
             [puppetlabs.trapperkeeper.testutils.logging :as logging]
             [puppetlabs.services.puppet-profiler.puppet-profiler-service :as profiler]
             [puppetlabs.services.jruby.jruby-puppet-core :as jruby-core]
```
nwolfe: You didn't introduce this, but I notice we're requiring this ns twice as `jruby-puppet-core` and `jruby-core`

cprice404: Yeah, that's fixed in my next PR.

nwolfe:
:+1:

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
```diff
@@ -14,18 +13,24 @@
...
+  (jruby-testutils/jruby-puppet-tk-config
+    (jruby-testutils/jruby-puppet-config {:max-active-instances pool-size})))
 
 (defn jruby-service-test-config-with-timeouts
   [connect-timeout idle-timeout]
-  (assoc jruby-service-test-config
-    :http-client {:connect-timeout-milliseconds connect-timeout
-                  :idle-timeout-milliseconds    idle-timeout}))
+  (merge (jruby-testutils/jruby-puppet-tk-config
+           (jruby-testutils/jruby-puppet-config {:max-active-instances 1}))
```
KevinCorcoran: looks like you can just re-use `jruby-service-test-config` here

cprice404: truth

PR #487, opened 2015-04-03
(SERVER-325) Add support for max-requests

cprice404:
This commit adds support for flushing an individual JRuby
instance after it has exceeded the number of requests
specified by the `max-requests-per-instance` setting.

There are a couple of additional tests I would like to add to this, but other than that I think it should be ready to go.  I can add those in a subsequent PR if this one gets reviewed and is deemed mergeable, and that might help us minimize merge conflicts.


src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -143,31 +154,27 @@
    pool :- jruby-schemas/pool-queue-type]
   (.pollFirst pool timeout TimeUnit/MILLISECONDS))
 
+(schema/defn borrowed-poison-pill
+  [pool :- jruby-schemas/pool-queue-type
+   instance :- PoisonPill]
+  (.putFirst pool instance)
+  (throw (IllegalStateException.
```
nwolfe: Is this missing an `if` or something? I don't imagine we want to throw an exception all the time

nwolfe: Oh, I see...this is really awkward looking as it's own function without the context.  I might just leave it where it was, or add a docstring maybe?

Or maybe it's just me?

KevinCorcoran: I suggest deleting this function and moving its code back into the `cond` below - that's the only place its used anyway, and as @nwolfe, it sure looks baffling out of context.

Unless this needs to be reused?

test/unit/puppetlabs/services/jruby/jruby_pool_test.clj
```diff
@@ -96,30 +96,63 @@
...
-          pool (create-pool-context config profiler nil)
+          pool (jruby-core/create-pool-context config profiler jruby-testutils/default-shutdown-fn)
           pool-state @(:pool-state pool)]
-      (is (= (core/default-pool-size (ks/num-cpus)) (:size pool-state))))))
+      (is (= (jruby-core/default-pool-size (ks/num-cpus)) (:size pool-state))))))
 
+(defn create-pool-context
+  [max-requests]
+  (let [config (-> (jruby-testutils/jruby-puppet-config {:max-active-instances 1})
+                   (assoc :max-requests-per-instance max-requests))
```
nwolfe: No biggie but you could get rid of the `assoc` and just put this key-value into the hash argument to `jruby-puppet-config`

cprice404:
This is rebased, and I have addressed @nwolfe and @KevinCorcoran 's initial stylistic suggestions.  I've also added another test based on a suggestion that @camlow325 in an earlier version of this PR.  I have one more integration test that I'm going to attempt to write but other than that I think this is finished, pending additional feedback.

cprice404:
Added a few more commits to improve the integration test coverage (specifically, testing an instance flush via max-requests-per-instance while a pool flush is already in progress).

I'm done with this PR, AFAIK, at this point.

KevinCorcoran:
> Increase the max heap in the project.clj until travis stops failing ( :no_good: )

Are there any downsides to this option?  Seems reasonable to me.  Although ... I wonder why you're not seeing the same behavior locally?  I wouldn't expect memory usage to differ *that much* between your laptop and travis.

KevinCorcoran:
> the max heap is set in the project file to 256MB

This is wrong, it's actually `MaxPermSize`.

Interestingly, the `OutOfMemoryError` in https://travis-ci.org/puppetlabs/puppet-server/jobs/58275376 does *not* seem to be about perm gen, so increasing that value in project.clj wouldn't have helped.

But, yeah, we could still try setting `Xmx`.

KevinCorcoran:
I tried setting `-Xmx` super low in project.clj and confirmed that I could reproduce the OutOfMemoryError during `lein test`, so, it seems like setting that value to something reasonable for our test suite could be a path to nirvana here.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -196,10 +199,24 @@
...
 (schema/defn ^:always-validate
   return-to-pool
   "Return a borrowed pool instance to its free pool."
   [instance :- jruby-schemas/JRubyPuppetInstanceOrRetry]
-  (.putFirst (:pool instance) instance))
+  (if (jruby-schemas/jruby-puppet-instance? instance)
+    (let [new-state (swap! (:state instance)
+                           update-in [:request-count] inc)
+          {:keys [max-requests flush-instance-fn pool]} instance]
+      (if (and (pos? max-requests)
```
KevinCorcoran: It looks like `max-requests` can only be a positive integer or `nil` here ... if that's the case, I think it's better to use `(not (nil? ...` instead of `pos?` here.

test/unit/puppetlabs/services/jruby/jruby_puppet_agents_test.clj
```diff
@@ -98,4 +98,16 @@
...
+  (let [pool-context (jruby-core/create-pool-context
+                       (jruby-testutils/jruby-puppet-config {:max-active-instances 8})
+                       jruby-testutils/default-profiler
+                       jruby-testutils/default-shutdown-fn)]
+    (testing "next instance id should be based on the pool size"
+      (is (= 10 (jruby-agents/next-instance-id 2 pool-context)))
+      (is (= 100 (jruby-agents/next-instance-id 92 pool-context))))
+    (testing "next instance id should wrap after max int"
+      (let [id (- Integer/MAX_VALUE 1)]
+        (is (= (mod id 8) (jruby-agents/next-instance-id id pool-context)))))))
```
KevinCorcoran: I'm having trouble understanding why `next-instance-id` is implemented this way.  Why not just start with 1 and increment from there?   Also I don't think we need to worry about overflowing Long .... http://stackoverflow.com/questions/15505515/java-long-primitive-type-maxumum-limit

KevinCorcoran:
I've reviewed all of the production code.  It looks mostly good, aside from my confusion around `next-instance-id`.  I'll take a look at the test code tomorrow.

cprice404:
> > Increase the max heap in the project.clj until travis stops failing ( :no_good: )
> Are there any downsides to this option? Seems reasonable to me. Although ... I wonder why you're not seeing the same behavior locally? I wouldn't expect memory usage to differ that much between your laptop and travis.

Downsides:
1. I don't really know what a sane value would be there, nor how high Travis will let us go, or even what Travis's default is
2. This test, on my laptop, increases the total run time of the test suite from 3 minutes to 5 minutes.  That seems like it might be pushing it.  We could always consider extracting the JRuby service to a separate git project at some point, but still...

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
cprice404: Why?  `pos?` seems like it makes it more obvious what kind of value we're expecting here, which makes the code more readable?

cprice404: Actually, the sentinal value for max-requests is `0`, not `nil`, so that change wouldn't work.

test/unit/puppetlabs/services/jruby/jruby_puppet_agents_test.clj
cprice404: I think we definitely need to worry about overflowing Long, in terms of "could it theoretically happen".  If you're suggesting that it would be OK to use negative integers as ids, then we could get away without worrying about it... but these ids will show up in log messages, so I thought it would be more user-friendly to not allow negatives.  I was actually thinking about limiting the range to something much smaller, like, 1000 or something, and overflowing back to 1 after we hit that limit.

KevinCorcoran: This seems like unnecessary complication to me.  I don't think negative IDs in log messages is a UX we should care about ... especially since that is very likely to never actually happen (unless somebody manages to run the server for 292 years without restarting it, while handling 1 request/nanosecond the entire time and having `max-requests = 1`).

Also, I was wondering why the ID grows by `pool-size` instead of just 1.

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
```diff
@@ -106,8 +105,16 @@
...
+        ;; borrow and return one more time
+        (with-jruby-puppet
+          jruby-puppet
+          service)
         (let [jruby (jruby-protocol/borrow-instance service)]
-          (is (= 2 (:request-count (jruby-core/instance-state jruby)))))))))
+          ;; the counter gets incremented when the instance is returned to the
+          ;; pool, so right now it should be at 2 since we've called
+          ;; `with-jruby-puppet` twice.
+          (is (= 2 (:request-count (jruby-core/instance-state jruby))))
```
KevinCorcoran: Perhaps rename this key to `:borrow-count` to maintain the separation of concerns between the JRuby service and request handling.

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
```diff
@@ -106,8 +105,16 @@
           (is (instance? JRubyPuppet jruby-puppet))
           (is (= 0 (jruby-protocol/free-instance-count service))))
         (is (= 1 (jruby-protocol/free-instance-count service)))
+        ;; borrow and return one more time
+        (with-jruby-puppet
+          jruby-puppet
+          service)
```
KevinCorcoran: I think it's a little funky to use the macro here and an explicit `borrow` and `return` would be better.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -49,21 +123,99 @@
...
-                        "begin; InstanceID; false; rescue NameError; true; end"))))))))
-
-;; TODO: test flush instance while pool flush is in progress
+        ;; set a ruby constant in each instance so that we can recognize them
+        (is (true? (set-constants-and-verify pool-context 4)))
+        (let [flush-complete (add-watch-for-flush-complete pool-context)]
+          (is (true? (trigger-flush)))
+          @flush-complete)
+        ;; now the pool is flushed, so the constants should be cleared
+        (is (true? (verify-no-constants pool-context 4)))))))
```
KevinCorcoran: Wolfe Principal violation!

> <script>:1 warning: already initialized constant InstanceID

ahpook: Sorenson Principle violation!  

A principal runs a school, a principle is an idea you live by.

KevinCorcoran: Sorenson Principle = use the word "principal" correctly

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -49,21 +123,99 @@
...
+
+            ;; now we're going to return instance2 to the pool.  This should cause it
+            ;; to get flushed, but after that, the main pool flush operation should
+            ;; pull it out of the old pool and create a new instance in the new pool
+            ;; to replace it.  So we should end up with 3 instances in the new pool,
+            ;; two of which should have the ruby constants and one of which should not.
+            (jruby-protocol/return-instance jruby-service instance2)
+            (is (true? (check-jrubies-for-constant-counts pool-context 2 1))))
+
+          ;; now we'll set the ruby constances on the 3 instances in the new pool
```
KevinCorcoran: s/constances/constants

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -49,21 +123,99 @@
...
+          (is (true? (set-constants-and-verify pool-context 3)))
+
+          ;; and finally, we return the last instance from the old pool
+          (jruby-protocol/return-instance jruby-service instance1)
+
+          ;; wait until the flush is complete
+          @flush-complete)
+
+        ;; we should have three instances with the constant and one without.
+        (is (true? (check-jrubies-for-constant-counts pool-context 3 1)))))))
```
KevinCorcoran: This is a helluva test!  :1234:

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
KevinCorcoran: oh, I mistakenly thought it was `nil`.

test/unit/puppetlabs/services/jruby/jruby_puppet_agents_test.clj
camlow325: I'd think the negative IDs in log messages could be confusing.  I'd generally expect them to increment and wrap around to a non-negative value at some point.  That being said, agree that it's highly unlikely that would ever be a problem unless we decided to lower the max to something much smaller at some point.

I believe the next ID grows by `pool-size` instead of 1 because we wouldn't want to reuse an ID for another instance in the pool that's still active, right?

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
cprice404: I'm not sure that this is going to be easy to work around because this is just the way that Ruby behaves, but I'll look into it.

test/unit/puppetlabs/services/jruby/jruby_puppet_agents_test.clj
cprice404: Yes, the initial ids are generated via a simple increment by 1, and subsequent ones need to be generated by adding the pool size to the id of the instance we're flushing, otherwise we'll end up with duplicate IDs.

So... it seems like we have two votes for not allowing negative IDs, and thus the function is necessary, and it's just a matter of determining what the maximum ID should be before we wrap around?  I wouldn't mind changing it to something like 10k instead of Long.MAX_VALUE.

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
cprice404: I like that idea.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
KevinCorcoran: Yeah, I figured ... I wonder if we could use a different ruby snippet which would work for these tests but not produce warnings.  Nothing's coming to mind, though ...

test/unit/puppetlabs/services/jruby/jruby_puppet_agents_test.clj
KevinCorcoran: Okay, glad to hear that I misread the code and that the IDs increase by 1 and not `pool-size`.  

Disallowing negative IDs is fine with me.  However, choosing some arbitrary value to cap them at seems arbitrary and unnecessary.

cprice404: In that case, it sounds like you're not proposing any changes to the current implementation of the next-id fn?

KevinCorcoran: Curveball:

What about using an atom, holding an integer, which can be used as an "ID factory"?  It would eliminate the need for the logic in the `next-instance-id` functionality entirely (ie, modulus, wrapping, etc.).

I think we're reaching the point of :bike:shed on this one ... so ... ya know, if we want to just call it and leave the code as-is, I won't lose any sleep.

camlow325: I'd actually prefer what's already in the PR to having to introduce any new state.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
camlow325: Is this not something we can whisk away with `with-test-logging`?

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -49,21 +123,99 @@
...
+      (let [jruby-service (tk-app/get-service app :JRubyPuppetService)
+            context (tk-services/service-context jruby-service)
+            pool-context (:pool-context context)]
+        ;; set a ruby constant in each instance so that we can recognize them
+        (is (true? (set-constants-and-verify pool-context 4)))
+        (let [flush-complete (add-watch-for-flush-complete pool-context)
+              ;; borrow an instance and hold the reference to it.
+              instance (jruby-protocol/borrow-instance jruby-service)]
+          ;; trigger a flush
+          (is (true? (trigger-flush)))
```
camlow325: Would it be better to have the `trigger-flush` return the status from the response and have the caller assert equivalence in the body of the test itself - so a failed assertion would show you the "bad" response code that the server returned?  The way you have it is more concise but a failure of this assertion would probably lead someone to have to step through / further instrument the `trigger-flush` function to figure out what happened.  Not a big deal to me...

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -37,6 +37,80 @@
    :ssl-key     localhost-key
    :ssl-ca-cert ca-cert})
 
+(def script-to-check-if-constant-is-defined
+  "begin; InstanceID; true; rescue NameError; false; end")
```
camlow325: To avoid the Wolfe Principle violation, could we do something like this?

~~~~
"if (defined? InstanceID); InstanceID; else; false; end"
~~~~

camlow325: n/m, totally misunderstood that the warning was coming from the constant being initialized twice from `set-constants-and-verify` rather than from checking from here if the constant is defined.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
camlow325: Tried that out - `with-test-logging` didn't help.

camlow325:
I'm basically :+1: on this PR, pending what we end up deciding to do about the double-flushing test.  I also still want to do some work around SERVER-251 to assess whether or not turning this feature on might lead to a Bad File Descriptor if a container is terminated while a file handle is open in another container.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
cprice404: I'll play with some ideas around the WP violation after we get the more important details squared away.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
cprice404: Yeah, I waffled on that.  I'll take another look at it because I agree in principle.

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
cprice404: sure

cprice404:
Summarizing what I believe to be the outstanding work on this PR:

* Look into mocking the JRubies for the one giant insane test
* Rename `:request-count` to `:borrow-count`
* Change `with-jruby-puppet` to `borrow`/`return` in j_p_service_test
* Attempt to resolve WP violation
* `constances` typo
* Look into having the `trigger-flush` test helper fn return something other than a boolean

@KevinCorcoran @camlow325 did I miss anything?

Also, @camlow325, were you thinking that we need to gate this PR on your BFD exploration?  Or that if the BFD exploration turns up something that we'll just need to do some additional work to solve that?

KevinCorcoran:
looks right to me

camlow325:
The summary all sounds right to me.

> Also, @camlow325, were you thinking that we need to gate this PR on your BFD exploration? Or that if the BFD exploration turns up something that we'll just need to do some additional work to solve that?

Probably better to just let the PR go in and handle the outcome of the BFD exploration later.  If it turns out this change would make the implementation more susceptible to BFDs and any residual work needed to overcome it wouldn't be doable in time for whenever our next release needs to land, we'll have to talk about how we want to deal with that - e.g., maybe even leaving the code in place but turning off the `max-requests` cap by default.  I understand that discussion would be very premature to have at this point.  Without significant performance penalties - which may make the whole feature untenable - I'd have a hard time envisioning that the discovery of BFD susceptibility would lead to any major architectural changes to what is currently in the PR.

So, yeah, I'm okay letting the PR come in once the items in your summary have been addressed.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
camlow325: Assuming you go with a "mock" approach for the `max-requests-flush-while-pool-flush-in-progress-test`, I'm thinking the WP violation might go away automatically - assuming you have to come up with an alternative way to the const for storing state like `set-constants-and-verify` does - one that hopefully wouldn't come with a warning message as a side effect.  I think this was the only test affected by the WP violation.

cprice404:
Rebased and added some commits.  This includes:

* Mocking JRubies for the one giant test; this shaves almost 2 minutes off of the test suite run on my laptop.
* Change `request-count` to `borrow-count`
* Fixed WP violation
* Fixed constances typo

There were two things in the list that I didn't end up doing.  Here's why, and I'm open to additional discussion on both:

* Did not change `with-jruby-puppet` to `borrow`/`return` in the jruby_puppet_service test, because the test in question is actually intended to test the `with-jruby-puppet` macro.  I did add a clarifying comment to try to make this more clear.
* I didn't change the `trigger-flush` test fn, because I didn't see anything that it could return that would actually be meaningful in the tests.  I started going down the road of returning the HTTP status code and having each test verify it, but it just seemed weird and out-of-place to me, so I rolled it back.  I can go ahead and do it anyway if you think it's important, or we can refactor as needed as the test namespace evolves in the future.

KevinCorcoran:
:+1: Left one more nitpick but we can sure :ship: this without it.  @camlow325 I will let you have one more look at this, feel free to merge.

camlow325:
Interesting on the mock JRubies for the big test - I'd been thinking that creating a `ScriptingContainer` and doing a `runScriptlet` on it would still be reasonably expensive - though certainly not nearly as expensive as initializing all of Puppet within that container, of course.  So if we're still including "real" `ScriptingContainer`s and running some Ruby code through these, doesn't seem like we lost much by going with "mocks" anyway.  Seems good to me.

camlow325:
:+1: :shipit:

PR #488, opened 2015-04-07
(SERVER-484) Update version to 2.0.0

jeffmccune:
This commit bumps the version to the 2.0.0 identifier.


jeffmccune:
Would someone mind taking a look at this so I could start the process of building the real 2.0.0 packages?  My only concern is if this is going to make it difficult to promote the 2.0.0 version into PE, but I don't think it matters since we'll hopefully be promoting something else.

nwolfe:
:+1:

PR #490, opened 2015-04-09
(SERVER-480) Test against puppet-agent 1.0.0

jeffmccune:
Without this patch puppetserver 2.0.0 is testing against a development build of
puppet agent.  This patch addresses the problem by moving the pin from a
development build to the real deal 1.0.0 version of the puppet-agent packages.


camlow325:
:+1:

PR #491, opened 2015-04-09
(SERVER-534) Bump puppet submodule to 3.7.5

dankreek:


camlow325:
I think the "puppet version" test in master_spec.rb will need to be bumped up to "3.7.5".

dankreek:
Ahh yes, of course. Thanks, @camlow325

camlow325:
Are there any changes to the Gemfile or Gemfile.lock files that we should make to keep in sync with 3.7.5?

camlow325:
Also, should our facter submodule be bumped as well - not sure if a new one was shipped along with 3.7.5?

dankreek:
Spec tests all currently pass with the current lock file, and I'm guessing that if there hasn't been any change to our acceptance tests then it should be fine as is. Does this sound correct?

dankreek:
Looks like the Puppet Gemfile.lock specifies Facter 2.4.1. Does that sounds like a sane version to bump to?

camlow325:
> Spec tests all currently pass with the current lock file, and I'm guessing that if there hasn't been any change to our acceptance tests then it should be fine as is.

It's probably ok.  I was just thinking that it would be best for all of our puppet-server CI jobs to be running with the same versions of gems as would the corresponding puppet platform CI jobs for 3.7.5.  For example, same beaker version.

dankreek:
Ahh ok, that makes sense.

camlow325:
> Looks like the Puppet Gemfile.lock specifies Facter 2.4.1. Does that sounds like a sane version to bump to?

Good question.  When I do an install of 'puppet' from the latest package repos right now, I get version 2.4.3 of Facter.

@kylog For our "stable" branch of Puppet Server - 1.x series supporting Puppet 3.x - what version of Facter would you suggest we test along with the Puppet 3.7.5 version that we're bumping to?

kylog:
Yeah I'd probably bump to 2.4.3. It just has misc fixes on top of 2.4.1 so should be low risk. 

If it does give you heartache relative to 2.4.1, please let us know!

dankreek:
Cool, facter's been bumped up to 2.4.3

camlow325:
:+1: Looks good to me.  @nwolfe - you good with this?

nwolfe:
:100:

PR #492, opened 2015-04-10
(maint) Pessimistically pin to beaker 2.9.0

jeffmccune:
Without this patch puppetserver is pinned to a specific GIT sha of beaker.
This is a problem because using a Git checkout has the side effect of transient
false positive failures as described in QENG-2138.  This patch addresses the
problem by updating the dependency expression to so-called pessimistic
dependency that pins beaker to released versions installed via rubygems, not
git, in the 2.9.x series.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ gem 'jira-ruby', :group => :development
 
 group :test do
   gem 'rspec'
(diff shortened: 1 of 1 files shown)


rlinehan:
:+1:

dankreek:
:+1:

PR #493, opened 2015-04-10
(SERVER-391) Provide better error message for borrow timeout

dankreek:


dankreek:
Adding the X-Puppet-version header and will repost this soon.

PR #495, opened 2015-04-13
(SERVER-391) Provide better error message for borrow timeout

dankreek:


src/clj/puppetlabs/services/request_handler/request_handler_service.clj
```diff
@@ -3,18 +3,28 @@
...
 
 (defn- handle-request
-  [request jruby-service config]
-  (jruby/with-jruby-puppet jruby-puppet jruby-service
-    (core/handle-request request jruby-puppet config)))
+  [request jruby-service config puppet-version]
+  (try
+    (jruby/with-jruby-puppet jruby-puppet jruby-service
+                             (core/handle-request request jruby-puppet config))
+    (catch IllegalStateException e
```
camlow325: In `request-handler-core/handle-request`, we already have a sling try/catch block for one case in which a response needs to be remapped.  I'm thinking it might be good to consolidate these try/catches together in one place rather than introducing a new one.  I'd probably lean toward pushing the implementation of this `handle-request` down into the `request_handler_core` namespace and doing the exception handling consolidation there.

src/clj/puppetlabs/services/request_handler/request_handler_service.clj
```diff
@@ -3,18 +3,28 @@
...
+      (let [message (.getMessage e)]
+        (log/error message)
+        {:status  503
+         :body    message
+         :headers {"Content-Type"     "text/plain"
+                   "X-Puppet-version" puppet-version}}))))
```
camlow325: Might be good to take the `wrap-with-puppet-version-header` middleware in the `certificate-authority-core` namespace and put it into a common place - maybe clj.puppetlabs.pupetserver.ringutils? - that both the CA and this function could call into to get the same wrapping?

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
```diff
@@ -76,7 +76,12 @@
   `(loop [pool-instance# (jruby/borrow-instance ~jruby-service)]
      (if (nil? pool-instance#)
        (throw (IllegalStateException.
-                "Error: Attempt to borrow a JRuby instance from the pool timed out")))
```
camlow325: Since this is now an `exception` that we'd be throwing and recatching from within Puppet Server, might be good to turn this into a Slingshot exception - like we do with the "bad requests" in the request-handler-core namespace.

src/clj/puppetlabs/services/request_handler/request_handler_service.clj
camlow325: I think the "bad request" exceptions that we throw from `request-handler-core` aren't getting an `X-Puppet-Version` header appended in the response either.  Would be nice to see if we could come up with a more general solution that would add the `X-Puppet-Version` header into the response delivered from `handle-request` for any case that we didn't get one already in the response to the core Puppet request.  This would hopefully help us avoid having to add this in every time we come up with future responses that we have to produce at the Clojure layer.

camlow325:
:+1: other than a few code reorganization suggestions.

src/clj/puppetlabs/services/request_handler/request_handler_service.clj
dankreek: As far as I can tell, the code in the `request-handler-core/handle-request` function runs after a successful borrow from the JRuby pool. So if we were to consolidate these (which I am totally in favaor of) I'm guess we'll need to do it here.

camlow325: Yeah, we'd definitely have to have the try/catch block wrapped around the invocation of `with-jruby-puppet` in order to be able to catch re-throw for a borrow timeout failure.  The only part I was debating was whether it would make sense to take the code in this function and push it down into the `request-handler-core` namespace or just leave it here - i.e., does this kind of logic more generally feel like something that should live in the *core* vs. the *service* namespace.  I'd be fine leaving it here - no strong opinion.  Maybe @cprice404 or @KevinCorcoran would have a preference?

KevinCorcoran: Offhand, I don't have a preference where the error handling lives.

However, I think it's wrong translate all `IllegalStateExceptions` into HTTP 503s.  In general, I would expect an `IllegalStateException` to be an HTTP 500.  Since the one particular error we actually want to catch here is thrown in our code (when the `borrow` times-out), I think it should throw a slingshot error, which can be caught instead and translated into an HTTP 503.

camlow325: > However, I think it's wrong translate all IllegalStateExceptions into HTTP 503s...

@KevinCorcoran Yep, totally agree with that and your suggestion about the slingshot error with translation to an HTTP 503.  That's also what I was expecting around a separate comment I made about this on the PR.

KevinCorcoran: d'oh!  I missed your comment above.

cprice404: Not sure where this landed and I wouldn't block the PR on it or anything, but in general, I'm in favor of keeping the code in the `service` namespace to an absolute minimum... it should mostly just be the interactions with TK to get references to services / contexts / etc., and everything else should be pushed down into core if possible.

jeffmccune:
:+1:

PR #496, opened 2015-04-14
(SERVER-442) Convert request vecor params to array

dankreek:


cprice404:
This looks reasonable to me, but I'm wondering if there's any kind of simple test we could add that would have caught the problematic behaviour?  If it's complicated, and not worth the ROI, I'm good with this w/o tests, but if it's easy it'd be a nice thing to add.

camlow325:
@cprice404 fyi, this code essentially recreates a commit, https://github.com/puppetlabs/puppet-server/commit/ce389e3a90e4b920ecc3764cceef68f6df13d105, from a prior PR, https://github.com/puppetlabs/puppet-server/pull/32.  I'd removed this code from the PR based on a comment thread between us with concerns about the performance impact of this work.  Not sure if the performance concerns are still valid here - and maybe we would have decided to let this through the first time anyway if we'd had a solid use case for this at the time.  Just wanted to bring that back up again in case you still had any concerns about this.

camlow325:
Hmmm, now that I look at the code in this PR again, I guess it is a little different than the original one I had put up.

In the original PR, I had also pulled in `ring-nested-params/nested-params-request`, which the current PR isn't proposing we do.  I think the only thing that the `nested-params-request` would add is the ability to interpret a key with square brackets in the name, e.g., "my_arr[]", as representing an "array" such that the key committed into the request map would be "my_arr" instead of "my_arr[]".  That change wouldn't matter for the bug that this PR is aimed at fixing - because the repeated key used in that request doesn't have square brackets in the name (only "ignore" and not "ignore[]").

Not sure if supporting the special array name awareness is a big deal / ever used in any Puppet API requests.  Suppose we could continue to leave that part out for now.

src/ruby/puppet-server-lib/puppet/server/master.rb
```diff
@@ -67,4 +68,24 @@ def puppetVersion()
...
+
+  def params_to_ruby(params)
+    Hash[params.collect do |key, value|
+      # Values for query string and/or form parameters which are specified
+      # with array-like syntax will be parsed by Ring into a Clojure
+      # PersistentVector, which derives from a Java List.  Need to
+      # translate the Java List into a Ruby Array so that the request
+      # handling logic in Ruby can make use of it.
+
+      # For example, a query string of 'arr[]=one&arr[]=two" will be translated
```
camlow325: Presuming we don't pull in the `ring-nested-params` middleware, it would be better to have an example here that doesn't have square brackets in the parameter name.

camlow325:
I agree with @cprice404 that it would be good to have some test coverage for this.  @dankreek would probably have better ideas about how to do this as a more integration-y kind of test that exercises pluginsync for real than I would.  I agree if we were to go that route that we should try to do something as simple as possible - not something we'd spend a lot of time on.

I'd be happy with at least a Ruby spec test for this.  To avoid having to do a bunch of mocking / test fixture setup for that, one way to do this would be:

1) Move the `params_to_ruby()` method and call from the Master class to be behind the `params` method in `Puppet::Server::Network::HTTP::Handler`.

This would happen before the call to `decode_params` that the `params` method makes.

Actually, I kind of like this change better anyway as it would consolidate our Ruby-level mangling of the params before they are handed off to the underlying route for processing to a single spot in the code.

2) Add a spec test or two to our `handler_spec.rb` in puppet-server-lib/puppet/jvm that would create a `TestHandler` pass in a map with/without a Java-type array param etc. into a call to the `params` method and confirm that the parameters were translated as expected.

We already have some tests like this for validating the conversion done to the request body for binary payloads.

camlow325:
:+1: other than adding some sort of test coverage for this.  I'm also good with avoiding use of the `ring-nested-params` middleware if we have not yet found a real use case where it would be needed for the Puppet APIs.

cprice404:
+1 on everything @camlow325 suggested:

* Now that we have a real use case, my previous concerns about perf are trumped for sure
* I'm in favor of avoiding the introduction of more upstream ring middleware if we don't have to do so yet
* That spec test strategy (and accompanying code reorg) sounds great.  I don't think we need to do any integration-y tests unless they are super simple.  Might be good to update the Jira ticket to let QA know what we did and didn't do for this, though.

dankreek:
@camlow325 @cprice404 Added a spec test for this.

camlow325:
The new tests look reasonable to me, although I think it might be a little better if we were to orient the test around the call to `params` rather than the call to `params_to_ruby`.  That way, we'd get the benefit of covering the new call that `params` makes to `params_to_ruby`.  Without the test going that far, the test could continue passing even if the call to `params_to_ruby` were removed or relocated to a different location later on.

camlow325:
:+1:  Merging since I think @cprice404 was basically :+1: on this already.

PR #498, opened 2015-04-15
(SERVER-544) stop duplicating request body for jruby

cprice404:
Due to some conflicting opinions about how request bodies should be processed between the Ruby Puppet code and ring's 'params' middleware, our current code has the unfortunate behavior of reading the body of the HTTP request into memory twice for any requests that are routed into the JRuby layer.

We can prevent this by duplicating a few functions from the ring middleware library in our own code base, which isn't awesome, but seems preferable to the unnecessary extra memory usage (which can be a pretty big chunk of RAM for requests like catalog/report).

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -27,7 +27,9 @@
                  ;; of its jar; please read the detailed notes above the
                  ;; 'uberjar-exclusions' example toward the end of this file.
                  [org.jruby/jruby-stdlib "1.7.19"]
-                 [clj-time "0.5.1" :exclusions [joda-time]]
+                 [joda-time "2.5"]
+                 [clj-time "0.6.0"]
+                 [ring/ring-core "1.3.2"]
                  [compojure "1.1.8" :exclusions [org.clojure/tools.macro]]
                  [liberator "0.12.0"]
                  [me.raynes/fs "1.4.5"]
--- a/project.clj
+++ b/project.clj
@@ -37,7 +37,6 @@
                  [commons-codec "1.9"]
                  [clj-yaml "0.4.0" :exclusions [org.yaml/snakeyaml]]
                  [slingshot "0.10.3"]
-                 [ring/ring-codec "1.0.0"]
                  [cheshire "5.3.1"]
                  [trptcolin/versioneer "0.1.0"]]
 
--- a/src/clj/puppetlabs/puppetserver/ring/middleware/params.clj
+++ b/src/clj/puppetlabs/puppetserver/ring/middleware/params.clj
@@ -1,0 +1,59 @@
+(ns puppetlabs.puppetserver.ring.middleware.params
+  (:require [ring.util.codec :as codec]
+            [ring.util.request :as req]))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; COPY OF RELEVANT FUNCTIONS FROM UPSTREAM ring.middleware.params LIBRARY
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defn parse-params [params encoding]
+  (let [params (codec/form-decode params encoding)]
+    (if (map? params) params {})))
+
+(defn content-type
+  "Return the content-type of the request, or nil if no content-type is set."
+  [request]
+  (if-let [type (get-in request [:headers "content-type"])]
+    (second (re-find #"^(.*?)(?:;|$)" type))))
+
+(defn urlencoded-form?
+  "True if a request contains a urlencoded form in the body."
+  [request]
+  (if-let [^String type (content-type request)]
+    (.startsWith type "application/x-www-form-urlencoded")))
+
+(defn assoc-query-params
+  "Parse and assoc parameters from the query string with the request."
+  [request encoding]
+  (merge-with merge request
+              (if-let [query-string (:query-string request)]
+                (let [params (parse-params query-string encoding)]
+                  {:query-params params, :params params})
+                {:query-params {}, :params {}})))
+
+(defn assoc-form-params
+  "Parse and assoc parameters from the request body with the request."
+  [request encoding]
+  (merge-with merge request
+              (if-let [body (and (urlencoded-form? request) (:body request))]
+                (let [params (parse-params (slurp body :encoding encoding) encoding)]
+                  {:form-params params, :params params})
+                {:form-params {}, :params {}})))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Public
+
+(defn params-request
+  "Adds parameters from the query string and the request body to the request
+  map. See: wrap-params."
+  {:arglists '([request] [request options])}
+  [request & [opts]]
+  (let [encoding (or (:encoding opts)
+                     (req/character-encoding request)
+                     "UTF-8")
+        request  (if (:form-params request)
+                   request
+                   (assoc-form-params request encoding))]
+    (if (:query-params request)
+      request
+      (assoc-query-params request encoding))))
--- a/src/clj/puppetlabs/puppetserver/ring/middleware/params.clj
+++ b/src/clj/puppetlabs/puppetserver/ring/middleware/params.clj
@@ -5,6 +5,27 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; COPY OF RELEVANT FUNCTIONS FROM UPSTREAM ring.middleware.params LIBRARY
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;
+;; This namespace is basically just here to provide an implementation of
+;; the `params-request` middleware function that supports a String representation
+;; of the body of a request.  The upstream library requires the body to be of
+;; a type that is compatible with Clojure's IOFactory, which forces us to read
+;; the request body into memory twice for requests that we have to pass down
+;; into the JRuby layer.  (Technically, the Ring specification states that the
+;; body must be an InputStream, so the maintainer of the upstream library was
+;; reluctant to accept any sort of upstream PR to work around this issue.)
+;;
+;; All of this code is copied from the upstream library, and there is just
+;; one very slight modification (see comment in `assoc-form-params` function)
+;; that allows us to avoid reading the body into memory twice.
+;;
+;; In the future, if we can handle the query parameter parsing strictly on the
+;; Clojure side and remove that code from the Ruby side, we should be able to
+;; get rid of this.  That will be much easier to consider doing once we're able
+;; to get rid of the Rack/Webrick support.
+;;
+;; If that happens, we should delete this namespace :)
+;;
 
 (defn parse-params [params encoding]
   (let [params (codec/form-decode params encoding)]
@@ -13,7 +34,11 @@
 (defn content-type
   "Return the content-type of the request, or nil if no content-type is set."
   [request]
-  (if-let [type (get-in request [:headers "content-type"])]
+  ;; NOTE: in the latest version of ring-core, they only look in
+  ;;  the headers map for the content type.  They no longer fall
+  ;;  back to looking for it in the main request map.
+  (if-let [type (or (get-in request [:headers "content-type"])
+                    (get request :content-type))]
     (second (re-find #"^(.*?)(?:;|$)" type))))
 
 (defn urlencoded-form?
@@ -36,7 +61,16 @@
   [request encoding]
   (merge-with merge request
               (if-let [body (and (urlencoded-form? request) (:body request))]
-                (let [params (parse-params (slurp body :encoding encoding) encoding)]
+                (let [params (parse-params
+                               ;; NOTE: this is the main difference between our
+                               ;;  copy of this code and the upstream version:
+                               ;;  the upstream always does a slurp here, while
+                               ;;  we only do a slurp if the body is not already
+                               ;;  a string.
+                               (if (string? body)
+                                 body
+                                 (slurp body :encoding encoding))
+                               encoding)]
                   {:form-params params, :params params})
                 {:form-params {}, :params {}})))
 
--- a/src/clj/puppetlabs/services/request_handler/request_handler_core.clj
+++ b/src/clj/puppetlabs/services/request_handler/request_handler_core.clj
@@ -6,10 +6,10 @@
             [clojure.string :as string]
             [puppetlabs.kitchensink.core :as ks]
             [puppetlabs.ssl-utils.core :as ssl-utils]
-            [ring.middleware.params :as ring-params]
             [ring.util.codec :as ring-codec]
             [ring.util.response :as ring-response]
             [slingshot.slingshot :as sling]
+            [puppetlabs.puppetserver.ring.middleware.params :as pl-ring-params]
             [puppetlabs.services.jruby.jruby-puppet-service :as jruby]))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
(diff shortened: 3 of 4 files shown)


jeffmccune:
In an ideal world the request body would never exist entirely in-memory and would instead be streamed through the whole system to conserve memory.  Why do we need to read the request body in the puppetserver layer, or stated another way, is it possible to avoid reading the request body and instead pass the file descriptor down into JRuby so it can be streamed there?

jeffmccune:
Also, it's not clear to me how the body is being read into a string twice.  Could you help me understand that part?

As it relates to the Puppet side of things, I think that portion of HTTP handling is well-behaved, at least in the sense that puppet [is accepting an IO stream and then yielding the body chunk by chunk](https://github.com/puppetlabs/puppet/blob/961f8424ab64527bb34454960547da2ec254756e/lib/puppet/network/http/rack/rest.rb#L14-L22).  This opens the possibility of true streaming, e.g. huge JSON objects without actually having to read the entire JSON string into memory at once.  It's worth noting the [#body method](https://github.com/puppetlabs/puppet/blob/961f8424ab64527bb34454960547da2ec254756e/lib/puppet/network/http/rack/rest.rb#L86-L88) appears to be poorly-behaved in that it reads everything into memory as well, which is less than ideal, but we do at least have the option of using the {{#each}} method instead of {{#body}} when we're dealing with things like huge JSON reports that could be streamed.

As this PR stands, the slurp call eliminates the possibility of end to end streaming, so I'm trying to understand how this moves us in the direction of a smaller memory footprint.

cprice404:
@jeffmccune prior to this PR, there were two slurps instead of one.  So that's the main way that it moves us in the direction of a smaller memory footprint...

There was one here:

https://github.com/puppetlabs/puppet-server/blob/e7fb6437b1e5c87643b9714d1196298a4b00fca6/src/clj/puppetlabs/services/request_handler/request_handler_core.clj#L86-L87

And then we created a `StringReader` wrapper around the resulting String:

https://github.com/puppetlabs/puppet-server/blob/e7fb6437b1e5c87643b9714d1196298a4b00fca6/src/clj/puppetlabs/services/request_handler/request_handler_core.clj#L109

The reason for that was that the ring middleware we were using to parse the request, here:

https://github.com/puppetlabs/puppet-server/blob/e7fb6437b1e5c87643b9714d1196298a4b00fca6/src/clj/puppetlabs/services/request_handler/request_handler_core.clj#L118

had an unconditional call to `slurp` inside of it.  So we were calling `slurp` twice, and creating two unique String objects in memory that each contained a full copy of the request.

This PR takes us one step in the right direction (IMO) by eliminating one of the two slurps.

I agree with the eventual goal of streaming the body through the whole system when appropriate, but it's really difficult to do that right now because there are three different code paths that can be exercised for processing the request body and initializing query/form parameters on our way to eventually ending up with an instance of `Puppet::Network::HTTP::Request` that we can pass around through the system: there's the Puppet Server code path, the Webrick code path, and the Rack code path.

When we start working on Puppet 5, we'll have the ability to eliminate two of those code paths, and then it will be much easier to reason about how a request flows through the system.  At that point we should absolutely be looking for opportunities to avoid the `slurp` entirely for requests where we can achieve that... so I totally agree with where you're coming from, but wasn't intending to try to solve the whole issue in one go with this PR.

cprice404:
Also, for reference, here's the upstream PR where I was trying to find a way to solve this on the ring side, so we wouldn't need to duplicate the code:

https://github.com/ring-clojure/ring/pull/193

I think his concerns about this being too much of an edge case to warrant acceptance of an upstream PR were pretty well-founded; I found myself having a hard time justifying what I was trying to do as the conversation progressed.

Another factor to note related to this PR: updating to the latest versions of compojure/bidi requires updating to a newer version of ring.  And in the newer versions of ring, there are some changes in how they support accessing the content-type for a request:

https://github.com/ring-clojure/ring/commit/bdffd464d2e803ae73707e46eb434e6b07ef0d21

So, we were going to have to either duplicate some of the ring code or make a bunch of changes around our code base either way.  Seemed like the right time to go ahead and eliminate the extra memory usage and kill two birds with one stone.

jeffmccune:
@cprice404 Thanks for the great information, I agree eliminating one full `slurp` is an improvement.  From a memory footprint standpoint I think this will have a significant improvement, especially since we're making a third full-copy of the body in the Ruby layer as well.

Just one last question, for my understanding a no other reason.  Could you help me understand the scenarios we need to pull parameters out of the body rather than headers or URL query parameters?  I'm trying to get a sense of how much existing behavior would need to change to get rid of all the full-copy-reads in the middleware.

This question is basically about focusing in a bit to a bit more detail on what requests we know about that trigger the behavior weavejester asked about:

> The params middleware only consumes the body if the body contains url-encoded parameters. So if I understand your use-case correctly, you sometimes want to parse a url-encoded body, and you sometimes want to keep it as an unparsed string. May I ask why?

cprice404:
@jeffmccune it'd be for any POST requests that have a content-type of url-form-encoded.  My memory is rusty on this but I think that the agent has some code where, for normal indirector HTTP requests, it does a check to see if the total length of the URL with all query params would exceed some character length, and if so, then it switches from a GET to a POST.  So I think there are multiple endpoints where we have to be able to get form data from either the query params or the body, and then put them into a single map before we create an instance of Puppet::Network::HTTP::Request to pass around through the system.

To address another comment that you made: I don't *think* we are actually doing a "third" read of the body in the Ruby layer.  There is a read in that Rack code that you linked to, but we don't call into that from Puppet Server.  We just create an instance of P::N::H::Request, and if my memory serves me correctly, Puppet expects the 'body' of that object to already be a String... and that's what we're giving it.  So I think we were just keeping two copies of the request bodies in memory before, not three... and I think this gets us down to one (thought it'd still be better, in the future, if we could stream it rather than read it all into memory, for endpoints where that is a possibility).

And I could be wrong about that, I'm just reciting from memory at the moment :)

project.clj
```diff
@@ -27,7 +27,9 @@
                  ;; of its jar; please read the detailed notes above the
                  ;; 'uberjar-exclusions' example toward the end of this file.
                  [org.jruby/jruby-stdlib "1.7.19"]
-                 [clj-time "0.5.1" :exclusions [joda-time]]
+                 [joda-time "2.5"]
+                 [clj-time "0.6.0"]
+                 [ring/ring-core "1.3.2"]
```
camlow325: Did we just move up to this because it's the latest?  Wasn't sure if there was something else we knew that we needed from later versions for these changes.

project.clj
```diff
@@ -37,7 +39,6 @@
                  [commons-codec "1.9"]
                  [clj-yaml "0.4.0" :exclusions [org.yaml/snakeyaml]]
                  [slingshot "0.10.3"]
-                 [ring/ring-codec "1.0.0"]
```
camlow325: So we get what we needed from `ring-codec` from `ring-core` now, I suppose?

src/clj/puppetlabs/puppetserver/ring/middleware/params.clj
```diff
@@ -0,0 +1,93 @@
...
+  (let [params (codec/form-decode params encoding)]
+    (if (map? params) params {})))
+
+(defn content-type
+  "Return the content-type of the request, or nil if no content-type is set."
+  [request]
+  ;; NOTE: in the latest version of ring-core, they only look in
+  ;;  the headers map for the content type.  They no longer fall
+  ;;  back to looking for it in the main request map.
+  (if-let [type (or (get-in request [:headers "content-type"])
```
camlow325: Not sure where all we'd be broken by not compensating for this elsewhere in Puppet Server but maybe it would be better for us to just "deal with it" elsewhere in our code?  I actually like what Ring did in eliminating the ambiguity around having this value represented in two different places in the map - which one should you honor if they happen to be different, etc?  One place I know this would affect is the hack that we did for the "file_bucket_file" binary content support.  To be compatible there, we might have to move the hack down from the `compojure/routes" into the master request handler, for example.

src/clj/puppetlabs/puppetserver/ring/middleware/params.clj
```diff
@@ -0,0 +1,93 @@
...
+;; All of this code is copied from the upstream library, and there is just
+;; one very slight modification (see comment in `assoc-form-params` function)
+;; that allows us to avoid reading the body into memory twice.
+;;
+;; In the future, if we can handle the query parameter parsing strictly on the
+;; Clojure side and remove that code from the Ruby side, we should be able to
+;; get rid of this.  That will be much easier to consider doing once we're able
+;; to get rid of the Rack/Webrick support.
+;;
+;; If that happens, we should delete this namespace :)
```
camlow325: Since a number of these functions come from different namespaces in the upstream `ring-core` library, might be good to leave more explicit references on each about where they come from - for future porting as may be needed, etc.

camlow325:
I'm definitely :+1: on the idea of eliminating the duplication of the request body string that currently is done for URL-encoded form posts.  I have a few ideas on the approach that I'd like to throw out for consideration...

First, some background:

My experience with the request handling, and this certainly could be short-sighted, has been that:

1) For the payload of a request that is not a URL encoded form POST, it must be of data type String and not IO at the point in the core Ruby Puppet code where it is handed off to the route for processing.

See the discussion thread on a previous PR, for example - https://github.com/puppetlabs/puppet-server/pull/345#discussion-diff-22479654

2) For URL encoded form POSTs, the expectation appears to be that the HTTP handler layer -- Rack, WEBrick, "Puppet Server" -- would be responsible for decoding the raw request body into "params" given to the underlying routes in Puppet.  Also, for the same URL encoded form POSTs, it appears that those underlying routes in Puppet only look at the "params" from the HTTP handler and not the "body" for the request data.

My recollection of why we went with the approach of ensuring that URL encoded form POST requests were translated by Puppet Server into both the decoded "params" and a String "body" is that this was the behavior I observed to be the case for WEBrick and Rack.  I'd then conservatively had suggested that we shouldn't deviate from that out of fear that there might be a downstream consumer of the data which might choose to read the request data from the "body" rather than from the "params".  I'm not sure that's really ever the case, though.

Given that we have a known/constrained set of endpoints that we allow through Puppet Server to core Ruby Puppet anyway, it seems like it wouldn't take too much work to determine if this is really a problem or not.  It might only be the "catalog" endpoint that this applies to.  Assuming this deviation wouldn't actually pose a problem, I think we could implement a simpler solution that wouldn't involve duplicating any of the Ring middleware code:

- Invoke `ring-params/params-request` before doing any slurping in Puppet Server code - so for URL encoded form posts, Ring middleware will slurp from the original input stream
- In Puppet Server code, case request content-type:
 - application/x-www-form-urlencoded
  -> Set body to nil (actually may not even matter what it's set to, but this seems reasonable)
 - nil "" "application/octet-stream"
  -> Set body as an InputStream (so it can be converted to a proper Ruby encoded string for binary transfer at the Puppet Server Ruby layer)
 - else
  -> Set body to slurped request body, considering encoding from the request

camlow325:
As mentioned above, avoiding slurping the request body entirely wouldn't be viable today without changed in core Ruby Puppet - at least for cases where routes in core Ruby Puppet code would need to be able to access the raw "body" directly.

Even for the case where Puppet Server is using Ring middleware to do the URL encoded form parameter decoding, it would be nice to get away from doing a full slurp of the request body before walking back over that to decompose the result into request parameters.  Maybe that would be best to pursue as an upstream contribution to Ring middleware if we ever decide to pursue it.

project.clj
cprice404: It's because the new bidi/compojure versions (which I am working on another PR for) require the newer ring... that is what led me down this rabbit hole in the first place.

project.clj
cprice404: I... think so?

src/clj/puppetlabs/puppetserver/ring/middleware/params.clj
cprice404: I didn't really think there was enough code here to warrant spending that much time on it, it's pretty easy to track down in the upstream... but if you feel strongly about it I can go back over it...

src/clj/puppetlabs/puppetserver/ring/middleware/params.clj
cprice404: I don't disagree, but I'd personally prefer to handle that as a separate ticket/PR.  This one gets us compatible with the new version of ring and eliminates a slurp, but there are definitely a lot of different kinds of iterative improvements we can make on it in the future.

My main goal for now is to unblock the bidi/compojure work, and it seemed like eliminating the second slurp in the process was worth a bit of extra effort.

cprice404:
My preference would be to handle the reconciliation of "when to slurp, when not to slurp", "do we need to maintain a copy of the body as a string?", "are there any routes in the puppet code that actually need to read the body?", etc., as part of the Puppet 5 efforts, when we know that we can make changes and decisions to the Puppet code that no longer force us to be concerned about differing / competing code paths / behaviors for three different web servers.

I fully agree with the thrust of the suggestions above, I just don't feel like now is the right time to invest the time it's going to take to make sure we come up with a robust solution for all of that.  That was not my intent with this PR.

I'd be happy to create or accept additional tickets for that stuff, but I don't think they can be prioritized over all of the other stuff we've already got scoped for 1.1/2.2.

So, as for this PR: if people feel like the temporary duplication of 50 lines of fairly simple code from ring-core is too high a price to pay for eliminating the extra slurp, as an incremental step towards a more holistic clean-up of our request handling, then I can close this, find a smaller temporary hack that makes us compatible with the newer version of ring, and unblocks the metrics work that I'm trying to get done.  Either way I would definitely expect us to improve upon the request handling in the future... it's just not the main focus of what I need to get done right now.

camlow325:
I wasn't so much concerned specifically about the way you went about the change for the slurp avoidance as at the prospect in general of needing to duplicate code from Ring.  But I suppose if you're going to have to do a fair amount of that anyway to preserve the legacy behavior for `content-type` because we don't want to spend the time finding / fixing all of the places in current code which would be affected by this - which I understand - then maybe my point is moot anyway.

I'm okay letting this come in as-is if others are.  I could then create follow-on tickets to cover:

1) Convert current code over to the way Ring handles `content-type` in `headers` now.

2) Assess whether or not a string body is really necessary for a URL encoded form post and, if not, discontinue setting it.  *I don't think this has to coincide with Puppet 5 even if it practically might work out that way anyway.*

3) Pending outcome 1 and 2, rip out the duplicated functions from Ring.

cprice404:
+1 to that plan (although I'm still totally willing to ditch this PR if people have other concerns about it).

As for whether the URL encoding stuff has to coincide with Puppet 5: the reason I'm proposing that is because we are already going to be spending a bunch of time in the relevant parts of the Ruby code base as part of our work for Puppet 5, and I think it would be wasteful to spend 2 different iterations going through that code, since it takes some time to build up a mental context around it when you haven't been in there for a while.  I think it would be much more efficient to group the work you're proposing with the other work that we already have planned and that we know will require someone to go into that part of the code.

src/clj/puppetlabs/puppetserver/ring/middleware/params.clj
```diff
@@ -0,0 +1,93 @@
...
+              (if-let [body (and (urlencoded-form? request) (:body request))]
+                (let [params (parse-params
+                               ;; NOTE: this is the main difference between our
+                               ;;  copy of this code and the upstream version:
+                               ;;  the upstream always does a slurp here, while
+                               ;;  we only do a slurp if the body is not already
+                               ;;  a string.
+                               (if (string? body)
+                                 body
+                                 (slurp body :encoding encoding))
```
KevinCorcoran: I don't understand why `body` would ever be a string here.  `(slurp my-string)` interprets `my-string` as a file path.

camlow325: If we go down this path in Puppet Server code, `body` will be a `string`.  We will have pre-slurped it from the `InputStream` handed down from the webserver for the request - see https://github.com/puppetlabs/puppet-server/blob/puppet-server-1.0.8/src/clj/puppetlabs/services/request_handler/request_handler_core.clj#L86.  I'm thinking the else clause is probably just defensive programming for a hypothetical future case where it could still be passed in as an `InputStream` - like maybe once we disprove the need to pre-slurp it for preservation as a string when passed down to core Ruby Puppet code.

KevinCorcoran: Ah, thanks for the explanation.

KevinCorcoran:
@cprice404 - I'll weigh in on this since it sounds like you're looking for some more opinions ...

I'd prefer to avoid duplicating the code from Ring and just leave the additional `slurp` in-place for now, since AFAIK it's never caused a problem, and since all parties seem to be in agreement that this is not the right long-term solution.  So, it just seems to be trading one kind of tech debt for another; and I think it might be easiest to address in the long-run if we don't complicate the codebase by pulling in the Ring code.

KevinCorcoran:
That said, it'll be okay with me if the consensus is to go the other way (and accept these changes).

camlow325:
I came away from this thinking that @cprice404 would probably have ended up duplicating at least some of the Ring code in the short term to cover backward compatibility around the `content-type` header anyway.  Given that, throwing the extra `slurp` avoidance in there seems like a reasonable, low enough turmoil thing to add on.  I don't know that we've had any complaints about the extra `slurp` to my knowledge but it can be a hefty chunk of RAM depending upon the request - several megabytes for catalog requests - so I appreciate the motivation.

camlow325:
Looks like this will need to be rebased if we're still going to bring it in.

KevinCorcoran:
@camlow325 Yeah, I certainly appreciate the motivation as well.  I've been waffling on this one, so I don't mind seeing it go either way, although my preference is still as stated above (at least, right now).  Which way do you lean?

camlow325:
@KevinCorcoran Yep, I've had the same internal debate that it sounds like you are.  :smile:  At this point, I'm leaning toward bringing the PR in as is stands and filing the follow-on "tech debt" tickets -- mentioned in a comment above -- to clean it up later when time / strategy permits.

cprice404:
My read of the comments above is that there is not super-strong objection to this path, so, given:

* It's only about 50 lines of code we're temporarily duplicating
* It's not very complex code
* I think the comment explains things reasonably well and it's not going to be hard to get rid of later
* It has the potential to reduce memory usage by over 2MB per agent run
* Memory usage is one of the more common questions / complaints that we field
* I don't feel confident that we're going to have bandwidth to implement the better, long-term solution before our 1.1/2.2 release...

I rebased :)

cprice404:
If it would help convince anyone I could spend some time with YourKit and/or Gatling to try to quantify the memory improvement, but I'd rather do that as a finishing-touches type of task when we get close to the release.

KevinCorcoran:
@camlow325 Chris and I just chatted about this PR.  He would like to get it merged ASAP and asked me to facilitate.  My understanding of the situation is ...
* 3 of us have weighed-in
* all of us are okay with either option
* I'm :-1: 
* You and Chris are :+1: 

So, given all of that, it sounds like we should go ahead and merge this.  Agree?

Also, you mention filing some tech-debt tickets  ... seems like a thing that should happen, if we merge this.

Ah, also, I see that @jeffmccune has commented above ... @jeffmccune do you have an opinion on :+1: / :-1: for this?

camlow325:
:+1: :shipit: - Will file those follow-up tickets and add another comment with references to the JIRA numbers after doing so.

camlow325:
I created just two follow-on tickets for this instead of the three I'd mentioned earlier.  Those are SERVER-589 (content-type changes) and SERVER-594 (string body for URL encoded requests).  Both tickets allude to the ability to rip out the duplicated Ring middleware code as a result of the work.  I figure we can wait and see how that plays out.

PR #499, opened 2015-04-15
(MAINT) Bump lein-ezbake dependency to 0.3.3

camlow325:
This commit bumps the puppet-server lein-ezbake dependency to 0.3.3,
needed to avoid Fedora 19 being used as a build target since that's now
been removed for our packaging jobs.


nwolfe:
:+1:

PR #501, opened 2015-04-15
(SERVER-157) Use the keylength setting from puppet.conf

nwolfe:
If none are provided, it will resort to the default puppet
value.

Note that if the keylength is not correct (too small or
too high) the underlying java security code will throw
an InvalidParameterSecurity message about RSA key size,
but not a puppet specific message.

Signed-off-by: Brice Figureau <brice-puppet@daysofwonder.com>

Conflicts:
	test/unit/puppetlabs/puppetserver/certificate_authority_test.clj

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -29,6 +29,7 @@
    :hostcert       schema/Str
    :hostprivkey    schema/Str
    :hostpubkey     schema/Str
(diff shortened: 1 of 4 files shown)


nwolfe:
@camlow325 Here's that cherry-pick we talked about.

@masterzen Just FYI...we probably wanted to re-target your PR against the stable branch before merging, but this works too :smile:

jeffmccune:
:+1:

camlow325:
:+1:

PR #505, opened 2015-04-17
(SERVER-391) Throw slingshot exception on timeout

dankreek:


src/clj/puppetlabs/services/master/master_service.clj
```diff
@@ -13,11 +13,12 @@
...
    (core/validate-memory-requirements!)
-   (let [path        (get-route this)
-         config      (get-config)
-         certname    (get-in config [:puppet-server :certname])
-         localcacert (get-in config [:puppet-server :localcacert])
-         settings    (ca/config->master-settings config)]
+   (let [path           (get-route this)
+         config         (get-config)
+         certname       (get-in config    [:puppet-server :certname])
+         localcacert    (get-in config    [:puppet-server :localcacert])
```
KevinCorcoran: extra spaces in the `get-in` args on these two lines

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
```diff
@@ -259,21 +260,35 @@
     (= (:type x)
        :puppetlabs.services.request-handler.request-handler-core/bad-request)))
 
+(defn jruby-timeout?
+  [x]
+  "Determine if the supplied slingshot message is for a JRuby borrow timeout."
+  (when (map? x)
+    (= (:type x)
+       ::jruby/jruby-timeout)))
```
KevinCorcoran: This doesn't seem like it should work, since `::jruby/jruby-timeout` here is equivalent to `::puppetlabs.services.request-handler.request-handler-core/jruby-timeout`, but since the `throw+` is in a different namespace,`

dankreek: Oh whoops, yeah that was a typo.

dankreek: So the `::` was initially a typo which, for some very mysterious reason, causes the test to pass. When I changed it to `:jruby/jruby-timeout` it failed, as expected. I'll change this to `:puppetlabs.services.jruby.jruby-puppet-service/jruby-timeout` to clear up the ambiguity. But this is super strange. Anyone have any ideas?

src/clj/puppetlabs/services/request_handler/request_handler_service.clj
```diff
@@ -2,19 +2,28 @@
   (:require [puppetlabs.trapperkeeper.core :as tk]
             [puppetlabs.services.protocols.request-handler :as handler]
             [puppetlabs.services.request-handler.request-handler-core :as core]
+            [puppetlabs.trapperkeeper.services :as tk-services]
             [puppetlabs.services.jruby.jruby-puppet-service :as jruby]
-            [puppetlabs.trapperkeeper.services :as tk-services]))
+            [clojure.tools.logging :as log]))
 
 (defn- handle-request
```
camlow325: This whole function should be removed now, right, since nothing should be calling it?

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
```diff
@@ -259,21 +260,36 @@
...
        :puppetlabs.services.request-handler.request-handler-core/bad-request)))
 
+(defn jruby-timeout?
+  [x]
+  "Determine if the supplied slingshot message is for a JRuby borrow timeout."
+  (when (map? x)
+    (= (:type x)
+       :puppetlabs.services.jruby.jruby-puppet-service/jruby-timeout)))
+
+
```
camlow325: Cosmetic - extra line?

src/clj/puppetlabs/services/request_handler/request_handler_service.clj
```diff
@@ -2,19 +2,28 @@
   (:require [puppetlabs.trapperkeeper.core :as tk]
             [puppetlabs.services.protocols.request-handler :as handler]
             [puppetlabs.services.request-handler.request-handler-core :as core]
+            [puppetlabs.trapperkeeper.services :as tk-services]
             [puppetlabs.services.jruby.jruby-puppet-service :as jruby]
-            [puppetlabs.trapperkeeper.services :as tk-services]))
+            [clojure.tools.logging :as log]))
```
camlow325: And this shouldn't be needed either if `handle-request` is removed from this namespace?

src/clj/puppetlabs/services/request_handler/request_handler_service.clj
```diff
@@ -2,19 +2,28 @@
...
+        (log/error message)
+        {:status  503
+         :body    message
+         :headers {"Content-Type"     "text/plain"
+                   "X-Puppet-version" puppet-version}}))))
 
 (tk/defservice request-handler-service
                handler/RequestHandlerService
-               [[:PuppetServerConfigService get-config]]
+               [[:PuppetServerConfigService get-config get-in-config]]
```
camlow325: Don't need to add this?

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
jeffmccune: Just to close the loop on this, @KevinCorcoran, @camlow325 and I were mystified for a bit how `::jruby/jruby-timeout` worked at all, we're pretty sure it's because `jruby` is aliased to the fully qualified namespace via `[puppetlabs.services.jruby.jruby-puppet-service :as jruby]`

KevinCorcoran: Yup, we learned that `::foo/bar` does a lookup for a namespace alias for `foo`, which is why `jruby` was working in that case.  Which actually seems like a reasonable syntax to me.

camlow325:
:+1: other than egregious Wolfe Principle™ violations.

dankreek:
All Wolfe Principle violations have been addressed.

camlow325:
:+1: @KevinCorcoran, you good with this?

KevinCorcoran:
Sorry y'all, I won't have time to look at this until next Thursday.  But you can feel free to merge it without waiting on me, I'm satisfied that we got to the bottom of the namespaced-keyword mystery.  :mag:

camlow325:
Well, here it is "next Thursday" and we still haven't merged this.  @KevinCorcoran - you still have time to take a quick pass at this and see if it looks reasonable to merge?

src/clj/puppetlabs/services/request_handler/request_handler_service.clj
```diff
@@ -2,14 +2,8 @@
...
             [puppetlabs.services.protocols.request-handler :as handler]
             [puppetlabs.services.request-handler.request-handler-core :as core]
-            [puppetlabs.services.jruby.jruby-puppet-service :as jruby]
             [puppetlabs.trapperkeeper.services :as tk-services]))
 
-(defn- handle-request
-  [request jruby-service config]
-  (jruby/with-jruby-puppet jruby-puppet jruby-service
-    (core/handle-request request jruby-puppet config)))
-
```
KevinCorcoran: Why did you make this change?  I try to keep interactions with services contained within "service" namespaces - which was the origin reason for getting the `jruby-puppet` instance here, instead of passing the JRuby service reference down into `handle-request`.

KevinCorcoran: I guess it's a convention/style thing, but I find it a bit odd to see service references being passed around into namespaces that otherwise have nothing to do with Trapperkeeper.  (e.g. `request-handler-core`).  Makes it a bit more onerous to test, etc.

camlow325: Yeah, @dankreek and I talked about that and waffled a bit.

I had found it a bit odd to be doing more seemingly domain specific logic like getting a jruby-puppet instance and possibly construct a ring response from the service namespace rather than core, which I kind of thought was where @cprice404 was on this as well.

I don't have a strong opinion on this, though - other than that I want to get this PR in soon so we can move on.

dankreek:
Yeah, in this case we either err on the side of putting a "service" thing in core code, or do code things in service code. This solution ended up being the least ugly, in my opinion.

src/clj/puppetlabs/services/request_handler/request_handler_service.clj
dankreek: Yeah, in this case we either err on the side of putting a "service" thing in core code, or do code things in service code. This solution ended up being the least ugly, in my opinion.

KevinCorcoran: I misunderstood what was going on here; this LGTM.

PR #506, opened 2015-04-22
(PE-8645) use comidi for routes

cprice404:
This PR ports our web routing library over from compojure to comidi.  This will allow us to introspect our web routes for things like metrics, middleware, documentation, etc. in the future.

### DIFF
--- a/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
+++ b/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
@@ -1,5 +1,6 @@
 (ns puppetlabs.services.ca.certificate-authority-core
-  (:import  [java.io InputStream])
+  (:import [java.io InputStream]
+           (clojure.lang IFn))
   (:require [puppetlabs.puppetserver.certificate-authority :as ca]
             [puppetlabs.puppetserver.ringutils :as ringutils]
             [puppetlabs.puppetserver.liberator-utils :as utils]
@@ -11,6 +12,7 @@
             [schema.core :as schema]
             [cheshire.core :as cheshire]
             [liberator.core :refer [defresource]]
+            ;[liberator.dev :as liberator-dev]
             [liberator.representation :as representation]
             [ring.util.response :as rr]))
 
@@ -253,8 +255,9 @@
          (handle-get-certificate-revocation-list ca-settings))))
 
 (schema/defn ^:always-validate
-  wrap-middleware
-  [handler puppet-version]
+  wrap-middleware :- IFn
+  [handler :- IFn
+   puppet-version :- schema/Str]
   (-> handler
     ;(liberator-dev/wrap-trace :header)           ; very useful for debugging!
     (ringutils/wrap-with-puppet-version-header puppet-version)
--- a/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
+++ b/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
@@ -3,16 +3,15 @@
   (:require [puppetlabs.puppetserver.certificate-authority :as ca]
             [puppetlabs.puppetserver.ringutils :as ringutils]
             [puppetlabs.puppetserver.liberator-utils :as utils]
+            [puppetlabs.comidi :as comidi :refer [GET ANY PUT]]
             [slingshot.slingshot :as sling]
             [clojure.tools.logging :as log]
             [clojure.java.io :as io]
             [clojure.string :as string]
             [schema.core :as schema]
             [cheshire.core :as cheshire]
-            [compojure.core :as compojure :refer [GET ANY PUT]]
             [liberator.core :refer [defresource]]
             [liberator.representation :as representation]
-            [liberator.dev :as liberator-dev]
             [ring.util.response :as rr]))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
@@ -53,7 +52,7 @@
       (rr/content-type "text/plain")))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
-;;; Compojure app
+;;; Web app
 
 (defn try-to-parse
   [body]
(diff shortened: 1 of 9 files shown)


fpringvaldsen:
This looks good to me! All the tests passed locally. I'm :+1: barring the other PRs getting merged and a subsequent rebase.

cprice404:
This is still blocked on #498.

cprice404:
This fails in travis because I have comidi set up to deploy to nexus, but tests should pass if you are on the VPN and pull down the source.  I'll OSS comidi on Monday if this PR gets reviewed and +1'd.  IMO probably not worth blocking the merge on that.

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
```diff
@@ -35,6 +35,12 @@
 (def localhost-cert
   (utils/pem->cert (test-pem-file "localhost-cert.pem")))
 
+(defn build-ring-handler
+  [settings puppet-version]
+  (-> (web-routes settings)
+       (comidi/routes->handler)
```
rlinehan: weird alignment?

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -3,16 +3,15 @@
...
             [slingshot.slingshot :as sling]
             [clojure.tools.logging :as log]
             [clojure.java.io :as io]
             [clojure.string :as string]
             [schema.core :as schema]
             [cheshire.core :as cheshire]
-            [compojure.core :as compojure :refer [GET ANY PUT]]
             [liberator.core :refer [defresource]]
             [liberator.representation :as representation]
-            [liberator.dev :as liberator-dev]
```
rlinehan: In another ns, this was left in but commented out... would be nice to be consistent about that?

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -236,29 +235,27 @@
...
-        (GET "/" []
-          (handle-get-certificate-request subject ca-settings))
-        (PUT "/" {body :body}
-          (handle-put-certificate-request! subject body ca-settings)))
-      (GET "/certificate_revocation_list/:ignored-node-name" []
-        (handle-get-certificate-revocation-list ca-settings))))
+  (comidi/context ["/" :environment]
+    (ANY ["/certificate_status/" :subject] [subject]
+         (certificate-status subject ca-settings))
+    (ANY ["/certificate_statuses/" :ignored-but-required] []
```
camlow325: Probably doesn't matter if we preserve the `do-not-use` reference here?  Can't remember why we had it there in the first place.  Maybe @KevinCorcoran remembers?

KevinCorcoran: No, we definitely don't need it.  I am not sure if there if Compojure requires it to be there - if not, it's only there because I didn't know it could be omitted.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -236,29 +235,27 @@
...
 (schema/defn ^:always-validate
-  build-ring-handler
-  [ca-settings :- ca/CaSettings
-   puppet-version :- schema/Str]
-  (-> (routes ca-settings)
-      ;(liberator-dev/wrap-trace :header)           ; very useful for debugging!
-      (ringutils/wrap-with-puppet-version-header puppet-version)
-      (ringutils/wrap-response-logging)))
+  wrap-middleware
+  [handler puppet-version]
```
camlow325: Useful to keep the schema types here since this is called from outside of this namespace?

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -106,11 +98,19 @@
...
+  "Creates all of the web routes for the master."
+  [request-handler]
+  (comidi/routes
+    (comidi/context "/v2.0"
+                     (v2_0-routes request-handler))
+    (comidi/context ["/" :environment]
+                     (legacy-routes request-handler))))
+
+(defn wrap-middleware
+  [handler puppet-version]
```
camlow325: Same question here about schema validation here as with the `wrap-middleware` function in `certificate_authority_core`.

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
```diff
@@ -20,7 +19,10 @@
      (log/info "CA Service adding a ring handler")
      (add-ring-handler
        this
-      (compojure/context path [] (core/build-ring-handler settings puppet-version))))
+       (-> (core/web-routes settings)
+           (#(comidi/context path %))
```
camlow325: Not a big deal but where these lines are repeated per service's ring-handler, maybe worth pulling this out into a helper function at some point - either common to Puppet Server or in comidi itself?

camlow325:
:+1: with just a few cosmetic comments, none of which absolutely have to be addressed.  I did a little functional testing of this with some puppet agent runs against a master using the comidi routes and everything seemed to work fine.

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
cprice404: I had that in `comidi` at one point (a fn called `context->handler`).  I was following compojure's convention for argument order, and for `context`, compojure puts the `path` arg first.  This means that there's no clean way to use the threading macro on it because all other middleware wrapping tends to expect the ring handler fn as the first arg :disappointed: .

I'd be OK with adding that back to comidi if people want it, but would probably rather deal with that as a separate PR.

cprice404:
I added a commit that addresses the formatting / schema issues from comments above.

Thanks for the review on that stuff!  Also, just to make sure everyone's on the same page: my number 1 priority for PR review here is to make sure everyone is on board with the switch to comidi in general.  You may need to look at the related PE PR (maybe run the PE server in the REPL and then run the new 'print-http-*' repl util fn) to get a sense for how bidi opens some doors for us.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -1,37 +1,37 @@
...
    excluding the CA-related endpoints, which are handled separately by the
    CA service."
   [request-handler]
-  (compojure/routes
-    (compojure/GET "/node/*" request
+  (comidi/routes
+    (comidi/GET ["/node/" :node] request
                    (request-handler request))
-    (compojure/GET "/facts/*" request
+    (comidi/GET ["/facts/" :node] request
```
rlinehan: One difference that's going to happen here and for the `/node` endpoint is that a request to e.g. `/production/node/foo/bar` with this code will return a 404 not found from puppet server, whereas a with the previous code it would match at puppet server's level, and then get passed down to the indirector code, which will return `Could not retrieve facts for foo/bar: invalid key`. I'm not sure that this is at all an issue, but it is a change in behavior that I wanted to point out.

rlinehan:
I played around with this in the REPL on pe-puppet-server-extensions, and I commented some there.  I'm generally +1 on this. I think that the syntax of this might take a little adjusting to (no more than using compojure did - I still have to look at compojure's destructuring docs every time I try to do anything with it, because I always end up confused), but seems generally pretty usable. It seems like it buys us a lot as far as being able to introspect our routes.

My only worry is what changes are we going to be inadvertently introducing by switching to this for Puppet Server 1.x/Puppet 3.x, and are we okay with potentially having some small changes, especially since our acceptance test isn't super great? It seems like any such changes are going to be pretty small and probably won't affect our users much, but there does seem to be some possibility of us introducing breaking changes with this.

camlow325:
Since this one's going into FOSS Puppet Server, would also be nice to have a SERVER ticket to link this to instead of a PE one so we have something to point the general public at for these changes.

src/clj/puppetlabs/services/master/master_core.clj
camlow325: This is a good point.  Not sure if this particular difference would be a big deal, but I think our overall approach to date has been to be as dumb about routing as possible for the master routes at the Puppet Server layer - letting core Ruby Puppet do the error handling.  Might be safer/better to just do the "*" for 'node', 'facts', and 'catalog' like is done for the other endpoints and not worry about destructuring the `:node` param that we don't appear to be using here for any of these anyway?  Hmmm ... although maybe your intent was to do it this way so that some of the metrics service stuff can make use of the destructured parameters?

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
camlow325: I'm okay leaving this as-is, not a big deal.

src/clj/puppetlabs/services/master/master_core.clj
cprice404: That's fair.  I think I mostly did it this way because the 'route id' strings that comidi generates are less disgusting if you don't have the regex crap in them, but that's probably not as compelling as the point about the behavior changing.  Will fix.

cprice404:
Moved the ticket to SERVER-598.  Will rebase and update commit messages.

cprice404:
Rebased, changed the ticket number, change the routes back as per @rlinehan 's comment.

camlow325:
OpenJDK cell failed Travis on the first run.  Looks like it might have been a transient / infrastructure thing.  Just rekicked it...

cprice404:
Anything else we are waiting on on this?  It is blocking a lot of other things for me...

rlinehan:
:shipit:

camlow325:
:rocket:

PR #507, opened 2015-04-24
(docs) Add mention of missing /status endpoint

ahpook:
Since a user spent some significant time tracking down this problem,
it seems worth preventing future frustration.


camlow325:
:+1: Looks reasonable to me.  @nfagerlund or @jbondpdx, you good with this?

nfagerlund:
:+1:  from me as well.

PR #508, opened 2015-04-27
(SERVER-545) Add ability to test Puppet 3.x agents

jeffmccune:
Without this patch there is no location to add acceptance tests that exercise
Puppet 3.x agents against Puppet Server 2.0.  This patch addresses the problem
by creating a new suite that maps to a new Jenkins job.  The pre_suite for this
new job will install Puppet 3.x legacy agents specified by the
PUPPET_LEGACY_VERSION environment variable.

### DIFF
--- a/acceptance/nodes/redhat7-64m-64a.yaml
+++ b/acceptance/nodes/redhat7-64m-64a.yaml
@@ -1,0 +1,30 @@
+---
+HOSTS:
+  redhat7-64-1:
+    roles:
+    - agent
+    - master
+    hypervisor: vcloud
+    pe_dir: 
+    pe_ver: 
+    pe_upgrade_dir: 
+    pe_upgrade_ver: 
+    platform: el-7-x86_64
+    template: Delivery/Quality Assurance/Templates/vCloud/redhat-7-x86_64
+  redhat7-64-2:
+    roles:
+    - agent
+    hypervisor: vcloud
+    pe_dir: 
+    pe_ver: 
+    pe_upgrade_dir: 
+    pe_upgrade_ver: 
+    platform: el-7-x86_64
+    template: Delivery/Quality Assurance/Templates/vCloud/redhat-7-x86_64
+CONFIG:
+  nfs_server: none
+  consoleport: 443
+  datastore: instance0
+  folder: Delivery/Quality Assurance/Enterprise/Dynamic
+  resourcepool: delivery/Quality Assurance/Enterprise/Dynamic
+  pooling_api: http://vcloud.delivery.puppetlabs.net/
--- a/acceptance/suites/pre_suite/puppet3_compat/70_install_puppet.rb
+++ b/acceptance/suites/pre_suite/puppet3_compat/70_install_puppet.rb
@@ -1,0 +1,22 @@
+# Agent running on the master is current, not legacy.
+legacy_agents = agents.reject { |agent| agent == master }
+
+step "Install Legacy Puppet Agents."
+legacy_agents.each do |host|
+  platform = host.platform
+
+  puppet_version = ENV['PUPPET_LEGACY_VERSION']
+  if not puppet_version
+    fail "PUPPET_LEGACY_VERSION is not set, e.g. '3.7.5'"
+  end
+  install_package host, 'puppet', puppet_version
+end
+
+step "Install Puppet Server."
+make_env = {
+  "prefix"  => "/usr",
+  "confdir" => "/etc/",
+  "rundir"  => "/var/run/puppetserver",
+  "initdir" => "/etc/init.d",
+}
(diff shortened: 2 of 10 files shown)


dev/scripts/server545_puppet3_compaitiblity.sh
```diff
@@ -0,0 +1,26 @@
+#! /bin/bash
```
nwolfe: s/compaitiblity/compatibility in filename

dev/scripts/server545_puppet3_compaitiblity.sh
```diff
@@ -0,0 +1,26 @@
...
+  # --post-suite acceptance/suites/post_suite_backwards_compatibility \
+bundle exec beaker \
+  --debug \
+  --root-keys \
+  --no-color \
+  --repo-proxy \
+  --preserve-hosts never \
+  --type aio \
+  --config acceptance/nodes/redhat7-64m-64a.yaml \
+  --pre-suite acceptance/suites/pre_suite/backwards_compatibility \
```
nwolfe: Old directory name

dev/scripts/server545_puppet3_compaitiblity.sh
```diff
@@ -0,0 +1,26 @@
...
+  --tests acceptance/suites/backwards_compatibility/010_hello_world \
```
nwolfe: and here

jeffmccune:
@nwolfe Thanks for catching those, I addressed the misaligned directories and misspelled file name in bbf0ac2

nwolfe:
@jeffmccune Should this be targeted at the 2.1.x branch instead of master? (noidea)

camlow325:
Good question.  I'd say we go ahead and target at 2.1.x and we'll merge back up to master later - since it looks like that's the direction we're going to end up following.

acceptance/nodes/redhat7-64m-64a.yaml
```diff
@@ -0,0 +1,30 @@
+---
```
camlow325: Should this file live under `.../acceptance/config/nodes` instead of this directory?  That's the only other one like this I found in the puppet-server repo.

acceptance/suites/pre_suite/puppet3_compat/70_install_puppet.rb
```diff
@@ -0,0 +1,22 @@
...
+    fail "PUPPET_LEGACY_VERSION is not set, e.g. '3.7.5'"
+  end
+  install_package host, 'puppet', puppet_version
+end
+
+step "Install Puppet Server."
+make_env = {
+  "prefix"  => "/usr",
+  "confdir" => "/etc/",
+  "rundir"  => "/var/run/puppetserver",
```
camlow325: Is this for the `rundir` on the master?  If so, should it be `/var/run/puppetlabs/puppetserver`?  Although it looks like this is the same for the `.../pre_suite/foss/70_install_puppet.rb`, so maybe it doesn't matter?

acceptance/nodes/redhat7-64m-64a.yaml
jeffmccune: Yes, I was thinking the same thing once I noticed we have existing node configurations.  I'll update the PR with the correct location.

acceptance/suites/pre_suite/puppet3_compat/70_install_puppet.rb
jeffmccune: I'm not entirely sure what this is for actually, it was cargo-culted from the FOSS pre-suite.

I think we can eliminate the make_env actually, so I'll try that now.

dev/scripts/server545_puppet3_compatibility.sh
```diff
@@ -0,0 +1,26 @@
...
+
+  # --post-suite acceptance/suites/post_suite_backwards_compatibility \
+bundle exec beaker \
+  --debug \
+  --root-keys \
+  --no-color \
+  --repo-proxy \
+  --preserve-hosts never \
+  --type aio \
+  --config acceptance/nodes/redhat7-64m-64a.yaml \
```
camlow325: Suppose this would need to change also if we move the `redhat7-64m-64a.yaml` file down to the `config` subdirectory?

jeffmccune:
Closing in favor of https://github.com/puppetlabs/puppet-server/pull/515 which should also address all of the comments made on this PR.

dev/scripts/server545_puppet3_compatibility.sh
jeffmccune: Yes, I've updated this location in https://github.com/puppetlabs/puppet-server/pull/515

PR #509, opened 2015-04-27
(maint) Pin 2.1.x against puppet-agent 1.0.0

jeffmccune:
Without this patch the smoke tests are failing in the 2.1.x branch, presumably
because the pinned version of the puppet-agent dependency has been pruned from
the builds server.  This patch addresses the problem by pinning to 1.0.0 which
is a released version and should not be pruned.


joshcooper:
:+1:

camlow325:
:+1:

PR #512, opened 2015-04-27
(SERVER-541) Bump lein-ezbake dependency to 0.2.9

camlow325:
This commit bumps the puppet-server lein-ezbake dependency to 0.2.9.
The primary change the new release has that Puppet Server is interested
in is quiet debug output by default during package installation.


dankreek:
:+1:

PR #513, opened 2015-04-28
(SERVER-571) Add puppet-agent libdir to ruby-load-path

jeffmccune:
Without this patch Hiera backend plugins distributed as modules are not
functioning as expected in Puppet Server 2.0.  The expected behavior is that
the plugin is usable in puppetserver after a successful agent run that has
synchronized plugins.

This patch addresses the problem by adding the
`/opt/puppetlabs/puppet/cache/lib` path to the ruby-load-path setting
which lines up with the [Puppet Installation Layout][layout].

[layout]: http://goo.gl/BWyXpo

### DIFF
--- a/acceptance/suites/tests/020_regression_tests/SERVER-571_hiera_backends_in_modules.rb
+++ b/acceptance/suites/tests/020_regression_tests/SERVER-571_hiera_backends_in_modules.rb
@@ -1,0 +1,53 @@
+test_name "(SERVER-571) Test hiera backend plugins located in modules"
+
+# The way this "works" is that the user installs a module on the server, the
+# agent run synchronizes the module lib/ plugins to the [Puppet Installation
+# Layout][layout] `libdir` path of /opt/puppetlabs/puppet/cache/lib  The
+# master then loads the synchronized plugin file using a ruby `require` which
+# searches the $LOAD_PATH
+#
+# [layout]: http://goo.gl/BWyXpo
+
+this_file = __FILE__
+# The local directory where resources are located, e.g. an environment directory
+rsrc_dir = File.expand_path('../../../../resources', this_file)
+
+# The target directory to place whole environments.
+# e.g. /etc/puppetlabs/code/environments
+env_path = on(master, puppet("config print environmentpath")).stdout.chomp
+target_dir = env_path.split(master['pathseparator']).first
+
+scp_options = {}
+step "Install hiera_test module in server571 environment on the master"
+Dir.chdir("#{rsrc_dir}/environments") do
+  # Create /etc/puppetlabs/code/environments/server571
+  # so that agents can run against this environment to sync plugins
+  master.do_scp_to "server571", target_dir, scp_options
+end
+
+# Save the hiera configuration
+step "Save hiera_config file to restore in teardown"
+hiera_config = on(master, puppet('config print hiera_config')).stdout.chomp
+on(master, "cp -p #{hiera_config} #{hiera_config}.bak")
+
+step "Write hiera configuration that uses the backend from the module"
+Dir.chdir("#{rsrc_dir}/files/server571") do
+  master.do_scp_to "hiera.yaml", hiera_config, scp_options
+end
+
+teardown do
+  on(master, "mv -f #{hiera_config}.bak #{hiera_config}")
+end
+
+with_puppet_running_on(master, {}) do
+  step "Run puppet agent on the master to pluginsync the hiera backend"
+  agent_run_cmd = "agent --test --server #{master} --environment=server571"
+  on(master, puppet(agent_run_cmd), :acceptable_exit_codes => [0,2])
+
+  # Assert the hiera backend returns a value
+  step "Verify data from the hiera backend is in the compiled catalog"
+  on(master, puppet(agent_run_cmd), :acceptable_exit_codes => [0,2]) do |result|
+    assert_match /4EB0033A-FC4E-4D7D-9453-5A9938B6FF61/, result.stdout,
+      "The agent output should contain the value from the Hiera backend"
+  end
+end
--- a/resources/ext/config/conf.d/puppetserver.conf
+++ b/resources/ext/config/conf.d/puppetserver.conf
@@ -2,7 +2,7 @@
 jruby-puppet: {
     # Where the puppet-agent dependency places puppet, facter, etc...
     # Puppet server expects to load Puppet from this location
-    ruby-load-path: [/opt/puppetlabs/puppet/lib/ruby/vendor_ruby]
+    ruby-load-path: [/opt/puppetlabs/puppet/lib/ruby/vendor_ruby, /opt/puppetlabs/puppet/cache/lib]
 
     # This setting determines where JRuby will look for gems.  It is also
     # used by the `puppetserver gem` command line tool.
--- a/acceptance/resources/environments/server571/manifests/site.pp
+++ b/acceptance/resources/environments/server571/manifests/site.pp
@@ -1,0 +1,3 @@
+node default {
+  include hiera_test
+}
--- a/acceptance/resources/environments/server571/modules/hiera_test/lib/hiera/backend/dummy_backend.rb
+++ b/acceptance/resources/environments/server571/modules/hiera_test/lib/hiera/backend/dummy_backend.rb
@@ -1,0 +1,11 @@
+Puppet.debug "Loaded dummy_backend! 8B06E7B6-8930-4CAB-B229-9A40A3F41AB8"
+class Hiera
+  module Backend
+   class Dummy_backend
+     def lookup(key, scope, order_override, resolution_type)
+       Puppet.debug "Lookup called! key=#{key} 8B06E7B6-8930-4CAB-B229-9A40A3F41AB8"
+       return "4EB0033A-FC4E-4D7D-9453-5A9938B6FF61"
+     end
+   end
+  end
+end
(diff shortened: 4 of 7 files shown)


cprice404:
+1 but @camlow325 and I had been discussing the possibility of putting this in for 2.1, in which case it will need to be re-targeted.

resources/ext/config/conf.d/puppetserver.conf
```diff
@@ -2,7 +2,7 @@
 jruby-puppet: {
     # Where the puppet-agent dependency places puppet, facter, etc...
     # Puppet server expects to load Puppet from this location
-    ruby-load-path: [/opt/puppetlabs/puppet/lib/ruby/vendor_ruby]
+    ruby-load-path: [/opt/puppetlabs/puppet/lib/ruby/vendor_ruby, /opt/puppetlabs/puppet/cache/lib]
```
joshcooper: This concerns me, because it means the environment that the agent (on the master) pluginsyncs & applies the catalog from, will affect how the master compiles catalogs for all other agents potentially in different environments.

cprice404: @joshcooper is the libdir not included in the load path for the master under Passenger?

camlow325: @joshcooper I understand that this is distasteful.  Is it a new issue with Puppet 4, though, or did we have this problem in the past as well?  Do you have an alternative suggestion for how hiera backends in modules could be loaded by the master if we don't use the pluginsync / agent path as has apparently been done in the past?

acceptance/suites/tests/020_regression_tests/SERVER-571_hiera_backends_in_modules.rb
```diff
@@ -0,0 +1,53 @@
...
+hiera_config = on(master, puppet('config print hiera_config')).stdout.chomp
+on(master, "cp -p #{hiera_config} #{hiera_config}.bak")
+
+step "Write hiera configuration that uses the backend from the module"
+Dir.chdir("#{rsrc_dir}/files/server571") do
+  master.do_scp_to "hiera.yaml", hiera_config, scp_options
+end
+
+teardown do
+  on(master, "mv -f #{hiera_config}.bak #{hiera_config}")
```
camlow325: Pardon my Beaker ignorance...

If the test fails before original hiera_config file is backed up, will this teardown step still run, fail, and the ultimate result of the test failure appear to be due to the "mv" failure as opposed to the initial failure?

resources/ext/config/conf.d/puppetserver.conf
camlow325: > @joshcooper is the libdir not included in the load path for the master under Passenger?

And, specifically, the agent libdir?

camlow325:
:+1: pending the outcome of the discussion with @joshcooper about the `ruby-load-path` libdir change.

I like the new test - might be good to try to get this into the core Ruby Puppet acceptance test suite at some point since I don't think there's anything Puppet Server-specific about it.

Also :+1: to @cprice404's suggestion that we target this at 2.1.x instead of master.

resources/ext/config/conf.d/puppetserver.conf
joshcooper: @camlow325 Prior to 4.0, the master (webrick, passenger, puppetserver) and agent shared the same libdir (`/var/lib/puppet/lib`). And yep, we did have this issue in the past (which is why I filed https://projects.puppetlabs.com/issues/18154). So I agree with you,  the current patch is no worse than it used to be.

joshcooper: @cprice404 good point about passenger. Prior to 4.0, we set `vardir` in [`config.ru`](https://github.com/puppetlabs/puppet/blob/3.7.5/ext/rack/config.ru#L18) so passenger and the agent shared the same vardir:

```
ARGV << "--vardir"  << "/var/lib/puppet"
```

But in 4.x, passenger is overriding `vardir` to `"/opt/puppetlabs/server/data/puppetmaster"` and will have the same problem as this.

joshcooper: @cprice404 @camlow325 given we're at puppet 4.0 and server 2.0, wouldn't now be the time to document the breaking change that we already implemented in SERVER-357? The master and agent no longer share libdir, so you can't use the agent to distribute hiera backends like this.

I think the technical issue preventing us from moving forward, is as @hlindberg mentioned, hiera loads its backends using require (since there is no puppet autoloader in that context). I think the way to fix that is for hiera to always use an absolute path to load the backend (like we do when loading helper code from a type, provider, report processor, etc, e.g.

```
require Pathname.new(__FILE__).dirname + "relative/path/to/backend"
```

Then we could just use r10k to distribute the module to the server.

jeffmccune: @joshcooper Thanks for following up on this.  @cprice404 raised an interesting question in the comments [SERVER-571](https://tickets.puppetlabs.com/browse/SERVER-571), "why isn't 1.0.8 affected?"

I assumed that puppetserver 1.0.8 would not suffer this issue because 1.0.x does not have the AIO pathing changes.  However, upon verification of that assumption I just found out 1.0.8 _is_ suffering from this issue.

Even more perplexing to me is that puppetserver 1.0.2 is _not_ experiencing this issue.

However, both 1.0.2 and 1.0.8 have the same ruby-load-path setting of `ruby-load-path: [/usr/share/ruby/vendor_ruby/]`.

I'll post more details in that ticket, but at this point I'm sufficient confused about the nature of this issue that I'm also concerned this pull request isn't the correct solution.

jeffmccune: @joshcooper Nevermind, the discrepancy I saw where 1.0.8 seemed to be affected was a result of me failing to restart the puppetserver service immediately after reconfiguring the hiera backend.

jeffmccune: @joshcooper Back to the discussion, I'm a bit confused at how switching to an absolute path in require statements would solve this issue.  What is the value of `__FILE__` in your example?

Are you thinking we'd return to a world where we load the code directly from the module directory structure, or are you thinking that the absolute path wouldn't be relative to `__FILE__` but rather relative to the agent's plugin sync directory?

Maybe something like hiera doing a `require "#{Puppet[:libdir]}/hiera/backends/#{backend_name}.rb"`?

jeffmccune: Not to create too many threads along the same conversation, and assuming Josh is thinking along the lines of constructing an absolute path to the agent's libdir...

Wouldn't this mean the _only_ way hiera could load backends would be if they're located in the agent plugin sync directory?  Do we have a sense of how end users distribute backend extensions, e.g. are backends located in Ruby gems supported?  If so, an absolute path solution might break support for that behavior, or similar behaviors.

acceptance/suites/tests/020_regression_tests/SERVER-571_hiera_backends_in_modules.rb
jeffmccune: I wasn't sure of the answer, so I inserted `on(master, 'false')` just before `on(master, "cp -p #{hiera_config} #{hiera_config}.bak")`, which simulates a failure.

In this situation, the teardown block doesn't cause a problem, presumably because it hasn't been evaluated yet.

```
Warning: acceptance/suites/tests/020_regression_tests/SERVER-571_hiera_backends_in_modules.rb errored in 1.16 seconds
      Test Suite: tests @ 2015-04-30 14:36:00 -0700

       Total Suite Time: 1.16 seconds
      Average Test Time: 1.16 seconds
              Attempted: 1
                 Passed: 0
                 Failed: 0
                Errored: 1
                Skipped: 0
                Pending: 0
                  Total: 1

      - Specific Test Case Status -

Failed Tests Cases:
Errored Tests Cases:
  Test Case acceptance/suites/tests/020_regression_tests/SERVER-571_hiera_backends_in_modules.rb reported: #<Beaker::Host::CommandFailure: Host 'h0lg8rdikjlazmx.delivery
.puppetlabs.net' exited with 1 running:
 false
Last 10 lines of output were:
>
Skipped Tests Cases:
Pending Tests Cases:


Failed: errored in TestSuite: report_and_raise_on_failure
#<RuntimeError: Failed while running the tests suite>
Begin acceptance/suites/post_suite/99_collect_data.rb
...
```

cprice404:
My +1 was premature.  :)  I hadn't understood the change in behavior between Puppet 3.x and Puppet 4.x, and now, having read @joshcooper 's comments, I'd prefer his proposal (consider this behavior to be a valid, breaking change) to the changes in this PR.  So I think I'm in favor of just closing this PR at the moment.

This thread is currently going on on both the Jira ticket and the PR, perhaps we could consolidate it on the Jira ticket since that will be easier to search for history in the future?

nwolfe:
@camlow325 Are we closing this? Just going through PRs and this hasn't seen activity in 15 days, and Chris is proposing to close it and perhaps continue the discussion on the ticket.

camlow325:
I feel like we're at an impasse on this based on the discussion continuing on in the JIRA ticket.  I'm going to reassign SERVER-571 to @ahpook and move it to "needs information" for now.  Assuming that takes some time to sort out, the best course may be to just close this PR for now.  @jeffmccune - you okay with that?

nwolfe:
@jeffmccune Closing this PR. Feel free to re-open if necessary.

PR #515, opened 2015-04-28
(SERVER-545) Add ability to test Puppet 3.x agents

jeffmccune:
Without this patch there is no location to add acceptance tests that exercise
Puppet 3.x agents against Puppet Server 2.0.  This patch addresses the problem
by creating a new suite that maps to a new Jenkins job.  The pre_suite for this
new job will install Puppet 3.x legacy agents specified by the
PUPPET_LEGACY_VERSION environment variable.

### DIFF
--- a/acceptance/config/beaker/jenkins/redhat7-64m-64a.cfg
+++ b/acceptance/config/beaker/jenkins/redhat7-64m-64a.cfg
@@ -1,0 +1,30 @@
+---
+HOSTS:
+  redhat7-64-1:
+    roles:
+    - agent
+    - master
+    hypervisor: vcloud
+    pe_dir: 
+    pe_ver: 
+    pe_upgrade_dir: 
+    pe_upgrade_ver: 
+    platform: el-7-x86_64
+    template: Delivery/Quality Assurance/Templates/vCloud/redhat-7-x86_64
+  redhat7-64-2:
+    roles:
+    - agent
+    hypervisor: vcloud
+    pe_dir: 
+    pe_ver: 
+    pe_upgrade_dir: 
+    pe_upgrade_ver: 
+    platform: el-7-x86_64
+    template: Delivery/Quality Assurance/Templates/vCloud/redhat-7-x86_64
+CONFIG:
+  nfs_server: none
+  consoleport: 443
+  datastore: instance0
+  folder: Delivery/Quality Assurance/Enterprise/Dynamic
+  resourcepool: delivery/Quality Assurance/Enterprise/Dynamic
+  pooling_api: http://vcloud.delivery.puppetlabs.net/
--- a/acceptance/suites/pre_suite/puppet3_compat/30_install_dev_repos.rb
+++ b/acceptance/suites/pre_suite/puppet3_compat/30_install_dev_repos.rb
@@ -1,0 +1,21 @@
+case test_config[:puppetserver_install_type]
+when :package
+  step "Setup Puppet Server dev repository on the master." do
+    package_build_version = ENV['PACKAGE_BUILD_VERSION']
+    if package_build_version
+      install_puppetlabs_dev_repo master, 'puppetserver', package_build_version
+    else
+      abort("Environment variable PACKAGE_BUILD_VERSION required for package installs!")
+    end
+  end
+
+  puppet_build_version = test_config[:puppet_build_version]
+  if puppet_build_version
+    confine_block :except, :platform => ['windows'] do
+      step "Setup Puppet dev repository on the master." do
+        install_puppetlabs_dev_repo master, 'puppet-agent', puppet_build_version
+      end
+    end
+  end
+end
+
--- a/acceptance/suites/pre_suite/puppet3_compat/70_install_puppet.rb
+++ b/acceptance/suites/pre_suite/puppet3_compat/70_install_puppet.rb
@@ -1,0 +1,16 @@
+# Agent running on the master is current, not legacy.
+legacy_agents = agents.reject { |agent| agent == master }
+
+step "Install Legacy Puppet Agents."
+legacy_agents.each do |host|
+  platform = host.platform
+
+  puppet_version = ENV['PUPPET_LEGACY_VERSION']
+  if not puppet_version
+    fail "PUPPET_LEGACY_VERSION is not set, e.g. '3.7.5'"
+  end
+  install_package host, 'puppet', puppet_version
+end
+
+step "Install Puppet Server."
+install_puppet_server master
--- a/acceptance/suites/puppet3_tests/010_hello_world/testtest.rb
+++ b/acceptance/suites/puppet3_tests/010_hello_world/testtest.rb
@@ -1,0 +1,14 @@
+test_name "Testing Master/Agent backwards compatibility"
+
+# Agent running on the master is current, not legacy.
+legacy_agents = agents.reject { |agent| agent == master }
+
+step "Check that legacy agents have Puppet 3.x installed"
+on(legacy_agents, puppet("--version")) do
+  assert(stdout.start_with? "3.", "puppet --version does not start with major version 3.")
+end
+
+step "Check that Puppet Server has Puppet 4.x installed"
+on(master, puppet("--version")) do
+  assert(stdout.start_with? "4.", "puppet --version does not start with major version 4.")
+end
--- a/dev/scripts/server545_puppet3_compatibility.sh
+++ b/dev/scripts/server545_puppet3_compatibility.sh
@@ -1,0 +1,26 @@
+#! /bin/bash
+
+# This script is expected to be run with a PWD of the root of the puppet-server
+# project.
+
+
+# Puppet Server package version (from http://builds.puppetlabs.lan/puppetserver/)
+export PACKAGE_BUILD_VERSION='2.1.0.SNAPSHOT.2015.04.17T0057'
+# Legacy means pre-aio (4.0) in this case
+export PUPPET_LEGACY_VERSION='3.7.5'
+
+  # --post-suite acceptance/suites/post_suite_backwards_compatibility \
+bundle exec beaker \
+  --debug \
+  --root-keys \
+  --no-color \
+  --repo-proxy \
+  --preserve-hosts never \
+  --type aio \
+  --config acceptance/config/beaker/jenkins/redhat7-64m-64a.yaml \
+  --pre-suite acceptance/suites/pre_suite/puppet3_compat \
+  --tests acceptance/suites/puppet3_tests \
+  --keyfile ~/.ssh/id_rsa-acceptance \
+  --helper acceptance/lib/helper.rb \
+  --options-file acceptance/config/beaker/options.rb \
+  --load-path acceptance/lib
--- a/acceptance/README.md
+++ b/acceptance/README.md
@@ -4,314 +4,137 @@ This directory is intended to manage acceptance testing for the Puppet Server.
 
 ## Acceptance Testing for Dummies
 
-This setup is intended for developers when running a VM on their local machine.
-This will not use Vagrant or a Rakefile, but instead a local pre-installed VM and the beaker CLI.
+The workflow described here is intended for developers who have access to the
+[vmpooler](http://vmpooler.delivery.puppetlabs.net), which currently requires
+[VPN Access](https://confluence.puppetlabs.com/display/HELP/VPN+access) and as
+such is only applicable to Puppet Labs employees.  This workflow is intended to
+help developers new to the puppet-server project.
+
+The two primary goals of this workflow are to enable the developer to modify
+the behavior of acceptance tests in as little time as possible while operating
+in a manner as similar to the production CI system as possible.
+
+If you would like to setup local VM's to develop and exercise acceptance tests,
+or would like to learn about additional options beyond the vmpooler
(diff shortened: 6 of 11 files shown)


acceptance/config/beaker/jenkins/redhat7-64m-64a.yaml
```diff
@@ -0,0 +1,30 @@
+---
```
camlow325: All of the other files in this directory have a ".cfg" extension and don't lead with the "---".  Not sure if it's more or less conventional for this to be a ".yaml" file instead of a ".cfg" file?

dev/scripts/server545_puppet3_compatibility.sh
```diff
@@ -0,0 +1,26 @@
...
+  --debug \
+  --root-keys \
+  --no-color \
+  --repo-proxy \
+  --preserve-hosts never \
+  --type aio \
+  --config acceptance/config/beaker/jenkins/redhat7-64m-64a.yaml \
+  --pre-suite acceptance/suites/pre_suite/puppet3_compat \
+  --tests acceptance/suites/puppet3_tests \
+  --keyfile ~/.ssh/id_rsa-acceptance \
```
nwolfe: How do we handle this elsewhere? Put the key in with the other dev files? Or does everybody just have this keyfile laying around? Or do we typically not put beaker scripts in VCS so we never end up needing to refer to this file?

camlow325:
:+1: This all seems okay to me - just the one cfg vs. yaml question.  Might be good to have @waynr could take a look at this if he has some time.

camlow325:
Could be as a follow on to this PR -- would be nice to have the `.../acceptance/README.md` updated with a blurb about anything unique to running this new set of acceptance tests.

dev/scripts/server545_puppet3_compatibility.sh
jeffmccune: We handle it poorly.  We really shouldn't be copying around an unprotected private key, but we do, sadly.

We should most definitely _not_ commit the private key to the git repo.  Doing so would mean anyone would have credentials to log into the CI cluster.

acceptance/config/beaker/jenkins/redhat7-64m-64a.yaml
jeffmccune: File extension shouldn't matter but I'll change it to *.cfg for consistency.  I'd like to leave the leading `---` as that's part of the YAML spec.

dev/scripts/server545_puppet3_compatibility.sh
nwolfe: So this file is expected to be there in Jenkins when/if this script is used?  And developers will just need to get this keyfile somehow if they want to run locally?

That's fine, just want to make sure I understand.

erikPrime: --keyfile defaults to ~/.ssh/id_rsa.  --root-keys copys all the public keys from puppetlabs/ssh-keys to the hosts it provisions.  This should let you use your own key, but I haven't tested it recently.

camlow325: > This should let you use your own key, but I haven't tested it recently.

Last time I tried, it still did - but that was months ago...

jeffmccune:
@camlow325 I went ahead and updated the README, but it was basically an overhaul.  I'm not sure if we actually want to remove all of the information about setting up a local VM cluster for acceptance testing, but the updated setup procedure seems like a big enough improvement in process that it warrants the overhaul.

Thoughts?

dev/scripts/server545_puppet3_compatibility.sh
jeffmccune: You have it correct Nate, the coordinator nodes that Jenkins uses to execute beaker processes already have the identity file on them.  Devs running locally need to get the private key "somehow" which has most commonly taken the form of asking someone who already copied it.

nwolfe:
> I'm not sure if we actually want to remove all of the information about setting up a local VM cluster for acceptance testing

I'd vote for keeping the README current and not worrying about historical workflows. Chances are if we ever want to set up a local VM cluster this document isn't going to reflect the current preferred approach (right?).

camlow325:
@jeffmccune I think the revised README with instructions is great for those who want to use the vmpooler.  I do think it would still be valuable to leave behind instructions on how to do beaker testing with your own local VM - although I'll admit I don't know if the instructions we have for that are still good at this point.  Would there be a clean way to preserve the vmpooler instructions separately from the "legacy" local VM instructions for now?

jeffmccune:
> I'd vote for keeping the README current and not worrying about historical workflows. Chances are if we ever want to set up a local VM cluster this document isn't going to reflect the current preferred approach (right?).

I'm of the same mind, until we run into some issue with the pooler workflow, or something better comes along, it seems like it's worthwhile to have one opinionated suggested workflow to help people get started and productive.

A lack of rapid iteration against snapshots seems like the only downside to the pooler workflow, but the price seems worth it to me at least given that the developer will always be starting with known-good-state which is identical to the state the production CI system starts with.

jeffmccune:
@camlow325 Sure, I could just move the previous information to another file.  I'll do that now.

jeffmccune:
@camlow325 @nwolfe should have all comments addressed now.

nwolfe:
:+1: 

@camlow325 @jb (button)

camlow325:
:+1: :shipit: 

@waynr - still would appreciate it if you could take a look at this if you have some time.  Could do a follow up PR to address any comments you may have...

dev/scripts/server545_puppet3_compatibility.sh
```diff
@@ -0,0 +1,26 @@
...
+
+  # --post-suite acceptance/suites/post_suite_backwards_compatibility \
+bundle exec beaker \
+  --debug \
+  --root-keys \
+  --no-color \
+  --repo-proxy \
+  --preserve-hosts never \
+  --type aio \
+  --config acceptance/config/beaker/jenkins/redhat7-64m-64a.yaml \
```
waynr: The static host config is confusing and unnecessary. For an example of a project that dynamically creates its beaker host configs based on command line parameters, please see:

https://github.com/puppetlabs/pe-razor-server/blob/3.8.x/acceptance/scripts/testrun_install-from-package.sh

acceptance/config/beaker/jenkins/redhat7-64m-64a.cfg
```diff
@@ -0,0 +1,30 @@
+---
+HOSTS:
+  redhat7-64-1:
+    roles:
+    - agent
+    - master
+    hypervisor: vcloud
```
waynr: The `vcloud` hypervisor is deprecated in favor of `vmpooler`.

acceptance/config/beaker/jenkins/redhat7-64m-64a.cfg
```diff
@@ -0,0 +1,30 @@
...
+    pe_upgrade_ver: 
+    platform: el-7-x86_64
+    template: Delivery/Quality Assurance/Templates/vCloud/redhat-7-x86_64
+CONFIG:
+  nfs_server: none
+  consoleport: 443
+  datastore: instance0
+  folder: Delivery/Quality Assurance/Enterprise/Dynamic
+  resourcepool: delivery/Quality Assurance/Enterprise/Dynamic
+  pooling_api: http://vcloud.delivery.puppetlabs.net/
```
waynr: With the new `vmpooler` hypervisor these last four attributes are unnecessary. Genconfig2 handles this properly.

acceptance/suites/puppet3_tests/010_hello_world/testtest.rb
```diff
@@ -0,0 +1,14 @@
+test_name "Testing Master/Agent backwards compatibility"
+
+# Agent running on the master is current, not legacy.
+legacy_agents = agents.reject { |agent| agent == master }
```
waynr: The nomenclature here is a little strange, since it's not the fact that the agent machine isn't also the master that makes it "legacy", but the fact that it has puppet 3.x installed. I imagine this might be confusing for future readers of this code.

acceptance/suites/pre_suite/puppet3_compat/70_install_puppet.rb
```diff
@@ -0,0 +1,16 @@
+# Agent running on the master is current, not legacy.
+legacy_agents = agents.reject { |agent| agent == master }
```
waynr: Previous comment about the variable name applies here.

dev/scripts/server545_puppet3_compatibility.sh
jeffmccune: The static configuration is necessary so long as the `genconfig2` command isn't normally available.  By "normal" I mean expressed as a non-special-case bundler dependency via the Gemfile.

@waynr I'm happy to switch this script over to dynamically creating the beaker configuration, but for that to happen we need to figure out a way eliminate the `GEM_SOURCE` special case behavior.  How do you recommend we eliminate this special-case? https://github.com/puppetlabs/pe-razor-server/blob/37ae577f789dc2e49101b9ce4d3b97b0c20efd20/acceptance/scripts/testrun_install-from-package.sh#L16-L17

waynr: @jeffmccune I wouldn't be opposed to releasing genconfig as its own publicly-available gem, maybe call it `beaker-genconfig`. I will bring this up with the QE team and let you know if we decide this is a viable option.

acceptance/config/beaker/jenkins/redhat7-64m-64a.cfg
jeffmccune: This value came from a recent run of `genconfig2` in version 0.12.1:

I think we're on the right path to get rid of these static config files, but perhaps `sqa-utils-gem` needs to be updated as well?

```
$ bundle exec genconfig2 redhat7-64m-64a
---
HOSTS:
  redhat7-64-1:
    roles:
    - agent
    - master
    hypervisor: vcloud
    pe_dir:
    pe_ver:
    pe_upgrade_dir:
    pe_upgrade_ver:
    platform: el-7-x86_64
    template: Delivery/Quality Assurance/Templates/vCloud/redhat-7-x86_64
  redhat7-64-2:
    roles:
    - agent
    hypervisor: vcloud
    pe_dir:
    pe_ver:
    pe_upgrade_dir:
    pe_upgrade_ver:
    platform: el-7-x86_64
    template: Delivery/Quality Assurance/Templates/vCloud/redhat-7-x86_64
CONFIG:
  nfs_server: none
  consoleport: 443
  datastore: instance0
  folder: Delivery/Quality Assurance/Enterprise/Dynamic
  resourcepool: delivery/Quality Assurance/Enterprise/Dynamic
  pooling_api: http://vcloud.delivery.puppetlabs.net/
```

acceptance/config/beaker/jenkins/redhat7-64m-64a.cfg
jeffmccune: Same comment as above, this file was generated from version 0.12.1 of sqa-utils which includes the last for attributes.

Are you sure genconfig2 is handling this as expected?  Maybe there's a version mis-match?  I see the most recent tag as 0.12.1 in git, and when I install the gem from http://rubygems.delivery.puppetlabs.net I get version 0.12.1 which produces the output shown above.

acceptance/suites/puppet3_tests/010_hello_world/testtest.rb
jeffmccune: I'm struggling to come up with better nomenclature.  Any suggestions?

acceptance/config/beaker/jenkins/redhat7-64m-64a.cfg
waynr: Apologies for the confusion, genconfig2 from 0.13 will produce the correct, non-deprecated output.

acceptance/suites/puppet3_tests/010_hello_world/testtest.rb
waynr: Maybe `standalone_agents`? `nonmaster_agents`?

nwolfe: `nonmaster_agents` :+1:

acceptance/suites/pre_suite/puppet3_compat/30_install_dev_repos.rb
```diff
@@ -0,0 +1,21 @@
...
+  step "Setup Puppet Server dev repository on the master." do
+    package_build_version = ENV['PACKAGE_BUILD_VERSION']
+    if package_build_version
+      install_puppetlabs_dev_repo master, 'puppetserver', package_build_version
+    else
+      abort("Environment variable PACKAGE_BUILD_VERSION required for package installs!")
+    end
+  end
+
+  puppet_build_version = test_config[:puppet_build_version]
```
waynr: I would recommend at some point settling on this style of environment variable use (https://github.com/puppetlabs/puppet-server/blob/master/acceptance/lib/helper.rb#L30) or just using ENV['varname'] wherever necessary. I don't have any opinion about which is better, except to say that if you are using the variable value in more than one place it is probably beneficial to extract from the environment in a central location like the helper file.

PR #517, opened 2015-04-29
(SERVER-545) Fix CI failure in post_suite log copy

jeffmccune:
Without this patch the post suite is failing in the new CI job created
as part of QENG-2227.  The post suite fails trying to copy the server
logs because the server has never started.  This patch addresses the
problem by doing a single agent run against the server, which was the
original behavior of the basic testtest file.


dankreek:
:+1:

PR #520, opened 2015-04-30
(SERVER-390) Acceptance test for foreground subcmd

jpinsonault:
The test starts up puppetserver with the `foreground` subcommand, kills it after 60s, and examines the output for some expected text.

I tested it with beaker like so:

```bash
export PACKAGE_BUILD_VERSION=2.0.0
export PUPPET_VERSION=1.0.0

bundle exec beaker \
  --debug \
  --root-keys \
  --repo-proxy \
  --preserve-hosts always \
  --type foss \
  --pre-suite acceptance/suites/pre_suite/foss \
  --config acceptance/config/beaker/jenkins/el-6-x86_64-mdca.cfg \
  --tests acceptance/suites/tests/010-puppetserver-cli/subcommand/foreground.rb \
  --helper acceptance/lib/helper.rb \
  --options-file acceptance/config/beaker/options.rb \
  --load-path acceptance/lib
```

### DIFF
--- a/acceptance/suites/tests/010-puppetserver-cli/subcommand/foreground.rb
+++ b/acceptance/suites/tests/010-puppetserver-cli/subcommand/foreground.rb
@@ -20,7 +20,7 @@ timeout_cmd = "timeout -s INT #{timout_length} #{foreground_cmd}"
 expected_messages = {
   /Initializing the JRuby service/ => "JRuby didn't initialize",
   /Starting web server/ => "Expected web server to start",
-  /Finished creating JRubyPuppet instance 4 of 4/ => "Expected to find 4 JRubyPuppet instances",
+  /Puppet Server has successfully started and is now ready to handle requests/ => "puppetserver never finished starting",
   /Finished shutdown sequence/ => "Test ended without puppetserver shutting down gracefully"
 }
 
--- a/acceptance/suites/tests/010-puppetserver-cli/subcommand/foreground.rb
+++ b/acceptance/suites/tests/010-puppetserver-cli/subcommand/foreground.rb
@@ -1,23 +1,15 @@
 test_name "Puppetserver 'foreground' subcommand tests."
 
-# --------------------------------------------------------------------------
-# This behavior needs to be baked into Beaker but is not yet. SERVER-79 tracks
-# the need to fix this once Beaker has been modified to make the paths to these
-# commands available.
-#
-if master.is_pe?
-  cli = "pe-puppetserver"
-else
-  cli = "puppetserver"
-end
-# --------------------------------------------------------------------------
+cli = "puppetserver"
+service = "puppetserver"
 
 # puppetserver seems to take about 45s to start up
 timout_length = "60s"
-foreground_cmd = "#{cli} foreground"
+foreground_cmd = "#{cli} foreground --debug"
 timeout_cmd = "timeout -s INT #{timout_length} #{foreground_cmd}"
 
 expected_messages = {
+  / DEBUG .*Debug logging enabled/ => "Debug logging isn't enabled",
   /Initializing the JRuby service/ => "JRuby didn't initialize",
   /Starting web server/ => "Expected web server to start",
   /Puppet Server has successfully started and is now ready to handle requests/ => "puppetserver never finished starting",
@@ -26,7 +18,7 @@ expected_messages = {
 
 # Start of test
 step "Stop puppetserver"
-on(master, puppet("resource service #{cli} ensure=stopped"))
+on(master, puppet("resource service #{service} ensure=stopped"))
 
 step "Run #{cli} with foreground subcommand, wait for #{timout_length}"
 on(master, timeout_cmd, :acceptable_exit_codes => [124]) do |result|
@@ -40,5 +32,5 @@ end
 
 teardown do
   step "Teardown: Start puppetserver again"
-  on(master, puppet("resource service #{cli} ensure=running"))
+  on(master, puppet("resource service #{service} ensure=running"))
 end
--- a/acceptance/suites/tests/010-puppetserver-cli/subcommand/foreground.rb
+++ b/acceptance/suites/tests/010-puppetserver-cli/subcommand/foreground.rb
@@ -1,0 +1,44 @@
+test_name "Puppetserver 'foreground' subcommand tests."
+
+# --------------------------------------------------------------------------
+# This behavior needs to be baked into Beaker but is not yet. SERVER-79 tracks
+# the need to fix this once Beaker has been modified to make the paths to these
+# commands available.
+#
+if master.is_pe?
+  cli = "pe-puppetserver"
+else
+  cli = "puppetserver"
+end
+# --------------------------------------------------------------------------
+
+# puppetserver seems to take about 45s to start up
+timout_length = "60s"
+foreground_cmd = "#{cli} foreground"
+timeout_cmd = "timeout -s INT #{timout_length} #{foreground_cmd}"
+
+expected_messages = {
+  /Initializing the JRuby service/ => "JRuby didn't initialize",
+  /Starting web server/ => "Expected web server to start",
+  /Finished creating JRubyPuppet instance 4 of 4/ => "Expected to find 4 JRubyPuppet instances",
+  /Finished shutdown sequence/ => "Test ended without puppetserver shutting down gracefully"
+}
+
+# Start of test
+step "Stop puppetserver"
+on(master, puppet("resource service #{cli} ensure=stopped"))
+
+step "Run #{cli} with foreground subcommand, wait for #{timout_length}"
+on(master, timeout_cmd, :acceptable_exit_codes => [124]) do |result|
+  assert_no_match(/error:/i, result.stderr, "Unexpected error running puppetserver!")
+
+  step "Check that #{cli} ran successfully and shutdown gracefully"
+  expected_messages.each do |message, explanation|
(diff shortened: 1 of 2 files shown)


acceptance/suites/tests/010-puppetserver-cli/subcommand/foreground.rb
```diff
@@ -0,0 +1,44 @@
...
+
+# puppetserver seems to take about 45s to start up
+timout_length = "60s"
+foreground_cmd = "#{cli} foreground"
+timeout_cmd = "timeout -s INT #{timout_length} #{foreground_cmd}"
+
+expected_messages = {
+  /Initializing the JRuby service/ => "JRuby didn't initialize",
+  /Starting web server/ => "Expected web server to start",
+  /Finished creating JRubyPuppet instance 4 of 4/ => "Expected to find 4 JRubyPuppet instances",
```
cprice404: I don't think we explicitly set a value for the number of JRubies in the acceptance test runs, so it'd probably be best to avoid hard-coding the `4`s into the test here.

I think there's a log message that Preben added that says something like "Puppet Server is now ready to handle requests".  That might be a good replacement for the test.

acceptance/suites/tests/010-puppetserver-cli/subcommand/foreground.rb
```diff
@@ -0,0 +1,44 @@
...
+#
+if master.is_pe?
+  cli = "pe-puppetserver"
+else
+  cli = "puppetserver"
+end
+# --------------------------------------------------------------------------
+
+# puppetserver seems to take about 45s to start up
+timout_length = "60s"
```
cprice404: It's possible that 60 seconds is not long enough depending on the hardware; on the other hand I'd hate to force the test to be longer than that unless we know it's necessary.

Can we assume that Puppet Server will have already been started at least once prior to this test running?  If so, we know that all of the SSL files should be generated before this test runs, which should hopefully prevent the startup time from being too crazy.

cprice404:
+1 other than the comment about hard-coding the number of JRuby instances.

@camlow325 do you think you could snag someone from your team to do a second review of this?

acceptance/suites/tests/010-puppetserver-cli/subcommand/foreground.rb
jpinsonault: Yeah, I was thinking the same thing. Puppet is installed just prior to this being run, and the script shuts it down first, so yeah I think it's reasonable to assume that it will always have been run at least once

jpinsonault:
@cprice404 Changed jruby check to checking for "ready to handle requests"

acceptance/suites/tests/010-puppetserver-cli/subcommand/foreground.rb
camlow325: 60s is probably okay since the run that generates SSL certs should happen in the pre-suite - 90_validate_sign_cert.rb.  We've been using 120 seconds as the `TimeoutStartSec` for the startup of the service without any problems in CI - although that is only gated to the webserver coming to ready, which would happen after 1 JRubyPuppet instance has been initialized but not necessarily any others.

camlow325: s/timout/timeout

acceptance/suites/tests/010-puppetserver-cli/subcommand/foreground.rb
```diff
@@ -0,0 +1,44 @@
+test_name "Puppetserver 'foreground' subcommand tests."
+
+# --------------------------------------------------------------------------
+# This behavior needs to be baked into Beaker but is not yet. SERVER-79 tracks
+# the need to fix this once Beaker has been modified to make the paths to these
+# commands available.
+#
+if master.is_pe?
```
camlow325: I don't think you need to do this.  The cli command is called "puppetserver" in both a FOSS and a PE install.

camlow325: Hmmm, actually it looks like "cli" is being overloaded later in the test to refer to either the service name (via resource service) or the command name (via ... foreground).  The service name is "puppetserver" for FOSS but is "pe-puppetserver" for PE.  The command name, however, is "puppetserver" for both FOSS and PE.  We don't ever run these tests against a PE install so the difference probably isn't a big deal anyway, but would be good to have the usages of the name reflect how it "should" work.

camlow325:
This test seems reasonable to me at a high-level.

I think it would be a little more useful / comprehensive if we had the test actually do some sort of network request to the puppetserver that was just started - or even a full agent run.  I think that would give us more confidence that the puppetserver is fully functional than would just inspecting some of the log messages from the master.  Also, if the test were rewritten to work that way, we could alter the timeout logic a bit to not have a fixed overhead.  For example, maybe the test could spawn the puppetserver java process, wait for the web server port to be available for up to a relatively higher max number of seconds (e.g., 120 seconds), do an agent run, send a SIGTERM to shut down the service, and confirm that the service was ultimately shut down properly.  With this approach, the test would only need to run as long as it would take for the service to be spun up and handle incoming web requests and for the server to then be shutdown - as opposed to the test requiring the server process to be running for a full 60 or so number of seconds.

I understand, though, that this approach would be considerably more complex to test so I'm willing to go with the test basically as written.  @cprice404, I'll defer to you on how much more bandwidth you'd be willing to dedicate to this one at this point.

acceptance/suites/tests/010-puppetserver-cli/subcommand/foreground.rb
jpinsonault: Ah ok, that makes sense. I copied it from the `gem.rb` test, so maybe that test should be changed at some point too.

cprice404:
@camlow325 IMO that's out of scope for this ticket.  The impetus for this ticket was just to catch potential regressions to the logging configuration (because @jeffmccune noticed a regression on that at one point), which I think he's done here.  Trying to figure out a way to run the `foreground` process in the background (whoa) does not seem worth the effort IMO; you'd have to still keep a handle to the STDOUT/STDERR of the process in order to validate the logging (which was the main goal of this ticket), plus find a way to find the pid, send the signal, etc.  Also I feel like we have sufficient coverage of the rest of the service code that if we see the log message about the server being ready, chances are very good that we are in good shape.

If we were going to change anything about this test, in the interest of keeping the runtime a bit shorter, I'd probably just change the timeout to something shorter and don't even worry about whether the server comes all the way up within the timeout; just check to see that something was logged to STDOUT.

Oh, one other thing that might be worth adding to the test would be a use of the `--debug` flag, and validating that it causes some DEBUG-level log messages to be logged to STDOUT.  That actually seems pretty important.

Beyond that I'm +1 and happy for this to be merged w/o waiting another 24hrs on my behalf.

acceptance/suites/tests/010-puppetserver-cli/subcommand/foreground.rb
camlow325: Yeah, good catch.  If it's not too much work -- like no more than a few minutes -- I'd probably just make the necessary updates to the `gem.rb` test to reflect this as well.

camlow325:
@cprice404 I'm okay with not doing any of the additional work I'd mentioned about making a request to the server to see that it is functional for fielding web requests.  And your idea of including some validation of a message for `--debug` sounds reasonable as well.

jpinsonault:
Ok, sounds like a plan

jpinsonault:
@camlow325 Added the debug flag and a check to see if a debug message was printed
Not overloading the cli variable anymore

Did the same for the gem.rb test

cprice404:
+1

acceptance/suites/tests/010-puppetserver-cli/subcommand/gem.rb
```diff
@@ -70,18 +59,10 @@
 
 on(master, "#{gem_cleanup}")
 
-step "Verify that current list matches initial list."
+step "Verify that current list matchs initial list."
```
camlow325: s/matchs/matches

acceptance/suites/tests/010-puppetserver-cli/subcommand/gem.rb
```diff
@@ -70,18 +59,10 @@
...
 
 final_installed_gems = get_gem_list(master, "#{gem_list}")
 
 initial_installed_gems.each do |gem_info|
   assert_send([final_installed_gems, :include?, gem_info])
 end
-
```
camlow325: I can't figure out why this and the removal of the `gem_env` definition above are showing up as diffs here since it doesn't look like this code is on the current master branch.

camlow325:
:+1: other than the one cosmetic spelling thing.  I guess I'm okay with the removal of the `is_pe?` check for determining the name of the service to use since we don't run these tests in PE today.  Would need to revisit that if we ever do in the future, though.

jpinsonault:
@camlow325 Phew, ok that was confusing. I think I hadn't done a pull --rebase before putting up the PR. I don't know why that typo and those extra lines at the bottom of `gem.rb` were there, since they're in the 2.1.x branch now. But pulled and rebased and now every looks good from my end

camlow325:
:+1: on the changes, although I think it might be better to target this at the stable branch and we can merge up to master from there.  @jpinsonault, you good with closing this and putting up a new PR targeted at the stable branch?

jpinsonault:
@camlow325 Yeah, sounds good

PR #521, opened 2015-05-03
(MAINT) merge april changes from stable/1.1

cprice404:
This PR merges up the changes from April from the stable/1.1 branch into master.  This includes all the changes around the big JRuby namespace re-org, as well as the new `max-requests-per-instance` setting.


Gemfile
```diff
@@ -17,7 +17,7 @@ group :test do
   gem 'rspec'
   gem 'beaker', *location_for(ENV['BEAKER_VERSION'] || '~> 2.9.0')
   if ENV['GEM_SOURCE'] =~ /rubygems\.delivery\.puppetlabs\.net/
-    gem 'sqa-utils', '~> 0.11'
+    gem 'sqa-utils', '0.12.1'
```
camlow325: @jeffmccune or @waynr - is this version of 'sqa-utils' appropriate for our master branch of Puppet Server - Puppet Server 2.x running on top of Puppet 4.x?

waynr: No, you should use 0.13.0 if you want to pin to a specific version or anything '~>0.13' if not.

ruby/facter
```diff
@@ -1 +1 @@
-Subproject commit 8ff4fa5e0251aaa1b3fea1e2e308a34b2ffe6027
+Subproject commit 691e2b164ec4144d0ce80c14d3132e89242f26cc
```
camlow325: Looks like this would bump up our Puppet Server 2.0 facter submodule to 2.4.3.  @kylog - does this sound like a reasonable version of Facter to be testing against for our latest Puppet Server 2 / AIO related work?

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -0,0 +1,164 @@
...
+
+  * :master-run-dir - path to the puppetmaster's run dir;
+    if not specified, will use the puppet default.
+
+  * :master-log-dir - path to the puppetmaster's log dir;
+    if not specified, will use the puppet default.
+
+  * :max-active-instances - The maximum number of JRubyPuppet instances that
+      will be pooled. If not specified, the system's
+      number of CPUs+2 will be used.
```
camlow325: This should be updated to the new algorithm, right?  (Probably need to make the same change to the "stable" branch, too, since I think this just came from there in the merge-up).

ruby/facter
kylog: Facter 2.4.3 is what is in the 1.0.z puppet-agent versions, the latest available at the moment, so yes, that makes sense to me.

Gemfile
cprice404: This is a merge-up commit.  None of our other branches are using `0.13`, and none of them are failing tests, so I don't think we should make that change as part of this PR.

That's not to say that we shouldn't do it soon/right away, but I'd like to keep this PR focused strictly on merging and addressing merge conflicts.

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
cprice404: You're correct.  Will address in a separate PR against stable (which will end up here in a subsequent merge-up.)

cprice404:
AFAICT none of the comments so far warrant changes to this PR.  Will file a separate PR against stable for the JRubyPuppet schema docstring, and will defer to whomever is touching acceptance tests next to make a call on the beaker 0.13 decision.

Gemfile
waynr: @cprice404 Actually, all the other branches are using `0.13`.

Notice that in this patch `~> 0.11` (less restrictive) is being replaced with `0.12.1`. Last week when I released `0.13.0` of the `sqa-utils` gem I had to submit a patch to puppet-server stable branch because the version of `beaker` being used was incompatible with `sqa-utils 0.13.0' which probably means `0.13.0` should have been a major version increase, but it's too late for that and (hopefully) in the near future we will be deprecating `sqa-utils` in favor of a more aptly-named `beaker-genconfig` gem.

While pinning to `0.12.1` in puppet-server master and 2.1.x branches won't cause any problems, I don't see any particular reason to do that given that these branches have been implicitly running off `0.13.0` since last week.

cprice404: https://github.com/puppetlabs/puppet-server/blob/master/Gemfile#L20
https://github.com/puppetlabs/puppet-server/blob/stable/Gemfile#L10

wat?

cprice404: and to be clear: I'm not voting for pinning to any specific version, I'm voting for not combining functional changes with merge commits.  So, we can totally update it, this PR just doesn't seem like the right place to do that.

waynr: `~> 0.11` in bundler land means "greater than 0.11.0 but less than 1.0.0" which means that puppet-server master and 2.1.x brances will accept 0.13.0 which is between 0.11.0 and 1.0.0.

cprice404: OK, so then what you are saying is that when dealing with the merge conflict I should have chosen the `~> 0.11` from `master` instead of the `0.12.1` from `stable`.  That I can do.

waynr: Well if you don't mind dealing with the churn of updating the Gemfile again after this on both master and 2.1.x branches, I recommend doing that...

waynr: @cprice404 yes @ your last comment

cprice404:
I squashed in a change to prefer the sqa-utils version from master instead of from stable, as per discussion with @waynr .

camlow325:
I'm basically :+1: on this - although it looks like it might need a rebase now.  Might do a little adhoc flush / memory leak testing just to see that things look sane with this given that we have some divergence between master and stable still (unrelated to this PR) - e.g., around the http client initialization.

cprice404:
Updated.  Rebasing this is not fun, so if there are no major objections with it it would be really nice to get it behind us.

waynr:
I am :+1: after the Gemfile fix but I also haven't reviewed any other part of the PR since I'm not particularly familiar with most of the code.

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
```diff
@@ -259,21 +248,35 @@
     (= (:type x)
        :puppetlabs.services.request-handler.request-handler-core/bad-request)))
 
+(defn jruby-timeout?
+  [x]
+  "Determine if the supplied slingshot message is for a JRuby borrow timeout."
```
nwolfe: Whoops, docstring in wrong place.  Probably not worth the rebase+wait though :sheep:

nwolfe:
:+1:

camlow325:
For posterity, I did do some basic flush testing with agent runs and an http report processor.  Both for the automatic pool flushing and for a manual pool flush via the admin API, I didn't see any `ScriptingContainer` or `CloseableHttpAsyncClient` objects being leaked.  So I'm satisfied that it doesn't look like there's been any regression here around memory usage per these changes.

PR #523, opened 2015-05-06
(MAINT) Fix docstring re: num cpus

cprice404:
This commit removes a reference to the old default value of
`num cpus + 2` for the `max-active-instances` setting.


fpringvaldsen:
:+1:

PR #526, opened 2015-05-07
(MAINT) Fix missed usages of renamed namespace

nwolfe:
The alias was changed in the namespace declaration but a couple usages
were missed.


camlow325:
:+1:

PR #529, opened 2015-05-08
(SERVER-547) Implement 3.x URL compatibility service

dankreek:


camlow325:
Travis failed - looks like we need to update the `master_spec` test to look for a Puppet version of '4.0.0' instead of '4.0.0-rc1' because of the Ruby Puppet submodule bump.

dev-resources/boostrap-with-legacy-service.cfg
```diff
@@ -0,0 +1,11 @@
+puppetlabs.services.request-handler.request-handler-service/request-handler-service
```
camlow325: s/boostrap/bootstrap in the name of the file.

dev/bootstrap.cfg
```diff
@@ -5,9 +5,11 @@ puppetlabs.trapperkeeper.services.webserver.jetty9-service/jetty9-service
...
+puppetlabs.services.legacy-routes.legacy-routes-service/legacy-routes-service
+
 puppetlabs.services.puppet-admin.puppet-admin-service/puppet-admin-service
 
 # To enable the CA service, leave the following line uncommented
 puppetlabs.services.ca.certificate-authority-service/certificate-authority-service
 # To disable the CA service, comment out the above line and uncomment the line below
-#puppetlabs.services.ca.certificate-authority-disabled-service/certificate-authority-disabled-service
\ No newline at end of file
+#puppetlabs.services.ca.certificate-authority-disabled-service/certificate-authority-disabled-service
```
camlow325: ␊!

camlow325:
I think it would be good to go ahead and change the `.../ezbake/config/conf.d/web-routes.conf` file as part of this PR to have the `legacy-routes` service registered and remove the `invalid-in-puppet-4` route.  That way when this PR first comes in we would hopefully be ready to start exercising it in CI.

camlow325:
Similarly, it would be good to go ahead and add the `legacy-routes` service to the `.../ezbake/config/bootstrap.cfg` file.

dev/user_repl.clj
```diff
@@ -61,7 +62,8 @@
                request-handler-service
```
camlow325: I'd also comment out / remove the `:invalid-in-puppet-4` route from the `master-service` route definition in the `puppet-server-conf` function.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
```diff
@@ -0,0 +1,144 @@
...
+            [compojure.core :as compojure]
+            [ring.util.codec :as ring-codec]
+            [compojure.route :as route]
+            [clojure.tools.logging :as log]))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; v3 => v4 header translation support functions
+
+; TODO: Improve the naive parsing here with a real library like
+; https://github.com/ToBeReplaced/http-accept-headers
```
camlow325: Seems odd that nothing else would be available in any of the Ring-y middleware libraries that we're already pulling in.  I did find a `parse-accepts-header` in liberator - https://github.com/clojure-liberator/liberator/blob/0.12.0/src/liberator/conneg.clj#L180.  Didn't immediately see a corresponding "flatten" function for that in the namespace, although it wouldn't seem too difficult to write that part, I guess.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
```diff
@@ -0,0 +1,144 @@
...
+(defn- map-accept-s-pson-to-binary
+  "Convert Accept: s, pson media types to Accept: binary given a ring request
+  map.  This method should only be used to munge file_bucket_file API requests
+  from legacy agents speaking to a contemporary master."
+  [request]
+  (if-let [accept (get-in request [:headers "accept"])]
+    (let [v4-accept-val "binary"
+          vals (str/split accept #"\s*,\s*")                ; BAD! see RFC-2616
+          munged-accept (->> vals
+                          (replace {"s" v4-accept-val "pson" v4-accept-val})
```
camlow325: Do we need to have special handling to filter out "pson" here?  If the agent sends "pson" but the master doesn't support it, is it just passively ignored by the master or does the master throw an error - e.g., 415 Unsupported Media Type?  If it's passively ignored, I'd be tempted to just let it pass through and not alter it here.

I think substituting "s" and "raw" with "binary" makes sense, but substituting "pson" with "binary" doesn't seem right to me.  If we have to filter out "pson" from the request, I'd be tempted to just remove it rather than translating it to something that doesn't seem like the right replacement.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
jeffmccune: Cool, thanks for finding that.  Do you think I should update this PR to use `parse-accepts-header` in liberator, or perhaps copy the function if we're planning on removing our dependency on liberator in the future?

camlow325: Good question.  I'd be tempted to just reference it directly to copying it but I haven't kept up on the conversations about removing liberator entirely.  I'd been thinking that was a soft goal and only in the sense of not automatically defaulting to layering new routes on top of Compojure/Bidi with it by default.  @cprice404 or @KevinCorcoran - do you think we'd we be going down a bad road by referencing liberator for a utility function here?

test/unit/puppetlabs/services/legacy_routes/legacy_routes_core_test.clj
```diff
@@ -0,0 +1,145 @@
...
+
+(deftest test-v3-header-munging
+  (testing "(SERVER-548) Header munging"
+    (testing "Accept: raw is munged to Accept: binary for all paths"
+      (doseq [[path methods] accept-header-test-data
+              [method examples] methods
+              [example expected] examples
+              :let [mockreq (mock/request method path)
+                    request (assoc-in mockreq accept-header example)]]
+        (test-munged-request request
```
jeffmccune: I'm curious to get feedback on this approach to testing the processed header before it hits the handler...  It feels pretty contorted, but it also seemed like an appropriate use of a functional approach to the problem rather than something more imperative (if that's the correct word).

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
```diff
@@ -0,0 +1,144 @@
...
+  and the such to be RFC compliant, but puppet agent does not use this format
+  so implementing the full spec is unnecessary."
+  [request]
+  (if-let [accept (get-in request [:headers "accept"])]
+    (let [vals (str/split accept #"\s*,\s*")]               ; BAD! See RFC-2616
+      (assoc-in request [:headers "accept"]
+        (str/join ", " (distinct (replace {"raw" "binary"} vals)))))
+    request))
+
+(defn- map-accept-s-pson-to-binary
```
camlow325: This function and the "raw" function seem very similar to me.  I'd be tempted to just combine them into a single function that walks through the collection once, replacing any "raw" or "s" references with "binary" (and maybe remove "pson" if that's the right thing to do with it).

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
```diff
@@ -0,0 +1,144 @@
...
+    (compojure/GET "/file_metadata/*" request (master-request-handler request))
+    (compojure/GET "/file_bucket_file/*" request
+      (master-request-handler (map-accept-s-pson-to-binary request)))
+    ;; TODO: file_bucket_file request PUTs from Puppet agents currently use a
+    ;; Content-Type of 'text/plain', which, per HTTP specification, would imply
+    ;; a default character encoding of ISO-8859-1 or US-ASCII be used to decode
+    ;; the data.  This would be incorrect to do in this case, however, because
+    ;; the actual payload is "binary".  Coercing this to
+    ;; "application/octet-stream" for now as this is synonymous with "binary".
+    ;; This should be removed when/if Puppet agents start using an appropriate
```
camlow325: I think we could just take the "TODO" and the "This should be removed when/if..." part off now because we're adding this back in specifically to support Puppet 3.X agent behavior which won't change in newer Puppet 3.X releases, for compatibility reasons.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
```diff
@@ -0,0 +1,144 @@
...
+    (compojure/POST "/catalog/*" request (master-request-handler request))
+    (compojure/PUT "/report/*" request (master-request-handler request))
+    (compojure/GET "/resource_type/*" request (master-request-handler request))
+    (compojure/GET "/resource_types/*" request (master-request-handler request))
+    (compojure/GET "/status/*" request (master-request-handler request))
+
+    ;; TODO: when we get rid of the legacy dashboard after 3.4, we should remove
+    ;; this endpoint as well.  It makes more sense for this type of query to be
+    ;; directed to PuppetDB.
+    (compojure/GET "/facts_search/*" request (master-request-handler request))
```
camlow325: @cprice404 or @jonathannewman - Do you know if the next release of PE - the first that would include Puppet Server 2 / Puppet 4 support - would be one in which the "facts_search" endpoint would no longer need to be used by the PE UI?  Since this PR is targeted for work to be used in that timeframe, I wonder if we can have this PR discontinue support for this endpoint.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
```diff
@@ -0,0 +1,144 @@
...
+
+(defn legacy-routes
+  "Creates all of the compojure routes for the master."
+  [master-request-handler ca-request-handler]
+  (compojure/routes
+    (compojure/context "/v2.0" request
+                       (v2_0-routes master-request-handler))
+    (compojure/context "/:environment" [environment]
+                       (legacy-master-routes master-request-handler ca-request-handler))
+    (route/not-found "Not Found")))
```
camlow325: The presence of a `not-found` route here makes me nervous because we know we have to mount this handler at "" on the server.  I'd worry that if this were running in a stack that had non-master / non-CA routes registered that the "not-found" route here could swallow up any requests to those other routes.  I think it would be safer to just not include that here, like we didn't for the implementations of these routes in the Puppet Server 1.x implementation.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
```diff
@@ -0,0 +1,144 @@
...
+  (compojure/routes
+    (compojure/GET "/environments" request
+                   (request-handler request))))
+
+(defn legacy-master-routes
+  [master-request-handler ca-request-handler]
+  (compojure/routes
+    (compojure/GET "/node/*" request (master-request-handler request))
+    (compojure/GET "/facts/*" request (master-request-handler request))
+    (compojure/GET "/file_content/*" request (master-request-handler request))
```
camlow325: I'm pretty sure that there's some Accept header remapping that needs to happen for this one too.  Can't remember if its an s->binary or raw->binary.

test/integration/puppetlabs/services/legacy_routes/legacy_routes_test.clj
```diff
@@ -0,0 +1,35 @@
...
+  :once
+  schema-test/validate-schemas
+  (jruby-testutils/with-puppet-conf (fs/file test-resources-dir "puppet.conf")))
+
+(defn http-get [path]
+  (http-client/get
+    (str "https://localhost:8140" path)
+    bootstrap/request-options))
+
+(deftest ^:integration legacy-routes
```
camlow325: Per the conversation in SERVER-526, I had been expecting that we'd try to have at least some level of coverage for validating that "auth.conf" rules created per the v3 routes would be evaluated against requests made to the corresponding legacy routes.  If that's going to cause a significant amount of rework -- like more than 1/2 hour -- I'm open to putting that in a follow-up ticket in the interest of getting this merged sooner.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
```diff
@@ -0,0 +1,144 @@
...
+(defn v2_0-routes
+  "Creates the compojure routes to handle the 3.X master's '/v2.0' routes."
+  [request-handler]
+  (compojure/routes
+    (compojure/GET "/environments" request
+                   (request-handler request))))
+
+(defn legacy-master-routes
+  [master-request-handler ca-request-handler]
+  (compojure/routes
```
camlow325: RE: the header remapping, I'm wondering if enough of the routes here require some header remapping and if the mappings are always 1->1, like just raw and s -> binary, if it would make sense to just have all of these routes call through a common function that does the header translation before calling through to `master-request-handler`.  Just concerned we might be missing some subtle case for some endpoints that the custom translation we're doing per endpoint here wouldn't cover.

test/unit/puppetlabs/services/legacy_routes/legacy_routes_core_test.clj
```diff
@@ -0,0 +1,145 @@
...
+(def accept-header-raw-examples
+  (assoc accept-header-common-examples
+    "s, pson, raw" "s, pson, binary"))
+
+(def accept-header-s-pson-examples
+  (assoc accept-header-common-examples
+    "s, pson, raw" "binary"
+    "s, pson, foo" "binary, foo"))
+
+; [path [method [example expected]]]
```
camlow325: Might be better to do this as a docstring?

test/unit/puppetlabs/services/legacy_routes/legacy_routes_core_test.clj
```diff
@@ -0,0 +1,145 @@
...
+(defn assertion-wrapper
+  "ring middleware that asserts by passing the request as the first argument
+  to assertion"
+  [handler assertion]
+  (fn [req]
+    (assertion req)
+    (handler req)))
+
+(defn validating-app
+  "Return a ring app with the results of build-ring-handler wrapped around a an
```
camlow325: s/a an/an

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
```diff
@@ -0,0 +1,144 @@
...
+
+; TODO: Improve the naive parsing here with a real library like
+; https://github.com/ToBeReplaced/http-accept-headers
+(defn- map-accept-raw-to-binary
+  "Convert Accept: raw media types to Accept: binary.  NOTE: This method does
+  not conform to RFC-2616 Accept: header specification, but then again
+  neither does Accept: raw or Accept: binary
+
+  The split on `,` is naive, the method should take into account `;` parameters
+  and the such to be RFC compliant, but puppet agent does not use this format
```
camlow325: s/and the such/and such

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
camlow325: Hmmm, I guess the `request-compatibility-wrapper`'s call to that covers this with its `map-accept-raw-to-binary` for the most part.  Maybe if `map-accept-raw-to-binary` and `map-accept-s-pson-to-binary` were combined together and called from there - assuming that's feasible - this would cover my comment here.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
```diff
@@ -0,0 +1,144 @@
...
+                          (distinct)
+                          (str/join ", "))]
+      (assoc-in request [:headers "accept"] munged-accept))
+    request))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Handler wrapper
+
+(defn request-compatibility-wrapper
+  "Given a Puppet 3.X CA request, convert it to a Puppet 4.X CA request."
```
camlow325: This is called for CA and master routes, right - so maybe say "agent" instead of "CA" here?

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
```diff
@@ -0,0 +1,144 @@
...
+                       (v2_0-routes master-request-handler))
+    (compojure/context "/:environment" [environment]
+                       (legacy-master-routes master-request-handler ca-request-handler))
+    (route/not-found "Not Found")))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Public
+
+(defn build-ring-handler
+  [master-handler master-mount master-api-verion
```
camlow325: s/verion/version

test/unit/puppetlabs/services/legacy_routes/legacy_routes_core_test.clj
```diff
@@ -0,0 +1,145 @@
...
+    (testing "Legacy ca routes"
+      (doseq [path ["certificate_status"
+                    "certificate_statuses"
+                    "certificate"
+                    "certificate_revocation_list"
+                    "certificate_request"
+                    "certificate_requests"]]
+        (let [resp (request (str "/" environment "/" path "/" route-val))]
+          (is (= 200 (:status resp)))
+          (is (= (str ca-mount "/" ca-api-version "/" path "/" route-val)))
```
camlow325: Should this assertion include `(:uri resp)`?

test/unit/puppetlabs/services/legacy_routes/legacy_routes_core_test.clj
```diff
@@ -0,0 +1,145 @@
...
+          (is (= (ring-codec/form-encode {:environment environment}))))))))
```
camlow325: Should this assertion include `(:query-string resp)`?

test/unit/puppetlabs/services/legacy_routes/legacy_routes_core_test.clj
```diff
@@ -0,0 +1,145 @@
...
+      handler master-mount master-api-version
+      identity ca-mount ca-api-version)))
+
+(defn test-munged-request
+  "Validate a request by threading it through the legacy-routes request handler
+  into an assertion-wrapper ring middleware."
+  [request assertion]
+  ((validating-app assertion) request))
+
+(deftest test-legacy-routes
```
camlow325: We should have at least some coverage of cases where we're making a request that has some query parameters in the original URL to see that the `:environment` one is added to the list rather than just replacing them entirely.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
camlow325: n/m, Looks like 'request-compatibility-wrapper' already has this covered.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
cprice404: I'd probably copy the function rather than add another dependency on liberator, personally.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
jonathannewman: @camlow325 I'm not sure exactly where that endpoint is used in the 3.8 UI, I'm assuming legacy dashboard somewhere.  We aren't using that endpoint currently, and I know of no plans to use it.  @scottyw any thoughts?

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
jeffmccune: @camlow325 I thought about doing this, but I'd have to implement the function to contain some way to either calculate, or be informed of, the type of request being processed.  As it stands, the only two headers that have 1:1 mappings are `Accept: raw` to `Accept: binary` and `Content-Type: text/plain` to `Content-Type: application/binary`.  The content-type header requires somewhat special handling in that it was previously a top level key instead of a `:headers` sub-key in the request map.

The processing of `Accept: s, pson` only happens for file bucket file requests.  I ended up breaking them out into discrete functions to avoid the need to duplicate the behavior of inspecting the request URI and changing behavior accordingly.

Given this additional information about the nature of the mapping, please let me know if you'd like to see this refactored into a common function.

test/unit/puppetlabs/services/legacy_routes/legacy_routes_core_test.clj
jeffmccune: Yes, added in a follow-up commit.

test/unit/puppetlabs/services/legacy_routes/legacy_routes_core_test.clj
jeffmccune: Fixed, thanks for the catch.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
jeffmccune: Fixed, thanks for the catch.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
jeffmccune: Fixed, thanks for the catch.

jeffmccune:
@dankreek I don't think I can push to your branch (might want to add me so I can?) or just fast-forward to ee33872 from my fork?  This patch addresses the minor comments @camlow325 made.  Some of the bigger ones I left.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
camlow325: @jeffmccune Yeah, I think the way you already have it with the `Content-Type` being forced to `application/octet-stream` for "file_bucket_file" only makes sense.  We know we have a problem where the agent "lies" about what the true content type is for that request specifically.

All I was suggesting is that we might be able to consolidate the `Accept` header transformation into a single function that just does one pass through the values - transforming `raw` or `s` to `binary` (because I think that can be done generically, for all endpoints) and preserving any other value as-is.  My question was whether that would cause a problem - like if the `file_bucket_file` endpoint were to receive an `Accept: binary, pson`, would that not work properly for some reason - like if the master were to have some logic which would see `pson` as a "bad" type - but maybe only for some endpoints? - and choke with an error - e.g., HTTP 415 / unsupported media type.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
scottyw: Same situation for us too.

camlow325: Thanks, @jonathannewman and @scottyw.  @cprice404 and @rlinehan - do you think it would be reasonable to start blacklisting this for Puppet Server 2.x?  Not sure if it's too late for that now that we cut Puppet Server 2.0 with this still in, though...

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
jeffmccune: I'm not sure how a request with `pson` is handled when it doesn't have `s` or `binary`.  The current behavior is based on the observation that `s, pson` have changed to `binary` in v4 requests.

I'll change the behavior to filter out `pson` and only map `s` to `binary` as that should have the exact same effect for valid requests.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
jeffmccune: I thought about doing this, but `raw` is mapped for all URI's while `s` is mapped only for the file bucket URI's.  As a result, I ended up with two different functions so I could call them from different locations respectively.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
camlow325: As a general practice, copying functionality from a non-PL library into a PL one rather than referencing the project directly as a dependency doesn't quite feel right to me.  Granted, there's a point at which if you're just talking about one small utility function which is unrelated to the overall purpose of the library it is coming from where this could make sense and this might be a little closer to that camp.

Ultimately, we need functionality here that would not only parse out the values for the header but be able to flatten them back into a valid string after we apply some changes.  It wasn't clear to me that the liberator implementation provided the latter.  Maybe the `ToBeReplaced/http-accept-headers` library does and could, therefore, be a better fit for our use case anyway - especially if we're still hoping to sever all ties with liberator from Puppet Server.

In any event, I think the simpler implementation in this PR that doesn't account for quality weights is probably sufficient for now as it should cover how actual agents populate the header.  So I'd be good with this implementation coming in as-is for now and maybe just file a separate ticket to explore pulling in a more complete upstream library later on.  I wouldn't block the next release for that latter work being done.

KevinCorcoran: @camlow325 - Right, the plan is to eventually remove the dependency on liberator.  So calling another function from it here would just be creating more of a mess that would need to be cleaned up eventually.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
camlow325: So is `pson` one that is still valid for some endpoints and not others in Puppet 4, or should we be filtering it out for all endpoints?

jeffmccune: Changed to remove `pson` rather than map it in 1cb8ff5  @dankreek could you please pick this one too?

jeffmccune: @dankreek I missed the HEAD request in 1cb8ff5 and amended it to 6fb8d29, so could you pick that amended commit please?

jeffmccune: @camlow325 

> So is pson one that is still valid for some endpoints and not others in Puppet 4, or should we be filtering it out for all endpoints?

It's still there for some requests in Puppet 4, e.g. NODE GET my-agent which looks like:

```
~~~~~ NODE GET my-agent ~~~~~
2015-04-17 12:43:14,908 INFO  [c.p.t.p.CharLoggingHandler] [id: 0x4135ecd2, /127.0.0.1:53807 => /127.0.0.1:8666]
INPUT
GET /puppet/v3/node/my-agent?environment=production&transaction_uuid=539ac6f4-a5ad-4970-ad7c-7c4ac675c08a&fail_on_404=true HTTP/1.1^M
Accept: pson, msgpack, yaml, binary^M
X-Puppet-Version: 4.0.0^M
Accept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3^M
User-Agent: Ruby^M
Host: localhost:8666^M
^M

2015-04-17 12:43:14,934 INFO  [c.p.t.p.CharLoggingHandler] [id: 0x4135ecd2, /127.0.0.1:53807 => /127.0.0.1:8666]
OUTPUT
HTTP/1.1 200 OK ^M
X-Puppet-Version: 4.0.0^M
Content-Type: text/pson^M
Server: WEBrick/1.3.1 (Ruby/1.9.3/2014-11-13) OpenSSL/0.9.8za^M
Date: Fri, 17 Apr 2015 19:43:14 GMT^M
ontent-Length: 46^M
Connection: Keep-Alive^M
^M
```

camlow325: @jeffmccune Out of curiosity, what happens if an agent includes `pson` in a `file_bucket_file` request to a Puppet 4-based master?  Does the master request fail?

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
jeffmccune: Addressed in 6fb8d29..2142fa5

jeffmccune:
> Travis failed - looks like we need to update the master_spec test to look for a Puppet version of '4.0.0' instead of '4.0.0-rc1' because of the Ruby Puppet submodule bump.

Fixed in 2142fa5..7a58761

dev-resources/boostrap-with-legacy-service.cfg
jeffmccune: Fixed in 7a58761..69b2591

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
jeffmccune: > @jeffmccune Out of curiosity, what happens if an agent includes pson in a file_bucket_file request to a Puppet 4-based master? Does the master request fail?

The agent run works when the header is munged to a value of `Accept: binary, pson`, so I think we could entirely ignore `pson` if we want to...

KevinCorcoran: Can we also purge it from our collective consciousness forever?

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
jeffmccune: Confirmed, Accept: raw => binary happens for all requests via `request-compatibility-wrapper`

jeffmccune:
I'm finished addressing comments in this branch for the time being, @dankreek @camlow325 please let me know if there are other things I need to address before this is able to be merged.  I'll be working on the auth.conf tests as a separate PR until then.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
camlow325: @jeffmccune Ok, thanks.  So my preference would still be - unless there's some dangerous reason not to do so - to just collapse the `map-accept-raw-to-binary` and `map-accept-s-pson-binary` functions into a single function which walks over the accept header values in the request once, substituting any instances of `s` or `raw` with `binary` and preserving everything else.  That function could just be called by `request-compatibility-wrapper` so that it all routes are automatically covered - to avoid us having to worry about which ones might use `s` vs `raw` and in what scenarios, etc.  Also, we'd only have to look at one spot in the code to know where any `accept` header translation is being done.

Would be curious to hear what others' thoughts are on this.  @dankreek - what do you think?

jeffmccune: @camlow325 My only concern about the course you prefer is that while it may technically work, it will result in a new pattern of API requests that are different from Puppet 3 and 4.  The implementation as it exists now results in requests that are identical to Puppet 4 requests from the perspective of the master service.  A new pattern of requests may technically work, but it might be more difficult to troubleshoot if we do run into issues.  With that said, perhaps there is a new pattern regardless of implementation by the nature of the compatibility service itself.

I'll go ahead and implement the single function as a single commit, we can take it or leave it at merge.

jeffmccune: @camlow325 Single `map-accept-header` comment addressed in 69b2591..a93294c.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
```diff
@@ -0,0 +1,120 @@
...
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Handler wrapper
+
+(defn request-compatibility-wrapper
+  "Given a Puppet 3.X CA request, convert it to a Puppet 4.X CA request."
+  [handler mount-point api-version request]
+  (let [{{environment :environment} :params
+         path-info                  :path-info
+         query-string               :query-string} request
+        query-params (if query-string (ring-codec/form-decode query-string) {})]
```
camlow325: I ran into a problem with this around one of the CA requests that a Puppet 3.x agent makes.  Here's the original request:

~~~~
GET /production/certificate/agent929?
~~~~

In this case, `query-string` ends up being an empty string, is passed into `form-decode`, is returned back as an empty string, and results in this error being returned to the client:

~~~~
Error: Could not request certificate: Error 500 on SERVER: <html>
<head>
<meta http-equiv="Content-Type" content="text/html;charset=ISO-8859-1"/>
<title>Error 500 </title>
</head>
<body>
<h2>HTTP ERROR: 500</h2>
<p>Problem accessing /production/certificate/agent929. Reason:
<pre>    java.lang.String cannot be cast to clojure.lang.IPersistentCollection</pre></p>
<hr /><i><small>Powered by Jetty://</small></i>
</body>
</html>
~~~~

I was able to work around this by changing this line to:

~~~~clj
query-params (if (str/blank? query-string)
                         {}
                         (ring-codec/form-decode query-string))
~~~~

Not sure if that's the best change but got me past this problem.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
```diff
@@ -0,0 +1,120 @@
...
+    ;; this endpoint as well.  It makes more sense for this type of query to be
+    ;; directed to PuppetDB.
+    (compojure/GET "/facts_search/*" request (master-request-handler request))
+
+    ;; legacy CA routes
+    (compojure/ANY "/certificate_status/*" request (ca-request-handler request))
+    (compojure/ANY "/certificate_statuses/*" request (ca-request-handler request))
+    (compojure/GET "/certificate/*" request (ca-request-handler request))
+    (compojure/GET "/certificate_revocation_list/*" request (ca-request-handler request))
+    (compojure/GET "/certificate_request/*" request (ca-request-handler request))
```
camlow325: Looks like we're missing this:

~~~~clj
(compojure/PUT "/certificate_request/*" request (ca-request-handler request))
~~~~

The `certificate_request` PUT re-routing didn't work without this for me anyway.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
```diff
@@ -0,0 +1,120 @@
...
+                       (legacy-master-routes master-request-handler ca-request-handler))
+    (route/not-found "Not Found")))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Public
+
+(defn build-ring-handler
+  [master-handler master-mount master-api-version
+   ca-handler ca-mount ca-api-version]
+  (log/info "Building ring handler for legacy service")
```
camlow325: Might be best to put this line above where the ring handler is registered for the first time and not here.  Otherwise, looks like we get this in the output every time a request is processed.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
```diff
@@ -0,0 +1,120 @@
...
+    ;; directed to PuppetDB.
+    (compojure/GET "/facts_search/*" request (master-request-handler request))
+
+    ;; legacy CA routes
+    (compojure/ANY "/certificate_status/*" request (ca-request-handler request))
+    (compojure/ANY "/certificate_statuses/*" request (ca-request-handler request))
+    (compojure/GET "/certificate/*" request (ca-request-handler request))
+    (compojure/GET "/certificate_revocation_list/*" request (ca-request-handler request))
+    (compojure/GET "/certificate_request/*" request (ca-request-handler request))
+    (compojure/GET "/certificate_requests/*" request (ca-request-handler request))))
```
camlow325: Is `certificate_requests` one that we need to support re-routing for?  Doesn't look like this one is even being handled in Puppet Server 1.0.8 - https://github.com/puppetlabs/puppet-server/blob/puppet-server-1.0.8/src/clj/puppetlabs/services/ca/certificate_authority_core.clj#L239-L255.  Maybe @nwolfe remembers more about this endpoint?

nwolfe: We explicitly didn't support this in the CA rewrite because nothing uses it - it existed only because of the magical plurality+indirector nonsense.  Plus, it does not offer any information that isn't available from the other endpoints.

Puppet Server has never supported it, you would just get a 404 Not Found.  Probably best to keep it that way and not introduce support for it now?

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
cprice404: @camlow325 I think we would be better off leaving this alone until we have time to go get rid of it from the ruby code as well, which is probably not right now.

I think that during the Puppet 5 sweep through the ruby code, one of our goals should be to reconcile all of the HTTP endpoints so that they either exist on both sides or neither.

rlinehan: IIRC it's been removed from the ruby code already (was deprecated in 3.x, removed in 4.0 https://docs.puppetlabs.com/guides/inventory_service.html#the-inventory-service-is-deprecated), so since it doesn't actually return anything it's probably reasonable to blacklist it...

camlow325: Cool, thanks, @rlinehan.  Yeah, if it doesn't actually return anything in 4.0, I'd also be inclined to blacklist it.  @nwolfe - does this jive with your recent explorations around the behavior of the different endpoints in Puppet Server 2 / Puppet 4?

nwolfe: bq. does this jive with your recent explorations around the behavior of the different endpoints in Puppet Server 2 / Puppet 4?

Yep - `facts` and `facts_search` are not available in Puppet 4.  You get `Not Found: Could not find indirection 'fact'` and `Not Found: No handler for facts` when hitting them, appropriately.

cprice404: well I'll be!  That's great.  Then yes, by all means, let's get rid of them.

test/unit/puppetlabs/services/legacy_routes/legacy_routes_core_test.clj
camlow325: It seems like another approach that would work for this would be to just do the expected/actual test against what is returned from `test-munged-request`.  That would have the headers in the Ring app as they were after munging from all of the other middleware - and might avoid the need to pass the assertion function all the way through.  I wouldn't hold up this PR to get that changed, but it seems like that approach might be a little easier to read.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
dankreek: I believe that's spelled L-I-L-B

dankreek:
| Travis failed - looks like we need to update the master_spec test to look for a Puppet version of '4.0.0' instead of '4.0.0-rc1' because of the Ruby Puppet submodule bump.

Whoops, still need to fix this.

dankreek:
Everything other than the travis issue should be ready to go now

ezbake/config/conf.d/web-routes.conf
```diff
@@ -2,11 +2,8 @@ web-router-service: {
     # These two should not be modified because the Puppet 4.x agent expects them to
     # be mounted at these specific paths
     "puppetlabs.services.ca.certificate-authority-service/certificate-authority-service": "/puppet-ca"
-    "puppetlabs.services.master.master-service/master-service": {master-routes:       "/puppet"
-
-                                                                 # The following controls 404 errors and should not
-                                                                 # be modified
-                                                                 invalid-in-puppet-4: "/" }
+    "puppetlabs.services.master.master-service/master-service": "/puppet"
+    "puppetlabs.services.legacy-routes.legacy-routes-service/legacy-routes-service": "/"
```
nwolfe: The value here needs to be `""` instead of `"/"`, otherwise I get a Jetty 404.  Works fine with `""` though

nwolfe: After making this change by hand, I was able to build packages off this branch and do a successful 3x->4x agent run! :godmode:

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
```diff
@@ -0,0 +1,124 @@
...
+
+(defn legacy-master-routes
+  [master-request-handler ca-request-handler]
+  (compojure/routes
+    (compojure/GET "/node/*" request (master-request-handler request))
+    (compojure/GET "/file_content/*" request (master-request-handler request))
+    (compojure/GET "/file_metadatas/*" request (master-request-handler request))
+    (compojure/GET "/file_metadata/*" request (master-request-handler request))
+    (compojure/GET "/file_bucket_file/*" request
+      ((respond-with-content-type-text-plain master-request-handler) request))
```
jeffmccune: @nwolfe @camlow325 This should fix the `puppet filebucket restore ...` issue you found.  :pray:

ezbake/config/conf.d/web-routes.conf
dankreek: Fixed that and push into the previous ez-bake commit.

KevinCorcoran:
The diff seems to say that you're adding an empty file at dev-resources/boostrap-with-legacy-service.cfg

dev/user_repl.clj
```diff
@@ -36,13 +37,13 @@
...
+     :certificate-authority {:certificate-status {:client-whitelist       []
                                                   :authorization-required false}}
      :web-router-service    {:puppetlabs.services.ca.certificate-authority-service/certificate-authority-service "/puppet-ca"
-                             :puppetlabs.services.master.master-service/master-service {:master-routes "/puppet"
-                                                                                        :invalid-in-puppet-4 "/"}
-                             :puppetlabs.services.puppet-admin.puppet-admin-service/puppet-admin-service "/admin"}
-     :puppet-admin          {:client-whitelist []
+                             :puppetlabs.services.master.master-service/master-service "/puppet"
+                             :puppetlabs.services.puppet-admin.puppet-admin-service/puppet-admin-service         "/admin"
+                             :puppetlabs.services.legacy-routes.legacy-routes-service/legacy-routes-service "/"}
```
camlow325: Should we mount this one on "" also?

jeffmccune: Yeah, I'll add that in now.

jeffmccune: Addressed in c2861bb..30fca82

test/unit/puppetlabs/services/legacy_routes/legacy_routes_core_test.clj
```diff
@@ -0,0 +1,134 @@
...
+(deftest test-v3-header-munging
+  (testing "(SERVER-548) Header munging"
+    (testing "Accept: raw is munged to Accept: binary for all paths"
+      (doseq [[path methods] accept-header-test-data
+              [method examples] methods
+              [example expected] examples
+              :let [mockreq (mock/request method path)
+                    request (assoc-in mockreq accept-header example)
+                    subject (get-in (munged-request request) accept-header)]]
+        (is (= expected subject))))))
```
camlow325: Sweet - I like this much better.  :smiley:

jeffmccune: Me too!

camlow325:
:+1: - will merge assuming Travis passes one more time.

PR #530, opened 2015-05-08
(SERVER-564) Re-enable legacy Puppet status endpoint

camlow325:
This commit re-enables the legacy Puppet Ruby status endpoint.  It
simply whitelists the endpoint in Puppet Server code so that the request
can be passed through to Ruby Puppet code.

### DIFF
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -53,6 +53,8 @@
                    (request-handler request))
     (compojure/GET "/environments" request
                    (request-handler request))
+    (compojure/GET "/status/*" request
+                   (request-handler request))
 
     ;; TODO: when we get rid of the legacy dashboard after 3.4, we should remove
     ;; this endpoint as well.  It makes more sense for this type of query to be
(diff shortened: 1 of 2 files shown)


fpringvaldsen:
:see_no_evil: 

I said this on https://github.com/puppetlabs/puppet-server/pull/531 as well, but should we add a simple test to validate this endpoint is working?

jeffmccune:
:+1:

KevinCorcoran:
Based on the discussion on #531, it looks like it was decided to forego adding a test for this.

@camlow325 - Does this really need to go into 2.1.x?  My only concern is merge conflicts when both 2.1.x and stable get merged into master ... It's only 3 lines, so not a big deal - just thought it was worth asking.

camlow325:
@KevinCorcoran It does need to go into 2.1.x, yeah, because product wants it in the Puppet Server 2.1.0 release, which we're still at present planning to do from this branch.  You're right that we'll need to make sure we cover this when we merge back to master later.  I'd like to get this in soon so that we don't have to make two passes back through this to merge up later.

KevinCorcoran:
:+1: that's all I needed to hear.

PR #531, opened 2015-05-08
(SERVER-564) Re-enable legacy Puppet status endpoint

camlow325:
This commit re-enables the legacy Puppet Ruby status endpoint.  It
simply whitelists the endpoint in Puppet Server code so that the request
can be passed through to Ruby Puppet code.

### DIFF
--- a/test/unit/puppetlabs/services/master/master_core_test.clj
+++ b/test/unit/puppetlabs/services/master/master_core_test.clj
@@ -32,6 +32,7 @@
                    "file_bucket_file"
                    "resource_type"
                    "resource_types"
+                   "status"
                    "facts_search"]
              :post ["catalog"]
              :put ["file_bucket_file"
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -64,6 +64,8 @@
                    (request-handler request))
     (comidi/GET ["/resource_types/" [#".*" :rest]] request
                    (request-handler request))
+    (comidi/GET ["/status/" [#".*" :rest]] request
+                   (request-handler request))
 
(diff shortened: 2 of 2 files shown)


test/unit/puppetlabs/services/master/master_core_test.clj
```diff
@@ -54,4 +55,4 @@
                             :content-type   "text/plain"
                             :uri            "/foo/file_bucket_file/bar"})]
       (is (= "application/octet-stream"
-             (get-in resp [:request :content-type]))))))
```
camlow325: ~mccuneprinciple !!!!

fpringvaldsen:
:see_no_evil: 

Should we add a simple test for this endpoint as well?

camlow325:
Via addition to the `test_master_routes` test, we would have a unit-level test which validates that a request to the endpoint returns a 200 OK / at least matching a route.  I could add an integration-level test which also validates the response payload.  We've typically shied away from going that far in Puppet Server Clojure tests, though, except for simple smoke tests and cases where we've implemented Puppet Server-level fixes for bugs, like the file_bucket_file binary output.  If others think adding the extra test on is compelling, I can do that.  Otherwise, though, I'd prefer not to bother with writing a test whose primary value would seem to be about validating core Ruby Puppet functionality.

test/unit/puppetlabs/services/master/master_core_test.clj
jeffmccune: moar newlines!

jeffmccune:
:+1:

camlow325:
@fpringvaldsen - you good with this coming in w/o the additional integration-level test?

fpringvaldsen:
@camlow325 Yup! The test coverage here is sufficient.

PR #532, opened 2015-05-09
(REVIEW ONLY) 3x->4x acceptance tests

nwolfe:
Just throwing this up in case anybody is curious. Other than a couple filebucket TODOs this is pretty much done, at least for the first working version.

Module: https://github.com/nwolfe/puppet-simmons

----

I was considering making each "Validate" step a separate acceptance test so that we can run through all validations even if one of them fails.

I also would like to name the validation steps after the feature they're validating rather than the file name used.

### DIFF
--- a/acceptance/suites/puppet3_tests/020_backwards_compat/test.rb
+++ b/acceptance/suites/puppet3_tests/020_backwards_compat/test.rb
@@ -1,6 +1,3 @@
-# require 'test/unit/assertions'
-# require 'puppet/acceptance/module_utils'
-
 test_name "3.x agent running against 4.x master"
 
 studio = "/tmp/simmons-studio-#{Process.pid}"
@@ -10,9 +7,8 @@ teardown do
 end
 
 step "Install modules" do
-  simmons_version = '0.2.0'
+  simmons_version = '0.2.4'
   on(master, puppet("module install nwolfe-simmons --version #{simmons_version}"))
-  on(master, puppet("module list"))
 end
 
 step "Configure file serving" do
@@ -35,11 +31,9 @@ AUTHCONF
   on(master, "chmod 644 #{authconf}")
 end
 
-step "Perform agent upgrade steps" do
-  step "Enable structured facts" do
-    agents.each do |agent|
-      on(agent, puppet("config set stringify_facts false --section agent"))
-    end
+step "Perform agent upgrade steps: enable structured facts" do
+  agents.each do |agent|
+    on(agent, puppet("config set stringify_facts false --section agent"))
   end
 end
 
@@ -69,13 +63,25 @@ end
 step "Validate source-file" do
   contents = on(agent, "cat #{studio}/source-file").stdout.chomp
   assert_equal('Static source file contents', contents)
-  # TODO test filebucket backup
+end
+
+step "Validate source-file filebucket backup" do
+  old_md5 = on(agent, "openssl dgst -md5 #{studio}/source-file-old | awk '{print $2}'").stdout.chomp
+  on(agent, puppet("filebucket restore #{studio}/source-file-backup #{old_md5} --server #{agent}"))
+  diff = on(agent, "diff #{studio}/source-file-old #{studio}/source-file-backup").exit_code
+  assert_equal(0, diff, 'source-file was not backed up to filebucket')
 end
 
 step "Validate binary-file" do
-  md5 = on(agent, "openssl dgst -md5 #{studio}/binary-file").stdout.chomp
-  assert_equal("MD5(#{studio}/binary-file)= 7cfc2db80222ef224d99648716cea8e4", md5)
-  # TODO test filebucket backup
+  md5 = on(agent, "openssl dgst -md5 #{studio}/binary-file | awk '{print $2}'").stdout.chomp
+  assert_equal('7cfc2db80222ef224d99648716cea8e4', md5)
+end
+
+step "Validate binary-file filebucket backup" do
+  old_md5 = on(agent, "openssl dgst -md5 #{studio}/binary-file-old | awk '{print $2}'").stdout.chomp
+  on(agent, puppet("filebucket restore #{studio}/binary-file-backup #{old_md5} --server #{agent}"))
+  diff = on(agent, "diff #{studio}/binary-file-old #{studio}/binary-file-backup").exit_code
+  assert_equal(0, diff, 'binary-file was not backed up to filebucket')
 end
 
 step "Validate recursive-directory" do
@@ -101,8 +107,8 @@ step "Validate mount-point-source-file" do
 end
 
 step "Validate mount-point-binary-file" do
-  md5 = on(agent, "openssl dgst -md5 #{studio}/mount-point-binary-file").stdout.chomp
-  assert_equal("MD5(#{studio}/mount-point-binary-file)= 4b392568e0c19886bf274663a63b7d18", md5)
+  md5 = on(agent, "openssl dgst -md5 #{studio}/mount-point-binary-file | awk '{print $2}'").stdout.chomp
+  assert_equal('4b392568e0c19886bf274663a63b7d18', md5)
 end
 
 step "Validate custom-fact-output" do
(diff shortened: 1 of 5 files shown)


acceptance/suites/puppet3_tests/020_backwards_compat/test.rb
```diff
@@ -0,0 +1,118 @@
...
+  fileserverconf = on(master, puppet("config print fileserverconfig")).stdout.chomp
+  create_remote_file(master, fileserverconf, <<FILESERVERCONF)
+[simmons_custom_mount_point]
+path /etc/puppetlabs/code/environments/production/modules/simmons/mount-point-files
+allow *
+FILESERVERCONF
+  on(master, "chmod 644 #{fileserverconf}")
+end
+
+step "Disable auth.conf" do
```
erikPrime: Do we disable auth.conf often enough that it ought to be in our helper library?

erikPrime:
I agree that we should move towards "making each "Validate" step a separate acceptance test so that we can run through all validations even if one of them fails."

There should probably be a comment with a linky to the simmons repo around here somewhere.

acceptance/suites/puppet3_tests/020_backwards_compat/test.rb
jeffmccune: This is mostly just for my own curiosity, but how difficult would it make these tests if we _didn't_ disable auth.conf?  My thinking is that the relaxed rules are pretty far from a representative production deployment, which makes me a little less confident about these tests mimicking the behavior a user would see in a typical deployment.

acceptance/suites/puppet3_tests/020_backwards_compat/test.rb
```diff
@@ -0,0 +1,118 @@
...
+  simmons_version = '0.2.0'
+  on(master, puppet("module install nwolfe-simmons --version #{simmons_version}"))
+  on(master, puppet("module list"))
+end
+
+step "Configure file serving" do
+  fileserverconf = on(master, puppet("config print fileserverconfig")).stdout.chomp
+  create_remote_file(master, fileserverconf, <<FILESERVERCONF)
+[simmons_custom_mount_point]
+path /etc/puppetlabs/code/environments/production/modules/simmons/mount-point-files
```
camlow325: To make this a little more generic would it make sense to use the `environmentpath` variable to find the base directory for the production environment?

acceptance/suites/puppet3_tests/020_backwards_compat/test.rb
camlow325: I'd generally agree with @jeffmccune here that executing with the default set of auth.conf rules in place would be ideal.  I'm also curious what specifically in the auth.conf has to be disabled in order for these tests to work.

acceptance/suites/puppet3_tests/020_backwards_compat/test.rb
```diff
@@ -0,0 +1,118 @@
...
+step "Perform agent upgrade steps" do
+  step "Enable structured facts" do
+    agents.each do |agent|
+      on(agent, puppet("config set stringify_facts false --section agent"))
+    end
+  end
+end
+
+step "Configure site.pp" do
+  sitepp = '/etc/puppetlabs/code/environments/production/manifests/site.pp'
```
camlow325: Similar comment here as above about how this path could be derived from `environmentpath` instead of being fully-qualified.

acceptance/suites/puppet3_tests/020_backwards_compat/test.rb
nwolfe: That returns a list of paths, so I'd have to split the string and then either assume it's in the first one, or search under all paths for the module.  Not sure which is better, but that's why I hardcoded it against the specification for the time being.

acceptance/suites/puppet3_tests/020_backwards_compat/test.rb
```diff
@@ -0,0 +1,118 @@
...
+  studio => "#{studio}",
+}
+SITEPP
+  on(master, "chmod 644 #{sitepp}")
+end
+
+agent = master
+
+step "Run agents" do
+  with_puppet_running_on(master, {}) do
```
camlow325: When this first Puppet agent run occurs, it should have to negotiate with the master to get a certificate - agent submits a cert request, master signs the cert, agent gets the signed cert from the master - right?  Does the master need to configured for { "autosign" => "true" } in order for the cert to be signed automatically.  Looks like we do that along with setting some other properties for the master in the `with_puppet_running_on` call which happens from `puppetserver_initialize_ssl` from `.../acceptance/lib/helper.rb`, for example.

acceptance/suites/puppet3_tests/020_backwards_compat/test.rb
nwolfe: I also tried setting the relevant path to a well-known one to avoid all of this discovery, but turns out doing that for modules specifically is a bit challenging and involved.  Coding against the specification turned out to be the path of least resistance (at least at the time), but I'm happy to improve this.

camlow325: I'm okay with it as is.  Had forgotten that the setting could have multiple values in it.

acceptance/suites/puppet3_tests/020_backwards_compat/test.rb
nwolfe: Yeah this was what I was referring to today in standup, about quirks that I'll need to work out once I actually run the 3.x agent against the 4.x master.  Right now, I have the 4.x agent running against the 4.x master for development, so that happens to avoid this issue since it's re-using the master's certificate.

Yeah I'll have to turn autosigning probably

nwolfe:
@camlow325 I just added a couple more commits that included filebucket tests - this is all I've got now.  I suppose we can continue to review this PR for the sake of feedback, but I'll probably just open a new PR with everybody prim and proper.

dev/scripts/server545_puppet3_compatibility.sh
```diff
@@ -5,7 +5,7 @@
 
 
 # Puppet Server package version (from http://builds.puppetlabs.lan/puppetserver/)
-export PACKAGE_BUILD_VERSION='2.1.0.SNAPSHOT.2015.04.17T0057'
+export PACKAGE_BUILD_VERSION='2.1.0.SNAPSHOT.2015.05.11T0411'
```
camlow325: This will be eventually done dynamically be Jenkins job parameter configuration so the latest will be picked up automatically, right?

nwolfe: This script isn't used at all by Jenkins actually.  Might end up removing it from the repo once we prove out the pipeline.

acceptance/suites/puppet3_tests/020_backwards_compat/test.rb
```diff
@@ -0,0 +1,124 @@
...
+step "Validate custom-fact-output" do
+  expected = on(agent, "hostname").stdout.chomp
+  content = on(agent, "cat #{studio}/custom-fact-output").stdout.chomp
+  assert_equal(expected, content)
+end
+
+step "Validate external-fact-output" do
+  expected = on(agent, "hostname").stdout.chomp
+  content = on(agent, "cat #{studio}/external-fact-output").stdout.chomp
+  assert_equal(expected, content)
```
camlow325: As discussed with @nwolfe, this test will fail on Puppet 4 because of a defect with external, executable facts not being set executable on the agent after having been downloaded.  We'll probably need to set this one to skip with a pointer to the PUP ticket against the agent that will fix this problem.

nwolfe:
Closing in favor of #542

acceptance/suites/puppet3_tests/020_backwards_compat/test.rb
nwolfe: Now that I've got everything stabilized I could probably flesh out a more real-world auth.conf in #542 if we think that's worthwhile.

PR #536, opened 2015-05-14
(QENG-2385) Use install_puppet to install puppet.

waynr:
As recommended by the Beaker developers.

PR #537, opened 2015-05-14
(QENG-2385) Use different install approach for each platform.

waynr:


waynr:
Tested successfully using a `redhat6-64ma-debian6-64a-windows2008r2-64a` SUT layout.

waynr:
Closing since #536 was merged

PR #538, opened 2015-05-14
(MAINT) Add simple beaker test run scripts.

waynr:
These scripts are what I used to fix the puppet3compat presuite, I know there has been reservation expressed regarding the use of genconfig since it isn't yet open source but they probably generally need a little additional work anyway. Just opening the PR in case anyone is interested in an alternate approach that makes it quite a bit easier to test against different platforms.

### DIFF
--- a/acceptance/scripts/generic/testrun-full.sh
+++ b/acceptance/scripts/generic/testrun-full.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+set -x
+
+export GEM_SOURCE=http://rubygems.delivery.puppetlabs.net
+
+export GENCONFIG_LAYOUT="${GENCONFIG_LAYOUT:-redhat6-64ma-debian6-64a-windows2008r2-64a}"
+export BEAKER_TESTSUITE="${BEAKER_TESTSUITE:-acceptance/suites/tests/00_smoke}"
+export BEAKER_PRESUITE="${BEAKER_PRESUITE:-acceptance/suites/pre_suite/foss}"
+export BEAKER_POSTSUITE="acceptance/suites/post_suite/"
+export BEAKER_OPTIONS="acceptance/config/beaker/options.rb"
+export BEAKER_CONFIG="acceptance/scripts/hosts.cfg"
+export BEAKER_KEYFILE="~/.ssh/id_rsa-acceptance"
+export BEAKER_HELPER="acceptance/lib/helper.rb"
+
+bundle install --path vendor/bundle
+
+bundle exec genconfig2 $GENCONFIG_LAYOUT > $BEAKER_CONFIG
+
+bundle exec beaker \
+  --config $BEAKER_CONFIG \
+  --type aio \
+  --helper $BEAKER_HELPER \
+  --options-file $BEAKER_OPTIONS \
+  --tests $BEAKER_TESTSUITE \
+  --post-suite $BEAKER_POSTSUITE \
+  --pre-suite $BEAKER_PRESUITE \
+  --keyfile $BEAKER_KEYFILE \
+  --load-path "acceptance/lib" \
+  --preserve-hosts always \
+  --debug \
+  --timeout 360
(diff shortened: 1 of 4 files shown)


acceptance/scripts/generic/testrun-full.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+set -x
+
+# static fixtures
+REDIS_HOSTNAME=redis.delivery.puppetlabs.net
```
KevinCorcoran: This appears to be unused?

waynr: Ah right, it was a copy-pasta error since I based much of this script on one from another project where I had to install the latest version of PE listed in redis.

waynr: This one: https://github.com/puppetlabs/pe-rbac-service/blob/master/integration/scripts/generic/testrun-full.sh#L9-L12

acceptance/scripts/test-noprovision-puppet3compat.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -x
+
+export BEAKER_TESTSUITE="acceptance/suites/puppet3_tests/}"
```
nwolfe: Trailing `}` here

nwolfe:
Ah, cool.  I was able to run the puppet 3 compatibility tests locally with these scripts, and I think in general they would be valuable to have around.

I was thinking I would have to manually install `genconfig2` somehow on my system, but I realized this is provided by the `sqa-utils` gem, which we already have in our Gemfile.  As long as this means developers won't need to manually install/configure this thing, I'm okay with using like this.

:+1:

waynr:
@nwolfe 
One objection to scripts like these that has been brought up in the past is the fact that they depend on a closed-source project--the `sqa-utils` gem is only available on our internal rubygems mirror. I have ticketed out QENG work to open source genconfig2 but it won't be anywhere near the top of my priority list at least until July or August.

waynr:
@nwolfe 
Without too much additional work we can set these scripts up to re-use the same SUTs also FYI, but that's usually a bigger deal for PE projects since PE takes so long to install.

nwolfe:
> One objection to scripts like these that has been brought up in the past is the fact that they depend on a closed-source project--the sqa-utils gem is only available on our internal rubygems mirror

I don't understand how that would really matter, unless we were somehow trying to provide these for community members to use?

> Without too much additional work we can set these scripts up to re-use the same SUTs also FYI, but that's usually a bigger deal for PE projects since PE takes so long to install.

Considering that we don't really expect to be running these that often I'm not sure it's worth the trouble.
If we find ourselves doing this manually often enough we can automate it.

nwolfe:
@waynr Whoops - could you please retarget this at master?

waynr:
@nwolfe sure, but you know you can open PRs from branches on my fork right?

nwolfe:
@waynr (woah) - you're totally right, I never even considered this a possibility! Good to know

jeffmccune:
> > One objection to scripts like these that has been brought up in the past is the fact that they depend on a closed-source project--the sqa-utils gem is only available on our internal rubygems mirror

> I don't understand how that would really matter, unless we were somehow trying to provide these for community members to use?

FWIW, it's not closed or open that I objected to.  It's the incompatibility with other tools like `pry`.  For example, try:

```shell
$ echo 'gem "pry"' > Gemfile.local
$ bundle install
```

Sometime later, you need genconfig2

```shell
$ export GEM_SOURCE=http://rubygems.delivery.puppetlabs.net
$ bundle install --path .bundle/gems/
Fetching source index from http://rubygems.delivery.puppetlabs.net/
Resolving dependencies...
Could not find gem 'pry-nav (>= 0) ruby' in the gems available on this machine.
```

It's this specific issue that I'd like to fix.

waynr:
@jeffmccune oh sorry, I misremembered but do recall you bringing that up before. For what it's worth, QENG-2050 + RE-4181 should allow us to replace stickler as our internal rubygems in a way that will remove the need for manually pushing every necessary gem manually.

PR #547, opened 2015-05-18
(SERVER-653) Skip foreground test on RHEL5

jpinsonault:
The `timeout` command isn't available on RH5, so skip it


camlow325:
:+1:

PR #548, opened 2015-05-19
(SERVER-655) Fix Bouncy Castle jar load failure

camlow325:
This commit has changes which allow Puppet Server to boot without
JRuby's jopenssl trying to load its own Bouncy Castle jar.

Prior to JRuby 1.7.20, we set the JARS_NO_REQUIRE environment variable
onto the ScriptingContainer in order to instruct JRuby's
jar-dependencies to not try to load the Bouncy Castle jar - which we
intentionally exclude from the Puppet Server uberjar so that
'jvm-ssl-utils' can pull in its own version.

In JRuby 1.7.20's jopenssl, jar-dependencies was not being required
implicitly and so the JARS_NO_REQUIRE variable was ignored.  This
led Puppet Server to fail to boot because the Bouncy Castle jar
could not be loaded.

This commit has Puppet Server do an explicit require of
jar-dependencies so that the JARS_NO_REQUIRE variable can be honored
properly again.  It also sets JARS_REQUIRE, as that appears to be
the environment variable which jar-dependencies will eventually
replace JARS_NO_REQUIRE with.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/cli/gem.clj
+++ b/src/clj/puppetlabs/puppetserver/cli/gem.clj
@@ -9,7 +9,9 @@
     (.setEnvironment
       (hash-map
         "GEM_HOME" (get-in config [:jruby-puppet :gem-home])
-        "JARS_NO_REQUIRE" "true"))
+        "JARS_NO_REQUIRE" "true"
+        "JARS_REQUIRE" "false"))
+    (.runScriptlet "require 'jar-dependencies'")
     (.runScriptlet "load 'META-INF/jruby.home/bin/gem'")))
 
 (defn -main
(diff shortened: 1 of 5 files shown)


KevinCorcoran:
This is disgusting, but :+1:

camlow325:
For consistency, I updated the gem, irb, and ruby subcommands to use the same approach for requiring 'jar-dependencies' and setting the additional JARS_REQUIRE environment variable.  I didn't bother duplicating the comments about why this was needed over to the subcommands - mostly since I'm hoping this stuff still will get consolidated soon as part of the work in SERVER-297.  In any event, I did exercise each of these subcommands from the command line locally to see that they all still basically work with the changes.  For that, I just ran the subcommands via an uberjar that I'd built locally with the changes.

nwolfe:
:+1:

PR #549, opened 2015-05-19
(SERVER-647) Make master-{code,run,log}-dir settings optional

jeffmccune:
Without this patch puppetserver fails to start with with an uncaught exception
when these three settings are not defined.  This is a problem because these
settings are strictly necessary because they are defined in the underlying
puppet implementation.  This patch addresses the problem by settings the values
to nil which will cause puppetserver to consult puppet for the values.

The exception this fixes looks like:

    clojure.lang.ExceptionInfo: Output of initialize-config does not match schema:
    {:master-code-dir missing-required-key,
     :master-log-dir missing-required-key,
     :ruby-load-path missing-required-key,
     :master-run-dir missing-required-key}
    at puppetlabs.services.jruby.jruby_puppet_core$eval14631$initialize_config__14632.invoke (jruby_puppet_core.clj:84)
    ...


camlow325:
:+1:

PR #552, opened 2015-05-20
(SERVER-655) Fix Bouncy Castle cli command failure

camlow325:
This commit adds onto the changes made in the first PR for SERVER-655.
In the first PR, the ruby subcommand was still errantly trying to load
the JRuby Bouncy Castle jar when a require was passed via a "-r" flag on
the command line.  The same behavior was not occurring for the irb
subcommand, though.  The difference was that the irb subcommand already
had logic to front load the Ruby load path with the list of libraries
that JRuby uses for loading its intenal libs.  One of those paths is the
path to jar-dependencies.  This commit addresses that and syncs up the
irb and ruby subcommands to basically have the same logic.


KevinCorcoran:
This looks good to me, gonna merge it so we can see what happens in CI.

PR #553, opened 2015-05-20
(maint) Remove legacy-service dependency on Master and CA for PE

jeffmccune:
Without this patch the legacy-routes-service specifies a dependency on the
master-service and the ca-service.  This is a problem because the
master-service is not implemented in Puppet Enterprise, instead the
pe-master-service is implemented.  This patch addresses the problem by removing
the dependency on the master service which will allow the legacy-service to
operate in PE.

The dependency is no longer necessary because the catch-all 404 handler has
been removed from the master-service which allows either service to handle
request before handing them to the other service.


KevinCorcoran:
:+1: assuming Travis CI passes

KevinCorcoran:
I know that @dankreek was good with this as well, so :shipit:

PR #551, opened 2015-05-19
(SERVER-648) Merge 2.1.x to master

dankreek:
Since SERVER-647 is still open the unit tests will probably fail.

PR #528, opened 2015-05-08
(SERVER-614) Initialize timeouts from ClientOptions

camlow325:
This commit corrects a problem with the initialization of HTTP client
options which had been introduced by SERVER-530, the merge-up of
changes from 1.0.8 into the 2.1.x branch.  The merge-up had caused
HTTP requests to try to call the `setConnectTimeoutMilliseconds` and
`setIdleTimeoutMilliseconds` methods on the `RequestOptions` class.
`RequestOptions` does not support those methods; they can only
be configured at the client and not a per-request level.  To adjust
for this, this PR moves the invocation of the timeout methods to the
client level via `ClientOptions`.

PR #554, opened 2015-05-21
(MAINT) Replace http client timeout settings

dankreek:


camlow325:
Code changes look good.

I did some functional testing of the connect and idle timeouts.  We had a bug in the master and 2.1.x branches where invoking the timeout methods fail because they are being called on the wrong class.  That was done in #528 for the 2.1.x branch and merged to the master branch in #551.  This branch wasn't rebased on master with those commits, though, so you'd see the same failure from before #528 when testing with this branch as is.  I locally cherry-picked this commit onto the latest code from the master branch, though, and saw that everything was working okay.

This problem exists in both the stable and master branches of Puppet Server.  I'm going to close this PR, cherry pick the commit into a new branch targetted at the stable branch, and push up a PR.  We can then merge up from stable to master from there.

dankreek:
Whoa dang, thanks for finding that, @camlow325

camlow325:
I replaced this with #559, targeted for stable.

PR #556, opened 2015-05-21
(SERVER-324) Add support for flushing a single environment

cprice404:
This commit extends our existing functionality around flushing
the Puppet environment cache.  Prior to this commit, the only
granularity that we allowed the cache to be flushed at was effectively
global.  This commit introduces functions for flushing an
individual environment, wires those functions all the way up through
the JRubyService protocol, and extends the puppet-admin-api HTTP endpoint
to support passing in an environment name as a query parameter.

### DIFF
--- a/documentation/admin-api/v1/environment-cache.markdown
+++ b/documentation/admin-api/v1/environment-cache.markdown
@@ -15,6 +15,14 @@ endpoint to the master's HTTP API:
 To trigger a complete invalidation of the data in this cache, make an HTTP
 request to this endpoint.
 
+### Query Parameters
+
+(Introduced in Puppet Server 1.1/2.2)
+
+This endpoint accepts an optional query parameter, `environment`, whose value
+may be set to the name of a specific Puppet environment.  If this parameter
+is provided, only the specified environment will be flushed from the cache,
+as opposed to all environments.
 
 ### Response
 
@@ -26,6 +34,9 @@ The response body will be empty.
 
 ~~~
 $ curl -i --cert <PATH TO CERT> --key <PATH TO KEY> --cacert <PATH TO PUPPET CA CERT> -X DELETE https://localhost:8140/puppet-admin-api/v1/environment-cache
+HTTP/1.1 204 No Content
+
+$ curl -i --cert <PATH TO CERT> --key <PATH TO KEY> --cacert <PATH TO PUPPET CA CERT> -X DELETE https://localhost:8140/puppet-admin-api/v1/environment-cache?environment=production
 HTTP/1.1 204 No Content
 ~~~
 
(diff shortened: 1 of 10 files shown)


documentation/admin-api/v1/environment-cache.markdown
```diff
@@ -15,6 +15,14 @@ endpoint to the master's HTTP API:
 To trigger a complete invalidation of the data in this cache, make an HTTP
 request to this endpoint.
 
+### Query Parameters
+
+(Introduced in Puppet Server 1.1/2.2)
```
KevinCorcoran: Should be 2.1, not 2.2.

test/integration/puppetlabs/services/jruby/puppet_environments_int_test.clj
```diff
@@ -193,3 +193,41 @@
...
+                                              {:max-active-instances 1}}
+      ;;; Validate that the catalog has `hello1`
+      (let [catalog1 (get-catalog)]
+        (is (catalog-contains? catalog1 "Notify" "hello1"))
+        (is (not (catalog-contains? catalog1 "Notify" "hello2"))))
+
+      ;; Now we modify the class definition to have a 'hello2' notify,
+      ;; instead of 'hello1'.
+      (write-foo-pp-file "class foo { notify {'hello2': } }")
+
```
KevinCorcoran: I think that this test should get another catalog right here, before flushing the env, and validate that it's the same as the previous catalog (indicating that the changes to the manifest aren't picked-up until the cache is flushed).

KevinCorcoran: When I wrote my comment above, I hadn't yet realized that the first request below specified a different environment - given that, I don't feel so strongly about adding the extra validation I mentioned - I'd be fine adding it or not.

KevinCorcoran:
Code changes look fine to me.  But I don't think this should be merged until it's decided whether or not it can be included in 1.1/2.1.

src/clj/puppetlabs/puppetserver/ring/middleware/params.clj
```diff
@@ -91,3 +91,19 @@
     (if (:query-params request)
       request
       (assoc-query-params request encoding))))
+
+(defn wrap-params
+  "Middleware to parse urlencoded parameters from the query string and form
```
jeffmccune: s/form/from/

jeffmccune:
:+1: other than my minor comments.

src/clj/puppetlabs/puppetserver/ring/middleware/params.clj
KevinCorcoran: @jeffmccune Actually, I think `form` is correct here - i.e. `url-encoded form`.

nwolfe: I think both `from` and `form` are correct here :smirk:

KevinCorcoran: :clap:

KevinCorcoran:
At @cprice404's request, I'm going to close this in favor of #558.

PR #557, opened 2015-05-21
(SERVER-657) Bump lein-ezbake dependency to 0.3.6

camlow325:
This commit bumps the lein-ezbake dependency to version 0.3.6.  This is
needed in order to allow Puppet Server packaging to include an updated
version of the foreground command which allows the HOME environment
variable to be set for the puppet user when the foreground subcommand is
executed.  Without this change, attempts to execute the foreground
subcommand on Ubuntu via sudo would fail because an inappropriate value
for the HOME environment variable was being used.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -94,7 +94,7 @@
              :ezbake {:dependencies ^:replace [[puppetlabs/puppetserver ~ps-version]
                                                [puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty-version]
                                                [org.clojure/tools.nrepl "0.2.3"]]
-                      :plugins [[puppetlabs/lein-ezbake "0.3.4"]]
+                      :plugins [[puppetlabs/lein-ezbake "0.3.6"]]
                       :name "puppetserver"}
              :uberjar {:aot [puppetlabs.trapperkeeper.main]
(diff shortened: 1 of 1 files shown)


KevinCorcoran:
:+1: 

Not going to wait for Travis CI since he wouldn't catch any problems with ezbake anyway.

jeffmccune:
I tossed up https://github.com/puppetlabs/pe-puppet-server-extensions/pull/204 because I think we'll want this for SG too, or not?

camlow325:
@jeffmccune Yeah, that makes sense.  Thanks for thinking of that!

KevinCorcoran:
Travis CI failures look like nothing to worry about ...
> $ git fetch origin +refs/pull/557/merge:
fatal: Couldn't find remote ref refs/pull/557/merge
The command "eval git fetch origin +refs/pull/557/merge:" failed. Retrying, 2 of 3.
fatal: Couldn't find remote ref refs/pull/557/merge
The command "eval git fetch origin +refs/pull/557/merge:" failed. Retrying, 3 of 3.
fatal: Couldn't find remote ref refs/pull/557/merge
The command "eval git fetch origin +refs/pull/557/merge:" failed 3 times.
The command "git fetch origin +refs/pull/557/merge:" failed and exited with 128 during .

PR #559, opened 2015-05-21
(SERVER-659) Restore broken http client timeout settings

camlow325:
This commit restores the ability for connect and idle timeout settings
configured for Puppet Server to be honored for HTTP client requests.
The code to support these settings had previously been inadvertently
removed as part of some refactor work for SERVER-325.

### DIFF
--- a/test/integration/puppetlabs/services/jruby/jruby_puppet_internal_test.clj
+++ b/test/integration/puppetlabs/services/jruby/jruby_puppet_internal_test.clj
@@ -1,0 +1,34 @@
+(ns puppetlabs.services.jruby.jruby-puppet-internal-test
+  (:require [clojure.test :refer :all]
+            [puppetlabs.services.jruby.jruby-puppet-internal :as jruby-internal]
+            [puppetlabs.services.jruby.jruby-testutils :as jruby-testutils])
+  (:import (java.util.concurrent LinkedBlockingDeque)))
+
+  (deftest ^:integration settings-plumbed-into-jruby-container
+    (testing "setting plumbed into jruby container for"
+      (let [pool (LinkedBlockingDeque. 1)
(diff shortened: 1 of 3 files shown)


test/integration/puppetlabs/services/jruby/jruby_puppet_internal_test.clj
```diff
@@ -0,0 +1,34 @@
+(ns puppetlabs.services.jruby.jruby-puppet-internal-test
+  (:require [clojure.test :refer :all]
+            [puppetlabs.services.jruby.jruby-puppet-internal :as jruby-internal]
+            [puppetlabs.services.jruby.jruby-testutils :as jruby-testutils])
+  (:import (java.util.concurrent LinkedBlockingDeque)))
+
+  (deftest ^:integration settings-plumbed-into-jruby-container
+    (testing "setting plumbed into jruby container for"
```
dankreek: So... I know I wrote this, but should we complete the doc string?

camlow325: Hmmm, I think I wrote this?  :smile: I did it here as "setting plumbed into jruby container for" so that when the individual assertions would fail below that the result would be concatenated into the output, i.e., if the "connect timeout milliseconds" test's assertion were to fail, the testing string written to test output would be "setting plumbed into jruby container for http_connect_timeout_milliseconds".  Does that make sense or were you getting at something else?

dankreek: Ohhhh hah. Wow that was maybe my most wrong comment ever. Carry on, carry on!

dankreek:
:+1: LGTM

jeffmccune:
Looks familiar...  :+1:

PR #560, opened 2015-05-21
(SERVER-324) #556 + a docs fix

KevinCorcoran:
@cprice404 asked me to make any small changes required to #556 in the interest of getting it merged ASAP. I've simply corrected the version number in the docs.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/ring/middleware/params.clj
+++ b/src/clj/puppetlabs/puppetserver/ring/middleware/params.clj
@@ -91,3 +91,19 @@
     (if (:query-params request)
       request
       (assoc-query-params request encoding))))
+
+(defn wrap-params
+  "Middleware to parse urlencoded parameters from the query string and form
+  body (if the request is a url-encoded form). Adds the following keys to
+  the request map:
+  :query-params - a map of parameters from the query string
+  :form-params  - a map of parameters from the body
+  :params       - a merged map of all types of parameter
+  Accepts the following options:
+  :encoding - encoding to use for url-decoding. If not specified, uses
+              the request character encoding, or \"UTF-8\" if no request
+              character encoding is set."
+  {:arglists '([handler] [handler options])}
+  [handler & [options]]
+  (fn [request]
+    (handler (params-request request options))))
--- a/src/clj/puppetlabs/services/jruby/puppet_environments.clj
+++ b/src/clj/puppetlabs/services/jruby/puppet_environments.clj
@@ -4,7 +4,8 @@
 
 (defprotocol EnvironmentStateContainer
   (environment-state [this])
-  (mark-all-environments-expired! [this]))
+  (mark-all-environments-expired! [this])
+  (mark-environment-expired! [this env-name]))
 
 (defn environment-registry
   []
@@ -36,4 +37,7 @@
       (environment-state [this] state)
       (mark-all-environments-expired! [this]
         (log/info "Marking all registered environments as expired.")
-        (swap! state mark-all-expired!)))))
+        (swap! state mark-all-expired!))
+      (mark-environment-expired! [this env-name]
+        (log/infof "Marking environment '%s' as expired." env-name)
+        (swap! state mark-expired! (keyword env-name))))))
--- a/src/clj/puppetlabs/services/puppet_admin/puppet_admin_core.clj
+++ b/src/clj/puppetlabs/services/puppet_admin/puppet_admin_core.clj
@@ -5,8 +5,10 @@
             [puppetlabs.puppetserver.liberator-utils :as liberator-utils]
             [schema.core :as schema]
             [liberator.core :refer [defresource]]
-            ;[liberator.dev :as liberator-dev]
-            [puppetlabs.comidi :as comidi]))
+    ;[liberator.dev :as liberator-dev]
+            [puppetlabs.comidi :as comidi]
+            [clojure.tools.logging :as log]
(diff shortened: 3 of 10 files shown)


src/clj/puppetlabs/puppetserver/ring/middleware/params.clj
```diff
@@ -91,3 +91,19 @@
     (if (:query-params request)
       request
       (assoc-query-params request encoding))))
+
+(defn wrap-params
```
camlow325: Looking forward to going back and nuking this someday.

src/clj/puppetlabs/services/puppet_admin/puppet_admin_core.clj
```diff
@@ -5,8 +5,10 @@
             [puppetlabs.puppetserver.liberator-utils :as liberator-utils]
             [schema.core :as schema]
             [liberator.core :refer [defresource]]
-            ;[liberator.dev :as liberator-dev]
-            [puppetlabs.comidi :as comidi]))
+    ;[liberator.dev :as liberator-dev]
+            [puppetlabs.comidi :as comidi]
+            [clojure.tools.logging :as log]
```
camlow325: Unused?  WP violation?

KevinCorcoran: Is that Wolfe's Second or Third Principle?

src/clj/puppetlabs/services/jruby/puppet_environments.clj
```diff
@@ -36,4 +37,7 @@
       (environment-state [this] state)
       (mark-all-environments-expired! [this]
         (log/info "Marking all registered environments as expired.")
-        (swap! state mark-all-expired!)))))
+        (swap! state mark-all-expired!))
+      (mark-environment-expired! [this env-name]
+        (log/infof "Marking environment '%s' as expired." env-name)
+        (swap! state mark-expired! (keyword env-name))))))
```
camlow325: So if an `env-name` passed in was not previously registered, should this return an error?  Seems like Ruby Puppet typically handles this case by returning an HTTP error with a message about the environment not existing - at least for the endpoints where the environment parameter is actually used.

KevinCorcoran: That's a good point.  It seems like this endpoint should return an error in the case when the specified environment doesn't exist.

camlow325:
:+1: other than the super-egregious WP violation and the question about what the behavior should be if an attempt is made to delete an unknown environment.  I'm willing to merge without resolving the unknown environment question and follow-up on that with an additional PR if others care enough to address that.

KevinCorcoran:
I tried to implement the behavior discussed above - check for existence of the environment, and error if not found.  I wasn't able to get it fully working - my attempt is here: https://github.com/puppetlabs/puppet-server/commit/8162df31235901c30b3e1ecc82b97082e7a3a301

I initially tried to just throw an Exception, but that doesn't work with liberator.  I think the problem with the current code has something to do with the liberator resource, either `:exists?` or `:handle-not-found`, but I am giving up and calling it a night.  A better option might be to just re-write this endpoint to not use liberator ... although that might be a bit more than we want to pull-in to the releases last-minute.

Alternatively, we could just call this "good enough" for now, ship it like this, and then come back and do the liberator removal + extra error handling.

cprice404:
@KevinCorcoran the approach you were trying would not work; the keys in that environment registry only get set when an environment is *loaded*, which happens lazily, when requests come in targeted a given environment.  There is no way to definitively know if an environment *exists* without calling into Puppet itself.

I do not believe that the costs associated with that implementation detail are worth the return on investment.  I also believe that we could change that behavior safely in a Z release if it were ever to be deemed important.  So I'd prefer that we do not burn time on that right now.

cprice404:
Also, I just did a curl for "/foo/node/mynodename", and Puppet 3 happily returned node data for environment 'foo' even though I have in no way defined that on disk.  That behavior might be different in Puppet 4, but our ecosystem's expected behavior for this case is at the very least ambiguous enough to where I don't feel it's a blocker for shipping this feature, personally.

camlow325:
> @KevinCorcoran the approach you were trying would not work; the keys in that environment registry only get set when an environment is loaded, which happens lazily, when requests come in targeted at a given environment.

Ah, sorry, my bad for suggesting he go in that direction.  I had been under the impression that core Ruby Puppet would call back into the hook we provide it to register / remove environments any time it becomes aware of a new one or disposes of one that it interprets to be going away, respectively.  If the environment cache in the Clojure-layer environment registry isn't kept in sync like that, though, I agree that validating the environment prior to a cache delete attempt would be significantly less useful / potentially error prone.

> There is no way to definitively know if an environment exists without calling into Puppet itself.

I definitely don't think we should be trying to do that here.  I think we should avoid calling into the core Ruby Puppet code to fulfill one of these requests - especially for cases where that request could get backed up for a significant amount of time on an attempt to borrow a JRubyPuppet pool instance, leaving any users of instances from that pool which get serviced ahead of the delete attempt to be using stale environment content for an indefinite amount of time.

> Also, I just did a curl for "/foo/node/mynodename", and Puppet 3 happily returned node data for environment 'foo' even though I have in no way defined that on disk.

Interesting, but not too surprising I guess.  I've definitely seen the environment name be validated on some other endpoints but I suppose the behavior as implemented on the PR right now would be no worse than the inconsistency in the existing endpoint implementations.

--

So... :+1: :ship:

PR #564, opened 2015-05-22
(MAINT) Add simple beaker test run scripts.

waynr:
@nwolfe here ya go, I removed the "noprovision" scripts since they actually depend on additional post-suite behavior.


nwolfe:
:+1: 

If we merge this we can get rid of https://github.com/puppetlabs/puppet-server/blob/master/dev/scripts/server545_puppet3_compatibility.sh

PR #565, opened 2015-05-22
(SERVER-325) Add docs for max-requests-per-instance

cprice404:


cprice404:
ping @camlow325 you were correct, I'd forgotten to add docs for the new setting.

ping @jbondpdx @nfagerlund for visibility; this is a new setting that is going in to the 1.1/2.1 releases of Puppet Server in the next week or so.  Let us know if you'd like us to create a ticket to track the docs work or anything.

ping @KevinCorcoran maybe you can help @camlow325 review and shepherd?  Will need to go into `master` as well.

KevinCorcoran:
:+1: 

As an aside, I wonder if it's a good idea to enable this behavior by default, with a reasonably-high value for `max-request-per-instance` - perhaps 100 or 1000.  The more instances we see of deployments running into memory leaks that would be fixed by this, the more I'd like to enable it by default - especially after we get some more validation of its usefulness in those scenarios.

camlow325:
:+1: and going to merge.  Will let @jbondpdx and @nfagerlund handle touch ups later, if desired.

nfagerlund:
OK cool, I just started on a minor edit. I'll PR in a sec.

PR #568, opened 2015-05-22
(MAINT) Merge stable to master (2015-05-22)

camlow325:
* stable:
      (docs) Minor revision of max-requests-per-instance description
      (SERVER-659) Restore broken http client timeout settings
      (SERVER-325) Add docs for max-requests-per-instance
      (SERVER-657) Bump lein-ezbake dependency to 0.2.10
      (MAINT) Updates to README_BRANCHING.md for 1.1/2.1 releases
      (maint) Correct release number in docs
      (SERVER-324) Add support for flushing a single environment

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
@@ -32,56 +32,63 @@
 
   The keys should have the following values:
 
-  * :ruby-load-path - a vector of file paths, containing the locations of puppet source code.
+    * :ruby-load-path - a vector of file paths, containing the locations of puppet source code.
 
-  * :gem-home - The location that JRuby gems are stored
+    * :gem-home - The location that JRuby gems are stored
 
-  * :master-conf-dir - file path to puppetmaster's conf dir;
-    if not specified, will use the puppet default.
+    * :master-conf-dir - file path to puppetmaster's conf dir;
+        if not specified, will use the puppet default.
 
-  * :master-code-dir - file path to puppetmaster's code dir;
-    if not specified, will use the puppet default.
+    * :master-code-dir - file path to puppetmaster's code dir;
+        if not specified, will use the puppet default.
 
-  * :max-active-instances - The maximum number of JRubyPuppet instances that
-      will be pooled.
-  
-  * :master-var-dir - path to the puppetmaster's var dir;
-    if not specified, will use the puppet default.
+    * :master-var-dir - path to the puppetmaster's var dir;
+        if not specified, will use the puppet default.
 
-  * :master-run-dir - path to the puppetmaster's run dir;
-    if not specified, will use the puppet default.
+    * :master-run-dir - path to the puppetmaster's run dir;
+        if not specified, will use the puppet default.
 
-  * :master-log-dir - path to the puppetmaster's log dir;
-    if not specified, will use the puppet default.
+    * :master-log-dir - path to the puppetmaster's log dir;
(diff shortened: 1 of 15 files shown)


camlow325:
@nfagerlund RE: the `documentation/configuration.markdown` changes in this PR, I noticed a couple of things that I didn't try to reconcile since this is just supposed to be a merge-up to master.  Wanted to mention those to you assuming you might want to follow-up on these with a separate PR:

1) With the unification of PE and FOSS paths that is going on around Puppet 4 and Puppet Server 2.0, we'll probably want to take a pass back through this to leave the appropriate paths in place, taking out the PE-specific references.

2) The “Enabling the Insecure SSLv3 Protocol” (master branch) and “Java Secure Socket Extension (JSSE) configuration” (stable branch) sections look very similar.  Probably should go with one format or the other for both branches?

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -32,56 +32,63 @@
...
-   :ruby-load-path                           (schema/both (schema/pred vector?) [schema/Str])
-   :gem-home                                 schema/Str
-   :master-conf-dir                          (schema/maybe schema/Str)
-   :master-code-dir                          (schema/maybe schema/Str)
-   :master-var-dir                           (schema/maybe schema/Str)
-   :master-run-dir                           (schema/maybe schema/Str)
-   :master-log-dir                           (schema/maybe schema/Str)
-   :http-client-ssl-protocols                [schema/Str]
-   :http-client-cipher-suites                [schema/Str]
+    * :ruby-load-path - a vector of file paths, containing the locations of puppet source code.
```
camlow325: Sorry for all of the spacing differences here.  Alignment and order was slightly different between the stable and master branches for these so the merge conflicts were hard to fix up without doing this.  On the plus side, hopefully any conflicts around these in the future should be easier to reconcile.

Highlights on the real differences, ignoring the spacing:

1) `:max-active-instances` was moved below `:master-log-dir` and the line within `:max-active-instances` *"If not specified, the system's number of CPUs+2 will be used"* was removed.

2) `:http-client-connect-timeout-milliseconds` and `:http-client-idle-timeout-milliseconds` text was added - should reflect stable branch implementation verbatim.

test/integration/puppetlabs/services/jruby/jruby_puppet_internal_test.clj
```diff
@@ -0,0 +1,35 @@
...
+              (is (= 5 (settings "http_idle_timeout_milliseconds"))))
+            (testing "cipher_suites"
+              (is (= ["TLS_RSA_WITH_AES_256_CBC_SHA256"
+                      "TLS_RSA_WITH_AES_256_CBC_SHA"]
+                    (into [] (settings "cipher_suites")))))
+            (testing "ssl_protocols"
+              (is (= ["TLSv1" "TLSv1.2"]
+                    (into [] (settings "ssl_protocols"))))))
+          (finally
+            (.terminate (:jruby-puppet instance))
```
camlow325: This line didn't exist in the stable branch implementation of this test.  The `terminate` function only exists in the master branch implementation of JRubyPuppet and should be called when disposing of a JRubyPuppet instance in order to ensure any resources it is holding on to are freed - most notably, the persistent Apache `CloseableHttpAsyncClient` instance that it can hold on to.

nfagerlund:
@camlow325 Cool. Actually, could you file DOC tickets for those, so we don't forget them? We can't get to them immediately.

jeffmccune:
:+1:

PR #572, opened 2015-05-26
(MAINT) Bump beaker dependency to 2.12.0

camlow325:
This commit bumps the beaker dependency to 2.12.0.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ gem 'jira-ruby', :group => :development
 
 group :test do
   gem 'rspec'
-  gem 'beaker', *location_for(ENV['BEAKER_VERSION'] || '~> 2.9.0')
+  gem 'beaker', *location_for(ENV['BEAKER_VERSION'] || '~> 2.12.0')
   if ENV['GEM_SOURCE'] =~ /rubygems\.delivery\.puppetlabs\.net/
     gem 'sqa-utils', '~> 0.11'
   end


nwolfe:
:+1:

Gemfile
```diff
@@ -15,7 +15,7 @@ gem 'jira-ruby', :group => :development
 
 group :test do
   gem 'rspec'
-  gem 'beaker', *location_for(ENV['BEAKER_VERSION'] || '~> 2.9.0')
+  gem 'beaker', *location_for(ENV['BEAKER_VERSION'] || '~> 2.12.0')
```
joshcooper: FWIW, we're tracking `'~> 2.8'` on the puppet side.

camlow325: Hmmm, is that what's actually being used for the platform side, though?  This recent job, for example, appears to be picking up 2.12.0:

https://jenkins.puppetlabs.com/view/All%20in%20One%20Agent/view/Master/view/Puppet/job/platform_aio-puppet_intn-sys_master/75/SLAVE_LABEL=beaker,TEST_TARGET=redhat-7-x86_64/consoleFull

joshcooper: Since we don't specify the z component, the y is allowed to float. So we're basically allowing versions >= 2.8 and less than 3

camlow325: Ah, k.  Missed the absent ".0" from your comment the first time I read it.  :smile:

PR #573, opened 2015-05-26
(MAINT) Add install options to 'install_dev_repos'

camlow325:
This commit configures options explicitly for the
'30_install_dev_repos.rb' for package installs.  This was done
specifically to configure the use of the 'PC1' repo since this is
required for use with AIO testing.  The need for this change was
introduced by the beaker refactoring done for BKR-184 which removed PC1
from the default list of repos that beaker checks during installation.


nwolfe:
@camlow325 Should we also update the corresponding file in the puppet3 compat pre-suite?

camlow325:
@nwolfe Pushed 98cd558 with an update to the corresponding file in the puppet3 compat pre-suite and also changed the options definition to merge in the one repos setting to override in place of the default options.  The latter change was made to avoid the need to redefine the base package url, for which Beaker already has a sane default.

nwolfe:
:+1:

PR #574, opened 2015-05-27
(SERVER-683) Dynamically get master service for legacy route service

camlow325:
This commit changes the logic in the legacy-routes-service to get the
route of the master-service via the service protocol name rather than by
a hard-coded service name.  This allows for the legacy-routes-service to
pull the route from whatever service implementing the MasterService
protocol happens to be in the service stack, i.e., master-service for
OSS or pe-master-service for PE.

### DIFF
--- a/src/clj/puppetlabs/services/legacy_routes/legacy_routes_service.clj
+++ b/src/clj/puppetlabs/services/legacy_routes/legacy_routes_service.clj
@@ -21,7 +21,14 @@
           puppet-version (get-in config [:puppet-server :puppet-version])
           ca-settings (ca/config->ca-settings (get-config))
           ca-mount (get-route (tk-services/get-service this :CaService))
-          master-mount (master-service/get-master-path config)
+          master-ns (keyword (tk-services/service-symbol
+                               (tk-services/get-service this :MasterService)))
+          master-route-config (master-core/get-master-route-config
+                                master-ns
+                                config)
+          master-mount (master-core/get-master-mount
+                         master-ns
+                         master-route-config)
           master-handler (-> (master-core/root-routes handle-request)
                              (#(comidi/context path %))
                              comidi/routes->handler
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -3,8 +3,6 @@
            (clojure.lang IFn))
   (:require [me.raynes.fs :as fs]
             [puppetlabs.puppetserver.ringutils :as ringutils]
-            [puppetlabs.dujour.version-check :as version-check]
-            [puppetlabs.services.config.puppet-server-config-core :as config]
             [puppetlabs.comidi :as comidi]
             [schema.core :as schema]))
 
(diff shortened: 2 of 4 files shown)


src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -64,19 +62,6 @@
     (comidi/GET ["/facts_search/" [#".*" :rest]] request
                    (request-handler request))))
 
-(defn construct-404-error-message
```
camlow325: Just pruning some dead code.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -142,11 +127,24 @@
       ringutils/wrap-response-logging
       (ringutils/wrap-with-puppet-version-header puppet-version)))
 
-(defn construct-invalid-request-handler
```
camlow325: More dead code.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -142,11 +127,24 @@
...
-  "Constructs a ring handler to handle an incorrectly formatted request and indicate to the user
-   they need to update to Puppet 4"
-  [error-message]
-  (fn
-    [_]
-    {:status 404
-     :body   error-message}))
+(schema/defn ^:always-validate get-master-route-config
```
camlow325: These functions were moved from the `master_service` namespace and altered a bit.  Thought it would be better for them to live here so that various users of these functions don't have to call into "service" layer code for a service that may not be loaded, e.g., when executed via the PE master service.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_service.clj
```diff
@@ -21,7 +21,14 @@
           puppet-version (get-in config [:puppet-server :puppet-version])
           ca-settings (ca/config->ca-settings (get-config))
           ca-mount (get-route (tk-services/get-service this :CaService))
-          master-mount (master-service/get-master-path config)
+          master-ns (keyword (tk-services/service-symbol
+                               (tk-services/get-service this :MasterService)))
```
KevinCorcoran: Do you intend to also change the PE Master Service to implement the `MasterService` protocol?  Currently, it implements `PEMasterService`, which means this won't work.  (hope I have that all sorted correctly ... could be that I'm missing something here.)

camlow325: Yep, that was my plan.  Would do that in a follow-up PR for PE.

fpringvaldsen:
This looks good to me, :+1:

jeffmccune:
:+1:

src/clj/puppetlabs/services/legacy_routes/legacy_routes_service.clj
```diff
@@ -21,7 +21,14 @@
...
+          master-route-config (master-core/get-master-route-config
```
dankreek: I know this has already been merged, but was curious if this functionality previously existed in the routing service and I just missed it?

camlow325: I don't think so.  IIRC, the current routing service would allow you to do a `get-route` for an arbitrary service - like is done for the `:CaService` above.  We still have the issue here, though, where the route payload for the `:MasterService` might be a map with the route value as one of the route ids vs. just being a string, and the `get-route` implementation forces you to specify in advance which one you want, throwing an `Exception` if the expectation isn't met.  I think this is why you filed https://tickets.puppetlabs.com/browse/TK-223, right?  To avoid having to write code that does up to two `get-route` calls and handles `Exception`s on failure, this PR basically preserves the logic that was in the implementation previously.

The only changes here are that the service symbol is determined dynamically from whatever service is found which implements the `:MasterService` protocol, and the implementation was moved down into the master core namespace.  The only reason I moved it into the core namespace is that it felt a little cleaner for it to be there given that we don't otherwise want to have any direct coupling between the `legacy_routes` service and the OSS `master` service since the OSS `master` service isn't even loaded in PE.

PR #575, opened 2015-05-28
(SERVER-682) Restore 'scan=true' to logback xml files

camlow325:
This commit restores the 'scan=true' attribute to the configuration
element in the logback and request-logging-dev xml files.  This had been
added previously as part of SERVER-275 but errantly lost in the
transition from ezbake to the lein-ezbake plugin between the Puppet
Server 1.0.2 and 1.0.8 releases.

### DIFF
--- a/dev/request-logging-dev.xml
+++ b/dev/request-logging-dev.xml
@@ -1,4 +1,4 @@
-<configuration scan="true" debug="false">
+<configuration debug="false" scan="true">
   <appender name="FILE" class="ch.qos.logback.core.FileAppender">
     <file>./target/logs/puppetserver-access.log</file>
     <encoder>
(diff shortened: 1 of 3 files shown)


camlow325:
I staged an ezbake build locally just to see that the "scan='true'" attribute should be present in the files with these changes.  I played around locally on my OSX laptop and didn't see that the presence of the "scan='true'" attribute actually allowed the logging configuration changes to be picked up without a restart - even when paired with a really short "scanPeriod" attribute.  Maybe the file watcher backing this has OS-specific issues, not sure.  In any event, this should restore us to compatibility with Puppet Server 1.0.2.

nwolfe:
:+1:

dankreek:
:+1:

PR #576, opened 2015-05-28
(SERVER-684) Add query param to file metadata requests

dankreek:


src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
```diff
@@ -38,6 +38,32 @@
...
+
+(defn- add-source-permissions-param
+  "Add the source_permissions header to the request. This is needed since
+  this was considered a default in Puppet 3.x and is now turned off by default
+  in Puppet 4.x. See SERVER-684."
+  [handler]
+  (fn [request]
+    (let [{:keys [query-string]} request
+          updated-request (assoc request :query-string
+                            (add-query-param
```
camlow325: If I understand this correctly, this would mean that if "source_permissions=X" were already in the query string that we would replace that with "source_permissions=use".  That may be irrelevant for Puppet 3.x agents if they never include the "source_permissions" parameter.  But if a client could include this parameter and the value it includes is trustworthy - i.e., we don't have the case where a client says "source_permissions=X" but really expects the server to process the request as though it had specified "source_permissions=use" - then I would rather have the implementation only add this query param if it wasn't already present in the request.  That way, we'd just be overriding the default behavior when the client doesn't specify a value rather than potentially processing the request with a value that the client did not intend.  @nwolfe - what do you think?

dankreek: Yeah, that's actually a much better idea in general. This comment has also pointed out that my docstring is incorrect. I'll get this fixed.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
```diff
@@ -46,21 +85,18 @@
   [handler mount-point api-version request]
   (let [{{environment :environment} :params
          uri :uri
-         query-string               :query-string} request
-        path-info (str "/" (-> (string/split uri #"/" 3) (nth 2)))
-        query-params (if (str/blank? query-string)
-                       {}
-                       (ring-codec/form-decode query-string))]
+         query-string :query-string} request
+        path-info (str "/" (-> (string/split uri #"/" 3) (nth 2)))]
```
camlow325: I guess this isn't new for this PR, but I'd prefer to not be doing our own hand-rolled URI parsing like this.  Could we instead construct a `java.net.URI` with `uri` as the argument and then call `.getRawPath` on that object to obtain the path?

jeffmccune: Bonus points for an idiomatic Clojure solution, maybe using [path-info](https://github.com/ring-clojure/ring/blob/1.3.1/ring-core/src/ring/util/request.clj#L66) from ring.util.request, or something similar?

jeffmccune:
:+1: other than the path-info comment.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
dankreek: Now that you pointed that out that smell, I think it stinks. I'll change it real quick.

dankreek: [It Stinks!](http://www.youtube.com/watch?v=DZR1gSvnOI0)

nwolfe:
Shoot, while you're in there I see we're pulling in `clojure.string` twice as `string` and `str` - maybe consolidate those? We're standardizing on `str` right?

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
dankreek: Actually now that I look back at it, the path is provided by ring. I'm just extracting the first element in the path using a regex. Not sure if `java.net.URI` would help much in that case.

dankreek: It's a bit misleading, but in this case the `:uri` key is actually just the path of the uri

dankreek:
That's legit nasty.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj
camlow325: Ah, okay that makes sense now.

camlow325:
"Exercised" the simmons module locally and saw that the external fact output worked properly.  The execute bits were preserved on the `external_fact_hostname` script that was downloaded into the client's .../var/facts.d directory and the hostname was dumped into the `external-fact-output` file.

So ... :+1: assuming Travis passes for the latest commits.

camlow325:
The current state of the branch ran against Travis once and passed the Clojure unit tests - https://travis-ci.org/puppetlabs/puppet-server/builds/64466580 - but failed "rake spec" tests due to the issue addressed in #579 .  The new run in process would probably result in the same outcome because it wasn't rebased after #579 was merged.  Local Clojure unit tests and "rake spec" tests - with the bundler version patched to 1.9.9 like in #579 - pass, so merging.

PR #579, opened 2015-05-28
(MAINT) Pin bundler version to 1.9.9

camlow325:
This commit pins the version of bundler that is used when executing spec
tests to 1.9.9.  Starting with version 1.10.0 of bundler, the .gemspec
file from core Ruby Puppet is automatically validated at 'bundle install'
startup.  The rubygems version in JRuby, 2.4.6, is not the same as the
rubygems version in the core Puppet repo's .gemspec file, 1.8.24.  This
causes validation to fail and, therefore, for the "bundle install" to
fail.  Pinning to bundler version 1.9.9 allows validation to be
bypassed for now.  See SERVER-697 for follow-up on long-term solution.


nwolfe:
:+1:

nwolfe:
Travis too slow

PR #581, opened 2015-05-29
(SERVER-682) Restore 'scan=true' to logback xml files

camlow325:
This commit restores the 'scan=true' attribute to the configuration
element in the logback and request-logging-dev xml files.  This had been
added previously as part of SERVER-275 but errantly lost in the
transition from ezbake to the lein-ezbake plugin between the Puppet
Server 1.0.2 and 1.0.8 releases.

### DIFF
--- a/dev/request-logging-dev.xml
+++ b/dev/request-logging-dev.xml
@@ -1,4 +1,4 @@
-<configuration scan="true" debug="false">
+<configuration debug="false" scan="true">
   <appender name="FILE" class="ch.qos.logback.core.FileAppender">
     <file>./target/logs/puppetserver-access.log</file>
     <encoder>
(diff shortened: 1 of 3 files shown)


camlow325:
Just realized that the previous PR I'd put up for this, #575, had errantly been targeted at master instead of stable and so wasn't picked up in the recent stable/master merge up.  This PR at least gets the changes on stable.

nwolfe:
> Just realized that the previous PR I'd put up for this, #575, had errantly been targeted at master instead of stable and so wasn't picked up in the recent stable/master merge up

Yep...dang.

:+1:

dankreek:
Dang, indeed! :+1:

PR #582, opened 2015-05-29
(SERVER-297) Consolidate JRuby environment handling

jeffmccune:
Without this patch the manner in which environment variables are handled in the
JRuby portions of puppet-server vary across the use cases of the `puppetserver
{gem,irb,ruby}` command line utilities in addition to the service itself.  This
is a problem because the behaviors have not kept in sync with one another over
time so there has been divergence at times, e.g. `puppetserver gem env`
currently fails because PATH is unset in that scenario but set in others, e.g.
the production service.

This patch addresses the problem by consolidating all of the environment
handling behavior into a single authoritative method.  The behavior itself
remains unchanged, GEM_HOME is set based on the config file value, PATH and
HOME are the only variables allowed to flow through to JRuby unmodified, all
other variables are filtered out, and JARS_REQUIRE and JARS_NO_REQUIRE are set
appropriately.

In addition to the environment handling, this patch also consolidates the JRuby
load path handling in a similar fashion across the scenarios of command line
utilities and the production service.

### DIFF
--- a/acceptance/suites/tests/010-puppetserver-cli/subcommand/SERVER-297_common_behavior.rb
+++ b/acceptance/suites/tests/010-puppetserver-cli/subcommand/SERVER-297_common_behavior.rb
@@ -1,0 +1,23 @@
+require 'test/unit/assertions'
+require 'json'
+
+test_name "Puppetserver subcommand consolidated ENV handling tests."
+
+step "ruby: Check that PATH, HOME, GEM_HOME JARS_REQUIRE and JARS_NO_REQUIRE are present"
+on(master, "puppetserver ruby -rjson -e 'puts JSON.pretty_generate(ENV.to_hash)'") do
+  env = JSON.parse(stdout)
+  assert(env['PATH'], "PATH missing")
+  assert(env['HOME'], "HOME missing")
+  assert(env['GEM_HOME'], "GEM_HOME missing")
+  assert(env['JARS_REQUIRE'], "JARS_REQUIRE missing")
+  assert(env['JARS_NO_REQUIRE'], "JARS_NO_REQUIRE missing")
+end
+
+step "irb: Check that PATH, HOME, GEM_HOME JARS_REQUIRE and JARS_NO_REQUIRE are present"
+on(master, "puppetserver irb -f -rjson -e 'puts JSON.pretty_generate(ENV.to_hash)'") do
+  assert_match(/\bPATH\b/, output, "PATH missing")
+  assert_match(/\bHOME\b/, output, "HOME missing")
+  assert_match(/\bGEM_HOME\b/, output, "GEM_HOME missing")
+  assert_match(/\bJARS_REQUIRE\b/, output, "JARS_REQUIRE missing")
+  assert_match(/\bJARS_NO_REQUIRE\b/, output, "JARS_NO_REQUIRE missing")
+end
--- a/src/clj/puppetlabs/puppetserver/cli/gem.clj
+++ b/src/clj/puppetlabs/puppetserver/cli/gem.clj
@@ -1,18 +1,10 @@
 (ns puppetlabs.puppetserver.cli.gem
-  (:import (org.jruby.embed ScriptingContainer))
-  (:require [puppetlabs.puppetserver.cli.subcommand :as cli]))
+  (:require [puppetlabs.puppetserver.cli.subcommand :as cli]
+            [puppetlabs.services.jruby.jruby-puppet-core :as jruby-core]))
 
 (defn run!
   [config args]
-  (doto (ScriptingContainer.)
-    (.setArgv (into-array String args))
-    (.setEnvironment
-      (hash-map
-        "GEM_HOME" (get-in config [:jruby-puppet :gem-home])
-        "JARS_NO_REQUIRE" "true"
-        "JARS_REQUIRE" "false"))
-    (.runScriptlet "require 'jar-dependencies'")
-    (.runScriptlet "load 'META-INF/jruby.home/bin/gem'")))
+  (jruby-core/cli-run! config "gem" args))
 
 (defn -main
   [& args]
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
@@ -51,7 +51,8 @@
    :size size})
 
 (schema/defn ^:always-validate managed-environment :- {schema/Str schema/Str}
-  "The environment variables that should be passed to the Puppet JRuby interpreters.
+  "The environment variables that should be passed to the Puppet JRuby
+  interpreters.
 
   We don't want them to read any ruby environment variables, like $RUBY_LIB or
   anything like that, so pass it an empty environment map - except - Puppet
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
@@ -16,8 +16,17 @@
 ;;; Definitions
 
 (def ruby-code-dir
-  "The name of the directory containing the ruby code in this project."
-  "src/ruby/puppet-server-lib")
+  "The name of the directory containing the ruby code in this project.
+
+  This directory is relative to `src/ruby` and works from source because the
+  `src/ruby` directory is defined as a resource in `project.clj` which places
+  the directory on the classpath which in turn makes the directory available on
+  the JRuby load path.  Similarly, this works from the uberjar because this
+  directory is placed into the root of the jar structure which is on the
+  classpath.
+
+  See also:  http://jruby.org/apidocs/org/jruby/runtime/load/LoadService.html"
+  "puppet-server-lib")
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Private
@@ -62,7 +71,7 @@
   (let [whitelist ["HOME" "PATH"]
         clean-env (select-keys env whitelist)]
     (assoc clean-env
-      "GEM_HOME" (fs/absolute-path gem-home)
+      "GEM_HOME" gem-home
       "JARS_NO_REQUIRE" "true"
       "JARS_REQUIRE" "false")))
 
@@ -73,15 +82,11 @@
   extensions extensions.  See SERVER-297 and SERVER-538 for more info."
   [ruby-load-path :- [schema/Str]
    jruby-home :- schema/Str]
-  (let [load-path (->> ruby-load-path
-                    (cons ruby-code-dir)
-                    (cons (str jruby-home "/lib/ruby/1.9"))
-                    (cons (str jruby-home "/lib/ruby/shared"))
-                    (cons (str jruby-home "/lib/ruby/1.9/site_ruby")))]
-    (->> load-path
-      (map fs/absolute-path)
-      (map fs/normalized-path)
-      (map str))))
+  (->> ruby-load-path
+    (cons ruby-code-dir)
+    (cons (str jruby-home "/lib/ruby/1.9"))
+    (cons (str jruby-home "/lib/ruby/shared"))
+    (cons (str jruby-home "/lib/ruby/1.9/site_ruby"))))
 
 (defn prep-scripting-container
   [scripting-container ruby-load-path gem-home]
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
@@ -80,7 +80,8 @@
                     (cons (str jruby-home "/lib/ruby/1.9/site_ruby")))]
     (->> load-path
       (map fs/absolute-path)
-      (map fs/normalized-path))))
+      (map fs/normalized-path)
+      (map str))))
 
 (defn prep-scripting-container
   [scripting-container ruby-load-path gem-home]
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
@@ -7,7 +7,7 @@
   (:import (com.puppetlabs.puppetserver PuppetProfiler JRubyPuppet)
            (puppetlabs.services.jruby.jruby_puppet_schemas JRubyPuppetInstance PoisonPill)
            (java.util HashMap)
-           (org.jruby CompatVersion RubyInstanceConfig$CompileMode)
+           (org.jruby CompatVersion Main RubyInstanceConfig RubyInstanceConfig$CompileMode)
            (org.jruby.embed ScriptingContainer LocalContextScope)
            (java.util.concurrent LinkedBlockingDeque TimeUnit)
            (clojure.lang IFn)))
@@ -15,20 +15,18 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Definitions
 
-(def jruby-puppet-env
-  "The environment variables that should be passed to the Puppet JRuby interpreters.
-  We don't want them to read any ruby environment variables, like $GEM_HOME or
-  $RUBY_LIB or anything like that, so pass it an empty environment map - except -
-  Puppet needs HOME and PATH for facter resolution, so leave those."
-  (select-keys (System/getenv) ["HOME" "PATH"]))
-
 (def ruby-code-dir
-  "The name of the directory containing the ruby code in this project.
-  This directory lives under src/ruby/"
-  "puppet-server-lib")
+  "The name of the directory containing the ruby code in this project."
+  "src/ruby/puppet-server-lib")
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Private
+
+(schema/defn get-system-env :- {schema/Str schema/Str}
+  "Same as System/getenv, but returns a clojure persistent map instead of a
+  Java unmodifiable map."
+  []
+  (into {} (System/getenv)))
 
 (defn instantiate-free-pool
   "Instantiate a new queue object to use as the pool of free JRubyPuppet's."
@@ -43,25 +41,55 @@
   {:pool (instantiate-free-pool size)
    :size size})
 
+(schema/defn ^:always-validate managed-environment :- {schema/Str schema/Str}
+  "The environment variables that should be passed to the Puppet JRuby interpreters.
+
+  We don't want them to read any ruby environment variables, like $RUBY_LIB or
+  anything like that, so pass it an empty environment map - except - Puppet
+  needs HOME and PATH for facter resolution, so leave those, along with GEM_HOME
+  which is necessary for third party extensions that depend on gems.
+
+  We need to set the JARS..REQUIRE variables in order to instruct JRuby's
+  'jar-dependencies' to not try to load any dependent jars.  This is being
+  done specifically to avoid JRuby trying to load its own version of Bouncy
+  Castle, which may not the same as the one that 'puppetlabs/ssl-utils'
+  uses. JARS_NO_REQUIRE was the legacy way to turn off jar loading but is
+  being phased out in favor of JARS_REQUIRE.  As of JRuby 1.7.20, only
+  JARS_NO_REQUIRE is honored.  Setting both of those here for forward
+  compatibility."
+  [env :- {schema/Str schema/Str}
+   gem-home :- schema/Str]
+  (let [whitelist ["HOME" "PATH"]
+        clean-env (select-keys env whitelist)]
+    (assoc clean-env
+      "GEM_HOME" (fs/absolute-path gem-home)
+      "JARS_NO_REQUIRE" "true"
+      "JARS_REQUIRE" "false")))
+
+(schema/defn ^:always-validate managed-load-path :- [schema/Str]
+  "Return a list of ruby LOAD_PATH directories built from the
+  user-configurable ruby-load-path setting of the jruby-puppet configuration
+  section and prepending specific JRuby directories necessary for third party
+  extensions extensions.  See SERVER-297 and SERVER-538 for more info."
+  [ruby-load-path :- [schema/Str]
+   jruby-home :- schema/Str]
+  (let [load-path (->> ruby-load-path
+                    (cons ruby-code-dir)
+                    (cons (str jruby-home "/lib/ruby/1.9"))
+                    (cons (str jruby-home "/lib/ruby/shared"))
+                    (cons (str jruby-home "/lib/ruby/1.9/site_ruby")))]
+    (->> load-path
+      (map fs/absolute-path)
+      (map fs/normalized-path))))
+
 (defn prep-scripting-container
   [scripting-container ruby-load-path gem-home]
-  (doto scripting-container
-    (.setLoadPaths (cons ruby-code-dir
-                         (map fs/absolute-path ruby-load-path)))
-    (.setCompatVersion (CompatVersion/RUBY1_9))
-    (.setCompileMode RubyInstanceConfig$CompileMode/OFF)
-    ;; We need to set the JARS..REQUIRE variables in order to instruct
-    ;; JRuby's 'jar-dependencies' to not try to load any dependent jars.  This
-    ;; is being done specifically to avoid JRuby trying to load its own version
-    ;; of Bouncy Castle, which may not the same as the one that
-    ;; 'puppetlabs/ssl-utils' uses. JARS_NO_REQUIRE was the legacy way to turn
-    ;; off jar loading but is being phased out in favor of JARS_REQUIRE.  As of
-    ;; JRuby 1.7.20, only JARS_NO_REQUIRE is honored.  Setting both of those
-    ;; here for forward compatibility.
-    (.setEnvironment (merge {"GEM_HOME" gem-home
-                             "JARS_NO_REQUIRE" "true"
-                             "JARS_REQUIRE" "false"}
-                            jruby-puppet-env))))
+  (let [jruby-home (.getHomeDirectory scripting-container)]
+    (doto scripting-container
+      (.setLoadPaths (managed-load-path ruby-load-path jruby-home))
+      (.setCompatVersion (CompatVersion/RUBY1_9))
+      (.setCompileMode RubyInstanceConfig$CompileMode/OFF)
+      (.setEnvironment (managed-environment (get-system-env) gem-home)))))
 
 (defn empty-scripting-container
   "Creates a clean instance of `org.jruby.embed.ScriptingContainer` with no code loaded."
@@ -250,3 +278,20 @@
         (.putFirst pool instance)))
     ;; if we get here, we got a Retry, so we just put it back into the pool.
     (.putFirst (:pool instance) instance)))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Public
+
+(schema/defn ^:always-validate new-main :- jruby-schemas/JRubyMain
+  "Return a new JRuby Main instance which should only be used for CLI purposes,
+  e.g. for the ruby, gem, and irb subcommands.  Internal core services should
+  use `create-scripting-container` instead of `new-main`."
+  [config :- jruby-schemas/JRubyPuppetConfig]
+  (let [jruby-config (RubyInstanceConfig.)
+        jruby-home   (.getJRubyHome jruby-config)
+        {:keys [ruby-load-path gem-home]} config]
+    (doto jruby-config
+      (.setEnvironment (managed-environment (get-system-env) gem-home))
+      (.setLoadPaths (managed-load-path ruby-load-path jruby-home))
+      (.setCompatVersion (CompatVersion/RUBY1_9)))
+    (Main. jruby-config)))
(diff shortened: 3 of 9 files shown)


src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -43,25 +41,55 @@
...
+  user-configurable ruby-load-path setting of the jruby-puppet configuration
+  section and prepending specific JRuby directories necessary for third party
+  extensions extensions.  See SERVER-297 and SERVER-538 for more info."
+  [ruby-load-path :- [schema/Str]
+   jruby-home :- schema/Str]
+  (let [load-path (->> ruby-load-path
+                    (cons ruby-code-dir)
+                    (cons (str jruby-home "/lib/ruby/1.9"))
+                    (cons (str jruby-home "/lib/ruby/shared"))
+                    (cons (str jruby-home "/lib/ruby/1.9/site_ruby")))]
```
jeffmccune: @camlow325 One question I had about this...  The ticket says:

> For Server 1.X only, add in the "front-loading" of the load path with embedded JRuby path libs, as is done for the irb subcommand today - see https://github.com/puppetlabs/puppet-server/blob/puppet-server-1.0.8/src/clj/puppetlabs/puppetserver/cli/irb.clj#L20-L24.

However, in the master branch both the `irb` and `ruby` commands, but not `gem` front-load these three directories relative to `JRUBY_HOME`.  I left them in the consolidated implemenetation, but I was wondering if you could clarify the "for 1.X only" piece.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -15,21 +15,19 @@
...
-  $RUBY_LIB or anything like that, so pass it an empty environment map - except -
-  Puppet needs HOME and PATH for facter resolution, so leave those."
-  (select-keys (System/getenv) ["HOME" "PATH"]))
-
 (def ruby-code-dir
-  "The name of the directory containing the ruby code in this project.
-  This directory lives under src/ruby/"
-  "puppet-server-lib")
+  "The name of the directory containing the ruby code in this project."
+  "src/ruby/puppet-server-lib")
```
jeffmccune: @camlow325 or @cprice404 or whomever feels confident about this...  Two questions I had while working with this bit of code:

 1. The path has been relative in the implementation up until this patch.  I'm not sure that's wise and/or secure for production since this is loading library code.  When my patch uses `ruby-code-dir` now, which is the only place it is used, the value is made absolute and canonical.  The question is: is this reasonable?  Is it a pre-mature improvement?
 2. Given that the path is relative and appears to be only intended to run things from source, does it make sense to include this path in the packaged releases of puppetserver?  I'm trying to come up with how this would be used in a production deployment and am falling short.

nwolfe: This is the path into the ruby code from the root of the `puppet-server-release.jar`.  

@jeffmccune I bet if you built packages with lein ezbake with this change that the CLI tools might not work.

src/clj/puppetlabs/puppetserver/cli/gem.clj
```diff
@@ -1,18 +1,10 @@
...
-  (doto (ScriptingContainer.)
-    (.setArgv (into-array String args))
-    (.setEnvironment
-      (hash-map
-        "GEM_HOME" (get-in config [:jruby-puppet :gem-home])
-        "JARS_NO_REQUIRE" "true"
-        "JARS_REQUIRE" "false"))
-    (.runScriptlet "require 'jar-dependencies'")
-    (.runScriptlet "load 'META-INF/jruby.home/bin/gem'")))
+  (jruby-core/cli-run! config "gem" args))
```
jeffmccune: I suppose I could kill this function entirely and reduce this to:

```clojure
(defn -main [& args] (cli/run jruby-core/cli-run! "gem" args))
```

Sane?

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
jeffmccune: Hrm.  Yeah, I wondered about that...

@nwolfe Putting that (it's broken) issue aside, does it make sense to you to have a relative directory on a LOAD_PATH?

jeffmccune: Well, after some digging I figured out how this works.  It's documented at a high level, but sufficient enough that I feel like I understand it, in the [LoadService](http://jruby.org/apidocs/org/jruby/runtime/load/LoadService.html) JRuby java doc.

Basically, JRuby will search each element of the classpath by taking the name being required, map it onto each element of Ruby's LOAD_PATH, then map it again onto each element of the classpath.  This seems like a ridiculous huge number of permutations, but hey it seems to work!

More concerning to me is this behavior seems to assume there will be relative directories located on the Ruby LOAD_PATH.  That seems terrible and just plain wrong, but it's also how upstream JRuby fundamentally behaves, so we're at least relying on intended behavior upstream, if not sound behavior.

KevinCorcoran: @jeffmccune Just to be clear - this is a relative path **inside the .jar we build**, not a path on the filesystem.  In general, yeah, a relative path on LOAD_PATH seems funky, but in this case it seems correct to me.

jeffmccune: @KevinCorcoran Yeah, understood that it's a relative path inside the jar, but the behavior is that it's added to the $LOAD_PATH (which is funky) and JRuby looks in JAR files by pre-pending each element of the LOAD_PATH to the value passed to "require" then peeking inside the JAR root.

So, even though it's used for a JAR file, it's added to the LOAD_PATH which affects both the behavior of filesystem and JAR resource loading.

KevinCorcoran: ohhh, my.  Gross.  Well, thanks for the explanation.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -250,3 +279,20 @@
         (.putFirst pool instance)))
     ;; if we get here, we got a Retry, so we just put it back into the pool.
     (.putFirst (:pool instance) instance)))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
```
camlow325: I think there are some other functions in this namespace which are called from outside of this namespace, e.g., `create-pool-from-config` appears to be called from `jruby-puppet-agents`.  I like the idea of adding the "public" section here, but I think it would be good to move any of the other functions which effectively are public down into this new section.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -43,25 +41,56 @@
...
-    ;; 'puppetlabs/ssl-utils' uses. JARS_NO_REQUIRE was the legacy way to turn
-    ;; off jar loading but is being phased out in favor of JARS_REQUIRE.  As of
-    ;; JRuby 1.7.20, only JARS_NO_REQUIRE is honored.  Setting both of those
-    ;; here for forward compatibility.
-    (.setEnvironment (merge {"GEM_HOME" gem-home
-                             "JARS_NO_REQUIRE" "true"
-                             "JARS_REQUIRE" "false"}
-                            jruby-puppet-env))))
+  (let [jruby-home (.getHomeDirectory scripting-container)]
+    (doto scripting-container
```
camlow325: The `scripting-container` initialization here and the `jruby-config` initialization in `new-main` below look very similar - other than just a difference in the underlying types being used.  What would you think about consolidating these into a single function which would operate either on the `ScriptingContainer` or `RubyInstanceConfig` passed into it?  If the idea of taking advantage of the duck-typing here seems too distasteful, could we at least factor out a common helper function for `compat-version` that both this function and `new-main` could call into to get the same `CompatVersion`?

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -43,25 +41,56 @@
...
+      (.setLoadPaths (managed-load-path ruby-load-path jruby-home))
+      (.setCompatVersion (CompatVersion/RUBY1_9))
+      (.setCompileMode RubyInstanceConfig$CompileMode/OFF)
```
camlow325: Any reason we would do this via the `ScriptingContainer` but not via a `RubyInstanceConfig` (through `new-main`).  Seems like we should be doing the same thing in both places - again through a common helper function that both paths would use so that we would just have to change the default value in one place.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
cprice404: Wow, that's a very unfortunate API design on their part :(

If they just had an interface that was implemented by both `ScriptingContainer` and `RubyInstanceConfig`... :(

The duck typing idea bugs me, quite a bit, but it does seem like our tests would catch it if the interface changed on either of the two objects in a future JRuby release.

I'm on the fence as to which I think is better, but I do agree with you that it would be a good idea to wrap the `compat-version` and `compile-mode` values in a function or a var and reference that from both of these initializers.

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -172,3 +181,12 @@
...
 
+(def JRubyMain
+  (schema/pred jruby-main-instance?))
+
+(def JRubyMainStatus
+  (schema/pred jruby-main-status-instance?))
+
+(def JRubyMainStatusOrNil
+  (schema/either JRubyMainStatus
+    (schema/pred nil? 'nil?)))
```
cprice404: this could just be `schema/maybe JRubyMainStatus`, right?

test/unit/puppetlabs/services/jruby/jruby_puppet_core_test.clj
```diff
@@ -13,6 +14,37 @@
...
+         captured-output# (new ByteArrayOutputStream)
+         capturing-print-stream# (new PrintStream captured-output#)]
+     (try
+       (System/setOut capturing-print-stream#)
+       (swap! return-map# assoc :return (do ~@body))
+       (finally
+         (.flush capturing-print-stream#)
+         (swap! return-map# assoc :out (.toString captured-output#))
+         (System/setOut system-output#)))
+     @return-map#))
```
cprice404: These macros are doing some pretty hard-core stuff (System/setOut, etc.) that makes me nervous.  That said, I looked at the tests and they do seem valuable, so I think I'm good with them... but... yikes.

test/unit/puppetlabs/services/jruby/jruby_puppet_core_test.clj
```diff
@@ -13,6 +14,37 @@
...
+  The return value of body is available as :return in the return map.
+
+  This macro is not thread safe."
+  [& body]
+  `(let [return-map# (atom {})
+         system-output# (System/out)
+         captured-output# (new ByteArrayOutputStream)
+         capturing-print-stream# (new PrintStream captured-output#)]
+     (try
+       (System/setOut capturing-print-stream#)
```
cprice404: I think that in Clojure you might normally do this with something like:

`(binding [*out* capturing-print-stream#]`

But, not sure if that would work for JRuby, so if we're messing w/stdin and stdout at all during the tests, it probably doesn't matter much which way we do it.

I wonder if it would be worth putting these tests into the `integration` test suite since they're a little riskier (because they much with stdout/stderr)?  Not sure about that one either.

test/unit/puppetlabs/services/jruby/jruby_puppet_core_test.clj
```diff
@@ -13,6 +14,37 @@
...
+(defmacro capture-out
+  "capture System.out and return it as the value of :out in the return map.
+  The return value of body is available as :return in the return map.
+
+  This macro is not thread safe."
+  [& body]
+  `(let [return-map# (atom {})
+         system-output# (System/out)
+         captured-output# (new ByteArrayOutputStream)
+         capturing-print-stream# (new PrintStream captured-output#)]
```
cprice404: nitpick: I think the `(ClassName.)` constructor form is more prevelant in Clojure than the `(new ClassName)` form.

cprice404:
I think I'm +1, except for the questions about possibly DRY'ing up the ScriptingContainer/RubyInstanceConfig stuff a bit.

test/unit/puppetlabs/services/jruby/jruby_puppet_core_test.clj
jeffmccune: Yeah, they make me nervous too.  The original implementation was using functions but was really awkward.  I'd like to use a binding in clojure, but my assumption is that won't help with `System.out` and `System.in`.

My main concerns are around thread safety, but I can't imagine anything else mucking with System.out unless we do something like [parallel-test](https://github.com/aredington/parallel-test) which requires tests to be marked as `^:parallel` before they'll run concurrently.

My thinking was this isn't really any worse than how we're doing other things like TCP 8140, but it's also far from ideal and has lots of room for improvement.

test/unit/puppetlabs/services/jruby/jruby_puppet_core_test.clj
jeffmccune: I've got them marked as integration tests already and was wondering if that alone was enough to warrant they move into the integration sub-directory.

jeffmccune:
> I think I'm +1, except for the questions about possibly DRY'ing up the ScriptingContainer/RubyInstanceConfig stuff a bit.

Cool, at this point the only thing that remains to change beyond addressing comments is to add 1 beaker test for each of the subcommands, which I hope to do today.

I also have a TODO item to build packages off this topic branch and validate we don't run into the hipchat gem problem described in [SERVER-538](https://tickets.puppetlabs.com/browse/SERVER-538).  If we do run into the issue my plan is to just pull out d7574cc.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
jeffmccune: Addressed in be6ed72

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
jeffmccune: fixed in  ef931c9 (maint) Add compile-mode, compat-version defs

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
jeffmccune: Yes, basically I didn't want to change the way we were initializing a ScriptingContainer which has been to not inject a RubyInstanceConfig into the initializer and instead configure it after initialization.

However, I needed to do so for the use of Main since that was the only way I could figure out to get the jruby home.

Upstream JRuby [says](https://github.com/jruby/jruby/blob/1.7.20/core/src/main/java/org/jruby/Main.java#L178-L184) that Main should only be used for command-line entry points for JRuby, and to use other mechanisms when embedding JRuby into an application.

That's why I kept both Main and ScriptingContainer around.

jeffmccune: I think the best way to solve this if we really want to is to figure out if we could have a single function that produces RubyInstanceConfig instances, then see if we can inject that instance into a ScriptingContainer like we're already doing it for Main.

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
jeffmccune: `maybe`...  I'm sure you're right, I just started writing schemas for functions so I'll check and fix it up if so.

jeffmccune:
Acceptance test added, this should be ready now.  No longer "for review" but still needs the normal review.

Chris, this is largely the same as when you looked at it earlier.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -15,53 +15,82 @@
...
+  See also:  http://jruby.org/apidocs/org/jruby/runtime/load/LoadService.html"
   "puppet-server-lib")
 
+(def compile-mode
+  "The JRuby compile mode to use."
+  RubyInstanceConfig$CompileMode/OFF)
+
+(def compat-version
+  "The JRuby compatibility version to use"
+  (CompatVersion/RUBY1_9))
```
nwolfe: Are these extra parens necessary? Just curious, since we don't have them for `compile-mode` above. I can't remember if there's a subtle difference

test/unit/puppetlabs/services/jruby/jruby_puppet_core_test.clj
nwolfe: > move into the integration sub-directory.

FWIW I don't see the value of making the distinction between unit and integration tests at the directory level.
:+1: for keeping the test here

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -71,7 +100,7 @@
          (string? gem-home)]
    :post [(instance? ScriptingContainer %)]}
   (-> (ScriptingContainer. LocalContextScope/SINGLETHREAD)
-      (prep-scripting-container ruby-load-path gem-home)))
+    (prep-scripting-container ruby-load-path gem-home)))
```
nwolfe: :cry:

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -15,53 +15,82 @@
...
+  RubyInstanceConfig$CompileMode/OFF)
+
+(def compat-version
+  "The JRuby compatibility version to use"
+  (CompatVersion/RUBY1_9))
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Private
 
+(schema/defn get-system-env :- {schema/Str schema/Str}
```
nwolfe: Since we have this `{schema/Str schema/Str}` in three places now I _might_ consider a schema for it. Just maybe though

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -15,53 +15,82 @@
...
+
+  See also:  http://jruby.org/apidocs/org/jruby/runtime/load/LoadService.html"
   "puppet-server-lib")
 
+(def compile-mode
+  "The JRuby compile mode to use."
+  RubyInstanceConfig$CompileMode/OFF)
+
+(def compat-version
+  "The JRuby compatibility version to use"
```
nwolfe: Your docstrings are kicking butt so far.  I think we should mention in these two that they're used for both the server and CLI.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -110,6 +139,21 @@
         (.put puppet-config dir (fs/absolute-path value))))
     puppet-config))
 
+(schema/defn borrow-with-timeout-fn :- jruby-schemas/JRubyPuppetBorrowResult
```
nwolfe: You didn't do this, but I'm wondering why this is named as `...-fn`.  Does it return a function?

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -110,6 +139,21 @@
...
+  [timeout :- schema/Int
+   pool :- jruby-schemas/pool-queue-type]
+  (.pollFirst pool timeout TimeUnit/MILLISECONDS))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Public
+
+(schema/defn ^:always-validate
+  create-pool-from-config :- jruby-schemas/PoolState
+  "Create a new PoolData based on the config input."
```
nwolfe: You didn't do this, but looks like we forgot to update the docstring when we renamed `PoolData`.  I might update it, but no big deal

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -250,3 +289,18 @@
         (.putFirst pool instance)))
     ;; if we get here, we got a Retry, so we just put it back into the pool.
     (.putFirst (:pool instance) instance)))
+
+(schema/defn ^:always-validate new-main :- jruby-schemas/JRubyMain
+  "Return a new JRuby Main instance which should only be used for CLI purposes,
+  e.g. for the ruby, gem, and irb subcommands.  Internal core services should
+  use `create-scripting-container` instead of `new-jruby-main`."
```
nwolfe: Looks like this is trying to reference itself, which is named `new-main`.

test/unit/puppetlabs/services/jruby/jruby_puppet_core_test.clj
nwolfe: Might be worthwhile to mention in the docstring that redirecting streams like this is necessary for JRuby, considering it's not quite standard Clojure (like Chris was mentioning).

acceptance/suites/tests/010-puppetserver-cli/subcommand/SERVER-297_common_behavior.rb
```diff
@@ -0,0 +1,33 @@
...
+
+step "irb: Check that PATH, HOME, GEM_HOME JARS_REQUIRE and JARS_NO_REQUIRE are present"
+cmd = "#{cli} irb -f -rjson -e 'puts JSON.pretty_generate(ENV.to_hash)'"
+on(master, cmd) do
+  assert_match(/\bPATH\b/, output, "PATH missing")
+  assert_match(/\bHOME\b/, output, "HOME missing")
+  assert_match(/\bGEM_HOME\b/, output, "GEM_HOME missing")
+  assert_match(/\bJARS_REQUIRE\b/, output, "JARS_REQUIRE missing")
+  assert_match(/\bJARS_NO_REQUIRE\b/, output, "JARS_NO_REQUIRE missing")
+end
```
nwolfe: :v: nice looking test

src/clj/puppetlabs/puppetserver/cli/gem.clj
nwolfe: I wouldn't be opposed to squashing these like you proposed.  Then at least all these minimal CLI namespaces will look the same, where they just call `cli/run` with a `jruby-core/cli-{run!,ruby!}` function.  No big deal though

test/unit/puppetlabs/services/jruby/jruby_puppet_core_test.clj
jeffmccune: Fixed in 72c6d4bd9e0d9a8c81d8b0b37139304a646bc805, thanks @nwolfe

test/unit/puppetlabs/services/jruby/jruby_puppet_core_test.clj
jeffmccune: Noted, thanks!  I'll use the conventional form from here on out.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
jeffmccune: Yes, CompatVersion/RUBY1_9 is a method, so we need to call it to get the value.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
jeffmccune: Yeah, sorry.  This snuck in with the public / private re-shuffle

test/unit/puppetlabs/services/jruby/jruby_puppet_core_test.clj
nwolfe: FWIW I like `(new CLASS...)` better as it looks like normal functional Clojure, whereas `(CLASS. ...)` looks like Java interop....which I suppose you could argue is what this code is anyway :smile:

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
KevinCorcoran: Hmm, actually, it's not a method, it's an enum value ...
```java
public enum CompatVersion {

    RUBY1_8, RUBY1_9, RUBY2_0, BOTH;
```
... which means that the parens can be removed, I think.  Should be same as accessing a field.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
nwolfe: I just meant that we think this formatting is a good idea

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
nwolfe: Don't worry about spending any time tracking this down and potentially changing it unless it's worthwhile.  I'm more just curious

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
jeffmccune: Fixed up in 9965399 `(maint) Add compile-mode, compat-version defs`  Thanks, great suggestion.

acceptance/suites/tests/010-puppetserver-cli/subcommand/SERVER-297_common_behavior.rb
```diff
@@ -0,0 +1,33 @@
...
+require 'json'
+
+test_name "Puppetserver subcommand consolidated ENV handling tests."
+
+# --------------------------------------------------------------------------
+# This behavior needs to be baked into Beaker but is not yet. SERVER-79 tracks
+# the need to fix this once Beaker has been modified to make the paths to these
+# commands available.
+#
+cli = "puppetserver"
```
nwolfe: Do we need this comment and `cli` variable anymore?  Let's just inline it below and delete this cruft if so

acceptance/suites/tests/010-puppetserver-cli/subcommand/SERVER-297_common_behavior.rb
```diff
@@ -0,0 +1,33 @@
+require 'test/unit/assertions'
+require 'json'
```
nwolfe: Also doesn't look like this is used?

nwolfe:
:+1: other than minor nitpick comments

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
jeffmccune: I'm convinced.  Fixed in 79711eb `(maint) Add jruby-schemas/EnvMap`.

acceptance/suites/tests/010-puppetserver-cli/subcommand/SERVER-297_common_behavior.rb
nwolfe: I'm stupid, it's right there

test/unit/puppetlabs/services/jruby/jruby_puppet_core_test.clj
jeffmccune: I wasn't going to argue the point, but I'm with @nwolfe, I found `(new Foo)` easier to read and understand as a newbie than `(Foo.)` which is less obvious about what's happening.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
jeffmccune: Oh, heh.  At least we don't have hash rockets to argue about, but I guess let bindings are effectively the same thing.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
jeffmccune: IDK, but looking at the schema it doesn't look to be a function:

```clojure
(def JRubyPuppetBorrowResult
  (schema/pred (some-fn nil?
                        poison-pill?
                        retry-poison-pill?
                        jruby-puppet-instance?)))
```

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
jeffmccune: Fixed up in 321237a `(SERVER-297) Consolidate JRuby environment handling`

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
jeffmccune: Fixed up in 54b7bb7 `(maint) Mark j_p_internal functions public / private`

acceptance/suites/tests/010-puppetserver-cli/subcommand/SERVER-297_common_behavior.rb
jeffmccune: It's used for `JSON.parse`

acceptance/suites/tests/010-puppetserver-cli/subcommand/SERVER-297_common_behavior.rb
jeffmccune: Fixed up in b14b711 `(SERVER-297) Add acceptance tests for env behavior`

acceptance/suites/tests/010-puppetserver-cli/subcommand/SERVER-297_common_behavior.rb
jeffmccune: Thanks!

jeffmccune:
@nwolfe I've addressed all of your comments.  Thanks for the eye to detail.

PR #584, opened 2015-05-29
(docs) Adds Debian specific config location

petems:
As someone not super familir to Debian file defaults, had to do a bit of searching to find out how to set puppetserver memory on Ubuntu! :smile:

I did it as a bulletpoint list for if this is at a different location on other OS's and can be easily added in the future


KevinCorcoran:
Thanks for the contribution!

PR #585, opened 2015-05-29
Add $codedir to the list of settings read from Puppet

KevinCorcoran:
Not actually needed in Puppet Server, but will be needed for file sync.


fpringvaldsen:
:+1:

PR #586, opened 2015-06-01
(SERVER-701) Cherry-pick jar exclusion acceptance test

dankreek:
JARS_NO_REQUIRE is an environment variable which, if set to `true`, will prevent
the loading of embedded jars using the `require_jar' method of the
`jar-dependencies` gem.

This is required to ensure that jruby-openssel, which is loaded by the JRuby
Kernel whenever ruby code requires 'openssl', does not attempt to load embedded
bouncy castle jars onto the classpath. The effect is to ensure that
jruby-openssl uses the bouncycastle included as a requirement of
jvm-certificate-authority rather than the version included with JRuby, as well
as preventing jvm-ceritificate-authority from using JRuby's bouncycastle.

This commit also adds an integration smoke test to ensure that the jar is
actually excluded by listing the contents of the jar then using the
`assert_no_match` minitest method to validate the listed output does not contain
BC jars.

Conflicts:
	src/clj/puppetlabs/puppetserver/cli/gem.clj
	src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -223,6 +223,30 @@ EOF
         raise "Package #{name} cannot be upgraded on #{host}"
     end
   end
+
+  def get_defaults_var(host, varname)
+    if master.is_pe?
+      package_name = "pe-puppetserver"
+    else
+      package_name = "puppetserver"
+    end
+
+    variant, version, _, _ = master['platform'].to_array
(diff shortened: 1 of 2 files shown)


camlow325:
:+1: :copyright: :spaghetti:

PR #593, opened 2015-06-02
(SERVER-692) Add defaults to master dir settings

fpringvaldsen:
Add hardcoded defaults for the :master-_-dir settings so that
the behavior of puppetserver will be in line with our
docs when these settings are absent.


KevinCorcoran:
:+1:

camlow325:
:+1:

PR #598, opened 2015-06-05
(maint) Fix failing FOO_DEBUG irb and ruby CLI tests

jeffmccune:
The smoke tests are failing checking that FOO_DEBUG is being preserved for the
irb and ruby subcommands.  This is a problem because the behavior has been
consolidated around the behavior of the production scripting container which is
to filter out variables like FOO_DEBUG.  This patch addresses the problem by
changing the test to assert the desired behavior.


camlow325:
:+1:

nwolfe:
:+1:

PR #600, opened 2015-06-05
(SERVER-717) Fix CI failures

fpringvaldsen:
* Remove the step from the pre-suites that installs the puppetlabs
release repositories, as this is now done in beaker's
install_puppet function and was causing failures in the
pre-suites for both the smoke tests and the 3.x compatibility
tests.
* Remove the puppet 3 compatibility suite script from the
dev folder
* Rename the testrun-full script, as by default it only runs the
smoke tests


camlow325:
:see_no_evil: @nwolfe or @jeffmccune - look ok to you?

jeffmccune:
:+1:

nwolfe:
:+1:

PR #602, opened 2015-06-06
(SERVER-718) Load facter jar into system classloader classpath

camlow325:
This commit adds some logic which searches for a file named 'facter.jar'
in the ruby load path and, if found, adds the file to the system
classloader's classpath.  This is needed for compatibility with Facter
3+.  The jar needs to be in the system classloader's classpath in order
for Facter's backing native shared library to only be loaded once by a
classloader, avoiding an error which would otherwise occur if attempts
were made to load the library into more than 1 JRuby container instance
classloader.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
@@ -1,5 +1,8 @@
 (ns puppetlabs.services.jruby.jruby-puppet-core
-  (:require [schema.core :as schema]
+  (:require [clojure.tools.logging :as log]
+            [me.raynes.fs :as fs]
+            [schema.core :as schema]
+            [puppetlabs.kitchensink.classpath :as ks-classpath]
             [puppetlabs.kitchensink.core :as ks]
             [puppetlabs.services.jruby.jruby-puppet-schemas :as jruby-schemas]
             [puppetlabs.services.jruby.puppet-environments :as puppet-env]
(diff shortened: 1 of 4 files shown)


src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -105,6 +108,25 @@
...
+  facter jar file.  The first one found is added to the system classloader's
+  classpath.  If no match is found, an info message is written to the log
+  but no failure is returned"
+  [ruby-load-path :- (schema/both (schema/pred vector?) [schema/Str]) ]
+  (if-let [facter-jar (first
+                        (filter fs/exists?
+                          (map #(fs/file % facter-jar) ruby-load-path)))]
+    (do
+      (log/debugf "Adding facter jar to classpath from: %s" facter-jar)
+      (ks-classpath/add-classpath facter-jar))
```
camlow325: One thing which bothers me about this is that reflection has to be used under the hood.  This is needed because `add-classpath` calls through to `URLClassLoader.addURL`, which is a protected method.  See https://github.com/tobias/dynapath/blob/v0.2.3/src/dynapath/defaults.clj#L17-L21.  Not sure there's any way to avoid this, though, without having the path to `facter.jar` be appended to the `java -cp` argument on the command line, which we specifically did not want to do for this work.

jeffmccune:
:+1:

dankreek:
I know this may be opening a can of difficult-to-maintain worms here and is probably more food for thought than anything, but it seems like this code will load up any ole' jar named "facter.jar" into the classpath. Is this a security concern? Should we maybe open up a ticket to later check the hash of the file against a list of blessed hashes or something?

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
dankreek: I've seen this same technique used on the Java-end for things like dynamically loading up a JDBC jar from the filesystem. I honestly have more beef with the Java API for making us do this dumb stuff than actually doing the dumb stuff ourselves. So, this looks good to me.

camlow325:
> it seems like this code will load up any ole' jar named "facter.jar" into the classpath

Not sure I see that as a security concern.  Puppet would already load up any Ruby code in the `ruby-load-path` without any specific validation of the content - e.g., whether it was "signed" in some manner as "trusted" code.  Also, maintaining that list of hashes could be difficult over time - e.g., would we need to roll Puppet Server every time the jar needed to be rebuilt with different content for puppet-agent?

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
camlow325: Yeah, understood.  Just seems kind of nasty, but I agree that there's already a lot of prior art out there for following this technique.

dankreek:
> Not sure I see that as a security concern. Puppet would already load up any Ruby code in the ruby-load-path without any specific validation of the content - e.g., whether it was "signed" in some manner as "trusted" code. Also, maintaining that list of hashes could be difficult over time - e.g., would we need to roll Puppet Server every time the jar needed to be rebuilt with different content for puppet-agent?

Cool, I just had the thought and wanted to raise it. This definitely falls under "if they have root on your system, you're already hosed" so I'm good with it.

dankreek:
:+1: LGTM!

PR #603, opened 2015-06-08
(SERVER-646) Allow charset for certificate_status content-type

camlow325:
This commit changes the validation for the 'certificate_status' endpoint
to allow for the request's content-type to include a "charset"
parameter.  Previously, the presence of a "charset" parameter in the
content-type would cause the content-type to fail validation with an
HTTP 415 (unsupported media type) error.

### DIFF
--- a/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
+++ b/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
@@ -14,6 +14,7 @@
             [liberator.core :refer [defresource]]
             ;[liberator.dev :as liberator-dev]
             [liberator.representation :as representation]
+            [ring.util.request :as request]
             [ring.util.response :as rr]))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
@@ -90,7 +91,7 @@
 
 (defn content-type-valid?
   [context]
-  (let [content-type (get-in context [:request :headers "content-type"])]
+  (let [content-type (request/content-type (:request context))]
     (or
       (nil? content-type)
       (media-types content-type))))
--- a/test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
+++ b/test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
@@ -455,6 +455,21 @@
         (is (true? (fs/exists? signed-cert-path)))
         (is (= 204 (:status response))))))
 
+  (testing "a signing request w/ a 'application/json' content-type and charset succeeds"
+    (let [settings (testutils/ca-sandbox! cadir)
+          test-app (-> (build-ring-handler settings "42.42.42")
+                     (wrap-with-ssl-client-cert))
+          signed-cert-path (ca/path-to-cert (:signeddir settings) "test-agent")]
+      (is (false? (fs/exists? signed-cert-path)))
+      (let [response (test-app
+                      {:uri "/production/certificate_status/test-agent"
+                       :request-method :put
+                       :headers {"content-type"
+                                 "application/json; charset=UTF-8"}
+                       :body (body-stream "{\"desired_state\":\"signed\"}")})]
+        (is (true? (fs/exists? signed-cert-path)))
+        (is (= 204 (:status response))))))
+
   (testing "a signing request w/ a 'text/pson' content-type succeeds"
     (let [settings         (testutils/ca-sandbox! cadir)
           test-app         (-> (build-ring-handler settings "42.42.42")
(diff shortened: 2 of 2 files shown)


KevinCorcoran:
I am so glad you fixed this!  It's been bugging me.

jonathannewman:
Thanks for putting this together!

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -14,6 +14,7 @@
             [liberator.core :refer [defresource]]
             ;[liberator.dev :as liberator-dev]
```
dankreek: I'm guessing we want this line removed?

rlinehan: @dankreek it seems like we've been leaving this in along with https://github.com/puppetlabs/puppet-server/pull/603/files#diff-bb1d8f2d842888913c49c1dddc70a5eaR263 because it's really helpful for debugging? It's at least been here a while, I think.

camlow325: Yeah, I actually uncommented that to debug this one initially, too, so guess it's still serving its purpose?  :smile:

nwolfe: Haha yeah @KevinCorcoran threw the comment in some liberator resource somewhere after a particularly frustrating week of development, if I remember correctly.

I'd say just start removing these comments from the codebase.  If there's really something to be shared to other developers then we should write formal documentation somewhere, even if it's only a paragraph or two.

camlow325: I can take them out.  @KevinCorcoran - what do you think?

KevinCorcoran: I left them in there because I thought they would be useful to people in the future.  They can be extremely useful in debugging issues with liberator resources, and I figured most people didn't know about the time (at the time, anyway).

If y'all don't think they're useful now, I am fine if they get removed.

camlow325: I guess since this PR didn't touch these and we're on the fence about keeping them in, I'd propose not to touch them for this PR.  I'm okay with @nwolfe filing a principle violation follow-up PR on this, if he'd like.

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
```diff
@@ -483,6 +513,20 @@
         (is (true? (fs/exists? signed-cert-path)))
         (is (= 204 (:status response))))))
```
rlinehan: This is maybe outside the scope of this PR, but it seems kind of bizarre that this test is still using `/production/certificate_status` while the other tests in this file are using `/v1/certificate_status`.

camlow325: Good call - I'll fix that up.

rlinehan: Looks like it's actually only the tests added in this PR that use `/production`, and also since this is against stable probably the new tests probably should be using `/production`, because otherwise the code just thinks that the environment is `/v1`... which doesn't actually matter since it throws out the environment, but is just confusing and kind of wrong.

KevinCorcoran: This is my favorite thing about the CA API.  :dancers:

camlow325: Yeah, duh, my bad.  I'd done this work against the `master` branch initially - which is why I had the new-style URLs in the tests - before remembering I probably should have done this against the stable branch instead.  But when I reapplied the changes to stable, I didn't change the URLs back to look like what they would for stable.  In any event, I'll replace the 'v1' with 'production' because I agree that looks better / more consistent with the other tests.

dankreek: Yeah, that whole "optional parameters first" thing is very amusing.

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
```diff
@@ -455,6 +455,22 @@
...
 
+  (testing "a signing request w/ a 'application/json' content-type and charset succeeds"
+    (let [settings         (testutils/ca-sandbox! cadir)
```
KevinCorcoran: speaking of principles, I believe that all of the alignment in these let bindings and maps is against the Linehan Principle.

KevinCorcoran:
:bike:shedding aside, this LGTM:  :+1:

camlow325:
Fixed up the spacing and replaced the "v1" values with "production" in b17a242.  I just squashed the prior commit since the differences seemed minor.

camlow325:
Does anyone see anything I missed here or does this look ok to merge?

dankreek:
:+1: LGTM

PR #604, opened 2015-06-08
(SERVER-721) Consolidate JRuby environment handling

jeffmccune:
Without this patch the manner in which environment variables are handled
in the JRuby portions of puppet-server vary across the use cases of the
`puppetserver {gem,irb,ruby}` command line utilities in addition to the
service itself.  This is a problem because the behaviors have not kept
in sync with one another over time so there has been divergence at
times, e.g. `puppetserver gem env` currently fails because PATH is unset
in that scenario but set in others, e.g. the production service.

This patch addresses the problem by consolidating all of the environment
handling behavior into a single authoritative method.  The behavior
itself remains unchanged, GEM_HOME is set based on the config file
value, PATH and HOME are the only variables allowed to flow through to
JRuby unmodified, all other variables are filtered out, and JARS_REQUIRE
and JARS_NO_REQUIRE are set appropriately.

In addition to the environment handling, this patch also consolidates
the JRuby load path handling in a similar fashion across the scenarios
of command line utilities and the production service.

(maint) Fix managed-load-path schema errors

Fix this error:

    Caused by: clojure.lang.ExceptionInfo: Output of managed-load-path does not match schema: [(not (instance? java.lang.String a-java.io.File)) (not (instance? java.lang.String a-java.io.File)) (not (instance? java.lang.String a-java.io.File)) (not (instance? java.lang.String a-java.io.File)) (not (instance? java.lang.String a-java.io.File)) (not (instance? java.lang.String a-java.io.File))]
     at puppetlabs.services.jruby.jruby_puppet_internal$eval9067$managed_load_path__9068.invoke (jruby_puppet_internal.clj:69)
        puppetlabs.services.jruby.jruby_puppet_internal$prep_scripting_container.invoke (jruby_puppet_internal.clj:89)
        puppetlabs.services.jruby.jruby_puppet_internal$empty_scripting_container.invoke (jruby_puppet_internal.clj:101)
        puppetlabs.services.jruby.jruby_puppet_internal$create_scripting_container.invoke (jruby_puppet_internal.clj:117)
        puppetlabs.services.jruby.jruby_puppet_internal$eval9122$create_pool_instance_BANG___9123$fn__9124.invoke (jruby_puppet_internal.clj:156)
        puppetlabs.services.jruby.jruby_puppet_internal$eval9122$create_pool_instance_BANG___9123.invoke (jruby_puppet_internal.clj:141)

(maint) Remove unused definitions from jruby-puppet-core

These defs moved over to jruby-puppet-internal in
23154f42f24351e55a329cf52663d291c4c5312f and are no longer used from
jruby-puppet-core.

(SERVER-297) Fix up LOAD_PATH behavior

Without this patch the load path behavior is incorrect because the use
of normalized and absolute paths is incompatible with the use of loading
resources from a sealed jar.  This patch addresses the problem by
restoring the previous behavior of implicitly relying in JRuby's
LoadService implementation.

For more information, particularly on the combination and interaction of
load path, class path, and ruby require, please see:

https://github.com/jruby/jruby/blob/1.7.20/core/src/main/java/org/jruby/runtime/load/LoadService.java

(SERVER-297) Add jruby subcommand integration tests

Without this patch there are not tests covering the behavior of the irb,
ruby, and gem JRuby subcommands.  This patch addresses the problem by
adding integration tests that assert against standard output and the
exit status.

NB this patch differs from the patch against master in that the
ruby-load-path setting is under the os-settings configuration section in
this patch while it is under the jruby-puppet section in the patch
against master for puppetserver 2.0

(SERVER-297) Check if subcommand exists before loading

Without this patch the subcommand string given to the cli-run! command
may not actually exist in the assumed directory of
META-INF/jruby.home/bin/  This patch checks before trying to execute the
given subcommand.

(maint) Mark j_p_internal functions public / private

Addressing PR comments, sorted by using "find usages" in cursive.

(maint) Add compile-mode, compat-version defs

Without this patch the initialization logic for the JRuby instances in
new-main and prep-scripting-container are almost identical.  This is a
problem because they should be identical and the compatibility version
and compile mode are not referenced in a way that changing one will also
change the other.

This patch addresses the problem by making the behavior identical
without going so far as to providing a single function that operates on
both classes of objects.

(maint) use schema/maybe instead of schema/pred nil?

(SERVER-297) Add acceptance tests for env behavior

Without this patch there aren't any acceptance tests that validate the
behavior of environment handling.  This patch addresses the problem by
adding a test for the irb and ruby commands.  The gem command is already
tested via the `gem env` command and is difficult to inspect the
environment from since it doesn't evaluate arbitrary code like irb and
ruby can.

(maint) Fix failing FOO_DEBUG irb and ruby CLI tests

The smoke tests are failing checking that FOO_DEBUG is being preserved
for the irb and ruby subcommands.  This is a problem because the
behavior has been consolidated around the behavior of the production
scripting container which is to filter out variables like FOO_DEBUG.
This patch addresses the problem by changing the test to assert the
desired behavior.

(maint) Fix failing smoke test on undefined var output

Fixes:

    #<NameError: undefined local variable or method `output'

(maint) Fix failing smoke test on irb -e

Fixes this error:

    IRB::UnrecognizedSwitch: Unrecognized switch: -e

irb reads from STDIN, it doesn't have a -e flag like the ruby command
does.

(maint) Add jruby-schemas/EnvMap

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
@@ -4,7 +4,8 @@
             [puppetlabs.services.jruby.jruby-puppet-schemas :as jruby-schemas]
             [puppetlabs.services.jruby.puppet-environments :as puppet-env]
             [puppetlabs.services.jruby.jruby-puppet-internal :as jruby-internal]
-            [puppetlabs.services.jruby.jruby-puppet-agents :as jruby-agents])
+            [puppetlabs.services.jruby.jruby-puppet-agents :as jruby-agents]
+            [clojure.java.io :as io])
   (:import (puppetlabs.services.jruby.jruby_puppet_schemas JRubyPuppetInstance)
            (com.puppetlabs.puppetserver PuppetProfiler)))
 
@@ -25,21 +26,6 @@
   "The default number of milliseconds that the client will allow for no data to
   be available on the socket. Currently set to 20 minutes."
   (* 20 60 1000))
-
-;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
-;;; Definitions
-
-(def jruby-puppet-env
-  "The environment variables that should be passed to the Puppet JRuby interpreters.
-  We don't want them to read any ruby environment variables, like $GEM_HOME or
-  $RUBY_LIB or anything like that, so pass it an empty environment map - except -
-  Puppet needs HOME and PATH for facter resolution, so leave those."
-  (select-keys (System/getenv) ["HOME" "PATH"]))
-
-(def ruby-code-dir
-  "The name of the directory containing the ruby code in this project.
-  This directory lives under src/ruby/"
-  "puppet-server-lib")
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Private
@@ -173,3 +159,20 @@
   "Return a borrowed pool instance to its free pool."
   [instance :- jruby-schemas/JRubyPuppetInstanceOrRetry]
   (jruby-internal/return-to-pool instance))
+
+(schema/defn ^:always-validate cli-ruby! :- jruby-schemas/JRubyMainStatus
+  "Run JRuby as though native `ruby` were invoked with args on the CLI"
+  [config :- {schema/Keyword schema/Any}
+   args :- [schema/Str]]
+  (let [main (jruby-internal/new-main (initialize-config config))
+        argv (into-array String (concat ["-rjar-dependencies"] args))]
+    (.run main argv)))
+
+(schema/defn ^:always-validate cli-run! :- (schema/maybe jruby-schemas/JRubyMainStatus)
+  "Run a JRuby CLI command, e.g. gem, irb, etc..."
+  [config :- {schema/Keyword schema/Any}
+   command :- schema/Str
+   args :- [schema/Str]]
+  (let [load-path (format "META-INF/jruby.home/bin/%s" command)]
+    (if-let [url (io/resource load-path (.getClassLoader org.jruby.Main))]
+      (cli-ruby! config (concat ["-e" (format "load '%s'" url) "--"] args)))))
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
@@ -5,7 +5,8 @@
             [puppetlabs.services.jruby.puppet-environments :as puppet-env]
             [puppetlabs.services.jruby.jruby-puppet-internal :as jruby-internal]
             [puppetlabs.services.jruby.jruby-puppet-agents :as jruby-agents]
-            [clojure.java.io :as io])
+            [clojure.java.io :as io]
+            [clojure.tools.logging :as log])
   (:import (puppetlabs.services.jruby.jruby_puppet_schemas JRubyPuppetInstance)
            (com.puppetlabs.puppetserver PuppetProfiler)))
 
@@ -173,6 +174,10 @@
   [config :- {schema/Keyword schema/Any}
    command :- schema/Str
    args :- [schema/Str]]
-  (let [load-path (format "META-INF/jruby.home/bin/%s" command)]
-    (if-let [url (io/resource load-path (.getClassLoader org.jruby.Main))]
-      (cli-ruby! config (concat ["-e" (format "load '%s'" url) "--"] args)))))
+  (let [bin-dir "META-INF/jruby.home/bin"
+        load-path (format "%s/%s" bin-dir command)
+        url (io/resource load-path (.getClassLoader org.jruby.Main))]
+    (if url
+      (cli-ruby! config
+        (concat ["-e" (format "load '%s'" url) "--"] args))
+      (log/errorf "command %s could not be found in %s" command bin-dir))))
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
@@ -7,7 +7,7 @@
   (:import (com.puppetlabs.puppetserver PuppetProfiler JRubyPuppet)
            (puppetlabs.services.jruby.jruby_puppet_schemas JRubyPuppetInstance PoisonPill)
            (java.util HashMap)
-           (org.jruby CompatVersion RubyInstanceConfig$CompileMode)
+           (org.jruby CompatVersion Main RubyInstanceConfig RubyInstanceConfig$CompileMode)
            (org.jruby.embed ScriptingContainer LocalContextScope)
            (java.util.concurrent LinkedBlockingDeque TimeUnit)
            (clojure.lang IFn)))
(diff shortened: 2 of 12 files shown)


src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -173,3 +159,20 @@
...
+        argv (into-array String (concat ["-rjar-dependencies"] args))]
+    (.run main argv)))
+
+(schema/defn ^:always-validate cli-run! :- (schema/maybe jruby-schemas/JRubyMainStatus)
+  "Run a JRuby CLI command, e.g. gem, irb, etc..."
+  [config :- {schema/Keyword schema/Any}
+   command :- schema/Str
+   args :- [schema/Str]]
+  (let [load-path (format "META-INF/jruby.home/bin/%s" command)]
+    (if-let [url (io/resource load-path (.getClassLoader org.jruby.Main))]
```
camlow325: Should this print an error or something if `url` ends up being nil?  Would that happen if the resource weren't found?

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -15,53 +15,91 @@
...
+(schema/defn ^:always-validate managed-load-path :- [schema/Str]
+  "Return a list of ruby LOAD_PATH directories built from the
+  user-configurable ruby-load-path setting of the jruby-puppet configuration
+  section and prepending specific JRuby directories necessary for third party
+  extensions extensions.  See SERVER-297 and SERVER-538 for more info."
+  [ruby-load-path :- [schema/Str]
+   jruby-home :- schema/Str]
+  (->> ruby-load-path
+    (cons ruby-code-dir)
+    (cons (str jruby-home "/lib/ruby/1.9"))
```
camlow325: To make these a little more generic, I wonder if we could emulate the logic in JRuby's `LoadService` class by using the methods that JRuby internally uses to construct these paths.  Looks like that's done here: https://github.com/jruby/jruby/blob/1.7.20/core/src/main/java/org/jruby/runtime/load/LoadService.java#L262-L279

- .../lib/ruby/1.9 -> `RbConfigLibrary.getRubyLibDir`
- .../lib/ruby/shared -> `RbConfigLibrary.getRubySharedLibDir`
- .../lib/ruby/1.9/site_ruby -> `RbConfigLibrary.getSiteDir`

Looks like this might work with the `ScriptingContainer` based initialization since it appears you can do a `.getProvider().getRuntime() ` call to get the `Ruby` runtime that those helper methods take.  But maybe not for the `RubyInstanceConfig` based approach, though, so maybe not worth the trouble.  Curious if you'd looked into this at all.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -15,53 +15,91 @@
...
+    (assoc clean-env
+      "GEM_HOME" gem-home
+      "JARS_NO_REQUIRE" "true"
+      "JARS_REQUIRE" "false")))
+
+(schema/defn ^:always-validate managed-load-path :- [schema/Str]
+  "Return a list of ruby LOAD_PATH directories built from the
+  user-configurable ruby-load-path setting of the jruby-puppet configuration
+  section and prepending specific JRuby directories necessary for third party
+  extensions extensions.  See SERVER-297 and SERVER-538 for more info."
```
fpringvaldsen: I think this is a typo

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
jeffmccune: Good call, probably best to print an error.  command should only ever be "irb" or "gem" but in case this gets used by other things it probably shouldn't just silently do nothing, which is what would happen if the url ends up being non-truthy.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
jeffmccune: I haven't looked into this approach, but I like it.  I'll take a time-boxed look at making this more robust and less assumptive about the layout of these directories.

fpringvaldsen:
Looking good! I don't have any comments beyond what @camlow325 provided, so I'm :+1: once that feedback is addressed. I'm really glad we're consolidating all that repeat code.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -15,53 +15,91 @@
...
-    ;; JRuby 1.7.20, only JARS_NO_REQUIRE is honored.  Setting both of those
-    ;; here for forward compatibility.
-    (.setEnvironment (merge {"GEM_HOME" gem-home
-                             "JARS_NO_REQUIRE" "true"
-                             "JARS_REQUIRE" "false"}
-                            jruby-puppet-env))))
+  (let [jruby-home (.getHomeDirectory scripting-container)]
+    (doto scripting-container
+      (.setLoadPaths (managed-load-path ruby-load-path jruby-home))
+      (.setCompatVersion (CompatVersion/RUBY1_9))
```
camlow325: Should this be calling the `compat-version` function?

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -15,53 +15,91 @@
...
+      (.setCompileMode RubyInstanceConfig$CompileMode/OFF)
```
camlow325: Should this be calling the `compile-mode` function?

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -96,6 +134,35 @@
     (.runScriptlet "require 'jar-dependencies'")
     (.runScriptlet "require 'puppet/server/master'")))
 
+(schema/defn ^:always-validate config->puppet-config :- HashMap
+  "Given the raw jruby-puppet configuration section, return a
+  HashMap with the configuration necessary for ruby Puppet."
+  [config :- jruby-schemas/JRubyPuppetConfig]
+  (let [puppet-config (new HashMap)]
+    (doseq [[setting dir] [[:master-conf-dir "confdir"]
+                           [:master-code-dir "codedir"]
```
camlow325: Probably don't need `master-code-dir`, `master-run-dir`, or `master-log-dir` for the stable branch, right?  Not sure if it hurts but the underlying Puppet Ruby code wouldn't support them.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -96,6 +134,35 @@
     (.runScriptlet "require 'jar-dependencies'")
     (.runScriptlet "require 'puppet/server/master'")))
 
+(schema/defn ^:always-validate config->puppet-config :- HashMap
```
camlow325: Hmmm, actually does this even get called from anywhere?

test/unit/puppetlabs/services/jruby/jruby_puppet_core_test.clj
```diff
@@ -19,3 +64,55 @@
     (is (= 4 (jruby-core/default-pool-size 16)))
     (is (= 4 (jruby-core/default-pool-size 32)))
     (is (= 4 (jruby-core/default-pool-size 64)))))
+
+(deftest ^:integration cli-run!-test
```
camlow325: Questions aside about whether or not the remapping of stdin and stdout is a "scary" thing to do in the tests, I think these tests are pretty cool / useful.  Good to know we can do testing of this kind in Clojure rather than having to just resort to using Beaker.

camlow325:
:+1: with just a few comments on code that may not apply to the stable branch that was copied over from the master branch implementation.  For the sake of getting this in (and because I'm not sure it's tenable anyway), I'm totally good with punting on the question about using the `RbConfigLibrary` class to derive the default library paths instead of hardcoding them.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
jeffmccune: Fixed in 928eee0 `(SERVER-721) Consolidate JRuby environment handling`

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
jeffmccune: Good catch, yes, thanks.  Fixed in 713bb6b `(SERVER-721) Consolidate JRuby environment handling`

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
jeffmccune: Yes, thanks, fixed in fed5f54 `(SERVER-721) Consolidate JRuby environment handling`

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
jeffmccune: No...  Weird.  Removed in it's own commit since this might be a pain to merge back up into master.

jeffmccune:
>I'm totally good with punting on the question about using the RbConfigLibrary class to derive the default library paths instead of hardcoding them.

Understood.  I've addressed all of the other comments so I'll forego jumping into that rabbit hole for the time being.

Ready for another pass or merge.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
jeffmccune: UPDATE: foregoing this, see other comments for reasoning why.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
jeffmccune: Yep, fixed in be422b8 `(maint) Remove config->puppet-config unused in 1.x`

camlow325:
:+1: Will let Travis stew for a bit and merge on passing.

PR #605, opened 2015-06-09
(SERVER-723) Return good Content-Type for CA errors

camlow325:
This commit wraps plain text error responses to CA requests with a
Content-Type of "text/plain" for 'error' cases in which liberator does
not automatically set the Content-Type.  Previously, these error cases
would have resulted in a Content-Type of ";charset=UTF-8" being
delivered in the response.  Some clients would subsequently fail to be
able to parse the error response correctly.


fpringvaldsen:
:+1:

KevinCorcoran:
:+1: Nice work, thanks for fixing another one of my bugs.

PR #606, opened 2015-06-09
(SERVER-724) Fix Fedora 21 3.x failures

fpringvaldsen:
Do not run the 3.x compatibility tests on Fedora 21 agents, as
Puppet 3.x does not support Fedora 21.

### DIFF
--- a/acceptance/lib/puppetserver/acceptance/compat_utils.rb
+++ b/acceptance/lib/puppetserver/acceptance/compat_utils.rb
@@ -1,5 +1,11 @@
 def nonmaster_agents()
-  agents.reject { |agent| agent == master }
+  agents.reject { |agent| agent.host_hash[:platform].start_with?('fedora-21') || agent == master }
+end
+
+# This is all agents minus those legacy agents running Fedora 21, as it is not supported
+# by Puppet 3.x
+def valid_agents
+  agents.reject{ |agent| agent.host_hash[:platform].start_with?('fedora-21') && agent != master }
 end
 
 def apply_simmons_class(agent, studio, classname)
@@ -19,7 +25,7 @@ end
 def rm_vardirs()
   # In order to prevent file caching and ensure agent-master HTTP communication
   # during agent runs we blow away the vardirs, which contains the cached files
-  hosts.each do |host|
+  valid_agents.each do |host|
     vardir = on(host, puppet("config print vardir")).stdout.chomp
     on(host, "rm -rf #{vardir}")
   end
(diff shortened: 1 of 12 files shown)


nwolfe:
Just driving by here, but...

Is there a way we can prevent Fedora 21 from being a host, rather than skip it everywhere in the code?  That'd be a lot cleaner, especially since the code isn't actually platform specific and runs fine once you have an installation - it's just that we don't have packages for Fedora.

camlow325:
I'd agree with @nwolfe that it would be better, if possible, to skip 'fedora-21' via host configuration rather than hard-coding the constraint into the test code.

fpringvaldsen:
@nwolfe @camlow325 That's definitely possible. I talked to @waynr about it this morning, apparently it's somewhat complicated and will require us to open a QENG ticket (all the jobs are using the same test matrix, and we need to test Fedora 21 with other jobs, so we'd need to somehow remove it from the 3.x jbo specifically).

waynr:
As part of work for a hypothetical QENG ticket to implement this behavior I would add a combination-filter JJB parameter to the beaker jobs that would allow us to conditionally exclude this platform.

nwolfe:
> As part of work for a hypothetical QENG ticket to implement this behavior I would add a combination-filter JJB parameter to the beaker jobs that would allow us to conditionally exclude this platform.

In the interim could we override the relevant variables for the 3.x compatibility job and remove the troublesome platform(s)?

waynr:
We can't over ride the variables for a specific job since the variables get passed in from the kick-off job--we run different sets of platforms on "mergely" vs "periodic" kickoffs.

fpringvaldsen:
I'll open a QENG ticket to not run the 3.x tests on Fedora 21. @camlow325 @nwolfe Do you want me to close out this PR?

camlow325:
@waynr How soon do you think you might be able to get to this if we go with JJB configuration route?  If this isn't something you can turn around like w/in a day, I'd propose we put this PR to get back to green for now, file the QENG ticket to cover the JJB configuration approach, and file another ticket to rip this workaround back out later when the QENG ticket could be completed.

acceptance/lib/puppetserver/acceptance/compat_utils.rb
```diff
@@ -19,7 +25,7 @@ class { 'simmons':
 def rm_vardirs()
   # In order to prevent file caching and ensure agent-master HTTP communication
   # during agent runs we blow away the vardirs, which contains the cached files
-  hosts.each do |host|
+  valid_agents.each do |host|
```
camlow325: @nwolfe Did we have this as `hosts` before because we wanted to blow away the `vardir` on the master as well as all agents?

nwolfe: Yes - this needs to blow away the master vardir too

waynr:
@camlow325 I could probably get to it today.

PR #609, opened 2015-06-11
(MAINT) Switch puppet submodule to core Puppet and remove bundler pin

camlow325:
This commit contains a couple of changes:

1) Switch the ruby/puppet submodule back from a commit on
   camlow325/puppet to one on puppetlabs/puppet.  The prior submodule
   was used temporarily because of broken tests being skipped per
   PUP-4659.  The PUP ticket has been fixed now, though, so these
   tests should be runnable again.

2) Unpin bundler version 1.9.9.  This was done previously because
   1.10.0 would fail because of a rubygems version mismatch between the
   Puppet .gemspec requirement and JRuby rubygems version.  The Puppet
   .gemspec rubygems requirement is no longer present with this new
   Puppet submodule and, therefore, the bundler pin should no longer be
   needed.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "ruby/puppet"]
 	path = ruby/puppet
-	url = https://github.com/camlow325/puppet.git
+	url = https://github.com/puppetlabs/puppet.git
 [submodule "ruby/facter"]
 	path = ruby/facter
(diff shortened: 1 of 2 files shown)


fpringvaldsen:
:+1:

rlinehan:
:+1:

PR #616, opened 2015-06-17
(MAINT) Bump version to 2.1.2-SNAPSHOT for next release

camlow325:


camlow325:
Merging so we're back on a SNAPSHOT.

PR #624, opened 2015-06-24
(docs) Store navigation sidebar alongside Puppet Server docs content

nfagerlund:
In the docs.puppetlabs.com website, we use a central config file to assign an
HTML fragment as the navigation sidebar for a group of pages.

It used to be that we had to store these fragments in one specific directory,
but we recently revised our tools so we could load an HTML fragment from any
content directory.

This makes it possible to maintain the navigation TOC right alongside the
content that it governs. The PuppetDB team have been doing this for a few months
now, and it makes it much easier to add or delete a page.

Once this PR is merged, there will be two new tasks for maintaining these docs:

* If you add a page, add it to the TOC. Its URL should be an absolute path from
  the root of the docs site, and should use a {{server_version}} placeholder
  to represent the version number.
* When it's time to start preparing for a new X.Y release, change the
  server_version value at the top of the nav fragment.

### DIFF
--- a/documentation/_puppetserver_nav.html
+++ b/documentation/_puppetserver_nav.html
@@ -1,0 +1,39 @@
+{% capture server_version %}2.1{% endcapture %}
+
+
+<h3 id="puppetserver-{{server_version}}-docs">Puppet Server {{ server_version }} Docs</h3>
+
+    <ul>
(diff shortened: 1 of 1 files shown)


fpringvaldsen:
:+1:

dankreek:
:+1: LGTM

PR #627, opened 2015-06-26
(SERVER-768) trapperkeeper-authorization POC [FOR REVIEW ONLY]

jeffmccune:
This PR is for review only, these changes should not be merged into a mainline
branch because preview packages will be built from this topic branch and made
available for evaluation and further scoping in
[SERVER-769](https://jira.puppetlabs.com/browse/SERVER-769).

This patch replaces the client whitelist authorization strategy with a
rules based strategy using Brice's [trapperkeeper-authorization
library][tk-authz].  This patch is not meant to be the final solution
but is instead intended to evaluate and scope additional work on the
authorization behavior.

This patch replaces the whitelist checking ring middleware with a
middleware that loads rules from /etc/puppetlabs/puppetserver/rules.conf
if present.  If the file does not exist then hard-coded default rules
which deny everything are used.

Hacks for the POC related to the complexity of the changes required are:

This patch removes the previous behavior of being able to turn off
access control entirely.  Similar behavior is provided using permissive
authorization rules, however this patch requires an SSL certificate from
the client.  If no client certificate is present then the request is
forbidden.

 * Decision needed: Do we need to preserve the behavior of disabling authz
   entirely?

The following tests are removed because they make assumptions about
client-whitelist.  These tests are expected to be fairly straightforward
to update.  The test set-up behavior needs to replace the
client-whitelist configuration with an approach that configures the
authz rules.  All of these tests should be able to be treated in the
same manner:

 * puppet-environments-int-test
 * puppet-admin-int-test
 * jruby-pool-int-test

There will be some complexity around this test, which may require
special handling relative to the mechanical nature of replacing the
client-whitelist configuration:

 * jruby-pool-int-test (The wait-for-new-pool function enters an
   infinite loop with the certificate whitelist implementation removed.)

Similarly, the legacy-routes-test/legacy-routes test completely disables
access control, however there should be no additional complexity fixing
this test beyond either adopting the new way to disable AuthZ entirely
(if we do that) or configuring the test to make an authorized request by
adjusting the rules.

 * legacy-routes-test/legacy-routes

[tk-authz]: https://github.com/masterzen/trapperkeeper-authorization


src/clj/puppetlabs/puppetserver/ringutils.clj
```diff
@@ -85,6 +109,28 @@
...
+  [request :- RingRequest]
+  (let [{:keys [ssl-client-cert]} request]
+    (if ssl-client-cert (get-cert-subject ssl-client-cert))))
+
+(schema/defn ^:always-validate authorized? :- schema/Bool
+  "(SERVER-768) replacement for client-allowed-access?"
+  [rules :- rules/Rules
+   request :- RingRequest]
+  (if-let [name (request->name request)]
+    (let [request-for-tk-authz-schema (merge {:remote-addr ""} request)
```
camlow325: Were you finding cases where the `:remote-addr` key wasn't present in the Ring request?  If it isn't a required key, might be better eventually to have `trapperkeeper-authorization` tolerate its absence.

jeffmccune: Ring documentation guarantees it will always be present [1].

I ran into this key being absent in our tests which construct mock request objects.  Presumably this means the mock objects aren't to spec, but I didn't go much further than this conclusion as the problematic tests got ripped out anyway.

[1]: https://github.com/mmcgrana/ring/blob/890ffee2e9d57a55a35804745fd1fee9205dc29c/SPEC#L46-L48

src/clj/puppetlabs/puppetserver/ringutils.clj
```diff
@@ -85,6 +109,28 @@
...
+  the request.  Return nil if the client cert is absent or cannot be parsed."
+  [request :- RingRequest]
+  (let [{:keys [ssl-client-cert]} request]
+    (if ssl-client-cert (get-cert-subject ssl-client-cert))))
+
+(schema/defn ^:always-validate authorized? :- schema/Bool
+  "(SERVER-768) replacement for client-allowed-access?"
+  [rules :- rules/Rules
+   request :- RingRequest]
+  (if-let [name (request->name request)]
```
camlow325: If you had a default "allow-all" rules.conf file and hit an endpoint with no SSL client certificate, should you be able to call into `rules/allowed?` with a `nil` value for `name` and have it return a truthy value for `authorized`?  Maybe `trapperkeeper-authorization` doesn't handle this today but seems like it should be capable of doing that.

src/clj/puppetlabs/puppetserver/ringutils.clj
jeffmccune: Errr...  Scratch that, there is at least one offending test left in place, otherwise this merge wouldn't be necessary.

camlow325: Ah, okay, makes sense.  When tests are resurrected to replace this, seems like it would be better to work this into the mocked request than to have to put this into production code.

jeffmccune: Agreed

src/clj/puppetlabs/puppetserver/ringutils.clj
jeffmccune: I spent a few minutes thinking about this...  My thinking was that an authorization rule that allows all clients still makes the assumption that the client is authenticated.  This led me to think a client still needs at least a client cert given the current design of trapperkeeper-authorization.

With that said, it seems like we'll need _some_ way to handle SSL clients that don't have a certificate and as such don't have a name.  I think this should be a change to the upstream authorization library as we need some first-class way to say, "this client is authentic, it's just authentic without any way to identify it, so it has no name."

I'd hate to overload something like nil or the empty string, but it may not be so terrible to do so.

camlow325: I noticed that the library has its own middleware function, `wrap-authorization-check`, which derives the `name` a little bit differently.  It pulls it from the SSL client certificate, if available, else falls back to a reverse lookup on the `:remote-addr` -- https://github.com/masterzen/trapperkeeper-authorization/blob/master/src/puppetlabs/trapperkeeper/authorization/ring_middleware.clj

Presumably, this is more like how this was done with legacy auth.conf handling.  If we were using this middleware instead of what is being done here, I think my comment here would be irrelevant.  Is there a reason you didn't go with this approach?

camlow325: Although if we go with the approach of allowing the `name` to be derived from a reverse lookup on `:remote-addr` as a fallback that we'd still also need something like the "auth" keyword from the old 'auth.conf' - https://docs.puppetlabs.com/puppet/latest/reference/config_file_auth.html#auth - in order to inform per rule as to whether or not an "unauthenticated" request should be allowed.

Where the fallback `:remote-addr` reverse lookup seems like it could open up a security hole for a client to trick the master into trusting it based on its ip address being that of a node that should legitimately be trusted, seems like it might be better to not carry forward with the fallback lookup.  Curious to get @ahpook's thoughts on that.

In any event, still a good discussion to have about whether we need to introduce some sort of a "authorized if actually authentic" distinction into the definition of the auth rules.  This might also inform how we approach compatibility for the use of `X-Client-*` headers for plaintext connections.

ahpook: as I put in [the mailing list thread](https://groups.google.com/d/msg/puppet-dev/P6vas3tvUzM/L7Gs1MvPGnIJ), I'm leaning pretty strongly towards the IP-based auth being both overly complicated and not necessary. It's more prone to spoofing, as you note, and every use-case seems to be better satisfied by having better access to certificate data. 

That said, I agree there should be a construct to permit requests with no client cert presented; at a minimum the retrieval of freshly-signed certs by nodes (though itself problematic) needs this kind of access.

masterzen: @camlow325 indeed, the library mimicks the puppet auth.conf behavior (ie reverse look up if no name is given).

camlow325:
@jeffmccune Since this is just for a POC, how long were you hoping to keep this PR open?  No rush from my side but wanted to see if you had already gotten what you were hoping to get out of it for now.

PR #629, opened 2015-06-29
(SERVER-251) Add a couple of more jruby flush tests

camlow325:
This commit adds a couple of more JRuby pool flush tests:

1) Make multiple JRuby flush requests back-to-back and ensure that
   nothing obviously *bad* happens.

2) Ensure that a file handle open from Ruby code run in a
   `ScriptingContainer` is not automatically invalidated when a
   pool flush is requested.


camlow325:
Ah, crap.  Wrong branch target.  Closing....

PR #630, opened 2015-06-29
(SERVER-251) Add a couple of more jruby flush tests

camlow325:
This commit adds a couple of more JRuby pool flush tests:

1) Make multiple JRuby flush requests back-to-back and ensure that
   nothing obviously *bad* happens.

2) Ensure that a file handle open from Ruby code run in a
   `ScriptingContainer` is not automatically invalidated when a
   pool flush is requested.

### DIFF
--- a/test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
+++ b/test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
@@ -141,6 +141,38 @@
         ;; now the pool is flushed, so the constants should be cleared
         (is (true? (verify-no-constants pool-context 4)))))))
 
+(deftest ^:integration admin-api-multiple-flush-jruby-pool-test
+  (testing "Flushing the jruby pool multiple times results in all new JRuby instances"
+    (bootstrap/with-puppetserver-running
+      app
+      {:puppet-admin {:client-whitelist ["localhost"]}
+       :jruby-puppet {:max-active-instances 4}}
+      (let [jruby-service (tk-app/get-service app :JRubyPuppetService)
+            context (tk-services/service-context jruby-service)
+            pool-context (:pool-context context)
+            times-to-flush 3]
+        ;; set a ruby constant in each instance so that we can recognize them
+        (is (true? (set-constants-and-verify pool-context 4)))
+        (let [times-flushed (atom 0)
+              _ (add-watch (:pool-agent pool-context)
+                           :flush-callback
+                           (fn [k _ _ _]
+                             (when (= k :flush-callback)
+                               (swap! times-flushed inc))))
+              wait-until (+ (System/currentTimeMillis) 300000)]
+          (dotimes [_ times-to-flush]
+            (is (true? (trigger-flush ssl-request-options))))
+          (while (and (< (System/currentTimeMillis) wait-until)
+                      (< @times-flushed times-to-flush))
+            (Thread/sleep 100))
+          (is (< (System/currentTimeMillis) wait-until)
+              (str "Pool not flushed expected number of times before timeout"))
+          (is (= @times-flushed times-to-flush)
+              "Pool flushed an unexpected number of times")
+          (remove-watch (:pool-agent pool-context) :flush-callback))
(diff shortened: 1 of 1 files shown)


test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -141,6 +141,38 @@
         ;; now the pool is flushed, so the constants should be cleared
         (is (true? (verify-no-constants pool-context 4)))))))
 
+(deftest ^:integration admin-api-multiple-flush-jruby-pool-test
```
camlow325: This test is kinda weak, but I couldn't think of a good system-level way to intercept the flushing process midstream and insert another request while being able to predict whether or not an instance from the new pool would be created before a second flush would come along and wipe it out first.  So this test doesn't ensure that the flushes aren't done synchronously - i.e., that the flush isn't completed before each HTTP response to the flush request is delivered - and doesn't determine the state of each of the pool instances after the first or second flush (only after the last one).  It could, however, expose any problems that might arise from the pool not being updated atomically like it should and that all of the flush requests are at least processed.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -141,6 +141,38 @@
...
+            times-to-flush 3]
+        ;; set a ruby constant in each instance so that we can recognize them
+        (is (true? (set-constants-and-verify pool-context 4)))
+        (let [times-flushed (atom 0)
+              _ (add-watch (:pool-agent pool-context)
+                           :flush-callback
+                           (fn [k _ _ _]
+                             (when (= k :flush-callback)
+                               (swap! times-flushed inc))))
+              wait-until (+ (System/currentTimeMillis) 300000)]
```
camlow325: I put the 5 minute time limit on this to ensure that the test doesn't run in an infinite loop if the flushes never happen.  5 minutes is fairly arbitrary but I figured it should be a reasonable upper bound.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -166,6 +198,42 @@
...
+                         (str "$unique_file = "
+                              "Puppet::FileSystem::Uniquefile.new"
+                              "('hold-instance-test-', './target')"))
+          (try
+            ;; trigger a flush
+            (is (true? (trigger-flush ssl-request-options)))
+            ;; wait for the new pool to become available
+            (wait-for-new-pool jruby-service)
+
+            (is (nil? (.runScriptlet sc "$unique_file.close"))
```
camlow325: This line is basically what would cause the IOException / BadFileDescriptor errors that we were seeing back before SERVER-38 was fixed, #227.  A nil return here is what we want - indicating that JRuby hasn't unintentionally stomped the file descriptor out from underneath the container as a result of any terminates being called on other `ScriptingContainer`s.

cprice404:
:see_no_evil:  +1

fpringvaldsen:
:+1:

cprice404:
hmmm.  I didn't run these locally before giving a +1.  It seems like they take like 20 minutes to run.... :(   not sure that's tolerable

cprice404:
yeah, on my laptop, the total runtime of the tests goes from 6 min to 12 min with this patch... @camlow325 do you have any thoughts on how we could address that?

camlow325:
I suspect that the `admin-api-multiple-flush-jruby-pool-test` test is the one that takes most of the time since it has to wait for 4 flushes of 4 pool instances to complete.  I'd thought about maybe mocking out some of the loading of Puppet code into instances in order to speed them up, but this would probably reduce the value of this test even further.  I'm not sure it adds enough value overall so I guess we could just delete it.

I think the `hold-file-handle-on-instance-while-pool-flush-in-progress-test` is potentially more useful since it is related to functionality that bit us in the past - the 'bad file descriptor' problem when a file was held open around a `ScriptingContainer.terminate`.  I could drop the number of JRuby instances it uses from 4 to 2 since that would probably be enough for what it would need to do.

PR #632, opened 2015-07-07
(PE-8226) log jruby events

cprice404:
This PR does the following:

* Refactors a few functions to make them more re-usable
* Adds an `action` argument to all of the functions related to borrowing a JRuby instance to the pool.  This `action` value is a description of what the JRuby instance will be used for, and can be used in log messages and other places to help provide insight into the state of the running system.
* Adds a `jruby-event-logging-service`, which logs debugging information when interesting events related to the JRuby pool occur.  This implementation is swappable, see PE-8226 for more details.

### DIFF
--- a/dev/bootstrap.cfg
+++ b/dev/bootstrap.cfg
@@ -1,5 +1,6 @@
 puppetlabs.services.request-handler.request-handler-service/request-handler-service
 puppetlabs.services.jruby.jruby-puppet-service/jruby-puppet-pooled-service
+puppetlabs.services.jruby.jruby-event-logger-service/jruby-event-logger-service
 puppetlabs.services.puppet-profiler.puppet-profiler-service/puppet-profiler-service
 puppetlabs.trapperkeeper.services.webserver.jetty9-service/jetty9-service
 puppetlabs.trapperkeeper.services.webrouting.webrouting-service/webrouting-service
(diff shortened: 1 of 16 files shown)


fpringvaldsen:
I think we should open a SERVER ticket for this and change the commit messages to reference that ticket instead.

fpringvaldsen:
I like this approach, :+1:.

KevinCorcoran:
This looks sane to me, but the `JRubyEventHandlerService` and extra dependency seems a bit heavyweight.  An alternative approach comes to mind ...

Add a new service function like `register-event-handler!` to the `JRubyPuppetService` protocol.  This function would simply take a function (callback) as a parameter, and it would document when it would invoke the callback and what data it would be passing in.  Perhaps it's cleaner as multiple functions - but the basic idea would be the same.

Along with feeling a bit "lighter-weight", this has the advantage of not requiring an extra service, in case we don't really need this extra behavior in certain cases.

I'm not sure it's a better idea, but seems worth consideration - and probably easy to hack-up a POC for comparison.

cprice404:
@KevinCorcoran I like that idea.  Will look into it.

cprice404:
Closing this until I have time to play around with the idea that @KevinCorcoran proposed.

PR #633, opened 2015-07-08
(maint) Fix dev sample config file

KevinCorcoran:
Without this change, the dev sample config file cannot be used without
modification - if they are, the server will crash during startup with an
error: "Permission denied - /var/log/puppetlabs"

This error was introduced by f7627a78aa25e647501c22dac4f6d318c6874137


cprice404:
:see_no_evil:

camlow325:
:+1:

PR #634, opened 2015-07-10
(SERVER-251) Tweak jruby-flush tests because of performance issues

camlow325:
This commit modifies some of the work previously committed for
SERVER-251, mostly due to concerns about the length of execution time
added to the suite.

1) The `admin-api-multiple-flush-jruby-pool-test` was eliminated.  The
value of the test was already a bit dubious due to some of the variable
timing in the execution of the test and limited ability to validate the
system's behavior with concurrent flush requests.  More importantly, the
test appeared to add 5 minutes or so to the overall execution time for
the test suite - increasing from 6 minutes or so to 11 minutes or so.
This commit removes the test completely.

2) The `hold-instance-while-pool-flush-in-progress-test` was modified to
just try to initialize 2 `jruby-puppet` instances instead of 4.  This
should speed the execution of the test up a bit.  The test overall may
still take around 10 seconds to run; measurements were inconsistent from
local runs.

### DIFF
--- a/test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
+++ b/test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
@@ -141,38 +141,6 @@
         ;; now the pool is flushed, so the constants should be cleared
         (is (true? (verify-no-constants pool-context 4)))))))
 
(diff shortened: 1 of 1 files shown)


cprice404:
+1

fpringvaldsen:
:+1:

PR #635, opened 2015-07-13
(SERVER-781) Add hiera as submodule

cprice404:
I chose version 1.3.4 as that was the version in the latest PE 3.8.1.  We can bump this up to newer versions for master branch, etc., as we do the merge-up from here.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "ruby/facter"]
 	path = ruby/facter
 	url = https://github.com/puppetlabs/facter.git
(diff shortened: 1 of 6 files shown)


camlow325:
:+1:

Also did a "functional" test of this by starting up a server from `lein run` on the branch using a "puppetserver.conf" which referenced the hiera submodule directory from the `ruby-load-path` and with some code loaded on the server that would perform a parameter lookup from hiera.  I didn't get any errors so seems reasonable to me.

rlinehan:
:+1: This looks good to me.

PR #637, opened 2015-07-13
(SERVER-765) Qualify endpoints supported by `allow-header-cert-info`

camlow325:
This commit includes some updates to the documentation to qualify
which endpoints are supported by the `allow-header-cert-info` setting
in the `master.conf` section.

### DIFF
--- a/documentation/configuration.markdown
+++ b/documentation/configuration.markdown
@@ -137,6 +137,14 @@ In a default installation, this file doesn't exist. You'll need to create it if 
     This setting is used to enable [external SSL termination.](./external_ssl_termination.markdown) If enabled, Puppet Server will ignore any actual certificate presented to the Jetty webserver, and will rely completely on header data to authorize requests. This is very dangerous unless you've secured your network to prevent any untrusted access to Puppet Server.
 
     You can change Puppet's `ssl_client_verify_header` setting to use another header name instead of `X-Client-Verify`; the `ssl_client_header` setting can rename `X-Client-CN`. The `X-Client-Cert` header can't be renamed.
+    
+    Note that this setting only applies to HTTP endpoints served by the "master"
+    service.  The applicable endpoints include those listed in the 
+    "Configuration Management Services", "Informational Services", and
+    "V2 HTTP API" sections of the [Puppet HTTP API](https://github.com/puppetlabs/puppet/blob/3.8.1/api/docs/http_api_index.md)
+    page.  Note that this setting does not apply to the endpoints listed under
+    the "SSL Certificate Related Services" section of the "Puppet HTTP API" page
+    or to any of the [Puppet Admin API](#puppetserverconf) endpoints. 
 
 ~~~
 master: {


camlow325:
@nfagerlund, I'd like to get your take on this one.  Would it be better to list out all of the endpoints individually or just link to them in other Puppet HTTP API documentation like I did on the first pass at this PR?

fpringvaldsen:
:see_no_evil:

nfagerlund:
Aw yikes. And this is for the 1.x versions of Puppet Server, right? And we'll need different text when we merge forward into master? >:[

camlow325:
@nfagerlund Yeah, we would need different text for when we merge forward to master.  We'd need different text whether we went with the "point to external links" approach or include references to the endpoints directly, though, since all of the endpoints changed.  Are you saying that you want to go with a different approach for this PR?  I'm definitely open to whatever you think is best.

nfagerlund:
I'm still thinking about it. just a sec!

documentation/configuration.markdown
```diff
@@ -137,6 +137,14 @@ In a default installation, this file doesn't exist. You'll need to create it if
     This setting is used to enable [external SSL termination.](./external_ssl_termination.markdown) If enabled, Puppet Server will ignore any actual certificate presented to the Jetty webserver, and will rely completely on header data to authorize requests. This is very dangerous unless you've secured your network to prevent any untrusted access to Puppet Server.
 
     You can change Puppet's `ssl_client_verify_header` setting to use another header name instead of `X-Client-Verify`; the `ssl_client_header` setting can rename `X-Client-CN`. The `X-Client-Cert` header can't be renamed.
+    
+    Note that this setting only applies to HTTP endpoints served by the "master"
+    service.  The applicable endpoints include those listed in the 
+    "Configuration Management Services", "Informational Services", and
+    "V2 HTTP API" sections of the [Puppet HTTP API](http://docs.puppetlabs.com/references/3.7.latest/developer/file.http_api_index.html)
```
nfagerlund: We should link to here instead: https://github.com/puppetlabs/puppet/blob/3.8.1/api/docs/http_api_index.md Docs team has decided to treat the YARD version of the API docs as abandoned, using the GitHub copy as canonical for 3.x and the version [here](http://docs.puppetlabs.com/puppet/latest/reference/http_api/http_api_index.html) as canonical for 4.x and up.

nfagerlund:
@camlow325 After considering for a bit, I think what you have here is good for 1.x. We can work on something more elegant for 2.x later. :+1:

documentation/configuration.markdown
camlow325: Thanks, @nfagerlund.  I updated the PR in 7707d07 to point at the 3.8.1 link that you listed.

FYI, in case you hadn't seen this recently, I noticed that the endpoint references under the sections on the github page don't function as links, e.g., "{file:api/docs/http_catalog.md Catalog}".  Not sure if that's intentional?

nfagerlund: @camlow325 It's an artifact of an ancient decision that we later reversed (because frames are a bad tool for displaying anything except ruby method docs). In https://github.com/nfagerlund/puppet/commit/9ce3dbfc27bc34f5b32945eed2ca9f3bc413b5df, I fixed all those on stable, but they're still there in historical versions like 3.8.1.

fpringvaldsen:
This has two :+1:'s so I'm going to go ahead and merge.

PR #638, opened 2015-07-16
(MAINT) Merge up from stable branch

cprice404:


KevinCorcoran:
It looks like this merge didn't quite go right, as it doesn't contain any of the commits ... just the result of the merge.  And the commit doesn't seem to be a merge commit either, since it only has one parent ... not sure exactly what's up with that.

Anywho, I opened #639 which seems like a more correct merge.

rlinehan:
Going to close this since #639 was merged.

PR #640, opened 2015-07-17
(maint) Refactor handle-request for reuse

KevinCorcoran:
This change doesn't have any value for this project on its own, but they
are necessary for downstream consumers (like PE) which will consume this
code directly.

### DIFF
--- a/src/clj/puppetlabs/services/request_handler/request_handler_core.clj
+++ b/src/clj/puppetlabs/services/request_handler/request_handler_core.clj
@@ -262,21 +262,34 @@
       (ring-response/status http-status)
       (ring-response/content-type "text/plain")))
 
+(defn handle-request*
+  [request jruby-instance config]
+  (->> request
+    wrap-params-for-jruby
+    (as-jruby-request config)
+    clojure.walk/stringify-keys
+    make-request-mutable
+    (.handleRequest jruby-instance)
+    response->map))
+
+(defmacro with-error-handling
+  [request & body]
+  `(sling/try+
+     ~@body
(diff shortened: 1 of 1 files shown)


fpringvaldsen:
:+1:

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
```diff
@@ -262,21 +262,34 @@
       (ring-response/status http-status)
       (ring-response/content-type "text/plain")))
 
+(defn handle-request*
```
camlow325: Assuming we're going to end up calling `handle-request*` and `with-error-handling` from a PE version of `handle-request`, I wonder if it would be better to leave `handle-request*` and `with-error-handling` in "public" - since a namespace outside of this one would be referencing them.

KevinCorcoran: sure, that sounds reasonable

camlow325:
Even if we don't end up implementing the PE side of the `handle-request` change in PE 3.8.x, I wonder if it might be better to target this change at the `stable` branch of `puppet-server` first and merge up to `master` from there.  Could help reduce merge conflicts later in the event we decide to do yet another big refactor around the jruby service which affects both `stable` and `master`.  Seems like @cprice404 might have been planning something like that soon with respect to some of the metrics work he's doing?

camlow325:
:+1: with a couple of minor comments.

KevinCorcoran:
@camlow325 - please see #642

PR #641, opened 2015-07-17
(SERVER-765) Updated endpoint links for `allow-header-cert-info` docs

camlow325:
This commit updates the endpoint links that the `allow-header-cert-info`
docs reference.  The links should now point to locations which are valid
for Puppet Server 2.x / Puppet 4.x.

### DIFF
--- a/documentation/configuration.markdown
+++ b/documentation/configuration.markdown
@@ -169,12 +169,11 @@ In a default installation, this file doesn't exist. You'll need to create it if 
     You can change Puppet's `ssl_client_verify_header` setting to use another header name instead of `X-Client-Verify`; the `ssl_client_header` setting can rename `X-Client-CN`. The `X-Client-Cert` header can't be renamed.
     
     Note that this setting only applies to HTTP endpoints served by the "master"
-    service.  The applicable endpoints include those listed in the 
-    "Configuration Management Services", "Informational Services", and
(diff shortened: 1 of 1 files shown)


camlow325:
@nfagerlund - This is just a simple update of links for the master branch of what was integrated to the stable branch for #637.  Did you have something else in mind for the master branch text?

nfagerlund:
@camlow325 We can simplify it even further by removing the text about "Configuration Management Services" et al. Instead, we should just say it only applies to endpoints under the `/puppet/v3` API (using the link you already switched it to).

camlow325:
@nfagerlund Good suggestion, thanks.  Updated this in 9c719f7.  Hope that looks okay.

nfagerlund:
The language about "page" is still a bit confusing, since it's all on the same page. Instead, I'd say

> Note that this setting only applies to HTTP endpoints in the [Puppet V3 HTTP API](https://docs.puppetlabs.com/puppet/4.2/reference/http_api/http_api_index.html#puppet-v3-http-api)
> ...
> ...

like that

camlow325:
@nfagerlund Sounds good.  Updated as 44a2db4.

nfagerlund:
:+1:

camlow325:
@fpringvaldsen, I think you reviewed the version of this that went into stable already.  You good merging this since we already have @nfagerlund's :+1:?

PR #642, opened 2015-07-20
(maint) Refactor handle-request for reuse

KevinCorcoran:
This change doesn't have any value for this project on its own, but it
is necessary for downstream consumers (like PE) which will consume this
code directly.


fpringvaldsen:
Looks like the Travis failure was due to a transient connection error, so I've restarted it.

fpringvaldsen:
:+1:

camlow325:
:+1:

PR #644, opened 2015-07-21
(maint) Update submodule and tests to Puppet 3.8.1

jeffmccune:
Without this patch the tests are mixing and matching the version of Puppet
being installed.  For example, the legacy puppet agent tests are installing
3.8.1 while the unit and integration tests run with a Puppet 3.7 version from
the submodule.

This patch addresses the problem by updating the behavior of the tests to use
the same version, 3.8.1.


fpringvaldsen:
:+1:

camlow325:
:+1:

PR #645, opened 2015-07-22
(SERVER-811) port request handler service to middleware-style

cprice404:
This commit modifies the `request-handler` service to use more
of a ring-middleware style of composing functions for building
up the final request handler function.  This should make it easier
to share code / extend the service in subsequent development.

### DIFF
--- a/src/clj/puppetlabs/services/request_handler/request_handler_core.clj
+++ b/src/clj/puppetlabs/services/request_handler/request_handler_core.clj
@@ -262,34 +262,44 @@
       (ring-response/status http-status)
       (ring-response/content-type "text/plain")))
 
+(defn wrap-with-error-handling
+  "Middleware that wraps a JRuby request with some error handling to return
+  the appropriate http status codes, etc."
+  [f]
+  (fn [request]
+    (sling/try+
+      (f request)
+      (catch bad-request? e
+        (output-error request e 400))
+      (catch jruby-timeout? e
+        (output-error request e 503)))))
+
+(defn wrap-with-jruby-instance
+  "Middleware fn that borrows a jruby instance from the `jruby-service` and makes
+  it available in the request as `:jruby-instance`"
+  [f jruby-service]
+  (fn [request]
+    (jruby/with-jruby-puppet jruby-instance jruby-service
+      (f (assoc request :jruby-instance jruby-instance)))))
+
+(defn jruby-request-handler
+  "Build a request handler fn that processes a request using a JRubyPuppet instance"
+  [config]
+  (fn [request]
+    (->> request
+      wrap-params-for-jruby
+      (as-jruby-request config)
+      clojure.walk/stringify-keys
+      make-request-mutable
+      (.handleRequest (:jruby-instance request))
+      response->map)))
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Public
 
-(defn handle-request*
-  [request jruby-instance config]
-  (->> request
-    wrap-params-for-jruby
-    (as-jruby-request config)
-    clojure.walk/stringify-keys
-    make-request-mutable
-    (.handleRequest jruby-instance)
-    response->map))
-
-(defmacro with-error-handling
-  [request & body]
-  `(sling/try+
-     ~@body
-     (catch bad-request? e#
-       (output-error ~request e# 400))
-     (catch jruby-timeout? e#
-       (output-error ~request e# 503))))
-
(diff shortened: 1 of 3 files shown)


cprice404:
@KevinCorcoran you and I are doing some of the same things w/rt attempting to extend the request handler in PE.  I think this will make it easier to avoid merge conflicts.

KevinCorcoran:
Yo,

Here's how I've been using the current version of the code in PE Puppet Server:
```clj
(core/with-error-handling request
  (jruby/with-jruby-puppet jruby-instance jruby-service
    ; do some stuff here ... 
    ;
    ; ... and then eventually
    (core/handle-request* request jruby-instance config))))
```

Looking at this diff, it looks like that would get converted to
```clj
(let [handler (fn [req]
                ; other stuff ...
                ((core/jruby-request-handler config) req))
      wrapped-handler (-> handler
                          core/wrap-with-error-handling
                          (core/wrap-with-jruby-instance jruby-service))]
  (wrapped-handler request)  
```
Ugly code / terrible variable names aside, does that seem about right to you?  (I tried to test this out, but couldn't easily do it since this is targeted at `stable` and doesn't apply cleanly on to `master`, which is what my work is based on.)

If that's about right, it seems quite a bit more difficult to understand than the previous version of the code ... but maybe I just have to live with that.

cprice404:
I looked at your PR for PE and it looked to me like you'd just need to write your logic as its own self-contained middleware function and then insert it into the list in the main `build-handler` function at the appropriate spot... I'm not seeing why you'd need all that extra `let` stuff above?

cprice404:
The merge conflicts for pulling this up to stable are.... two ruby spec tests.  :/

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
```diff
@@ -262,34 +262,44 @@
...
+  "Build a request handler fn that processes a request using a JRubyPuppet instance"
+  [config]
+  (fn [request]
+    (->> request
+      wrap-params-for-jruby
+      (as-jruby-request config)
+      clojure.walk/stringify-keys
+      make-request-mutable
+      (.handleRequest (:jruby-instance request))
+      response->map)))
```
cprice404: Add a function here that returns a one-arg fn, accepting a 'request', and doing whatever stuff you need it to do.

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
```diff
@@ -262,34 +262,44 @@
...
 
-(defn handle-request
-  [request jruby-service config]
-  (with-error-handling request
-    (jruby/with-jruby-puppet jruby-instance jruby-service
-      (handle-request* request jruby-instance config))))
+(defn build-request-handler
+  "Build the main request handler fn for JRuby requests."
+  [jruby-service config]
+  (-> (jruby-request-handler config)
```
cprice404: Add your function to the middleware chain here.

cprice404:
Ah, something else goes wrong with the merge up to master that doesn't cause a merge conflict.  I'll see if I can put a branch together for you.

cprice404:
The only other test failure from a merge of this up to master was an issue with the test, so, it should be safe to test this code via a simple merge into master on your local box, or, if you prefer, here's a branch that fixes the 3 test problems mentioned above:

https://github.com/cprice404/puppet-server/tree/scratch/master/jruby-request-handler-middleware

cprice404:
@KevinCorcoran here's a glimpse of what my pe branch of this looks like:

https://github.com/cprice404/pe-puppet-server-extensions/blob/ebddb56d1622de423e65ba72877d370c342b6950/src/clj/puppetlabs/enterprise/services/request_handler/pe_request_handler_core.clj#L37-L62

I'd guess that yours would end up being really similar.

KevinCorcoran:
> I looked at your PR for PE and it looked to me like you'd just need to write your logic as its own self-contained middleware function and then insert it into the list in the main build-handler function at the appropriate spot... I'm not seeing why you'd need all that extra let stuff above?

Yeah, this occurred to me as well, after I wrote-up that initial version.  I wasn't getting it at first, but in retrospect it's obvious what you were going for with the middleware approach.

KevinCorcoran:
:+1: 

Looks fine to me.  I'm confident that this will work out for the changes I'm working on in PE.

cprice404:
OK well if this doesn't get merged soon perhaps you should rebase your branch on it?  If there are cosmetic fixes feel free to close this, add them, and get your own PR merged, I can rebase my stuff.

I'm not sure how much I'll be online from now until when I get back to Belfast but I'll be working some offline and will be proceeding under the assumption that we'll move forward with a middleware approach like this.  Thanks!

KevinCorcoran:
Cool, that all sounds good.

rlinehan:
:+1:

PR #646, opened 2015-07-25
(PE-8226) jruby events

cprice404:
This PR adds event notifications for whenever something interesting happens related to the JRuby pool.  It's a pre-cursor to PE-8226, and replaces my earlier attempt at this that was in PR #632 .

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
@@ -165,7 +165,6 @@
 
 (def JRubyPuppetBorrowResult
   (schema/pred (some-fn nil?
-                        poison-pill?
                         retry-poison-pill?
                         jruby-puppet-instance?)))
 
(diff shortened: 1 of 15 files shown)


fpringvaldsen:
This looks good to me, and I definitely like this approach. :+1:

jpinsonault:
This looks really good, I'm :+1: It looks like there was some discussion on https://github.com/puppetlabs/pe-puppet-server-extensions/pull/283 that affects this, so I won't hit the button yet. Maybe @KevinCorcoran knowns

KevinCorcoran:
I would like to have a look at quick look at this before it gets merged, but if there is any rush, you don't need to wait for me.

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
```diff
@@ -32,16 +31,21 @@
...
-            (assoc :borrow-timeout (:borrow-timeout config))))))
+            (assoc :borrow-timeout (:borrow-timeout config))
+            (assoc :event-callbacks (atom []))))))
 
   (borrow-instance
-    [this]
-    (let [pool-context (:pool-context (tk-services/service-context this))
-          borrow-timeout (:borrow-timeout (tk-services/service-context this))]
-      (core/borrow-from-pool-with-timeout pool-context borrow-timeout)))
+    [this action]
```
nwolfe: The `action` seems like something that would be optional.  Should we provide another arity these functions, or perhaps like what we did in jvm-ssl-utils and support an `options` map?

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -183,3 +182,39 @@
...
+
+(defn event-type-borrowed?
+  [e]
+  (= :instance-borrowed (:type e)))
+
+(defn event-type-returned?
+  [e]
+  (= :instance-returned (:type e)))
+
+(def JRubyEventAction
```
nwolfe: While looking over this PR it took me a bit to realize what `action` represented.  I thought it was going to be a function, considering the name, but then the `schema/Any` seemed a bit odd.

Turns out it represents a "reason", and it's generally a map (at least that's what the docstrings below say).

Since the data type can be anything, what are we thinking we would do with the data?  I see you mention logging and metrics, but if it can be anything what more can we do than call `str` on it?

So I guess my questions are:
- What about a different name, like `JRubyEventReason`?  The term "action" seems overloaded here
- Why is the datatype `schema/Any`?

nwolfe:
:see_no_evil:, just my naive questions

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
jpinsonault: For what it's worth I was also confused by `action`, thinking it was related to an `event` or a callback at first

cprice404: Yeah.  Not in love with that name.  To address @nwolfe 's earlier question, it is definitely required, because for metrics and logging and other things we always want to be able to show *some* indication of what triggered the borrow/return event.

If "Reason" seem like a better name than "Action" to everyone, I would definitely be happy to change it.  @jpinsonault does that seem more clear to you?  Or, do you have other suggestions for the name?

jpinsonault: I like "reason", it descriptive and simple.

nwolfe:
This will be closed when I adjust the stable branch for SERVER-828.  I'll re-open the PR against the new 1.x branch

nwolfe:
@cprice404 Reopened against 1.x: #658

PR #647, opened 2015-07-27
(PE-10385) Modify acceptance tests to run under PE 

dankreek:
A few modifications were needed to allow running theses acceptance tests from a PE installation with the pe-puppetserver package installed.


acceptance/lib/helper.rb
```diff
@@ -167,12 +167,7 @@ def upgrade_package(host, name)
   end
 
   def get_defaults_var(host, varname)
-    if master.is_pe?
```
dankreek: `is_pe?` is actually broken in the current Beaker, so this is a bit of a workaround.

acceptance/suites/tests/00_smoke/validate-excluded-bc-jars.rb
```diff
@@ -1,12 +1,7 @@
 
 test_name "Validate excluded BC jars are not in packaged uberjar."
 
-if master.is_pe?
```
dankreek: Same as above.

acceptance/suites/tests/010-puppetserver-cli/subcommand/gem.rb
```diff
@@ -1,5 +1,4 @@
 require 'test/unit/assertions'
-require 'puppetserver/acceptance/gem_utils'
```
dankreek: This appears to be in the load path already as a helper, and it is manually added as a Beaker helper in the PE acceptance script.

acceptance/scripts/hosts.cfg
```diff
@@ -1,37 +0,0 @@
----
```
camlow325: Just curious why this is going away.  Does nothing use it anymore?

acceptance/lib/helper.rb
camlow325: Is there an expectation that this will get fixed at some point?  Wondering if we should create a follow-on ticket for our backlog to move back to that later on if so?

camlow325:
:see_no_evil: @waynr or @erikPrime - this look okay to you?

acceptance/scripts/hosts.cfg
dankreek: It gets overwritten every time the `testrun.sh` script is executed.

waynr: Yeah this was only ever meant to be a transient `hosts.cfg` anyway, looks like an errant `git add` included it. I recommend adding it to `.gitignore`.

acceptance/lib/helper.rb
nwolfe: This actually seems like an improvement over the if-else checking everywhere

acceptance/config/beaker/options.rb
```diff
@@ -6,6 +6,7 @@
   "service-prefix" => 'service ',
   "service-num-retries" => 1500,
   "puppetservice" => 'puppetserver',
```
nwolfe: I take it that during PE testing the PE version of this `options.rb` will be used, right?  Otherwise the service name wouldn't match (`pe-puppetserver`)

acceptance/suites/tests/010-puppetserver-cli/subcommand/foreground.rb
```diff
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
```
nwolfe: Oops.

Oh well

acceptance/suites/tests/010-puppetserver-cli/subcommand/gem.rb
nwolfe: Pretty sure we don't need the `test/unit/assertions` one either

acceptance/suites/tests/010-puppetserver-cli/subcommand/foreground.rb
dankreek: EMACS!!!!

waynr: Also, if for any reason we had to pass this file through a Python interpreter this would let the interpreter know to read it in with "utf-8" encoding.

acceptance/config/beaker/options.rb
dankreek: That's correct

acceptance/lib/helper.rb
camlow325: I'm actually okay with this and agree that this definitely feels better than proliferating conditional logic across tests - although presumably we could have also done this by just putting the conditional check in a helper function that would be called from the different tests instead of defining a new `option`.

I wasn't sure if `is_pe?` is some sort of "standard" recommended for this kind of differentiation in the Beaker world.  In any event, I'm happy with what is on the PR for this right now and am not suggesting we change it in the future.

dankreek:
@nwolfe everything's addressed on here if you want to mash that button.

nwolfe:
:+1:

PR #648, opened 2015-07-27
(SERVER-808) Enable containerized test running

dankreek:
The `sudo: false` flag indicated to Travis-CI that the tests should be
run in a container instead of VM. This should hopefully allow for faster
test results.


camlow325:
:+1: but might be better to do a PR for this against the stable branch first and let that then be merged up to the master branch.

dankreek:
Ok, I reopened this as https://github.com/puppetlabs/puppet-server/pull/650

PR #649, opened 2015-07-28
Merge stable into master

KevinCorcoran:
Merge conflicts:
* a couple places where we're testing against the Puppet or puppet-agent version number (obviously, different across branches)
* `puppetlabs.puppetserver.error-handling-int-test/test-HTTP-500` had to be updated to no longer refer to the `handle-request` function which was deleted in #645.  This test does not exist on stable.


camlow325:
:+1:

PR #650, opened 2015-07-28
(SERVER-808) Enable containerized test running

dankreek:
The `sudo: false` flag indicated to Travis-CI that the tests should be
run in a container instead of VM. This should hopefully allow for faster
test results.


camlow325:
:+1: :shipit:

PR #651, opened 2015-07-30
(SERVER-815) Add getClassInfo methods to JRubyPuppet interface

camlow325:
This commit adds two new methods to the JRubyPuppet interface -
`getClassInfoForEnvironment` and `getClassInfoForAllEnvironments` - to
the JRubyPuppet interface.  A corresponding implementation of those in
the `Puppet::Server::Master` Ruby class uses the
`Puppet::InfoService::ClassInformationService` to get class information
for available manifests in environments.

### DIFF
--- a/src/ruby/puppet-server-lib/puppet/server/http_client.rb
+++ b/src/ruby/puppet-server-lib/puppet/server/http_client.rb
@@ -87,8 +87,8 @@ class Puppet::Server::HttpClient
   end
 
   def self.terminate
-    unless self.client.nil?
-      self.client.close
+    unless @client.nil?
+      @client.close
     end
   end
 
--- a/src/ruby/puppet-server-lib/puppet/server/master.rb
+++ b/src/ruby/puppet-server-lib/puppet/server/master.rb
@@ -1,5 +1,7 @@
 require 'puppet/server'
 
+require 'puppet/info_service'
+
 require 'puppet/network/http'
 require 'puppet/network/http/api/master/v3'
 
@@ -32,6 +34,7 @@ class Puppet::Server::Master
                           any.
                           chain(Puppet::Network::HTTP::API::Master::V3.routes)
     register([master_routes])
+    @env_loader = Puppet.lookup(:environments)
   end
 
   def handleRequest(request)
@@ -57,6 +60,53 @@ class Puppet::Server::Master
         response["X-Puppet-Version"])
   end
 
+  def getClassInfoForAllEnvironments()
+    # The clear_environment_settings(env) ensures that the calls which happen
+    # later on to enumerate all of the manifests for each environment are
+    # using the latest environment.conf settings for each environment.  Without
+    # this call, cached (and, therefore, possibly stale) environment.conf
+    # settings would be used.  We want to return the latest data for any calls
+    # made to this method.
+    @env_loader.list.each do |env|
+      Puppet.settings.clear_environment_settings(env)
+    end
+
+    environments =
+      Hash[@env_loader.list.collect do |env|
+       [env.name, self.class.getManifests(env)]
+      end]
+
+    classes_per_env =
+        Puppet::InfoService::ClassInformationService.new.classes_per_environment(environments)
+    Hash[classes_per_env.collect {|key, value| [key.to_s, value]}]
+  end
+
+  def getClassInfoForEnvironment(env)
+    # The clear_environment_settings(env) ensures that the calls which happen
+    # later on to enumerate all of the manifests for each environment are
+    # using the latest environment.conf settings for each environment.  Without
+    # this call, cached (and, therefore, possibly stale) environment.conf
+    # settings would be used.  We want to return the latest data for any calls
+    # made to this method.
+    Puppet.settings.clear_environment_settings(env)
+
+    # It would be more direct and less expensive to call `@env_loader.get(env)`
+    # here.  The problem with doing that, though, is that a Cached `env_loader`
+    # could return an Environment object with cached settings.  The `list` call
+    # to the Cached loader results in new Environment objects being created
+    # and so would not be subject to any potentially stale settings data being
+    # used to enumerate manifests.
+    environment = @env_loader.list.find do |env_from_loader|
+      env_from_loader.name == env
+    end
+    unless environment.nil?
+      environments = Hash[env, self.class.getManifests(environment)]
+      classes_per_env =
+          Puppet::InfoService::ClassInformationService.new.classes_per_environment(environments)
+      classes_per_env[env]
+    end
+  end
+
   def getSetting(setting)
     Puppet[setting]
   end
@@ -73,4 +123,19 @@ class Puppet::Server::Master
     Puppet::Server::Config.terminate_puppet_server
   end
 
+  private
+
+  def self.getManifests(env)
+    manifests = []
+    if env.manifest != Puppet::Node::Environment::NO_MANIFEST
+      if File.directory?(env.manifest)
+        manifests = Dir.glob(File.join(env.manifest, '**/*.pp')).sort
+      else
+        manifests = [env.manifest]
+      end
+    end
+
+    module_manifests = env.modules.collect {|mod| mod.all_manifests}
+    manifests.concat(module_manifests).flatten.uniq
+  end
 end
--- a/test/integration/puppetlabs/services/jruby/class_info_test.clj
+++ b/test/integration/puppetlabs/services/jruby/class_info_test.clj
@@ -1,0 +1,229 @@
+(ns puppetlabs.services.jruby.class-info-test
+  (:require [clojure.test :refer :all]
+            [puppetlabs.kitchensink.core :as ks]
+            [puppetlabs.services.jruby.jruby-testutils :as jruby-testutils]
+            [puppetlabs.services.jruby.jruby-puppet-internal :as jruby-internal]
+            [me.raynes.fs :as fs]
+            [cheshire.core :as cheshire])
+  (:import (java.util.concurrent LinkedBlockingDeque)))
+
+(deftest ^:integration class-info-test
+  (testing "class info properly enumerated for"
+    (let [pool (LinkedBlockingDeque. 1)
+          code-dir (ks/temp-dir)
+          conf-dir (ks/temp-dir)
+          config (jruby-testutils/jruby-puppet-config
+                   {:master-code-dir (.getAbsolutePath code-dir)
+                    :master-conf-dir (.getAbsolutePath conf-dir)})
+          instance (jruby-internal/create-pool-instance!
+                     pool 0 config #() nil)
+          jruby-puppet (:jruby-puppet instance)
+          container (:scripting-container instance)
+
+          env-dir (fn [env-name]
+                    (fs/file code-dir "environments" env-name))
+          create-file (fn [file content]
+                        (ks/mkdirs! (fs/parent file))
+                        (spit file content))
+          gen-classes (fn [[mod-dir manifests]]
+                        (let [manifest-dir (fs/file mod-dir "manifests")]
+                          (ks/mkdirs! manifest-dir)
+                          (doseq [manifest manifests]
+                            (spit (fs/file manifest-dir (str manifest ".pp"))
+                                  (str
+                                    "class " manifest "($" manifest "_a, Integer $"
+                                    manifest "_b, String $" manifest
+                                    "_c = 'c default value') { }\n"
+                                    "class " manifest "2($" manifest "2_a, Integer $"
+                                    manifest "2_b, String $" manifest
+                                    "2_c = 'c default value') { }\n")))))
+          create-env-conf (fn [env-dir content]
+                            (create-file (fs/file env-dir "environment.conf")
+                                         (str "environment_timeout = unlimited\n"
+                                              content)))
+          create-env (fn [[env-dir manifests]]
+                       (create-env-conf env-dir "")
+                       (gen-classes [env-dir manifests]))
+
+          _ (create-file (fs/file conf-dir "puppet.conf")
+                         "[main]\nenvironment_timeout=unlimited\nbasemodulepath=$codedir/modules\n")
+
+          env-1-dir (env-dir "env1")
+          env-1-dir-and-manifests [env-1-dir ["foo" "bar"]]
+          _ (create-env env-1-dir-and-manifests)
+
+          env-2-dir (env-dir "env2")
+          env-2-dir-and-manifests [env-2-dir ["baz" "bim" "boom"]]
+          _ (create-env env-2-dir-and-manifests)
+
+          env-1-mod-dir (fs/file env-1-dir "modules")
+          env-1-mod-1-dir-and-manifests [(fs/file env-1-mod-dir
+                                                  "envmod1")
+                                         ["envmod1baz" "envmod1bim"]]
+          _ (gen-classes env-1-mod-1-dir-and-manifests)
+          env-1-mod-2-dir (fs/file env-1-mod-dir "envmod2")
+          env-1-mod-2-dir-and-manifests [env-1-mod-2-dir
+                                         ["envmod2baz" "envmod2bim"]]
+          _ (gen-classes env-1-mod-2-dir-and-manifests)
+
+          env-3-dir-and-manifests [(env-dir "env3") ["dip" "dap" "dup"]]
+
+          base-mod-dir (fs/file code-dir "modules")
+          base-mod-1-and-manifests [(fs/file base-mod-dir "basemod1")
+                                    ["basemod1bap"]]
+          _ (gen-classes base-mod-1-and-manifests)
+
+          bogus-env-dir (env-dir "bogus-env")
+          _ (create-env [bogus-env-dir []])
+          _ (gen-classes [bogus-env-dir ["envbogus"]])
+          _ (gen-classes [(fs/file base-mod-dir "base-bogus") ["base-bogus1"]])
+
+          roundtrip-via-json (fn [obj]
+                               (-> obj
+                                   (cheshire/generate-string)
+                                   (cheshire/parse-string)))
+          expected-class-info (fn [class]
+                                    {"name" class
+                                     "params" [{"name" (str class "_a")}
+                                               {"name" (str class "_b"),
+                                                "type" "Integer"}
+                                               {"name" (str class "_c"),
+                                                "type" "String",
+                                                "default_literal" "c default value"}]})
+
+          expected-manifests-info (fn [manifests]
+                                        (into {}
+                                              (apply concat
+                                                     (for [[dir names] manifests]
+                                                       (do
+                                                         (for [name names]
+                                                           [(.getAbsolutePath
+                                                              (fs/file dir
+                                                                       "manifests"
+                                                                       (str name ".pp")))
+                                                            [(expected-class-info name)
+                                                             (expected-class-info
+                                                               (str name "2"))]]))))))
+          get-class-info (fn []
+                           (-> (.getClassInfoForAllEnvironments
+                                jruby-puppet)
+                              (roundtrip-via-json)))
+          get-class-info-for-env (fn [env]
+                                   (-> (.getClassInfoForEnvironment jruby-puppet
+                                                                    env)
+                                       (roundtrip-via-json)))]
+        (try
+          (testing "initial parse for"
+            (let [expected-envs-info {"env1" (expected-manifests-info
+                                               [env-1-dir-and-manifests
+                                                env-1-mod-1-dir-and-manifests
+                                                env-1-mod-2-dir-and-manifests
+                                                base-mod-1-and-manifests])
+                                      "env2" (expected-manifests-info
+                                               [env-2-dir-and-manifests
+                                                base-mod-1-and-manifests])}]
+              (testing "all environments"
+                (is (= expected-envs-info (get-class-info)))
+              (testing "one environment by name"
+                (is (= (expected-envs-info "env1")
+                       (get-class-info-for-env "env1"))
+                    "Unexpected info retrieved for 'env1'")
+                (is (= (expected-envs-info "env2")
+                       (get-class-info-for-env "env2"))
+                    "Unexpected info retrieved for 'env2'")))))
+
+          (testing "changes to module and manifest paths for"
+            (create-env-conf env-1-dir (str "manifest="
+                                            (.getAbsolutePath (fs/file env-1-dir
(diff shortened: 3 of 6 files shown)


camlow325:
Travis appears to be failing because of a checksum issue on the Maven Central server:

~~~
Could not transfer artifact org.jruby.jcodings:jcodings:pom:1.0.13 from/to central (https://repo1.maven.org/maven2/): Checksum validation failed, expected 7dcf937c53aec4b39481ec27df3437fa201bb951 but is 20262d29aa3b63aa867ef46daea3c7f596fc7aef
~~~

See https://issues.sonatype.org/browse/MVNCENTRAL-782

src/ruby/puppet-server-lib/puppet/server/http_client.rb
```diff
@@ -87,8 +87,8 @@ def get(url, headers)
   end
 
   def self.terminate
-    unless self.client.nil?
-      self.client.close
+    unless @client.nil?
```
camlow325: For a couple of the tests that I'd added for this PR - as well as for the tests in the pre-existing `jruby-puppet-internal-test` namespace - I was hoping to avoid setting up / generating SSL certs and keys in order simply to exercise a JRubyPuppet instance since it is irrelevant to what is being tested.  Before this code accesses `self.client.nil?`, `@client` would be nil.  `@self.client`, however, would end up creating `@client` because it does `@client ||= create_client`.  The `create_client` call ends up creating an SSL context - which would require good pre-existing certs and keys - and creating an Apache HTTP client.  For the tests that don't setup the good SSL certs and keys - for cases where the certs/keys weren't initialized by side-effect from some other test anyway - this code will lead into an `Exception` being thrown.

While we could work around this for tests, it seemed to me that having to do all of this work to create instances of objects that would immediately get thrown away anyway would be wasteful.  This could happen in production code, for example, in cases where no outbound HTTP calls are ever made from the master - in which case the client never gets created.  Granted, those cases in production may not be all that common anyway.  I'm open to just leaving this code the way it is and working around it in the tests if this is too concerning to others.

src/ruby/puppet-server-lib/puppet/server/master.rb
```diff
@@ -57,6 +60,53 @@ def handleRequest(request)
         response["X-Puppet-Version"])
   end
 
+  def getClassInfoForAllEnvironments()
```
camlow325: I'm pretty mixed at this point about the work that this PR is doing to try to work around the fact that environment objects and settings can get cached.  My hope is that when this code gets hooked up to other Clojure-level code in Puppet Server later on that that code will always be marking all of the environments associated with a JRubyPuppet instance as "expired".  Granted, we have a separate defect, SERVER-813, where the expiration isn't applied across all instances in some cases.  That defect will need to be fixed at some point, though.  If we assume that the expiration is done before calls to either of the "getClassInfo" methods added by this PR are made, I don't think any of the cache avoidance work would be necessary - including the `clear_environment_settings` call or the `@env_loader.list.find()` call made in `getClassInfoForEnvironment`.  Not knowing how the rest of the dependencies are going to fall out at this point, I went with including the cache avoidance logic for now.  Hopefully / maybe we could rip this out later, though?

test/integration/puppetlabs/services/jruby/class_info_test.clj
```diff
@@ -0,0 +1,229 @@
...
+            [puppetlabs.kitchensink.core :as ks]
+            [puppetlabs.services.jruby.jruby-testutils :as jruby-testutils]
+            [puppetlabs.services.jruby.jruby-puppet-internal :as jruby-internal]
+            [me.raynes.fs :as fs]
+            [cheshire.core :as cheshire])
+  (:import (java.util.concurrent LinkedBlockingDeque)))
+
+(deftest ^:integration class-info-test
+  (testing "class info properly enumerated for"
+    (let [pool (LinkedBlockingDeque. 1)
```
camlow325: This `let` block is obviously pretty massive.  Even though they don't get used outside of this `deftest` block, I could move some of the "helper" type functions out to separate `defn`s just to shorten this up.

The `testing` sub-blocks are intentionally ordered the way they are to validate that the nature of what class information is extracted from parsing is updated appropriately as the underlying content on disk changes from one call to the next.  By design, then, the `testing` sub-blocks aren't intended to be run independently.

test/integration/puppetlabs/services/jruby/class_info_test.clj
```diff
@@ -0,0 +1,229 @@
...
+                (is (= expected-envs-info (get-class-info))))
+              (testing "one environment by name"
+                (is (= (expected-envs-info "env2")
+                       (get-class-info-for-env "env2"))
+                    "Unexpected info retrieved for 'env2'")
+                (is (= (expected-envs-info "env3")
+                       (get-class-info-for-env "env3"))
+                    "Unexpected info retrieved for 'env3'"))))
+
+          (testing "all environments by key, without serialization"
```
camlow325: I mentioned above that the `testing` sub-blocks in this `deftest` are intentionally ordered, not meant to be run in isolation.  The final two blocks here, though, I think could be separated out into their own `deftest`, though - assuming that some of the setup from this `deftest` were repeated.  I didn't do that initially because that would involve initializing even more JRubyPuppet instances for the additional tests.  I was thinking about the performance overhead that comes with doing that and the slow performance that we're already seeing with this suite overall.

src/ruby/puppet-server-lib/puppet/server/master.rb
```diff
@@ -57,6 +60,53 @@ def handleRequest(request)
...
+    end
+
+    environments =
+      Hash[@env_loader.list.collect do |env|
+       [env.name, self.class.getManifests(env)]
+      end]
+
+    classes_per_env =
+        Puppet::InfoService::ClassInformationService.new.classes_per_environment(environments)
+    Hash[classes_per_env.collect {|key, value| [key.to_s, value]}]
```
camlow325: @dankreek - Here's that logic I mentioned to you where I decoded the environment Symbol that appears as the key for the outermost Hash back into a String before returning back across the Java interop layer.  I did this here because I was thinking that the consumers of this in Clojure would likely need to reference elements from the HashMap by environment key but likely would treat anything in the value per environment as a more opaque blob - handed off to cheshire or some other JSON serializer as needed.  This could help keep the Clojure-level handling of the return structure a little more generic - i.e., not having to know that the environment was a JRuby `RubySymbol` before handling it.   The "all environments by key, without serialization" integration test in "class_info_test.clj" which is associated with this validates that a Clojure consumer would only need to do a simple `get` call to access the class info for a specific environment.  Still open to doing this differently.

dankreek: Yeah, that all sounds good. I don't think we need to convert the keys to Clojure keywords since environments will be asked for by string anyway. And cheshire will just `.toString` any Ruby keyword objects it finds.

src/ruby/puppet-server-lib/puppet/server/master.rb
camlow325: One other thing I had forgotten, though...

When getting `Environment` instances back from the `@env_loader.list`, the cache expiration logic is not implicitly executed.  That only happens when `@env_loader.get(env)` is called.  So I think the `Puppet.settings.clear_environment_settings()` calls are safer to make regardless of whether the environments in the cache had been marked for expiration just before this code is invoked.

ruby/puppet
```diff
@@ -1 +1 @@
-Subproject commit d93f59152eeb30c292b5e134d39967e03e41ff61
+Subproject commit f6f511d4046b05566e5495defa59832fe58a1fa4
```
dankreek: being as we're relying on some new functionality in Puppet, I'm assuming this was intentional?

camlow325: Yep, this is the first commit in the core Ruby Puppet repo where the underlying `ClassInformationService` that this PR uses came in.  https://github.com/puppetlabs/puppet/commit/f6f511d4046b05566e5495defa59832fe58a1fa4.

src/ruby/puppet-server-lib/puppet/server/http_client.rb
dankreek: Oh wow, good catch @camlow325

camlow325:
Had a merge conflict on the ruby/puppet submodule update after the PR came in.  Rebased and pushed 6c73930.

src/ruby/puppet-server-lib/puppet/server/master.rb
dankreek: This all sounds find to me.

test/integration/puppetlabs/services/jruby/class_info_test.clj
dankreek: I couldn't agree more. Our test suite already overstays its welcome and I'd say these are still logically a member of the `class-info-test`

dankreek:
:+1: This all LGTM. Thanks for the annotations, @camlow325

test/integration/puppetlabs/services/jruby/class_info_test.clj
fpringvaldsen: I think it would be worth pullilng the functions out into their own `defn`s outside of the `deftest` since there's a large number of them.

I'm happy to be overruled on this point, though, if others disagree.

fpringvaldsen:
I had one small-ish comment, other than that I'm :+1:

test/integration/puppetlabs/services/jruby/class_info_test.clj
```diff
@@ -0,0 +1,229 @@
...
+            (fs/delete-dir env-1-dir)
+            (let [_ (create-env env-3-dir-and-manifests)
+                  expected-envs-info {"env2" (expected-manifests-info
+                                               [env-2-dir-and-manifests])
+                                      "env3" (expected-manifests-info
+                                               [env-3-dir-and-manifests
+                                                base-mod-1-and-manifests])}]
+              (testing "all environments"
+                (is (= expected-envs-info (get-class-info))))
+              (testing "one environment by name"
```
fpringvaldsen: Would it be worth trying to get the class info for `"env1"` to ensure that it won't be returned since the environment has been deleted?

camlow325: Yeah, good call.  I'll add that.

test/integration/puppetlabs/services/jruby/class_info_test.clj
camlow325: I'd be okay with doing that.  @dankreek - what do you think?

dankreek: Yeah, it would increase the readability quite a bit. I'm in favor of it.

camlow325:
Added one more commit, d14b2ca, to address prior comments - splitting out a few of the functions from the test block and adding one more test assertion that no class info is returned for a just-deleted environment.

test/integration/puppetlabs/services/jruby/class_info_test.clj
```diff
@@ -0,0 +1,242 @@
...
+          base-mod-1-and-manifests [(fs/file base-mod-dir "basemod1")
+                                    ["basemod1bap"]]
+          _ (gen-classes base-mod-1-and-manifests)
+
+          bogus-env-dir (env-dir "bogus-env")
+          _ (create-env [bogus-env-dir []])
+          _ (gen-classes [bogus-env-dir ["envbogus"]])
+          _ (gen-classes [(fs/file base-mod-dir "base-bogus") ["base-bogus1"]])
+
+          get-class-info (fn []
```
camlow325: I left these two functions in the let block rather than separating them because they close over `jruby-puppet` and didn't want to have to instead repeat `jruby-puppet` as an argument down where they are used inside of the testing sub-blocks.

camlow325:
Added one more commit, 25a206b, which bumps up the puppet-agent and ruby/puppet submodules to a version that has recently been seen to go through the AIO agent CI pipeline successfully - https://jenkins.puppetlabs.com/view/All%20in%20One%20Agent/view/Master/view/Puppet/job/platform_aio-puppet_pkg-promote_master/108/.

fpringvaldsen:
I had one nitpick (feel free to ignore it), other than that I'm :+1: assuming Travis passes.

camlow325:
Thanks, @fpringvaldsen.  I fixed up the alignment issue in 1df8927.

cprice404:
@camlow325 based on recent branching conversations, should we hold off on merging this until the third branch is created?  Or do you expect we'd create that branch off of an older commit?

src/ruby/puppet-server-lib/puppet/server/master.rb
```diff
@@ -57,6 +60,53 @@ def handleRequest(request)
...
+    # settings would be used.  We want to return the latest data for any calls
+    # made to this method.
+    @env_loader.list.each do |env|
+      Puppet.settings.clear_environment_settings(env)
+    end
+
+    environments =
+      Hash[@env_loader.list.collect do |env|
+       [env.name, self.class.getManifests(env)]
+      end]
```
cprice404: @camlow325 If the goal of the two previous stanzas is just to get a list of strings of environment names, I think it would be much better to push this logic into Puppet itself.  @hlindberg  how much work would it be for you to give us an API for this?  I get really worried about putting any logic that relies on the internal implementation details of Puppet into the Puppet Server code.

src/ruby/puppet-server-lib/puppet/server/master.rb
```diff
@@ -57,6 +60,53 @@ def handleRequest(request)
...
+
+    # It would be more direct and less expensive to call `@env_loader.get(env)`
+    # here.  The problem with doing that, though, is that a Cached `env_loader`
+    # could return an Environment object with cached settings.  The `list` call
+    # to the Cached loader results in new Environment objects being created
+    # and so would not be subject to any potentially stale settings data being
+    # used to enumerate manifests.
+    environment = @env_loader.list.find do |env_from_loader|
+      env_from_loader.name == env
+    end
```
cprice404: Same here.  @hlindberg could we get some API for this?

src/ruby/puppet-server-lib/puppet/server/master.rb
camlow325: @cprice404 It's not just to get the "names" of the environments.  It also iterates through the available {{Puppet::Node::Environment}} objects to get all of the manifest files that need to be parsed for each environment, building the result into a Hash that is passed into the new {{Puppet::InfoService::ClassInformationService}} to parse out the class info from all of those manifests.  And a key piece of that is ensuring that the list which is returned contains "latest data" - i.e., would include all of the environments currently on disk and also use the {{environment.conf}} settings as currently specified on disk.

By "new API" what are you thinking of as being the scope?  Would that API essentially just allow the code in Puppet Server here to make some sort of {{getManifests}} call down into core Ruby Puppet code, returning the Hash of environment => manifest files from "latest data" -- essentially having all of the work to enumerate manifests from the environments be done behind that new {{getManifests}} API?  Or were you just thinking of some sort of construct that would allow for a list of {{Puppet::Node::Environment}} objects compromising "latest data" that can be mined out like is done here behind the `self.class.getManifests` call?

Or - going even further - were you thinking of just thinning this logic out almost entirely to where all this does is a passthrough to core Ruby Puppet code that does the manifest enumeration and parsing via the `ClassInformationService` all in one step?

src/ruby/puppet-server-lib/puppet/server/master.rb
camlow325: I alluded to this in a comment earlier, but I think we could get replace the logic here with an `@env_loader.get(env)` call so long as we ensure that the cache expiration registry's value for the environment's `isExpired` would return true when called.  So, assuming that right before this call is made that Puppet Server were to first `mark-all-environments-expired!`, the `Puppet::Node::Environment` that we get back should have been constructed from "latest data".

camlow325:
> @camlow325 based on recent branching conversations, should we hold off on merging this until the third branch is created? Or do you expect we'd create that branch off of an older commit?

I had been thinking we would do one of those - depending upon the timing of when this PR were to come in.  Based on the questions you have above about the current content on the PR and request for new APIs from @hlindberg, sounds like I should just close this out for now, wait until you and he have agreed on what the new API is, and reopen something later that uses that new API?  In the meantime, I imagine we will have already done the branch reorientation work.

src/ruby/puppet-server-lib/puppet/server/master.rb
cprice404: I don't really care about the implementation details other than the following:

* I am terrified of the idea of you calling {{Puppet.settings.clear_environment_settings(env)}}, and to a lesser degree, even {{Puppet.lookup(environments)}}, from our code.  I would much prefer that we tell @hlindberg exactly what data we need to ask them for and he provides a method for us to call to get it, so that he is free to change implementation details w/o us needing to worry about it.  IMO we should not be mucking directly with anything related to Puppet settings, we should let that happen on the Puppet side of the API.
* In the grand and ideal world of the future, we would be able to parse the environment configs directly from Clojure and do the mapping from "list of changed files" to "list of affected environments" on our own, rather than having to call into Ruby for that.  So whatever we do now, it'd be nice for it to be encapsulated in some way that made it fairly easy to switch to that model if we're ever able to do so.

Beyond those constraints it doesn't matter much to me what the Ruby API ends up looking like as long as both you and @hlindberg are in agreement on it, so feel free to work it out with his team however you think it would be best.

src/ruby/puppet-server-lib/puppet/server/master.rb
cprice404: Sure, but is there a compelling reason that our code should need to be that aware of that detail of the caching?  As opposed to just saying "we need an API that gives us the latest config for all environments", and calling that?

camlow325: No, if we had such an API then I agree this would be unnecessary here.  I wasn't arguing that point.

src/ruby/puppet-server-lib/puppet/server/master.rb
camlow325: Ok, I agree with you, @cprice404, that having an API into core Ruby Puppet that Puppet Server would call into would greatly simplify and make the logic more maintainable here.

So, @hlindberg, what I would envision is that two new helper methods could be created in core Ruby Puppet code that Puppet Server-level code could call into.  Something like what you did for the `ClassInformationService.classes_per_environment` method would work fine.  The methods would be something like:

* environment_manifests() 

This would require no arguments and when called would return back a `Hash` with keys being the string name of each environment and the value being an `Array` of fully-qualified file names of the manifest files that can be referenced by that environment.

* environment_manifests_for_environment(env_name)

The `env_name` would be a string representing the name of the environment for which manifests files should be found.  The return payload from this call would be an `Array` of fully-qualified names of manifest files that can be referenced by that environment.

--

For both methods above, the list of manifest files returned for an environment should include any manifests that the environment could reference directly - via the directory or file that the `manifest` setting in the environment's `environment.conf` file would either implicitly or explicitly be set to.  The list should also include any manifests that the environment could reference via any of the modules its the `modulepath`.

Also, for both methods above, no "cached" data should be used to compute the result.  No cached values of settings sourced from each environment's `environment.conf` file should be used, e.g., for computing module and manifest paths.  Also, none of the manifests that the module objects return should be from previously cached data.  (I understand that the setting values coming directly from `puppet.conf` will be cached, but I think that's an artifact of Puppet Server's current reloading of global configuration that will need to be looked at outside of the effort for the "NC Class Refresh" epic.) Essentially, we want the result to include the full list of manifest files for each environment as currently defined on disk at the time the method call is made.

In terms of the format of the response data, I'm assuming it should be sufficient for our Puppet Server-level code to just turn right back around and hand the data into the `ClassInformationService.classes_per_environment` method to get parsed class info for each of the manifest files.

--

@hlindberg - does the above make sense?  I can file a PUP ticket with this request and assign to your team for scoping if that makes sense as a next step.

camlow325: So, @cprice404, assuming that plumbing the new API up from core Ruby Puppet and integrating with Puppet Server will take some time, what do you suggest we do with this PR in the meantime?  Commit it as is, with a separate ticket to replace this work with the new API as it becomes available later, or just shelve it and come back to it later.  I suppose over the next few days that we may mostly be consumed with work outside of the "NC Class Refresh" effort, but this will quickly become a blocker for making further progress on that epic as we do have cycles to put toward it.

cprice404: @camlow325 can you please paste that last big comment with your requirements / goals into a new PUP ticket and assign to @hlindberg ?  And maybe create a placeholder ticket for SERVER to link to it?  After that I'm fine w/you merging this (pending review/consensus by whomever you would like to have review it, I haven't really reviewed it thoroughly and probably won't unless you ask me to) and moving on as long as we have the other ticket flagged as a release blocker.

Also FYI I believe that @hlindberg is on PTO for another week or so.  You could check w/Kylo on that.

camlow325: @cprice404 Ok, thanks.  I created PUP-5011 to capture the request for new manifest enumeration methods in core Ruby Puppet and SERVER-834 to cover the follow-on rework in Puppet Server to use those new methods whenever they become available.

camlow325:
@fpringvaldsen, @dankreek - so I think this is basically ready to merge now.  I did create a couple of follow-on tickets to capture the concerns that @cprice404 had expressed about pushing some of the functionality down into core Ruby Puppet

As far as the Travis CI failures that we had been seeing earlier on, I haven't had the time to fully root-cause them, but it seems to me that they are due to a memory leak which is pervasive across at least the tests which do a full Trapperkeeper service bootstrap.  In a later commit for the PR, I bumped the JVM heap size up from 1 Gb to 2 Gb and the tests started passing.  I ran just the tests that were added for this PR in a local loop about 50 times or so locally with no failures, so I'm not sure that the new tests were necessarily exacerbating the memory problems that were already in the suite.  In any event, I added some comments about this to SERVER-797, the "tests are too slow" ticket, in the event that someone is able to get around to looking at that soon.

It may make sense to hold off on merging this until @nwolfe is done with the branch reorbiting that I think he'll be doing today.  Any additional issues anyone sees with this PR?

camlow325:
@fpringvaldsen, @dankreek - In 89e4ed, I bumped the Puppet submodule that this references up to a newer version, one that we need in order to get past some of the Puppet acceptance test failures seen related to SERVER-832.

Since the branch reorientation work on Puppet Server is done for now, then, assuming this is still passing w/ Travis, I think this is ready to be merged now.

PR #652, opened 2015-07-30
(MAINT) Add require statements to main helper

dankreek:
This will allow us to cleanup the Beaker command line a bit

### DIFF
--- a/acceptance/suites/tests/010-puppetserver-cli/subcommand/gem.rb
+++ b/acceptance/suites/tests/010-puppetserver-cli/subcommand/gem.rb
@@ -1,3 +1,6 @@
+require 'test/unit/assertions'
+require 'puppetserver/acceptance/gem_utils'
+
 test_name "Puppetserver 'gem' subcommand tests."
 
 cli = "puppetserver"


acceptance/lib/helper.rb
```diff
@@ -1,4 +1,6 @@
 require 'beaker/dsl/install_utils'
+require 'puppetserver/acceptance/compat_utils'
+require 'puppetserver/acceptance/gem_utils'
```
camlow325: My preference would be to just put the `gem_utils` require statement back into the `.../subcommand/gem.rb` since that's what uses the one function in `gem_utils.rb`.  And the `compat_utils` one probably isn't strictly needed since the individual `puppet3_tests` that need it all require it on their own, right?

nwolfe: +1 for putting the require's everywhere they're actually needed rather than always just in the `helper.rb`

dankreek: well, this bring up the original problem. The relative path to the helper is different depending upon how you're running the tests. (from pe-puppet-server-extensions or from puppet-server itself )

dankreek: So I can put it back on the command line or leave it here.

nwolfe: I think we need to add another directory to the load path; shouldn't have to mess with the helper file at all.

dankreek: The first person to chime in wins.

camlow325: > I think we need to add another directory to the load path; shouldn't have to mess with the helper file at all.

:+1:

dankreek:
Okay, put everything back as it was. We'll just need to modify the load path in the pe-puppet-server-extensions bit.

camlow325:
:+1:

camlow325:
Travis failed due to a known problem with the Maven Central server - a checksum failure when downloading org.jruby.jcodings:jcodings:pom:1.0.13.  I don't see anything in this PR which would change the outcome of a Travis run since these changes are in acceptance tests only, which don't run in Travis.

Merging...

PR #655, opened 2015-08-04
(PE-11124) Define env cache structure

dankreek:


src/clj/puppetlabs/puppetserver/environment_cache.clj
```diff
@@ -0,0 +1,66 @@
...
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Schemas
+
+(def EnvironmentKey
+  "Environments are keyed by their names as strings."
+  schema/Str)
+
+(def EnvironmentClassInfo
+  "A data structure describing one or more envrionments, keyed by
```
camlow325: s/envrionments/environments

test/unit/puppetlabs/puppetserver/environment_cache_test.clj
```diff
@@ -0,0 +1,79 @@
...
+        "/Users/batmanandrobin/.puppet-server/2.x/code/environments/production/modules/simmons/manifests/source_file.pp"
+          [{"name" "simmons::source_file"
+            "params" [{"name" "studio"}]}]}}))
+
+(deftest environment-cache
+  (testing "Create a new empty cache"
+    (is (= {} (create-cache))))
+
+  (testing "Initializing and clearing a cache"
+    (let [cache (atom (create-cache))
```
camlow325: Do we need to use an atom in these tests?

dankreek: Not specifically, no. But I kind of wanted to illustrate intended usage as tests tend to be useful as developer documentation.

camlow325:
I think another operation we might eventually need is some sort of `remove-environment-info`.  For example, I could see a call being made to `get-environment(env)` for an environment that may exist in the cache but that no longer exists on disk on the master.  In that case, I think we'd want to end up removing the environment from the cache.

src/clj/puppetlabs/puppetserver/environment_cache.clj
```diff
@@ -0,0 +1,66 @@
...
+
+(schema/defn ^:always-validate
+  get-all-environments-info :- EnvironmentClassInfo
+  "Return the info for the environments in the cache. If the cache is empty then
+  an empty map is returned. "
+  [cache :- EnvironmentClassInfoCache]
+  cache)
+
+(schema/defn ^:always-validate
+  get-environment-info :- (schema/maybe EnvironmentClassInfo)
```
camlow325: This probably works fine as is, but I might go with a design that just returns back the `info` payload, the `{schema/Any schema/Any}` as opposed to wrapping it back up with its `env` key since the caller already knows what that key is.

dankreek:
> I think another operation we might eventually need is some sort of remove-environment-info.

This has been added.

src/clj/puppetlabs/puppetserver/environment_cache.clj
dankreek: Sounds good, this has been addressed.

camlow325:
:+1:

Sorry, just realized that it would have been better for this work to be done for a SERVER ticket instead of a PE one.  Would you mind moving the JIRA ticket to the SERVER project and revising the commit message to have the SERVER ticket in the description.

Also, work on SERVER-828 (branch renaming) will be starting soon.  This is work that we'll only want to have in be in the "new" master branch and not the "new" stable branch.  @nwolfe - given that, do you think would it be better to hold off on letting this be merged until you have SERVER-828 done?

src/clj/puppetlabs/puppetserver/environment_cache.clj
```diff
@@ -0,0 +1,78 @@
...
+  [cache :- EnvironmentClassInfoCache
+   env :- EnvironmentKey]
+  (dissoc cache env))
+
+(schema/defn ^:always-validate
+  get-all-environments-info :- EnvironmentClassInfo
+  "Return the info for the environments in the cache. If the cache is empty then
+  an empty map is returned. "
+  [cache :- EnvironmentClassInfoCache]
+  cache)
```
nwolfe: Man...this function does nothing?

KevinCorcoran: global warming

nwolfe:
So many whats

src/clj/puppetlabs/puppetserver/environment_cache.clj
dankreek: That's correct. What's important here is that there's a function named `get-all-environments-info` that performs the documented functionality. The internal details of these functions will likely change over time, right now this is the "just get something working" implementation.

dankreek:
> So many whats

Sorry you're having trouble with this.

nwolfe:
> Sorry you're having trouble with this.

Haha yeah sorry about that...Friday code review and all.  I later realized we're basically defining an API (schemas+functions) with an initial, simple implementation using a map.  The API part is what's more important right now.

:gemini: :+1: :money_with_wings:

dankreek:
> :gemini: :+1: :money_with_wings:

I understand the :+1:  and the :money_with_wings:.. but :gemini: ?

PR #658, opened 2015-08-07
(PE-8226) jruby events

nwolfe:
This PR adds event notifications for whenever something interesting happens related to the JRuby pool. It's a pre-cursor to PE-8226, and replaces my earlier attempt at this that was in PR #632 .

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -75,6 +75,12 @@
   :deploy-repositories [["releases" ~(deploy-info "http://nexus.delivery.puppetlabs.net/content/repositories/releases/")]
                         ["snapshots" ~(deploy-info "http://nexus.delivery.puppetlabs.net/content/repositories/snapshots/")]]
 
+  ;; By declaring a classifier here and a corresponding profile below we'll get an additional jar
+  ;; during `lein jar` that has all the code in the test/ directory. Downstream projects can then
+  ;; depend on this test jar using a :classifier in their :dependencies to reuse the test utility
+  ;; code that we have.
+  :classifiers [["test" :testutils]]
+
   :profiles {:dev {:source-paths  ["dev"]
                    :dependencies  [[org.clojure/tools.namespace "0.2.4"]
                                    [puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty-version]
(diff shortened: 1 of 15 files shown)


cprice404:
From the previous incarnation of this PR, I believe the only outstanding feedback was @fpringvaldsen 's comment about `default-services`, and the suggestion to rename `action` to `reason`.  Anyone have anything else before I address those?

KevinCorcoran:
Looks good overall!  :+1:

cprice404:
I pushed some new commits that I believe handle all of the feedback, with one exception: I started looking into the possibility of moving the `instance-requested` / `instance-borrowed`, etc., calls out of the service namespace and into the core namespace, and the refactor was giving me some pause.  Basically, there are multiple 'borrow' functions in core, and a decent number of tests and a few other things that are consuming the functions in core directly, and it didn't really seem correct to force all of the event-notification-related arguments to be wired through all of that stuff.

The way I'm reasoning about it right now is that the event notification stuff is part of the service contract, more than part of the core.  (e.g. you really only register callbacks through the service protocol, so it seems maybe OK to have the implementation also coupled to the borrow/return functions in the service protocol.)

I could be convinced that it's the right thing to do anyway, and I don't mind doing it if that's consensus (I generally agree that keeping the service namespace as thin as possible is the way to go), but it gave me enough pause that I decided not to do it right this minute.

Since I know that this PR may be blocking other work, maybe we could file a follow-up ticket about that change if it seems important enough?

KevinCorcoran:
My opinion is that it is still worth the extra effort to move the calls to `instance-requested` et. al. into the core namespace.  I think of the core namespace as being simply the underlying implementation of the service namespace, so it makes sense to me that these calls would be pushed into core.

I understand the pain of dealing with the tests, but I don't think that's a big enough issue to force our hand here.  Perhaps the refactor can be done in a way that doesn't affect the tests so drastically - like, simply rename `core/borrow-from-pool-with-timeout`  to `borrow-from-pool-with-timeout*` and add a new function which can be called from the service namespace to do the event handling and the borrow.  That way, the code can be moved into the service namespace but the tests won't need any updating.

I don't think this is important enough to file a separate ticket (which would inevitably just fall into the JIRA black hole), but it does seem important enough to just take care of now, especially since I don't think this PR will be blocking anything for a couple more days.  I don't think it's a 'must-have', but it would be nice to get it right, if we have the time.

KevinCorcoran:
I'm :+1: on this PR, pending resolution of the issue about moving some calls into the `core` namespace.  @fpringvaldsen - I will be out for a day or so - you can feel free to merge this tomorrow if you and @cprice404 agree that it's ready to be merged.

fpringvaldsen:
This looks good to me, I'm :+1: on the changes that have been introduced since I last gave this a review.

I agree with @KevinCorcoran that it might be a good idea to move those function calls into the `core` namespace. 

Also, since this is going into Open Source Puppet Server, do we want a PE ticket associated with it, or should we open a SERVER ticket and associate the commit messages with that instead?

Pending the resolution of those two issues I'm good with this being merged.

cprice404:
Closing this PR because it needed a rebase to change all of the commit messages to refer to the SERVER ticket, and I don't want to blow away the comment threads.

PR #660, opened 2015-08-07
(SERVER-828) Maximum supported puppet-agent <1.3.0

nwolfe:
This commit caps the puppet-agent dependency to < 1.3.0 (i.e. 1.2.z)

### DIFF
--- a/resources/ext/ezbake.conf
+++ b/resources/ext/ezbake.conf
@@ -7,7 +7,7 @@
 ezbake: {
    pe: {}
    foss: {
-      redhat: { dependencies: ["puppet-agent"],
+      redhat: { dependencies: ["puppet-agent < 1.3.0"],
                # This is terrible, but we need write access to puppet's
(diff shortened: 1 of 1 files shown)


nwolfe:
Ping @camlow325 for review

camlow325:
:+1:

resources/ext/ezbake.conf
```diff
@@ -34,7 +34,7 @@ ezbake: {
                ]
              }
 
-      debian: { dependencies: ["puppet-agent"],
+      debian: { dependencies: ["puppet-agent < 1.3.0"],
```
haus: The deprecated forms < and > were confusingly used to mean earlier/later or equal, rather than strictly earlier/later, and must not appear in new packages (though dpkg still supports them with a warning).

I think you want some `<<` here instead.

nwolfe:
Updated with `<<`

nwolfe:
We decided it's safer not to pin the puppet-agent package for the time being

PR #661, opened 2015-08-10
(SERVER-832) Bump to new puppet version to deal with acceptance failures

rlinehan:


camlow325:
:+1: Before we merge this, though, should we have a corresponding ci-job-configs PR which bumps the puppet-agent version up?  Usually we try to have the puppet-server and ci-job-configs go in around the same time.  The ci-job-configs puppet-agent version will trump the one in puppet-server anyway for the jobs that get run through Jenkins in CI.

rlinehan:
Sure, I can do that. Just curious, what's the point of having the puppet-agent version in both places? That just seems really non-obvious, confusing, and redundant, but as I haven't worked on the puppet-server acceptance tests in a while, I'm probably missing something.

camlow325:
> Just curious, what's the point of having the puppet-agent version in both places? That just seems really non-obvious, confusing, and redundant

It is - and I don't like it either.

My understanding is that we have the one hardcoded in Puppet Server as an aid for those who want to run tests directly via beaker scripts - where JJB parameters don't come into play - and just have the "right" version for the latest stuff that we're testing be picked up.  Since we have that restriction anyway, my preference would be that we just have the one in Puppet Server and only define an override in the JJB configs for special cases where we want to try out a run from a different version number without having to bump the source - like if that would otherwise cause us to burn a tag when it's important that we don't for whatever reason.

@nwolfe and @waynr could probably give some more history on how we got to this point, though.

rlinehan:
Thanks for the explanation @camlow325! I agree, from what you've said it seems like it would make more sense to just have this in Puppet Server, but I'll defer to Nate or Wayne.

In any case, I've filed https://github.com/puppetlabs/ci-job-configs/pull/455.

camlow325:
Looks simple enough, so I'm just going to merge this one.

PR #663, opened 2015-08-11
(MAINT) Prefer clj-time/seconds over clj-time/secs

nwolfe:
The latter is deprecated in favor of the former; we should no longer see
deprecation warnings during `lein test` for this.


src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -163,7 +163,7 @@
   [ca-ttl :- schema/Int]
   (let [now        (time/now)
         not-before (time/minus now (time/days 1))
-        not-after  (time/plus now (time/secs ca-ttl))]
```
dankreek: ...in a world where `secs` is deprecated...

dankreek:
:+1: LGTM, all tests passed on my machine.

PR #664, opened 2015-08-11
(SERVER-813) env cache flush while borrowed

cprice404:
This commit fixes a bug where, if any JRuby instances were borrowed from the pool when a request came in to flush the environment cache, those instances would not have their cache flushed.

The implementation of the cache flush was simply grabbing a reference to all of the instances *currently* in the pool, and then manipulating some Clojure state associated with those instances to notify the Ruby code that the environment cache should be flushed.  It did not take into account any instances that weren't in the pool at the time when this action transpired.

This commit adds a simple subclass of the java.util.concurrent data
structure that we have been using to implement the pool, and a "register" method that can be used to add items to the data structure in a way where the list of all known instances is available for retrieval at any time, regardless of which instances are currently in the pool and which are borrowed.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
@@ -69,9 +69,10 @@
   (jruby-internal/get-pool context))
 
 (schema/defn ^:always-validate
-  pool->vec :- [JRubyPuppetInstance]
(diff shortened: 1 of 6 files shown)


cprice404:
@camlow325 ping

cprice404:
Crap, I didn't account for our branch changes; this should be targeted at 1.x.  Closing, will re-open a new one targeted there.

PR #670, opened 2015-08-11
(maint) Update puppet version to fix acceptance test failures

rlinehan:
Previously, puppet-server stable tests were running against a build of
puppet-agent that was from the per-commit pipeline. This meant that it was not
available on all platforms we run puppet-server stable acceptance tests against.
The first commit in this PR bumps the puppet build version to a puppet-agent package version in
nightlies, which is built for all the platforms.

PUP-4824 had some incorrectly written tests that caused errors for our CI.
These have now been fixed in puppet stable. The second commit updates the submodule
(and therefore the acceptance tests we are running) for these fixes.

The second commit does not update the puppet agent/puppet build version in
`acceptance/lib/helper.rb` since the only changes that have been made since the
package we currently have specified there are test changes.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -29,7 +29,7 @@ module PuppetServerExtensions
     # http://builds.delivery.puppetlabs.net/puppet-agent/
     puppet_build_version = get_option_value(options[:puppet_build_version],
                          nil, "Puppet Development Build Version",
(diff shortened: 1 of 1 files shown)


rlinehan:
Corresponding ci-job-configs PR: https://github.com/puppetlabs/ci-job-configs/pull/458

nwolfe:
:+1:

PR #671, opened 2015-08-12
(maint) Update puppet version for acceptance test fix

rlinehan:
PUP-4824 had some incorrectly written tests that caused errors for our CI.
These have now been fixed in puppet stable. This commit updates the submodule
(and therefore the acceptance tests we are running) for these fixes.

This commit does not update the puppet agent/puppet build version in
`acceptance/lib/helper.rb` since the only changes that have been made since the
package we currently have specified there are test changes.


rlinehan:
closing in favor of https://github.com/puppetlabs/puppet-server/pull/670. I didn't realize that had not yet been merged and so just pushed this commit onto that PR.

PR #674, opened 2015-08-13
Merge stable to master 2015 08 13

camlow325:


camlow325:
Most of these commits were no-ops from the merge-up.  The "bump to ssl-utils" and "prefer clj-time/seconds" ones were already in master, just being reapplied again because we had forgotten to put them into 1.x the first time and are just merging back up now.

The "acceptance test fix" ones have to do with pinning puppet-agent versions and we don't want to inherit the ones from stable into master.  I accepted the changes as-is from master for this merge conflict.

So most of the work being merged up here is related to the changes from SERVER-813 to fix the env cache flush problem.

The only thing I had to add on is one commit to change a reference from LinkedBlockingDeque to RegisteredLinkedBlockingDeque in one test.  This was necessary here because that test doesn't exist in the stable branch.

rlinehan:
:+1: this looks good to me.

camlow325:
@cprice404 or @fpringvaldsen - would appreciate it if one of you could give this a quick look at merge if you think it looks okay.  Mostly is just merging up code changes from @cprice404's PR for SERVER-813.

fpringvaldsen:
:shipit:

PR #677, opened 2015-08-18
(MAINT) merge 1.x to stable, 2015-08-18

cprice404:
FWIW I believe this is the last change to the OSS code that I will need for metrics stuff, and also the last patch I intended for 1.x at all, so I think maybe we can consider that branch to be in maintenance mode as soon as this gets merged all the way up.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -77,6 +77,12 @@
   :deploy-repositories [["releases" ~(deploy-info "http://nexus.delivery.puppetlabs.net/content/repositories/releases/")]
                         ["snapshots" ~(deploy-info "http://nexus.delivery.puppetlabs.net/content/repositories/snapshots/")]]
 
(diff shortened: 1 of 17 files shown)


cprice404:
@camlow325 @KevinCorcoran @fpringvaldsen if this gets merged today, I'll do the merge from stable->master tomorrow.

KevinCorcoran:
:see_no_evil: except for my question above about log messages.

camlow325:
I'd appreciate it if @fpringvaldsen could take a look at this since I didn't review the original PR for this which had been merged to the 1.x branch.

fpringvaldsen:
Also :+1: barring the question about the log messages

KevinCorcoran:
I'm just going to merge this; we can easily add back the logging if we find it to be useful in the future.

PR #680, opened 2015-08-21
(MAINT) Fix repl per break in SERVER-813

camlow325:
This commit fixes the user_repl.clj, previously broken by changes
introduced in SERVER-813.  The primary change in the commit is to rename
`pool->vec` to `registered-instances` per the name change made as part
of the work on that ticket.  The commit also includes some web-routing
and puppet-admin sections missing from the sample repl configuration.


nwolfe:
:yellow_heart:

PR #682, opened 2015-08-22
Validates SERVER-816

erikPrime:
This PR ports the External CA that runs from QATests against pe-puppetserver to FOSS puppet-server.
- removed the pe-specific certificate operations
- added Key Usage, Extended Key Usage, and Microsoft Universal Principal Name field to openssl.cnf


acceptance/suites/tests/020-external-ca/010_setup_remote_ca.rb
```diff
@@ -0,0 +1,136 @@
...
+pm_domain = fact_on(master, "domain").chomp
+ca_fqdn = fact_on(ca, "fqdn").chomp
+ca_hostname = fact_on(ca, "hostname").chomp
+
+#Pre-Setup: Define Directories used on Puppet Master and Root CA
+ca_dir = "/root/rakeca"
+int_ca_dir = "/root/rakeca/intermediate"
+p_ssl = "/etc/puppetlabs/puppet/ssl"
+
+step 'Push External CA public key onto all hosts' do
```
jeffmccune: Might be good to disambiguate this as an SSH user account public key rather than an SSL based public key.

acceptance/suites/tests/020-external-ca/020_retrieve_new_cert.rb
```diff
@@ -0,0 +1,21 @@
+test_name 'QA-1393 - C62572 - Register Agent with Puppet Master using an External CA'
+
+lockfile = '/opt/puppetlabs/puppet/cache/state/agent_catalog_run.lock'
+lockcopy = '/tmp/foo'
```
jeffmccune: Better to use `mkstemp` or something similar versus hard-coded temp directories.

Is `lockcopy` actually used?  I only see it in the teardown.

jeffmccune:
Only a few minor comments.  Overall, looks good to me.

acceptance/suites/tests/020-external-ca/020_retrieve_new_cert.rb
erikPrime: Jeff, it's also in line 17.

nwolfe: Line 17 is using `lockfile` - I don't see any other usages of `lockcopy` outside of teardown

acceptance/suites/tests/020-external-ca/010_setup_remote_ca.rb
camlow325: @erikPrime Were you going to update the step description to include something about this being the SSH public key, like Jeff had suggested?

erikPrime: I misunderstood and renamed the variable, not the Step name...  fixing.

acceptance/suites/tests/020-external-ca/010_setup_remote_ca.rb
```diff
@@ -0,0 +1,136 @@
...
+  hosts.each do |h|
+    on(h, "echo #{ca_ssh_key} >> /root/.ssh/authorized_keys")
+  end
+end
+
+step 'Install Git & Pull Down rakeca on External CA' do
+  install_package ca, 'git'
+  on(ca, "gem install rake")
+  on(ca, "rm -rf rakeca")
+  on(ca, "git clone https://github.com/LuvCurves/rakeca")
```
camlow325: Would be really nice for this to be under a PL-based project.

acceptance/suites/tests/020-external-ca/010_setup_remote_ca.rb
```diff
@@ -0,0 +1,136 @@
...
+end
+
+step 'turn on the fields defined in SERVER-816' do
+  #TODO: Someday we will deal with this string of hacks in a more graceful way
+  keyusage = 'keyUsage=critical,nonRepudiation,digitalSignature,keyEncipherment,dataEncipherment'
+  #extendedKeyUsage = 'extendedKeyUsage=serverAuth,clientAuth'
+  extendedKeyUsage = 'extendedKeyUsage=serverAuth,clientAuth,codeSigning,emailProtection,timeStamping,msCodeInd,msCodeCom,msCTLSign,msEFS'
+  on(ca, "cd #{ca_dir};sed -i -e 's/keyUsage = nonRepudiation, digitalSignature, keyEncipherment/#{keyusage}\\n#{extendedKeyUsage}/g' openssl.cnf.tmp")
```
camlow325: Where does the `openssl.cnf.tmp` file first get created?

acceptance/suites/tests/020-external-ca/010_setup_remote_ca.rb
```diff
@@ -0,0 +1,136 @@
...
+end
+
+step 'Correct owner and permissions inside the certs directory' do
+  on(master, 'chown -R puppet:puppet /etc/puppetlabs/puppet/ssl/certs')
+  on(master, 'chmod -R 0644 /etc/puppetlabs/puppet/ssl/certs')
+end
+
+
+step 'Restart Puppet Services' do
+  on(master, "service puppetserver start")
```
camlow325: To make this work when run in the PE pipeline, I think we'll want to determine the name of the service to restart more dynamically.  Could do this like the foreground.rb test does - https://github.com/puppetlabs/puppet-server/blob/e600a8c4909b1dbe046b5b807e4b766af988e8ba/acceptance/suites/tests/010-puppetserver-cli/subcommand/foreground.rb#L25.

acceptance/suites/tests/020-external-ca/020_retrieve_new_cert.rb
```diff
@@ -0,0 +1,16 @@
+test_name 'QA-1393 - C62572 - Register Agent with Puppet Master using an External CA'
+
+lockfile = '/opt/puppetlabs/puppet/cache/state/agent_catalog_run.lock'
```
camlow325: Seems okay, but maybe would be better to just stop the agent service so that it doesn't unintentionally run when we don't want it to.  Not sure how PE tests typically do this sort of thing...

acceptance/suites/tests/020-external-ca/030_revoke_agents.rb
```diff
@@ -0,0 +1,37 @@
...
+      on(ca, "cd /root/rakeca/intermediate;rake revoke CN=#{agent_fqdn}.cert")
+    end
+  end
+end
+
+step 'Push CRL to master & restart puppetserver' do
+  pm_fqdn = fact_on(master, "fqdn").chomp
+  on(ca, "cd /root/rakeca/intermediate;scp -o stricthostkeychecking=no ca_crl.pem root@#{pm_fqdn}:/etc/puppetlabs/puppet/ssl/ca/ca_crl.pem")
+  on(master, puppet('agent','--test', "--server #{master}"), :acceptable_exit_codes => [0,2])
+  on(master, "service puppetserver restart")
```
camlow325: Same comment here on the name being used for starting the service.

camlow325:
:see_no_evil: Just a few minor comments.

PR #685, opened 2015-08-26
(maint) Add handler for /environment/:environment route

nicklewis:
This route is used for retrieving information about an environment
catalog. This just passes through handling of the route to puppet, as
with all the similar routes.

### DIFF
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -51,6 +51,8 @@
                    (request-handler request))
     (comidi/GET ["/resource_types/" [#".*" :rest]] request
                    (request-handler request))
+    (comidi/GET ["/environment/" [#".*" :rest]] request
+                   (request-handler request))
     (comidi/GET "/environments" request
                    (request-handler request))
     (comidi/GET ["/status/" [#".*" :rest]] request
--- a/test/unit/puppetlabs/services/master/master_core_test.clj
+++ b/test/unit/puppetlabs/services/master/master_core_test.clj
@@ -24,6 +24,7 @@
     (doseq [[method paths]
             {:get ["catalog"
                    "node"
+                   "environment"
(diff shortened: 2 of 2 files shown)


KevinCorcoran:
:see_no_evil:

KevinCorcoran:
From the Travis CI build ...
```
FAIL in (admin-api-flush-jruby-pool-test) (jruby_pool_int_test.clj:142)
  actual: (not (true? false))
```
@cprice404 - Any thoughts?

camlow325:
I think I saw that once, too, but wasn't able to reproduce it later.

cprice404:
@camlow325 did you see that on your box?  Or in Travis?

@KevinCorcoran seems like we should probably file a separate ticket for that failure and deal with it in triage, and consider this PR independently.

@nicklewis I have some reservations about this (just in terms of timing) but would be willing to discuss; seems like it might be more appropriate to do that in Jira for posterity, do you have a ticket number?

camlow325:
> @camlow325 did you see that on your box? Or in Travis?

Pretty sure I saw that on my local box but I can't remember now.  I had some other local changes in place at the time I'd run the tests so I chalked up my lack of an ability to reproduce it to some of that at the time but it sounds now like there might be a real problem here.

nicklewis:
@cprice404 The placeholder is APPMGMT-94.

cprice404:
@nicklewis thanks, will add some comments over there.

cprice404:
After some discussion and clarification on the Jira ticket, I'm +1 on this.  @camlow325 can you merge this today unless you have any other concerns?

cprice404:
I don't want to hold up the app management team and this seems pretty straightforward to me, so I'm going to merge it.  If anyone has any concerns we can deal with them retroactively.

camlow325:
FWIW now that it's already been merged, :+1: from me.

PR #687, opened 2015-08-28
(SERVER-835) Integrate trapperkeeper-authorization and associated settings

camlow325:
This PR integrates Puppet Server with trapperkeeper-authorization for authorizing access to both Clojure-level and core Ruby Puppet-hosted endpoints.  By default, the client-whitelists are still honored, if present, for puppet-admin and certificate-status(es) Clojure endpoints.

A new optional setting, use-legacy-auth-conf, can be set to 'false' in order to have tk-auth used for authorizing core Ruby Puppet-hosted endpoints.  The default for the use-legacy-auth-conf setting, however, is 'true'.


camlow325:
Rebased due to merge conflicts at 7971c1a

camlow325:
Rebased again due to merge conflicts at 9d1b876.

src/clj/puppetlabs/services/master/master_service.clj
```diff
@@ -38,12 +41,17 @@
...
+                            (-> (core/root-routes handle-request)
+                                (#(comidi/context path %))
+                                comidi/routes->handler)
+                            wrap-with-authorization-check
+                            use-legacy-auth-conf
+                            puppet-version))]
        (if (map? route-config)
-         (add-ring-handler this ring-handler {:route-id :master-routes})
+         (add-ring-handler this ring-handler
+                           {:route-id :master-routes})
```
nwolfe: :hand: :hand: :gun:

src/clj/puppetlabs/services/puppet_admin/puppet_admin_core.clj
```diff
@@ -101,24 +99,40 @@
...
-                  "configuration settings")))))
+  (:puppet-admin config))
+
+(schema/defn ^:always-validate puppet-admin-settings->whitelist-settings
+  :- ringutils/WhitelistSettings
+  "Given the full Puppet Admin web service settings, extract and return the
+  embedded client whitelist settings."
+  [admin-settings :- (schema/maybe PuppetAdminSettings)]
+  (select-keys admin-settings [:client-whitelist
+                               :authorization-required]))
```
nwolfe: Not that we expect the keys to change, but you might get the keys from `ringutils/WhitelistSettings` rather than hardcode them again here.

src/ruby/puppet-server-lib/puppet/server/auth_provider.rb
```diff
@@ -0,0 +1,13 @@
+## This is a "dummy" AuthProvider that Puppet Server registers with core
+## Puppet for cases where Puppet Server decides that it should be in charge of
+## authorizing requests at the Clojure / Ring handler level - depending upon
+## the configuration of the 'use_legacy_auth_conf' setting - and not core
+## Puppet.
+
+class Puppet::Server::AuthProvider
```
nwolfe: Hm, we might not want this to be the same name as the Puppet class, although that may never technically bite us?

nwolfe: Oh just noticed this is in a server-specific namespace

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -718,11 +717,6 @@
    service return a map with of all the CA settings.
    Throws an exception if any required configuration is not found."
```
nwolfe: Might delete this from the docstring now - no biggie though

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -286,3 +280,28 @@
...
+   puppet-version :- schema/Str]
+  ;; For backward compatibility, requests to the .../certificate_status* endpoint
+  ;; will be authorized by a client-whitelist, if one is configured for
+  ;; 'certificate_status'.  When we are able to drop support for
+  ;; client-whitelist authorization later on, we should be able to get rid of
+  ;; the 'wrap-with-trapperkeeper-or-client-whitelist-authorization' function
+  ;; and replace with a line chaining the handler into a call to
+  ;; 'authorization-fn'.
+  (let [whitelist-path (str path
+                            (if (not= \/ path) "/")
```
nwolfe: (woah)

src/clj/puppetlabs/services/legacy_routes/legacy_routes_service.clj
```diff
@@ -6,46 +6,57 @@
...
+                                              wrap-with-authorization-check
+                                              use-legacy-auth-conf
+                                              puppet-version)
+                               :api-version master-core/puppet-API-version}
           real-ca-service? (= (namespace (tk-services/service-symbol ca-service))
                               "puppetlabs.services.ca.certificate-authority-service")
+          ca-settings (ca/config->ca-settings (get-config))
+          ca-route-handler (-> ca-settings
+                               (ca-core/web-routes)
+                               (#(comidi/context path %))
```
nwolfe: Could `partial` the numerous occurrences of this like `(partial comidi/context path)`

src/clj/puppetlabs/services/puppet_admin/puppet_admin_core.clj
nwolfe: Oh I see those keys aren't really accessible as they're `schema/optional-key`s

nwolfe:
:+1: 

Functionally reviewed using `curl` - works as expected.

Agent runs aren't possible yet as the CA endpoints can't be accessed without a SSL cert (e.g. catch 22).
This is expected though.

As @camlow325 and I discussed IRL, we'll have to wait to merge this until:
* @jeffmccune 's PR for allow-unauthenticated is merged
* new tk-authz release
* @camlow325 updates this PR with `:match-request` changes and `allow-unauthenticated` rules for CA endpoints.

camlow325:
Added e6c251a to bump puppet-server dependency up to newer trapperkeeper-authorization version, 0.1.2.  This version has support for `allow-authenticated` and the new `:match-request` rule structure.  This commit includes an updated ezbake `auth.conf` file that will permit appropriate access to the certificate authority endpoints so that a new agent run can succeed.

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
```diff
@@ -92,19 +92,19 @@
...
+  (let [body-for-jruby (body-for-jruby request)
+        request-with-params (-> request
+                                (assoc :body body-for-jruby)
+                                pl-ring-params/params-request)]
+    ; :params may or may not been populated by ring middleware at this point.
+    ; Even if they were, though, they may have key/value pairs in them that
+    ; originated from the comidi route destructuring.  Ignore whatever may
+    ; have been there and repopulate it with what ring pulled from the form
+    ; and query params.
+    (assoc request-with-params :params
```
camlow325: I did this to resolve a conflict introduced by the inclusion of tk-authz 0.1.2.  tk-authz now wraps the Ring middleware which destructures request params into the Ring request map - in order to be able to pull out query params from the request query string.  Previously, this code would clear out the `:params` key before "re-running" the Ring params middleware, which would leave the `:params` not set because `:query-params` and `:form-params` had been pre-populated.  With this change, the `:params` are now always recomposed from the `:query-params` and `:form-params`, whether the `:query-params` and `:form-params` had originally been populated onto the request from tk-authz or from within this `wrap-params-for-jruby` function.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_service.clj
camlow325: Done in c456582.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
camlow325: This wasn't right.  I was intending to have this just add "/" if the `path` didn't already end with one.  I changed this to ...

~~~~clojure
(if (not= \/ (last path)) "/")
~~~~

... in c456582.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
camlow325: Done in c456582.

camlow325:
I realized one other problem when attempting to integrate this with PE Puppet Server.  The `config->ca-settings` function needs to not try to associate any content from the `certificate-authority` section other than the `certificate-status` key/value.  This produced a schema violation with PE Puppet Server when the `reverse-proxy-ca-service` was configured with the proxy-related content that it manages in the `certificate-authority-section`.  Addressed this in 628f77e.

src/clj/puppetlabs/puppetserver/ringutils.clj
```diff
@@ -119,4 +111,40 @@
...
+  starts with the value provided to this function for 'whitelist-path' and if
+  the 'whitelist-settings' are non-empty.  In all other cases, requests are
+  routed to the handler constructed from the 'authorization-fn'"
+  [base-handler :- IFn
+   authorization-fn :- IFn
+   whitelist-path :- schema/Str
+   whitelist-settings :- (schema/maybe WhitelistSettings)]
+  (let [handler-with-trapperkeeper-authorization (authorization-fn base-handler)]
+    (if-let [handler-with-client-whitelist-authorization
+             (if (not-empty whitelist-settings)
```
dankreek: It seems that using `(when)` when you don't have an else clause is more idiomatic, but this doesn't need to be fixed.

dankreek:
Everything LGTM. :+1:

PR #688, opened 2015-08-31
(QENG-2770) Use default option value when env var is empty string.

waynr:


acceptance/lib/helper.rb
```diff
@@ -61,7 +61,9 @@ def self.get_option_value(value, legal_values, description,
 
     # precedence is environment variable, option file, default value
     value = ((env_var_name && ENV[env_var_name]) || value || default_value)
-    if value
+    if value == ""
+      value = default_value
```
camlow325: For what cases should we be doing a `to_sym` on the value?  Looks like if we get here via the environment variable not being defined and value = "" that what we'll use as the value, `default_value`, won't have `to_sym` applied to it.  If we were to get to using `default_value` because the environment variable was not defined and `value` supplied to the function was nil, though, I think we'd get `default_value.to_sym`.

nwolfe: It seems like `to_sym` is already intelligently applied in this function.  We won't get `default_value.to_sym` when `default_value` is nil because it won't enter either the `if` or `elsif`, it'll fall through to the nil check below.

nwolfe:
:+1:

acceptance/lib/helper.rb
camlow325: I'm wondering if we should eliminate the `elsif` and just require the value to have been converted to a symbol before calling into the `get_option_value` function.  Seems confusing to only conditionally convert to a symbol from within this function.  I haven't looked at the callers of this function, though, so maybe I'm offbase on this.

nwolfe: Hm, yeah, there's probably room for improvement in here but is it worth the trouble?  You could argue that changing all the usages (and missing a few?) is worse than making the one function smarter.

camlow325: Seems like this function makes an assumption that `value` will always come from an options file and that, in all of the cases where the option is set, the `value` needs to be converted to a symbol?  Wasn't clear to me that this is a safe assumption to make in all cases, in which case doing the conversion closer to the caller might make more sense.

I'll leave it up to @waynr as to what's best to do here.  Just seemed a bit confusing / fragile to me.  But I'll admit that I don't have a very good grasp of what's going on here.

waynr: @camlow325 I agree that the `get_option_value` behavior is confusing and I'm not sure what exactly can be done to make it better. We could add an optional argument to `get_option_value` so that it knows what the preferred type of its return value should be (`:string`, `:symbol`, etc)

waynr:
@camlow325 I pushed a commit that demonstrates what I mentioned in my last comment. I'm not sure it really makes `get_option_value` any more understandable and I don't think it's worth spending more time on unless you want me to revert the head of this PR back to the previous commit.

acceptance/lib/helper.rb
```diff
@@ -9,27 +9,28 @@ def self.initialize_config(options)
...
 
     puppetserver_version =
         get_option_value(options[:puppetserver_version],
                          nil, "Puppet Server Version",
-                         "PUPPETSERVER_VERSION", nil)
+                         "PUPPETSERVER_VERSION", nil, :string)
 
     puppet_version = get_option_value(options[:puppet_version],
-                         nil, "Puppet Version", "PUPPET_VERSION", nil) ||
+                         nil, "Puppet Version", "PUPPET_VERSION", nil, :symbol) ||
```
camlow325: Should the `value_type` for this one be a `:string` instead of a `:symbol`?

camlow325:
@waynr I think your follow-up commit is an improvement over what was there.  @nwolfe - what do you think?

camlow325:
Seems like this may need a rebase now?

nwolfe:
> @nwolfe - what do you think?

Sure, let's try it!

acceptance/lib/helper.rb
waynr: Fixed.

waynr:
@camlow325 okay, I have rebased and made changes per your most recent comment.

nwolfe:
@waynr rebase again?

nwolfe:
@waynr Actually, could you please retarget this at stable so we can then merge it up to master?

waynr:
@nwolfe just targeted at stable in #690

PR #689, opened 2015-09-01
(MAINT) Upgrade puppet-agent package

nwolfe:
The version we were pinned to has been purged from the nightly repo, so
we need to bump it up to the newest (green) build. This also bumps the
puppet submodule to match the version expected by puppet-agent.


camlow325:
:100: :fast_forward:

PR #690, opened 2015-09-01
(QENG-2770) Make it possible to use default value of option when given value is empty string.

waynr:


nwolfe:
:+1:

Once this lands, I can update our ci-job-configs to remove the version and see if this actually works.

camlow325:
:rocket: :hammer: :racehorse: :zap:

PR #693, opened 2015-09-09
(PUP-4536) Add docs comment about module code compatibility

ahpook:
Previously we didn't have a message to users describing the different
Ruby runtime versions between Puppet Server and the Agent.

This commit adds a section to the Apache/Passenger->Puppet Server
doc describing the Ruby compatibility requirements for code that
runs on the server vs agent.

### DIFF
--- a/documentation/puppetserver_vs_passenger.markdown
+++ b/documentation/puppetserver_vs_passenger.markdown
@@ -30,6 +30,10 @@ If you have server-side Ruby code in your modules, Puppet Server will run it via
 JRuby. Generally speaking, this only affects custom parser functions and report
 processors. For the vast majority of cases, this shouldn't pose any problems, as JRuby is highly compatible with vanilla Ruby.
 
+## Ruby Compatibility For Extensions
+
+Ruby extension code in your modules needs to run under both Ruby 1.9 and Ruby 2.1. This is because Puppet Server runs Puppet functions and custom resource types under JRuby (which is a Ruby 1.9-compatible interpreter), and the official `puppet-agent` releases run custom facts and types/providers under MRI Ruby 2.1.
+
 ## Installing And Removing Gems
 
 We isolate the Ruby load paths that are accessible to Puppet Server's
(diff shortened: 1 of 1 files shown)


ahpook:
ping @nfagerlund @kylog since y'all commented on the ticket - how's this look?

documentation/puppetserver_vs_passenger.markdown
```diff
@@ -30,6 +30,10 @@ If you have server-side Ruby code in your modules, Puppet Server will run it via
 JRuby. Generally speaking, this only affects custom parser functions and report
 processors. For the vast majority of cases, this shouldn't pose any problems, as JRuby is highly compatible with vanilla Ruby.
 
+## Ruby Compatibility For Extensions
+
+Because the Puppet Server runs master under JRuby, extension code that runs on the server -- parser functions and custom Ruby types -- needs to work under JRuby, which is a Ruby 1.9-compatible interpreter. The All-in-One Puppet 4 agent, where custom facts and types/providers run, uses MRI Ruby 2.1. So for maximum compatibility, Ruby code in your modules should run under both Ruby 1.9 and Ruby 2.1.
```
nfagerlund: I suggest trimming / rearranging a bit:

> Ruby extension code in your modules needs to run under both Ruby 1.9 and Ruby 2.1. This is because Puppet Server runs Puppet functions and custom resource types under JRuby (which is a Ruby 1.9-compatible interpreter), and the official `puppet-agent` releases run custom facts and types/providers under MRI Ruby 2.1.

ahpook: That's better thanks. Amended and pushed.

kylog:
I'd say `JRuby 1.7` because that's the version that has 1.9.3 compat. (As contrasted with JRuby 9000, which introduces Ruby 2.2 compat.)

/cc @cprice404 to keep me honest

camlow325:
Yep, what @kylog said is true.

nwolfe:
:+1:

PR #698, opened 2015-09-21
(SERVER-860) (WIP) Introduce ReentrantReadWriteLock into JRubyPool da…

rlinehan:
…ta structure

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
@@ -4,7 +4,8 @@
             [puppetlabs.services.jruby.puppet-environments :as puppet-env]
             [me.raynes.fs :as fs]
             [clojure.tools.logging :as log])
-  (:import (com.puppetlabs.puppetserver PuppetProfiler JRubyPuppet RegisteredLinkedBlockingDeque)
+  (:import (com.puppetlabs.puppetserver PuppetProfiler JRubyPuppet)
+           (com.puppetlabs.puppetserver.pool JRubyPool)
            (puppetlabs.services.jruby.jruby_puppet_schemas JRubyPuppetInstance PoisonPill)
            (java.util HashMap)
            (org.jruby CompatVersion Main RubyInstanceConfig RubyInstanceConfig$CompileMode)
@@ -60,7 +61,7 @@
   "Instantiate a new queue object to use as the pool of free JRubyPuppet's."
   [size]
   {:post [(instance? jruby-schemas/pool-queue-type %)]}
-  (RegisteredLinkedBlockingDeque. size))
+  (JRubyPool. size))
 
 (schema/defn ^:always-validate managed-environment :- jruby-schemas/EnvMap
   "The environment variables that should be passed to the Puppet JRuby
@@ -153,7 +154,7 @@
 (schema/defn borrow-with-timeout-fn :- JRubyPuppetInternalBorrowResult
   [timeout :- schema/Int
    pool :- jruby-schemas/pool-queue-type]
-  (.pollFirst pool timeout TimeUnit/MILLISECONDS))
+  (.borrowItemWithTimeout pool timeout TimeUnit/MILLISECONDS))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Public
@@ -211,7 +212,7 @@
                                                            JRubyPuppet)
                         :scripting-container  scripting-container
                         :environment-registry env-registry})]
-        (.registerLast pool instance)
+        (.register pool instance)
         instance))))
 
 (schema/defn ^:always-validate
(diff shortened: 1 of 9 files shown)


rlinehan:
@cprice404 didn't quite manage to finish this up today (mostly need to write a commit message and fix the git history - I just realized this is missing the other test file -  and clean up a couple things), but I wanted to put this up for now just to get feedback. In particular I don't really know what I'm doing with java, so if there's any comments you have there that would be great.

src/java/com/puppetlabs/puppetserver/pool/LockablePool.java
```diff
@@ -0,0 +1,25 @@
...
+
+    public void returnItem(E e) throws InterruptedException;
+
+    public void insertPill(E e) throws InterruptedException;
+
+    public void clear();
+
+    public int remainingCapacity();
+
+    public int size();
```
rlinehan: I'm not sure whether `clear`, `remainingCapacity`, and `size` should be in this interface?

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -299,9 +300,10 @@
                      (:id instance)
                      max-requests)
           (flush-instance-fn pool instance))
-        (.putFirst pool instance)))
-    ;; if we get here, we got a Retry, so we just put it back into the pool.
-    (.putFirst (:pool instance) instance)))
+        (.returnItem pool instance)))
+    ;; if we get here, it was from a borrow and we got a Retry, so we just
+    ;; return it to the pool.
+    (.returnItem (:pool instance) instance)))
```
cprice404: I think it's probably functionally equivalent, but, technically I think we're putting a Pill back into the pool here, so I wonder if just in terms of semantics it would be better to call your {{insertPill}} method?

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Override
+    public E borrowItemWithTimeout(long timeout, TimeUnit unit) throws InterruptedException, IllegalStateException {
+        if (lock.isWriteLockedByCurrentThread()) {
+          throw new IllegalStateException("The current implementation has a risk of deadlock if you attempt to borrow a JRuby instance while holding the write lock!");
+        }
+        E item = live_queue.pollFirst(timeout, unit);
+        lock.readLock().lock();
+        return item;
+    }
```
cprice404: It'd be nice if we could refactor the two methods above to share some common code, but, in Java 7 there's no way to do that without a bunch of extra interface junk that would almost certainly end up being less maintainable than the duplication you've got here now, so it's probably better to leave it like this.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Override
+    public void returnItem(E e) throws InterruptedException {
+        live_queue.putFirst(e);
+        lock.readLock().unlock();
```
cprice404: Typically whenever you're holding a lock it's a good idea to put the `unlock` in a `try`/`finally` block, to be as certain as possible that the lock will be released, because if an exception is thrown somewhere else and the lock isn't released it will cause problems.

In this case it probably won't make much of a difference because there's only one line of code where an exception could happen, and if something went wrong on that line of code, we'd probably already be hosed... but I might be inclined to put the `try`/`finally` in anyway, just because it's generally a good pattern to follow.

src/java/com/puppetlabs/puppetserver/pool/LockablePool.java
cprice404: Well, ideally, the only things that would be in the interface would be public methods that we really wanted to allow callers to use as part of the API, and if this were pure Java it's possible we'd leave those out of the interface and just make them `protected` methods that we called in the few spots where we need them for internal purposes, but clojure doesn't really play nicely with `protected` methods and it would be debatable whether that was a better way to do it even if clojure did play well.  So I think this is fine.

cprice404:
This all looks good to me so far!

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
rlinehan: I spent a while trying to puzzle over this, and I think (but am not convinced) that this needs to be a `returnItem`. I changed the `putFirst` in `flush-pool!` to `insertPill` (https://github.com/puppetlabs/puppet-server/pull/698/files#diff-af969cd26304257b62d766cb4c5325faR118), because there there isn't anything holding a read lock; we're just adding the `RetryPoisonPill` to the pool for the first time.

Here, on the other hand, it's possible that a borrow could have gotten the retry pill (https://github.com/puppetlabs/puppet-server/blob/master/src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj#L106). In that case, since it did a borrow it holds a read lock, which we need to unlock, which is why `returnItem`.

rlinehan:
Closing in favor of #699.

PR #699, opened 2015-09-22
(SERVER-860) Introduce ReentrantReadWriteLock into JRubyPool data structure

rlinehan:
Rename RegisteredLinkedBlockingDeque to JRubyPool and move the locking logic relating to atomic code deploys into this data structure, so that we have one encapsulated datastructure managing the JRuby pool.

Add tests for our required behavior of this encapsulated data structure.

I have tried to separate out the commits so that adding/moving of files is in a separate commit from modifying those commits and is hopefully easier to understand.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
@@ -4,7 +4,8 @@
             [puppetlabs.services.jruby.puppet-environments :as puppet-env]
             [me.raynes.fs :as fs]
             [clojure.tools.logging :as log])
-  (:import (com.puppetlabs.puppetserver PuppetProfiler JRubyPuppet RegisteredLinkedBlockingDeque)
+  (:import (com.puppetlabs.puppetserver PuppetProfiler JRubyPuppet)
+           (com.puppetlabs.puppetserver.pool JRubyPool)
            (puppetlabs.services.jruby.jruby_puppet_schemas JRubyPuppetInstance PoisonPill)
            (java.util HashMap)
            (org.jruby CompatVersion Main RubyInstanceConfig RubyInstanceConfig$CompileMode)
@@ -60,7 +61,7 @@
   "Instantiate a new queue object to use as the pool of free JRubyPuppet's."
   [size]
   {:post [(instance? jruby-schemas/pool-queue-type %)]}
-  (RegisteredLinkedBlockingDeque. size))
+  (JRubyPool. size))
 
 (schema/defn ^:always-validate managed-environment :- jruby-schemas/EnvMap
   "The environment variables that should be passed to the Puppet JRuby
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
@@ -154,7 +154,7 @@
 (schema/defn borrow-with-timeout-fn :- JRubyPuppetInternalBorrowResult
   [timeout :- schema/Int
    pool :- jruby-schemas/pool-queue-type]
-  (.pollFirst pool timeout TimeUnit/MILLISECONDS))
+  (.borrowItemWithTimeout pool timeout TimeUnit/MILLISECONDS))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Public
@@ -212,7 +212,7 @@
                                                            JRubyPuppet)
                         :scripting-container  scripting-container
                         :environment-registry env-registry})]
-        (.registerLast pool instance)
+        (.register pool instance)
         instance))))
 
 (schema/defn ^:always-validate
@@ -235,7 +235,7 @@
 
 (schema/defn borrow-without-timeout-fn :- JRubyPuppetInternalBorrowResult
   [pool :- jruby-schemas/pool-queue-type]
-  (.takeFirst pool))
+  (.borrowItem pool))
 
 (schema/defn borrow-from-pool!* :- jruby-schemas/JRubyPuppetBorrowResult
   "Given a borrow function and a pool, attempts to borrow a JRuby instance from a pool.
@@ -247,7 +247,7 @@
   (let [instance (borrow-fn pool)]
     (cond (instance? PoisonPill instance)
           (do
-            (.putFirst pool instance)
+            (.returnItem pool instance)
             (throw (IllegalStateException.
                      "Unable to borrow JRuby instance from pool"
                      (:err instance))))
@@ -300,9 +300,10 @@
                      (:id instance)
                      max-requests)
           (flush-instance-fn pool instance))
-        (.putFirst pool instance)))
-    ;; if we get here, we got a Retry, so we just put it back into the pool.
-    (.putFirst (:pool instance) instance)))
+        (.returnItem pool instance)))
+    ;; if we get here, it was from a borrow and we got a Retry, so we just
+    ;; return it to the pool.
+    (.returnItem (:pool instance) instance)))
 
 (schema/defn ^:always-validate new-main :- jruby-schemas/JRubyMain
   "Return a new JRuby Main instance which should only be used for CLI purposes,
--- a/src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
+++ b/src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
@@ -44,8 +44,8 @@ public final class JRubyPool<E> implements LockablePool<E> {
     */
     @Override
     synchronized public void register(E e) throws InterruptedException {
-        liveQueue.putLast(e);
         registeredElements.add(e);
+        liveQueue.putLast(e);
     }
 
     /**
--- a/src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
+++ b/src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
@@ -166,9 +166,9 @@ public final class JRubyPool<E> implements LockablePool<E> {
     /**
     * Acquires a write lock from <tt>ReentrantReadWriteLock</tt>.
     *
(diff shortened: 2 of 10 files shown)


src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -299,9 +300,10 @@
                      (:id instance)
                      max-requests)
           (flush-instance-fn pool instance))
-        (.putFirst pool instance)))
-    ;; if we get here, we got a Retry, so we just put it back into the pool.
-    (.putFirst (:pool instance) instance)))
+        (.returnItem pool instance)))
+    ;; if we get here, it was from a borrow and we got a Retry, so we just
+    ;; return it to the pool.
+    (.returnItem (:pool instance) instance)))
```
rlinehan: Copying this comment thread over from #698 (https://github.com/puppetlabs/puppet-server/pull/698#discussion_r40011586) because that was closed.

Chris:
>I think it's probably functionally equivalent, but, technically I think we're putting a Pill back into the pool here, so I wonder if just in terms of semantics it would be better to call your `insertPill` method?

My response:
>I spent a while trying to puzzle over this, and I think that this needs to be a `returnItem`. I changed the `putFirst` in `flush-pool!` to `insertPill` (https://github.com/puppetlabs/puppet-server/pull/698/files#diff-af969cd26304257b62d766cb4c5325faR118), because there there isn't anything holding a read lock; we're just adding the `RetryPoisonPill` to the pool for the first time.

>Here, on the other hand, it's possible that a borrow could have gotten the retry pill (https://github.com/puppetlabs/puppet-server/blob/master/src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj#L106). In that case, since it did a borrow it holds a read lock, which we need to unlock, which is why `returnItem`.

@cprice404 does what I'm saying here make sense, and should I stick with using `returnItem` here?

cprice404: OK, yep, I think you're right.  Good catch.

cprice404:
@camlow325 if you end up with any bandwidth to review this PR, we'd definitely appreciate your input.

src/java/com/puppetlabs/puppetserver/pool/LockablePool.java
```diff
@@ -0,0 +1,25 @@
...
+    public void clear();
+
+    public int remainingCapacity();
+
+    public int size();
+
+    public void lock() throws InterruptedException;
+
+    public void unlock() throws InterruptedException;
+}
```
KevinCorcoran: very minor Java note: all of the `public` modifiers on these methods are redundant; methods in an interface are all always public.

KevinCorcoran: It would also be nice to have javadocs on these methods.

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
```diff
@@ -0,0 +1,193 @@
...
+    (.returnItem pool instance)))
+
+(deftest pool-lock-is-blocking-test
+  (let [pool (create-populated-pool 3)
+        instances (borrow-n-instances pool 3)
+        future-started? (promise)
+        lock-acquired? (promise)
+        unlock? (promise)]
+    (is (= 3 (count instances)))
+    (let [lock-thread (future (do (deliver future-started? true)
```
KevinCorcoran: This `do` is unnecessary (and there are several more of these below, which are also unnecessary).

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
```diff
@@ -0,0 +1,193 @@
...
+      (is (true? true))
+      (let [borrow-thread-1 (future (do
+                                      (let [instance (.borrowItem pool)]
+                                        (.returnItem pool instance))))
+            borrow-thread-2 (future (do
+                                      (let [instance (.borrowItem pool)]
+                                        (.returnItem pool instance))))]
+        ;; this is racey, but the only ways i could think of to make it non-racey
+        ;; depended on knowledge of the implementation
+        (Thread/sleep 500)
```
KevinCorcoran: @cprice404 - Do you remember what those other ways were?  It seems quite possible that it would preferable for this test to depend on the implementation if that enabled the elimination of this race condition.

(sorry, I realize this is asking a question about code you *didn't* write several weeks ago).

KevinCorcoran: ... especially since the test below is dependent on the current implementation of `JRubyPool` and specifically calls that out.

KevinCorcoran:
:see_no_evil:

rlinehan:
I've pushed up several commits addressing feedback; mostly cleanup things and adding javadocs. I think I've addressed all feedback except for dealing with the race condition in `pool-lock-reentrant-with-many-borrows-test`.

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
```diff
@@ -0,0 +1,191 @@
...
+        (is (not (realized? borrow-thread-2)))
+
+        ;; Current implementation of JRubyPool is not reentrant. This assertion
+        ;; tests that the error we expect is thrown, rather than a deadlock
+        ;; happening. The tests for the behavior we want have been commented out
+        ;; until we change the implementation.
+        (is (thrown? IllegalStateException (.borrowItem pool)))
+        ;; (let [instance (.borrowItem pool)]
+        ;;   (is (true? true))
+        ;;   (.returnItem pool instance))
```
cprice404: So, this whole test is effectively commented out because we don't acutually support re-entrant borrows... so it seems like it would probably make sense to just comment out those lines as well?  And worry about race conditions when we move to an implementation that is re-entrant?

KevinCorcoran: that would be fine with me

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
```diff
@@ -0,0 +1,194 @@
...
+    * This method is analagous to <tt>putLast</tt> in the
+    * <tt>LinkedBlockingDeque</tt> class, but also causes the element to be
+    * added to the list of "registered" elements that will be returned by
+    * <tt>getRegisteredInstances</tt>.
+    *
+    * Note that this method is synchronized to try to ensure that the addition
+    * to the queue and the list of registered instances are visible roughly
+    * atomically to consumers, but because the underlying queue uses its own
+    * lock, it is possible for it to be modified on another thread while this
+    * method is being executed.
```
KevinCorcoran: I got pretty hung-up trying to understand this paragraph.  It particular, I think the phrase `visible roughly atomically` is pretty confusing.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
```diff
@@ -0,0 +1,194 @@
...
+    */
+    @Override
+    public int size() {
+        return liveQueue.size();
+    }
+
+    /**
+    * Acquires a write lock from <tt>ReentrantReadWriteLock</tt>.
+    *
+    * Note that this implementation does not fulfill requirement (e); with
```
KevinCorcoran: What is `requirement (e)`?

KevinCorcoran: Oh, I see, it's in the javadocs on the interface.  Might want to explain that.

KevinCorcoran:
:+1:

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
rlinehan: I just copied back in the paragraph that Chris had here before. TBH I also find it pretty hard to understand. I'll try to see if I can improve it though...

rlinehan: In the process of trying to come up with a better description of this, I discovered that I messed up in a previous commit and accidentally reordered the `.putLast` and the `.add` to `registeredElements` (https://github.com/rlinehan/puppet-server/commit/67afeb70eb1bf813009bdf2a8d5c28137f135eb7#diff-42395ae4a364917085c26d57d9e4cf85R20). I've added a commit 8a8e636 fixing this. (Thanks @scottyw for helping me work through this!)

rlinehan: Trying to come up with a better explanation of this...

~~~
This method is synchronized not for thread safety, but to try to ensure that
to consumers of this class adding an instance to the queue and adding it to
the set of registered elements is visible roughly atomically. `synchronize`
guarantees that registration is atomic, so (ignoring borrows happening)
`RegisteredElements` and `liveQueue` can only diverge by 1. "This is only
"roughly" atomic, as the underlying queue uses its own lock, so it is
possible for it to be modified on another thread while this method is being
executed.
 ~~~

suggestions/improvements welcome.

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
rlinehan: addressed in 9eb12e1aea53421b6fa8a038ef7938449fc08f04

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
cprice404: That sounds fine (and better).  TBH I don't think it is worth spending a bunch of time on this right now so I'd say change it to that and then let's put this to bed.

cprice404:
@camlow325 what's the likelihood you'd be able to give this a quick skim tomorrow?  I think we need to get it merged before EOW so if you won't have bandwidth, we'll just have to merge it anyway and maybe you can retroactively review later?

camlow325:
@cprice404 Have a few things I want to take care of this morning but will aim to take a look at this today.  If I haven't gotten to it by then, though, I'm good with ye just merging it anyway, and I could then take a look back at it later.

cprice404:
@camlow325 merging this as it is blocking other work.  Would still appreciate your input.

PR #702, opened 2015-09-22
(MAINT) Bump Puppet submodule to 3.8.3

dankreek:


camlow325:
I think we may also need to update the `PUPPET_BUILD_VERSION` in `acceptance/lib/helper.rb` to `3.8.3`?  Not totally sure how that gets used, TBH.

camlow325:
Also, there's a test in `puppet-server-lib/puppet/jvm/master_spec.rb` that would need to have an assertion reference of `3.8.1` bumped up to `3.8.3`, I think.

camlow325:
Also, it looks like Travis wasn't running for this PR.  May need to add "1.x" to the `.travis.yml` file to make this happen - https://github.com/puppetlabs/puppet-server/blob/1.x/.travis.yml#L19-L20?  Would be good to just take care of that on this PR so we don't forget for next time.

camlow325:
:+1: other than bumping `PUPPET_BUILD_VERSION`, the version being picked up in the `master_spec.rb`, and re-enabling Travis runs for the 1.x branch.

camlow325:
:+1:

PR #703, opened 2015-09-22
(MAINT) Bump Puppet submodule to 4.2.2

dankreek:


camlow325:
I also think we should bump the `PUPPET_BUILD_VERSION` in `acceptance/lib/helper.rb` up to `1.2.5` as part of this PR.  `1.2.5` is the latest `puppet-agent` package, and it includes Puppet `4.2.2`.  This should ensure that we're getting the right version when installing from packages in CI.

camlow325:
:+1: other than bumping up to a newer version of `puppet-agent`.

camlow325:
:+1:

PR #704, opened 2015-09-23
(maint) Bump dev profile dependency on tools.namespace to 0.2.5 to ad…

kylog:
…dress dependency conflict

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -91,7 +91,7 @@
   :classifiers [["test" :testutils]]
 
   :profiles {:dev {:source-paths  ["dev"]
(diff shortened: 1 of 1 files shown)


kylog:
This made the pain go away when following the directions to set up a dev environment. It seems pretty likely that I've done something foolish, so please feel free to set me straight ...

no_idea

nwolfe:
I don't see any harm in bumping this dependency up if it helped you get up and running.  Looks like there's a version `0.2.11` out, but we don't need to rush to it; this namespace is just used for the REPL dev utilities.

:+1:

PR #705, opened 2015-09-23
(SERVER-846) Default tk-authz auth.conf rules

nwolfe:
This commit fills out the new auth.conf with default rules as they're
specified in the legacy Ruby auth.conf. Tests and developer resources
have also been updated to include new required fields.

Planning to do a follow-up pull request including some simple acceptance
tests for these default rules.

### DIFF
--- a/ezbake/config/conf.d/auth.conf
+++ b/ezbake/config/conf.d/auth.conf
@@ -1,39 +1,116 @@
 authorization: {
-    # TODO: Fill these in with real default rule definitions - SERVER-846.
-    # Putting just the ones for the certificate authority in for now since
-    # these are always evaluated by trapperkeeper-authorization and are
-    # needed to complete a successful agent run.
+    version: 1
     rules: [
         {
+            # Allow nodes to retrieve their own catalog
             match-request: {
-                path: "/puppet-ca/v1/certificate_revocation_list/ca"
-                type: path
-                method: get
+                path: "^/puppet/v3/catalog/([^/]+)$"
+                type: regex
+                method: [get, post]
             }
-            allow: "*"
+            allow: "$1"
+            sort-order: 500
+            name: "puppetlabs catalog"
         },
         {
-            match-request: {
-                path: "/puppet-ca/v1/certificate/ca"
-                type: path
-                method: get
-            }
-            allow-unauthenticated: true
-        },
-        {
+            # Allow nodes to retrieve the certificate they requested earlier
             match-request: {
                 path: "/puppet-ca/v1/certificate/"
                 type: path
                 method: get
             }
             allow-unauthenticated: true
+            sort-order: 500
+            name: "puppetlabs certificate"
         },
         {
+            # Allow all nodes to access the certificate revocation list
+            match-request: {
+                path: "/puppet-ca/v1/certificate_revocation_list/ca"
+                type: path
+                method: get
+            }
+            allow: "*"
+            sort-order: 500
+            name: "puppetlabs crl"
+        },
+        {
+            # Allow nodes to request a new certificate
             match-request: {
                 path: "/puppet-ca/v1/certificate_request"
                 type: path
+                method: [get, put]
             }
             allow-unauthenticated: true
+            sort-order: 500
+            name: "puppetlabs csr"
+        },
+        {
+            match-request: {
+                path: "/puppet/v3/environments"
+                type: path
+                method: get
+            }
+            allow: "*"
+            sort-order: 500
+            name: "puppetlabs environments"
+        },
+        {
+            # Allow nodes to access all file services; this is necessary for
+            # pluginsync, file serving from modules, and file serving from
+            # custom mount points (see fileserver.conf). Note that the `/file`
+            # prefix matches requests to file_metadata, file_content, and
+            # file_bucket_file paths.
+            match-request: {
+                path: "/puppet/v3/file"
+                type: path
+            }
+            allow: "*"
+            sort-order: 500
+            name: "puppetlabs file"
+        },
+        {
+            # Allow nodes to retrieve only their own node definition
+            match-request: {
+                path: "^/puppet/v3/node/([^/]+)$"
+                type: regex
+                method: get
+            }
+            allow: "$1"
+            sort-order: 500
+            name: "puppetlabs node"
+        },
+        {
+            # Allow nodes to store only their own reports
+            match-request: {
+                path: "^/puppet/v3/report/([^/]+)$"
+                type: regex
+                method: put
+            }
+            allow: "$1"
+            sort-order: 500
+            name: "puppetlabs report"
+        },
+        {
+            match-request: {
+                path: "/puppet/v3/status"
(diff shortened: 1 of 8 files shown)


ezbake/config/conf.d/auth.conf
```diff
@@ -1,39 +1,116 @@
 authorization: {
```
nwolfe: It might be a good idea to include some meaningful documentation at the top of this file, like the old one has: https://github.com/puppetlabs/puppet/blob/master/conf/auth.conf#L1-L57

Rather than copy/paste the docs or write special new ones, we might be able to get away with just linking to the official published docs site (which we haven't published yet)?

@nfagerlund got any suggestions here?

ezbake/config/conf.d/auth.conf
```diff
@@ -1,39 +1,116 @@
...
+                method: [get, post]
+            }
+            allow: "$1"
+            sort-order: 500
+            name: "puppetlabs catalog"
+        },
+        {
+            # Allow nodes to retrieve the certificate they requested earlier
+            match-request: {
+                path: "/puppet-ca/v1/certificate/"
```
nwolfe: This rule covers both the CA and everything else.  Since the rule was exactly the same, we don't actually need to represent it twice like the old one does: https://github.com/puppetlabs/puppet/blob/master/conf/auth.conf#L102-L113

ezbake/config/conf.d/auth.conf
```diff
@@ -1,39 +1,116 @@
...
         {
+            # Allow nodes to retrieve their own catalog
+            match-request: {
+                path: "^/puppet/v3/catalog/([^/]+)$"
+                type: regex
+                method: [get, post]
+            }
+            allow: "$1"
+            sort-order: 500
+            name: "puppetlabs catalog"
```
nwolfe: The order of the rules in this file is different that the old order, since we now have the notion of sorting the rules and a warning will be printed out if they're not sorted (on sort-order & name) in the file.

I tried to choose meaningful names for the rules, and then sorted them alphabetically (since they all have the same sort-order 500).

nwolfe:
Ping @ahpook in case you're interested to see the new default `auth.conf`

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
```diff
@@ -53,24 +53,28 @@
...
-              (is (= 403 (:status response))
-                  (ks/pprint-to-string response)))))))))
+         :authorization         {:version 1
+                                 :rules [{:match-request
+                                          {:path "/puppet-ca/v1/certificate"
+                                           :type "path"}
+                                          :allow ["nonlocalhost"]
+                                          :sort-order 1
+                                          :name "cert"}]}}
+        (logutils/with-test-logging
```
camlow325: Could wrap this just around the `http-get` in the "are denied when..." test, right?  Shouldn't expect any messages to be logged for the "are allowed" test?

nwolfe: Sure. FYI there's already a top-level `with-test-logging` that'll swallow those things, but I could push it down further

camlow325: In my experience, the top-level one didn't swallow any of the authorization errors generated at run-time, like the ones as the `http-get` calls are made.  If we push this down into the "are denied..." test, though, we'd avoid swallowing any that might unexpectedly happen for the `http-get` call on the "are allowed..." test, though - which could be good.

nwolfe: That's true. Yeah pushing it down is better.

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
```diff
@@ -79,34 +83,41 @@
...
-              (is (= 200 (:status response))
-                  (ks/pprint-to-string response)))))))))
+         :authorization         {:version 1
+                                 :rules [{:match-request
+                                          {:path "/puppet-ca/v1/certificate"
+                                           :type "path"}
+                                          :allow ["localhost"]
+                                          :sort-order 1
+                                          :name "cert"}]}}
+        (logutils/with-test-logging
```
camlow325: Same comment here on pushing the `with-test-logging` down to just being around the `http-get` for the `denied` test.

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
```diff
@@ -79,34 +83,41 @@
...
+    (bootstrap/with-puppetserver-running
+      app
+      {:authorization {:version 1
+                       :rules [{:match-request
+                                {:path "^/puppet-ca/v1/certificate_status(?:es)?/([^/]+)$"
+                                 :type "regex"}
+                                :allow ["$1"]
+                                :sort-order 1
+                                :name "cert status"}]}}
+      (logutils/with-test-logging
```
camlow325: Same here.

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
nwolfe: Amended in

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
nwolfe: Amended in

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
nwolfe: Amended in

camlow325:
:+1: Did some local functional testing with the new rules - from source - with an agent running against a master where no certs/keys had previously been generated and all worked fine.

@dankreek - whatcha think?

ezbake/config/conf.d/auth.conf
```diff
@@ -1,39 +1,116 @@
...
+            name: "puppetlabs crl"
         },
         {
+            # Allow nodes to request a new certificate
             match-request: {
-                path: "/puppet-ca/v1/certificate/ca"
+                path: "/puppet-ca/v1/certificate_request"
                 type: path
-                method: get
+                method: [get, put]
```
dankreek: It probably doesn't exactly matter right now, but had the TK-Auth that supports this been released yet?

nwolfe: That would definitely matter, wouldn't it? But yes, it has - you'll see the version bump change in `project.clj`

dankreek: Well it would matter if you wanted to run from a package, yeah. But I'm guessing this file doesn't actually get used during development.

dankreek:
:+1: Swingalong sing song.

PR #706, opened 2015-09-24
[FOR REVIEW] (PE-11857) Add `with-lock` function to JRuby service

rlinehan:
Add `with-lock` function to JRuby service. 

This is based on #699 and will need to be rebased once that is merged. https://github.com/rlinehan/puppet-server/commit/434e77e3dbf0e95e678f0b24a38624ff03aa601c is the relevant commit for this PR.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
@@ -57,7 +57,7 @@
                        id count))))
       (catch Exception e
         (.clear pool)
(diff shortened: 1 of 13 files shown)


rlinehan:
Does this function need to do anything with registering events for metrics or anything like that?

cprice404:
@rlinehan I reviewed your last few commits and I think everything looks OK, so AFAICT you'll just need a squash and maybe add a few tests as described above and then I'm +1.

rlinehan:
This is replaced by #710.

PR #707, opened 2015-09-25
(MAINT) simplify dev env setup

cprice404:


documentation/dev_running_from_source.markdown
```diff
@@ -4,6 +4,25 @@ title: "Puppet Server: Running From Source"
...
+
+Before we get started, let's highlight a few gotchas that might save you some time.
+The things that you are most likely to run into trouble with are pretty much the
+same things that were tricky about running the Webrick server from source:
+
+* Knowing where your confdir/vardir/codedir are
+* Knowing what your certname is
+* Making sure that your certname is resolvable
+
+In Puppet Server, confdir/vardir/codedir *always* come from your config.  They
```
KevinCorcoran: I don't think this is true, is it?  Those settings aren't required, we still have defaults in code, I think.  The larger point is good but I think this sentence is a little misleading.

cprice404: OK, I guess I should say we always set these explicitly in the config files we ship if we want to be 100% accurate.

documentation/dev_running_from_source.markdown
```diff
@@ -14,23 +33,62 @@ Step 0: Quick Start for Developers
...
+    $ export RUBYLIB="./ruby/puppet/lib:./ruby/facter/lib"
+    $ export PATH=$PATH:./ruby/puppet/bin:./ruby/facter/
+    # Create the modules directory for the production environment, to avoid the
+    #  annoying module error during pluginsync
+    $ mkdir -p ./target/master-code/environments/production/modules
+    # Run the agent
+    $ puppet agent --no-daemonize --debug --trace --verbose \
+         --confdir ./target/master-conf-dir \
+         --vardir=./target/master-var-dir \
+         --server `facter fqdn` --onetime
```
KevinCorcoran: I'm not too familiar with this stuff, and it's been a while, but last time I recall running the agent from source I used `bundle exec puppet agent ...` and I imagine that is what client-side devs would be more accustomed to ... and I think also what the puppet docs for running from source use.

cprice404: Why?  You don't need any gems to run the agent from source?

cprice404: Also they can file PRs to change that part if they like since they know more about it than we do, but that ^^ is how I've always run it, and I just tested it.  Works On My Machine (tm)

dev/user.clj.sample
```diff
@@ -3,17 +3,33 @@
...
 (defn puppet-server-conf
   []
-  (config/load-config "/Users/cprice/.puppet-server/puppet-server.conf"))
+  (let [conf-file (user-repl/initialize-default-config-file)]
+    (config/load-config conf-file))
+  ;; a better way to handle config for a long-term dev environment is to create
+  ;; your own copy of the dev/puppet-server.conf.sample file somewhere, edit it
+  ;; carefully to reflect your needs, and then load it directly, like this:
+  ;(config/load-config "/Users/cprice/work/puppet-server/conf/puppet-server.conf")
+  )
```
KevinCorcoran: :+1: I am indeed a fan.

Since this is usable OOTB now (I think, right?), whacha think about moving it to just user.clj?

KevinCorcoran: oh, n/m

documentation/dev_running_from_source.markdown
nwolfe: Without the `bundle exec` part, `puppet` will need to be configured on your system path somehow, otherwise where will it find the executable?  Using `bundle exec` requires that you run it from a specific directory now (puppet-server), but it means that you don't need to rely on your global system configuration, which may or may not have `puppet` available.

The `bundle exec` ensures consistency in this case.  You can probably get away without it since AIO and all 3.x compatibility support goodness.

Either way, it's such a common pattern around here I'm sure client-side devs will know what to do.

cprice404:
@KevinCorcoran I added a commit that re-words the part about the magical config.

documentation/dev_running_from_source.markdown
cprice404: I set the path explicitly in this example, like 5 lines above the call to `puppet`...

nwolfe: Sure, and `bundle exec` would just avoid the need for that

cprice404: Do we even have a meaningful bundle setup in the root of the puppet server repo?  Anyway, we can change this later if we want.  I would like to have this code in for the demo today... :/

KevinCorcoran:
Alrighty, :see_no_evil: 

@nwolfe anything else on this before :ship: ?

nwolfe:
:anchor: :arrow_up:

PR #708, opened 2015-09-25
(MAINT) fix bug in user.clj

cprice404:
user.clj is a special namespace and can't reference anything
complicated or anything that requires java compilation directly.


nwolfe:
:tada:

PR #710, opened 2015-09-28
(PE-11857) Add with-lock macro to JRuby Service

rlinehan:
Add `with-lock` macro to JRuby service, which can be used by other services to
acquiring a lock on the JRubyPool, execute the body, then release the lock.

Add tests that `with-lock` actually does lock the write lock from the
JRubyPool, and add basic tests that `borrow-instance` and `return-instance`
affect the lock in the ways we expect. Add a `get-lock` helper function to get
access to the lock for tests.

### DIFF
--- a/test/unit/puppetlabs/services/jruby/jruby_testutils.clj
+++ b/test/unit/puppetlabs/services/jruby/jruby_testutils.clj
@@ -4,6 +4,7 @@
             [puppetlabs.services.jruby.puppet-environments :as puppet-env]
             [puppetlabs.services.jruby.jruby-puppet-schemas :as jruby-schemas]
             [puppetlabs.services.jruby.jruby-puppet-internal :as jruby-internal]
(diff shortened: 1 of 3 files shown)


rlinehan:
This replaces #706.

test/unit/puppetlabs/services/jruby/jruby_testutils.clj
```diff
@@ -174,3 +175,21 @@
...
+                       :JRubyPuppetService
+                       :pool-context
+                       jruby-core/get-pool)
+        f (->> jruby-pool
+               .getClass
+               .getDeclaredFields
+               (filter #(= "lock" (.getName %)))
+               first)]
+    (.setAccessible f true)
+    (.get f jruby-pool)))
```
KevinCorcoran: :scream:

KevinCorcoran:
:see_no_evil: except for the really evil thing.

cprice404:
+1

PR #712, opened 2015-09-28
(MAINT) Remove default "deny all" rule

nwolfe:
We can longer have a "deny all" rule in the default auth.conf as users
need to be able to provide custom rules that come after the default
Puppet Labs rules.


nwolfe:
Ping @camlow325

PR #713, opened 2015-09-28
Bump puppet-agent to c3140803ca769f0f1a43ddaf9c3628e6e6a1315e

dankreek:


camlow325:
:+1: LGTM.  @nwolfe - I think this should all we need to change for this now, right - no ci-job-config bumps to go with it or anything...?

nwolfe:
Nope, nothing in ci-job-configs that I'm aware of.

:+1:

PR #716, opened 2015-09-28
(MAINT) Re-order default "deny all" rule

nwolfe:
The default "deny all" rule needs to come at the very end and not the
end of the Puppet Labs range, otherwise the entire "user space" range
would be rejected.


nwolfe:
TO: @camlow325 @erikPrime 
CC: @ahpook

erikPrime:
Looks good.

ahpook:
:100: ... errr, :nine: :nine: :nine:

camlow325:
:+1:

PR #717, opened 2015-09-29
(SERVER-914) Allow middleware params in jruby requests

cprice404:
Prior to this commit, the JRuby request handler code would
filter out any :params that had been added to the ring request
map by other middleware.  We now have concrete use cases for
this capability, so this commit loosens the restrictions and
allows other middlewares to add to :params, and have those
params still be visible to JRuby.

### DIFF
--- a/test/unit/puppetlabs/services/request_handler/request_handler_core_test.clj
+++ b/test/unit/puppetlabs/services/request_handler/request_handler_core_test.clj
@@ -59,7 +59,8 @@
                              :content-type "text/plain"
                              :query-string "one=1%201&two=2&arr[]=3&arr[]=4"
                              :params       {:bogus ""}})]
-      (is (= {"one" "1 1", "two" "2", "arr[]" ["3", "4"]}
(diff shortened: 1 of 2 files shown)


test/unit/puppetlabs/services/request_handler/request_handler_core_test.clj
```diff
@@ -97,7 +99,7 @@
                              :content-type       "text/plain"
                              :character-encoding "UTF-16"
                              :params             {:bogus ""}})]
-      (is (= {} (:params wrapped-request))
+      (is (= {:bogus ""} (:params wrapped-request))
```
KevinCorcoran: 'tis a bit weird to have `:bogus` as the expected value now.  Perhaps change that key to something else like `:notbogus`.

KevinCorcoran:
:+1: fine with me, @camlow325 I imagine you'll want to have a peek at this?

camlow325:
:+1: Did a couple of agent runs against the branch for good measure and all seemed to work fine.

PR #718, opened 2015-09-29
(MAINT) Update puppet to fix hiera-template test

cprice404:
The last version of puppet that we updated to had some test
failures upstream.  hiera_in_templates, and an mcollective
service test.  These failures should be fixed in the latest
version, which we are updating to here.


KevinCorcoran:
What could go wrong?  :godmode:

PR #719, opened 2015-09-29
(SERVER-846) Acceptance test for default auth.conf

nwolfe:


nwolfe:
FYI @erikPrime

erikPrime:
@nwolfe This looks great to me, but I am having difficulty getting it to run; 
yyd1wdbllj88eh8.delivery.puppetlabs.net (debian6-64-1) executed in 0.01 seconds
Fetching: http://builds.delivery.puppetlabs.net/puppet-agent/c3140803ca769f0f1a43ddaf9c3628e6e6a1315e/repo_configs/deb/pl-puppet-agent-c3140803ca769f0f1a43ddaf9c3628e6e6a1315e-squeeze.list
  and saving to tmp/repo_configs/debian/pl-puppet-agent-c3140803ca769f0f1a43ddaf9c3628e6e6a1315e-squeeze.list
#<OpenURI::HTTPError: 404 Not Found>
...
cceptance/suites/pre_suite/foss/30_install_dev_repos.rb errored in 6.58 seconds

acceptance/suites/tests/authorization/default_rules.rb
```diff
@@ -0,0 +1,163 @@
...
+## and then one of the two assertion functions to validate allowed/denied.
+## The assertion functions look for regexes and status codes in the stdout
+## of the previous curl invocation.
+
+test_name 'Default auth.conf rules'
+
+step 'Turn on new auth support' do
+  modify_tk_config(
+    master,
+    File.join(master['puppetserver-confdir'], 'puppetserver.conf'),
```
camlow325: Where does `puppetserver-confdir` come from?  Is it guaranteed to be where `puppetserver`'s ".../conf.d" directory is created?

acceptance/suites/tests/authorization/default_rules.rb
```diff
@@ -0,0 +1,163 @@
...
+    {'jruby-puppet' => {'use-legacy-auth-conf' => true}})
+end
+
+def curl_authenticated(path)
+  curl = 'curl '
+  curl += '--cert $(puppet config print hostcert) '
+  curl += '--key $(puppet config print hostprivkey) '
+  curl += '--cacert $(puppet config print localcacert) '
+  curl += "--write-out '\\nSTATUSCODE=%{http_code}\\n' "
```
camlow325: Nice - didn't know you could do that.

acceptance/suites/tests/authorization/default_rules.rb
nwolfe: Comes from Beaker proper: https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/install_utils/foss_defaults.rb#L23

acceptance/suites/tests/authorization/default_rules.rb
```diff
@@ -0,0 +1,163 @@
...
+  curl += 'https://$(hostname -f):8140'
+  curl += path
+  on(master, curl)
+end
+
+def curl_unauthenticated(path)
+  curl = 'curl --insecure '
```
camlow325: Could also do this by just having the `--cacert ...` setting and not provide the `--cert` or `--key`.  But seems okay either way for this test since it's not about having the client validate the server.

acceptance/suites/tests/authorization/default_rules.rb
camlow325: Ok, makes sense.  Although the value it is using looks wrong for Puppet Server 2 / AIO.  It has `/etc/puppetserver/conf.d` but in the Puppet Server 2 / AIO world, it should be `/etc/puppetlabs/puppetserver/conf.d`.  Hopefully we're picking up a version of Beaker that will use the right path?

nwolfe: Oh, yeah it gets set to the proper path somewhere. FWIW this is used by `with_puppet_running_on`: https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/helpers/puppet_helpers.rb#L158

nwolfe: We set it here: https://github.com/puppetlabs/puppet-server/blob/master/acceptance/config/beaker/options.rb#L12

camlow325: Cool - looks good.

acceptance/suites/tests/authorization/default_rules.rb
```diff
@@ -0,0 +1,163 @@
...
+    {'jruby-puppet' => {'use-legacy-auth-conf' => true}})
+end
+
+def curl_authenticated(path)
+  curl = 'curl '
+  curl += '--cert $(puppet config print hostcert) '
+  curl += '--key $(puppet config print hostprivkey) '
+  curl += '--cacert $(puppet config print localcacert) '
+  curl += "--write-out '\\nSTATUSCODE=%{http_code}\\n' "
+  curl += 'https://$(hostname -f):8140'
```
camlow325: Is this the hostname of the node that the tests will run on or the hostname of the target master - which are generally different in CI, right?

acceptance/suites/tests/authorization/default_rules.rb
```diff
@@ -0,0 +1,163 @@
...
+
+    curl_unauthenticated('/puppet/v3/environments')
+    assert_denied(/denied by rule 'puppetlabs environments'/)
+  end
+
+  step 'catalog endpoint' do
+    curl_authenticated('/puppet/v3/catalog/$(hostname -f)?environment=production')
+    assert_allowed
+
+    curl_authenticated('/puppet/v3/catalog/notme?environment=production')
```
camlow325: (notit)

acceptance/suites/tests/authorization/default_rules.rb
nwolfe: It's the hostname of the master.  I essentially want this to just be "localhost:8140" since the test only needs a single node and it can curl itself.

acceptance/suites/tests/authorization/default_rules.rb
```diff
@@ -0,0 +1,163 @@
...
+  step 'certificate endpoint' do
+    curl_unauthenticated('/puppet-ca/v1/certificate/ca?environment=production')
+    assert_allowed
+  end
+
+  step 'certificate_request endpoint' do
+    # We'd actually need to store a CSR file on the server in order to get
+    # back a 200, but we know that a 404 means we got past authorization
+    curl_unauthenticated('/puppet-ca/v1/certificate_request/foo?environment=production')
+    assert_allowed(404)
```
camlow325: Might be good to include an example of a `put` here, too, to validate that the auth rule will allow it - since that's part of the rule definition.  Suppose if we did that that we could do a `get` right after that which would be more meaningful as well.

acceptance/suites/tests/authorization/default_rules.rb
```diff
@@ -0,0 +1,163 @@
...
+
+  step 'catalog endpoint' do
+    curl_authenticated('/puppet/v3/catalog/$(hostname -f)?environment=production')
+    assert_allowed
+
+    curl_authenticated('/puppet/v3/catalog/notme?environment=production')
+    assert_denied(/denied by rule 'puppetlabs catalog'/)
+
+    curl_unauthenticated('/puppet/v3/catalog/$(hostname -f)?environment=production')
+    assert_denied(/denied by rule 'puppetlabs catalog'/)
```
camlow325: Might be good to add an example that does a catalog `post` to confirm that it can be successful / fail as expected - since that's what a real-world agent would do and `post` is explicitly part of the rule definition.

camlow325:
:+1: With a couple of suggestions on additional tests that could be added.  Willing to do those later getting this ticket through and since @erikPrime will have some more thorough, real-world kinds of tests that will cover those suggestions later anyway.

@nwolfe - Presumably you agree that the failure @erikPrime was seeing in trying to run these tests is unrelated to the new tests?

nwolfe:
> Presumably you agree that the failure @erikPrime was seeing in trying to run these tests is unrelated to the new tests?

Yeah, not sure what that failure is.  Looks like the puppet-agent package referenced in the error is still available; maybe it was just a transient issue?  Either way don't think that error is related to this PR.

acceptance/suites/tests/authorization/default_rules.rb
nwolfe: Yeah, thought about it but didn't want to spend the time generating a CSR, and I was hesitant to start taking an exhaustive approach like that (Where do we draw the line? How exhaustive should we be?) considering these are meant to be more like smoke tests.

acceptance/suites/tests/authorization/default_rules.rb
nwolfe: Yeah, like I mentioned above, I kind of wanted to do the least amount of work possible in these tests - only enough to verify our new rules are being honored.  I already dislike hard-coding the JSON body to the `report` endpoint, and I think that would require more of the same here.

camlow325:
:shipit:

PR #720, opened 2015-09-30
(SERVER-964) Support multi-webserver config

fpringvaldsen:
Previously, Puppet Server explicitly checked the contents of the
`:webserver` key in determining whether or not it needed to
override webserver settings with those from ruby puppet, meaning
that the settings would always be overridden if a multi-webserver
style configuration is used.

This commit modifies Puppet Server to use the contents of the
configuration at `[:webserver :puppet-server]`, or `:webserver` if
`[:webserver :puppet-server]` does not exist so as to support
multi-webserver configs.

### DIFF
--- a/src/clj/puppetlabs/services/config/puppet_server_config_service.clj
+++ b/src/clj/puppetlabs/services/config/puppet_server_config_service.clj
@@ -26,7 +26,9 @@
           "Initializing with the following settings from core Puppet:\n%s"
           (ks/pprint-to-string puppet-config))
         (core/init-webserver! override-webserver-settings!
-                              (get-in tk-config [:webserver])
+                              (get-in tk-config
+                                      [:webserver :puppet-server]
+                                      (get-in tk-config [:webserver]))
                               puppet-config)
         (assoc context :puppet-config
                        {:puppet-server puppet-config}))))
--- a/test/unit/puppetlabs/services/config/puppet_server_config_service_test.clj
+++ b/test/unit/puppetlabs/services/config/puppet_server_config_service_test.clj
@@ -91,3 +91,27 @@
                                                             :cacert "meow")
                 ; do nothing - bootstrap should throw the exception
                 )))))))
+
+(deftest multi-webserver-setting-override
+  (let [webserver-config {:ssl-cert "thehostcert"
+                          :ssl-key "thehostprivkey"
+                          :ssl-ca-cert "thelocalcacert"
+                          :ssl-crl-path "thecacrl"
+                          :port 8081}]
+    (testing (str "webserver settings not overridden when mult-webserver config is provided"
+                  "and full ssl cert configuration is available")
+      (with-test-logging
+        (tk-testutils/with-app-with-config
+          app
+          service-and-deps
+          (assoc required-config :webserver {:puppet-server webserver-config})
+          (is (logged? #"Not overriding webserver settings with values from core Puppet")))))
+
+    (testing (str "webserver settings not overridden when single webserver is provided"
+                  "and full ssl cert configuration is available")
+      (with-test-logging
+        (tk-testutils/with-app-with-config
+          app
+          service-and-deps
+          (assoc required-config :webserver webserver-config)
+          (is (logged? #"Not overriding webserver settings with values from core Puppet")))))))


src/clj/puppetlabs/services/config/puppet_server_config_service.clj
```diff
@@ -26,7 +26,9 @@
           "Initializing with the following settings from core Puppet:\n%s"
           (ks/pprint-to-string puppet-config))
         (core/init-webserver! override-webserver-settings!
-                              (get-in tk-config [:webserver])
+                              (get-in tk-config
+                                      [:webserver :puppet-server]
+                                      (get-in tk-config [:webserver]))
                               puppet-config)
```
camlow325: I think this is a pretty straightforward way to accomplish this, which is good.  The only thing that bothers me a little bit is that it's not entirely configuration data-driven.  I had originally been envisioning something like:

1) Get the server that the master-service route from web-routes uses.
2) Get the webserver config for that server.

This would allow you to change the server config in web-routes.conf / webserver.conf without having anything in the code know anything about the server name - or lack thereof - directly.

This approach would be more involved, though, and maybe just the idea of looking up a specific service's route from the web-routes.conf - especially from a "common" service like the config one which doesn't manage its own route(s) - would make the approach of directly coding in knowledge of the server name a better option.

Would be curious what @KevinCorcoran or @cprice404 thinks about this, though.

KevinCorcoran: @camlow325 Would that involve hard-coding the fully-qualified `master-service` namespace into this namespace?  That sounds a little gross.

But ... agreed that it would be nice to not hard-code the `:puppet-server` server name here.

camlow325: @KevinCorcoran We had to do something like that for the legacy-routes-service.  For that case, we used `tk-services/get-service` to get the `:MasterService` and then got its route from that.  So we didn't have to hardcode any namespaces (just the service protocol name) - which was good because the master service's namespace for PE is different from OSS.  But it does still couple the config service to the master service, which isn't great either.

Unless others have a better idea, suppose I'm okay with just hardcoding `:puppet-server` for now.

KevinCorcoran: Ah yes, I see:

https://github.com/puppetlabs/puppet-server/blob/master/src/clj/puppetlabs/services/legacy_routes/legacy_routes_service.clj#L26-26

That service should really have a dependency on the master service - otherwise that call is not really safe.

And, if we were to apply that principle to the current situation as well, we'd end up with a circular dependency - the master service already has a dependency on `:PuppetServerConfigService`.

cprice404: I don't immediately have any better ideas and agree with @KevinCorcoran about the circular dependency.  This seems at least better than what we currently have.  It seems like there should be some docs somewhere that we update to reflect this change, though?

fpringvaldsen: There's also the issue of the different configuration styles allowed for webserver configuration, as well as the presence of a default webserver, meaning the user does not always need to explicitly state which webserver they want a route added to. Parsing out the configuration for the webserver that Puppet Server is adding routes to is doable, but it requires a lot of specific knowledge about how the webserver configuration is structured, which makes me think that if we were to do that it should land in tk-webserver-jetty9 as a new service function rather than here in Puppet server. 

Saying that, we'd still have the circular dependency issue, which I don't have a solution for.

cprice404: So, maybe we go with this for now and file a follow-up ticket about coming up with a tk-j9 improvement to try to solve this sort of issue?

fpringvaldsen: @cprice I just created [TK-284](https://tickets.puppetlabs.com/browse/TK-284)

camlow325: I'm good with this as it stands for now.  Separate tk-j9 improvement for later sounds like a good thing, too, although I'm still not sure what an improvement to it which would solve the problem we have here would look like.  Just wish we could get rid of all of this override stuff....

cprice404:
+1 , @camlow325 @KevinCorcoran one of you want to polish this off

camlow325:
Do we think we only need this for the master branch or might we need it for the stable branch as well?  If we think we might need it for both, probably better to retarget at stable, merge it, and merge up.  If we don't think we'll need it for anything but master going forward, though, I'm good with merging it.

cprice404:
There's nothing that I'm aware of that would cause major changes to the arity of the webserver config on the old releases; this is pretty much only a PE issue.  So master seems fine to me.

camlow325:
Works for me.  :+1: and merging then.

@cprice404 - was there more you were hoping we'd do before we close out SERVER-402 as well?

cprice404:
@camlow325 I don't think so, seems like the new ticket that @fpringvaldsen created is the best representation of the current state of the world... probably?

camlow325:
Sounds good to me.  @fpringvaldsen - you want to close our SERVER-402 as a dupe of the ticket you just fixed, then?

fpringvaldsen:
@camlow325 Sure, I'll go ahead and close out that ticket now.

PR #721, opened 2015-09-30
(MAINT) Bump puppet to 6bc15cc

dankreek:


nwolfe:
Looks like this is the wrong (older?) commit considering the version spec test expected `4.2.2` but got `4.2.1`

camlow325:
I'm closing this in favor of #723 since I think it would pick up the Puppet submodule changes that we're looking for on the stable branch.

PR #722, opened 2015-10-01
(MAINT) Fix default_rules acceptance test

nwolfe:
Replace usages of '(hostname -f)' with the Beaker variable 'master'
since the former doesn't always return the FQDN on all platforms.

### DIFF
--- a/acceptance/suites/tests/authorization/default_rules.rb
+++ b/acceptance/suites/tests/authorization/default_rules.rb
@@ -31,16 +31,14 @@ def curl_authenticated(path)
   curl += '--key $(puppet config print hostprivkey) '
   curl += '--cacert $(puppet config print localcacert) '
   curl += "--write-out '\\nSTATUSCODE=%{http_code}\\n' "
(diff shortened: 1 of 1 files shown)


acceptance/suites/tests/authorization/default_rules.rb
```diff
@@ -58,7 +56,8 @@ def assert_denied(expected_stdout)
 def report_query(node)
   curl = "/puppet/v3/report/#{node}?environment=production "
   curl += '-X PUT -H "Content-Type: text/pson" '
-  curl += '--data "{\"host\":\"$(hostname -f)\", \"metrics\":{}, \"logs\":[], \"resource_statuses\":{}}"'
+  curl += '--data "{\"host\":\"' + node
+  curl += '\",\"metrics\":{},\"logs\":[],\"resource_statuses\":{}}"'
```
nwolfe: Escaping sucks.  All I did was s/$(hostname -f)/node and then broke it into two lines just to avoid a long line

camlow325:
:rocket:

PR #726, opened 2015-10-02
(SERVER-915) Enable trapperkeeper-authorization for empty whitelists

camlow325:
This commit changes the 'puppet-admin' and 'certificate-status'
endpoints to use trapperkeeper-authorization instead of the client
whitelists for authorization when the whitelist is present in
configuration but empty.

This commit also has a couple of somewhat unrelated changes:

1) Adds an explicit ring-core 1.4.0 conflict resolution dependency to
   the project.clj.  comidi 0.1.1 pulls in ring-core 1.3.2, which
   conflicts with the ring-core version that
   trapperkeeper-authorization pulls in, 1.4.0.

2) Allows the `X509Certificate` for an `ssl-client-cert` to be optional
   in a Ring request map.  This is needed because Ring will populate
   this key with 'nil' when no SSL client certificate was provided for
   the request.  Found this after adding a test for calling
   'puppet-admin' with 'authorization-required: false'.

### DIFF
--- a/src/clj/puppetlabs/services/ca/certificate_authority_service.clj
+++ b/src/clj/puppetlabs/services/ca/certificate_authority_service.clj
@@ -15,14 +15,33 @@
    [this context]
    (let [path           (get-route this)
          settings       (ca/config->ca-settings (get-config))
-         puppet-version (get-in-config [:puppet-server :puppet-version])]
-     (if (not-empty (:access-control settings))
-       (log/warn
-         "The 'client-whitelist' and 'authorization-required' settings in the"
-         "'certificate-authority.certificate-status' section are deprecated and"
-         "will be removed in a future release.  Remove these settings and create"
-         "an appropriate authorization rule in the"
-         "/etc/puppetlabs/puppetserver/conf.d/auth.conf file."))
+         puppet-version (get-in-config [:puppet-server :puppet-version])
+         certificate-status-access-control (get-in settings
+                                                   [:access-control
+                                                    :certificate-status])
+         certificate-status-whitelist (:client-whitelist
+                                       certificate-status-access-control)]
+     (cond
+       (or (false? (:authorization-required certificate-status-access-control))
+           (not-empty certificate-status-whitelist))
+         (log/warn
+          "The 'client-whitelist' and 'authorization-required' settings in the"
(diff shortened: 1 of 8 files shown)


src/clj/puppetlabs/services/ca/certificate_authority_service.clj
```diff
@@ -15,14 +15,33 @@
...
+          "'certificate-authority.certificate-status' section are deprecated and"
+          "will be removed in a future release.  Remove these settings and create"
+          "an appropriate authorization rule in the"
+          "/etc/puppetlabs/puppetserver/conf.d/auth.conf file.")
+       (not (nil? certificate-status-whitelist))
+         (log/warn
+          "The 'client-whitelist' and 'authorization-required' settings in the"
+          "'certificate-authority.certificate-status' section are deprecated"
+          "and will be removed in a future release.  Because the"
+          "'client-whitelist' is empty and 'authorized-required' is set to"
```
fpringvaldsen: Should be `authorization-required`

src/clj/puppetlabs/services/puppet_admin/puppet_admin_service.clj
```diff
@@ -16,14 +16,27 @@
...
+           "the 'puppet-admin' section are deprecated and will be removed"
+           "in a future release.  Remove these settings and create an"
+           "appropriate authorization rule in the"
+           "/etc/puppetlabs/puppetserver/conf.d/auth.conf file.")
+        (not (nil? client-whitelist))
+          (log/warn
+           "The 'client-whitelist' and 'authorization-required' settings in"
+           "the 'puppet-admin' section are deprecated and will be removed"
+           "in a future release.  Because the 'client-whitelist' is empty"
+           "and 'authorized-required' is set to 'false', the 'puppet-admin'"
```
fpringvaldsen: Same here

fpringvaldsen:
:+1:

camlow325:
Thanks, @fpringvaldsen.  I updated the branch with the two s/authorized-required/authorization-required changes in the log warnings.  Since it was just these two small changes, I just squashed it into the original commit - now as 171d9b3.

dankreek:
:+1: LGTM

PR #728, opened 2015-10-06
(MAINT) Extract puppet server config file name

nwolfe:
This commits moves the name of the config file to options.rb so we can
set it to the PE-specific name in pe-puppet-server-extensions.

### DIFF
--- a/acceptance/suites/tests/authorization/default_rules.rb
+++ b/acceptance/suites/tests/authorization/default_rules.rb
@@ -12,17 +12,13 @@
 test_name 'Default auth.conf rules'
 
 step 'Turn on new auth support' do
-  modify_tk_config(
-    master,
-    File.join(master['puppetserver-confdir'], 'puppetserver.conf'),
-    {'jruby-puppet' => {'use-legacy-auth-conf' => false}})
+  modify_tk_config(master, options['puppetserver-config'],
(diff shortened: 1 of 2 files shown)


nwolfe:
Once this is merged I'll need to do a voom build of puppet-server and then upgrade the dependency in pe-p-s-e

acceptance/suites/tests/authorization/default_rules.rb
```diff
@@ -12,17 +12,13 @@
 test_name 'Default auth.conf rules'
 
 step 'Turn on new auth support' do
-  modify_tk_config(
-    master,
-    File.join(master['puppetserver-confdir'], 'puppetserver.conf'),
-    {'jruby-puppet' => {'use-legacy-auth-conf' => false}})
+  modify_tk_config(master, master['puppetserver-config'],
```
nwolfe: I guess the other code is referencing these through `options['puppetserver-config']`...this works but I just noticed that, should probably change it to be consistent?

erikPrime:
Looks good to me...

acceptance/suites/tests/authorization/default_rules.rb
nwolfe: updated

PR #729, opened 2015-10-06
Tk auth test

erikPrime:
I am still doing substantial work against this PR; It isn't ready to merge.  I would appreciate feedback about the broad direction of this test...


acceptance/scripts/generic/testrun.sh
```diff
@@ -1,10 +1,12 @@
 #!/bin/bash
 set -x
 
-export GEM_SOURCE=http://rubygems.delivery.puppetlabs.net
+# get to the correct directory
+while [ ! -d acceptance ] ; do cd ..; done
```
camlow325: Seems like this could get into an infinite loop depending upon where you are when it starts?  I generally don't like scripts changing my current working directory while they execute.  My preference would be to remove this.

acceptance/scripts/generic/testrun.sh
```diff
@@ -14,9 +16,58 @@ export BEAKER_HELPER="acceptance/lib/helper.rb"
...
+BEAKER="$BEAKER --options-file $BEAKER_OPTIONS"
+BEAKER="$BEAKER --post-suite $BEAKER_POSTSUITE"
+BEAKER="$BEAKER --load-path acceptance/lib"
+
+
+case $1 in
+  -p | --p* )
+  bundle exec genconfig2 $GENCONFIG_LAYOUT > $BEAKER_CONFIG
+    
+  if [ -z "$PACKAGE_BUILD_VERSION" ];
```
camlow325: Does something down below this not already blow up with an intelligible error message if this isn't set?  If not, should we be looking to add that check there instead of here?

acceptance/scripts/generic/testrun.sh
```diff
@@ -14,9 +16,58 @@ export BEAKER_HELPER="acceptance/lib/helper.rb"
...
+  BEAKER="$BEAKER --config hosts_preserved.yml"
+  BEAKER="$BEAKER --preserve-hosts"
+  BEAKER="$BEAKER --tests $BEAKER_TESTSUITE"
+  $BEAKER
+  ;;
+
+
+* )
+  # run it the old way  
+  bundle exec beaker \
```
camlow325: Might be nice to have this build off of `$BEAKER` like the other cases do - to cut down on the repetition a bit.

acceptance/scripts/generic/testrun.sh
erikPrime: Yes, it blows up, but the error message isn't intelligible, and it takes a few minutes to get into the test before you discover that something is wrong.

acceptance/suites/tests/030-tk-auth/010-tk-auth-smoke.rb
```diff
@@ -0,0 +1,146 @@
...
+  step 'Install hocon gem' do
+    #TODO: Is there a way we can calculate this location from puppet config?
+    on(master, '/opt/puppetlabs/puppet/bin/gem install hocon')
+  end
+
+#TODO: Make this look in forge first, and then proceed to git if forge fails.
+  step 'Install puppetlabs-hocon and puppetlabs-http_authorization modules from git' do
+    environmentpath=on(master, 'puppet config print environmentpath').stdout.chomp
+    environment='production'
+    $manifest_path=[environmentpath,environment,'manifests','site.pp'].join('/')
```
camlow325: Wasn't sure why this is named with a `$` at the start?  Also, maybe could use Ruby's `File.join` to put the full path together here - http://ruby-doc.org/core-1.9.3/File.html#method-c-join?

acceptance/suites/tests/030-tk-auth/010-tk-auth-smoke.rb
```diff
@@ -0,0 +1,146 @@
...
+    modulepath=[environmentpath,environment,'modules'].join('/')
+    manifest_text = <<-END
+#vcsrepo { '#{modulepath}/hocon' :
```
camlow325: Any value in keeping the commented out one here?

acceptance/suites/tests/030-tk-auth/010-tk-auth-smoke.rb
```diff
@@ -0,0 +1,146 @@
...
+    create_remote_file(master, $manifest_path, manifest_text)
+    on(master, "chmod +r #{$manifest_path}")
+    on(master, 'puppet agent -t', :acceptable_exit_codes => [0,2])
+  end
+
+  #TODO: file a ticket?
+  #check if version is 9.4 and fail if version isn't so we are forced to fix this 
+  #when it is fixed upstream.
+  step 'WORKAROUND: prerelease puppetlabs-hocon...' do
+    on(master, 'perl -pi -e \'s/"version": "0.9.2"/"version": "0.9.4"/g\' /etc/puppetlabs/code/environments/production/modules/hocon/metadata.json')
```
camlow325: Is this because `puppetlabs-http_authorization` currently depends upon a version of `puppetlabs-hocon` which doesn't exist yet?

acceptance/suites/tests/030-tk-auth/010-tk-auth-smoke.rb
```diff
@@ -0,0 +1,146 @@
...
+  end
+
+
+  step 'WORKAROUND: Disable legacy auth.conf' do
```
camlow325: Is this a WORKAROUND?

acceptance/suites/tests/030-tk-auth/050-tk-auth.rb
```diff
@@ -0,0 +1,216 @@
...
+    if @rule['allow'] != nil then
+      return true if @rule['allow'].include?('*') #TODO: express that -k won't work
+      return true if @rule['allow'].include?('$1') #TODO and its a regex and the cn is signed and the cn is in the url regex thing
+    end
+    return false
+  end
+
+
+  def calculate_expected_result
+    if method_match(@method, @rule['match-request']['method']) &&
```
camlow325: Just an fyi - `method` is not a required field in a rule `match-request`.  We do happen to include it for each rule that we put in our default `auth.conf` file for Puppet Server, though, as you've seen.

acceptance/scripts/generic/testrun.sh
```diff
@@ -2,9 +2,8 @@
 set -x
 
 export GEM_SOURCE=http://rubygems.delivery.puppetlabs.net
-
-export GENCONFIG_LAYOUT="${GENCONFIG_LAYOUT:-redhat6-64ma-debian6-64a-windows2008r2-64a}"
-export BEAKER_TESTSUITE="${BEAKER_TESTSUITE:-acceptance/suites/tests/00_smoke}"
+export GENCONFIG_LAYOUT="${GENCONFIG_LAYOUT:-redhat6-64ma-ubuntu1404-64a-windows2008r2-64a}"
+export BEAKER_TESTSUITE="${BEAKER_TESTSUITE:-acceptance/suites/tests/}"
```
kurtwall: @erikPrime Is everyone on-board with changing debian6 to ubuntu1404?

erikPrime: I have no idea.  I don't remember making this change.
Ultimately, it doesn't really matter because CI controls what OSes get tested.

acceptance/scripts/generic/testrun.sh
kurtwall: +1 to that. I'd like it to fail and bail as early as possible.

acceptance/scripts/generic/testrun.sh
```diff
@@ -14,18 +13,60 @@ export BEAKER_HELPER="acceptance/lib/helper.rb"
...
+  if [ -z "$PACKAGE_BUILD_VERSION" ]; 
+    #TODO: curl builds.puppetlabs.lan/puppetserver.  Parse HTML, find most recent folder name.
+    then echo "$0: PACKAGE_BUILD_VERSION not set.  Please export PACKAGE_BUILD_VERSION."
+    exit -1;
+  fi
+  
+  BEAKER="$BEAKER --pre-suite $BEAKER_PRESUITE"
+  BEAKER="$BEAKER --tests $BEAKER_TESTSUITE"
+  BEAKER="$BEAKER --config $BEAKER_CONFIG"
+  BEAKER="$BEAKER --preserve-hosts"
```
kurtwall: Recommend explicitly stating the ```--preserve-hosts``` argument (```onfail```, ```never```, ```always```, etc).

acceptance/scripts/generic/testrun.sh
```diff
@@ -14,18 +13,60 @@ export BEAKER_HELPER="acceptance/lib/helper.rb"
...
+  $BEAKER
+  echo "Beaker exited with $?"
+  cp log/latest/hosts_preserved.yml .
+  ;;
+
+
+-r | --r* )
+  if [ ! -s ./hosts_preserved.yml ];
+  then echo "$0: Can not find hosts_preserved.yml; can not run without presuite./n \
```
kurtwall: My Ruby stinks, but there seems to be extra or missing characters before the line-continuation: ```...presuite./n \```. Is that a typo or does my Ruby really stink?

acceptance/scripts/generic/testrun.sh
```diff
@@ -14,18 +13,60 @@ export BEAKER_HELPER="acceptance/lib/helper.rb"
...
+    Either put a hosts_preserved.yml or use this script with -p to create new hosts and run the pre-suite against them"
+    exit -1;
+  fi
+  
+  BEAKER="$BEAKER --config hosts_preserved.yml"
+  BEAKER="$BEAKER --preserve-hosts"
```
kurtwall: Recommend explicitly stating the ```--preserve-hosts``` argument (```onfail```, ```never```, ```always```, etc).

acceptance/scripts/generic/testrun.sh
```diff
@@ -14,18 +13,60 @@ export BEAKER_HELPER="acceptance/lib/helper.rb"
...
+* )
+  # run it with the old options.
+  BEAKER="$BEAKER --pre-suite $BEAKER_PRESUITE"
+  BEAKER="$BEAKER --config $BEAKER_CONFIG"
+  BEAKER="$BEAKER --pre-suite $BEAKER_PRESUITE"
+  BEAKER="$BEAKER --tests $BEAKER_TESTSUITE"
+  BEAKER="$BEAKER --preserve-hosts onfail \
   --debug \
-  --timeout 360
+  --timeout 360"
```
kurtwall: I don't understand why this last argument is continued from the previous line when all of the others are built up.

acceptance/lib/puppetserver/acceptance/Auth_Test.rb
```diff
@@ -0,0 +1,119 @@
...
+    # returns true if the rule authorizes the request
+    #TODO: accept a host name and calculate if the rule authorizes the host
+    #we have to handle the case where path contains ([^/]+ and allow => $1 in here.
+    #in this case, we need to seperate
+    # 1.  the path we are going to test
+    # 2.  the path in the rule under test
+    # So,
+    # If cn=hostBob is looking for /puppet/v3/report/hostBob Than we AUTH = True
+    # If cn=SomeOtherHost is looking for /puppet/v3/report/hostBob Than AUTH = FALSE.
+    return true if @rule.keys.include?('allow-unauthenticated') &&
```
camlow325: Will be good with this style of tests to do variations of these like:

- Where `allow-unauthenticated` is set to `true`, should be able to curl with or without an SSL client certificate and still get access to the endpoint
- Where the path is a regex with a capture group and backreference in the allow keyword, request should succeed if the SSL client certificate CN matches the backreference but fail if it doesn't.

camlow325:
Done looking at these for now.  I don't have any major issues with them - although I also didn't try actually running them either.  Will defer to @kurtwall as to when they'd be in good enough shape to run/merge.

acceptance/scripts/generic/testrun.sh
erikPrime: This is bash.

acceptance/scripts/generic/testrun.sh
erikPrime: Mostly because I was trying to preserve the behavior from the original file when no options are passed.

acceptance/scripts/generic/testrun.sh
kurtwall: Is that a typo or does my shell really stink?

kurtwall:
This is a pretty big change to testrun.sh, so I would feel more comfortable merging this after a look by @nwolfe or someone else familiar with it.

kurtwall:
I'm not a big fan of the double newlines. They create breake where there are none.

PR #730, opened 2015-10-07
(SERVER-910) make private keys unreadable by all

dankreek:


src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -141,6 +144,54 @@
...
+  the form of the standard 9 character posix format, ie \"rwxr-xr-x\"."
+  [path :- schema/Str
+   permissions :- schema/Str]
+  (-> (get-path-obj path)
+      (Files/setPosixFilePermissions
+        (PosixFilePermissions/fromString permissions))
+      (.toFile)))
+
+(schema/defn get-file-perms :- schema/Str
+  "Returns the current "
```
camlow325: Incomplete thought?  :smiling_imp:

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -141,6 +144,54 @@
...
+(def p-key-perms
+  "Posix permissions for all private keys on disk."
+  "rw-r-----")
+
+(def p-key-dir-perms
+  "Posix permissions for the private key directory on disk."
+  "rwxr-x---")
+
+(def empty-string-array
+  "The API for Paths/get requires a string array which empty for all of our needs
```
camlow325: ... which *is* empty....?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -141,6 +144,54 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Internal
 
+(def p-key-perms
```
camlow325: Cosmetic - might be good to just call these `private-key...` instead of `p-key...`.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -141,6 +144,54 @@
...
+
+(def empty-string-array
+  "The API for Paths/get requires a string array which empty for all of our needs
+  in this namespace. "
+  (into-array String []))
+
+(defn get-path-obj
+  "Create a Path object from the provided path."
+  [path]
+  (Paths/get path empty-string-array))
```
camlow325: java--

src/clj/puppetlabs/puppetserver/certificate_authority.clj
dankreek: > 'He who is valiant and pure of spirit may find the Holy Grail in the Castle of....

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -557,11 +610,31 @@
...
+      (log/warnf (str "The private key directory at '%s' had permissions set as "
+                      "'%s' and should be set to '%s' this has been corrected")
+                 p-key-dir cur-dir-perms p-key-dir-perms))
+    (doseq [key-file (fs/find-files p-key-dir #".*pem$")]
+      (let [key-path (.getPath key-file)
+            cur-key-perms (get-file-perms key-path)]
+        (when-not (= p-key-perms cur-key-perms)
+          (set-file-perms key-path p-key-perms)
+          (log/warnf (str "The private key located at '%s' had permissions set as "
+                          "'%s' and should be set to '%s' this has been fixed.")
```
dankreek: Note to self: change 'fixed' to 'corrected'

camlow325: Cosmetic - would be good to have as separate sentences like with the ca-file-perms message, like "and should be set to '%s'.  This has been fixed."

camlow325: I guess practically it will be hard for any of this code to be hit since the underlying Ruby initialization code will probably correct the permissions anyway before we get here.  I'm still good with this code being here as a sanity check anyway, though - i.e., not just trusting the Ruby layer to *magically* fix things up.

dankreek: Yeah, these log messages should be more consistent.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -890,6 +965,18 @@
       (when-not (fs/exists? path)
         (ks/mkdirs! path)))))
 
+(schema/defn ensure-ca-file-perms!
+  "Ensure that the CA's private key file has the correct permissions set. If it
+  doesn not, then correct them."
```
camlow325: s/doesn not/doesn't

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -431,7 +444,14 @@
...
                (re-pattern (str "Missing:\n" path))
-               (initialize! settings))))))))
+               (initialize! settings)))))))
+
+  (testing "The CA private key has its permissions properly reset"
+    (let [settings (testutils/ca-sandbox! cadir)]
+      (set-file-perms (:cakey settings) "rw-r--r--")
+      (logutils/with-test-logging
+        (initialize! settings)
+        (is (logged? #"/ca/ca_key.pem' was found to have the wrong permissions set as 'rw-r--r--'. This has been corrected to 'rw-r-----'."))))))
```
camlow325: Would seem good to check the file's permissions as well to see that they were reset, not just check for a log message.

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -1049,3 +1069,30 @@
                         (utils/get-extension-value utils/subject-alt-name-oid)
                         (:dns-name))]
       (is (= #{"puppet" "master"} (set alt-names))))))
+
+(deftest file-permissions
+  (let [tmpdir (System/getProperty "java.io.tmpdir")]
```
camlow325: Can use `raynes.fs/tmpdir` for this instead?

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -1049,3 +1069,30 @@
...
+    (testing "A newly created file contains the properly set permissions"
+      (doseq [u all-perms
+              g no-write-perms
```
camlow325: Not a big deal, but curious why you only did `no-write-perms` instead of `all-perms` for g and o.

dankreek: AHh yes, I probably should have documented this. On OS X, it won't actually allow me to set write perms for group or other.

dankreek: No error occurs, but the write permissions don't stick.

camlow325: My guess is that this is due to a more restrictive default umask causing the attempt to set the write bit for g and o to not be honored for the `Files/createFile` call.  From some text about `Files/createFile` in https://docs.oracle.com/javase/tutorial/essential/io/fileAttr.html#posix:

> Note that the umask also applies, so the new file might be more secure than the permissions that were requested.

I've found that if I turn around and call `Files/setPosixFilesPermissions` on the newly created file with a file permission mask that sets the write bit for g and o that the full mask is honored properly.  I modified the implementation of `create-file-with-perms` to:

~~~~clojure
(schema/defn create-file-with-perms :- File
 ...
  [path :- schema/Str
   permissions :- schema/Str]
  (let [permissions-set (PosixFilePermissions/fromString permissions)
        path-obj (-> (get-path-obj path)
                     (Files/createFile
                      (into-array FileAttribute
                                  [(PosixFilePermissions/asFileAttribute
                                    permissions-set)])))]
    (Files/setPosixFilePermissions path-obj permissions-set)
    (.toFile path-obj)))
~~~~

With the above and `all-perms` used both for g and o in both of the `file-permissions` tests, the tests all pass for me on my OS X laptop.  I think it would be good to add something like this to ensure that the *exact* permissions we're intending to set are applied for the newly created file.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
camlow325: On second thought, I think it might be better to remove the logic from this PR which resets the permissions on the `privatekeydir` and master/agent private key pem files within it.  I had forgotten that the Puppet settings which map to this directory and files can be specified with an explicit mode in the `puppet.conf` file.  From https://docs.puppetlabs.com/references/latest/configuration.html#configuration-settings:

> Settings that take a single file or directory can optionally set the owner, group, and mode for their value: rundir = $vardir/run { owner = puppet, group = puppet, mode = 644 }

If we were to reset these permissions at startup, we could overwrite an alternate mode that the user may have set in their `puppet.conf` file.  To avoid this, I don't think we should try to figure out how to have Puppet Server extract the mode value for Ruby Puppet settings and use that to set permissions.  Since Ruby Puppet code isn't managing the CA private key mode when running under Puppet Server, I do still think it would be good to retain the logic which resets the permissions to 0640 for the CA private key mode at startup.  For the `privatekeydir` and its subordinate pems, though, I now think it might be better to just let Ruby Puppet handle resetting permissions, if needed, at startup as it already does / knows how to do.  From Puppet Server's perspective, then, the only thing we'd additionally need to handle for the `privatekeydir` and subordinate pems is setting world-inaccessible permissions as the directory/files are first generated.

camlow325:
:+1: overall with a few comments, the major ones being:

1) I think it would be good to back out the logic which resets the permissions on the master/agent `privatekeydir` and subordinate PEM files at startup, when needed, since there's a possibility that by adding this that we may break a user's overridden value for the mode in the "puppet.conf" file.  Instead, Ruby Puppet can continue to take care of this.

2) I also think it would be good to explicitly reset permissions for files after they are just created via a call to `create-file-with-perms`, as opposed to just using permissions on the `Files/createFile` call.  This would compensate for the possibly reduced set of permissions first set on the `Files/createFile` call based on the value of the user's umask.

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
dankreek: Yep, we talked about this yesterday. Thanks.

camlow325:
I'm pretty much :+1: on these changes.  Although, I recently discovered that there's a Puppet setting called `manage_internal_file_permissions` that we may want to honor as part of all of this work.  I posted a comment to the JIRA ticket about this - soliciting some others opinions on that.  Would like to see how that conversation unfolds before we merge this but would appreciate reviews on the existing code from others in the meantime.

dankreek:
@camlow325 @nwolfe The `manage-internal-file-permissions` setting is now respected.

camlow325:
:+1: @nwolfe - what do you think?

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -13,7 +14,8 @@
             [clojure.string :as string]
             [clj-time.core :as time]
             [clj-time.coerce :as time-coerce]
-            [me.raynes.fs :as fs]))
+            [me.raynes.fs :as fs]
+            [puppetlabs.puppetserver.certificate-authority :as ca]))
```
nwolfe: This is already brought in above with `:refer :all`

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -1049,3 +1075,30 @@
...
                         (:dns-name))]
       (is (= #{"puppet" "master"} (set alt-names))))))
+
+(deftest file-permissions
+  (let [tmpdir (fs/tmpdir)]
+    (testing "A newly created file contains the properly set permissions"
+      (doseq [u all-perms
+              g all-perms
+              o all-perms]
+        (let [tmp-file (str tmpdir (UUID/randomUUID) (.getTime (Date.)))
```
nwolfe: We can just use `(fs/temp-name tmpdir)` instead of building the temporary name manually.

Can then get rid of the `UUID` and `Date` imports too

test/integration/puppetlabs/puppetserver/bootstrap_int_test.clj
```diff
@@ -1,12 +1,23 @@
...
-      (is (true? true)))
-    (is (true? true))))
\ No newline at end of file
+    (is (true? true))
+    (testing "Private keys have the correct permissions."
+      (let [pk-dir (str bootstrap/master-conf-dir "/ssl/private_keys")
+            pks (fs/find-files pk-dir #".*pem$")]
+        (is (= ca/private-key-dir-perms (ca/get-file-perms pk-dir)))
+        (is (= ca/private-key-perms (ca/get-file-perms
+                               (str bootstrap/master-conf-dir
```
nwolfe: :gun: :gun:

:smoking: 

:skull:

nwolfe:
:+1: other than `fs/temp-name` comment

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
dankreek: I'm going go have to blame IntelliJ....

test/integration/puppetlabs/puppetserver/bootstrap_int_test.clj
dankreek: I think we're allowed to break the McCune principle now....

dankreek: this is weird though.. I'll fix it up

nwolfe:
:+1:

PR #727, opened 2015-10-02
(SERVER-849) Add deprecated features doc to Puppet Server

camlow325:
This commit adds a deprecated features document to Puppet Server.  The
only features (settings) covered in the doc are those related to the
deprecated of legacy authorization systems in favor of
`trapperkeeper-authorization`'s "auth.conf".

PR #732, opened 2015-10-08
Implement tk-auth documentation changes.

gguillotte:
Revise and expand tk-auth predocs. Notable changes:

- Split `configuration.markdown` into separate files.
- Added deprecation notices where applicable.
- ~~Added a `deprecated_settings.md` similar to the Puppet docs to consolidate deprecation references.~~ Linked to the `deprecated_features` document proposed in #727.
- Expanded explanations of `auth.conf` parameters.
- Moved several links to Puppet API documentation from GitHub to the docs site.


documentation/config_file_auth.md
```diff
@@ -0,0 +1,151 @@
...
+# Use GET and POST.
+method: [get, post]
+
+# Use PUT.
+method: put
+~~~
+
+The legacy `auth.conf` format uses indirector names for the `method` value instead of the request's HTTP method:
+
+Indirector | HTTP
```
camlow325: Good idea putting the whole table in - thanks!

documentation/config_file_auth.md
```diff
@@ -0,0 +1,151 @@
+---
+layout: default
+title: "Puppet Server Configuration Files: auth.conf"
+canonical: "/puppetserver/latest/config_file_auth.html"
+---
+
+[`trapperkeeper-authorization`]: https://github.com/puppetlabs/trapperkeeper-authorization
```
camlow325: Have you thought any further about how we want to cover / link to the complete set of settings in the `trapperkeeper-authorization` "auth.conf" file format?  This document covers a subset of those - most of the interesting ones that users converting from the deprecated format to the new format would care about.  For more complete information - including information on the new required settings like `sort-order` and `name` that they'll need to implement - how are we thinking we'll direct users?  I'm definitely onboard with the idea of not linking directly to the "authorization-config.md" doc in the tk-authz github repo.  Do we want to eventually publish that doc to somewhere on docs.puppetlabs.com and link to that from this document?  Open to however you and @nfagerlund think would be best to do this...

documentation/config_file_master.md
```diff
@@ -0,0 +1,57 @@
...
+[new `auth.conf`]: ./conf_file_auth.html
+[Puppet `auth.conf`]: /puppet/latest/reference/config_file_auth.html
+[deprecated]: ./deprecated_features.html
+[`puppetserver.conf`]: ./conf_file_puppetserver.html
+
+The `master.conf` file configures how Puppet Server handles legacy authorization methods for master endpoints. For a broader overview of Puppet Server configuration, see the [configuration documentation](./configuration.html).
+
+> **Deprecation Note:** This file only supports the `allow-header-cert-info` parameter, and is deprecated as of Puppet Server 2.2 in favor of new authorization methods configured in the [new `auth.conf`][] file. Since this configuration file is deprecated, Puppet Server no longer includes a default `master.conf` file in the Puppet Server package.
+
+* In `master.conf`, `allow-header-cert-info` determines whether Puppet Server should use authorization info from the `X-Client-Verify`, `X-Client-CN`, and `X-Client-Cert` HTTP headers. Defaults to `false`.
```
camlow325: All of the references in this doc to `X-Client-CN` should actually be `X-Client-DN`.  Nothing you introduced; we must have missed that in the current docs as well.  The CN is one of the attributes that we extract from the DN.  The expectation, though, is that requests are providing a full DN format for this header value.

documentation/config_file_master.md
```diff
@@ -0,0 +1,57 @@
...
+            sort-order: 500
+            name: "puppetlabs catalog"
+        },
+        ...
+    ]
+}
+~~~
+
+For more information on this configuration file format, see
+[the `trapperkeepr-authorization` documentation](https://github.com/puppetlabs/trapperkeeper-authorization/blob/master/doc/authorization-config.md).
```
camlow325: s/trapperkeepr/trapperkeeper

documentation/config_file_puppetserver.md
```diff
@@ -0,0 +1,93 @@
...
+    #connect-timeout-milliseconds: 120000
+}
+
+# Settings related to profiling the puppet Ruby code.
+profiler: {
+    enabled: true
+}
+~~~
+
+> **Note:** The `puppet-admin` setting and `client-whitelist` parameter are deprecated in favor of new authorization methods provided in Puppet Server 2.2. For details, see the [`auth.conf` documentation][./config_file_auth.html].
```
camlow325: I think it might make sense to change `[./config_file_auth.html]` to just `[]` here?

documentation/config_file_webserver.md
```diff
@@ -0,0 +1,40 @@
+---
```
camlow325: I think the prior documentation in this file was slightly out of date.

In our current Puppet Server 2.x packaging, we split the `webserver` and `web-router-service` sections into two different files - `webserver.conf` and `web-routes.conf`, respectively.

The default content of the `webserver.conf` file is:

~~~~hocon
webserver: {
    access-log-config = /etc/puppetlabs/puppetserver/request-logging.xml
    client-auth = want
    ssl-host = 0.0.0.0
    ssl-port = 8140
}
~~~~

The default content of the `web-routes.conf` file is:

~~~~hocon
web-router-service: {
    # These two should not be modified because the Puppet 4.x agent expects them to
    # be mounted at these specific paths
    "puppetlabs.services.master.master-service/master-service": "/puppet"
    "puppetlabs.services.legacy-routes.legacy-routes-service/legacy-routes-service": ""

    # This controls the mount point for the puppet admin API.
    "puppetlabs.services.puppet-admin.puppet-admin-service/puppet-admin-service": "/puppet-admin-api"
}
~~~~

I'm okay with having this split done as part of this PR or as a follow-on one, whichever you prefer.

documentation/configuration.markdown
```diff
@@ -5,558 +5,42 @@ canonical: "/puppetserver/latest/configuration.html"
...
-#literal string
-path /puppet/v3/report/
-~~~~
-
-For a `trapperkeeper-authorization` "auth.conf" rule, the distinction between
-a regex vs. a literal "starts-with" string is done via the use of an explicit
-`type` attribute in the rule.
-
-For example:
+#### Example (Legacy)
```
camlow325: Would it be better for this example to move over to the new `config_file_ca.html` file?

documentation/configuration.markdown
```diff
@@ -568,10 +52,7 @@ Puppet Server relies on `logrotate` to manage the log file, and installs a confi
 
 ### HTTP Traffic
 
-Puppet Server logs HTTP traffic in a format similar to Apache, and to a separate
-file than the main log file. By default, this is located at
-`/var/log/puppetlabs/puppetserver/puppetserver-access.log` (open source releases) and
-`/var/log/pe-puppetserver/puppetserver-access.log` (Puppet Enterprise).
+Puppet Server logs HTTP traffic in a format similar to Apache, and to a separate file than the main log file. By default, this is located at `/var/log/puppetlabs/puppetserver/puppetserver-access.log` (open source releases) and `/var/log/pe-puppetserver/puppetserver-access.log` (Puppet Enterprise).
```
camlow325: This path is now the same for both OSS Puppet Server 2.x and PE 2015.2.x - `/var/log/puppetlabs/puppetserver/puppetserver-access.log`.

camlow325:
Overall :+1: on these edits.  Great job cleaning up the language, and I really like the breakout of the `configuration.markdown` doc into separate documents.

I left a few comments, mostly related to some text that appeared to be wrong in the original docs, not related to additions for this PR.  Would be nice to clean all of those up on this PR but if you want to do some of those on later PRs, I'd be good with that.

I think it would be good to either provide a link from the new `auth.conf` document over to the detailed settings page doc in `trapperkeeper-authorization` or, even better, absorb all of the docs for those settings onto the docs site somehow when this stuff is published so we don't have to link back to the docs in the trapperkeeper-authorization github repo.  Mostly, I think it will be good for users to get more information about new fields like `sort-order` and `name` which are required for tk-auth rules that did not apply to the legacy Puppet `auth.conf` feature.

Also, we may want to consider porting some of these changes - especially the split up of the configuration sections - back to the docs on the latest `stable` and `1.x` branches as well.  The new trapperkeeper-authorization `auth.conf` information won't apply to the older branches.  Not sure if it is worth doing this.  Will defer to you, @kayayarai, and @nfagerlund on that question.

gguillotte:
Thanks for the feedback, @camlow325. I'll also defer to @nfagerlund on splitting `configuration.markdown` on older versions, but he might take a while to get to it.

Splitting up `configuration.markdown` helps with navigation, but there's enough links from elsewhere that changing all prior versions of Puppet Server to do that might ripple out to an onerous number of changes to the other docs that already point here. (That said, at least links to sections of `configuration.html` won't be _completely_ broken.) 

Splitting `configuration` up also makes it easier to remove the docs for `master.conf` and `ca.conf` when those deprecated files are completely removed, which isn't a concern for past versions.

I'll address the feedback, push, and notify you, at which point I'll also be :+1:. After talking to @kayayarai, I think we're good with merging this PR with your approval, and then Kari will branch and edit the prose in a separate ticket for a follow-up PR.

documentation/config_file_auth.md
gguillotte: I do think the parts of `trapperkeeper-authorization` that are _relevant to Puppet Server_ should be on our docs site, and documented in the context of a Puppet Server install. If there's more settings required in `auth.conf`, for instance, we should document them in this `config_file_auth` document. 

If that means we need to add `sort-order` and `name` in addition to the changes in this PR, I can add those in a follow-up PR.

If there's any optional parts of `trapperkeeper-authorization` that _can_ work with Puppet Server but aren't _required_, I'm OK with linking to GitHub _or_ adding an "advanced options" section to the appropriate configuration file documents, as long as we emphasize that those options are not required.

documentation/config_file_puppetserver.md
gguillotte: Actually, I did a few things wrong here: you're right that I could use the reference link, but the reference link at the top points to the wrong place (`conf_file_auth` sted `config_file_auth`). And those should've been parens instead of brackets for a local link anyway.

gguillotte:
@camlow325 Implemented feedback.

documentation/config_file_auth.md
camlow325: @gguillotte Okay, that makes sense.

We do cover most of the required elements in this section.  `sort-order` and `name` are the major ones that are missing.  We don't touch on the `version` setting underneath the `authorization` section either, which is required but always expected to be "1" for now.

Also, we don't explicitly touch on the `allow` or `deny` settings in this doc.  `deny` is probably more of a niche case but `allow` would definitely be worth covering - especially with respect to the extra level of detail around the forms that an `allow` entry can take on.  One of `allow`, `deny`, or `allow-unauthenticated` is required at a minimum for each rule; `allow` and `deny` can also be used together for the same rule.

I'm open to you covering these others on a separate PR if that's how you'd prefer to do it.

camlow325:
Also, I think this may need to be rebased again because of a merge conflict, sorry.  :frowning: 

I hope the merge conflict resolution would be a little easier this time - might just be "one space" change to "configuration.markdown" which was recently merged in as part of #727.

camlow325:
@gguillotte I'm good with your latest updates on the PR.

Assuming you would want to defer the additional work to add documentation on some of the other rule fields - `sort-order`, `name`, `allow` - as a later PR, I'd be good with merging this.  We would still need to have you rebase/resolve the current merge conflicts so that we can push the Green button.  Sound okay?

Regarding the work to add more documentation for the rule settings in the `auth.conf` document, I'd be happy to take an initial swing at that, too, if you'd like.  Let me know.  Thanks!

documentation/config_file_webserver.md
```diff
@@ -0,0 +1,42 @@
...
+layout: default
+title: "Puppet Server Configuration Files: webserver.conf"
+canonical: "/puppetserver/latest/config_file_webserver.html"
+---
+
+The `webserver.conf` file configures the Puppet Server `webserver` service. For a broader overview of Puppet Server configuration, see the [configuration documentation](./configuration.html). To configure the mount points for the Puppet administrative AI web applications, see the [`web-routes.conf` documentation](./config_file_web-routes.md).
```
nwolfe: s/AI/API

documentation/config_file_webserver.md
```diff
@@ -0,0 +1,42 @@
...
+    ...
+    ssl-cert    : /path/to/master.pem
+    ssl-key     : /path/to/master.key
+    ssl-ca-cert : /path/to/ca_bundle.pem
+    ssl-cert-chain : /path/to/ca_bundle.pem
+    ssl-crl-path : /etc/puppetlabs/puppet/ssl/crl.pem
+}
+~~~
+
+Configuring an external CA requires additional steps, which are described in the  [External CA Configuration documentation](./external_ca_configuration.markdown).
```
nwolfe: s/the  [/the [

gguillotte:
Rebased, and fixed more links to the individual configuration file docs.

camlow325:
:+1:

PR #733, opened 2015-10-13
(SERVER-972) Register dummy AuthConfigLoader when not using legacy auth

camlow325:
This commit registers a dummy AuthConfigLoader with core Ruby Puppet
when Puppet Server is configured with a `use-legacy-auth-conf` of
`false`.  This is done to avoid having the auth.conf be parsed (and
corresponding messages logged) by core Ruby Puppet.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -30,7 +30,7 @@ module PuppetServerExtensions
     puppet_build_version = get_option_value(options[:puppet_build_version],
                          nil, "Puppet Development Build Version",
                          "PUPPET_BUILD_VERSION",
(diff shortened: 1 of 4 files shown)


camlow325:
This will fail Travis CI because it is dependent upon some changes that still need to land in core Puppet - https://github.com/puppetlabs/puppet/pull/4334.

This PR will need to have the Puppet submodule bumped up to a newer version before it can be merged.

camlow325:
I just bumped the puppet-agent and puppet submodule dependencies to pick up the work merged from https://github.com/puppetlabs/puppet/pull/4334.  This PR is ready to review for merge now.

nwolfe:
:+1: easy

cprice404:
:see_no_evil:

PR #739, opened 2015-10-15
(doc) Update Puppet Server 2.x release notes.

gguillotte:
Add release notes for Puppet Server 2.2.0 and 2.1.2.

### DIFF
--- a/documentation/release_notes.markdown
+++ b/documentation/release_notes.markdown
@@ -10,11 +10,100 @@ canonical: "/puppetserver/latest/release_notes.html"
 [configuration directory]: /puppet/latest/reference/dirs_confdir.html
 [code directory]: /puppet/latest/reference/dirs_codedir.html
 [cache directory]: /puppet/latest/reference/dirs_vardir.html
+[`trapperkeeper-authorization`]: https://github.com/puppetlabs/trapperkeeper-authorization
+[Configuration documentation]: ./configuration.html
+[deprecated]: ./deprecated_features.html
+[Puppet `auth.conf` file]: /puppet/latest/reference/config_file_auth.html
+[`auth.conf`]: ./config_file_auth.html
+[`ca.conf`]: ./config_file_ca.html
+[`master.conf`]: ./config_file_master.html
 
+## Puppet Server 2.2
+
+Released October 15, 2015.
+
+### Supported Platforms
+
+Puppet Server 2.2 supports the following platforms:
+
+* Enterprise Linux 7
+* Enterprise Linux 6
+* Ubuntu 14.04
+* Ubuntu 12.04
+* Debian 7
+
+### What's New
+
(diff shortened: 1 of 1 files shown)


documentation/release_notes.markdown
```diff
@@ -10,11 +10,100 @@ canonical: "/puppetserver/latest/release_notes.html"
...
+
+### Supported Platforms
+
+Puppet Server 2.2 supports the following platforms:
+
+* Enterprise Linux 7
+* Enterprise Linux 6
+* Ubuntu 14.04
+* Ubuntu 12.04
+* Debian 7
```
gguillotte: Can someone confirm this list of supported platforms?

camlow325: @gguillotte Those look right.  We're also adding "Debian 8" for this release.

documentation/release_notes.markdown
```diff
@@ -10,11 +10,100 @@ canonical: "/puppetserver/latest/release_notes.html"
...
+### Bug Fixes
+
+#### Make Certificate Authority and Master Private Keys Inaccessible to "World" Users
+
+Previous versions of Puppet Server would not explicitly set file permissions for certificate authority (CA) and Master private keys, which could leave both keys' readable by "world" users. Puppet Server 2.2 resolves this bug by automatically setting and enforcing a permissions change that limits read access to the Puppet Server user and group.
+
+This fix was also applied to [Puppet Server 2.1.x](#puppet-server-212) and [1.2.x](#puppet-server-122).
+
+* [SERVER-910](https://tickets.puppetlabs.com/browse/SERVER-910)
+
```
camlow325: The fixes for SERVER-813 and SERVER-813 are also going into the 2.2.0 release, so probably would be good to mention those here.

camlow325:
:+1: These look good to me.  Just left the one note about including a couple of the other fixes in the description for the 2.2.0 release.

I noticed that this PR is targeted at the puppet-server master branch but includes a roll-up of fixes for both the 2.1.2 and 2.2.0 release.  Not sure how the doc source branches are pinned right now for Puppet Server, but I was thinking we were going to want to have the release notes for the 2.1.2 release targeted at the puppet-server 'stable' branch vs. the ones for 2.2.0 (which could include the 2.1.2 ones for historical reference) targeted at the 'master' branch.  Does that match your understanding?

In terms of timing, we're expecting 2.1.2 to go out on Monday, October 19th.  The release date for 2.2.0 is a bit in flux right now, though.  It should be sometime in early November, but we don't have an exact date yet because we're dependent upon puppet-agent 1.3.0 going out first.

Maybe it makes most sense to just close this PR for now and put up one with just the 2.1.2 release notes in it, targeted at the stable branch.  We could then merge that as we do the release this Monday, hopefully.  Then a new PR could be put back up sometime in early November against the master branch for the 2.2.0 notes.  Does that make sense?

PR #740, opened 2015-10-15
(doc) Add 1.1.2 release notes.

gguillotte:
Add Puppet Server 1.1.2 release notes.

### DIFF
--- a/documentation/release_notes.markdown
+++ b/documentation/release_notes.markdown
@@ -4,6 +4,43 @@ title: "Puppet Server: Release Notes"
 canonical: "/puppetserver/latest/release_notes.html"
 ---
 
+## Puppet Server 1.1.2
+
+Released October 19, 2015.
+
+This is a security and bug fix release in the Puppet Server 1.1 series; no new
+features have been added since 1.1.0. We recommend that all users upgrade.
+
+### Bug Fixes
+
+#### Make Certificate Authority and Master Private Keys Inaccessible to "World" Users
+
+Previous versions of Puppet Server would not explicitly set file permissions for certificate authority (CA) and Master private keys, which could leave both keys' readable by "world" users. Puppet Server 2.2 resolves this bug by automatically setting and enforcing a permissions change that limits read access to the Puppet Server user and group.
+
+This fix was also applied to [Puppet Server 2.1.x](#puppet-server-212) and [1.2.x](#puppet-server-122).
+
+* [SERVER-910](https://tickets.puppetlabs.com/browse/SERVER-910)
+
+#### Resolve an Issue with FreeIPA 4.x CA Signed Certificates
+
+When running Puppet agent against a Puppet Server with a FreeIPA 4.x CA signed certificate that contains a `General Name` of type `other`, Puppet agent fails with an error:
+
+    [puppet-server] Puppet java.util.ArrayList cannot be cast to java.lang.String
+
+Puppet Server 1.1.2 resolves this issue.
+
+* [SERVER-816](https://tickets.puppetlabs.com/browse/SERVER-816)
+
+#### Correctly Flush Caches of Actively Used JRuby Instances
+
+In previous versions of Puppet Server, Puppet Server could not flush the caches of JRuby instances if they were borrowed from a pool when the cache flush request was issued. This version resolves the issue.
+
+* [SERVER-813](https://tickets.puppetlabs.com/browse/SERVER-813)
+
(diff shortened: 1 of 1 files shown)


camlow325:
:+1: on the content - looks good.

We decided to hold back the release until Monday, October 19th - so would appreciate it if you could update the release date in the notes.

Also, I presume you want us to hold off from merging this until right before we do the release so that this doesn't get published prematurely?  At least, that's what @nfagerlund has had us do in the past.

nfagerlund:
There's kind of two options. 

* Hold off on merging it, and merge on release. 
* Go ahead and merge it, lock the wiring in puppet-docs/source/_config.yml to a commit sha prior to the merge, and file a PR to puppet-docs that reverts the lock commit, then merge THAT when it's release time. 

@gguillotte, I'm cool with either of those, so it's down to what you'd rather do. The benefit of the second one is that you can handle multiple releases at once with one PR.

documentation/release_notes.markdown
```diff
@@ -4,6 +4,43 @@ title: "Puppet Server: Release Notes"
...
+
+Previous versions of Puppet Server would not explicitly set file permissions for certificate authority (CA) and Master private keys, which could leave both keys' readable by "world" users. Puppet Server 2.2 resolves this bug by automatically setting and enforcing a permissions change that limits read access to the Puppet Server user and group.
+
+This fix was also applied to [Puppet Server 2.1.x](#puppet-server-212) and [1.2.x](#puppet-server-122).
+
+* [SERVER-910](https://tickets.puppetlabs.com/browse/SERVER-910)
+
+#### Resolve an Issue with FreeIPA 4.x CA Signed Certificates
+
+When running Puppet agent against a Puppet Server with a FreeIPA 4.x CA signed certificate, Puppet agent fails with an error:
```
dankreek: This issue was actually specific to X.509 `General Name` extensions.  The bug was triggered when a user had a cert that contained a `General Name` of type `other`.  In the case of that user's FreeIPA (:beer:) server, it was generating certs with kerberos and some Microsoft style names which include this type `General Name` extension.

gguillotte: Amended commit to try to clarify this.

camlow325:
I'm good with this whenever @dankreek is.  @gguillotte - when do you want us to merge this?  Should we wait until right around when the packages have gone live to the public or just whenever we're good on the review?

gguillotte:
@camlow325 I can push whenever the packages are live.

camlow325:
We've been told packages are live now, so merging.

documentation/release_notes.markdown
```diff
@@ -4,6 +4,43 @@ title: "Puppet Server: Release Notes"
...
+This is a security and bug fix release in the Puppet Server 1.1 series; no new
+features have been added since 1.1.0. We recommend that all users upgrade.
+
+### Bug Fixes
+
+#### Make Certificate Authority and Master Private Keys Inaccessible to "World" Users
+
+Previous versions of Puppet Server would not explicitly set file permissions for certificate authority (CA) and Master private keys, which could leave both keys' readable by "world" users. Puppet Server 2.2 resolves this bug by automatically setting and enforcing a permissions change that limits read access to the Puppet Server user and group.
+
+This fix was also applied to [Puppet Server 2.1.x](#puppet-server-212) and [1.2.x](#puppet-server-122).
```
camlow325: @gguillotte Sorry for the belated comment here, but the references in this sentence don't look right to me.  We don't have plans yet for a 1.2.x release, so I think that part of this should be removed.  (We will include this fix in the 2.2.0 release later, but I'm not sure if you want to mention that here yet?)

Also, the link for `#puppet-server-212` seems like it might not be right.  When I click on it from within http://docs.puppetlabs.com/puppetserver/1.1/release_notes.html#puppet-server-112, my browser tries to go to http://docs.puppetlabs.com/puppetserver/1.1/release_notes.html#puppet-server-212, which doesn't exist.  Instead, I think it should be going to http://docs.puppetlabs.com/puppetserver/2.1/release_notes.html#puppet-server-212.

Since I merged this already, we'd need to have another PR put up to address this.  Let me know if you want one of us to do this or if you have it covered.  Thanks!

gguillotte: I'm in meetings/trainings for the next 6 hours so if it's easier for you to do it, go for it. Otherwise I'll swing back around to this ~4 p.m. Pacific today.

PR #741, opened 2015-10-15
(MAINT) Delete `upgrade_package` method from helper.rb

camlow325:
This commit deletes the `upgrade_package` method from the helper.rb file.
With it present, the call to `rm -f` that it made for EL-7 systems was
failing because it wasn't handling directories to be removed instead of
files.  By removing this, the test helpers should now be able to use the
beaker-provided `upgrade_package` function instead.


nwolfe:
:+1:

PR #751, opened 2015-10-21
(MAINT) Remove facts and facts_search endpoints from routing

camlow325:
This commit removes the facts and facts_search endpoints from the master
service routing layer.  These endpoints are no longer supported as of
Puppet 4 and so are no longer necessary to route through from Puppet
Server.


nwolfe:
:+1: don't see any references to these endpoints in puppet-server or pe-p-s-e

PR #752, opened 2015-10-21
(MAINT) Allow unauthenticated requests to /status

nwolfe:
This commit changes the default auth.conf to allow unauthenticated
requests to the /puppet/v3/status endpoint. This same change will be
made to the auth.conf in pe-puppetserver.

### DIFF
--- a/acceptance/suites/tests/authorization/default_rules.rb
+++ b/acceptance/suites/tests/authorization/default_rules.rb
@@ -129,11 +129,8 @@ with_puppet_running_on(master, {}) do
   end
 
   step 'status endpoint' do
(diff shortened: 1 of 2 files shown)


nwolfe:
This change came out of the discussion here: https://github.com/puppetlabs/puppetlabs-puppet_enterprise/pull/635#discussion_r42445704

nwolfe:
We're making this endpoint allow-unauthenticated in the PE module, so we were thinking it should be open in FOSS as well.

nwolfe:
FYI @camlow325 @ahpook

camlow325:
:+1:

PR #755, opened 2015-10-28
(PDB-2071) Add PuppetDBTerminusState class

mullr:
In order to support high availability, the PuppetDB terminus needs
to implement sticky failover. That is, after failing over to another
PuppetDB instance, it should keep using that instance for future
requests as long as possible. This is tricky when running in
PuppetServer, since it hosts JRuby instances. 

This adds the PuppetDBTerminusState java class as a place for the
PuppetDB terminus to put the data it needs to share between all
JRubies in the process.

### DIFF
--- a/src/java/com/puppetlabs/puppetserver/PuppetDBTerminusState.java
+++ b/src/java/com/puppetlabs/puppetserver/PuppetDBTerminusState.java
@@ -1,0 +1,12 @@
+package com.puppetlabs.puppetserver;
+
+import clojure.lang.Atom;
+
+/**
+  The PuppetDB terminus sometimes has occasion to store data that should be
+  shared between all JRuby instances. This class provides a place to put it.
+ */
+public final class PuppetDBTerminusState {
+    private static final Atom sharedStateAtom = new Atom(null);
+    public static final Atom getSharedStateAtom() { return sharedStateAtom; }
+}


mullr:
The corresponding commit in PuppetDB: https://github.com/mullr/puppetdb/commit/f5363832ba591ffc34e37a4496df915352efc238

kbarber:
This is obviously tickling my 'action-at-a-distance' funny bone. I don't suppose we have any thoughts about making this kind of thing more generic and less pdb-specific?

mullr:
@kbarber Yeah, I know what you mean. I'm of two minds here; while it would be nice to make this more generic (a map in an atom?), I also don't know of anybody else who needs such a thing.

mullr:
@cprice404 : we talked about this in standup and agreed (I think) that this is ok given that we're going to a Clojure terminus at some point. @kbarber are we good here?

cprice404:
@mullr OK; we probably aren't going to be able to do anything with this for a little while, we're blocking Ankeny right now and we don't have a Burnside branch yet.  Will review ASAP.

cprice404:
Sorry it's taken so long to get to this.

I have a handful of comments, but I don't know that any of them are things I'd consider mandatory for this PR.  I probably have more comments for the terminus-side code, but I'll put those on that PR/commit.

* I kind of do like the idea of trying to make this more generic, and maybe have just a `TerminusContext` class or something.  It could be a map whose keys map to individual terminii, and then they reserve the right to decide what the values are and use them as they see fit.
* This may be a little over-protective, but I have an instinct to be really careful about muddying the waters between languages whenever we have to do cross-language interop stuff like this.  For that reason, we've tried pretty hard to make sure we do not introduce any dependencies on Clojure in the JRuby code.  This PR would break that paradigm, and it's not clear to me that it's necessary.  It seems like a Java AtomicInteger would be sufficient here, and that there is no big advantage to introducing a dependency on the Clojure runtime to the Ruby side.
* For all of the patterns where we've done this sort of thing in the past, we inject things into the JRuby world from the Clojure side, rather than reaching out from the Ruby side into the JVM.  I think that would be doable and desirable here as well.  Relevant code (there are a couple of different approaches in play in these examples):

https://github.com/puppetlabs/puppet-server/blob/573e915979dc0ade9cc82cc50d6987410954f40f/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj#L207-L212

https://github.com/puppetlabs/puppet-server/blob/573e915979dc0ade9cc82cc50d6987410954f40f/src/ruby/puppet-server-lib/puppet/server/master.rb#L28-L29

And perhaps the most relevant:

https://github.com/puppetlabs/puppet-server/blob/573e915979dc0ade9cc82cc50d6987410954f40f/src/ruby/puppet-server-lib/puppet/server/config.rb#L29

If you did something like this last example, to override the default class used to store your state, then you could have a default class implemented in pure ruby in your terminus, and puppet server would override it for the case where you are being run under puppet server.  I think that would clean up a lot of that stuff you're currently doing with NameError, and also result in a much less tangled dependency graph between the various functional units of code.

I'll try to add some comments to the terminus pr to clarify, and happy to discuss in person if that would help (though I'm out all of next week).

cprice404:
@mullr going through our PR backlog... any new thoughts on this?  Is it blocking you?

I think I added some comments on the terminus PR as well.

Would it be easier if we set up some time to round up a few people and discuss?

mullr:
@cprice404 ack, sorry, this is a simple case of neglect. I'll dive back into this later today or early tomorrow.

mullr:
> I kind of do like the idea of trying to make this more generic, and maybe have just a TerminusContext class or something. It could be a map whose keys map to individual terminii, and then they reserve the right to decide what the values are and use them as they see fit.

I thought about that. The tricky bit is deciding what kind of map to use, and which language's semantics it needs to obey. I'm sure we can answer this question, but I didn't want to go into it just for this. I would totally understand if that's what's needed though, and am willing to rewrite to make it so. 

> This may be a little over-protective, but I have an instinct to be really careful about muddying the waters between languages whenever we have to do cross-language interop stuff like this. For that reason, we've tried pretty hard to make sure we do not introduce any dependencies on Clojure in the JRuby code. This PR would break that paradigm, and it's not clear to me that it's necessary. It seems like a Java AtomicInteger would be sufficient here, and that there is no big advantage to introducing a dependency on the Clojure runtime to the Ruby side.

Totally agree on the dependency direction smell. I took this approach because I wanted to avoid bleeding the semantic concerns of the pdb terminus into puppet server. That is, the thing PuppetDB really wants in order to implement this functionality is synchronized place to store some chunk of data. That is exactly a Clojure atom. 

It is true that the data is an Integer, so it could be implemented with AtomicInteger. But my feeling is that spreading this implementation detail across two code bases is more harmful than introducing a terminus -> clojure runtime dependency. (the dependency is soft, fwiw; it'll store the state in the ruby vm if it can't find the java stuff)

> Relevant code (there are a couple of different approaches in play in these examples):

The big difference between these samples and the requirement here is that we need a storage location at the PuppetServer level. The `static` implementation is really a hack; it would be better to have an atom whose lifetime follows that of PuppetServer (getting cleared on HUP) and is passed in to the terminus. The http client class business is kind of like that, but it appears to be stored in a static location rather than with the app itself.

mullr:
@cprice404 ⬆️

cprice404:
> Totally agree on the dependency direction smell. I took this approach because I wanted to avoid bleeding the semantic concerns of the pdb terminus into puppet server. That is, the thing PuppetDB really wants in order to implement this functionality is synchronized place to store some chunk of data. That is exactly a Clojure atom. 

But there are other classes in `java.util.concurrent` that would satisfy your requirements without introducing the JRuby->Clojure dependency, right?

cprice404:
@mullr after our meatspace convo the other day, would it make sense for us to go ahead and close this out and wait for your follow-up PR?

Also please let us know if you need a hand with the other PR, I know it kind of gets into the gross Ruby guts of Puppet Server, and wasn't trying to totally punt that over to you :)

mullr:
@cprice404 yep, closing this. Will submit a new PR.

cprice404:
@mullr thanks!

PR #758, opened 2015-11-03
Merge up from stable 2015-11-3

camlow325:


camlow325:
Merge conflicts were mostly related to versions of things from stable - like the puppet-agent version it uses vs. the one master uses.  I deferred to what was already being used in master in most places.  I did let the sqa-utils bump come through, since we want that one specifically.

nwolfe:
:+1:

documentation/release_notes.markdown
```diff
@@ -10,11 +10,60 @@ canonical: "/puppetserver/latest/release_notes.html"
 [configuration directory]: /puppet/latest/reference/dirs_confdir.html
 [code directory]: /puppet/latest/reference/dirs_codedir.html
 [cache directory]: /puppet/latest/reference/dirs_vardir.html
+[`trapperkeeper-authorization`]: https://github.com/puppetlabs/trapperkeeper-authorization
```
cprice404: We have stuff about tk-auth in stable?

camlow325: No.  I think this might have been an error from the release notes writeup.  Nothing references this link in the docs.  /CC @gguillotte.

cprice404:
+1 this all looks fine, not sure what (if anything) should be done about that tk-auth link existing in `stable`; it doesn't seem like it should be there.

camlow325:
Submitted #759 to nuke the tk-auth link from the release_notes.markdown doc.

PR #760, opened 2015-11-04
(MAINT) Cache AuthConfig to reduce logging

nwolfe:
To avoid excessive logging, this commit moves the log statement into the
object constructor and then caches the object from the AuthConfigLoader
class. The log statement is also lowered from INFO level to DEBUG.


nwolfe:
Tested on a live PE installation.

Ping @camlow325

nwolfe:
> Tested on a live PE installation.

Customer was thrilled.

camlow325:
:ship: Way to make that customer happy!

PR #761, opened 2015-11-04
(MAINT) Bump Puppet submodule for upstart vivid test fix

camlow325:
This commit bumps the Ruby Puppet submodule up in order to pick up a
Puppet acceptance test "fix" which skips execution of the handle_upstart
test case when running on later systemd-based Ubuntus.


nwolfe:
:+1:

PR #764, opened 2015-11-09
(MAINT) Remove 'finish' assertion from the foreground acceptance test

camlow325:
This commit removes an assertion that the "Finished shutdown sequence"
message appears in the output of the puppetserver foreground test.  This
message has been seen to sometimes not appear when puppetserver is shut
down - because an OS may kill the process early if puppetserver's
shutdown hook takes too long to run.  For the OSes where this is
problematic - apparently RedHat-based ones but maybe not Debian-based
ones? - there doesn't appear to be any way to prolong the wait time
when running via "timeout" or pressing CTRL-C in the foreground.

### DIFF
--- a/acceptance/suites/tests/010-puppetserver-cli/subcommand/foreground.rb
+++ b/acceptance/suites/tests/010-puppetserver-cli/subcommand/foreground.rb
@@ -17,7 +17,7 @@ expected_messages = {
   /Initializing the JRuby service/ => "JRuby didn't initialize",
   /Starting web server/ => "Expected web server to start",
   /Puppet Server has successfully started and is now ready to handle requests/ => "puppetserver never finished starting",
-  /Finished shutdown sequence/ => "Test ended without puppetserver shutting down gracefully"
+  /Beginning shutdown sequence/ => "Test ended without puppetserver triggering shutdown"
 }
 
 # Start of test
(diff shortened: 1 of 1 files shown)


acceptance/suites/tests/010-puppetserver-cli/subcommand/foreground.rb
```diff
@@ -17,7 +17,7 @@
   /Initializing the JRuby service/ => "JRuby didn't initialize",
   /Starting web server/ => "Expected web server to start",
   /Puppet Server has successfully started and is now ready to handle requests/ => "puppetserver never finished starting",
-  /Finished shutdown sequence/ => "Test ended without puppetserver shutting down gracefully"
+  /Beginning shutdown sequence/ => "Test ended without puppetserver triggering shutdown"
```
camlow325: Figured we could at least keep an assertion to see that the shutdown hook is getting invoked even though we're removing the check for "finished".

camlow325:
As a follow-up for this, I added a task to create a replacement "graceful shutdown" test that would use the service framework rather than `puppetserver foreground`.  This would seem to be a much more useful test for validating a production workflow for puppetserver than `puppetserver foreground`, which is typically only expected to be used for troubleshooting.  That ticket is https://tickets.puppetlabs.com/browse/SERVER-997.

nwolfe:
:+1:

PR #767, opened 2015-11-11
(MAINT) Bump puppet submodule to 28a10df

camlow325:
This commit bumps the Puppet submodule up to version 28a10df and
corresponding puppet-agent dependency up to d5a07cfe7.


nwolfe:
:metal:

PR #771, opened 2015-11-14
(MAINT) Upgrade lein-ezbake

nwolfe:
Utopic has reached EOL and we no longer build packages for it, yet our
version of ezbake was old enough that it still attempted to build
packages for it and would fail.

This commit bumps us up to a newer version of lein-ezbake which will
allow for local ezbake builds to work properly. Jenkins CI isn't failing
because we override the platforms in ci-job-configs and thus don't rely
on the default platforms defined in lein-ezbake.


nwolfe:
@jb

nwolfe:
Hmm...never done this before but I just closed my previous PR against master and re-opened it against stable.  My branch was forked off master though...not sure if that's going to be problematic for merges in the future or something.

camlow325:
:+1:

PR #773, opened 2015-11-16
(docs) Add 2.2 release notes.

gguillotte:


camlow325:
:+1: I assume we'll want to hold off on merging this until right around when we do the release?

gguillotte:
@camlow325 Yep!

camlow325:
@gguillotte One other thing I thought of which will be unique about Puppet Server 2.2.0 is that it will require a minimum of version 1.3.0 of the puppet-agent to be installed as a dependency.  Previous versions of Puppet Server 2.x had no base version dependency for the puppet-agent package.  The rpm/deb packaging systems will enforce this dependency so may not strictly be necessary to document it.  I'll defer to your judgment as to whether or not it would be useful to include this.

gguillotte:
@camlow325 Totally, good point. I'll mention it in the release notes and check for good places to mention it elsewhere in the 2.2 docs.

PR #774, opened 2015-11-17
(SERVER-1006) Unregister jruby instances on flush

cprice404:
Prior to this commit, when a JRuby instance was added to the pool,
it was also added to the list of "registered" instances for the pool
so that we could find an authoritative list of all of the instances
even when some were out on loan.

However, we did not have the corresponding logic to *unregister*
instances from this list when an instance was flushed via
max-requests-per-instance.  This resulted in a reference being
kept to the flushed instances even after they were flushed,
meaning that we could not reclaim all of the memory during
garbage collection.  This caused memory to leak over time.

This commit adds code to unregister instances when they are
flushed, so that we do not retain the extra reference to the
flushed instances, and we are able to garbage collect them
and free up the memory.

### DIFF
--- a/src/java/com/puppetlabs/puppetserver/RegisteredLinkedBlockingDeque.java
+++ b/src/java/com/puppetlabs/puppetserver/RegisteredLinkedBlockingDeque.java
@@ -40,6 +40,21 @@ public class RegisteredLinkedBlockingDeque<E> extends LinkedBlockingDeque<E> {
     }
 
     /**
+     * This method removes an element from the list of "registered" elements,
+     * such that it will no longer be returned by calls to
+     * <tt>getRegisteredInstances</tt>.
+     *
+     * This method does not remove the element from the underlying queue; it
+     * is assumed that the caller has already done so via the methods of the
+     * parent class.
+     *
+     * @param e the element to remove from the list of registered instances.
+     */
+    synchronized public void unregister(E e) throws InterruptedException {
+        registeredElements.remove(e);
+    }
+
+    /**
      * @return a set of all of the known elements that have been registered with
      *         this queue.
      */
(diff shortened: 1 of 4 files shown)


camlow325:
:+1:

cprice404:
Going to add one more commit to this to null out the http client references in the jrubies after a terminate.

fpringvaldsen:
:+1:, will give this another review once @cprice404's new commit is pushed up

cprice404:
@fpringvaldsen whoops, as @camlow325 pointed out, that change re: the http client is not relevant in the 1.x branch.  So I'm done with this PR, and we will need to add the commit that I hinted at in the 2.x branch once this is merged up

camlow325:
@KevinCorcoran I'm going to go ahead and merge this.  We can follow-up in later PRs if you find any issues from your review.

src/java/com/puppetlabs/puppetserver/RegisteredLinkedBlockingDeque.java
```diff
@@ -40,6 +40,21 @@ synchronized public void registerLast(E e) throws InterruptedException {
     }
 
     /**
+     * This method removes an element from the list of "registered" elements,
+     * such that it will no longer be returned by calls to
+     * <tt>getRegisteredInstances</tt>.
+     *
+     * This method does not remove the element from the underlying queue; it
+     * is assumed that the caller has already done so via the methods of the
+     * parent class.
```
KevinCorcoran: Should it?  Would it be better if this method also removed the element from the underlying queue?  That would make it consistent with `register`.

cprice404: How is the caller going to know what instance they want to unregister if they don't already have a reference to it?

KevinCorcoran:
retroactive :+1:

PR #777, opened 2015-11-17
(MAINT) null out http client reference on close

cprice404:
This commit nulls out the reference to the http client object
inside of the JRuby instances when we close them, just to
make 100% sure that they are eligible for garbage collection.

### DIFF
--- a/src/ruby/puppet-server-lib/puppet/server/http_client.rb
+++ b/src/ruby/puppet-server-lib/puppet/server/http_client.rb
@@ -87,8 +87,9 @@ class Puppet::Server::HttpClient
   end
 
   def self.terminate
-    unless self.client.nil?
-      self.client.close
+    unless @client.nil?
+      @client.close
+      @client = nil
     end
   end
 


cprice404:
@camlow325 I'm trying to test this locally but I can't seem to run the specs. :/

src/ruby/puppet-server-lib/puppet/server/http_client.rb
```diff
@@ -89,6 +89,7 @@ def get(url, headers)
   def self.terminate
     unless self.client.nil?
       self.client.close
+      self.client = nil
```
camlow325: Does this need to be `@client = nil`?

cprice404: oh, duh

cprice404: or i need to implement `client=`

src/ruby/puppet-server-lib/puppet/server/http_client.rb
```diff
@@ -165,6 +166,10 @@ def self.client
     @client ||= create_client
   end
 
+  def self.client=(new_client)
+    @client = new_client
+  end
```
cprice404: @camlow325 I added this setter method, which... is kind of crappy because I wouldn't want it to be called from anywhere else, but, maintains the abstraction of the class variable so that knowledge doesn't leak out into the other code.  This is kind of gross, but I think the other way would be gross in a different way.  If you have a strong opinion I can flip it, though.

src/ruby/puppet-server-lib/puppet/server/http_client.rb
camlow325: Hmm, now that I think about it, we made some changes around this stuff in the stable and master branches that hasn't gotten back to 2.1.x yet - https://github.com/puppetlabs/puppet-server/blob/master/src/ruby/puppet-server-lib/puppet/server/http_client.rb#L89-L93:

~~~ruby
def self.terminate
  unless @client.nil?
    @client.close
  end
end
~~~~

Before that, a `self.client.nil?` call would cause the client to be created if it hadn't already - and we'd turn right around and throw it away.  The change above was just an optimization for that case.

I'm okay with your `client=`, if that's what you'd prefer.  But I think it would be good to end up with the same implementations for these methods on all of our branches at the end of all of this.

cprice404: Cool, agree, will fix.

cprice404:
@camlow325 this look more like what you were hoping for?

camlow325:
:+1:

PR #778, opened 2015-11-17
(MAINT) merge up 2.1.x 2015 11 17

cprice404:
@camlow325 I was already working on this before I saw your last comments on the HTTP client nil patch, so I just went ahead and finished it.  I'll do another merge from 2.1.x to stable to include that after it's fixed/merged.


camlow325:
:+1:

PR #781, opened 2015-11-18
(SERVER-836) Bump puppet-agent dep to 1.3.0, puppet submodule to 4.3.0

camlow325:
This commit bumps the puppet-agent dependency to 1.3.0 and the
corresponding puppet submodule dependency to 4.3.0.


camlow325:
Travis passed on both cells that it ran on.  Would like to get this in ahead of the nightly runs, so merging.

PR #782, opened 2015-11-18
(MAINT) Fix broken puppet-agent debian dependency

camlow325:
This commit fixes a syntax problem in how the minimum puppet-agent
dependency was previously expressed for Debian - moving a paren to
the left of the version string.


camlow325:
Merging this so it goes in ahead of the nightly runs.

PR #783, opened 2015-11-18
(MAINT) Change puppet-agent minimum dep version to 1.3.0

camlow325:
This commit changes the minimum puppet-agent version dependency to
1.3.0.  Previous commits had used 4.3.0.  4.3.0 is the core Puppet
version that is packaged into puppet-agent 1.3.0.


camlow325:
Merging this to keep things moving for the SERVER 2.2.0 release.

PR #788, opened 2015-11-24
(MAINT) Introduce PuppetDB testing 

nwolfe:
This commit adds support for installing PuppetDB and writing tests.
Since PuppetDB packages aren't built for as many platforms as Puppet
Server we need to keep a list of supported platforms to protect the
pre-suite step and any tests against PuppetDB from breaking CI.

The PuppetDB in this commit is a relatively simple smoke test that
validates Puppet Server can communicate with PuppetDB and send it agent
information.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -28,9 +28,15 @@ module PuppetServerExtensions
     # puppet-agent version corresponds to packaged development version located at:
     # http://builds.delivery.puppetlabs.net/puppet-agent/
     puppet_build_version = get_option_value(options[:puppet_build_version],
-                         nil, "Puppet Development Build Version",
+                         nil, "Puppet Agent Development Build Version",
                          "PUPPET_BUILD_VERSION",
                          "2b6b9db87bb5e6329a5cc6520b6c6db569d4113f", :string)
+
+    # puppetdb version corresponds to packaged development version located at:
+    # http://builds.delivery.puppetlabs.net/puppetdb/
+    puppetdb_build_version =
+      get_option_value(options[:puppetdb_build_version], nil,
+                       "PuppetDB Version", "PUPPETDB_BUILD_VERSION", "3.2.1", :string)
 
     @config = {
       :base_dir => base_dir,
@@ -39,11 +45,27 @@ module PuppetServerExtensions
       :puppetserver_version => puppetserver_version,
       :puppet_version => puppet_version,
       :puppet_build_version => puppet_build_version,
+      :puppetdb_build_version => puppetdb_build_version,
     }
 
     pp_config = PP.pp(@config, "")
 
     Beaker::Log.notify "Puppet Server Acceptance Configuration:\n\n#{pp_config}\n\n"
+  end
+
+  # PuppetDB development packages aren't available on as many platforms as
+  # Puppet Server's packages, so we need to restrict the PuppetDB-related
+  # testing to a subset of the platforms.
+  # This guards both the installation of the PuppetDB package repository file
+  # and the running of the PuppetDB test(s).
+  def puppetdb_supported_platforms()
+    [
+      /debian-7/,
+      /debian-8/,
+      /el/, # includes cent6,7 and redhat6,7
+      /ubuntu-12/,
+      /ubuntu-14/,
+    ]
   end
 
   class << self
--- a/acceptance/suites/pre_suite/foss/30_install_dev_repos.rb
+++ b/acceptance/suites/pre_suite/foss/30_install_dev_repos.rb
@@ -25,3 +25,24 @@ if puppet_build_version
     end
   end
 end
+
+confine_block(:to, {:platform => puppetdb_supported_platforms}, master) do
+  step "Install PuppetDB repository" do
+    install_puppetlabs_dev_repo(
+      master, 'puppetdb', test_config[:puppetdb_build_version],
+      repo_config_dir, install_opts)
+
+    # Internal packages on ubuntu/debian aren't authenticated and thus apt
+    # will fail to install PuppetDB on those platforms.
+    # This hack tells apt that we can trust the PuppetDB packages until RE-6014
+    # is resolved, at which point this [trusted=yes] will already be in the file
+    # and we can delete the on(master, "sed ...") block below.
+    # This should make the puppetlabs-puppetdb module happily install PuppetDB on
+    # ubuntu/debian.
+    on(master, <<TRUSTPACKAGES)
+if [ -e /etc/apt/sources.list.d/pl-puppetdb* ]; then
+  sed -i -e 's/deb/deb [trusted=yes]/1' /etc/apt/sources.list.d/pl-puppetdb*
+fi
+TRUSTPACKAGES
+  end
+end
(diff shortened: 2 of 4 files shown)


acceptance/lib/helper.rb
```diff
@@ -28,10 +28,16 @@ def self.initialize_config(options)
...
-                         nil, "Puppet Development Build Version",
+                         nil, "Puppet Agent Development Build Version",
                          "PUPPET_BUILD_VERSION",
                          "2b6b9db87bb5e6329a5cc6520b6c6db569d4113f", :string)
 
+    # puppetdb version corresponds to packaged development version located at:
+    # http://builds.delivery.puppetlabs.net/puppetdb/
+    puppetdb_build_version =
+      get_option_value(options[:puppetdb_build_version], nil,
+                       "PuppetDB Version", "PUPPETDB_BUILD_VERSION", "3.2.1", :string)
```
nwolfe: I pinned to the latest release, but I tested with SNAPSHOT builds too (like today's 4.0.0.SNAPSHOT.2015.11.24T1346) and all is well.

nwolfe:
@camlow325 here's the puppetdb test that we've typically done by hand during FOSS release smoke testing.

CC @cprice404

acceptance/suites/tests/00_smoke/puppetdb_integration.rb
```diff
@@ -0,0 +1,52 @@
...
+## We can just run the agent that's on the master for this.
+
+confine :to, {:platform => puppetdb_supported_platforms}, master
+
+require 'json'
+
+test_name 'PuppetDB integration'
+
+step 'Install PuppetDB module' do
+  on(master, puppet("module install puppetlabs-puppetdb --version 5.0.0"))
```
camlow325: Would it make sense to also make this an option - like with the repo config - with a hard-coded default of 5.0.0 in the helper.rb?

acceptance/suites/tests/00_smoke/puppetdb_integration.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+end
+
+step 'Install PuppetDB with agent run' do
+  with_puppet_running_on(master, {}) do
+    on(master, puppet_agent("--test --server #{master}"), :acceptable_exit_codes => [0,2])
+  end
+end
+
+step 'Validate server sent agent report to PuppetDB' do
```
camlow325: Seems like when I've done this manually that I've had to do one more agent run after the first one in order to have a report be stored with puppetdb.  But maybe I'm remembering this wrong or was doing something wrong before.  If this works as is, sounds good to me.

nwolfe: You're remembering correctly, and at first I had 2 agent runs in here: one to install puppetdb via module, and one to do our test against.  But, during development the second one was always a noop and the first did what we wanted, so I just got rid of the second agent run.

acceptance/suites/pre_suite/foss/30_install_dev_repos.rb
```diff
@@ -25,3 +25,24 @@
...
+    # Internal packages on ubuntu/debian aren't authenticated and thus apt
+    # will fail to install PuppetDB on those platforms.
+    # This hack tells apt that we can trust the PuppetDB packages until RE-6014
+    # is resolved, at which point this [trusted=yes] will already be in the file
+    # and we can delete the on(master, "sed ...") block below.
+    # This should make the puppetlabs-puppetdb module happily install PuppetDB on
+    # ubuntu/debian.
+    on(master, <<TRUSTPACKAGES)
+if [ -e /etc/apt/sources.list.d/pl-puppetdb* ]; then
+  sed -i -e 's/deb/deb [trusted=yes]/1' /etc/apt/sources.list.d/pl-puppetdb*
```
camlow325: As a sanity check, would it be good to write this in such a way that if the `deb [trusted=yes]` is already in the file that it won't do a second substitution?  Not sure what would happen if the line ended up having `deb [trusted=yes] [trusted=yes]` in it?  Maybe dpkg would just handle that gracefully, ignoring the second one or last one winning?

acceptance/suites/tests/authorization/default_rules.rb
```diff
@@ -57,6 +57,8 @@ def report_query(node)
 end
 
 with_puppet_running_on(master, {}) do
+  masterfqdn = on(master, '/opt/puppetlabs/bin/facter fqdn').stdout.chomp
```
camlow325: Should we also use this in the URL hostname which is added to the curl command in `curl_authenticated` and `curl_unauthenticated` instead of `master`?

acceptance/suites/pre_suite/foss/30_install_dev_repos.rb
nwolfe: It never seemed to be problematic, and I actually think I would prefer this breaks, because then it would draw our attention to it and we could delete the hack (since presumably it breaks because RE-6014 is resolved).

acceptance/suites/tests/00_smoke/puppetdb_integration.rb
```diff
@@ -0,0 +1,52 @@
...
+  fqdn = on(master, '/opt/puppetlabs/bin/facter fqdn').stdout.chomp
+  query = "curl http://localhost:8080/pdb/query/v4/nodes/#{fqdn}"
```
camlow325: Is the plaintext port hosted by default packages?  Can't remember if I had to enable that manually, as opposed to using the SSL one?

acceptance/suites/tests/00_smoke/puppetdb_integration.rb
nwolfe: We could, yeah.  Or we could unpin it and see how far our luck goes?  I didn't want to spread this code out too much in the first attempt, but if we think it's worthwhile to expose a bit more I'd be happy to.

camlow325:
Overall, :+1:.  This is awesome - long overdue!

acceptance/suites/tests/00_smoke/puppetdb_integration.rb
camlow325: I'd probably go with either unpinning it or pinning it with the option to configure it externally like we do with the repo configs.  I'm good however you want to go with this for now.

acceptance/suites/tests/00_smoke/puppetdb_integration.rb
nwolfe: Yes, I think so ( https://github.com/puppetlabs/puppetdb/blob/master/documentation/api/query/curl.markdown#using-curl-from-localhost-non-sslhttp )

I never had a problem with this during development, and I tried a few releases of puppetdb on several platforms.

acceptance/suites/tests/authorization/default_rules.rb
nwolfe: It wasn't necessary in those curl calls, and I can't just define a top-level variable since there's no outer class/module.  Since it was only necessary in these calls, and since I could define the variable in this scope, I just did that and moved on.

acceptance/suites/tests/00_smoke/puppetdb_integration.rb
nwolfe: I think I'd like to unpin it actually.  I'll make the change.

camlow325:
:ship:

PR #789, opened 2015-12-01
(SERVER-1016) Add pool releaseItem variant for unlocking item w/o return

camlow325:
This commit renames the returnItem method in the JRubyPool and
LockablePool to releaseItem and adds a variant of the method which can
be used to determine whether the item should be released back into the
pool or just discarded.  For cases where an item is flushed, the item is
just discarded without going back into the pool - but with the
associated release lock for the thread being unlocked.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
@@ -299,8 +299,10 @@
                           "maximum number of requests (%s)")
                      (:id instance)
                      max-requests)
-          (flush-instance-fn pool instance)
-          (.releaseItem pool instance false))
+          (try
+            (flush-instance-fn pool instance)
+            (finally
+              (.releaseItem pool instance false))))
         (.releaseItem pool instance)))
     ;; if we get here, it was from a borrow and we got a Retry, so we just
     ;; return it to the pool.
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
@@ -247,7 +247,7 @@
   (let [instance (borrow-fn pool)]
     (cond (instance? PoisonPill instance)
           (do
-            (.returnItem pool instance)
+            (.releaseItem pool instance)
             (throw (IllegalStateException.
                      "Unable to borrow JRuby instance from pool"
                      (:err instance))))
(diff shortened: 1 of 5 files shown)


src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -299,11 +299,12 @@
                           "maximum number of requests (%s)")
                      (:id instance)
                      max-requests)
-          (flush-instance-fn pool instance))
-        (.returnItem pool instance)))
+          (flush-instance-fn pool instance)
+          (.releaseItem pool instance false))
```
KevinCorcoran: Should there be a `finally` here?  Maybe
```clj
(try
  (flush-instance-fn pool instance)
  (finally
    (.releaseItem pool instance false)))
```

camlow325: Yeah, that's probably a good idea.  Will add that.

KevinCorcoran:
:see_no_evil: except for `finally` mentioned in other comment (and lack of tests, obviously, which are hopefully coming later).

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
camlow325: Added this in 11a3704.

KevinCorcoran:
:+1:

camlow325:
I added a couple of lockable-pool unit tests for the overloaded `releaseItem` in  d75614e.

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
```diff
@@ -185,9 +185,33 @@
...
+    (let [pool (create-populated-pool 1)
+          instance (.borrowItem pool)]
+      (is (= 0 (.size pool)))
+      (.releaseItem pool instance false)
+      (is (= 0 (.size pool)))
+      (.lock pool)
+      (is (true? true))
+      (.unlock pool)
+      (is (true? true))))
+  (testing (str "releaseItem call with value 'true' does not return item to "
```
cprice404: I think this should say `*does* return item to pool`.

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
```diff
@@ -185,9 +185,33 @@
...
+                "pool and allows pool to still be lockable")
+    (let [pool (create-populated-pool 1)
+          instance (.borrowItem pool)]
+      (is (= 0 (.size pool)))
+      (.releaseItem pool instance true)
+      (is (= 1 (.size pool)))
+      (.lock pool)
+      (is (true? true))
```
cprice404: Would be worth a quick check to see if we could replace these with some sort of 'is lock held' assertion.  Ruth had a pattern for this in one of her tests but I think it might rely on having access to the full service context, so maybe not worth trying to duplicate here.

cprice404:
+1 ; going to merge this so our tests will run.  Will also spend some time playing with it locally, and maybe address my own comments from the tests above, and maybe look into adding another test that actually exercises some of the flushing stuff.

PR #796, opened 2015-12-02
remove -java from gem request

erikPrime:


kurtwall:
Not sure about this - my understanding is that we need the jruby version. Going to defer to @camlow325.

camlow325:
:+1:

camlow325:
My recollection is that gem, when running under jruby, will look for a "-java" variant of the base version of the gem it is asked to install.  If it doesn't find that, it will just try to install whatever non-"-java" version it finds.  So I think this change is good.  Will see what happens in CI, I suppose....

PR #797, opened 2015-12-02
pin nokogiri version, this time with correct syntax

erikPrime:


camlow325:
@erikPrime This variation won't allow the test to pass as-is because the test will try to find "nokogiri -v 1.6.7" in the output of a `gem list` command to see that the gem was installed but the list output instead will have "nokogiri (1.6.7 java)".

I made a slightly different attempt at a fix in #798.  I ran this test via beaker and confirmed that the test passed.

What do you think about going with that instead?

PR #799, opened 2015-12-02
(MAINT) Change enumeration in gem test cleanup to use key as name

camlow325:
This commit changes the enumeration of gems in the gem test cleanup to
use just the key name instead of the full key/value pair.  The original
approach would have used the key/value pair as an array representing the
name of the gem to uninstall, which was incorrect.


haus:
:+1:

erikPrime:
Looks good to me, but given our success rate around what seemed to be such a simple issue, lets have @nwolfe have a brief look to.

nwolfe:
:metal:

PR #801, opened 2015-12-04
(docs) change allowed words in auth.conf to match the actual code

fiddyspence:
The actual puppetserver code does not allow save or destroy as methods

### DIFF
--- a/documentation/config_file_auth.md
+++ b/documentation/config_file_auth.md
@@ -79,8 +79,8 @@ Indirector | HTTP
 -----------|------
 find       | GET and POST
 search     | GET and POST, for endpoints with names that end in "s" or "_search"
-save       | PUT
-destroy    | DELETE
+put        | PUT
+delete     | DELETE
 
 For more details, see the [Puppet `auth.conf` documentation](/puppet/latest/reference/config_file_auth.html#method).
 


camlow325:
@fiddyspence Thanks for this PR submission.

The intention of that section of the documentation was to indicate that where one would have used "indirector" names like find, search, save, destroy, etc. as the value for the `method` parameter in rules with the deprecated auth.conf format - like here (https://github.com/puppetlabs/puppet/blob/4.3.1/conf/auth.conf#L118) - one would now need to use the actual HTTP method instead with the new auth.conf format for Puppet Server.  To make this more clear, maybe we should change the header columns on the table to "Indirector name used for deprecated auth.conf" and "Corresponding HTTP method(s) to use for new auth.conf", or something like that.

@gguillotte - can you take a look at this and make a recommendation on how we might reword this to make it more clear?

gguillotte:
@camlow325 The line immediately preceding the table starts with "In the deprecated `auth.conf` format", and the line immediately after the table points users to the deprecated Ruby Puppet `auth.conf` docs. Revising this table's header row and the Ruby docs link could re-emphasize that it's deprecated, but I think we should consider restructuring the `auth.conf` doc to fully segregate the deprecated Ruby content from the supported tk-auth/HOCON content.

camlow325:
> I think we should consider restructuring the auth.conf doc to fully segregate the deprecated Ruby content from the supported tk-auth/HOCON content.

@gguillotte I agree with that suggestion.  We have a separate docs ticket, DOC-2396, which alludes to doing something like this.  For community visibility, would it make sense to you to move (or make a clone of) that ticket in a public project?  I could do either, per your recommendation.

gguillotte:
@camlow325 Good idea. I've moved it to DOCUMENT-466 for now.

I'll close this PR. @fiddyspence, would you like me to give you a heads-up when I have a follow-up PR?

PR #804, opened 2015-12-08
(maint) Update puppet, hiera, agent versions to latest

haus:
This updates the submodule versions as well as the packaged agent
versions to latest to ensure that we are testing and working against the
latest released puppet-agent.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,4 +18,5 @@ branches:
   only:
     - master
     - stable
+    - 2.1.x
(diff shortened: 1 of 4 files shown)


haus:
Seeing some failures in my local test. Please hold. Your call is important to us.

haus:
The heap size increase seems to have addressed the problems i was seeing.

camlow325:
:shipit:

PR #812, opened 2015-12-11
(MAINT) Remove one httparty gem reference from the Gemfile

camlow325:
The prior commit added an extra reference to the httparty gem in the
puppet-server Gemfile which is unnecessary.  This commit removes that
reference.


nwolfe:
:gem:

PR #813, opened 2015-12-11
(SERVER-862) Enable jruby pool write locker to borrow pool instances

camlow325:
This PR reworks the `JRubyPool` implementation to pull in all of the
concurrent constructs around a single `ReentrantLock`, the primary benefit
being to allow the holder of a write lock to be able to borrow instances
from the pool without running the risk of a deadlock due to competing
borrows that may be occurring from other threads.  The implementation
uses three condition variables to gate borrower and locker access to the
pool, similar to how `LinkedBlockingDeque` does for managing queue takers.

Where the underlying lock construct in the `JRubyPool` class would be a
`ReentrantLock` instead of a `ReentrantReadWriteLock`, this commit also
reworks a few of the unit tests to attempt borrows from non-locking
threads as an indication that the write lock is held, as opposed to
trying to interrogate the lock construct itself.

This PR also cleans up the Javadocs for the `LockablePool` interface and
`JRubyPool` class to be less redundant and cover only details that the
consumer of the two would need to know, pulling internal details about the
`JRubyPool` class into comments within that class.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
@@ -94,8 +94,9 @@
   (let [{:keys [config profiler pool-state]} pool-context
         new-pool-state (jruby-internal/create-pool-from-config config)
         new-pool (:pool new-pool-state)
-        old-pool @pool-state
-        count    (:size old-pool)]
+        old-pool-state @pool-state
+        old-pool (:pool old-pool-state)
+        count    (:size old-pool-state)]
     (log/info "Replacing old JRuby pool with new instance.")
     (reset! pool-state new-pool-state)
     (log/info "Swapped JRuby pools, beginning cleanup of old pool.")
@@ -104,11 +105,11 @@
         (let [id        (inc i)
               instance  (jruby-internal/borrow-from-pool!*
                           jruby-internal/borrow-without-timeout-fn
-                          (:pool old-pool))]
+                          old-pool)]
           (try
             (flush-instance! pool-context instance new-pool id config profiler)
             (finally
-              (.releaseItem (:pool old-pool) instance false)))
+              (.releaseItem old-pool instance false)))
           (log/infof "Finished creating JRubyPuppet instance %d of %d"
                      id count))
         (catch Exception e
@@ -119,7 +120,7 @@
                    e)))))
     ;; Add a "RetryPoisonPill" to the pool in case something else is in the
     ;; process of borrowing from the old pool.
-    (.insertPill (:pool old-pool) (RetryPoisonPill. (:pool old-pool)))))
+    (.insertPill old-pool (RetryPoisonPill. old-pool))))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Public
--- a/src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
+++ b/src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
@@ -1,62 +1,100 @@
 package com.puppetlabs.puppetserver.pool;
 
+import java.util.LinkedList;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.CopyOnWriteArraySet;
-import java.util.concurrent.LinkedBlockingDeque;
-import java.util.concurrent.locks.ReentrantReadWriteLock;
-import com.puppetlabs.puppetserver.pool.LockablePool;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.ReentrantLock;
 
 /**
- * A data structure to be used as a Pool, encapsulating a LinkedBlockingDeque
- * and a ReentrantReadWriteLock. Implements the <tt>LockablePool</tt>
- * interface such that when a <tt>borrowItem</tt> is performed an instance is
- * taken from the deque and a read lock is acquired, when a
- * <tt>releaseItem</tt> is performed an instance is put back onto the deque and
- * a read lock is released, and <tt>lock</tt> locks a write lock.
+ * A data structure to be used as a Pool, encapsulating a ReentrantLock around
+ * a LinkedList.
  *
- * @param <E> the type of element that can be added to the LinkedBlockingDeque.
+ * @param <E> the type of element that can be added to the LinkedList.
  */
 public final class JRubyPool<E> implements LockablePool<E> {
-    private final LinkedBlockingDeque<E> liveQueue;
-    private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock(true);
+    // Underlying queue which holds the elements that clients can borrow.
+    private final LinkedList<E> liveQueue;
+
+    // Lock which guards all accesses to the underlying queue and registered
+    // element set.  Constructed as "nonfair" for performance, like the lock
+    // that a <tt>LinkedBlockingDeque</tt> does.  Not clear that we need this
+    // to be a "fair" lock.
+    private final ReentrantLock lock = new ReentrantLock();
+
+    // Condition signaled when all elements that have been registered have been
+    // returned to the queue.  Awaited when a lock has been requested but
+    // one or more registered elements has been borrowed from the pool.
+    private final Condition allRegisteredInQueue = lock.newCondition();
+
+    // Condition signaled when an element has been added into the queue.
+    // Awaited when a request has been made to borrow an item but no elements
+    // currently exist in the queue.
+    private final Condition notEmpty = lock.newCondition();
+
+    // Condition signaled when the pool has been unlocked.  Awaited when a
+    // request has been made to borrow an item or lock the pool but the pool
+    // is currently locked.
+    private final Condition notLocked = lock.newCondition();
+
+    // Holds a reference to all of the elements that have been registered.
+    // Newly registered elements are also added into the <tt>liveQueue</tt>.
+    // Elements only exist in the <tt>liveQueue</tt> when not currently
+    // borrowed whereas elements that have been registered (but not
+    // yet unregistered) will be accessible via <tt>registeredElements</tt>
+    // even while they are borrowed.
     private final Set<E> registeredElements = new CopyOnWriteArraySet<>();
 
+    // Maximum size that the underlying queue can grow to.
+    private int maxSize;
+
+    // Id of the thread which currenly holds the write lock.  -1 indicates that
+    // there is no current write lock holder.
+    private volatile long lockThreadId = -1;
+
+    /**
+     * Create a JRubyPool
+     *
+     * @param size maximum capacity for the pool.
+     */
     public JRubyPool(int size) {
-        liveQueue = new LinkedBlockingDeque<>(size);
+        liveQueue = new LinkedList<>();
+        maxSize = size;
     }
 
     /**
-    * This method is analagous to <tt>putLast</tt> in the
-    * <tt>LinkedBlockingDeque</tt> class, but also causes the element to be
-    * added to the list of "registered" elements that will be returned by
-    * <tt>getRegisteredInstances</tt>.
-    *
-    * This method is synchronized not for thread safety, but to try to ensure that
-    * to consumers of this class adding an instance to the queue and adding it to
-    * the set of registered elements is visible roughly atomically. `synchronize`
-    * guarantees that registration is atomic, so (ignoring borrows happening)
-    * `RegisteredElements` and `liveQueue` can only diverge by 1. "This is only
-    * "roughly" atomic, as the underlying queue uses its own lock, so it is
-    * possible for it to be modified on another thread while this method is being
-    * executed.
-    *
-    * @param e the element to register and put at the end of the queue.
-    *
-    * @throws InterruptedException
-    */
+     * This method is analogous to <tt>addLast</tt> in the <tt>LinkedList</tt>
+     * class, but also causes the element to be added to the list of
+     * "registered" elements that will be returned by
+     * <tt>getRegisteredInstances</tt>.
+     *
+     * @param e the element to register and put at the end of the queue.
+     * @throws IllegalStateException if an attempt is made to register an
+     *                               element but the number of registered
+     *                               instances is already equal to the maximum
+     *                               capacity for the pool.
+     */
     @Override
-    synchronized public void register(E e) throws InterruptedException {
-        registeredElements.add(e);
-        liveQueue.putLast(e);
+    public void register(E e) {
+        lock.lock();
+        try {
+            if (registeredElements.size() == maxSize)
+                throw new IllegalStateException(
+                        "Unable to register additional instance, pool full");
+            registeredElements.add(e);
+            liveQueue.addLast(e);
+            signalNotEmpty();
+        } finally {
+            lock.unlock();
+        }
     }
-
 
     /**
      * This method removes an element from the list of "registered" elements,
      * such that it will no longer be returned by calls to
      * <tt>getRegisteredInstances</tt>.
-     *
+     * <p>
      * This method does not remove the element from the underlying queue; it
      * is assumed that the caller has already done so via the methods of the
      * parent class.
@@ -64,163 +102,253 @@ public final class JRubyPool<E> implements LockablePool<E> {
      * @param e the element to remove from the list of registered instances.
      */
     @Override
-    synchronized public void unregister(E e) throws InterruptedException {
-        registeredElements.remove(e);
+    public void unregister(E e) {
+        lock.lock();
+        try {
+            registeredElements.remove(e);
+            signalIfAllRegisteredInQueue();
+        } finally {
+            lock.unlock();
+        }
     }
 
-
     /**
-    * This method is analagous to <tt>takeFirst</tt> in the
-    * <tt>LinkedBlockingDeque</tt> class, but also causes a read lock to be
-    * locked from the <tt>ReentrantReadWriteLock</tt>.
-    *
-    * @return The head of the deque.
-    *
-    * @throws IllegalStateException if the thread holding the write lock
-    *         attempts to call this method.
-    * @throws InterruptedException
-    */
+     * This method is analogous to <tt>removeFirst</tt> in the
+     * <tt>LinkedList</tt> class.  An element will not be returned until
+     * the queue is unlocked and an element can be pulled out of the queue
+     * for return.
+     *
+     * @return The head of the queue.
+     * @throws InterruptedException if the calling thread is interrupted
+     *                              while waiting for the pool to be
+     *                              unlocked or for an element to
+     *                              be available in the queue for borrowing.
+     */
     @Override
-    public E borrowItem() throws InterruptedException, IllegalStateException {
-        if (lock.isWriteLockedByCurrentThread()) {
-          throw new IllegalStateException("The current implementation has a risk of deadlock if you attempt to borrow a JRuby instance while holding the write lock!");
+    public E borrowItem() throws InterruptedException {
+        E item = null;
+
+        lock.lock();
+        try {
+            long currentThreadId = getThreadId(Thread.currentThread());
+            while (item == null) {
+                if (lockThreadId != -1 && lockThreadId != currentThreadId) {
+                    notLocked.await();
+                } else if (liveQueue.size() < 1) {
+                    notEmpty.await();
+                } else {
+                    item = liveQueue.removeFirst();
+                }
+            }
+        } finally {
+            lock.unlock();
         }
-        E item = liveQueue.takeFirst();
-        lock.readLock().lock();
+
         return item;
     }
 
     /**
-    * This method is analagous to <tt>pollFirst</tt> in the
-    * <tt>LinkedBlockingDeque</tt> class, but also causes a read lock to be
-    * locked from the <tt>ReentrantReadWriteLock</tt>.
-    *
-    * @param timeout how long to wait before giving up, in units of unit
-    * @param unit a <tt>TimeUnit</tt> determining how to interpret the
-    *        <tt>timeout parameter</tt>
-    *
-    * @return The head of the deque, or <tt>null</tt> if the specified waiting
-    *         time elapses before an element is available.
-    *
-    * @throws IllegalStateException if the thread holding the write lock
-    *         attempts to call this method.
-    * @throws InterruptedException
-    */
+     * This method is analogous to <tt>removeFirst</tt> in the
+     * <tt>LinkedList</tt> class but with a timed maximum wait for an element
+     * to be available in the queue for borrowing.
+     *
+     * @param timeout how long to wait before giving up, in units of unit
+     * @param unit    a <tt>TimeUnit</tt> determining how to interpret the
+     *                <tt>timeout parameter</tt>
+     * @return The head of the queue, or <tt>null</tt> if the specified waiting
+     *         time elapses before an element is available.
+     * @throws InterruptedException if the calling thread is interrupted while
+     *                              waiting for the pool to be unlocked or for
+     *                              an element to be available in the queue
+     *                              for borrowing.
+     */
     @Override
-    public E borrowItemWithTimeout(long timeout, TimeUnit unit) throws InterruptedException, IllegalStateException {
-        if (lock.isWriteLockedByCurrentThread()) {
-          throw new IllegalStateException("The current implementation has a risk of deadlock if you attempt to borrow a JRuby instance while holding the write lock!");
+    public E borrowItemWithTimeout(long timeout, TimeUnit unit) throws
+            InterruptedException {
+        E item = null;
+
+        long remainingMaxTimeToWait = unit.toNanos(timeout);
+        lock.lockInterruptibly();
+        try {
+            long currentThreadId = getThreadId(Thread.currentThread());
+            while (item == null && remainingMaxTimeToWait > 0) {
+                if (lockThreadId != -1 && lockThreadId != currentThreadId) {
+                    remainingMaxTimeToWait =
+                            notLocked.awaitNanos(remainingMaxTimeToWait);
+                } else if (liveQueue.size() < 1) {
+                    remainingMaxTimeToWait =
+                            notEmpty.awaitNanos(remainingMaxTimeToWait);
+                } else {
+                    item = liveQueue.removeFirst();
+                }
+            }
+        } finally {
+            lock.unlock();
         }
-        E item = liveQueue.pollFirst(timeout, unit);
-        lock.readLock().lock();
+
         return item;
     }
 
     /**
-    * This method is analogous to <tt>putFirst</tt> in the
-    * <tt>LinkedBlockingDeque</tt> class, but also causes a read lock to be
-    * unlocked from the <tt>ReentrantReadWriteLock</tt>.  This method should
-    * only be called from the same thread that was running when the item was
-    * originally borrowed.
-    *
-    * @param e the element to return to the pool
-    * @throws InterruptedException
-    */
+     * This method is analogous to <tt>addLast</tt> in the <tt>LinkedList</tt>
+     * class.
+     *
+     * @param e the element to return to the pool
+     */
     @Override
-    public void releaseItem(E e) throws InterruptedException {
+    public void releaseItem(E e) {
         releaseItem(e, true);
     }
 
     /**
-    * This method is analogous to <tt>putFirst</tt> in the
-    * <tt>LinkedBlockingDeque</tt> class, but also causes a read lock to be
-    * unlocked from the <tt>ReentrantReadWriteLock</tt>.  This method should
-    * only be called from the same thread that was running when the item was
-    * originally borrowed.
-    *
-    * @param e the element to return to the pool
-    * @param returnToPool whether to return the element to the pool (true)
-    *                     or just throw the element away (false)
-    * @throws InterruptedException
-    */
+     * This method is analogous to <tt>addFirst</tt> in the <tt>LinkedList</tt>
+     * class.
+     *
+     * @param e            the element to return to the pool
+     * @param returnToPool whether to return the element to the pool (true)
+     *                     or just throw the element away (false)
+     */
     @Override
-    public void releaseItem(E e, boolean returnToPool)
-            throws InterruptedException {
+    public void releaseItem(E e, boolean returnToPool) {
+        lock.lock();
         try {
             if (returnToPool) {
-                liveQueue.putFirst(e);
+                addFirst(e);
             }
         } finally {
-            lock.readLock().unlock();
+            lock.unlock();
         }
     }
 
     /**
-    * This method is analagous to <tt>putFirst</tt> in the
-    * <tt>LinkedBlockingDeque</tt> class. It should only ever be used to
-    * insert a `PoisonPill` or `RetryPoisonPill` to the pool, which is an
-    * operation that should be done without acquiring a read lock.
-    *
-    * @param e the element to add to the queue
-    * @throws InterruptedException
-    */
+     * This method is analogous to <tt>addFirst</tt> in the
+     * <tt>LinkedList</tt> class.  It should only ever be used to
+     * insert a `PoisonPill` or `RetryPoisonPill` to the pool.  The
+     * implementation of this method is equivalent to calling
+     * releaseItem(e, true).
+     *
+     * @param e the element to add to the queue
+     */
     @Override
-    public void insertPill(E e) throws InterruptedException {
-        liveQueue.putFirst(e);
+    public void insertPill(E e) {
+        lock.lock();
+        try {
+            addFirst(e);
+        } finally {
+            lock.unlock();
+        }
     }
 
     /**
-    * This method is analagous to <tt>clear</tt> in the
-    * <tt>LinkedBlockingDeque</tt> class.
-    *
-    * @throws InterruptedException
-    */
+     * This method is analogous to <tt>clear</tt> in the
+     * <tt>LinkedList</tt> class.
+     */
     @Override
     public void clear() {
-        liveQueue.clear();
+        lock.lock();
+        try {
+            liveQueue.clear();
+        } finally {
+            lock.unlock();
+        }
     }
 
     /**
-    * This method is analagous to <tt>remainingCapacity</tt> in the
-    * <tt>LinkedBlockingDeque</tt> class.
-    *
-    * @return the number of additional elements that the queue can ideally
-    *         accept without blocking.
-    */
+     * This method calculates the remaining capacity in the queue from
+     * the supplied <tt>size</tt> at construction minus the current number of
+     * elements in the underlying <tt>LinkedList</tt>.
+     *
+     * @return the number of additional elements that the queue can accept.
+     */
     @Override
     public int remainingCapacity() {
-        return liveQueue.remainingCapacity();
+        int remainingCapacity;
+        lock.lock();
+        try {
+            remainingCapacity = maxSize - liveQueue.size();
+        } finally {
+            lock.unlock();
+        }
+        return remainingCapacity;
     }
 
     /**
-    * This method is analagous to <tt>size</tt> in the
-    * <tt>LinkedBlockingDeque</tt> class.
-    *
-    * @return the number of elements in the queue
-    */
+     * This method is analogous to <tt>size</tt> in the <tt>LinkedList</tt>
+     * class.
+     *
+     * @return the number of elements in the queue
+     */
     @Override
     public int size() {
-        return liveQueue.size();
+        int size;
+        lock.lock();
+        try {
+            size = liveQueue.size();
+        } finally {
+            lock.unlock();
+        }
+        return size;
     }
 
     /**
-    * Acquires a write lock from <tt>ReentrantReadWriteLock</tt>.
-    *
-    * Note that this implementation does not fulfill requirement (e) in the
-    * LockablePool javadocs for lock(); with this implementation, the thread
-    * holding the lock cannot perform a borrow while it's holding the lock.
-    */
+     * Acquires a "lock" on the underlying queue, preventing any would-be
+     * future borrowers from acquiring an element or any future pool lockers
+     * until the lock is freed.
+     *
+     * @throws InterruptedException if the calling thread is interrupted while
+     *                              waiting for the pool to be unlocked.
+     */
     @Override
     public void lock() throws InterruptedException {
-        lock.writeLock().lock();
+        lock.lock();
+        try {
+            long currentThreadId = getThreadId(Thread.currentThread());
+            while (currentThreadId != lockThreadId) {
+                if (lockThreadId == -1) {
+                    lockThreadId = currentThreadId;
+                } else {
+                    notLocked.await();
+                }
+            }
+            while (registeredElements.size() != liveQueue.size()) {
+                allRegisteredInQueue.await();
+            }
+        } finally {
+            lock.unlock();
+        }
     }
 
     /**
-    * Releases the write lock.
-    */
+     * Releases the write lock.
+     *
+     * @throws IllegalStateException if the calling thread does not currently
+     *                               hold the write lock.
+     */
     @Override
     public void unlock() throws InterruptedException {
-        lock.writeLock().unlock();
+        lock.lock();
+        try {
+            long currentThreadId = getThreadId(Thread.currentThread());
+            if (currentThreadId != lockThreadId) {
+                String threadIdForMessage;
+                if (lockThreadId == -1) {
+                    threadIdForMessage = "not held by any thread";
+                } else {
+                    threadIdForMessage = "held by thread id " + lockThreadId;
+                }
+                throw new IllegalStateException(
+                        "Unlock requested from thread not holding the lock.  " +
+                                "Requested from thread id " + currentThreadId +
+                                " but lock " + threadIdForMessage);
+            }
+            lockThreadId = -1;
+            // Need to use 'signalAll' here because there might be multiple
+            // waiters (e.g., multiple borrowers) queued up, waiting for the
+            // pool to be unlocked.
+            notLocked.signalAll();
+        } finally {
+            lock.unlock();
+        }
     }
 
     /**
@@ -231,4 +359,40 @@ public final class JRubyPool<E> implements LockablePool<E> {
         return registeredElements;
     }
 
+    private void addFirst(E e) {
+        liveQueue.addFirst(e);
+        signalNotEmpty();
+    }
+
+    private void signalNotEmpty() {
+        // Could use 'signalAll' here instead of 'signal' but 'signal' is
+        // less expensive in that only one waiter will be woken up.  Can use
+        // signal here because the thread being awoken will be able to borrow
+        // a pool instance and any further waiters will be woken up by
+        // subsequent posts of this signal when instances are added/returned to
+        // the queue.
+        notEmpty.signal();
+        signalIfAllRegisteredInQueue();
+    }
+
+    private void signalIfAllRegisteredInQueue() {
+        // Could use 'signalAll' here instead of 'signal'.  Doesn't really
+        // matter though in that there will only be one waiter at most which
+        // is active at a time - a caller of lock() that has just acquired
+        // the lock but is waiting for all registered elements to be returned
+        // to the queue.
+        if (registeredElements.size() == liveQueue.size()) {
+            allRegisteredInQueue.signal();
+        }
+    }
+
+    /**
+     * Returns the thread id for a given thread.  For safety, we might want /
+     * need to replace this with the scary-looking method that the OpenJDK uses
+     * in <tt>ReentrantReadWriteLock</tt> - see
+     * http://hg.openjdk.java.net/jdk8/jdk8/jdk/rev/1919c226b427#l1.124.
+     */
+     private static long getThreadId(Thread thread) {
+         return thread.getId();
+     }
 }
--- a/src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
+++ b/src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
@@ -240,14 +240,18 @@ public final class JRubyPool<E> implements LockablePool<E> {
     }
 
     /**
-     * This method is analogous to <tt>clear</tt> in the
-     * <tt>LinkedList</tt> class.
+     * This method is analogous to <tt>clear</tt> in the <tt>LinkedList</tt>
+     * class.  This method clears all elements currently in the queue and
+     * also unregisters them from the set of registered elements.
      */
     @Override
     public void clear() {
         lock.lock();
         try {
-            liveQueue.clear();
+            int queueSize = liveQueue.size();
+            for (int i=0; i<queueSize; i++) {
+                registeredElements.remove(liveQueue.removeFirst());
+            }
         } finally {
             lock.unlock();
         }
--- a/src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
+++ b/src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
@@ -317,6 +317,16 @@ public final class JRubyPool<E> implements LockablePool<E> {
         // waiters (e.g., multiple borrowers) queued up, waiting for the
         // pool to be unlocked.
         notWriteLocked.signalAll();
+        // Borrowers that are woken up when an instance is returned to the
+        // pool and the write lock is held would then start waiting on a
+        // 'notWriteLocked' signal instead.  Re-signalling 'notEmpty' here
+        // allows any borrowers still waiting on the 'notEmpty' signal to be
+        // reawoken when the write lock is released, compensating for any
+        // 'notEmpty' signals that might have been essentially ignored from
+        // when the write lock was held.
+        if (liveQueue.size() > 0) {
+            notEmpty.signalAll();
+        }
     }
 
     private void signalNotEmpty() {
--- a/src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
+++ b/src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
@@ -8,18 +8,39 @@ import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.ReentrantLock;
 
 /**
- * A data structure to be used as a Pool, encapsulating a ReentrantLock around
- * a LinkedList.
+ * An implementation of LockablePool for managing a pool of JRubyPuppet
+ * instances.
  *
- * @param <E> the type of element that can be added to the LinkedList.
+ * @param <E> the type of element that can be added to the pool.
  */
 public final class JRubyPool<E> implements LockablePool<E> {
+    // The `LockingPool` contract requires some synchronization behaviors that
+    // are not natively present in any of the JDK deque implementations -
+    // specifically to allow one calling thread to call lock() to supercede
+    // and hold off any pending pool borrowers until unlock() is called.
+    // This class implementation fulfills the contract by managing the
+    // synchronization constructs directly rather than deferring to an
+    // underlying JDK data structure to manage concurrent access.
+    //
+    // This implementation is modeled somewhat off of what the
+    // `LinkedBlockingDeque` class in the OpenJDK does to manage
+    // concurrency.  It uses a single `ReentrantLock` to provide mutual
+    // exclusion around offer and take requests, with condition variables
+    // used to park and later reawaken requests as needed, e.g., when pool
+    // items are unavailable for borrowing or when the write lock is
+    // unavailable.
+    //
+    // See http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/concurrent/LinkedBlockingDeque.java#l157
+    //
+    // Because access to the underlying deque is synchronized within
+    // this class, the pool is backed by a non-synchronized JDK `LinkedList`.
+
     // Underlying queue which holds the elements that clients can borrow.
     private final LinkedList<E> liveQueue;
 
     // Lock which guards all accesses to the underlying queue and registered
     // element set.  Constructed as "nonfair" for performance, like the lock
-    // that a <tt>LinkedBlockingDeque</tt> does.  Not clear that we need this
+    // that a `LinkedBlockingDeque` does.  Not clear that we need this
     // to be a "fair" lock.
     private final ReentrantLock lock = new ReentrantLock();
 
@@ -36,22 +57,26 @@ public final class JRubyPool<E> implements LockablePool<E> {
     // Condition signaled when the pool has been unlocked.  Awaited when a
     // request has been made to borrow an item or lock the pool but the pool
     // is currently locked.
-    private final Condition notLocked = lock.newCondition();
+    private final Condition notWriteLocked = lock.newCondition();
 
     // Holds a reference to all of the elements that have been registered.
-    // Newly registered elements are also added into the <tt>liveQueue</tt>.
-    // Elements only exist in the <tt>liveQueue</tt> when not currently
+    // Newly registered elements are also added into the `liveQueue`.
+    // Elements only exist in the `liveQueue` when not currently
     // borrowed whereas elements that have been registered (but not
-    // yet unregistered) will be accessible via <tt>registeredElements</tt>
+    // yet unregistered) will be accessible via `registeredElements`
     // even while they are borrowed.
     private final Set<E> registeredElements = new CopyOnWriteArraySet<>();
 
     // Maximum size that the underlying queue can grow to.
     private int maxSize;
 
-    // Id of the thread which currenly holds the write lock.  -1 indicates that
-    // there is no current write lock holder.
-    private volatile long lockThreadId = -1;
+    // Thread which currently holds the write lock.  null indicates that
+    // there is no current write lock holder.  Using the current Thread
+    // object for tracking the lock owner is comparable to what the JDK's
+    // `ReentrantLock` class does:
+    //
+    // http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/concurrent/locks/ReentrantLock.java#l164
+    private Thread writeLockThread = null;
 
     /**
      * Create a JRubyPool
@@ -63,20 +88,9 @@ public final class JRubyPool<E> implements LockablePool<E> {
         maxSize = size;
     }
 
-    /**
-     * This method is analogous to <tt>addLast</tt> in the <tt>LinkedList</tt>
-     * class, but also causes the element to be added to the list of
-     * "registered" elements that will be returned by
-     * <tt>getRegisteredInstances</tt>.
-     *
-     * @param e the element to register and put at the end of the queue.
-     * @throws IllegalStateException if an attempt is made to register an
-     *                               element but the number of registered
-     *                               instances is already equal to the maximum
-     *                               capacity for the pool.
-     */
     @Override
     public void register(E e) {
+        final ReentrantLock lock = this.lock;
         lock.lock();
         try {
             if (registeredElements.size() == maxSize)
@@ -90,19 +104,9 @@ public final class JRubyPool<E> implements LockablePool<E> {
         }
     }
 
-    /**
-     * This method removes an element from the list of "registered" elements,
-     * such that it will no longer be returned by calls to
-     * <tt>getRegisteredInstances</tt>.
-     * <p>
-     * This method does not remove the element from the underlying queue; it
-     * is assumed that the caller has already done so via the methods of the
-     * parent class.
-     *
-     * @param e the element to remove from the list of registered instances.
-     */
     @Override
     public void unregister(E e) {
+        final ReentrantLock lock = this.lock;
         lock.lock();
         try {
             registeredElements.remove(e);
@@ -112,28 +116,16 @@ public final class JRubyPool<E> implements LockablePool<E> {
         }
     }
 
-    /**
-     * This method is analogous to <tt>removeFirst</tt> in the
-     * <tt>LinkedList</tt> class.  An element will not be returned until
-     * the queue is unlocked and an element can be pulled out of the queue
-     * for return.
-     *
-     * @return The head of the queue.
-     * @throws InterruptedException if the calling thread is interrupted
-     *                              while waiting for the pool to be
-     *                              unlocked or for an element to
-     *                              be available in the queue for borrowing.
-     */
     @Override
     public E borrowItem() throws InterruptedException {
         E item = null;
-
+        final ReentrantLock lock = this.lock;
         lock.lock();
         try {
-            long currentThreadId = getThreadId(Thread.currentThread());
+            final Thread currentThread = Thread.currentThread();
             while (item == null) {
-                if (lockThreadId != -1 && lockThreadId != currentThreadId) {
-                    notLocked.await();
+                if (isWriteLockHeldByAnotherThread(currentThread)) {
+                    notWriteLocked.await();
                 } else if (liveQueue.size() < 1) {
                     notEmpty.await();
                 } else {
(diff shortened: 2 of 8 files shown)


src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
```diff
@@ -94,8 +94,9 @@
   (let [{:keys [config profiler pool-state]} pool-context
         new-pool-state (jruby-internal/create-pool-from-config config)
         new-pool (:pool new-pool-state)
-        old-pool @pool-state
-        count    (:size old-pool)]
+        old-pool-state @pool-state
```
camlow325: No functional differences for this namespace, just reducing some repetition.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
```diff
@@ -1,234 +1,355 @@
...
-        lock.writeLock().lock();
+        final ReentrantLock lock = this.lock;
+        lock.lock();
+        try {
+            final Thread currentThread = Thread.currentThread();
+            while (!isWriteLockHeldByCurrentThread(currentThread)) {
+                if (!isWriteLockHeld()) {
+                    writeLockThread = currentThread;
+                } else {
+                    notWriteLocked.await();
```
camlow325: @cprice404 and I talked about this a bit.  Contrary to the method signature, the implementation of this method on the master branch currently does not appear to lead to an `InterruptedException` being thrown.  To preserve the semantic behavior with the prior implementation, the line above and `allRegisteredInQueue.await()` call below could use `awaitUninterruptibly` instead.

Overall, though, I think it might be better, however, to just let this method be interruptible.  For example, a process shutdown might try to interrupt the thread.  Where these `await` calls could run for a pretty long time - e.g., while some particularly long code updates or catalog compilations are going on - it would seem better to just let the thread be interrupted so that the process can shut down sooner.  May need to rethink this later if we have more specific use cases for thread interruption across pool consumers.

test/integration/puppetlabs/services/jruby/puppet_environments_int_test.clj
```diff
@@ -57,20 +57,6 @@
     {:headers     {"Accept" "pson"}
      :as          :text}))
 
-(defn service-context
```
camlow325: I had a need to use these for some of the lockable pool tests, so I moved these down into the common `test/unit/puppetlabs/services/jruby/jruby_testutils`.

test/unit/puppetlabs/services/jruby/jruby_testutils.clj
```diff
@@ -176,25 +176,18 @@
...
-               .getDeclaredFields
-               (filter #(= "lock" (.getName %)))
-               first)]
-    (.setAccessible f true)
-    (.get f jruby-pool)))
-
-(defn get-lock-from-app
-  "Given a trapperkeeper application with the JRubyPuppetService running,
-  returns the ReentrantReadWriteLock instance used by its JRubyPool."
+(defn wait-for-jrubies
```
camlow325: Relocated - with some consolidation - from `test/integration/puppetlabs/services/jruby/puppet_environments_int_test.clj`.

test/unit/puppetlabs/services/jruby/jruby_testutils.clj
```diff
@@ -176,25 +176,18 @@
     (fill-drained-pool jrubies)
     result))
 
-(defn get-lock-from-pool
```
camlow325: Where the use of a `ReentrantReadWriteLock` would go away completely with this PR, I wanted to rewrite some of the tests to try to figure out whether the "write lock" is held or not by validating the behavior of public methods of the `JRubyPool` class rather than validating the state held internally for the "write lock" - e.g., the "write lock" must be "locked" if a non-writer thread borrower is unable to borrow instance while the "lock" is supposed to be held or must be "unlocked" if borrows from any threads are granted.

cprice404:
@camlow325 was there a previous PR that you closed that had our earlier conversations on it?

cprice404:
@KevinCorcoran @rlinehan FYI this is probably the next PR that needs to land in puppetserver.  I've already mostly reviewed the production code, and am pretty much +1.  I need to carve out some time to review the tests, hoping to do that next week.  If I'm able to, I may merge it, but if either of you are interested in reviewing it'd certainly be good to get some more eyeballs on it.

camlow325:
> @camlow325 was there a previous PR that you closed that had our earlier conversations on it?

@cprice404 No PR, but the commits on my old branch with your review comments should still be here for reference:

https://github.com/puppetlabs/puppet-server/commit/5548d980f4d1e9037bbe5042a5c3693b7a375b23
https://github.com/puppetlabs/puppet-server/commit/36ccd4e6cb3ceb9ddf408aea5d894d81095af7b9

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
```diff
@@ -1,234 +1,371 @@
...
+    // See http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/concurrent/LinkedBlockingDeque.java#l157
+    //
+    // Because access to the underlying deque is synchronized within
+    // this class, the pool is backed by a non-synchronized JDK `LinkedList`.
 
-    public JRubyPool(int size) {
-        liveQueue = new LinkedBlockingDeque<>(size);
-    }
+    // Underlying queue which holds the elements that clients can borrow.
+    private final LinkedList<E> liveQueue;
```
KevinCorcoran: Why `LinkedList` instead of just `List`?

KevinCorcoran: I guess that question is actually two parts:

1) Can this type be just `List` instead of `LinkedList` or does the implementation depend on some methods on `LinkedList`?

2) If the answer to that is "yes", why `LinkedList` instead of the more commonly-used `ArrayList`?

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
```diff
@@ -1,234 +1,371 @@
...
-    @Override
-    synchronized public void register(E e) throws InterruptedException {
-        registeredElements.add(e);
-        liveQueue.putLast(e);
-    }
+    // Lock which guards all accesses to the underlying queue and registered
+    // element set.  Constructed as "nonfair" for performance, like the lock
+    // that a `LinkedBlockingDeque` does.  Not clear that we need this
+    // to be a "fair" lock.
+    private final ReentrantLock lock = new ReentrantLock();
```
KevinCorcoran: If it's important that this lock be non-fair (and I infer that it is from the comment immediately above) it is probably better to explicitly specify that in the constructor call (i.e. `new ReentrantLock(false)`) just to make it obvious.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
```diff
@@ -1,234 +1,371 @@
...
+    // is currently locked.
+    private final Condition notWriteLocked = lock.newCondition();
+
+    // Holds a reference to all of the elements that have been registered.
+    // Newly registered elements are also added into the `liveQueue`.
+    // Elements only exist in the `liveQueue` when not currently
+    // borrowed whereas elements that have been registered (but not
+    // yet unregistered) will be accessible via `registeredElements`
+    // even while they are borrowed.
+    private final Set<E> registeredElements = new CopyOnWriteArraySet<>();
```
KevinCorcoran: Why `CopyOnWriteArraySet` instead of the more commonly-used `HashSet`?  Is that because it needs to be thread-safe?

In particular, this line from the javadocs has me a little concerned: "Mutative operations (add, set, remove, etc.) are expensive since they usually entail copying the entire underlying array."

Given that this is a mutable data structure, it seems perhaps a little bad to have a `getRegisteredElements` method which just returns this data structure, allowing any caller to potentially modify it.  I see that this is only used by tests but ... still.  I think it's more idiomatic to return a copy in this kind of method.

aaaand now I realize that you didn't actually introduce or change any of this code in this PR, so feel free to ignore me.  :hand:

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
```diff
@@ -1,234 +1,371 @@
...
+    public JRubyPool(int size) {
+        liveQueue = new LinkedList<>();
+        maxSize = size;
+    }
+
     @Override
-    synchronized public void unregister(E e) throws InterruptedException {
-        registeredElements.remove(e);
+    public void register(E e) {
+        final ReentrantLock lock = this.lock;
```
KevinCorcoran: This is unnecessary and looks kinda weird to my :eyes:

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
```diff
@@ -1,234 +1,371 @@
...
+            final Thread currentThread = Thread.currentThread();
+            do {
+                if (isWriteLockHeldByAnotherThread(currentThread)) {
+                    notWriteLocked.await();
+                } else if (liveQueue.size() < 1) {
+                    notEmpty.await();
+                } else {
+                    item = liveQueue.removeFirst();
+                }
+            } while (item == null);
```
KevinCorcoran: When would `item` be `null`?

KevinCorcoran: Oh, I see, that would be the case when one of the first two branches was used.  My confusion was that the `removeFirst` call is inside the loop.  It might be cleaner to just do something like
```java
    public E borrowItem() throws InterruptedException {
        final ReentrantLock lock = this.lock;
        lock.lock();
        try {
            final Thread currentThread = Thread.currentThread();
            boolean ready = false;
            do {
                if (isWriteLockHeldByAnotherThread(currentThread)) {
                    notWriteLocked.await();
                } else if (liveQueue.size() < 1) {
                    notEmpty.await();
                } else {
                    ready = true;
                }
            } while (!ready);
            return liveQueue.removeFirst();
        } finally {
            lock.unlock();
        }
    }
```

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
```diff
@@ -1,234 +1,371 @@
...
+            InterruptedException {
+        E item = null;
+        final ReentrantLock lock = this.lock;
+        long remainingMaxTimeToWait = unit.toNanos(timeout);
+
+        // `lock.lockInterruptibly()` is called here as opposed to just
+        // `lock.lock` to follow the pattern that the JDK's
+        // `LinkedBlockingDeque` does for a timed poll from a deque.  See:
+        // http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/concurrent/LinkedBlockingDeque.java#l516
+        lock.lockInterruptibly();
```
KevinCorcoran: Is it weird that `lockInterruptibly` isn't used elsewhere, especially `borrowItem`?

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
KevinCorcoran: Oh, I see, this is copied outta `LinkedBlockingDeque`.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
```diff
@@ -1,234 +1,371 @@
...
     @Override
     public void clear() {
-        liveQueue.clear();
+        final ReentrantLock lock = this.lock;
+        lock.lock();
+        try {
+            int queueSize = liveQueue.size();
+            for (int i=0; i<queueSize; i++) {
+                registeredElements.remove(liveQueue.removeFirst());
+            }
```
KevinCorcoran: Instead of this loop, could it just be something like this?
```java
liveQueue.clear();
registeredElements.clear();
```

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
```diff
@@ -1,234 +1,371 @@
...
+    // Maximum size that the underlying queue can grow to.
+    private int maxSize;
 
+    // Thread which currently holds the write lock.  null indicates that
+    // there is no current write lock holder.  Using the current Thread
+    // object for tracking the lock owner is comparable to what the JDK's
+    // `ReentrantLock` class does:
+    //
+    // http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/concurrent/locks/ReentrantLock.java#l164
+    private Thread writeLockThread = null;
```
KevinCorcoran: Maybe drop the `write` prefix from this field name (and associated method names) since, AFAICT, this class no longer has the concept of a "read lock" vs. "write lock" - it's just got the one lock.

KevinCorcoran: Disclaimer: my knowledge of the `volatile` keyword is super fuzzy.

That said, I think this needs to be marked as `volatile` to prevent pre-thread caching.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
```diff
@@ -1,234 +1,371 @@
...
+        // pool to be unlocked.
+        notWriteLocked.signalAll();
+        // Borrowers that are woken up when an instance is returned to the
+        // pool and the write lock is held would then start waiting on a
+        // 'notWriteLocked' signal instead.  Re-signalling 'notEmpty' here
+        // allows any borrowers still waiting on the 'notEmpty' signal to be
+        // reawoken when the write lock is released, compensating for any
+        // 'notEmpty' signals that might have been essentially ignored from
+        // when the write lock was held.
+        if (liveQueue.size() > 0) {
```
KevinCorcoran: Could this check be eliminated?  Is there any harm in always signaling?

KevinCorcoran:
Overall, this looks good to me.  I left lots of questions/nitpicks but the only one that's terribly important is if `writeLockThread` should `volatile`.  Otherwise, :see_no_evil:

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
camlow325: > 1) Can this type be just List instead of LinkedList or does the implementation depend on some methods on LinkedList?

The specific methods the implementation depends upon come from the `Deque<E>` interface.  I'll change the definition to use that instead to make it a little more generic.

> 2) If the answer to that is "yes", why LinkedList instead of the more commonly-used ArrayList?

The properties of being able to insert an element at the front and back of the queue are what I believe had led us to migrating from having this structure be a queue to a deque - currently as a `LinkedBlockingDeque`.  Migrating the concurrency protection up into the `JRubyPool` class, I wanted to go with a non-synchronized backing class as a replacement, to avoid the performance overhead of double-synchronization.  I had been thinking a `LinkedList` would be better than an `ArrayList` because the operation of inserting an item at the beginning of the collection, which is what we most commonly do when returning a borrowed element to the pool, could be significantly slower with the `ArrayList` since it would involve shifting all of the existing elements down to accommodate the new one.

I hadn't, however, considered `ArrayDeque` instead, added in Java 6.  `ArrayDeque` satisfies the `Deque<E>` interface and, according to a few references I've seen, seems to be superior to a `LinkedList` from a performance perspective for the general operations we would use.  From https://docs.oracle.com/javase/tutorial/collections/implementations/deque.html:

> In terms of efficiency, ArrayDeque is more efficient than the LinkedList for add and remove operation at both ends. The best operation in a LinkedList implementation is removing the current element during the iteration. LinkedList implementations are not ideal structures to iterate.  The LinkedList implementation consumes more memory than the ArrayDeque implementation.

`ArrayDeque` might be a worse choice than `LinkedList` if we were having to resize / copy content frequently but where our queue is fixed size (at least today), `ArrayDeque` might be a better choice.  Maybe if we need to do dynamic resizing later and the need for / cost of resizing the `ArrayDeque` frequently were prohibitively expensive, it would be best to go with a `LinkedList`.  But we could wait until later if that turned out to be an issue.

I'll try out `ArrayDeque` instead and go with that instead if it looks workable.  What do you think?

cprice404: I don't understand how `ArrayDeque` would be better at double-ended insertions, and I also doubt that the difference between these two data structures is going to have enough of a performance difference to warrant spending more time on this PR given everything else we have to do.  I'd be OK with changing references to use the interface where appropriate, but anything beyond that seems like an optimization that we can afford to wait on.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
camlow325: I don't consider it important that the lock be non-fair.  Functionally, I don't see a problem with either a non-fair or fair lock.  Where the performance with a non-fair lock is reputed to be better, though, I figured it would be best to just go with that.  I can add an explicit "false" parameter to make it more obvious which behavior we are getting, though.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
cprice404: The `write` prefix is appropriate, this is a conceptually different thing from the lock that guards the methods in the data structure.

I don't *think* that `volatile` is necessary since all of the read operations for this variable are guarded by the same lock that the write operation uses...

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
camlow325: Yeah, @cprice404 and I debated this a bit here - https://github.com/puppetlabs/puppet-server/commit/5548d980f4d1e9037bbe5042a5c3693b7a375b23#commitcomment-14884322.  I was also leaning toward using a `HashSet` instead with an invocation of the `getRegisteredElements` method doing a synchronized clone.  We couldn't convince ourselves that there would be a significant benefit to doing that though, so we just kept it the way it is.  Does keeping it as it is sound reasonable to you, @KevinCorcoran?

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
camlow325: I've seen some comments on StackOverflow posts about this, like from http://stackoverflow.com/questions/6163166/why-is-arraydeque-better-than-linkedlist:

> @bestsss why would adding/removing on the head be worse off for the LinkedList?
> @DavidT. b/c it involves GC costs of the freed node, assigning the head may also require card marking (for the GC again, if the LinkedList is already in the tenured gen)... and that's on top of the extra indirection (cache-miss) to return the element and relink.

But fair enough concern about what benefit in performance we might actually see from going with this instead.  I'll leave the concrete implementation as `LinkedList` for now but update the declaration to use the `Deque<E>` instead.  Can revisit this later if needed to boost performance.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
camlow325: Yeah, just trying to be idiomatic per how the Java concurrency class implementations are written.  There may be some subtle benefits to this that I don't understand but this didn't seem harmful in any event.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
camlow325: I could go with that instead.  Not sure if there's a big benefit for readability there.  @cprice404, what do you think?

cprice404: Doesn't make much difference to me, so I'd probably err on the side of leaving it alone so we can get this closed out sooner, since it's been open for a while.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
camlow325: I tried to describe why I did that with the comment above.  Essentially, this is again copying from a pattern in `LinkedBlockingDeque`.

For the methods that involve performing an operation with a timeout, `lock.lockInterruptibly()` is called:

http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/concurrent/LinkedBlockingDeque.java#l516

For the methods that don't supply a timeout, however, `lock.lock()` is used.  For example:

http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/concurrent/LinkedBlockingDeque.java#l488

I haven't found any other references in source code that go into further detail about why the implementation works this way.

Part of the rationale might be that if you supplied a timeout that you want the implementation to return a failure sooner if necessary.  Ideally, this wouldn't be noticeable anyway, though.  The `lock` should only be held for a very short period of time and so the benefit of failing `lock` attempts sooner in the event of an interrupt request - which I don't know that we have a reliable way to trigger in Puppet Server today anyway - would seem dubious at best.

Again, I was just thinking here that it was best to follow the JDK's example for this.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
cprice404: If this is how the current implementation works and it hasn't been causing us problems, then I'd suggest that we push any proposed improvements to a follow-up ticket

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
camlow325: @cprice404 also mentioned this here - https://github.com/puppetlabs/puppet-server/commit/36ccd4e6cb3ceb9ddf408aea5d894d81095af7b9#commitcomment-14854753.

Basically, I was thinking of the scenario where an instance may have been borrowed from the `liveQueue` at the time `clear` was called.  It would seem weird to me to have `clear` remove the associated element from `registeredElements` for the currently borrowed item and later possibly have that borrowed item inserted back into the `liveQueue` with no associated element available in `registeredElements` anymore.  There are probably a number of different approaches for dealing with this, but I thought it might be simplest to only remove the `registeredElements` that are likewise being removed from the `liveQueue`.

FWIW, I don't think the `clear` method is even used in any production code and the only test code reference I'm aware of is in the test I added for it in this PR.  I only kept it here because it exists in current code.

Does that make sense?

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
camlow325: > I don't think that volatile is necessary since all of the read operations for this variable are guarded by the same lock that the write operation uses...

Yeah, I waffled on this a bit as well.  At one point, I had a `volatile` keyword in the declaration.  I took it out later when I saw that the JDK class which backs `ReentrantLock`, which does this same sort of thing to track an owner thread among callers synchronized from various threads, likewise does not use the `volatile` keyword.  See:

http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/concurrent/locks/AbstractOwnableSynchronizer.java#l64

This thread object is declared as `transient`.  But I think that is only because the class it is held within implements `java.io.Serializable` and the contents of the original thread object isn't something that would make sense to survive across serialization / deserialization.  Since the `JRubyPool` class doesn't implement `java.io.Serializable` - and I don't think we have a need for it to either - it doesn't seem like adding `transient` would be necessary either.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
camlow325: No, I wouldn't see any harm with that.  This just eliminates a little extra overhead in unnecessarily waking up borrowers that wouldn't be able to do anything but go right back to sleep in the case that the queue might be empty when the write lock is released.  I don't have a strong opinion on this one.  If @cprice404 feels strongly that this would be an improvement, I could eliminate the extra if check.

cprice404: The check seems like a lesser evil than falsely triggering context switches in the JVM, but I don't have a strong opinion

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
camlow325: >  I can add an explicit "false" parameter to make it more obvious which behavior we are getting, though.

Did this in d9e4bcf.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
camlow325: > I'll leave the concrete implementation as LinkedList for now but update the declaration to use the Deque<E> instead.

Did this in d9e4bcf.

camlow325:
@cprice404, @KevinCorcoran I think I've hit all of the feedback so far.  Thanks so much for that!  Anything else you suggest we look at before merging?  You okay with the tests as written?

cprice404:
@camlow325 I still haven't really reviewed the tests.  That's pretty much at the top of my list today after e-mail.

Re: the volatile thing, do you agree with my assessment that all of the possible reads are happening in places where we know that a lock will be held, and that it's the same lock that is held for the write?

Also I might need to dust off my knowledge of the JMM and validate that that ^^ (holding the same lock) is sufficient to address @KevinCorcoran 's question, and that a `synchronized` block isn't necessary to trigger the JMM guarantees that I'm thinking about.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
KevinCorcoran: Sure, we can leave this alone for now.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
KevinCorcoran: I don't really understand the JVM stuff about `per-thread` caching ... it's a little spooky, but if the JDK didn't use `volatile` in the same situation, then I guess we probably don't need it either.

@cprice404, I don't understand your statement about the `write` prefix.  To make my point another way: nothing in the `LockablePool` interface uses the word "write" - the pool is just "locked".  So, why does this implementation use the term "write lock"?  It seems like it is the *only* lock, or maybe the "pool lock".

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
KevinCorcoran: word

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
KevinCorcoran: Yeah, makes sense.  Perhaps it would be good to add a comment with that explanation.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
KevinCorcoran: I don't really care, was mostly just curious.

KevinCorcoran:
I also did not look at the tests.  Production code looks :shipit: to me.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
cprice404: @KevinCorcoran this isn't a lock.  It's an indicator of which thread, if any, holds a conceptual write lock on the pool.  It's distinct from the lock that is used to guard access to the mutators, which is just an implementation detail that doesn't actually have anything to do with the conceptual locking of the pool as far as API consumers are concerned.

camlow325: FWIW, I added "write" into the name per @cprice404's suggestion here - https://github.com/puppetlabs/puppet-server/commit/5548d980f4d1e9037bbe5042a5c3693b7a375b23#commitcomment-14854496.  Mostly was supposed to help disambiguate between the `sharedLock` that gets used for any operation - which might involve taking a conceptual "write lock" or a borrow - and the conceptual "write lock" used to hold off borrowers.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
camlow325: k, I'll do that.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
camlow325: @cprice404, @KevinCorcoran - I could go with "poolLockThread" instead if that sounds better.  No strong preferences on my part.

cprice404: Since it's only ever used to indicate that a write lock is held, I don't see a good reason not to name it that :P

camlow325:
> Re: the volatile thing, do you agree with my assessment that all of the possible reads are happening in places where we know that a lock will be held, and that it's the same lock that is held for the write?

@cprice404, yeah, that was my take as well.  And the code that I saw in the JDK's `ReentrantLock` and backing `AbstractOwnableSynchronizer` where the "owner thread" is tracked as a variable not marked with `volatile` also led me to believe that this is a safe thing to do.

That said, I'm not averse to adding `volatile` to the variable.  It probably wouldn't hurt anything from a functionally perspective - although it may hurt performance?

@cprice404, what do you think?

cprice404:
> And the code that I saw in the JDK's ReentrantLock and backing AbstractOwnableSynchronizer where the "owner thread" is tracked as a variable not marked with volatile also led me to believe that this is a safe thing to do.

Does that code also use a common lock object to guard all of the read/write access?  If so then I say we call this good as-is.

camlow325:
> Does that code also use a common lock object to guard all of the read/write access? If so then I say we call this good as-is.

Essentially, yes.  It manages whether the "lock" is held or not as a volatile int called `state` - http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/concurrent/locks/AbstractQueuedSynchronizer.java#l533.  That `state` variable is generally checked before the lock thread object is accessed.

The lock thread object is set (http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/concurrent/locks/AbstractQueuedSynchronizer.java#l212) or cleared (http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/concurrent/locks/AbstractQueuedSynchronizer.java#l222) without a secondary wrapping lock / synchronized block once the current state has been established.  The lock thread object itself isn't marked volatile but could be accessed from different threads over time.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
camlow325: Added in 0380cb1.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
KevinCorcoran: I understand that this is not the actual lock itself.

What I'm taking issue with is use of the word "write" anywhere in this file; my point is that, since there are no longer any "read" locks, calling this a "write" lock is misleading.

>  It's distinct from the lock that is used to guard access to the mutators, which is just an implementation detail that doesn't actually have anything to do with the conceptual locking of the pool as far as API consumers are concerned.

Yup, I get that this is distinct from the `ReentrantLock` used elsewhere in this class; however, this thread is the thread which called `lock` on the pool, which doesn't necessarily have anything to do with "writing", does it?  It's that word, "write", which bothers me - it doesn't show up in the interface so it seems out of place here too.

Don't feel like you need to help up merging this PR on behalf of my nitpick about naming.

cprice404:
@camlow325 after reading through the links you pasted above, and reading through some of the source and some docs about the JMM, I am now leaning towards saying that we *should* mark that variable as volatile, as @KevinCorcoran suggested.

Reading the source code of those `AbstractSynchronizer` classes, particularly the comments on methods like this one:

http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/concurrent/locks/AbstractOwnableSynchronizer.java#l73

Leads me to believe that these classes are intended to be used from other classes that *do* use `synchronized` blocks.  I think that without those `synchronized` blocks, there would be a data race on that `exclusiveOwnerThread` variable.  ( https://www.cs.umd.edu/~pugh/java/memoryModel/jsr-133-faq.html#incorrectlySync ).

I was thinking that maybe the ReentrantLock that you are using to lock things would trigger the same behavior as a `synchronized` block, but I'm somewhat doubting that now, and I also don't have a good understanding of how the `synchronized` semantics would work in combination with the `await` / `signal` stuff we're now relying on.  :(

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
cprice404: OK, I understand what you are saying now.  I do still think it is important to choose names that clearly distinguish the 'concurrent access to mutators' lock from anything that has to do with the objects / methods that are used to manage the conceptual pool lock that is exposed via the API.  So, IMO, we would probably need to come up with a better name for the class variable that is currently called simply `lock` if we are going to change the other names.

KevinCorcoran:
If there is any doubt, it's probably best to just go ahead and mark it `volatile`.  Seems like the trade-off is a likely very small performance hit vs. potential for bugs which may be extremely difficult to debug.

cprice404:
@KevinCorcoran still think no one needs to read JCIP?  :P

camlow325:
@cprice404 I think if the owner thread needed to be marked `volatile` in order to avoid the problems with accessing partially written values across thread that the `LinkedBlockingDeque` implementation that we are using in Puppet Server today would already be broken.  It relies upon a `ReentrantLock`, which uses the `AbstractOwnableSynchronizer` with the underlying non-volatile write thread object to manage the lock owner.  I think the implementation is made safe from these problems by guarding access to the write thread object with prior accesses of the volatile `state` object.  AFAIK, we haven't seen problems with the pool queue getting periodically corrupted, which seems like a possible consequence of failing to use volatile properly here.  Maybe we've just been incredibly lucky so far?

All that said, though, I don't have any hard evidence of what the effect of marking the thread object as `volatile` would be.  I suppose I could just do that.  Do you suggest that we do another Gatling run with this change or just go with it and see what happens?

test/integration/puppetlabs/services/jruby/jruby_locking_test.clj
```diff
@@ -23,16 +32,15 @@
     [jruby-service/jruby-puppet-pooled-service
      profiler/puppet-profiler-service]
     (jruby-service-test-config 1)
-    (let [jruby-service (tk-app/get-service app :JRubyPuppetService)
-          lock (jruby-testutils/get-lock-from-app app)]
-
+    (jruby-testutils/wait-for-jrubies app)
```
KevinCorcoran: This is unnecessary, since there is only a single JRuby in-use by this test, right?

test/integration/puppetlabs/services/jruby/jruby_locking_test.clj
```diff
@@ -89,24 +98,35 @@
...
+                       :with-lock-and-borrow-contention-test)]
+         (let [lock-thread-started? (promise)
+               unlock-thread? (promise)
+               lock-thread (future (jruby-service/with-lock
+                                    jruby-service
+                                    :with-lock-and-borrow-contention-test
+                                    (deliver lock-thread-started? true)
+                                    @unlock-thread?))]
+           (testing "lock not granted yet when instance still borrowed"
+             (Thread/sleep 500)
```
KevinCorcoran: This seems bad.  It seems like this should be replaced with `(deref lock-thread-started?)`

KevinCorcoran: Oh, oops, never mind, seems like that won't work ...

cprice404:
>  I think the implementation is made safe from these problems by guarding access to the write thread object with prior accesses of the volatile state object. 

What I'm worried about is caching, like, at the level of an individual processor.  If a non-volatile variable is written to, outside of a `synchronized` block, by one thread, then it is possible for the write to be done at the level of a CPU cache, and another thread could then read the variable on a different CPU and not see the new value.  It doesn't seem like the `state` objects would prevent this possibility of a stale value being read.

Perhaps the semantics of the class are resilient to the case where a stale value is read.  Otherwise I have no good explanation for why `volatile` wouldn't be necessary in `LinkedBlockingQueue`.

>  Do you suggest that we do another Gatling run with this change or just go with it and see what happens?

How much effort is it to do a run?

test/integration/puppetlabs/services/jruby/jruby_locking_test.clj
KevinCorcoran: Is `Thread/sleep` necessary?

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
```diff
@@ -167,54 +207,332 @@
...
+        (let [instance (.borrowItem pool)]
+          (is (true? true))
+          (.releaseItem pool instance))
 
         (is (true? true))
         (.unlock pool)
         @borrow-thread-1
         @borrow-thread-2))))
 
+(deftest pool-lock-reentrant-for-one-lock-test
```
KevinCorcoran: This test seems to be subsumed by the next one.

camlow325:
> How much effort is it to do a run?

I'll make the change and get another run setup.  Probably take a day to get the results back.

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
camlow325: I can eliminate this one.  I sometimes like to have a discrete "1 instance" vs. a "many instance" test since there might be a problem in the implementation which is brought out in one case but not the other.  I didn't have any specific code paths in mind in this situation, though.

test/integration/puppetlabs/services/jruby/jruby_locking_test.clj
camlow325: This was just to wait a bit longer to ensure that we didn't get into code on the future that shouldn't have been run because the lock shouldn't have been acquired yet.  Yeah, this line could be eliminated but it would make it a bit more likely that if the lock actually were erroneously granted early that the following assertion wouldn't catch it.  Although I guess that could happen anyway - like if we waited for 5 minutes and the lock were erroneously granted after 5 minutes and 1 second.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
camlow325: What if we went with something like `borrowLockThread` - or something similar that works the word "borrow" into the name?  Seems like the thing we are fundamentally locking is the ability for any threads other than the locker to do borrows.

cprice404: I'd be OK with `borrowLock` or `poolLock` if we also come up with a better name for the thing that is currently just called `lock`.  Maybe... `dequeLock`?

camlow325: >  Maybe... dequeLock?

Yeah, I suppose that would work - although it's also technically a lock around the `registeredElements`, too.

KevinCorcoran:
I reviewed the test changes.  Very thorough!  Didn't see any big problems, just a few nits to pick, as always.  :+1:

test/integration/puppetlabs/services/jruby/jruby_locking_test.clj
KevinCorcoran: Yeah.  And half a second seems like a long time to wait.  Also it seemed like maybe there were other cases in the unit tests where the same reasoning applies, but `Thread/sleep` was not used in those cases ... but I could be wrong about that.

camlow325: Yeah, I think there are other places where this could apply but where we haven't done the `sleep`.  I'll just remove the line for the sake of keeping the tests more snappy.  Hopefully the test will get to the `with-lock` call on the `future` often enough before the assertion occurs that we'll get some value from the assertion.

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
camlow325: Removed in `864f9ed`.

test/integration/puppetlabs/services/jruby/jruby_locking_test.clj
camlow325: Removed in 864f9ed.

test/integration/puppetlabs/services/jruby/jruby_locking_test.clj
camlow325: Agreed.  Removed in 864f9ed.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
cprice404: `dataStructureLock`?  I don't know, I'm flailing here.  @KevinCorcoran do you have any ideas

test/integration/puppetlabs/services/jruby/jruby_locking_test.clj
```diff
@@ -89,24 +96,34 @@
...
+       (let [instance (jruby-protocol/borrow-instance
+                       jruby-service
+                       :with-lock-and-borrow-contention-test)]
+         (let [lock-thread-started? (promise)
+               unlock-thread? (promise)
+               lock-thread (future (jruby-service/with-lock
+                                    jruby-service
+                                    :with-lock-and-borrow-contention-test
+                                    (deliver lock-thread-started? true)
+                                    @unlock-thread?))]
```
cprice404: nit, but couldn't these three let blocks be collapsed into one?

test/integration/puppetlabs/services/jruby/jruby_locking_test.clj
```diff
@@ -89,24 +96,34 @@
...
+     app
+     [jruby-service/jruby-puppet-pooled-service
+      profiler/puppet-profiler-service]
+     (jruby-service-test-config 2)
+     (jruby-testutils/wait-for-jrubies app)
+     (let [jruby-service (tk-app/get-service app :JRubyPuppetService)]
+       (let [instance (jruby-protocol/borrow-instance
+                       jruby-service
+                       :with-lock-and-borrow-contention-test)]
+         (let [lock-thread-started? (promise)
```
cprice404: maybe rename this var to `lock-acquired?`?  technically the thread has been started before then.

/me ducks

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
KevinCorcoran: I think `poolLockThread` is a pretty good name for this variable.

For the other variable ... hmm ... that one is harder ... I'd prefer something with the word "queue" instead of "deque" to match the convention established by the other variables in this class, like `liveQueue`.  So, some ideas ...
* `queueLock`
* `queueAndElementSetLock`
* `elementsLock`
* `internalLock`

And just to be contrarian, because why not, I don't hate just leaving it as `lock` since it is fundamental to the implementation, and that's what the JDK classes do (I think), and `poolLockThread` seems to distinguish it from this lock.

If I had to choose it would probably be `lock` or `queueLock`.

camlow325:
> Perhaps the semantics of the class are resilient to the case where a stale value is read.  Otherwise I have no good explanation for why volatile wouldn't be necessary in LinkedBlockingQueue.

Yeah, I think the semantics are what allow this implementation to work.

The owner thread only gets set to that of the current thread if the current thread is first successful in setting the `state` variable to the number of acquires it desires.  The `state` variable set is atomic, via the use of `volatile`.

After the locking thread resets the owner thread to null at release time, the `state` variable is reset to 0 to indicate that no acquires are held.

It does seem possible that a thread trying to acquire the lock might interpret that the thread is still locked if it tries to check `state` after the owning thread has set the owner thread object to null but before `state` is reset.  In that case, it seems like the thread trying to obtain the lock would just end up waiting a bit longer before getting the lock.

In any event, if we assume that `ReentrantLock` is doing what is supposed to be doing, only one thread should be able to hold the lock at a time and, therefore, only one thread should be trying to access the value of the `writeLockThread` variable in the `JRubyPool` class at a time.  I still think that the `LinkedBlockingDeque` would inherently have problems if `ReentrantLock` weren't managing its underlying `exclusiveOwnerThread` object in such a way that more than one thread were to think that it was holding the lock at a time.

I fired off some more performance tests using `volatile` in the variable declaration.  If that doesn't show any significant changes, I'm good just going with that for now.

test/integration/puppetlabs/services/jruby/jruby_locking_test.clj
camlow325: Yeah, I think they could.  Left over from when there were assertions in the middle of those, I guess.  I'll change that.

test/integration/puppetlabs/services/jruby/jruby_locking_test.clj
camlow325: Yeah, that name would be better here.  Will update that.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
camlow325: @cprice404 and @KevinCorcoran I'll just go with `poolLockThread` and `queueLock` if that works for you.

test/integration/puppetlabs/services/jruby/jruby_locking_test.clj
```diff
@@ -15,7 +15,16 @@
 (defn jruby-service-test-config
   [pool-size]
   (jruby-testutils/jruby-puppet-tk-config
-    (jruby-testutils/jruby-puppet-config {:max-active-instances pool-size})))
+    (jruby-testutils/jruby-puppet-config {:max-active-instances pool-size
+                                          :borrow-timeout 1})))
```
cprice404: is this milliseconds?  Is there any risk of setting it so low causing transient failures?  Or are we safe because we're never attempting multiple simultaneous borrows in this test?

test/integration/puppetlabs/services/jruby/jruby_locking_test.clj
```diff
@@ -89,24 +96,34 @@
...
+                                    (deliver lock-thread-started? true)
+                                    @unlock-thread?))]
+           (testing "lock not granted yet when instance still borrowed"
+             (is (not (realized?
+                       lock-thread-started?))))
+           (jruby-protocol/return-instance
+            jruby-service instance :with-lock-and-borrow-contention-test)
+           @lock-thread-started?
+           (testing "cannot borrow from non-locking thread when locked"
+             (is (not (can-borrow-from-different-thread? jruby-service))))
```
cprice404: My reading of the implementation of `can-borrow-from-different-thread?` makes it look like it will block until the timeout fires here, so presumably we're relying on the timeout being really low.  I guess that's OK here because there are two JRubies in the pool so we should be able to know for sure that one is available when this line is hit?

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
```diff
@@ -60,6 +85,27 @@
...
+
+        (testing (str "other threads may successfully unregister instances "
+                      "while pool.lock() is being executed")
+          (.unregister pool (first instances))
+          (is (not (realized? lock-thread)))
+          (.unregister pool (second instances)))
+
+        @lock-thread
+        ;; make sure we got here
+        (is (true? true))))))
```
cprice404: This test seems legit, but, I'm curious as to whether you had a use-case in mind where you expected this behavior to be tickled.  Also, if we're testing this, should we also test for expected behavior for the case where the last registered instance is unregistered while a lock request is pending?

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
```diff
@@ -167,54 +207,314 @@
...
+      (is (= 1 (.size pool)))
       (.releaseItem pool instance false)
-      (is (= 0 (.size pool)))
+      (.unregister pool instance)
+      (is (= 1 (.size pool)))
       (.lock pool)
-      (is (.isWriteLocked lock))
+      (is (nil? @(future (.borrowItemWithTimeout pool
+                                                 1
+                                                 TimeUnit/MICROSECONDS))))
```
cprice404: 1 microsecond?!  :)  I presume the implementation doesn't start the timer until it determines that an item is not immediately available?  Otherwise that seems frighteningly short :)

cprice404: Other thing I was going to mention is that it seems like this pattern of specifying an explicit timeout with the borrow operation might be better than the more global `borrow-timeout: 1` that you put in the config of one of the other tests, but, not sure that's actually a big deal.

camlow325:
> I fired off some more performance tests using volatile in the variable declaration. If that doesn't show any significant changes, I'm good just going with that for now.

Well, not only did I not see any degradation with this set of tests but instead the mean time per agent run with my test setup actually showed about a ~120ms improvement with the `volatile` keyword applied to the thread object vs. previous experiments which did not include the `volatile` keyword.  I saw about the same mean time per agent run on two different bare metal machines using similar specs.  The runs were 8 hours, simulating a scale up to about 1000 agents.  The standard deviation per agent run for each of these experiments is in the 300 - 350 ms range, so the difference may not be statistically meaningful.

I'll go ahead and include the `volatile` keyword in a follow-up commit.

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
```diff
@@ -167,54 +207,314 @@
...
+          borrow-thread-borrowed? (promise)
+          borrow-thread (future (deliver borrow-thread-started? true)
+                                (let [instance (.borrowItem pool)]
+                                  (deliver borrow-thread-borrowed? true)
+                                  (.releaseItem pool instance)))]
+      @borrow-thread-started?
+      (is (not (realized? borrow-thread-borrowed?)))
+      (.releaseItem pool item)
+      @borrow-thread
+      (is (true? true)))))
```
cprice404: Just making sure I understand: this test doesn't really have anything to do with your new changes, it's just some additional validation of the expected behavior of the pool overall, right?

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
```diff
@@ -167,54 +207,314 @@
...
+            borrow-thread-borrowed? (promise)
+            borrow-thread (future (deliver borrow-thread-started? true)
+                                  (let [instance (.borrowItem pool)]
+                                    (deliver borrow-thread-borrowed? true)
+                                    (.releaseItem pool instance)))]
+        @borrow-thread-started?
+        (is (not (realized? borrow-thread-borrowed?)))
+        (.releaseItem pool item)
+        @borrow-thread
+        (is (true? true))))))
```
cprice404: Did you have a situation in mind where this kind of test might fail?

test/integration/puppetlabs/services/jruby/jruby_locking_test.clj
camlow325: If an instance is available in the queue and the timeout is anything greater than 0, the instance will be returned.  The wait timer doesn't start unless/until it is determined that no instance is immediately available to be borrowed.

test/integration/puppetlabs/services/jruby/jruby_locking_test.clj
camlow325: It doesn't make any specific assumptions about the timeout other than that there is one.  The `borrow-timeout` is configured to 1 millisecond in this case, which allows the test to run quickly, but it seems like this test would work fine with a much larger timeout.  We know that there are 2 JRubies in the pool by this point because of the `(jruby-testutils/wait-for-jrubies app)` call earlier on, which loops until all of the expected instances are available in the pool.

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
camlow325: I was thinking something similar to this behavior could be tickled if a .lock() request were made while Puppet Server is in the process of flushing an instance that had hit its max-requests counter.  In this case, the old instance would just be unregistered but not put back into the pool.  I was looking to ensure that the .lock() would still succeed after instances had been unregistered.  This test did help me find a reference counting bug in my initial implementation.

I agree with the idea about adding a test for when the last registered instance is unregistered.  I'll do that.

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
```diff
@@ -167,54 +207,314 @@
...
+
+(deftest pool-clear-test
+  (testing (str "pool clear removes all elements from queue and only matching"
+                "registered elements")
+    (let [pool (create-populated-pool 3)
+          instance (.borrowItem pool)]
+      (is (= 2 (.size pool)))
+      (is (= 3 (-> pool
+                   (.getRegisteredElements)
+                   (.size))))
```
cprice404: nit but you could do this as just `(.. pool getRegisteredElements size)`

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
camlow325: > 1 microsecond?! :) I presume the implementation doesn't start the timer until it determines that an item is not immediately available?

Yep, that's right.

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
```diff
@@ -167,54 +207,314 @@
...
+      (.clear pool)
+      (is (= 0 (.size pool)))
+      (let [registered-elements (.getRegisteredElements pool)]
+        (is (= 1 (.size registered-elements)))
+        (is (identical? instance (-> registered-elements
+                                     (.iterator)
+                                     iterator-seq
+                                     first)))))))
```
cprice404: This test seems valuable in terms of validating current behavior, but I'm not sure whether this behavior is actually what we'd desire to have happen here.  WDYT?  It seems like `clear` should not behave this way, but, if we're only using it in tests, then I'm probably +1 on just having this test assert that the behavior is weird and we can file a separate ticket about improving the actual behavior in the future if it becomes necessary?

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
camlow325: > Other thing I was going to mention is that it seems like this pattern of specifying an explicit timeout with the borrow operation might be better than the more global borrow-timeout: 1 that you put in the config of one of the other tests

That makes sense.  I did the `borrow-timeout: 1` setting in the app configuration for the jruby_locking_test tests because those tests go through the JRuby service to borrow instances and there's no function on that service that allows the timeout to be specified per request.  I think for the lockable-pool-tests that I specified a timeout in the `.borrowItemWithTimeout` call directly where a timeout was needed.  I had the ability to do that at this level since these tests use the `JRubyPool` class directly.

cprice404:
Finished my first pass at reviewing the test code, looks good other than the (mostly nitpick) comments that I left.  Will go back and respond to the latest comments that have come in, ASAP.

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
cprice404: OK, that makes sense... I think your current approach is probably the best way to handle those, then.

~legit

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
camlow325: Yeah, I don't think this found a problem with my implementation.  Just some additional validation of the overall behavior.

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
camlow325: Doesn't look like the code changes in the commit where this test came in were related but I'm pretty sure this test was failing earlier on with the code changes on my branch - not on the current master branch.  There were a few tests like this that I added for cases where some of the condition variables weren't being signaled appropriately for waiters when needed, leading to borrow requests being stuck indefinitely.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
cprice404: sounds good to me.  naming fatigue

test/integration/puppetlabs/services/jruby/jruby_locking_test.clj
cprice404: ~legit

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
camlow325: @KevinCorcoran and I also talked about how `clear` should work here - https://github.com/puppetlabs/puppet-server/pull/813#discussion_r48579770.  I made the argument that it would be desirable for `clear` to work this way because it would seem wrong for a borrower to return an instance to the pool after the `clear` call was made and end up in the situation where we have an instance in the `liveQueue` that was not also in `registeredElements`.

We don't actually use `clear` in any production code anyway, AFAICT.  I doubt we ever will either since we need to do some custom cleanup processing on any elements that we want to dump from the pool, which involves borrowing the individual instances.

Does this make sense?

test/integration/puppetlabs/services/jruby/jruby_locking_test.clj
cprice404: ~legit

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
cprice404: So we're presuming that if this situation occurs, there should be another `register` call happening shortly?

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
cprice404: k

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
camlow325: Yeah, I would expect that today.  I suppose if we ever were to add in the ability to dynamically size up / down the pool that this attribute would be more useful.

camlow325: Or maybe a better current case would be around the full JRuby pool flush.  In that case, if the lock were requested against the soon-to-be "old" pool, the lock() would be held off until all instances were back in the "old" pool.  I think the logic in the pool flush will borrow the instance from the "old" pool, unregister it without putting it back into the "old" pool, create a replacement instance, and put the replacement instance in the "new" pool.  There is some logic in the pool flush to put a poison pill in the "old" pool, but it only does one rather than one per instance being disposed.  Maybe it would have been better to do one pill per instance being disposed?  In any event, I think the logic to awaken the lock requester after all of the pool instances have been unregistered would prevent the lock request from otherwise getting stuck indefinitely.

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
cprice404: k

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
cprice404: Yeah, that's what I was getting at, is wondering what would happen if the lock was requested and then the last instance was unregistered.  It seems like the only case where that could currently happen is what you're describing, where we're trying to lock the old pool while a full pool flush is in progress, and so the lock should probably fail when the last instance is unregistered... but maybe it needs to happen via some kind of exception that we can catch and recognize so that we can retry the lock acquisition?

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
cprice404: It makes sense.  It seems like the desired behavior would be that `clear` would have to block until it had freed all registered elements, but that is probably not worth dealing with right now.  I'm probably fine leaving this as is.  One other idea, though, might be to simply delete the `clear` method altogether for now?  I suspect that maybe the only reason it was there before was because we were directly inheriting from a queue class that already defined it, and we needed to override the behavior?

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
camlow325: The way the code works in the PR currently, the pending lock request would be granted after the last instance is unregistered.  But, then, the locker would have a lock on a pool that is now empty - at least temporarily, because the poison pill would end up being dumped into the pool right after that.  I don't have any logic in the implementation that would prevent a pill from being inserted while the pool lock is held.  Not sure if we need that.

I'm not sure I agree with the semantics of having a lock request be aborted if instances are unregistered from the pool while the lock request is still pending.  Seems like it still makes sense for the locker to get access to the pool in that case.

This may be bleeding over into the question I'd asked you today about what could happen if a file sync write lock were to occur right around the time that a full JRuby pool flush is being performed.  I think it might be better to deal with that question separately from this PR.  I'll file a separate ticket for that and link that back to this PR.  Sound okay?

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
camlow325: I'm +1 on the idea of just deleting the `clear` method altogether since we don't seem to have a specific need for it.  I'll go ahead and do that if there are no other objections.  @KevinCorcoran, you good with that?

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
cprice404: Yeah, I'm definitely good with deferring any additional changes, just trying to make sure I understand the current semantics.  If you're confident that, after the final unregister, the lock will be granted and there will be a poison pill... agh probably faster to discuss this in IRL

KevinCorcoran: :fast_forward:

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
KevinCorcoran: yes plz!

camlow325: > We don't actually use clear in any production code anyway, AFAICT. 

Well, turns out I was wrong about that.  :disappointed:  There are a couple of places where we use it in production code:

https://github.com/puppetlabs/puppet-server/blob/puppet-server-2.2.1/src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj#L59

https://github.com/puppetlabs/puppet-server/blob/puppet-server-2.2.1/src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj#L115

In both cases, it is called after an exception is caught when a pool instance fails to be initialized.

I think calling `.clear` in these cases is wrong.  If we were to just do this without calling `.terminate` on any of the ScriptingContainers for the instances that had been created prior to the one that had failed, we would be leaking resources.

I'm thinking it might be best to just file a separate ticket for this one where we'd look into eliminating the `.clear` method and ensuring that any previously created instances had been pulled out of the pool and disposed of properly.  So for this PR, we could just leave the implementation of `.clear` as is.  I don't think the implementation as it stands on the branch currently would be a regression in these cases.  The next attempt to borrow from the pool after either of the code paths above are executed would still return a poison pill to the caller.

cprice404: :dizzy_face:

okay, I agree with your assessment.

camlow325: I filed the follow-up ticket for this as [SERVER-1080](https://tickets.puppetlabs.com/browse/SERVER-1080).

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
camlow325: > Yeah, I'm definitely good with deferring any additional changes, just trying to make sure I understand the current semantics. If you're confident that, after the final unregister, the lock will be granted

I added the suggested test from earlier in c5d08f to confirm that this is the case.

> and there will be a poison pill...

I didn't add anything to this PR to make that happen, but this line of code already in the master branch appears to do that:

https://github.com/puppetlabs/puppet-server/blob/puppet-server-2.2.1/src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj#L122

--

After some additional IRL discussion about this, I filed a follow-up ticket that would cover looking into this further - [PE-13552](https://tickets.puppetlabs.com/browse/PE-13552).  That one will probably need to have a SERVER project ticket spun off of it as well for the open-source work but thought it would be better to start with it as a PE ticket to research the interaction between the pool flush and file-sync pool locking.

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
camlow325: Sounds good.  Did this in c5d08f1.

test/integration/puppetlabs/services/jruby/jruby_locking_test.clj
camlow325: Did this in c5d08f1.

cprice404:
OK I think those new tickets and that last test were all that was outstanding from me, so I think I'm +1 now.  @KevinCorcoran ?

test/integration/puppetlabs/services/jruby/jruby_locking_test.clj
camlow325: Did this in c5d08f1.

src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
camlow325: Did this in e9e42e1.

camlow325:
Marked the `writeLockThread` (later renamed to `poolLockThread`) as `volatile` in 9f64a65.

camlow325:
@cprice404, @KevinCorcoran Yeah, I think I've addressed all of the outstanding comments, either with code changes or separate JIRA tickets filed.  I don't have anything else planned to add onto the PR at this point.

KevinCorcoran:
Nothing more from me on this!

FWIW, I'm glad we got rid of the word "write" in `JRubyPool`.  I think that it'll be easier to understand for any poor souls who venture in there in the future.

:shipit:

cprice404:
OK I'm ready to merge this, just waiting for jank to clear out a bit.  Also, travis failure... hrm

cprice404:
awwww yiss comment record

cprice404:
stupid robot

camlow325:
I think the Travis failure was due to my removal of a wait for the JRuby pool to be ready from the `with-lock-test` before the test performs a borrow.

I had previously removed a couple of those waits from 864f9ed but those waits are actually necessary since the JRuby pool is not guaranteed to have at least one instance in it after the services have been started in the tests.  It would be if a service that required a JRuby instance to be borrowed as part of its service startup, e.g., the PuppetServerConfigService, were in the list of services being bootstrapped.  For the minimal set of services being bootstrapped for these tests, though, pool initialization is asynchronous from the startup of any services and so we need to have a wait in the tests so that they don't try to access the pool before it is ready.

I re-added two JRuby pool waits in 61a88fa.  We can wait and see if these pass through Travis this time...

cprice404:
roger roger

cprice404:
also I think @nathanielksmith and I would both benefit from a voom build before we merge this.

cprice404:
voom is done.

PR #814, opened 2015-12-12
(docs) Revise auth.conf documentation.

gguillotte:
To further emphasize the new supported `trapperkeeper-authorization`
configuration file, format, and methods over the deprecated methods,
divorce them from the deprecated content, elevate them to the primary
part of the `auth.conf` documentation, lower the deprecated content, and
further emphasize the deprecation.

In the process, add more supported settings from the
`trapperkeeper-authorization` documentation to avoid having to point
users away from the Puppet Server documentation.

### DIFF
--- a/documentation/config_file_auth.md
+++ b/documentation/config_file_auth.md
@@ -9,47 +9,187 @@ canonical: "/puppetserver/latest/config_file_auth.html"
 [deprecated]: ./deprecated_features.html
 [`puppetserver.conf`]: ./config_file_puppetserver.html
 
-The `auth.conf` file contains rules for authorizing access to Puppet Server's HTTP API endpoints. For an overview, see [Puppet Server Configuration](./configuration.html).
+Puppet Server's `auth.conf` file contains rules for authorizing access to Puppet Server's HTTP API endpoints. For an overview, see [Puppet Server Configuration](./configuration.html).
 
-The new Puppet Server authentication configuration and functionality is similar to the legacy method in that you define rules in `auth.conf`, and Puppet Server applies the settings when a request's endpoint matches a rule. 
+The new Puppet Server authentication configuration and functionality is similar to the legacy method in that you define rules in a file named `auth.conf`, and Puppet Server applies the settings when a request's endpoint matches a rule. 
 
-However, the new HOCON format provides different parameters and syntax, and the new method provides slightly different functionality. The following sections describe how to configure both methods.
+However, Puppet Server now has its own `auth.conf` file that uses a new HOCON format with different parameters, syntax, and functionality.
 
 > ### Aside: Changes to Authorization in Puppet Server 2.2.0
 > 
 > Puppet Server 2.2.0 introduces a significant change in how it manages authentication to API endpoints. It uses [`trapperkeeper-authorization`][] for authentication, which is configured by rules and settings in Puppet Server's own `auth.conf`, with a HOCON configuration file format in a different location than the [Puppet `auth.conf`][] file.
 >
-> The older Puppet `auth.conf` file and whitelist-based authorization method are [deprecated][]. Puppet Server's new `auth.conf` file, documented below, also uses a different format for authorization rules. Additionally, there are several conditions that can affect whether Puppet Server uses the new or old authorization methods:
+> The older Puppet `auth.conf` file and whitelist-based authorization method are [deprecated][]. Puppet Server's new `auth.conf` file, documented below, also uses a different format for authorization rules.
 >
-> * Requests to Puppet master service endpoints already manageable through the older authorization methods and [Puppet `auth.conf`][] file---such as catalog, node, and report---will use Puppet Server's new `auth.conf` rules **only** if the `use-legacy-auth-conf` setting in `puppet-server.conf` is set to false. Also, Puppet Server warns you that the legacy authentication method is deprecated if `use-legacy-auth-conf` is set to true (which is its default).
-> * Requests to certificate status and administration endpoints will use the new `auth.conf` rules **only** if the corresponding `client-whitelists` setting is empty or unspecified and the `authorization-required` flag is set to true (which is its default).
-> * Requests to other certificate administration endpoints---such as certificate, certificate_request, and certificate_revocation_list---will **always** use the new `auth.conf` rules. This happens regardless of the `client-whitelist`, `authorization-required`, or `use-legacy-auth-conf` settings, as versions of Puppet Server before 2.2.0 can't manage those endpoints.
+> Puppet Server follows the following logic when determining whether to use the new or old authorization methods:
 >
-> **Note:** You can also use the [`puppetlabs-puppet_authorization`](https://forge.puppetlabs.com/puppetlabs/puppet_authorization) module to manage the new `auth.conf` file's authorization rules in the new HOCON format, and the [`puppetlabs-hocon`](https://forge.puppetlabs.com/puppetlabs/hocon).
+> * Requests to Puppet master service endpoints already manageable through the deprecated authorization methods and [Puppet `auth.conf`][] file --- such as `catalog`, `node`, and `report` --- use Puppet Server's new `auth.conf` rules **only** if the `use-legacy-auth-conf` setting in `puppet-server.conf` is set to `false`. If `use-legacy-auth-conf` is set to true (which is its default), Puppet Server warns you that the legacy authentication method is deprecated.
+> * Requests to certificate status and administration endpoints use the new `auth.conf` rules **only** if the corresponding `client-whitelists` setting is empty or unspecified **and** the `authorization-required` flag is set to `true` (which is its default).
+> * Requests to other certificate administration endpoints --- such as `certificate`, `certificate_request`, and `certificate_revocation_list` --- **always** use the new HOCON `auth.conf` rules in Puppet Server's `auth.conf` file. This happens regardless of the `client-whitelist`, `authorization-required`, or `use-legacy-auth-conf` settings, as versions of Puppet Server before 2.2.0 can't manage those endpoints.
+>
+> **Note:** You can also use the [`puppetlabs-puppet_authorization`](https://forge.puppetlabs.com/puppetlabs/puppet_authorization) module to manage the new `auth.conf` file's authorization rules in the new HOCON format, and the [`puppetlabs-hocon`](https://forge.puppetlabs.com/puppetlabs/hocon) module to use Puppet to manage HOCON-formatted settings in general.
 
-## Parameters
+You have two options when configuring how Puppet Server authenticates requests:
 
-Use the following parameters when writing or migrating custom authorization rules.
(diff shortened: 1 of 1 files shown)


gguillotte:
CC @camlow325

documentation/config_file_auth.md
```diff
@@ -9,142 +9,168 @@ canonical: "/puppetserver/latest/config_file_auth.html"
...
 
-Use the following parameters when writing or migrating custom authorization rules.
+* If you opt into using Puppet Server's new, supported HOCON `auth.conf` file and authorization methods, use the parameters and rule definitions in the [HOCON Parameters](#hocon-parameters) section.
+* If you continue using the deprecated Ruby [Puppet `auth.conf`][] file and authorization methods, see the [Deprecated Ruby Parameters](#deprecated-ruby-parameters) section.
 
-### `path`
+## HOCON Parameters
 
-Rules with a `path` parameter apply only to endpoints with URLs that start with the parameter's value.
+Use the following parameters when writing or migrating custom authorization rules using the new HOCON forat.
```
camlow325: s/forat/format

documentation/config_file_auth.md
```diff
@@ -9,142 +9,168 @@ canonical: "/puppetserver/latest/config_file_auth.html"
...
-path ~ ^/puppet/v3/report/([^/]+)$
+If this setting is `true`, Puppet Server ignores any presented certificate and relies completely on header data to authorize requests. 
 
-# Literal string to match at the start of a URL's path.
-path /puppet/v3/report/
-~~~
+> **Warning!** This is very insecure; **do not enable this parameter** unless you've secured your network to prevent **any** untrusted access to Puppet Server.
 
-### `method`
+You cannot rename any of the `X-Client` headers when this setting is enabled, and you must specified identity through the `X-Client-Verify`, `X-Client-DN`, and `X-Client-Cert` headers.
```
camlow325: s/specified/specify

documentation/config_file_auth.md
```diff
@@ -9,142 +9,168 @@ canonical: "/puppetserver/latest/config_file_auth.html"
...
 
-~~~
-# Use GET and POST.
-method: [get, post]
+The required `rules` array of a Puppet Server's HOCON `auth.conf` file determines how Puppet Server responds to a request. Each element is a map of settings pertaining to a rule, and when Puppet Server receives a request, it evaluates that request against each rule looking for a match.
 
-# Use PUT.
-method: put
-~~~
+You define each rule by adding parameters to the rule's [`match-request`](#match-request) section. A `rules` array can contain as many `match-request` sections as you need.
```
camlow325: Would it be better to say "A `rules` array can contain as many rules as you need"?  I read this initially as indicating that you can add more than one `match-request` section to an individual rule as opposed to meaning that you can have multiple rules but each with only one `match-request` section.  The implementation only supports the latter.

documentation/config_file_auth.md
```diff
@@ -9,142 +9,168 @@ canonical: "/puppetserver/latest/config_file_auth.html"
...
+After each rule's `match-request` section, it must also have an `allow`, `allow-unauthenticated`, or `deny` parameter. (You can set both `allow` and `deny` parameters for a rule, though Puppet Server always prioritizes `deny` over `allow` when a request matches both.) 
 
-### `allow-unauthenticated` and `auth`
+If a request matches the rule, Puppet Server checks the request's authenticated "name" (see [`allow-header-cert-info`](#allow-header-cert-info)) against these parameters to determine what to do with the request.
 
-In the HOCON rule format, Puppet Server determines whether to allow or deny requests _after_ a rule is matched. By default, this authentication method refuses unauthenticated requests. You can force Puppet Server to allow requests from unauthenticated clients by setting the Boolean `allow-unauthenticated` parameter to true:
+* **`allow-unauthenticated`**: If this Boolean parameter is set to `true`, Puppet Server allows the request --- even if it can't determine an authenticated name. **This is an insecure configuration and not recommended for production environments.** A rule with this parameter set to `true` can't also contain the `allow` or `deny` parameters.
```
camlow325: We may want to soften or eliminate the language about "... not recommended for production environments" for the `allow-unauthenticated: true` setting.  There are some situations in our own rule defaults where we have to do this.  The endpoints for submitting a certificate request and retrieving back the signed certificate, for example, use `allow-unauthenticated: true` because the submitter has to hit those without a certificate first in order to get the certificate that it will use for other requests later on.

documentation/config_file_auth.md
```diff
@@ -9,142 +9,168 @@ canonical: "/puppetserver/latest/config_file_auth.html"
...
-> **Note:** If you've enabled the new authentication method introduced in Puppet Server 2.2, Puppet Server ignores the setting of the same name in [`master.conf`](./config_file_master.html) in favor of this setting in the new `auth.conf` file. If you use the deprecated authentication method and legacy [Puppet `auth.conf`][] file, you must instead configure this setting in `master.conf`.
+~~~ hocon
+name: "my path"
+~~~
 
-This setting determines whether to enable [external SSL termination](./external_ssl_termination.markdown) on all HTTP endpoints that Puppet Server handles, including those served by the "master" service, the certificate authority API, and the Puppet Admin API. It also controls how Puppet Server derives the user's identity for authorization purposes.
+> **Note:** If multiple rules have the same `name` value, Puppet Server will fail to launch.
 
-If this setting is enabled, Puppet Server ignores any presented certificate and relies completely on header data to authorize requests. **This is very insecure; do not do this unless you've secured your network to prevent _any_ untrusted access to Puppet Server.**
+#### `allow-unauthenticated`
```
camlow325: Would it be better to eliminate this section, in favor of the `allow, allow-unauthenticated, and deny` section?  Maybe the Notes about not supporting `allow_ip` and `deny_ip` and the `auth: on` parameter could instead be pulled up into the `allow-unauthenticated` bullet under the `allow, allow-unauthenticated, and deny` section?

documentation/config_file_auth.md
```diff
@@ -159,10 +185,71 @@ Puppet Server allows agents to make requests at the old URLs and internally tran
...
 * `certificate-status.authorization-required` and `certificate-status.client-whitelist` in the `certificate-authority` section: If `authorization-required` is set to `false` or `client-whitelist` has one or more entries, Puppet Server handles requests made to its [Certificate Status](/puppet/latest/reference/http_api/http_certificate_status.html) API according to the parameters' values. See the [`ca.conf` documentation](./config_file_ca.html) for more information on these settings. If `authorization-required` is set to `true` or not set and the `client-whitelist` is set to an empty list or not set, Puppet Server authorizes requests using the authorization method introduced in Puppet Server 2.2.
+
+## Deprecated Ruby Parameters
```
camlow325: I wonder it might be better to write this section from the perspective of "migrating your existing custom rules from the deprecated Ruby auth.conf file over to Puppet Server's auth.conf".

In each of the subsections, we could have an example which shows how the rule was written in Ruby's auth.conf and the change that would have to be made to express the same configuration in Puppet Server's auth.conf.

For example, the `path` section could say that where you had to express a regex path in a Ruby auth.conf rule like this...

````
path ~ ^/puppet/v3/report/([^/]+)$
````
... that you would have to convert it to the following for a Puppet Server auth.conf rule:

````
path: "^/puppet/v3/report/([^/]+)$"
type: regex
````

Comment (camlow325) [inline documentation/config_file_auth.md]: And if it were to look like that, this section could also point out the new "required" settings that you'd have to add to your Puppet Server auth.conf rule that did not exist for a Ruby auth.conf rule - `sort-order` and `name`.

File: documentation/config_file_auth.md [hunk 7]
```diff
@@ -159,10 +185,71 @@ Puppet Server allows agents to make requests at the old URLs and internally tran
...
+
+# Literal string to match at the start of a URL's path.
+path /puppet/v3/report/
+~~~
+
+### `method`
+
+If a rule contains the `method` parameter, it only applies to requests that use the value's corresponding HTTP methods. In the deprecated `auth.conf` format, use indirector names for the `method` value:
+
+Indirector | HTTP
```
camlow325: As a further example of the "migrating your settings..." approach, this table's column names could have `Deprecated Ruby auth.conf` vs. `New Puppet Server auth.conf` - or something similar.  The point would be to show users that where they had to use "indirector" based abstractions, which could be roughly translated into corresponding HTTP methods, in the deprecated Ruby auth.conf format, users would instead need to express the HTTP method names directly in the new Puppet Server auth.conf format.

camlow325:
Done reviewing for now.  I really like the new `HOCON Parameters` section.  Great, concise coverage of all of the settings in the new `auth.conf` file.

Would be curious to see what you think about maybe reorienting the `Deprecated Ruby Parameters` section to be more of a tutorial for users on how to migrate / translate their custom auth.conf rules from the deprecated to the new format.

documentation/config_file_auth.md
nwolfe: Yeah, we should adjust this wording a bit.  If we really wanted to tie together the rules array and match-request sections in the same breath, we could even say something like `"A rules array can contains as many rules as you need, each with a single match-request section"`.

nwolfe:
:see_no_evil:

documentation/config_file_auth.md
gguillotte: Is it possible to have a rule without a `match-request` section? Like:

~~~
authorization: {
    version: 1
    rules: [
            {
                deny: "*"
                sort-order: 1
                name: "no way"
            },
    ]
}
~~~

nwolfe: No, `match-request` is required (https://github.com/puppetlabs/trapperkeeper-authorization/blob/master/doc/authorization-config.md#match-request) (https://github.com/puppetlabs/trapperkeeper-authorization/blob/master/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj#L139-L141)

documentation/config_file_auth.md
gguillotte: :+1:

documentation/config_file_auth.md
gguillotte: I'd want that in a different page. This one's long enough as a reference to both rule formats, IIRC there are other considerations in migrating between formats, and a separate migration document would be easier to remove when the older format is removed.

documentation/config_file_auth.md
gguillotte: :bow: Thanks!

gguillotte:
Revised; can squash if OK.

documentation/config_file_auth.md
camlow325: @gguillotte Ok, that's fair.  Do you think it's worthwhile to have a separate migration document - and, then, maybe a separate JIRA ticket to cover it? - or just stick with this page as it is for now and see if any users provide further feedback indicating that it would be useful?  I'm good either way.

camlow325:
:+1: and can merge.  @gguillotte - did you want any other docs folks to weigh in on this first?

gguillotte:
@camlow325 I'll get started on the auth rule migration doc today, and once I have an idea of what it'll look like we can figure out if we want to merge it into the `Deprecated Ruby Parameters` section or have it stand alone.

PR #815, opened 2015-12-14
(MAINT) Bump project.clj version to 2.2.2-master-SNAPSHOT

camlow325:
This commit bumps the Puppet Server version in the project.clj up to
2.2.2-master-SNAPSHOT given that we've released 2.2.1 already.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,7 +1,7 @@
 (def tk-version "1.1.1")
 (def tk-jetty-version "1.3.1")
 (def ks-version "1.1.0")
-(def ps-version "2.2.0-master-SNAPSHOT")
+(def ps-version "2.2.2-master-SNAPSHOT")
 
 (defn deploy-info
   [url]


haus:
Any reason this can't be 2.3.0-SNAPSHOT ?

camlow325:
I don't think we know for sure yet that the next release we do off of this branch will be a Y bump.  We've tried to be conservative about that in the past anyway - only bump to the next Z.

cprice404:
I'm +1 on this ending up as 2.2.2-master-SNAPSHOT, but it seems like if it currently says 2.2.0 then we are delinquent on some merge-ups (which is probably my fault).  In any case I'd have no big problem with merging this now as long as we're careful about not screwing it up when the merge-ups come in.

camlow325:
@cprice404 I think we had already merged up code from `stable` recently.  Just did a separate merge up PR to catch everything else that was in stable but not master yet - just docs.  That PR is #817.  Figured it made sense to have the version bump and merge up be on separate PRs.

nwolfe:
:+1:

PR #824, opened 2015-12-15
(MAINT) Upgrade ezbake to remove utopic target

nwolfe:
The version of lein-ezbake we were on was a few releases behind and
would fail to build packages when doing a
`lein with-profile ezbake ezbake build` since utopic was removed from
somewhere.

This commit just brings us up to a newer version that will no longer
attempt to build packages for utopic.


camlow325:
:+1:

PR #828, opened 2015-12-18
(SERVER-537) Cleanup jruby pools during shutdown

haus:
With https://github.com/puppetlabs/trapperkeeper/pull/198, puppet-server may begin receiving HUP signals which will cause it to shut down and restart. Without these commits, that results in leaks of scripting containers. This PR adds a stop function to the jruby puppet service that cleans up the existing jruby pools during a shutdown. Any agents connecting during a shutdown will be given a 503 if the webserver is still up.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
@@ -76,6 +76,13 @@
                                         profiler))
 
 (schema/defn ^:always-validate
+  pool-initialized? :- schema/Bool
+  "Determine if the current pool has been fully initialized."
+  [expected-pool-size :- schema/Int
+   pool :- jruby-schemas/pool-queue-type]
+  (= expected-pool-size (count (.getRegisteredElements pool))))
+
+(schema/defn ^:always-validate
   flush-pool!
   "Flush of the current JRuby pool.  NOTE: this function should never
   be called except by the pool-agent."
@@ -94,39 +101,42 @@
         new-pool (:pool new-pool-state)
         old-pool-state @pool-state
         old-pool (:pool old-pool-state)
-        count (:size old-pool-state)]
+        old-pool-size (:size old-pool-state)]
     (log/info "Replacing old JRuby pool with new instance.")
-    (when (= reason :shutdown)
-      (.insertPill new-pool (ShutdownPoisonPill. new-pool)))
-    (reset! pool-state new-pool-state)
-    (log/info "Swapped JRuby pools, beginning cleanup of old pool.")
-    (doseq [i (range count)]
-      (try
-        (let [id (inc i)
-              instance (jruby-internal/borrow-from-pool!*
-                        jruby-internal/borrow-without-timeout-fn
-                        old-pool)]
+    (if (and (= reason :shutdown) (not (pool-initialized? old-pool-size old-pool)))
+      (throw (IllegalStateException. "Attempting to flush a pool that does not appear to have successfully initialized. Aborting."))
+      (do
+        (when (= reason :shutdown)
+          (.insertPill new-pool (ShutdownPoisonPill. new-pool)))
+        (reset! pool-state new-pool-state)
+        (log/info "Swapped JRuby pools, beginning cleanup of old pool.")
+        (doseq [i (range old-pool-size)]
           (try
-            (jruby-internal/cleanup-pool-instance! instance)
-            (when (= reason :flush-requested)
-              (jruby-internal/create-pool-instance! new-pool id config
-                                                    (partial send-flush-instance! pool-context)
-                                                    profiler))
-            (finally
-              (.releaseItem old-pool instance false)))
-          (log/infof "Finished creating JRubyPuppet instance %d of %d"
-                     id count))
-        (catch Exception e
-          (.clear new-pool)
-          (.insertPill new-pool (PoisonPill. e))
-          (throw (IllegalStateException.
-                  "There was a problem adding a JRubyPuppet instance to the pool."
-                  e)))))
-    ;; Add a "RetryPoisonPill" to the pool in case something else is in the
-    ;; process of borrowing from the old pool.
-    (.insertPill old-pool (RetryPoisonPill. old-pool)))
-  (finally
-    (deliver flush-complete? true)))
+            (let [id (inc i)
+                  instance (jruby-internal/borrow-from-pool!*
+                            jruby-internal/borrow-without-timeout-fn
+                            old-pool)]
+              (try
+                (jruby-internal/cleanup-pool-instance! instance)
+                (when (= reason :flush-requested)
+                  (jruby-internal/create-pool-instance! new-pool id config
+                                                        (partial send-flush-instance! pool-context)
+                                                        profiler))
+                (finally
+                  (.releaseItem old-pool instance false)))
+              (log/infof "Finished creating JRubyPuppet instance %d of %d"
+                         id old-pool-size))
+            (catch Exception e
+              (.clear new-pool)
+              (.insertPill new-pool (PoisonPill. e))
+              (throw (IllegalStateException.
+                      "There was a problem adding a JRubyPuppet instance to the pool."
+                      e)))))
+        ;; Add a "RetryPoisonPill" to the pool in case something else is in the
+        ;; process of borrowing from the old pool.
+        (.insertPill old-pool (RetryPoisonPill. old-pool))))
+        (finally
+          (deliver flush-complete? true))))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Public
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
@@ -6,7 +6,7 @@
             [puppetlabs.services.jruby.jruby-puppet-schemas :as jruby-schemas])
   (:import (clojure.lang IFn)
            (com.puppetlabs.puppetserver PuppetProfiler)
-           (puppetlabs.services.jruby.jruby_puppet_schemas PoisonPill RetryPoisonPill JRubyPuppetInstance)))
+           (puppetlabs.services.jruby.jruby_puppet_schemas PoisonPill RetryPoisonPill JRubyPuppetInstance ShutdownPoisonPill)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Private
@@ -95,6 +95,8 @@
         old-pool (:pool old-pool-state)
         count    (:size old-pool-state)]
     (log/info "Replacing old JRuby pool with new instance.")
+    (when (= reason :shutdown)
+      (.insertPill new-pool (ShutdownPoisonPill. new-pool)))
     (reset! pool-state new-pool-state)
     (log/info "Swapped JRuby pools, beginning cleanup of old pool.")
     (doseq [i (range count)]
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
@@ -89,7 +89,7 @@
   [pool-context :- jruby-schemas/PoolContext
    old-pool-state :- jruby-schemas/PoolState
    new-pool-state :- jruby-schemas/PoolState
-   refill :- schema/Bool]
+   refill? :- schema/Bool]
   (let [{:keys [config profiler pool-state]} pool-context
         new-pool (:pool new-pool-state)
         old-pool (:pool old-pool-state)
@@ -105,7 +105,7 @@
                         old-pool)]
           (try
             (jruby-internal/cleanup-pool-instance! instance)
-            (when refill
+            (when refill?
               (jruby-internal/create-pool-instance! new-pool id config
                                                     (partial send-flush-instance! pool-context)
                                                     profiler)
@@ -125,13 +125,14 @@
 
 (schema/defn ^:always-validate
   flush-pool-for-shutdown!
-  "Flush of the current JRuby pool when shutting down during a stop."
+  "Flush of the current JRuby pool when shutting down during a stop.
+  Delivers the on-complete promise when the pool has been flushed."
   ;; Since this function is only called by the pool-agent, we know that if we
   ;; receive multiple flush requests before the first one finishes, they will
   ;; be queued up and we don't need to worry about race conditions between the
   ;; steps we perform here in the body.
   [pool-context :- jruby-schemas/PoolContext
-   flush-complete? :- IDeref]
+   on-complete :- IDeref]
   (try
     (log/info "Flush request received; creating new JRuby pool.")
     (let [{:keys [config pool-state]} pool-context
@@ -140,12 +141,12 @@
           old-pool-state @pool-state
           old-pool (:pool old-pool-state)
           old-pool-size (:size old-pool-state)]
-      (if-not (pool-initialized? old-pool-size old-pool)
+      (when-not (pool-initialized? old-pool-size old-pool)
         (throw (IllegalStateException. "Attempting to flush a pool that does not appear to have successfully initialized. Aborting.")))
       (.insertPill new-pool (ShutdownPoisonPill. new-pool))
       (swap-and-drain-pool! pool-context old-pool-state new-pool-state false))
     (finally
-      (deliver flush-complete? true))))
+      (deliver on-complete true))))
 
 (schema/defn ^:always-validate
   flush-and-repopulate-pool!
@@ -189,8 +190,8 @@
   send-flush-pool-for-shutdown! :- jruby-schemas/JRubyPoolAgent
   "Sends requests to the agent to flush the existing pool to prepare for shutdown."
   [pool-context :- jruby-schemas/PoolContext
-   flush-complete? :- IDeref]
-  (send-agent (:pool-agent pool-context) #(flush-pool-for-shutdown! pool-context flush-complete?)) )
+   on-complete :- IDeref]
+  (send-agent (:pool-agent pool-context) #(flush-pool-for-shutdown! pool-context on-complete)) )
 
 (schema/defn ^:always-validate
   send-flush-instance! :- jruby-schemas/JRubyPoolAgent
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
@@ -83,60 +83,84 @@
   (= expected-pool-size (count (.getRegisteredElements pool))))
 
 (schema/defn ^:always-validate
-  flush-pool!
-  "Flush of the current JRuby pool.  NOTE: this function should never
-  be called except by the pool-agent."
+  swap-and-drain-pool!
+  "Replace the current pool with a new pool and drain the old pool,
+  optionally refilling the new pool with fresh jrubies."
   [pool-context :- jruby-schemas/PoolContext
-   reason :- jruby-schemas/FlushReason
-   flush-complete? :- IDeref]
-  ;; Since this function is only called by the pool-agent, and since this
-  ;; is the only entry point into the pool flushing code that is exposed by the
-  ;; service API, we know that if we receive multiple flush requests before the
-  ;; first one finishes, they will be queued up and the body of this function
-  ;; will be executed atomically; we don't need to worry about race conditions
-  ;; between the steps we perform here in the body.
-  (log/info "Flush request received; creating new JRuby pool.")
+   old-pool-state :- jruby-schemas/PoolState
+   new-pool-state :- jruby-schemas/PoolState
+   refill :- schema/Bool]
   (let [{:keys [config profiler pool-state]} pool-context
-        new-pool-state (jruby-internal/create-pool-from-config config)
         new-pool (:pool new-pool-state)
-        old-pool-state @pool-state
         old-pool (:pool old-pool-state)
         old-pool-size (:size old-pool-state)]
     (log/info "Replacing old JRuby pool with new instance.")
-    (if (and (= reason :shutdown) (not (pool-initialized? old-pool-size old-pool)))
-      (throw (IllegalStateException. "Attempting to flush a pool that does not appear to have successfully initialized. Aborting."))
-      (do
-        (when (= reason :shutdown)
-          (.insertPill new-pool (ShutdownPoisonPill. new-pool)))
-        (reset! pool-state new-pool-state)
-        (log/info "Swapped JRuby pools, beginning cleanup of old pool.")
-        (doseq [i (range old-pool-size)]
+    (reset! pool-state new-pool-state)
+    (log/info "Swapped JRuby pools, beginning cleanup of old pool.")
+    (doseq [i (range old-pool-size)]
+      (try
+        (let [id (inc i)
+              instance (jruby-internal/borrow-from-pool!*
+                        jruby-internal/borrow-without-timeout-fn
+                        old-pool)]
           (try
-            (let [id (inc i)
-                  instance (jruby-internal/borrow-from-pool!*
-                            jruby-internal/borrow-without-timeout-fn
-                            old-pool)]
-              (try
-                (jruby-internal/cleanup-pool-instance! instance)
-                (when (= reason :flush-requested)
-                  (jruby-internal/create-pool-instance! new-pool id config
-                                                        (partial send-flush-instance! pool-context)
-                                                        profiler))
-                (finally
-                  (.releaseItem old-pool instance false)))
+            (jruby-internal/cleanup-pool-instance! instance)
+            (when refill
+              (jruby-internal/create-pool-instance! new-pool id config
+                                                    (partial send-flush-instance! pool-context)
+                                                    profiler)
               (log/infof "Finished creating JRubyPuppet instance %d of %d"
                          id old-pool-size))
-            (catch Exception e
-              (.clear new-pool)
-              (.insertPill new-pool (PoisonPill. e))
-              (throw (IllegalStateException.
-                      "There was a problem adding a JRubyPuppet instance to the pool."
-                      e)))))
-        ;; Add a "RetryPoisonPill" to the pool in case something else is in the
-        ;; process of borrowing from the old pool.
-        (.insertPill old-pool (RetryPoisonPill. old-pool))))
-        (finally
-          (deliver flush-complete? true))))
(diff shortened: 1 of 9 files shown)


haus:
The HUP test included depends on https://github.com/puppetlabs/trapperkeeper/pull/198.

haus:
This won't be testable by travis until the related TK PR is merged and released, but these tests are passing locally against the TK PR, so feel free to review.

src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
```diff
@@ -83,43 +79,57 @@
...
-    (.insertPill (:pool old-pool) (RetryPoisonPill. (:pool old-pool)))))
+  (try
+    (log/info "Flush request received; creating new JRuby pool.")
+    (let [{:keys [config profiler pool-state]} pool-context
+          new-pool-state (jruby-internal/create-pool-from-config config)
+          new-pool (:pool new-pool-state)
+          old-pool @pool-state
+          old-pool-size (:size old-pool)]
+      (log/info "Replacing old JRuby pool with new instance.")
+      (if (and (= reason :shutdown) (not= old-pool-size (count (.getRegisteredElements (:pool old-pool)))))
```
KevinCorcoran: This is a bit inscrutable.  It might be a good idea to write a function like `pool-initialized?` which could be used here.

haus: oooh. good call

src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
```diff
@@ -83,43 +79,57 @@
...
+        (log/warn "Attempting to flush a pool that does not appear to have successfully initialized. Aborting.")
```
KevinCorcoran: Should this be an exception instead of just a warning?

src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
```diff
@@ -83,43 +79,57 @@
...
+                (.clear new-pool)
+                (.insertPill new-pool (PoisonPill. e))
+                (throw (IllegalStateException.
+                        "There was a problem adding a JRubyPuppet instance to the pool."
+                        e)))))
+          ;; Add a "RetryPoisonPill" to the pool in case something else is in the
+          ;; process of borrowing from the old pool.
+          (.insertPill (:pool old-pool) (RetryPoisonPill. (:pool old-pool))))))
+    (finally
+      (deliver flush-complete? true))))
```
KevinCorcoran: There's now a non-trivial amount of code in this function which is only applicable to one of two cases: flushing the pool for a shutdown, or flushing and re-creating the pool while the server is running.  Instead of simply extending this function to handle both cases, I wonder if it would be cleaner to introduce a new function like `flush-pool-for-shutdown!`  (and/or perhaps name this to `flush-and-repopulate-pool!`) and factoring-out bits of this function as-needed for reuse.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -167,6 +168,15 @@
    :size size})
 
 (schema/defn ^:always-validate
+  cleanup-pool-instance! :- nil
```
KevinCorcoran: It's common practice to omit `:- nil` here.

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
```diff
@@ -42,6 +42,13 @@
             (assoc :pool-context pool-context)
             (assoc :borrow-timeout (:borrow-timeout config))
             (assoc :event-callbacks (atom []))))))
+  (stop
+   [this context]
+   (let [{:keys [pool-context]} (tk-services/service-context this)
+         flush-complete? (promise)]
+     (jruby-agents/send-flush-pool! pool-context :shutdown flush-complete?)
+     @flush-complete?)
```
KevinCorcoran: I think this `deref` should have use a timeout.

jpinsonault:
This looks good to me. I'm not super knowledgable this area of the code so I reviewed it mostly for my own education.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -283,3 +292,84 @@
...
+       (conj (tk-bootstrap/parse-bootstrap-config! bootstrap/dev-bootstrap-file) bonus-service)
+       {:jruby-puppet {:max-active-instances 1}}
+       (tk-internal/register-sighup-handler [app])
+       (beckon/raise! "HUP")
+       (let [start (System/currentTimeMillis)]
+         (while (and (not= (count @call-seq) 5)
+                     (< (- (System/currentTimeMillis) start) 90000))
+           (Thread/yield)))
+       ; 5 elements in the call-seq corresponds to init, start, stop, init, start for the BonusService
+       (is (= (count @call-seq) 5))
```
KevinCorcoran: Any reason to not just assert `(is (= @call-seq [:init-bonus-service :start- ...]))` ?

haus: Nope. That makes sense.

src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
haus: The only time @cprice404 and I saw this was in a test that replaced the pool creation function with an exception being thrown. I'm not sure what the right thing to do in that case is, as the pool creation exception will have already started a shutdown.

src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
haus: I'll take a stab at breaking this up on those lines and see how it looks.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
haus: good to know.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -283,3 +292,84 @@
...
+          stop-complete? (future (tk-app/stop app))]
+        (let [start (System/currentTimeMillis)]
+          (while (and
+                  (< (- (System/currentTimeMillis) start) 10000)
+                  (not= 503 (:status (ping-environment))))
+            (Thread/yield))
+          (is (= 503 (:status (ping-environment))))
+          (jruby-protocol/return-instance jruby-service jruby-instance :i-want-this-instance)
+          @stop-complete?
+          (is (= 503 (:status (ping-environment))))))))
```
KevinCorcoran: How does this work?  The app should be shutdown at this point in the test, right?

src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
KevinCorcoran: That makes me thing an exception is even more appropriate here, given that this is something that we expect to never happen in production.

haus: Sounds reasonable. IllegalStateException or something then?

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
fpringvaldsen: I'm also curious, it seems like if the app is shutdown this shouldn't return a 503 response

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -283,3 +292,84 @@
...
+         (while (and
+                 (< (- (System/currentTimeMillis) start) 10000)
+                 (not= 503 (:status (ping-environment))))
+           (Thread/yield)))
+       (is (= 503 (:status (ping-environment))))
+       (jruby-protocol/return-instance jruby-service jruby-instance :i-want-this-instance)
+       @stop-complete?
+       (let [app-context (tk-app/app-context app)]
+         (swap! app-context assoc :JRubyPuppetService {})
+         (tk-internal/run-lifecycle-fn! app-context tk-services/init "init" :JRubyPuppetService jruby-service)
```
fpringvaldsen: What is the purpose for these two calls?

fpringvaldsen:
I just had one question, other than that I didn't have any additional feedback on top of what @KevinCorcoran already provided. I'm :+1: once his feedback gets addressed.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
haus: Because we reached in and stopped the jruby-service earlier in the test, we need to reinitialize it here for the tk stop to actually succeed.

haus: I can add a comment to explain that, as it is pretty funky.

fpringvaldsen: Ah, okay, that makes sense to me. :+1: on adding a comment

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
haus: My understanding here, and @cprice404 may have more insight, was that the TK shutdown  unregistered services but didn't truly stop them (or maybe I have that backwards).

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
haus: comment added

haus:
@KevinCorcoran i've broken down the flush-pool! function some.

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
haus: do you have any idea what sort of timeout would make sense here?

src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
KevinCorcoran: Yeah, that sounds right.

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
```diff
@@ -42,6 +42,13 @@
             (assoc :pool-context pool-context)
             (assoc :borrow-timeout (:borrow-timeout config))
             (assoc :event-callbacks (atom []))))))
+  (stop
+   [this context]
+   (let [{:keys [pool-context]} (tk-services/service-context this)
+         flush-complete? (promise)]
+     (jruby-agents/send-flush-pool-for-shutdown! pool-context flush-complete?)
+     @flush-complete?)
```
KevinCorcoran: > do you have any idea what sort of timeout would make sense here?

Not really.  I'd be tempted to go with something like 1 minute.  I wonder if @camlow325 or @cprice404 have an opinion on this.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
KevinCorcoran: That sounds backwards ... what do you mean, "unregister" them?  TK shutdown definitely calls `stop` on all the services.

src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
```diff
@@ -65,62 +65,101 @@
...
 
+(schema/defn ^:always-validate
+  flush-pool-for-shutdown!
+  "Flush of the current JRuby pool when shutting down during a stop."
+  ;; Since this function is only called by the pool-agent, we know that if we
+  ;; receive multiple flush requests before the first one finishes, they will
+  ;; be queued up and we don't need to worry about race conditions between the
+  ;; steps we perform here in the body.
+  [pool-context :- jruby-schemas/PoolContext
+   flush-complete? :- IDeref]
```
KevinCorcoran: It's probably better to name this something like `on-complete` to indicate that it's expected to be a promise and not a boolean.  Also, could use a mention in the docstring.

src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
```diff
@@ -65,62 +65,101 @@
...
+  (try
+    (log/info "Flush request received; creating new JRuby pool.")
+    (let [{:keys [config pool-state]} pool-context
+          new-pool-state (jruby-internal/create-pool-from-config config)
+          new-pool (:pool new-pool-state)
+          old-pool @pool-state
+          old-pool-size (:size old-pool)]
+      (if-not (pool-initialized? old-pool-size (:pool old-pool))
```
KevinCorcoran: This is a nitpick, but it's a bit more idiomatic to use `when-not` in situations like this, when there is no "else" branch.

src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
```diff
@@ -65,62 +65,101 @@
...
+  (= expected-pool-size (count (.getRegisteredElements pool))))
+
+(schema/defn ^:always-validate
+  swap-and-drain-pool!
+  "Replace the current pool with a new pool and drain the old pool,
+  optionally refilling the new pool with fresh jrubies."
+  [pool-context :- jruby-schemas/PoolContext
+   old-pool :- jruby-schemas/PoolState
+   new-pool-state :- jruby-schemas/PoolState
+   refill :- schema/Bool]
```
KevinCorcoran: Should be named `refill?`

KevinCorcoran: And, it'd be nice if we could refactor this out even further and get rid of this parameter (definitely a code smell to have a boolean parameter which toggles significant functionality on/off) but that said, I think this looks a lot better and it'd probably be something of a pain to keep going, so :+1:

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -283,3 +293,88 @@
...
+                 (not= 503 (:status (ping-environment))))
+           (Thread/yield)))
+       (is (= 503 (:status (ping-environment))))
+       (jruby-protocol/return-instance jruby-service jruby-instance :i-want-this-instance)
+       @stop-complete?
+       (let [app-context (tk-app/app-context app)]
+         ;; We have to re-initialize the JRubyPuppetService here because
+         ;; otherwise the tk-app/stop that is included in the
+         ;; with-puppetserver-running macro will fail, as the
+         ;; JRubyPuppetService is already stopped.
```
KevinCorcoran: That's unfortunate.  It would probably be better if this test, instead of using `with-puppetserver-running`, started up TK manually, without using a macro - shouldn't be too hard, I think there might even be helper functions in the TK testutils that will do this.  That should eliminate the need for this cleanup.

KevinCorcoran:
This is getting close!  :+1:  I'm definitely a fan of the refactor to split out those flush functions.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
cprice404: I don't think we can do that.  There were a lot of gymnastics involved in getting this working the way we needed it to.  I will try to weigh in with additional details when I get caught up, but I'm kind of going through the PRs in order by the date they were opened.

src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
```diff
@@ -65,62 +65,101 @@
...
+   flush-complete? :- IDeref]
+  (try
+    (log/info "Flush request received; creating new JRuby pool.")
+    (let [{:keys [config pool-state]} pool-context
+          new-pool-state (jruby-internal/create-pool-from-config config)
+          new-pool (:pool new-pool-state)
+          old-pool @pool-state
+          old-pool-size (:size old-pool)]
+      (if-not (pool-initialized? old-pool-size (:pool old-pool))
+        (throw (IllegalStateException. "Attempting to flush a pool that does not appear to have successfully initialized. Aborting.")))
```
camlow325: Do we expect this to happen in production use cases?  What happens if the shutdown is requested via a HUP while the JRubyPool is only partially initialized - say 5 of the 8 expected instances having been initialized?  Will those instances get cleaned up and the service restarted okay?

camlow325: ah, maybe the partial initialization couldn't happen here.  I think `prime-pool` would be run on the same pool agent as the `flush-pool-for-shutdown!` would - so a pending `flush-pool-for-shutdown!` call would be held off until the pool is fully primed.  So probably good here then.

haus: The stop uses the pool-agent, not the flush-instance-agent, so I think this should be okay. I think the pool will finish initializing before the stop begins.
This is based on my limited understanding of agents.

camlow325: Yep, yep.  That makes sense to me, too.

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -262,3 +275,6 @@
     event-type-lock-requested? JRubyLockRequestedEvent
     event-type-lock-acquired? JRubyLockAcquiredEvent
     event-type-lock-released? JRubyLockReleasedEvent))
+
+(def FlushReason
```
camlow325: Is this used anywhere?

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
camlow325: Yeah, not sure.  Could take a super-long time depending upon what's going on within the server.  I'd probably be inclined to not try to put a timeout on this.  Otherwise, the server might be in an unknown state if we continue on with the rest of the shutdown process.

camlow325: Maybe it would be good to log starting/completing messages at a DEBUG level around the dereference here so we can see from the log if this gets stuck for an exorbitant amount of time?

camlow325:
I did some functional testing of this with puppetserver built into a jar, running on a CentOS VM.  Everything seemed to work the way I had expected it to in terms of how the process responded to HUP signals.

I noticed that when I do a 'service puppetserver stop' while the pool is still being initialized that the process won't shut down completely until after the pool has been fully initialized.  That all seems to be working the way we would expect it to be per the current implementation.  One aspect of this to consider is that we have a number of acceptance tests that involve starting up Puppet Server, doing some quick operation that only requires one pool instance to have been initialized (e.g., a puppet agent run), and shutting down the service - often before the pool is fully initialized.  In today's world, the process is shut down right away, since none of the Trapperkeeper service `stop` functions are gated on completion of pool initialization.  When this PR comes in, though, it will impose a few extra seconds in some cases to allow for the pool to be fully initialized before it is torn down.  Not sure if this is going to end up making our acceptance test runs longer as a result.  May just have to see how noticeable the difference is.

If this ends up being a problem, we might consider filing a follow-on ticket that would allow for the pool initialization to be terminated if a stop has been triggered before the pool is done fully initializing.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -14,7 +15,16 @@
             [puppetlabs.services.puppet-admin.puppet-admin-service :as puppet-admin]
             [puppetlabs.trapperkeeper.services.authorization.authorization-service :as authorization]
             [puppetlabs.http.client.sync :as http-client]
-            [me.raynes.fs :as fs]))
+            [me.raynes.fs :as fs]
+            [puppetlabs.trapperkeeper.internal :as tk-internal]
+            [puppetlabs.trapperkeeper.testutils.logging :as logutils]
```
camlow325: Unused?

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
camlow325: I'm okay with it how it is now.  If others think this test is too onerous, I'd be okay just having the `test-503-when-app-shuts-down` one added for this PR and eliminating this one.  I understand the desire to have a test that exercises the actual production stack, including HTTP network calls to the real Jetty web service.

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
```diff
@@ -38,27 +40,26 @@
 
 (deftest test-error-during-init
   (testing
-      (str "If there as an exception while putting a JRubyPuppet instance in "
-           "the pool the application should shut down.")
+   (str "If there as an exception while putting a JRubyPuppet instance in "
```
camlow325: s/as/is

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
```diff
@@ -38,27 +40,26 @@
...
-                     (is (true? @got-expected-exception)
-                                "Did not get expected exception."))
-                   (is (logged?
-                         #"^shutdown-on-error triggered because of exception!"
-                                :error))))))
+     (with-redefs [jruby-internal/create-pool-instance!
+                   (fn [& _] (throw (Exception. "42")))]
+       (let [got-expected-exception (atom false)]
+         (try
+           ; We use this instead of with-app-with-config because with-app-with-config has an implicit stop
```
camlow325: Did this behavior change with the changes in the HUP-related PRs or is this just something you noticed?

haus: Once the HUP changes were in we started seeing a double-stop, which broke the logging check. However, now that the partial-pool initialization throws an error instead of just logging a warning, we still have a double stop, so this could probably be reverted to the old test with the logged? check removed.

camlow325:
Overall :+1: from me.  Looks great, works great!  I just left a few cosmetic comments, mostly about code that didn't appear to be used anywhere.

src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
haus: cool. will fix.

src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
haus: will do.

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
haus: good idea.

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
haus: it was until the flush-pool got refactored. it can go away now.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
haus: Yup.

haus:
@KevinCorcoran @camlow325 I believe i've now responded to all of your feedback.

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
camlow325: > Once the HUP changes were in we started seeing a double-stop, which broke the logging check.

Ah, okay.  Took me a while to wrap my head around what this meant.  I think what happens with this test now is that the ShutdownService's `shutdown-on-error` is triggered twice - once for the failure in `prime-pool!` when the initial JRuby instance cannot be created and once again from `swap-and-drain-pool!` via `flush-pool-for-shutdown!` when the service is being stopped.  So it's probably not that the service's `stop` is being called twice that breaks the logging check but that the `shutdown-on-error triggered because of exception` being logged from both of those failed operations now - whereas the service `stop` prior to changes in this PR didn't try to drain the pool.  Since the `logged?` macro only allows you to match one instance - and not two - makes sense that you couldn't use it here anymore.  I'm good with removing the check for the log message.  I generally don't like to have those kinds of assertions in unit tests anyway.

I think this could be reverted back to using the `with-app-with-config` macro now, as you mentioned.  Seems like the remaining assertion for "@got-expected-exception" would work fine here.  Could also just have the try/catch wrapped around the `run-app` call since that's the one we expect to propagate the exception up, right?

KevinCorcoran:
needs a rebase

haus:
rebase complete. running tests then i'll push out a thing.

haus:
okay, tests are passing, but puppetlabs.services.jruby.jruby-puppet-service-test/test-error-during-init is intermittently showing a stacktrace during execution. i haven't been able to suppress it so far.

camlow325:
One other thing that I wasn't totally clear how we've covered is determining whether any memory has leaked between service HUP restarts.  I think we've done a good job focusing specifically on whether or not any `ScriptingContainer` objects have leaked across the restarts.  In my anecdotal functional testing of this on a VM, I didn't see any evidence that `ScriptingContainer` objects were leaking across restarts - but I didn't go any further than that.  I'd be curious to see if there are any other ones that may be leaking that we don't know about - seems like there was some suspicion about Jetty / JMX-related resources leaking but not sure if there are others lurking.  Do we feel like we're well covered for this already?

haus:
@camlow325 I don't have a super good answer for that. I know that @erikPrime had some ideas about beaker tests around this. The best idea I've heard so far would be to spin up this branch, and alternate between jmap and hup for a few times and then aggregate the results and see if anything isn't being cleaned up correctly. That's how I spotted the jetty related objects previously.

camlow325:
@haus Yeah, I wasn't expecting that we'd have an answer yet for having a long-term automated way of doing this.  Even just doing the manual jmap/hup a few times and compare snapshots for potentially leaking objects would be enough for me for now.  If you've already done that and didn't spot anything that isn't accounted for - like the Jetty ones that you found - then I'm good with this for now.

haus:
@camlow325 I'll do another check today and throw anything suspicious in a gist for review

cprice404:
@camlow325 I have a ticket about the jetty thing and was planning on doing some more exhaustive testing sometime around whenever I work on that.  I think it's probably OK for this PR to just be concerned with the scriptingcontainers for now.

haus:
The HUP test has been disabled until the TK support for HUP lands. That should enable this PR to get merged independent of the TK PR.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
cprice404: +1 to what @camlow325 said.

cprice404:
+1 Imma merge this today unless anyone has any last thoughts or concerns... @camlow325 @KevinCorcoran ?

camlow325:
No more concerns from me.  I'm good with it.

PR #829, opened 2015-12-18
(FOR REVIEW ONLY)(DO NOT MERGE)(SERVER-1064) Port to JRuby 9.0.4.0

camlow325:
This commit bumps Puppet Server's JRuby dependency up to 9.0.4.0.  This
commit also updates a couple of dependencies related to the JRuby bump,
including:

* Explicitly specify a dependency on version 1.14 of org.yaml/snakeyaml.
  The clj-yaml 0.4.0 dependency was pulling in version 1.13 of
  snakeyaml, but JRuby 9K bundles version 1.14 in its stdlib jar.  This
  commit also follows the pattern we used with bouncycastle where we
  exclude the snakeyaml jar from the uberjar.

* Bump the joda-time dependency up to 2.8.2 since this is what JRuby
  9.0.4.0 uses.

* Fixed a few test failures that occurred with the JRuby bump.  JRuby 9K
  apparently does not make stdout available from one of its Main
  containers for tests to capture.  Refactored some of the CLI tests
  that read from stdout to inject a stdout stream directly into the
  construction of the Main container.  Also changed a couple of places
  where JRuby 9K fails to equate a symbol to a string.  This actually
  seems like a good change in JRuby 9K, compatible with MRI.  Not
  exactly sure why these were succeeding on old code.


camlow325:
We haven't made a call on when we might want to switch Puppet Server over to JRuby 9000K.  Just putting this up for review for now - and for anyone who would want to play around with it.  Will probably close after a few people have looked at it.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -27,18 +28,13 @@
   classpath.
 
   See also:  http://jruby.org/apidocs/org/jruby/runtime/load/LoadService.html"
-  "puppet-server-lib")
+  "classpath:/puppet-server-lib")
```
camlow325: With this just including the relative path, JRuby will fail to find any of the Puppet Server-specific Ruby files inside of the jar at run-time.  I was able to get this to work, however, by using a prefix of either `classpath:` or `uri:classloader:` to work.  `classpath:` seemed more appropriate since we just want JRuby to find the Puppet Server Ruby files from within the classpath.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -100,7 +96,6 @@
   ; Note, this behavior should remain consistent with new-main
   (doto scripting-container
     (.setLoadPaths (managed-load-path ruby-load-path))
-    (.setCompatVersion compat-version)
```
camlow325: Ruby version is hardcoded to "2.2" in JRuby 9K.  The `ScriptingContainer` still supports the `.setCompatVersion` method but it's deprecated and does nothing - https://github.com/jruby/jruby/blob/9.0.4.0/core/src/main/java/org/jruby/embed/ScriptingContainer.java#L503-L505.  Seemed best to just not call it anymore.

src/ruby/puppet-server-lib/puppet/server/master.rb
```diff
@@ -97,7 +97,7 @@ def getClassInfoForEnvironment(env)
     # and so would not be subject to any potentially stale settings data being
     # used to enumerate manifests.
     environment = @env_loader.list.find do |env_from_loader|
-      env_from_loader.name == env
+      env_from_loader.name.to_s == env
```
camlow325: `env_from_loader.name` is a `Symbol` but `env` is a `String`.  For reasons I don't understand, the original comparison passes under JRuby 1.7.X but fails on JRuby 9K - but seemingly unique to this code for some reason.

The JRuby 9K behavior seems correct to me - and is consistent with what I see when doing simple `Symbol` to `String` comparisons from `irb` under MRI Ruby 1.9.3 and 2.1 and `jirb` under JRuby 1.7.22 (under Ruby 1.9.3) and 9.0.4.0.  I also did this against code pulled from the master branch:

````
# puppetserver ruby --version
jruby 1.7.20.1 (1.9.3p551) 2015-06-10 d7c8c27 on OpenJDK 64-Bit Server VM 1.7.0_51-mockbuild_2014_06_09_19_19-b00 +indy +jit [linux-amd64]
# puppetserver ruby -e "puts :mysym == 'mysym'"
false
````

File: test/unit/puppetlabs/services/jruby/jruby_interpreter_test.clj [hunk 4]
```diff
@@ -57,5 +57,7 @@
           jruby-env (.runScriptlet jruby-interpreter "ENV")]
 
       ; $HOME and $PATH are left in by `jruby-puppet-env`
-      (is (= #{"HOME" "PATH" "GEM_HOME" "JARS_NO_REQUIRE" "JARS_REQUIRE"}
+      ; $RUBY is added by JRuby:
+      ; https://github.com/jruby/jruby/blob/9.0.4.0/core/src/main/java/org/jruby/RubyInstanceConfig.java#L680
```
camlow325: I don't know what the deal with this is.  May just have to live with it?

spec/puppet-server-lib/puppet/jvm/puppet_config_spec.rb
```diff
@@ -17,7 +17,7 @@
     describe "the puppet log level (Puppet::Util::Log.level)" do
       subject { Puppet::Util::Log.level }
       it 'is set to debug (the highest) so messages make it to logback' do
-        expect(subject).to eq('debug')
+        expect(subject).to eq(:debug)
```
camlow325: Appears to be due to the same `Symbol` -> `String` comparison issue mentioned earlier.

camlow325:
Failed Travis on all three JDKs because it timed out waiting for the `puppet-admin` tests to finish on two of them.  On OpenJDK 1.7, https://travis-ci.org/puppetlabs/puppet-server/jobs/97753914, it encountered this error:

````
FAIL in (admin-api-flush-jruby-pool-test) (jruby_pool_int_test.clj:140)

actual: (not (true? false))
````
This issue has been encountered on previous Travis runs, captured as - https://tickets.puppetlabs.com/browse/SERVER-1012.

Will try to re-run Travis...

**Comment by camlow325** [comment]
This PR has failed Travis runs on all JDKs multiple times now.  Usually, the failures happen due to an output timeout while running the puppet-admin-int-tests:

~~~~
lein test puppetlabs.services.puppet-admin.puppet-admin-int-test

2015-12-21 01:50:21,188 WARN  [main] [o.e.j.s.h.ContextHandler] Empty contextPath
...
2015-12-21 02:09:00,928 WARN  [main] [o.e.j.s.h.ContextHandler] Empty contextPath

No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.

The build has been terminated
~~~~

The tests pass when I run them locally.  I haven't timed them formally but it does seem like the tests are considerably slower on JRuby 9000 vs. what is on the master branch with JRuby 1.7.20.1 right now.

The puppet-admin-int-tests involve exercising Puppet Ruby code pretty heavily - mostly around flushing the JRuby pool.  Not sure at this point if this indicates some significant performance degradation in general with JRuby 9000 vs. JRuby 1.7.20.1, at least in terms of how Puppet Server uses it.  Or maybe there's heavier memory consumption involved that would lead to more Garbage Collection / thrashing as the heap nears exhaustion later in the run.

We would obviously need to get to the bottom of this before considering moving to JRuby 9000.

**Comment by cprice404** [comment]
@camlow325 would be interesting to see if this changes with a larger heap size or with @haus 's scriptingcontainer cleanup fixes.

**Comment by camlow325** [comment]
@cprice404 Yeah, I'm not sure that just bumping up memory or cleaning up the ScriptingContainers is going to be enough to fix the performance discrepancy.  On my local laptop, full runs of the Puppet Server Clojure unit tests take about 10 minutes or so against the master branch currently.  For runs done against this branch, I tried bumping the maximum heap size that the tests currently use from 2g to 4g.  There did appear to be some improvement from doing that but not too much.  One 2g run took about 32 minutes and a 4g run took about 29 minutes.  I suppose maybe there's still a memory issue that explains the discrepancy, not sure - like if the amount of memory is left behind by an un-terminated ScriptingContainer has gone way up with JRuby 9000.  I may try to run this with the ScriptingContainer cleanup fixes from #828 when I get some time and see how that performs.

File: project.clj [hunk 6]
```diff
@@ -27,21 +28,14 @@
...
                  [puppetlabs/dujour-version-check "0.1.2" :exclusions [org.clojure/tools.logging]]
                  [puppetlabs/http-client "0.4.4"]
-                 [org.jruby/jruby-core "1.7.20.1"
-                  :exclusions [com.github.jnr/jffi com.github.jnr/jnr-x86asm]]
-                 ;; jffi and jnr-x86asm are explicit dependencies because,
-                 ;; in JRuby's poms, they are defined using version ranges,
-                 ;; and :pedantic? :abort won't tolerate this.
-                 [com.github.jnr/jffi "1.2.9"]
-                 [com.github.jnr/jffi "1.2.9" :classifier "native"]
-                 [com.github.jnr/jnr-x86asm "1.0.2"]
```
cprice404: does this mean that in JRuby9k they've fixed the version range stuff for JFFI?

cprice404:
This seems relatively straightforward to me.  The handful of behavioral changes seem mildly surprising, but for the most part the way that you've refactored things to handle them seem like improvements to the code anyway.

I am curious about the JFFI version ranges thing if you have any info about it.

I'm going to go ahead and close the PR, and we can pick it back up when we have time to get more serious about the upgrade (which hopefully will be very soon).

project.clj
camlow325: Well, I left `:pedantic? :abort` in our project.clj and it didn't barf when I ran `lein` anything against it.  And it did appear that their poms pin specific versions for jffi...

https://github.com/jruby/jruby/blob/9.0.4.0/pom.xml#L116
https://github.com/jruby/jruby/blob/9.0.4.0/core/pom.xml#L166

... and jnr-x86asm ...

https://github.com/jruby/jruby/blob/9.0.4.0/core/pom.xml#L117

Figured it would be better to just pull in `jruby-core` and not mention those other dependencies since we don't have anything that calls into them directly AFAIK.

cprice404: Yeah if we don't have to do that crap anymore that's great.

PR #832, opened 2015-12-22
Merge stable into master

KevinCorcoran:
There were no merge conflicts to resolve in this merge-up.

This PR merges-up the fix for the gem test which has been failing in CI.  It also appears to pull in docs updates from `stable` but I assume that's okay.


cprice404:
@camlow325 @gguillotte I'm going to merge this so that we can get tests passing, but, if either or both of you want to review this to make sure that the doc merge-up is valid, that'd be appreciated.

gguillotte:
@cprice404 :+1:

PR #834, opened 2015-12-30
(QENG-3181) Use beaker-hostgenerator instead of sqa-utils.

waynr:


kurtwall:
:see_no_evil: 
@waynr Same question for all of these PRs: Did they run and pass?

PR #827, opened 2015-12-16
(FOR REVIEW) (MAINT) update to clojure 1.7

cprice404:
This is blocked on a `comidi` release.

PR #844, opened 2016-01-03
(SERVER-1052) allow configurable jruby compile mode

cprice404:
This PR is mostly cleanup, refactoring, whitespace changes.  The interesting commit is c56509d , which adds support for configuring JRuby's `CompileMode`.

### DIFF
--- a/documentation/config_file_puppetserver.md
+++ b/documentation/config_file_puppetserver.md
@@ -30,6 +30,7 @@ The `puppetserver.conf` file contains settings for Puppet Server software. For a
     JRuby flushing can be useful for working around buggy module code that would otherwise cause memory leaks, but it slightly reduces performance whenever a new JRuby instance reloads all of the Puppet Ruby code. If memory leaks from module code are not an issue in your deployment, the default value of 0 performs best.
     * `borrow-timeout`: Optional. The timeout in milliseconds, when attempting to borrow an instance from the JRuby pool. The default is 1200000.
     * `use-legacy-auth-conf`: Optional. The method to be used for authorizing access to the HTTP endpoints served by the "master" service. The applicable endpoints are listed in [Puppet v3 HTTP API](/puppet/latest/reference/http_api/http_api_index.html#puppet-v3-http-api).
+    * `compile-mode`: Optional, experimental.  Used to control JRuby's "CompileMode", which may improve performance.  The default value is `:off`, which is the most conservative value.  A value of `:jit` will enable JRuby's "just in time" compilation of Ruby code.  A value of `:force` will cause JRuby to attempt to pre-compile all Ruby code ahead of time.
 
     If this setting is set to `true` or is not specified, Puppet uses the [deprecated][] Ruby `puppet-agent` authorization method and [Puppet `auth.conf`][`auth.conf` documentation] format, which will be removed in a future version of Puppet Server.
     
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
@@ -95,10 +95,13 @@
   [ruby-load-path :- [schema/Str]]
   (cons ruby-code-dir ruby-load-path))
 
-(defn prep-scripting-container
-  [scripting-container ruby-load-path gem-home]
-  ; Note, this behavior should remain consistent with new-main
-  (doto scripting-container
+(schema/defn ^:always-validate init-jruby-config :- jruby-schemas/ConfigurableJRuby
+  "Applies configuration to a JRuby... thing.  See comments in `ConfigurableJRuby`
+  schema for more details."
+  [jruby-config :- jruby-schemas/ConfigurableJRuby
+   ruby-load-path :- [schema/Str]
+   gem-home :- schema/Str]
+  (doto jruby-config
     (.setLoadPaths (managed-load-path ruby-load-path))
     (.setCompatVersion compat-version)
     (.setCompileMode compile-mode)
@@ -112,7 +115,7 @@
          (string? gem-home)]
    :post [(instance? ScriptingContainer %)]}
   (-> (ScriptingContainer. LocalContextScope/SINGLETHREAD)
-    (prep-scripting-container ruby-load-path gem-home)))
+      (init-jruby-config ruby-load-path gem-home)))
 
 (defn create-scripting-container
   "Creates an instance of `org.jruby.embed.ScriptingContainer` and loads up the
@@ -132,7 +135,7 @@
     ;; we need to explicitly require 'jar-dependencies' so that it is used
     ;; to manage jar loading.  We do this so that we can instruct
     ;; 'jar-dependencies' to not actually load any jars.  See the environment
-    ;; variable configuration in 'prep-scripting-container' for more
+    ;; variable configuration in 'init-jruby-config' for more
     ;; information.
     (.runScriptlet "require 'jar-dependencies'")
     (.runScriptlet "require 'puppet/server/master'")))
@@ -313,12 +316,9 @@
   e.g. for the ruby, gem, and irb subcommands.  Internal core services should
   use `create-scripting-container` instead of `new-main`."
   [config :- jruby-schemas/JRubyPuppetConfig]
-  (let [jruby-config (RubyInstanceConfig.)
-        {:keys [ruby-load-path gem-home]} config]
-    ; Note, this behavior should remain consistent with prep-scripting-container
-    (doto jruby-config
-      (.setLoadPaths (managed-load-path ruby-load-path))
-      (.setCompatVersion compat-version)
-      (.setCompileMode compile-mode)
-      (.setEnvironment (managed-environment (get-system-env) gem-home)))
+  (let [{:keys [ruby-load-path gem-home]} config
+        jruby-config (init-jruby-config
+                      (RubyInstanceConfig.)
+                      ruby-load-path
+                      gem-home)]
     (Main. jruby-config)))
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
@@ -29,11 +29,6 @@
   See also:  http://jruby.org/apidocs/org/jruby/runtime/load/LoadService.html"
   "puppet-server-lib")
 
-(def compile-mode
-  "The JRuby compile mode to use for all ruby components, e.g. the master
-  service and CLI tools."
-  RubyInstanceConfig$CompileMode/OFF)
-
 (def compat-version
   "The JRuby compatibility version to use for all ruby components, e.g. the
   master service and CLI tools."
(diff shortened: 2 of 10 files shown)


documentation/config_file_puppetserver.md
```diff
@@ -30,6 +30,7 @@ The `puppetserver.conf` file contains settings for Puppet Server software. For a
     JRuby flushing can be useful for working around buggy module code that would otherwise cause memory leaks, but it slightly reduces performance whenever a new JRuby instance reloads all of the Puppet Ruby code. If memory leaks from module code are not an issue in your deployment, the default value of 0 performs best.
     * `borrow-timeout`: Optional. The timeout in milliseconds, when attempting to borrow an instance from the JRuby pool. The default is 1200000.
     * `use-legacy-auth-conf`: Optional. The method to be used for authorizing access to the HTTP endpoints served by the "master" service. The applicable endpoints are listed in [Puppet v3 HTTP API](/puppet/latest/reference/http_api/http_api_index.html#puppet-v3-http-api).
+    * `compile-mode`: Optional, experimental.  Used to control JRuby's "CompileMode", which may improve performance.  The default value is `:off`, which is the most conservative value.  A value of `:jit` will enable JRuby's "just in time" compilation of Ruby code.  A value of `:force` will cause JRuby to attempt to pre-compile all Ruby code ahead of time.
```
rlinehan: I don't think having these as keyword works - the config service doesn't parse values as keywords, only as strings, so actually trying to set this setting in my config file currently results in an error.

cprice404: Crap, good call, thanks.  Will fix.  I wonder if there's a not-insane way to add test coverage for that... :(

jpinsonault:
This looks good to me barring Ruth's comment

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -95,44 +91,55 @@
...
+  [config-compile-mode :- jruby-schemas/SupportedJRubyCompileModes]
+  (case config-compile-mode
+    :jit RubyInstanceConfig$CompileMode/JIT
+    :force RubyInstanceConfig$CompileMode/FORCE
+    :off RubyInstanceConfig$CompileMode/OFF))
+
+(schema/defn ^:always-validate init-jruby-config :- jruby-schemas/ConfigurableJRuby
+  "Applies configuration to a JRuby... thing.  See comments in `ConfigurableJRuby`
+  schema for more details."
+  [jruby-config :- jruby-schemas/ConfigurableJRuby
```
camlow325: FWIW, Jeff and I had talked about doing this as a common helper that called the same on either type regardless of the grossness of not having a common Java interface available to enforce the contract.  I think we chose not to do it then because you were concerned about not having that common Java interface.  I'm still okay with doing it this way for the sake of keeping the initialization common between the `ScriptingContainer` and JRuby `Main` for the CLI use cases.

test/integration/puppetlabs/services/jruby/jruby_puppet_internal_test.clj
```diff
@@ -1,35 +1,60 @@
...
-              (is (= ["TLS_RSA_WITH_AES_256_CBC_SHA256"
-                      "TLS_RSA_WITH_AES_256_CBC_SHA"]
-                    (into [] (settings "cipher_suites")))))
-            (testing "ssl_protocols"
-              (is (= ["TLSv1" "TLSv1.2"]
-                    (into [] (settings "ssl_protocols"))))))
-          (finally
-            (.terminate (:jruby-puppet instance))
-            (.terminate container))))))
+(deftest get-compile-mode-test
```
camlow325: Not a big deal but the `get-compile-mode-test` feels more like a "unit" test to me than an "integration" one.  I think I may have done the `settings-plumbed-into-jruby-container` one originally.  That one felt more like an "integration" one because it was validating actions that occurred down inside of the Ruby code run within the container whereas the `get-compile-mode-test` one just deals with the immediate function in the internal namespace and has no side effects.  Think it's worth separating the new one out into a unit namespace instead?

src/java/com/puppetlabs/puppetserver/jruby/ScriptingContainer.java
```diff
@@ -0,0 +1,35 @@
...
+        extends org.jruby.embed.ScriptingContainer {
+    public ScriptingContainer(LocalContextScope scope) {
+        super(scope);
+    }
+    public ScriptingContainer(LocalContextScope scope, LocalVariableBehavior behavior) {
+        super(scope, behavior);
+    }
+
+    /**
+     * This method delegates to a specific signature of #callMethod from the
```
camlow325: This makes sense.  Just curious whether you ran into a specific problem that led you to do this - i.e., you found out that our Clojure code was actually calling the wrong method - and what problem that may have caused.

camlow325:
To get around the problem that @rlinehan found, I tested this branch out by just changing one line in `initialize-config` from "jruby_puppet_core.clj" to...

~~~~clj
(update-in [:compile-mode] #(or (keyword %) default-jruby-compile-mode))
~~~~

... and using "compile-mode: [string value]" (for example, "compile-mode: jit") in my config file.

This all seemed to work okay.

Not sure if this is how you would want to make this work.  If you did this, though, could just add another test like this one - https://github.com/puppetlabs/puppet-server/blob/puppet-server-2.2.1/test/unit/puppetlabs/services/jruby/jruby_pool_test.clj#L34-L38 - to ensure that the string value specified in the original config is translated into a keyword during the initial config validation.

camlow325:
Mostly :+1: pending the fix for the problem @rlinehan found and merging in #827 ahead of this one.  Just left a few minor comments.

fpringvaldsen:
Also :+1: pending the keyword fix

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
cprice404: Yep, I remember that, and I know I'm going back on my original stance.  I think that the tipping point for me was that we would be adding an additional coupling with a config setting, that we didn't have before.

I still feel like the new approach is disgusting, and could be talked out of it.  But with the new coupling with the config setting it started to feel like the lesser of two evils.

camlow325: Another thought that occurred to me is that we could just implement some new adapter Java classes that inherit from a new common interface that we'd create with the setters that we know that we'll need to use but that derive from the appropriate JRuby concrete implementations.  That way, we could just refer to the interface in the schema validation here.  Where this PR already introduces a derived class implementation for `ScriptingContainer`, maybe it wouldn't be too onerous to do a new one for `RubyInstanceConfig` as well.  WDYT?

test/integration/puppetlabs/services/jruby/jruby_puppet_internal_test.clj
cprice404: I totally agree with this and had the same confusion / concern when writing this.

Where I ended up landing on it was:

1. I kind of hate having multiple files with the same name in a project, which has been a factor in how we've chosen to name a lot of stuff so far, and I was hesitant to create another `jruby_puppet_internal_test` namespace.
2. I don't feel like we have solid alignment / agreement on the best way to organize unit vs. integration tests as a team, and have been thinking for a while that that would be a good topic for a festivus sometime soon.  So I made a somewhat arbitrary decision in the meantime.

I did have the same thought though, and am willing to change it up if anyone has a brilliant suggestion that is actionable prior to us having a team conversation about best practices.

src/java/com/puppetlabs/puppetserver/jruby/ScriptingContainer.java
cprice404: I did.  I've run into it several times in the past.  What happens is that Clojure calls the wrong arity of the Java method, which results in the underlying Ruby method being passed a single array as an argument instead of being passed multiple arguments (or vice versa).  That surfaces in Clojure as a JRuby exception about passing the wrong number of arguments to the Ruby method.

I've run into this enough times in the past where I decided it was probably worth just eliminating the issue once and for all.  I tried a few approaches around type-hinting things in Clojure but they were horrible and didn't seem like they were 100% reliable anyway.  This one seems 100% reliable to me... even though I kind of hate it.

cprice404:
Rebased and added a commit that fixes the string/keyword issue.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
cprice404: That seems reasonable but I'm not sure if I consider it urgent enough to gate this PR on.  As much as I hate this code I do think that it would explode pretty obviously if the upstream API changed, don't you?  I'm definitely willing to file a follow-up ticket that suggests what you said, and could be talked into gating this PR on that work, but would probably lean towards not blocking this one.

test/integration/puppetlabs/services/jruby/jruby_puppet_internal_test.clj
camlow325: No particularly brilliant ideas.  I was thinking that it's best to reserve the reuse of namespaces from production code only in the corresponding "unit" tests that would reference them - so, in this case, there would be a "unit" namespace of "puppetlabs.services.jruby.jruby-puppet-internal-test".  But for integration tests, we could consider changing the namespace to something more unique to what's being tested, or just decorate it with "integration" in this case to make it unique if we can't come up with a better name, e.g., "puppetlabs.services.jruby.jruby-puppet-internal-integration-test".

Dunno.  I'm okay keeping it like you have it if no one comes up with an idea that you like better.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
camlow325: > As much as I hate this code I do think that it would explode pretty obviously if the upstream API changed, don't you?

I do, yeah, and that was the same argument for a common helper that I'd made when you, Jeff, and I had talked about this previously.  :smile:

I'm good just keeping it as you have it for now and not creating another ticket that we'll probably never get around to anyway...

src/java/com/puppetlabs/puppetserver/jruby/ScriptingContainer.java
camlow325: Yeah, that all makes sense.  Was just curious if something about the changes on this PR caused this problem where we weren't seeing it before since the code changes at the Clojure level didn't seem obviously different.  I'm good with the changes you made here.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
cprice404: > I do, yeah, and that was the same argument for a common helper that I'd made when you, Jeff, and I had talked about this previously. :smile:

:) fair

camlow325:
Latest code changes look good to me so I'm :+1:  @cprice404 if you're good with us merging this, I'll do the deed.

src/java/com/puppetlabs/puppetserver/jruby/ScriptingContainer.java
cprice404: Believe it or not, adding a schema to the return value of the `create-scripting-container` causes the error.  When that function is not schematized, or when the schema omits a return value specification, Clojure calls the correct Java signature.  With a schematized return value it calls the wrong signature.

camlow325: (woah)

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
cprice404: Created SERVER-1083

cprice404:
@camlow325 I created that ticket, so I'm done unless someone tells me otherwise

camlow325:
Sounds good.  :rocket:

documentation/config_file_puppetserver.md
camlow325: Would the user need to specify these as strings like...

~~~~
compile-mode: jit
~~~~

... or would doing this as a keyword work?

~~~~
compile-mode: :jit
~~~~

I was thinking they would have to be in string form for the config file.

cprice404: You're right, I missed the docs update w/my last commit

PR #848, opened 2016-01-06
(MAINT) Fix ezbake clojure version conflict

cprice404:
This commit adds an explicit clojure version to our ezbake configuration.
Without this, leiningen may bring in its own version of clojure, which
can vary depending on which version of lein you are running.

Also, this commit updates us to the latest trapperkeeper for improved
logging testutils.


camlow325:
Huh, bummer.  :see_no_evil:

nwolfe:
:+1:

PR #849, opened 2016-01-06
(SERVER-885) Add execute-command function

haus:
This commit adds an execute-command function to puppetserver that can be
used to execute an external command and return the exit code, stdout and
stderr.


haus:
So this uses the existing java ProcessWrapper class, but my work seems to be basically reimplementing clojure.java.shell/sh so it may be cleaner to just use that directly.

fpringvaldsen:
I think my preference would be to just go ahead and use `clojure.java.shell/sh`, since it's already in use in the CA codebase. 

If others feel that we shouldn't use `clojure.java.shell/sh`, then I'm :+1: on this code.

haus:
Tests are going to get very weird because of pathing of binaries. Yay.

haus:
@fpringvaldsen I agree with you. It might be worth wrapping sh in the absolutepath and executable checks to provide a more meaningful error-message in that case.

cprice404:
Hrm.  We probably need to at least think about whether we care about the ruby code using a different execution path than this new clojure code, and whatever we decide, we should probably reconcile with the CA stuff.  Sorry @haus I missed that, didn't realize we were already using the clojure.sh stuff.

Also we probably need to vet the clojure.sh stuff in terms of how it buffers the output streams...

KevinCorcoran:
I'd prefer to use `clojure.java.sh` if it meets our needs.

cprice404:
Then someone needs to do an audit on how it buffers output streams...

haus:
clojure.java.shell does not buffer streams. It deadlocks when trying to cat a large file.

haus:
If we decide to go down the apache commons exec route, I'll open a library request for it.

haus:
And...I've now added a third implementation here. This has the java side refactored to use the commons-exec library, and the clojure side just calls into ExecutionStubImpl/executeCommand the same as the jruby puppet-server code does.

haus:
The ExecutionStubImpl refactoring will mean some straightforward updates to the jruby side to ensure that commands and command arguments are being passed correctly.

src/clj/puppetlabs/puppetserver/shell_utils.clj
```diff
@@ -0,0 +1,90 @@
...
+  validate-command
+  "Checks the command string to ensure that is is an absolute path, executable
+  and that the file exists. An exception is thrown is any of those are not the
+  case."
+  [command :- schema/Str]
+  (let [command-file (io/as-file command)]
+    (cond
+      (not (.isAbsolute command-file)) (throw (IllegalArgumentException. (format "An absolute path is required, but %s is not an absolute path" command)))
+      (not (.exists command-file)) (throw (IllegalArgumentException. (format "The referenced command '%s' does not exist" command)))
+      (not (.canExecute command-file)) (throw (IllegalArgumentException. (format "The referenced command '%s' is not executable" command))))))
```
KevinCorcoran: These lines are really long, it'd be nice to shorten them up to 80 characters or so.

src/java/com/puppetlabs/puppetserver/ExecutionStubImpl.java
```diff
@@ -10,25 +14,78 @@
...
      * @throws InterruptedException
      * @throws IOException
      */
     public static ExecutionStubResult executeCommand(String command)
             throws InterruptedException, IOException {
+        org.apache.commons.exec.CommandLine commandLine = org.apache.commons.exec.CommandLine.parse(command);
+        ByteArrayOutputStream errStream = new ByteArrayOutputStream();
+        ByteArrayOutputStream outStream = new ByteArrayOutputStream();
+
+        Executor executor = setupExecutor(errStream, outStream);
```
KevinCorcoran: I think you can push the rest of the code in this method below this line into `setupExecutor`, since it's the same in both versions of `executeCommand`.  In which case, `setupExecutor` probably warrants a rename, but it'd be a little DRY-er that way.

src/java/com/puppetlabs/puppetserver/ExecutionStubImpl.java
```diff
@@ -10,25 +14,78 @@
...
+     */
+    private static Executor setupExecutor(ByteArrayOutputStream errStream, ByteArrayOutputStream outStream) {
+        DefaultExecutor executor = new DefaultExecutor();
+        PumpStreamHandler streamHandler = new PumpStreamHandler(outStream, errStream);
+
+        // Don't throw exception on non-zero exit code
+        executor.setExitValues(null);
+
+        // Set up the handlers
+        executor.setStreamHandler(streamHandler);
```
KevinCorcoran: I think this bit with the `PumpStreamHandler` is the crucial bit of this implementation.  I took a quick glance at the code/docs and it seems to do what we want (i.e. what we were previously doing with `StreamBoozer`.  @cprice404 whatcha think?

KevinCorcoran:
I'm +1 on switching to the implementation based on `commons-exec`.  It seems to satisfy our current requirements and have a good path forward to consolidating the CA code / handling environment variables and such.  But, @cprice404, happy to let you make the call on this one.

src/java/com/puppetlabs/puppetserver/ExecutionStubImpl.java
haus: Updated

src/clj/puppetlabs/puppetserver/shell_utils.clj
haus: Updated. Approaching 80 now.

haus:
(library request filed at https://tickets.puppetlabs.com/browse/RE-6241)

src/clj/puppetlabs/puppetserver/shell_utils.clj
```diff
@@ -0,0 +1,96 @@
...
+  {:exit-code schema/Int
+   :stderr schema/Str
+   :stdout schema/Str})
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Internal
+
+(schema/defn ^:private ^:always-validate
+  validate-command
+  "Checks the command string to ensure that is is an absolute path, executable
```
nwolfe: s/is is/it is

src/clj/puppetlabs/puppetserver/shell_utils.clj
```diff
@@ -0,0 +1,96 @@
...
+  and that the file exists. An exception is thrown is any of those are not the
```
nwolfe: s/is any/if any

src/clj/puppetlabs/puppetserver/shell_utils.clj
```diff
@@ -0,0 +1,96 @@
...
+  "A map that contains the details of the result of executing a command."
+  {:exit-code schema/Int
+   :stderr schema/Str
+   :stdout schema/Str})
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Internal
+
+(schema/defn ^:private ^:always-validate
+  validate-command
```
nwolfe: FWIW I believe our other `validate-foo` functions are explicitly marked with a trailing `!` to indicate we're using it for the side-effect (of throwing an exception)

src/clj/puppetlabs/puppetserver/shell_utils.clj
```diff
@@ -0,0 +1,96 @@
...
+  "Checks the command string to ensure that is is an absolute path, executable
+  and that the file exists. An exception is thrown is any of those are not the
+  case."
+  [command :- schema/Str]
+  (let [command-file (io/as-file command)]
+    (cond
+      (not (.isAbsolute command-file))
+      (throw (IllegalArgumentException.
+              (format "An absolute path is required, but %s is not an absolute path" command)))
```
nwolfe: Should probably throw a pair of `'` around `%s` for readability like you did below

src/clj/puppetlabs/puppetserver/shell_utils.clj
```diff
@@ -0,0 +1,96 @@
...
+      (throw (IllegalArgumentException.
+              (format "The referenced command '%s' is not executable" command))))))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Public
+
+(schema/defn ^:always-validate
+  execute-command :- ExecutionResult
+  "Execute the specified fully qualified command and any included options and
+  return the exit-code, and the contents of stdout and stderr for the command."
```
nwolfe: Are the `options` for the command itself, or for this function?  I wonder because there's a general pattern where the last argument of a function can be a map that allows for all sorts of various "options" without having to create an explicit function signature for each optional argument.

So, we should probably just mention that the options will be given to the command and aren't going to be used by this function itself (which was initially what I was expecting).

src/clj/puppetlabs/puppetserver/shell_utils.clj
```diff
@@ -0,0 +1,96 @@
...
+   (try
+     (let [process (ProcessWrapper. (. (Runtime/getRuntime) exec (into-array (concat [command] options))))]
+       {:exit-code (.getExitCode process)
+        :stderr (.getErrorString process)
+        :stdout (.getOutputString process)})
+     (catch IOException e
+       (throw (IllegalStateException. (format "Exception while executing '%s': %s" command (.getMessage e))))))))
+
+(schema/defn ^:always-validate
+execute-command' :- ExecutionResult
```
nwolfe: whitespace

nwolfe: Whoa, didn't notice the `'` at the end of the function name.  

If this is meant to be a non-public or subordinate function that's used by `execute-command`, then there are existing naming patterns like `do-execute-command` or similar that we should probably stick to.

If this is supposed to be an alternate execute-command that deals with streams instead of strings, we should at least mention that in the docstring and probably rename it appropriately too

src/clj/puppetlabs/puppetserver/shell_utils.clj
```diff
@@ -0,0 +1,96 @@
...
+         (. setStreamHandler (PumpStreamHandler. outstream errstream))
+         (. setExitValues nil))
+       (let [exit-code
+             (. executor execute (. commandline addArguments (into-array options) false))]
+         {:exit-code exit-code
+          :stderr (slurp (ByteArrayInputStream. (.toByteArray errstream)))
+          :stdout (slurp (ByteArrayInputStream. (.toByteArray outstream)))})))))
+
+(schema/defn ^:always-validate
+execute-command'' :- ExecutionResult
```
nwolfe: ahh

src/java/com/puppetlabs/puppetserver/ExecutionStubImpl.java
```diff
@@ -10,25 +14,70 @@
...
-     * @return A 2-element array; the first element is the output of the process.
-     *      The second element is the exit code of the process.
+     * @param commandLine commandLine instance to execute
+     * @return An ExecutionStubResult with output, error, and the exit code of the process
      *
      * @throws InterruptedException
      * @throws IOException
      */
-    public static ExecutionStubResult executeCommand(String command)
+    private static ExecutionStubResult executeExecutor(org.apache.commons.exec.CommandLine commandLine)
```
nwolfe: We should just add a normal `import` for this like we did the others

src/java/com/puppetlabs/puppetserver/ExecutionStubImpl.java
```diff
@@ -10,25 +14,70 @@
...
+        // Don't throw exception on non-zero exit code
+        executor.setExitValues(null);
 
-        ProcessWrapper wrapper = new ProcessWrapper(Runtime.getRuntime().exec(command));
+        // Set up the handlers
+        executor.setStreamHandler(streamHandler);
+
+        Integer exitCode = executor.execute(commandLine);
+
+        String stdErr = errStream.toString();
```
nwolfe: Do we need to close the streams when we're done with them?

Can't remember, been a while since I worked with java streams

src/java/com/puppetlabs/puppetserver/ExecutionStubImpl.java
```diff
@@ -10,25 +14,70 @@
...
     /**
-     * Executes the given command in a separate process.
+     * Takes a prepared CommandLine instance and executes it using some sane
+     * defaults and a DefaultExecutor. Also makes a delicious pancake.
      *
-     * @param command the command to execute
-     * @return A 2-element array; the first element is the output of the process.
-     *      The second element is the exit code of the process.
+     * @param commandLine commandLine instance to execute
+     * @return An ExecutionStubResult with output, error, and the exit code of the process
```
nwolfe: I think in all of these docstrings (both clojure and java) we need to be specific with the data types being returned and say whether they're a string or stream

nwolfe:
This is awesome! :metal:

src/clj/puppetlabs/puppetserver/shell_utils.clj
nwolfe: Aw had a discussion IRL with @haus - now I understand we're going to choose between one of these 3 implementations before merging

haus:
@nwolfe since we've now settled on commons-exec. I'm going to do a pass and clean up all of the unneeded functions and code.

haus:
@nwolfe this is getting better. I think i responded to most of your feedback. I'm going to do another pass on the docstrings still.

src/java/com/puppetlabs/puppetserver/ExecutionStubImpl.java
haus: I am super rusty here. Do not know.

project.clj
```diff
@@ -50,7 +50,8 @@
                  [puppetlabs/ssl-utils "0.8.1"]
                  [puppetlabs/dujour-version-check "0.1.2" :exclusions [org.clojure/tools.logging]]
                  [puppetlabs/http-client "0.4.4"]
-                 [puppetlabs/comidi "0.3.1"]]
+                 [puppetlabs/comidi "0.3.1"]
+                 [org.apache.commons/commons-exec "1.3"]]
```
nwolfe: Not a big deal but to keep these ordered this would probably go above, underneath `liberator` in that section.
Ones at the bottom are for our own PL libs.

More info: https://confluence.puppetlabs.com/pages/viewpage.action?pageId=48956980

haus: cool

src/java/com/puppetlabs/puppetserver/ExecutionStubImpl.java
```diff
@@ -1,8 +1,12 @@
...
+import org.apache.commons.exec.DefaultExecutor;
+import org.apache.commons.exec.PumpStreamHandler;
+import org.apache.commons.exec.CommandLine;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 
 public class ExecutionStubImpl {
```
haus: As this class is no longer a Stub (it will be used in more than just the one jruby puppet method call), we should probably rename it. Any suggestions? Here are a few ideas: ExecutionShell, ShellExecutor, ShellCommons...

haus: ShellUtils

nwolfe: Haus mentioned this code could end up extracted into its own library so that code manager can use it as well.  And the clojure side of this code is called `shell-utils`.  Also, it sounded like it could end up doing more than just *executing* shell commands....

So I think I like ShellUtils

src/clj/puppetlabs/puppetserver/shell_utils.clj
```diff
@@ -0,0 +1,56 @@
...
+           (java.io IOException)))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Schemas
+
+(def ExecutionResult
+  "A map that contains the details of the result of executing a command."
+  {:exit-code schema/Int
+   :stderr schema/Str
+   :stdout schema/Str})
```
cprice404: At some point we will need a way to get `Stream` objects back out of stderr/stdout, instead of Strings, but that doesn't need to be in scope for this PR as long as you can see a path forward.  I can imagine possibly an alternate function that you'd call to initiate the execution, and a different schema that it returns.

src/clj/puppetlabs/puppetserver/shell_utils.clj
```diff
@@ -0,0 +1,56 @@
...
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Public
+
+(schema/defn ^:always-validate
+  execute-command :- ExecutionResult
+  "Execute the specified fully qualified command (string) and any included
+  command-arguments (vector of strings) and return the exit-code (integer),
+  and the contents of the stdout (string) and stderr (string) for the command."
+  ([command :- schema/Str]
+   (execute-command command [""]))
```
cprice404: Any reason you need an empty string in the default arg list here, as opposed to just an empty vector?

src/clj/puppetlabs/puppetserver/shell_utils.clj
```diff
@@ -0,0 +1,56 @@
...
+  ([command :- schema/Str
+    command-arguments :- [schema/Str]]
+   (validate-command! command)
+   (try
+     (let [process (ShellUtils/executeCommand command (into-array command-arguments))]
```
cprice404: Might be worth doing `(into-array String command-arguments)` here, so there's no guesswork involved as to the type of the array.

src/java/com/puppetlabs/puppetserver/ExecutionStubImpl.java
```diff
@@ -1,34 +0,0 @@
...
-        ProcessWrapper wrapper = new ProcessWrapper(Runtime.getRuntime().exec(command));
-
-        String stdErr = wrapper.getErrorString();
-        if ( ! stdErr.isEmpty() ) {
-            log.warn("Executed an external process which logged to STDERR: " + stdErr);
-        }
-
-        return new ExecutionStubResult(wrapper.getOutputString(), wrapper.getExitCode());
-    }
-}
```
cprice404: yay!

src/java/com/puppetlabs/puppetserver/ProcessWrapper.java
```diff
@@ -1,73 +0,0 @@
...
-                    if (reader != null) {
-                        reader.close();
-                    }
-                } catch (IOException e) {
-                    log.warn("Attempt to close stream failed", e);
-                }
-            }
-        }
-    }
-}
```
cprice404: omg yay!

src/clj/puppetlabs/puppetserver/shell_utils.clj
haus: The java side expects an array of strings to be handed to addArguments. I could do some sort of check to see if the vector is empty and then do the right thing.

src/java/com/puppetlabs/puppetserver/ProcessWrapper.java
KevinCorcoran: :fireworks:

cprice404:
:see_no_evil: but didn't review super-thoroughly.  I'm fine w/going this route as long as:

a. we have a rough idea of how we will add stream support in the future, and
b. we wire up the ruby code to use this and validate that the server runs OK with the changes

src/clj/puppetlabs/puppetserver/shell_utils.clj
cprice404: I think `(into-array String [])` would resolve this.

haus: Yep. Totally does.

src/java/com/puppetlabs/puppetserver/ShellUtils.java
```diff
@@ -0,0 +1,88 @@
...
+     *
+     * @throws InterruptedException
+     * @throws IOException
+     */
+    private static ExecutionResult executeExecutor(CommandLine commandLine)
+            throws InterruptedException, IOException {
+        DefaultExecutor executor = new DefaultExecutor();
+        ByteArrayOutputStream errStream = new ByteArrayOutputStream();
+        ByteArrayOutputStream outStream = new ByteArrayOutputStream();
+        PumpStreamHandler streamHandler = new PumpStreamHandler(outStream, errStream);
```
cprice404: @haus did you find any docs on this class?  (I mean, ones that weren't useless)

haus: yesssssss

haus: There are these: https://commons.apache.org/proper/commons-exec/apidocs/org/apache/commons/exec/PumpStreamHandler.html. I don't know how they compare, but they seemed reasonable.

cprice404: Cursory glance at the javadocs of this class makes it sound like it's doing the 'pumping' work on background threads, which is what we were doing in our other implementation, and what is necessary in order to prevent the OS buffer from filling up.

test/unit/puppetlabs/puppetserver/shell_utils_test.clj
```diff
@@ -0,0 +1,36 @@
...
+  (testing "passes the expected number of args to cmd"
+    (is (= 5 (:exit-code (sh-utils/execute-command (script-path "num-args") ["a" "b" "c" "d" "e"]))))))
+
+(deftest throws-exception-for-non-absolute-path
+  (testing "Commands must be given using absolute paths"
+    (is (thrown? IllegalArgumentException (sh-utils/execute-command "echo")))))
+
+(deftest throws-exception-for-non-existent-file
+  (testing "The given command must exist"
+    (is (thrown? IllegalArgumentException (sh-utils/execute-command "/usr/bin/footest")))))
```
cprice404: would it be crazypants to add a test that executed a command whose stdout was larger than the typical size of an OS command output buffer?

haus: No, that seems sane to me. I don't know that size off the top of my head, but i can probably find out.

cprice404: `man pipe` on my linux box shows the default buffer size being 65536.  Doubling that for a test would only use 128k of memory, which seems not-entirely-crazypants.

cprice404: [StackOverflow post](http://unix.stackexchange.com/questions/11946/how-big-is-the-pipe-buffer) from earlier this year says Mac defaults to 16384 and can go up to 65536, so 128k seems reasonable to me.

nwolfe: :+1:

KevinCorcoran:
@haus did you do the following (quoth @cprice404):

> wire up the ruby code to use this and validate that the server runs OK with the changes

dev-resources/puppetlabs/puppetserver/shell_utils_test/echo
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env sh
+
+echo $@
```
nathanielksmith: would it be bad to have newlines at the end of these?

KevinCorcoran:
add clojure test which exercises execution stub by writing custom parser function which calls `Puppet::Util::Execute` or whatev

dev-resources/puppetlabs/puppetserver/shell_utils_test/echo
haus: I don't think so. I can add them.

src/java/com/puppetlabs/puppetserver/ShellUtils.java
```diff
@@ -0,0 +1,88 @@
...
+
+        // Don't throw exception on non-zero exit code
+        executor.setExitValues(null);
+
+        // Set up the handlers
+        executor.setStreamHandler(streamHandler);
+
+        Integer exitCode = executor.execute(commandLine);
```
nathanielksmith: Does the executor have a timeout? Is it configurable? Might we want end-users to be able to set a timeout?

haus: Yes, this will be desirable for code manager's use case I think. I was considering opening a ticket to capture that improvement.

nathanielksmith: yeah, I'd def open that ticket.

haus: https://tickets.puppetlabs.com/browse/SERVER-1100 opened.

haus:
@KevinCorcoran test added. I think the functions I call should be broken out into test utilities, but I don't know if that needs to happen as part of this PR.

KevinCorcoran:
I left some nitpicks on the new test but I'm going to merge this PR since they are not show stoppers and other work depends upon it.

PR #850, opened 2016-01-06
[DO NOT MERGE] (SERVER-1062) puppetserver config subcommand

nwolfe:
Initial attempt at a `puppetserver config print|set` subcommand.
Supports reading and writing, batches, and optionally printing as JSON.

Writing assumes configuration files are HOCON.

---

* REQUIRES JAVA 8 FOR NEWER TYPESAFE CONFIG VERSION
  * Need to ensure our CI is set up for this, which it might already be
* Only works with HOCON files
  * Which is all that we happen to be shipping, but lower-level TK utilities this is built upon support more formats

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -16,9 +16,11 @@
   :dependencies [[org.clojure/clojure "1.7.0"]
 
                  ;; begin version conflict resolution dependencies
-                 [puppetlabs/typesafe-config "0.1.4"]
+                 [puppetlabs/typesafe-config "0.1.4" :exclusions [com.typesafe/config]]
                  [org.clojure/tools.macro "0.1.5"]
                  ;; end version conflict resolution dependencies
+
+                 [com.typesafe/config "1.3.0"]
 
                  [cheshire "5.3.1"]
                  [slingshot "0.10.3"]
@@ -53,7 +55,7 @@
 
   :main puppetlabs.trapperkeeper.main
 
-  :pedantic? :abort
+  ;; :pedantic? :abort
 
   :source-paths ["src/clj"]
   :java-source-paths ["src/java"]
--- a/src/clj/puppetlabs/puppetserver/cli/config.clj
+++ b/src/clj/puppetlabs/puppetserver/cli/config.clj
@@ -1,0 +1,78 @@
+(ns puppetlabs.puppetserver.cli.config
+  (:import [com.typesafe.config ConfigUtil]
+           [com.typesafe.config.parser ConfigDocumentFactory ConfigDocument])
+  (:require [cheshire.core :as json]
+            [puppetlabs.puppetserver.cli.subcommand :as cli]
+            [slingshot.slingshot :refer [throw+]]))
+
+(defn read-config-settings!
+  [config cmd-args]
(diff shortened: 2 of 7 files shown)


src/clj/puppetlabs/puppetserver/cli/config.clj
```diff
@@ -0,0 +1,78 @@
...
+  Get multiple configuration settings as a JSON structure:
+    $ puppetserver config print jruby-puppet.master-code-dir jruby-puppet.master-conf-dir --as-json
+    {'jruby-puppet.master-code-dir':'/etc/puppetlabs/code','jruby-puppet.master-conf-dir':'/etc/puppetlabs/puppet'}
+
+  Change (or add) a configuration setting:
+    $ puppetserver config set profiler.enabled true
+
+  Change (or add) multiple configuration settings:
+    $ puppetserver config set profiler.enabled true jruby-puppet.gem-home /my/gems
+  "
```
nwolfe: API above

src/clj/puppetlabs/puppetserver/cli/subcommand.clj
```diff
@@ -30,16 +30,25 @@
...
    line arguments specific to the subcommand.
 
-   The `run-fn` must accept a configuration map and argument sequence."
+   The `run-fn` must accept a configuration map and argument sequence.
+   The configuration map will look like:
+    {
+      :config       {<the fully parsed & merged configuration map>}
+      :config-path  '/opt/puppetlabs/server/app/conf.d/'
+      :config-files [<File foo.conf>, <File bar.conf>, <File qux.conf>]
+    }
```
nwolfe: I struggled to determine how best to support the data hand-off between this general function and the specific subcommand "run" functions.  
The problem was that it only provided the final, merged map of configuration settings to the subcommand, but I needed the actual file names so I could write to them.  For read-only actions, I could reuse the final, merged map though, so I kind of wanted access to both.

That's how I ended up with generalizing the existing "config" parameter (that's passed down to subcommand run functions) from just the final, merged map to instead encompassing all the configuration-related info (which is why I put the :config-path in there even though I wasn't using it in this subcommand).

I also played with the idea of having multiple `cli/run` functions that subcommand could call, but in the end I went with the option that required the fewest code changes.

Suggestions very much welcome for this part.

project.clj
```diff
@@ -16,10 +16,12 @@
   :dependencies [[org.clojure/clojure "1.7.0"]
 
                  ;; begin version conflict resolution dependencies
-                 [puppetlabs/typesafe-config "0.1.4"]
+                 [puppetlabs/typesafe-config "0.1.4" :exclusions [com.typesafe/config]]
                  [org.clojure/tools.macro "0.1.5"]
                  ;; end version conflict resolution dependencies
 
+                 [com.typesafe/config "1.3.0"]
```
nwolfe: The version of `com.typesafe/config` we're using in our `puppetlabs/typesafe-config` wrapper is outdated by a couple minor versions. It doesn't preserve formatting or comments during file modifications.

Should be easy to upgrade our wrapper to the newer version and get rid of this override.

nwolfe:
Replaced by #851

PR #852, opened 2016-01-07
(MAINT) Exclude org.clojure from lein-ezbake

nwolfe:
Instead of overriding org.clojure in ezbake profile with an explicit
version, just exclude the one dragged in by the lein-ezbake plugin.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,3 @@
-(def clojure-version "1.7.0")
 (def tk-version "1.1.3")
 (def tk-jetty-version "1.3.1")
 (def ks-version "1.1.0")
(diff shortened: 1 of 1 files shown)


nwolfe:
Follow-up to #848

cprice404:
+1, I agree this is better, thanks for doing it @nwolfe .  Would prefer to hold off on merging this until we get the voom build out though.

camlow325:
I started having second thoughts about this a bit.  Everywhere else, we've moved to a policy of resolving dependency version conflicts by specifying the explicit version to be used at the top level.  @cprice404's #848 PR seems to follow that same policy.  If we were to merge in this PR, seems like we'd be going against that policy.

That said, I don't have super-strong feelings about this so if @cprice404 and @nwolfe are both onboard with the exclusion approach, I won't object to merging it in.

cprice404:
Generally speaking I prefer avoiding exclusions, but for clojure specifically, and ezbake specifically, it seems like a reasonable place to make an exception, and I really hate having that extra version variable at the time.  So I'm +1 but I don't care all that much.  Maybe someone else could weigh in as a tiebreaker.

fpringvaldsen:
@cprice404 @camlow325 @nwolfe I also don't have a very strong opinion on this, but I think I'm leaning towards the approach here in this PR. Making an exception for clojure seems worthwhile to me

PR #855, opened 2016-01-11
(MAINT) Rename `clojure-version` to avoid shadowing

cprice404:
In a recent PR, we introduced a var called `clojure-version` in
the project.clj, to coordinate our references to our clojure
dependency version.  It turns out that the name `clojure-version`
is already in use in `clojure.core`, so this var results in a
warning that we are shadowing the upstream var.

This commit changes the var name to `clj-version`, so that we
will no longer be shadowing the built-in name.


nwolfe:
:+1:

PR #856, opened 2016-01-11
Should resolve CI failure.  Improves TKAuth test...

erikPrime:
- modifies the authorized? method to respect the case where the hostname is authorized.
- modifies the calculate_expected_result method to treat any HTTP response in the 200-299 range as a PASS.


camlow325:
:see_no_evil:  These look fine to me.  Did you run these and see them pass at least locally?

erikPrime:
@camlow325 Yes, I ran this change locally against the puppet-server branch.  I have not run it against the pe-puppet-server branch.

camlow325:
Works for me.  Merging this...

PR #857, opened 2016-01-11
(SERVER-1089) Skip PuppetDB installation in PE

nwolfe:
The puppetdb integration test can assume that PuppetDB is already
installed and configured when testing PE.


camlow325:
:+1:

PR #860, opened 2016-01-12
(SERVER-1091) Add .isLocked() method for JRubyPool

camlow325:
This commit adds an .isLocked() method to the JRubyPool, mostly for ease
of use in upstream tests for detecting whether or not the pool is
locked.

### DIFF
--- a/src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
+++ b/src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
@@ -307,6 +307,19 @@ public final class JRubyPool<E> implements LockablePool<E> {
     }
 
     @Override
+    public boolean isLocked() {
+        boolean locked;
+        final ReentrantLock lock = this.queueLock;
+        lock.lock();
+        try {
+            locked = isPoolLockHeld();
+        } finally {
+            lock.unlock();
+        }
+        return locked;
+    }
+
+    @Override
     public void unlock() {
         final ReentrantLock lock = this.queueLock;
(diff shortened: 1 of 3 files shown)


src/java/com/puppetlabs/puppetserver/pool/JRubyPool.java
```diff
@@ -307,6 +307,19 @@ public void lock() throws InterruptedException {
     }
 
     @Override
+    public boolean isLocked() {
+        boolean locked;
+        final ReentrantLock lock = this.queueLock;
+        lock.lock();
+        try {
```
camlow325: A try/finally here may not strictly be necessary since `.isPoolLockHeld` just does a variable comparison.  Thought it would be safest to follow the pattern in other methods and future-proof this in the event that `.isPoolLockHeld` might ever have to do something more involved that could throw an exception.

cprice404:
+1

src/java/com/puppetlabs/puppetserver/pool/LockablePool.java
```diff
@@ -149,6 +149,15 @@ E borrowItemWithTimeout(long timeout, TimeUnit unit)
     */
     void lock() throws InterruptedException;
 
+    /**
+     * Returns whether or not the pool is currently locked.  Note that the
+     * value returned may no longer be accurate by the time it is consumed by
```
KevinCorcoran: lol, awesome

KevinCorcoran:
:see_no_evil:

PR #864, opened 2016-01-14
(SERVER-1097) Use timed waits for lockable-pool-test promises

camlow325:
This commit modifies logic in the `lockable-pool-test` tests to use a
timed wait with assertions for all promises other than just ones related
to thread start synchronization.  Hopefully, this would help better
pinpoint any test failures when they occur, as opposed to just having a
test potentially hang and be aborted by a test runner without any better
information to help diagnose the problem.

### DIFF
--- a/test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
+++ b/test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
@@ -3,6 +3,10 @@
   (:import (com.puppetlabs.puppetserver.pool JRubyPool)
            (java.util.concurrent TimeUnit ExecutionException)))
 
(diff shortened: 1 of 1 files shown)


camlow325:
With the changes in this branch, I ran the `lockable-pool-test` suite 200 times locally and didn't get any failures.  With all of the timing sensitivity, thought that would be good to try.

camlow325:
I didn't bother changing the promise derefs that were about just seeing that a launched thread was started to be timed.  Figured we're not trying to test Clojure/Java themselves here and that those should be trustworthy.  I really hope anyway...

cprice404:
+1

PR #867, opened 2016-01-15
(SERVER-1072) Add `with-puppet-conf-files` test utility

cprice404:
This commit just has a minor refactor to our JRuby test utilities
that is useful in downstream projects like pe-puppet-server-extensions
and pe-aio.

### DIFF
--- a/test/unit/puppetlabs/services/jruby/jruby_testutils.clj
+++ b/test/unit/puppetlabs/services/jruby/jruby_testutils.clj
@@ -11,7 +11,8 @@
            (org.jruby.embed LocalContextScope)
            (puppetlabs.services.jruby.jruby_puppet_schemas JRubyPuppetInstance)
            (clojure.lang IFn)
-           (com.puppetlabs.puppetserver.jruby ScriptingContainer)))
+           (com.puppetlabs.puppetserver.jruby ScriptingContainer)
+           (java.io File)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; Constants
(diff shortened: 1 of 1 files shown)


test/unit/puppetlabs/services/jruby/jruby_testutils.clj
```diff
@@ -29,6 +30,39 @@
...
+  * values are paths to a test-specific file that should be copied to puppet's confdir,
+    a la `$confdir/<key>`.
+
+  For each entry in the map, copies the files into $confdir, runs the test function,
+  and then deletes all of the files from $confdir.
+
+  Optionally accepts a second argument `dest-dir`, which specifies the location
+  of Puppet's $confdir.  If this argument is not provided, defaults to
+  `jruby-testutils/conf-dir`."
+  ([puppet-conf-files]
```
haus: Probably worth copying the below schema into this section.

cprice404: I had it like that before, but since that schema is long and gross and this fn arity just calls into that one, I thought maybe it was better not to repeat it.

test/unit/puppetlabs/services/jruby/jruby_testutils.clj
```diff
@@ -29,6 +30,39 @@
...
+   (with-puppet-conf-files puppet-conf-files conf-dir))
+  ([puppet-conf-files :- {schema/Str (schema/pred (some-fn string? #(instance? File %)))}
```
cprice404: Also, for the record, in the past I've fought against having schemas like this that tolerate a string or a File because we used to use them in some production code and it caused a lot of very-hard-to-debug issues... but I kind of caved on this one because it's just a testutil.

test/unit/puppetlabs/services/jruby/jruby_testutils.clj
haus: :+1:

nwolfe:
:metal:

PR #868, opened 2016-01-18
(SERVER-1071) Make legacy routes tolerant of multi-server config

cprice404:
Prior to this commit, the legacy routes services made some
hard-coded assumptions about the master's route configuration that
made it incompatible with the recommended webrouting configuration
style for multi-webserver configurations.  This commit adds a few
more checks to detect that case, and make the services compatible
with the multi-server configuration style.  This is necessary in
order for us to be able to run PE Puppet Server in-process with
other PE web apps.

### DIFF
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -134,8 +134,16 @@
   [master-ns :- schema/Keyword
    config-route]
   (cond
-    (and (map? config-route) (contains? config-route :master-routes))
-    (:master-routes config-route)
+    ;; if the route config is a map, we need to determine whether it's the
+    ;; new-style multi-server config (where there will be a `:route` key and a
+    ;; `:server`, key), or the old style where there is a single key that is
+    ;; assumed to be our hard-coded route id (`:master-routes`).
+    ;; It should be possible to delete this hack (perhaps this entire function)
+    ;; when we remove support for legacy routes.
+    (and (map? config-route) (or (contains? config-route :route)
+                                 (contains? config-route :master-routes)))
+    (or (:route config-route)
+        (:master-routes config-route))
 
     (string? config-route)
     config-route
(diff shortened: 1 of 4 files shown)


src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -130,8 +130,16 @@
...
   (cond
-    (and (map? config-route) (contains? config-route :master-routes))
-    (:master-routes config-route)
+    ;; if the route config is a map, we need to determine whether it's the
+    ;; new-style multi-server config (where there will be a `:route` key and a
+    ;; `:server`, key), or the old style where there is a single key that is
+    ;; assumed to be our hard-coded route id (`:master-routes`).
+    ;; It should be possible to delete this hack (perhaps this entire function)
+    ;; when we remove support for legacy routes.
+    (and (map? config-route) (or (contains? config-route :route)
```
fpringvaldsen: Do we also want to support multi-route, multi-server configs? In that case I believe it would look something like

```
{
  :web-router-service {
    :puppetlabs.services.master.master-service/master-service {
      :master-routes {
        :route "/puppet"
        :server "puppet-server"
      }
    }
  }
}
```
It would let us mount multiple routes through the master service, but if we don't plan to ever do that then this is fine as-is.

cprice404: Since I'm hoping that we can kill all of this with fire soon-ish, I think I'd be OK with erring on the side of "add in a hack to support that as soon as a situation necessitates it", but I can try to do it now if people think it's necessary.

camlow325:
:+1: other than that there now appears to be a merge conflict, needs a rebase?

src/clj/puppetlabs/services/master/master_core.clj
camlow325: I'm fine with waiting until the need arises.  Like you mentioned, I'm hoping that the legacy routes service would die long before / if we'd ever have a need to host multiple route-ids under the master-service.  Seems like if that were to come up that any failures by the legacy routes service to handle it would be pretty obvious.

cprice404:
@camlow325 rebased

camlow325:
:+1: from me.  @fpringvaldsen - were you also good with merging this?

fpringvaldsen:
@camlow325 Yup, I'm :+1: as well. I'll go ahead and merge this

PR #874, opened 2016-01-22
(SERVER-1069) bundle logback json encoder

cprice404:
This PR adds a dependency on the logback logstash JSON encoder library, so that we will bundle it with our uberjar.  This allows users to opt-in to configuring their Puppet Server logs to write output in a JSON format that can be imported into logstash.

The PR also includes documentation with examples of how to configure Puppet Server's logging accordingly.

We were almost certainly going to end up with a dependency on this library in the future anyway, because it is a dependency of the structured logging library that PuppetDB is now using.

### DIFF
--- a/documentation/config_logging_advanced.md
+++ b/documentation/config_logging_advanced.md
@@ -1,0 +1,130 @@
+---
+layout: default
+title: "Puppet Server: Advanced Logging Configuration"
+canonical: "/puppetserver/latest/config_logging_advanced.html"
+---
+
+Puppet Server uses the very flexible, very powerful [Logback](http://logback.qos.ch/) library under the hood to handle all of our logging needs.
+
+Among the many interesting and useful capabilities of logback is the ability to configure it to log messages in a JSON format, which makes it easy to send them over to other logging backends such as logstash.
+
+## Configuring Puppet Server for use with Logstash
+
+There are a few steps necessary to get your Puppet Server logging set up for use with logstash.  The first step is to modify your logging configuration so that Puppet Server is logging in a JSON format.  After that, you'll need to configure an external tool to monitor these JSON files and send the data to logstash (or another remote logging system).
+
+### Configuring Puppet Server to log to JSON
+
+There are a handful of considerations you may want to take into account before configuring Puppet Server to log to JSON:
+
+* Do you want to configure Puppet Server to *only* log to JSON?  In lieu of the normal text file logging?  Or do you want to add JSON logging *in addition to* your normal text logging?
+* Do you want to set up JSON logging only for the normal Puppet Server logs (i.e. `puppetserver.log`), or also for the HTTP access logs (i.e. `puppetserver-access.log`)?
+* What kind of log rotation strategy do you want to use for the new JSON log files?
+
+In this section we'll show some examples of setting up your configuration for logging both the normal logs and the HTTP access logs in JSON, *in addition to* leaving the normal text-based log files in place.  We'll also configure logback to do log rotation on these new JSON files.  Hopefully you can modify these examples to suit your own needs, based on your answers to the three questions above.
+
+**NOTE**: Puppet Server's packaging sets up `logrotate` management of the main `puppetserver.log` file automatically, which is why you don't see any of the `logback` support for log rotation configured in our default `logback.xml` file.  However, the Puppet Server packaging does *not* include any `logrotate` integration for these new JSON log files that we'll be setting up, so that is why we will show examples of how to use `logback` to do that.  If you were so inclined, you could omit those portions of these configuration examples, and configure `logrotate` to manage the log rotation for your new JSON files.
+
+#### Adding a JSON version of the "normal" Puppet Server logs
+
+To configure Puppet Server to log its "normal" logs to a second log file, in JSON format, simply add a section like this to your `logback.xml` file:
+
+``` xml
+<appender name="LOGSTASH" class="ch.qos.logback.core.rolling.RollingFileAppender">
+        <file>/var/log/puppetlabs/puppetserver/puppetserver.log.json</file>
+
+        <rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
+            <fileNamePattern>/var/log/puppetlabs/puppetserver/puppetserver.log.json.%d{yyyy-MM-dd}</fileNamePattern>
+            <maxHistory>5</maxHistory>
+        </rollingPolicy>
+
+        <encoder class="net.logstash.logback.encoder.LogstashEncoder">
+            <fieldNames>
+                <timestamp>timegenerated</timestamp>
+                <message>logmsg</message>
+            </fieldNames>
+        </encoder>
+    </appender>
+```
+
+Note that the above configuration includes the `RollingFileAppender`, to rotate the log files and ensure they don't fill up your disk.  Also, note that the `fieldNames` section above is optional, and in this case is restricting the number of fields that will be included in the JSON to two fields per log entry.  There are additional fields available, and if you omitted the `fieldNames` section you would end up with several additional fields per log entry by default.  For more info see the [Logstash Logback Encoder Docs](https://github.com/logstash/logstash-logback-encoder/blob/master/README.md#loggingevent-fields).
+
+The above XML snippet will create an appender that will log to the specified JSON file when active.  In order to activate the appender, you also need to add an `appender-ref`, e.g. in the `<root>` section of your `logback.xml`:
+
+``` xml
+<root level="info">
+        <appender-ref ref="FILE"/>
+        <appender-ref ref="LOGSTASH"/>
+</root>
+```
+
+You could also comment out or remove the other `appender-ref`s, if you decided you *only* wanted to log the JSON format.
+
+#### Adding a JSON version of the Puppet Server HTTP Acces logs
+
+To add JSON logging for Puppet Server's HTTP requests, modify the `request-logging.xml` file.  Here's an example of adding a logback appender there:
+
+``` xml
+   <appender name="LOGSTASH" class="ch.qos.logback.core.rolling.RollingFileAppender">
+        <file>/var/log/puppetlabs/puppetserver/puppetserver-access.log.json</file>
+
+        <rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
+            <fileNamePattern>/var/log/puppetlabs/puppetserver/puppetserver-access.log.json.%d{yyyy-MM-dd}</fileNamePattern>
+            <maxHistory>30</maxHistory>
+        </rollingPolicy>
+
+      <encoder class="net.logstash.logback.encoder.AccessEventCompositeJsonEncoder">
+          <providers>
+              <version/>
+              <pattern>
+                  <pattern>
+                    {
+                      "timegenerated":"%date{yyyy-MM-dd'T'HH:mm:ss.SSSXXX}",
+                      "clientip":"%remoteIP",
+                      "auth":"%user",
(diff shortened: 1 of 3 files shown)


cprice404:
This replaces #787

cprice404:
@Ziaunys @danieldreier here's the PR I mentioned to you guys the other day, finally :)  If you have time to take a peek that'd be appreciated.

@gguillotte ping since this'll end up on the public docs site

cprice404:
@mfournier if you are interested in reviewing this to make sure it meets your needs and that the docs make sense, please do!

gguillotte:
Thanks, @cprice404. I'll review and make sure there's a docs repo update to add this to the site nav, and I might have a follow-up PR post-merge for small style tweaks.

documentation/config_logging_advanced.md
```diff
@@ -0,0 +1,130 @@
+---
+layout: default
+title: "Puppet Server: Advanced Logging Configuration"
+canonical: "/puppetserver/latest/config_logging_advanced.html"
+---
+
+Puppet Server uses the very flexible, very powerful [Logback](http://logback.qos.ch/) library under the hood to handle all of our logging needs.
```
gguillotte: Can we add a line to note that it's configured in `logback.xml`, located by default at `/etc/puppetserver/logback.xml` or `/etc/puppetlabs/puppetserver/logback.xml`? I know it's redundant with what's in `configuration.html`, but if people wind up on this page through a search result they won't have that context.

documentation/config_logging_advanced.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Configuring Puppet Server for use with Logstash
+
+There are a few steps necessary to get your Puppet Server logging set up for use with logstash.  The first step is to modify your logging configuration so that Puppet Server is logging in a JSON format.  After that, you'll need to configure an external tool to monitor these JSON files and send the data to logstash (or another remote logging system).
+
+### Configuring Puppet Server to log to JSON
+
+There are a handful of considerations you may want to take into account before configuring Puppet Server to log to JSON:
+
+* Do you want to configure Puppet Server to *only* log to JSON?  In lieu of the normal text file logging?  Or do you want to add JSON logging *in addition to* your normal text logging?
```
gguillotte: What does "normal" mean? If it means plain text, can we say that instead of "normal"? For instance:

> * Do you want to configure Puppet Server to *only* log to JSON in lieu of the default plain-text file logging? Or do you want to add JSON logging *in addition to* the default plain-text logging?

documentation/config_logging_advanced.md
```diff
@@ -0,0 +1,130 @@
...
+
+**NOTE**: Puppet Server's packaging sets up `logrotate` management of the main `puppetserver.log` file automatically, which is why you don't see any of the `logback` support for log rotation configured in our default `logback.xml` file.  However, the Puppet Server packaging does *not* include any `logrotate` integration for these new JSON log files that we'll be setting up, so that is why we will show examples of how to use `logback` to do that.  If you were so inclined, you could omit those portions of these configuration examples, and configure `logrotate` to manage the log rotation for your new JSON files.
+
+#### Adding a JSON version of the "normal" Puppet Server logs
+
+To configure Puppet Server to log its "normal" logs to a second log file, in JSON format, simply add a section like this to your `logback.xml` file:
```
gguillotte: Does the location where this is added matter? Can you add it to the beginning of the file, or the end, or anywhere? Is there a part of the file where it _shouldn't_ be added? (I'm assuming it shouldn't be inside another XML element?)

documentation/config_logging_advanced.md
```diff
@@ -0,0 +1,130 @@
...
+        <encoder class="net.logstash.logback.encoder.LogstashEncoder">
+            <fieldNames>
+                <timestamp>timegenerated</timestamp>
+                <message>logmsg</message>
+            </fieldNames>
+        </encoder>
+    </appender>
+```
+
+Note that the above configuration includes the `RollingFileAppender`, to rotate the log files and ensure they don't fill up your disk.  Also, note that the `fieldNames` section above is optional, and in this case is restricting the number of fields that will be included in the JSON to two fields per log entry.  There are additional fields available, and if you omitted the `fieldNames` section you would end up with several additional fields per log entry by default.  For more info see the [Logstash Logback Encoder Docs](https://github.com/logstash/logstash-logback-encoder/blob/master/README.md#loggingevent-fields).
```
gguillotte: Can we briefly summarize what an appender is? For example:

> Logback writes logs using components called [**appenders**](http://logback.qos.ch/manual/appenders.html). Note that the above...

documentation/configuration.markdown
```diff
@@ -35,6 +35,8 @@ The default Logback configuration file is at `/etc/puppetserver/logback.xml` or
 
 Puppet Server relies on `logrotate` to manage the log file, and installs a configuration file at `/etc/logrotate.d/puppetserver` or `/etc/logrotate.d/pe-puppetserver`.
 
+For some tips on advanced logging configuration, including information about configuring your system to send log data to logstash, see the [Advanced Logging Configuration](./config_logging_advanced.html) documentation.
```
gguillotte: The logging doc doesn't really cover sending log data to logstash, it just refers to third-party tools that do it. And theoretically, third-party tools could send log data to anything, not just logstash.

Calling it "advanced logging configuration" implies there's a basic way to configure Puppet Server logging. We should probably describe (either in this section, in config_logging_advanced, or a separate doc) how to make basic changes, like switching logging modes from info to debug. I've got a branch [on my repo](https://github.com/gguillotte/puppet-server/blob/logback-docs/documentation/config_file_logbackxml.md) to try to tackle that.

Ziaunys:
This all looks very good. I think the configuration guide is detailed enough for people who aren't experienced with Java logging frameworks. I think users will like the fact that we've considered log rotation. The only small concern I have is including Logstash in the name of the appender. There's nothing that functionally ties it to Logstash specifically. Of course I don't mind being prescriptive about users sending the data to Logstash. Other than that I'm +1

danieldreier:
+1

@Ziaunys it looks to me like they're vendoring in the encoder that the logstash project produces, which I'm assuming is why it has that name.

@gguillotte has already covered all the issues I noticed, no need for duplicate feedback.

documentation/config_logging_advanced.md
cprice404: Added, though, since this is only on the `master` branch, I don't think I necessarily need to add the old path?  So I only added the `/etc/puppetlabs/puppetserver` one.

documentation/configuration.markdown
cprice404: I agree that we could use more coverage of other logback features, and would love to see your branch get merged in wherever is appropriate.  One minor thing to keep in mind is that much of this documentation is relevant to all of the other Clojure projects as well, but we could at least start by getting it into the Puppet Server docs.  Your branch looks great, let us know how we can help.

I will tweak the mention of logstash above to be a bit more general.

cprice404:
@Ziaunys I agree with your comment about using LOGSTASH as the appender name, and I think your comment also aligns with some of @gguillotte's comments.  I changed it to JSON.

cprice404:
Added a commit that attempts to address all of the PR feedback.

documentation/config_logging_advanced.md
```diff
@@ -0,0 +1,130 @@
...
+``` xml
+<root level="info">
+        <appender-ref ref="FILE"/>
+        <appender-ref ref="JSON"/>
+</root>
+```
+
+You could also comment out or remove the other `appender-ref`s, if you decided you *only* wanted to log the JSON format.
+
+#### Adding a JSON version of the Puppet Server HTTP Acces logs
```
mfournier: typo: *Access*

mfournier:
@cprice404 nice ! A couple of comments:

 - You copy-pasted the configs I pointed you to, but I realise they contain something which isn't the standard way of doing in the logstash world, and would probably confuse users. We label the timestamps `timegenerated` to be able to monitor the processing backlog by substracting them from a `timeprocessed` field we add downstream. By default, logstash (and kibana, the web gui) expect to find timestamps in a field called `@timestamp`. See comments inline.
 - indeed, the "logstash-logback-encoder" name is a bit confusing. From the project's github README: `Originally written to support output in logstash's JSON format, but has evolved into a highly-configurable, general-purpose, JSON logging mechanism`. Maybe would you want to add a short sentence making clear that using logstash is not mandatory ?
 - one nice thing about this encoder is it's ability to ship logs directly over the network. See https://github.com/logstash/logstash-logback-encoder/blob/master/README.md#tcp and https://github.com/logstash/logstash-logback-encoder/blob/master/README.md#udp. Maybe also mention this in the "Sending to Logstash" section ? In this case, I'd recommend warning that UDP can silently drop or truncate logs, and that TCP could stall the server in the case logstash inflicts back-pressure.

documentation/config_logging_advanced.md
```diff
@@ -0,0 +1,130 @@
...
+            <fileNamePattern>/var/log/puppetlabs/puppetserver/puppetserver.log.json.%d{yyyy-MM-dd}</fileNamePattern>
+            <maxHistory>5</maxHistory>
+        </rollingPolicy>
+
+        <encoder class="net.logstash.logback.encoder.LogstashEncoder">
+            <fieldNames>
+                <timestamp>timegenerated</timestamp>
+                <message>logmsg</message>
+            </fieldNames>
+        </encoder>
```
mfournier: IMHO this whole block can be skipped. The encoder will generate proper json-formatted events for logstash consumption by default. We added this because of custom processing we do in our logstash setup.

documentation/config_logging_advanced.md
```diff
@@ -0,0 +1,130 @@
...
+            <maxHistory>30</maxHistory>
+        </rollingPolicy>
+
+      <encoder class="net.logstash.logback.encoder.AccessEventCompositeJsonEncoder">
+          <providers>
+              <version/>
+              <pattern>
+                  <pattern>
+                    {
+                      "timegenerated":"%date{yyyy-MM-dd'T'HH:mm:ss.SSSXXX}",
```
mfournier: As mentioned, `s/timegenerated/@timestamp/`

cprice404:
@mfournier thank you so much for the feedback!  I'll add another commit attempting to address your input.

cprice404:
@mfournier @gguillotte @Ziaunys @danieldreier I added another commit that includes some tweaks based on @mfournier 's feedback.  Specifically I didn't know that you could just configure logback to export the data directly.

documentation/config_logging_advanced.md
```diff
@@ -0,0 +1,124 @@
...
+            <maxHistory>30</maxHistory>
+        </rollingPolicy>
+
+      <encoder class="net.logstash.logback.encoder.AccessEventCompositeJsonEncoder">
+          <providers>
+              <version/>
+              <pattern>
+                  <pattern>
+                    {
+                      "@timestamp":"%date{yyyy-MM-dd'T'HH:mm:ss.SSSXXX}",
```
Ziaunys: This field is generated by Logstash, but I wouldn't say it's required or standard. Newer versions of Kibana let you choose a time field. If you were interested in knowing when Logstash processed the event and when it was generated you would want two different time fields. However, it's really up to the user. I think if they're already Logstash users they'll understand this either way.

cprice404: Cool, thanks.

Ziaunys:
All I was going to add was to make sure and address the TCP/UDP transport issue, but it looks like you've done that. +1

cprice404:
@gguillotte are you satisfied with this for now?  We can always touch it up later if necessary?

gguillotte:
@cprice404 :+1:

PR #875, opened 2016-01-22
(SERVER-1056) Allow unauthenticated access to CRL

cprice404:


camlow325:
:+1: @nwolfe - look okay to you?

camlow325:
@cprice404 - hmm, maybe it would be good to target this at stable and merge up from there since the stable branch has tk-auth support?

nwolfe:
:+1: 

There's few more places where we've got this rule defined:
* https://github.com/puppetlabs/pe-puppetserver/blob/2016.1.x/config/conf.d/auth.conf#L33
* https://github.com/puppetlabs/puppetlabs-puppet_enterprise/blob/2016.1.x/manifests/master/tk_authz.pp#L50

cprice404:
@nwolfe I believe that I already have PRs for both of those?

camlow325:
I talked with @cprice404 about this IRL.  We agreed to let this change just happen on the master branch for now.  We can consider backporting it to the stable branch later if it becomes an issue for users.  So, merging...

PR #876, opened 2016-01-22
(SERVER-876) Revert code id changes

haus:
The previous PR worked fine for GET requests, where the environment would be part of the params map, but would not work on POST requests. It needs more work.


haus:
I think i have a fix, so maybe let's wait on this revert.

haus:
Yea I have a fix now. I'll have a PR open Monday.

PR #877, opened 2016-01-23
(SERVER-1094) consolidate shell execution libs

cprice404:
This PR migrates the CA autosign code over to use the new shellutils.  There is some cleanup and refactoring to the versioned code service and shell utils in the first two commits.  The third commit adds support for passing environment variables and a STDIN stream to the shellutils, and ports the CA code over to use it instead of `clojure.java.shell`.  It may be easiest to review the commits individually.


haus:
:+1:

PR #878, opened 2016-01-25
(SERVER-876) Only call wrap-params-for-jruby once

haus:
In a previous commit, wrap-params-for-jruby was called twice for catalog
requests. This meant that the body of the request would be consumed by
the first call (via body-for-jruby), and then the body would be gone for
the second call. This commit avoids that problem by refactoring
root-routes and the wrap-with-code-id to call wrap-params-for-jruby only
once in all routes. It also adds a test for the failing case where
catalog POST requests had no body.

### DIFF
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -20,9 +20,10 @@
   "Creates the routes to handle the master's '/v3' routes, which
    includes '/environments' and the non-CA indirected routes. The CA-related
    endpoints are handled separately by the CA service."
-  [request-handler
+  [request-handler-fn
    get-code-id-fn]
-  (let [request-handler-with-code-id (request-core/wrap-with-code-id request-handler get-code-id-fn)]
+  (let [request-handler (comp request-handler-fn request-core/wrap-params-for-jruby)
+        request-handler-with-code-id (request-core/wrap-with-code-id request-handler-fn get-code-id-fn)]
     (comidi/routes
      (comidi/GET ["/node/" [#".*" :rest]] request
                  (request-handler request))
--- a/test/unit/puppetlabs/services/master/master_core_test.clj
+++ b/test/unit/puppetlabs/services/master/master_core_test.clj
@@ -54,17 +54,19 @@
                 "is not overwritten, and simply passed through unmodified.")
     (let [handler     (fn ([req] {:request req}))
           app         (build-ring-handler handler "1.2.3")
-          resp        (app {:request-method :put
-                            :content-type   "application/octet-stream"
-                            :uri            "/v3/file_bucket_file/bar"})]
+          resp        (app (-> {:request-method :put
+                                :content-type "application/octet-stream"
+                                :uri "/v3/file_bucket_file/bar"}
+                               (mock/body "foo")))]
       (is (= "application/octet-stream"
              (get-in resp [:request :content-type])))
 
       (testing "Even if the client sends something insane, "
                "just pass it through and let the puppet code handle it."
-        (let [resp (app {:request-method :put
-                         :content-type   "something-crazy/for-content-type"
-                         :uri            "/v3/file_bucket_file/bar"})]
+        (let [resp (app (-> {:request-method :put
+                          :content-type "something-crazy/for-content-type"
+                          :uri "/v3/file_bucket_file/bar"}
+                            (mock/body "foo")))]
           (is (= "something-crazy/for-content-type"
                  (get-in resp [:request :content-type]))))))))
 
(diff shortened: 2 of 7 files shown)


camlow325:
Looks like multiple Travis failures occurred against this PR...

haus:
Yay!

haus:
Updated to fix test failures. Take that travis.

nathanielksmith:
:+1:

jpinsonault:
:+1:

jpinsonault:
just waiting on Mr. Travis

haus:
one travis cell failed fetching the facter gem for spec tests, so i rekicked it.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -20,9 +20,10 @@
   "Creates the routes to handle the master's '/v3' routes, which
    includes '/environments' and the non-CA indirected routes. The CA-related
    endpoints are handled separately by the CA service."
-  [request-handler
+  [request-handler-fn
    get-code-id-fn]
-  (let [request-handler-with-code-id (request-core/wrap-with-code-id request-handler get-code-id-fn)]
+  (let [request-handler (comp request-handler-fn request-core/wrap-params-for-jruby)
+        request-handler-with-code-id (request-core/wrap-with-code-id request-handler-fn get-code-id-fn)]
```
camlow325: @haus, @KevinCorcoran It feels a little strange to me to have `master_core` be coupled directly to `request-core` like this.  One other option that occurs to me is that we could just have this layer add something to the request map - maybe just an `add-code-id-to-params: {boolean}` or something - but then have the logic that actually calls back on the versioned code service to get the code id if necessary, etc.., could all just be done down inside of the request handler service's `jruby-request-handler`.

A potentially related question, does it matter when the service function to get the code-id is called relative to when the request reserves the jruby instance that it will be using to perform the catalog request?  If the code-id is added to the request as this PR currently has it, it seems like the Ruby code running the catalog request could be changed out between the time that the code-id is computed and the catalog compilation takes place - i.e., due to a file sync update being done in between.  Could that mean that the effective code-id related to the content on disk might be different than that supplied with the request down into the Ruby code?  In that case, doing the request for the code-id down in the `jruby-request-handler` could ensure that no file sync update would occur while the code-id is being computed - because the file sync update would be held off while the JRuby instance is held.

I don't have much of a handle on what code-id means at this point, so very possible that my comments here are all garbage and can be ignored.  :)

test/unit/puppetlabs/services/master/master_core_test.clj
```diff
@@ -24,11 +24,12 @@
         app         (build-ring-handler handler "1.2.3")
         request     (partial app-request app)]
     (is (= 200 (:status (request "/v3/environments"))))
+    (is (= 200 (:status (request "/v3/catalog/bar?environment=environment1234"))))
+    (is (= 200 (:status (request :post "/v3/catalog/bar?environment=environment1234"))))
```
camlow325: Should this one mock putting the environment parameter in the request body instead of just appending it as a query string at the end of the URL so that it would end up being run more like how a real catalog request would do it?

haus: Yes, it should.

src/clj/puppetlabs/services/master/master_core.clj
KevinCorcoran: That's a really good point, and raises an interesting question;

I think you are right, in the context of PE - it would be desirable to only make the function call to retrieve the `code_id` when a JRuby instance is held, thus ensuring that it matches what the catalog will be compiled from.  That's probably a reasonably easy change to make.

However, in OSS Puppet Server, computing `code_is` means shelling-out to an external process., and it seems like it'd be nice to avoid doing that while holding a JRuby instance.

Offhand, I am not sure how feasible it is to satisfy both of those requirements without some pretty major refactoring and possibly duplicating some code into PE.

I lean towards the PE requirement being more important, and so my suggestion would probably be that we:
1. move the call to the function which returns the `current-code-id` so that it happens when a JRuby instance is held.
2. add tests which cover the `code_id` changing due to a File Sync update happening in the background
3. acknowledge that this means we'll be shelling-out to `code-id-command` while holding a JRuby (and perhaps look into how much work it would be to avoid that, although I don't expect that to pan out).
4. not block this PR on those changes since CI is red and this should fix it.

Thoughts, @camlow325 @haus @cprice404 ?

camlow325: @KevinCorcoran That all sounds fine to me.

camlow325:
:+1: Tests pass locally so merging...

src/clj/puppetlabs/services/master/master_core.clj
cprice404: Unrelated to all of the previous comments: I think this code would be more readable if `wrap-with-code-id` didn't implicitly call `wrap-params-for-jruby` and instead we just composed it in right here...

PR #879, opened 2016-01-25
(SERVER-876) Compute code_id when holding a JRuby

KevinCorcoran:
This commit refactors the code which computes the code_id to only do so
while holding a JRuby instance.  This ensures that other code which may
result in the code_id changing (such as File Sync) and which requires
use of the JRubyPool's lock cannot be run in between the time when the
code_id is computed and when the catalog is compiled.

### DIFF
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -20,42 +20,39 @@
   "Creates the routes to handle the master's '/v3' routes, which
    includes '/environments' and the non-CA indirected routes. The CA-related
    endpoints are handled separately by the CA service."
-  [request-handler-fn
-   get-code-id-fn]
-  (let [request-handler (comp request-handler-fn request-core/wrap-params-for-jruby)
-        request-handler-with-code-id (request-core/wrap-with-code-id request-handler-fn get-code-id-fn)]
-    (comidi/routes
-     (comidi/GET ["/node/" [#".*" :rest]] request
-                 (request-handler request))
-     (comidi/GET ["/file_content/" [#".*" :rest]] request
-                 (request-handler request))
-     (comidi/GET ["/file_metadatas/" [#".*" :rest]] request
-                 (request-handler request))
-     (comidi/GET ["/file_metadata/" [#".*" :rest]] request
-                 (request-handler request))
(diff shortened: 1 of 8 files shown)


KevinCorcoran:
I'm a little uneasy with this since it doesn't include a test which specifically tests that `code_id` is computed while a JRuby is held.  But, @haus and I talked through a couple options on how to test that, and they all seemed difficult, time-consuming, and were sufficiently janky that we more-or-less decided to just punt on that.

Also, we'll need to voom this and update PE to account for these changes, but that change should be real easy.

haus:
I'm :+1:

camlow325:
Might be a bit Janky, but you could write a test that bootstraps just the RequestHandlerService and a custom VersionedCodeService and confirm that while inside of the body of the `current-code-id` function on the VersionedCodeService that the number of instances in the pool behind the RequestHandlerService is "total - 1"?

KevinCorcoran:
@camlow325 Yeah, I considered that ... thinking about that again, it sounds not completely terrible.  @haus wanna take a swing at writing that test?

camlow325:
LGTM - I like where this ended up so far.  :+1: other than the question about the additional test we could write.

haus:
@KevinCorcoran sure. love to. probably worth merging this in the meantime to unblock PE?

cprice404:
I haven't had time to finish reviewing this but the title makes me very nervous and I would like to hear the backstory before we consider merging this.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -20,42 +20,39 @@
...
+   (comidi/GET ["/resource_type/" [#".*" :rest]] request
+               (request-handler request))
+   (comidi/GET ["/resource_types/" [#".*" :rest]] request
+               (request-handler request))
+   (comidi/GET ["/environment/" [#".*" :rest]] request
+               (request-handler request))
+   (comidi/GET "/environments" request
+               (request-handler request))
+   (comidi/GET ["/status/" [#".*" :rest]] request
+               (request-handler request))))
```
cprice404: This is *really* hard to review... can we please try to be more diligent about stashing and doing whitespace changes in separate commits?

cprice404:
OK, I read the comments on the previous PR and I think I understand this all now.  Seems a little fragile but I don't have any better ideas.

PR #880, opened 2016-01-26
(SERVER-876) Add test for current-code-id function

haus:
A previous commit moved where the current-code-id function was called
during request handling into the jruby request handler. In order to
ensure that current-code-id is always calculated with the same jruby
that is handling the request, this commit adds a test that makes a
request and inspects the state of the jruby pool while code id is being
calculated during the request.

### DIFF
--- a/test/integration/puppetlabs/puppetserver/testutils.clj
+++ b/test/integration/puppetlabs/puppetserver/testutils.clj
@@ -6,6 +6,12 @@
   (:import (java.io File)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Schemas
+
+(def PuppetConfFiles
+  {schema/Str (schema/pred (some-fn string? #(instance? File %)))})
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Default settings
 
 (def conf-dir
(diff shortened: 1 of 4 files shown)


haus:
The need for a macro would go away here if I made a new test namespace for request_handler_core integration tests, but I don't see a precedent for that in the current codebase.

haus:
oops. i forgot to add a file also.

haus:
rebased!

fpringvaldsen:
:+1: as long as T-Pain is good with the changes

test/integration/puppetlabs/puppetserver/testutils.clj
```diff
@@ -101,7 +136,31 @@
...
+  of with-puppet-conf.
+
+  Accepts a map with the following characteristics:
+
+  * keys are simple filenames (no paths) that will be created in puppet's $confdir
+  * values are paths to a test-specific file that should be copied to puppet's confdir,
+    a la `$confdir/<key>`.
+
+  For each entry in the map, copies the files into $confdir, runs the test function,
+  and then deletes all of the files from $confdir.
```
jpinsonault: Maybe reword this a bit. I spent a few confusing seconds thinking it would run the body separately for each file in the map. Something like `Copies each of the files from the map into $confdir...`

haus: cool. i copied this directly from the test fixture function, but I can absolutely tweak the wording.

jpinsonault: Just notice that, I can blame chris instead

haus:
```For each entry in the map, copies the files into $confdir.
  Then it runs the body.
  Then after running the body it deletes all of the files from $confdir.```

is that ^^ better?

jpinsonault: Looks good

haus:
@jpinsonault updated.

jpinsonault:
This looks good to me, and it cleaned some stuff up, I'm :+1:

PR #881, opened 2016-01-27
(MAINT) wrap IOException when throwing in shellutils

cprice404:


haus:
:+1: waiting on a sanity check from @cprice404

PR #883, opened 2016-01-28
(SERVER-1115) Tolerate invalid-in-puppet-4 key in master webroute config

camlow325:
This commit re-adds the ability to tolerate the now obsolete
`invalid-in-puppet-4` key in the webroute config for the master-service.
The work for SERVER-1071 had removed this support, causing the server to
fail to start up if it still had this key in the config.


fpringvaldsen:
:+1:

PR #884, opened 2016-01-28
(SERVER-1094) Add logging re: JDK8 "stream closed" bug

cprice404:


test/unit/puppetlabs/puppetserver/shell_utils_test.clj
```diff
@@ -1,8 +1,9 @@
 (ns puppetlabs.puppetserver.shell-utils-test
   (:require [clojure.test :refer :all]
             [puppetlabs.puppetserver.shell-utils :as sh-utils]
+            [puppetlabs.trapperkeeper.testutils.logging :as logging]
```
nwolfe: leftovers

nwolfe:
:+1:

fpringvaldsen:
:+1:

src/java/com/puppetlabs/puppetserver/ShellUtils.java
```diff
@@ -66,9 +66,7 @@ private static ExecutionResult executeExecutor(CommandLine commandLine,
      */
     public static ExecutionResult executeCommand(String command)
             throws InterruptedException, IOException {
-        CommandLine commandLine = CommandLine.parse(command);
```
haus: This parse is needed for the jruby layer. It passes the command and arguments in as a single string. So rather than trying to parse them sanely in ruby, I chose to use the commons-exec facility to do so.

haus: Sorry, they aren't always a single string. They are either a command as a string, or an array where the command is the first element and the arguments are the rest of the array elements (could be one string or one array element per argument).

cprice404:
this has a bug, as @haus pointed out, and I'd rather rebase it on top of @nathanielksmith 's PR #885 .

PR #885, opened 2016-01-28
(SERVER-887) Implement get-code-content

nathanielksmith:
This implements the open source half of get-code-content. This work is very similar to the current-code-id work with the important caveat that it switches ShellUtils's ExecutionResult to using a ByteArrayInputStream internally for stdout instead of a string.

There is some minor refactoring of the current-code-id stuff to have it share error message code with get-code-content.

### DIFF
--- a/src/clj/puppetlabs/services/protocols/versioned_code.clj
+++ b/src/clj/puppetlabs/services/protocols/versioned_code.clj
@@ -5,4 +5,7 @@
 
   (current-code-id
    [this environment]
-   "Returns the current code id (representing the freshest code) for the given environment."))
+   "Returns the current code id (representing the freshest code) for the given environment.")
+  (get-code-content
+   [this environment code-id file-path]
+   "Returns the content of file-path in the given environment at the revision represented by code-id."))
--- a/src/clj/puppetlabs/services/versioned_code_service/versioned_code_core.clj
+++ b/src/clj/puppetlabs/services/versioned_code_service/versioned_code_core.clj
@@ -3,14 +3,38 @@
             [clojure.tools.logging :as log]
             [puppetlabs.puppetserver.shell-utils :as shell-utils]
             [clojure.string :as string])
-  (:import (java.io IOException)))
+  (:import (java.io IOException InputStream)
+           (org.apache.commons.io IOUtils)))
+
+(schema/defn ^:always-validate
+  success-with-stderr-msg :- schema/Str
+  [cmd :- schema/Str
+   stderr :- schema/Str]
+  (format "Error output generated while running '%s'. stderr: '%s'"
+          cmd stderr))
+
+(schema/defn ^:always-validate
+  nonzero-msg :- schema/Str
+  [cmd :- schema/Str
+   exit-code :- schema/Int
+   stdout :- schema/Str
+   stderr :- schema/Str]
+  (format (str "Non-zero exit code returned while running '%s'. "
+               "exit-code: '%d', stdout: '%s', stderr: '%s'")
+          cmd exit-code stdout stderr))
+
+(schema/defn ^:always-validate
+  execution-error-msg :- schema/Str
+  [cmd :- schema/Str
+   e :- Exception]
+  (format (str "Running script generated an error. "
+               "Command executed: '%s', error generated: '%s'")
+              cmd (.getMessage e)))
 
 (schema/defn ^:always-validate execute-code-id-script! :- (schema/maybe schema/Str)
   [code-id-script :- schema/Str
    environment :- schema/Str]
-  (let [error-msg #(log/errorf (str "Calculating code id generated an error. "
-                                    "Command executed: '%s', error generated: '%s'")
-                               code-id-script (.getMessage %1))]
+  (let [log-execution-error! #(log/error (execution-error-msg code-id-script %))]
     (try
       (let [{:keys [exit-code stderr stdout]} (shell-utils/execute-command
                                                code-id-script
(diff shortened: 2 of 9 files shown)


src/clj/puppetlabs/services/versioned_code_service/versioned_code_core.clj
```diff
@@ -27,20 +51,44 @@
...
+        (log-execution-error! e))
+      (catch IOException e
+        (log-execution-error! e))
+      (catch InterruptedException e
+        (log-execution-error! e)))))
+
+(schema/defn ^:always-validate
+  execute-code-content-script! :- InputStream
+  "Given a string path to an executable script and the environment, code-id, and
+  file-path of a desired revision of a file of code, returns that code as a
```
haus: "file of code" => "file"? and "returns that code" => "returns that file"?

src/clj/puppetlabs/services/versioned_code_service/versioned_code_service.clj
```diff
@@ -21,4 +21,12 @@
    [this environment]
    (if-let [code-id-script (get-in-config [:versioned-code :code-id-command])]
      (vc-core/execute-code-id-script! code-id-script environment)
-     nil)))
\ No newline at end of file
+     nil))
+
+  (get-code-content
+   [this environment code-id file-path]
+   (if-let [code-content-script (get-in-config [:versioned-code :code-content-command])]
```
haus: Would it be worth logging a message at init/startup as was done with code-id-command if code-content-command isn't set?

src/java/com/puppetlabs/puppetserver/ShellUtils.java
```diff
@@ -43,13 +50,14 @@ private static ExecutionResult executeExecutor(CommandLine commandLine,
...
 
+        ByteArrayInputStream stdoutInputStream = new ByteArrayInputStream(stdoutOutputStream.toByteArray());
         String stdErr = errStream.toString();
 
         if ( ! stdErr.isEmpty() ) {
             log.warn("Executed an external process which logged to STDERR: " + stdErr);
         }
 
-        return new ExecutionResult(outStream.toString(), stdErr, exitCode);
+        return new ExecutionResult(stdoutInputStream, stdErr, exitCode);
```
haus: I'm not sure what the standard practice is here, but you could inline `new ByteArrayInputStream(stdoutOutputStream.toByteArray())` here and not have the assignment above, as all it is used for is the return.

haus:
This looks awesome.

src/java/com/puppetlabs/puppetserver/ShellUtils.java
nathanielksmith: I think this is clearer, that line would get pretty long if I did that

src/clj/puppetlabs/services/versioned_code_service/versioned_code_service.clj
nathanielksmith: yup, I had that but lost it in rebase thrashing, will add back

src/clj/puppetlabs/services/versioned_code_service/versioned_code_core.clj
nathanielksmith: updated

src/java/com/puppetlabs/puppetserver/ShellUtils.java
haus: that's true

haus:
This looks totally reasonable to me. I'm :+1:.

I do think the commit message could probably use a little more detail before merging.

src/java/com/puppetlabs/puppetserver/ExecutionResult.java
```diff
@@ -15,7 +19,11 @@ public int getExitCode() {
         return exitCode;
     }
 
-    public String getOutput() {
+    public String getOutput() throws IOException {
+        return IOUtils.toString(output, "UTF-8");
```
cprice404: Good man, always specify the encoding :)

nathanielksmith:
Updated the commit message to have words.

src/java/com/puppetlabs/puppetserver/ExecutionResult.java
nathanielksmith: i live in fear of mis interpreted bytes

cprice404:
:see_no_evil:

src/clj/puppetlabs/services/protocols/versioned_code.clj
```diff
@@ -5,4 +5,7 @@
 
   (current-code-id
    [this environment]
-   "Returns the current code id (representing the freshest code) for the given environment."))
+   "Returns the current code id (representing the freshest code) for the given environment.")
+  (get-code-content
+   [this environment code-id file-path]
```
jpinsonault: Seems like `environment` will be an optional parameter right? Should we have two versions of the service function, one with it and one without where it defaults to `nil`?

haus: when would environment be optional?

jpinsonault: When they're getting a file out of code-dir that's not in an environment. I think I might be :confused: about something

haus: So there are 2 content endpoints. static-file-content will use this endpoint and require an environment. the regular content endpoint won't need an environment and will just hit the current file content endpoint afaik. I'm probably messing the actual api endpoint names up here, but puppet will be redirecting based on whether or not the desired file is inside code dir or not.

jpinsonault: Ah, ok awesome, makes sense

PR #886, opened 2016-01-28
(PE-13485) Implement a consistent request logging format

npwalker:
Prior to this commit puppetserver used a different request
logging format than puppetdb and console-services.

After this commit the request logging will be of the suggested
formatting for all trapper-keeper projects.


cprice404:
FWIW there is probably another one of these in pe-puppet-server-extensions or pe-puppetserver

cprice404:
@npwalker thanks!  I think we will also want to do this here:

https://github.com/puppetlabs/pe-puppetserver/tree/2016.1.x/config

npwalker:
https://github.com/puppetlabs/pe-puppetserver/pull/60

PR #887, opened 2016-01-29
(SERVER-880) Add static_file_content endpoint

fpringvaldsen:
Add a new /static_file_content web endpoint that exercises the
`get-code-content` service function from the VersionedCodeService.

This is based on top of https://github.com/puppetlabs/puppet-server/pull/885, which should be merged first. The relevant commit is 2bba3d2

### DIFF
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -81,7 +81,9 @@
    ring/wrap-params))
 
 (defn static-file-content-request-handler
-  "Build the main request handler for the /static_file_content endpoint"
+  "Returns a function which is the main request handler for the
+  /static_file_content endpoint, utilizing the provided implementation of
+  `get-code-content`"
   [get-code-content]
   (fn [req]
     (let [environment (get-in req [:params "environment"])
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -9,7 +9,8 @@
             [ring.util.response :as rr]
             [schema.core :as schema]
             [puppetlabs.services.protocols.jruby-puppet :as jruby-protocol]
-            [puppetlabs.puppetserver.jruby-request :as jruby-request]))
+            [puppetlabs.puppetserver.jruby-request :as jruby-request]
+            [puppetlabs.services.request-handler.request-handler-core :as request-core]))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Constants
@@ -86,8 +87,9 @@
   "Creates the routes to handle the master's '/v3' routes, which
    includes '/environments' and the non-CA indirected routes. The CA-related
    endpoints are handled separately by the CA service."
-  [request-handler jruby-service]
-  (let [environment-class-handler (environment-class-handler jruby-service)]
+  [request-handler jruby-service get-code-content-fn]
+  (let [environment-class-handler (environment-class-handler jruby-service)
+        static-file-content-handler (request-core/static-file-content-request-handler get-code-content-fn)]
     (comidi/routes
      (comidi/GET ["/node/" [#".*" :rest]] request
                  (request-handler request))
@@ -122,7 +124,9 @@
                  (request-handler request))
 
      (comidi/GET ["/environment_classes" [#".*" :rest]] request
-                 (environment-class-handler request)))))
+                 (environment-class-handler request))
+     (comidi/GET ["/static_file_content/" [#".*" :rest]] request
+                 (static-file-content-handler request)))))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Lifecycle Helper Functions
@@ -173,10 +177,12 @@
 
 (defn root-routes
   "Creates all of the web routes for the master."
-  [request-handler jruby-service]
+  [request-handler jruby-service get-code-content-fn]
   (comidi/routes
     (comidi/context "/v3"
-                    (v3-routes request-handler jruby-service))))
+                    (v3-routes request-handler
+                               jruby-service
+                               get-code-content-fn))))
 
 (schema/defn ^:always-validate
   wrap-middleware :- IFn
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -80,6 +80,17 @@
    jruby-request/wrap-with-error-handling
    ring/wrap-params))
 
+(defn static-file-content-request-handler
+  "Build the main request handler for the /static_file_content endpoint"
+  [get-code-content]
+  (fn [req]
+    (let [environment (get-in req [:params "environment"])
+          code-id (get-in req [:params "code-id"])
+          file-path (get-in req [:params :rest])
+          content-stream (get-code-content environment code-id file-path)]
+      {:status 200
+       :body content-stream})))
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Routing
 
@@ -89,7 +100,7 @@
    endpoints are handled separately by the CA service."
   [request-handler jruby-service get-code-content-fn]
   (let [environment-class-handler (environment-class-handler jruby-service)
-        static-file-content-handler (request-core/static-file-content-request-handler get-code-content-fn)]
+        static-file-content-handler (static-file-content-request-handler get-code-content-fn)]
     (comidi/routes
      (comidi/GET ["/node/" [#".*" :rest]] request
                  (request-handler request))
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -85,11 +85,15 @@
   [get-code-content]
   (fn [req]
     (let [environment (get-in req [:params "environment"])
-          code-id (get-in req [:params "code-id"])
-          file-path (get-in req [:params :rest])
-          content-stream (get-code-content environment code-id file-path)]
-      {:status 200
-       :body content-stream})))
+          code-id (get-in req [:params "code_id"])
+          file-path (get-in req [:params :rest])]
+      (if (or (nil? environment)
+              (nil? code-id)
+              (or (nil? file-path) (= "" file-path)))
+        {:status 400
+         :body "Error: A /static_file_content request requires an environment, a code-id, and a file-path"}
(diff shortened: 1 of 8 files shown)


haus:
@fpringvaldsen for the wrap-with-code-id work, there was some concern (https://github.com/puppetlabs/puppet-server/pull/878#discussion_r50751427) about tightly coupling master_core and request_handler_core, and it looks like this has the same coupling.

KevinCorcoran:
#885 has ben merged, so this is ready for a rebase.

fpringvaldsen:
@haus I can look into decoupling that. I wasn't sure of a good way to do it since the `static_file_content` endpoint isn't a JRuby request and the `wrap-with-code-id` work was done under that assumption, but I should be able to find a way to do it.

haus:
@fpringvaldsen yea I didn't see any obvious way of doing that either. possibly @camlow325 's PR might be a path forward #895 as it introduces a non-jruby based endpoint.

fpringvaldsen:
@haus Hmm, might be worth rebasing on top of that PR then

haus:
@fpringvaldsen no idea if it's worth holding this up for that or circling back around once that work is in

fpringvaldsen:
@KevinCorcoran Rebased!

camlow325:
@fpringvaldsen I don't have any context on this PR but it would be good if you could at least review the portion of #895 that covers how the new "environment_classes" endpoint would be plumbed through the comidi route / handler, just to see that it looks copacetic with what you would be doing for the new `static_file_content` endpoint.

fpringvaldsen:
This PR has been rebased. I also added a new commit that decouples request-core from master-core.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -79,6 +80,17 @@
...
 
+(defn static-file-content-request-handler
+  "Build the main request handler for the /static_file_content endpoint"
+  [get-code-content]
+  (fn [req]
+    (let [environment (get-in req [:params "environment"])
+          code-id (get-in req [:params "code-id"])
+          file-path (get-in req [:params :rest])
+          content-stream (get-code-content environment code-id file-path)]
+      {:status 200
```
haus: What happens if the file/environment/code-id doesn't exist?

src/clj/puppetlabs/services/master/master_service.clj
```diff
@@ -45,7 +46,9 @@
            path (core/get-master-mount ::master-service route-config)
            ring-handler (when path
                           (core/get-wrapped-handler
-                            (-> (core/root-routes handle-request jruby-service)
+                            (-> (core/root-routes handle-request
+                                                  jruby-service
+                                                  get-code-content)
```
KevinCorcoran: @haus this is what we were expecting to happen, the last time we were working on this code, right?

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -79,6 +80,17 @@
    jruby-request/wrap-with-error-handling
    ring/wrap-params))
 
+(defn static-file-content-request-handler
+  "Build the main request handler for the /static_file_content endpoint"
```
KevinCorcoran: Nitpick about docstring: "build" is maybe a little unclear, instead you could say "Returns a function which is the main request handler ..."

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -79,6 +80,17 @@
...
+  [get-code-content]
+  (fn [req]
+    (let [environment (get-in req [:params "environment"])
+          code-id (get-in req [:params "code-id"])
+          file-path (get-in req [:params :rest])
```
KevinCorcoran: I think there's a way to specify the comidi route below that will allow us to refer to this as `[:params :file-path]` or something like that.  Using `:rest` here seems less than awesome.

src/clj/puppetlabs/services/master/master_service.clj
haus: I think so.

src/clj/puppetlabs/services/master/master_core.clj
KevinCorcoran: should probably be a 404, although the current implementation does not seem to account for that.

src/clj/puppetlabs/services/master/master_core.clj
KevinCorcoran: Oh, I see what's going on now.  I thought that `file-path` was a query parameter like the others, which is incorrect, so this is probably good as-is.

test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
```diff
@@ -71,3 +72,30 @@
...
+(deftest ^:integration static-file-content-endpoint-test
+  (logging/with-test-logging
+    (testing "the /static_file_content endpoint successfully streams file content"
+      (bootstrap/with-puppetserver-running
+        app
+        {:jruby-puppet
+         {:max-active-instances num-jrubies}
+         :versioned-code
+         {:code-content-command (script-path "echo")}}
+        (let [response (http-client/get "https://localhost:8140/puppet/v3/static_file_content/foo/bar/?code-id=foobar&environment=test"
```
KevinCorcoran: @joshcooper Could you have a quick peek at this URL - does this match your expectation of what the agent will be using?  Primarily I'm wondering about:
* specifying the file path as the end of the path (i.e. `/static_file_content/my/file/path/goes/here`)
* I am guessing that `code-id-` should actually be `code_id` here, right?

test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
```diff
@@ -71,3 +72,30 @@
...
+      (bootstrap/with-puppetserver-running
+        app
+        {:jruby-puppet
+         {:max-active-instances num-jrubies}
+         :versioned-code
+         {:code-content-command nil}}
+        (let [response (http-client/get "https://localhost:8140/puppet/v3/static_file_content/foo/bar/?code-id=foobar&environment=test"
+                                        (assoc testutils/ssl-request-options
+                                          :as :stream))]
+          (is (= 500 (:status response))))))))
```
KevinCorcoran: What's the response body in this case?  It would be nice if it was something like "code-content-command is not set".  Also, I almost hesitate to say this, but what do we think about a 500 in this case?  Maybe a 409 would be more appropriate?

fpringvaldsen: Basically the response body will just be the error message thrown by the `get-code-content` fn, so with the current implementation, it will return

```
Internal Server Error: java.lang.IllegalStateException: Cannot fetch code content without :code-content-command being set.
```

src/clj/puppetlabs/services/master/master_core.clj
fpringvaldsen: I wasn't entirely sure what to do in this case. Basically, our `get-code-content` function will rely on the execution of the set script. According to the docstring:

```
Does no exception catching: Being unable to compute code content is considered a fatal error and should be handled by calling code appropriately.
```

Judging from the code, it looks like we're expecting an exception to be thrown by `get-code-content` if the file/environment/code-id doesn't exist (we catch certain exceptions and rethrow as `IllegalStateException`s, and throw an `IllegalStateException` if the script has a non-zero exit code) so that will manifest as an HTTP 500 error and include the Exception's message.

I think we could feasibly change that to return a 404 error, but we'd have to do some exception catching in this method. That wouldn't be too bad, but `get-code-content` will throw an `IllegalArgumentException` in all cases, and there doesn't seem to really be a way to differentiate between when a script failed due to an unexpected error and when a failure occurred because the file/environment/code-id doesn't exist.

KevinCorcoran: How about we update `get-code-content` to throw a slingshot exception with a specific `:type`, and then we catch that here and return a 404 ?

fpringvaldsen: @KevinCorcoran Hmm, that could work. Then we can only catch the case in which the script returns a non-zero exit code (assuming that's the case in which file content can't be found).

KevinCorcoran: On further thought ...

Maybe this code should simply be responsible for checking that environment and code ID are not `nil`, and that `file-path` is actually a file which exists on disk?  Although, for that last case, this code can't easily know that, since it's actually inside an environment.  This is more complex than I had originally realized.

test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
fpringvaldsen: I based the URL structure off of [PUP-5114](https://tickets.puppetlabs.com/browse/PUP-5114). Looks like that should definitely be `code_id` rather than `code-id`. @joshcooper Is the URL structure in PUP-5114 still what you all are looking for?

src/clj/puppetlabs/services/master/master_core.clj
fpringvaldsen: Yup. Not only that, but even if we could check the file inside the environment, it's entirely possible that the file exists at the revision indicated by `code-id` but was deleted in a later revision, and thus does not exist in the most recent code.

test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
joshcooper: @fpringvaldsen :+1: on `code_id`

Also the request path looks correct, though to be a bit more specific, I'm assuming the `path` will be rooted in a specific per-environment directory and will be taken directly from the `content_uri` parameter described in https://docs.google.com/document/d/1bZU1WJAvVQfyoBHf-5kXGYivg7r_ul5o9boimCdpYuM/edit#heading=h.lhexacaboszr. So I'm thinking it'd be something like:

https://localhost:8140/puppet/v3/static_file_content/modules/ntp/files/ntpd.conf`

I also believe the agent will send the `environment` query parameter (duplicating information in `code_id`), because it's hard to get the agent to not send it.

fpringvaldsen: @joshcooper Yup, that assumption is correct. The request handler will treat the path after `/static_file_content` as a relative path to a file in the specified environment.

fpringvaldsen:
@KevinCorcoran I've added a new commit to address what we talked about. I've also rebased.

test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
KevinCorcoran: Okay, that's reasonable.  I think it'd be good for that to be tested here as well.  And while, you're at it, what do you think about changing that error message in `get-code-content` to something a little better like:

"Cannot retrieve code content because the "versioned-code.code-content-command" setting is not present in configuration."

KevinCorcoran:
:+1: I left one more nitpick about improving an error message and associated test in a comment above (currently hidden as an 'outdated diff') but other than that, I'm good with this and think this is merge-able as-id.

@haus if you could finish this off whenever you have a chance, that'd be swell.

haus:
Also :+1:.

fpringvaldsen:
@KevinCorcoran @haus I've pushed up a new commit to fix that error message

PR #889, opened 2016-01-29
(docs) Escape Liquid-ish tags.

gguillotte:


documentation/config_logging_advanced.md
```diff
@@ -105,6 +106,7 @@ To add JSON logging for Puppet Server's HTTP requests, modify the `request-loggi
             </providers>
         </encoder>
     </appender>
+{% endraw %}
 ```
```
cprice404: There are a few more of the ``` sections below

cprice404: it looks to me like there are 4 total.  I can add a commit and copypaste what you've done if you like.  At least, I'm assuming that it needs to happen for the other sections as well... but maybe not?

gguillotte: We only have to do it around the example that includes `{%` in the code example, and that's the only one I see.

PR #890, opened 2016-01-29
(SERVER-1117) Update class-info tests per PUP-5713 changes

camlow325:
This commit contains two changes:

1) Update the class-info tests to expect a 'default_source' parameter to
   be present in the return from a call to the
   `ClassInformationService.classes_per_environment` method per expected
   changes made in PUP-5713.

2) Update the code in .getClassInfoForEnvironment from "master.rb" to
   ensure that the value of a candidate Puppet::Node::Environment's
   "name" is coerced to a String before comparing to the supplied String
   type environment argument.  Previously, the "name" was not coerced
   and so was being compared as a Symbol rather than a String, leading
   the comparison to always fail and for the target environment to not
   be found.

### DIFF
--- a/src/ruby/puppet-server-lib/puppet/server/master.rb
+++ b/src/ruby/puppet-server-lib/puppet/server/master.rb
@@ -97,7 +97,7 @@ class Puppet::Server::Master
     # and so would not be subject to any potentially stale settings data being
     # used to enumerate manifests.
     environment = @env_loader.list.find do |env_from_loader|
-      env_from_loader.name == env
+      env_from_loader.name.to_s == env
     end
     unless environment.nil?
       environments = Hash[env, self.class.getManifests(environment)]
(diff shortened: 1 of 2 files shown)


src/ruby/puppet-server-lib/puppet/server/master.rb
```diff
@@ -97,7 +97,7 @@ def getClassInfoForEnvironment(env)
     # and so would not be subject to any potentially stale settings data being
     # used to enumerate manifests.
     environment = @env_loader.list.find do |env_from_loader|
-      env_from_loader.name == env
+      env_from_loader.name.to_s == env
```
camlow325: I can't really explain why this change is needed now but these tests were able to pass without it previously.  This allows the comparison to be made as a String to a String rather than as a Symbol to a String.  Locally, I haven't seen Symbol to String comparisons succeed in other contexts and it doesn't appear that the type of `Puppet::Node::Environment.name` has changed from String to Symbol, at least recently.

PR #891, opened 2016-01-29
(SERVER-1094) log message about stdin jdk8 bug

cprice404:
This commit adds some extra error handling and logging around an edge case that we hit with JDK8 and commons-exec, where passing data on STDIN to a process that doesn't consume it can cause commons-exec to throw a not-very-helpful exception.  For more info see SERVER-1094 and linked tickets.


haus:
:+1: looks sane. where did travis go?

fpringvaldsen:
:+1:

PR #894, opened 2016-02-01
(MAINT) Bump jruby-pool-code-id test timeout

camlow325:
This commit bumps up the maximum amount of time that the
`test-jruby-pool-not-full-during-code-id-generation` test will wait for
a catalog request to complete from 5 seconds to 10 seconds.  This test
frequently failed for me locally when run in isolation from the command
line because the first catalog request takes right around 5 seconds,
although subsequent ones were much faster (~100 milliseconds).

### DIFF
--- a/test/unit/puppetlabs/services/request_handler/request_handler_core_test.clj
+++ b/test/unit/puppetlabs/services/request_handler/request_handler_core_test.clj
@@ -311,9 +311,13 @@
         (jruby-bootstrap/with-puppetserver-running-with-services
          app services {:jruby-puppet {:max-active-instances 1}}
          (jruby-testutils/wait-for-jrubies app)
-         (let [jruby-service (tk-app/get-service app :JRubyPuppetService)]
-           (future (testutils/get-catalog))
-           (is (deref first-promise 5000 false))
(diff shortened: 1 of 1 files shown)


test/unit/puppetlabs/services/request_handler/request_handler_core_test.clj
```diff
@@ -311,9 +311,13 @@
...
          (jruby-testutils/wait-for-jrubies app)
-         (let [jruby-service (tk-app/get-service app :JRubyPuppetService)]
-           (future (testutils/get-catalog))
-           (is (deref first-promise 5000 false))
+         (let [in-catalog-request-future (promise)
+               jruby-service (tk-app/get-service app :JRubyPuppetService)]
+           (future
+            (deliver in-catalog-request-future true)
+            (testutils/get-catalog))
+           (deref in-catalog-request-future)
```
camlow325: I added in this bit just so that the timer doesn't include the overhead of launching/waiting for the future's thread to start.

haus:
:+1:

fpringvaldsen:
:+1:

PR #895, opened 2016-02-01
(SERVER-1110) Introduce environment_classes endpoint

camlow325:
This PR introduces a new HTTP endpoint, environment_classes, which
returns information about the available classes on a per environment
basis, including the names of the classes, associated parameter names
and types, and the default values for any parameters, where applicable.

This PR also moves some of the JRuby request handling helper
functions from the request-handler-core namespace into a new namespace,
jruby-request, so that the backend implementation of the new
environment_classes endpoint can reuse some of those same functions
without having to be processed by the request-handler-service.

This PR also removes some code that had been added in previous tickets --
SERVER-815 and PE-11124 -- in anticipation of what the environment_classes
endpoint would need.  That code no longer seemed to be appropriate, so it
seemed best to just remove it.


src/clj/puppetlabs/puppetserver/jruby_request.clj
```diff
@@ -0,0 +1,71 @@
+(ns puppetlabs.puppetserver.jruby-request
```
camlow325: I should have done this in a separate commit for better reviewability, sorry, but all of the functionality here was just moved over from the `request-handler-core` namespace.  I did that so that the environment classes handling in `master-core` could use this same functionality, keeping the basic request handling functionality simliar to that of other Puppet Ruby-bound HTTP requests.

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
```diff
@@ -259,73 +251,13 @@
   "Make the request mutable.  This is required by the ruby layer."
   (HashMap. request))
 
-(defn bad-request?
```
camlow325: All of the deleted functions here moved over to the new `jruby-request` namespace.  Other changes in this namespace are mostly just about referencing the functions from their new location.

src/java/com/puppetlabs/puppetserver/JRubyPuppet.java
```diff
@@ -15,7 +15,6 @@
  *
  */
 public interface JRubyPuppet {
-    Map<String, Map> getClassInfoForAllEnvironments();
```
camlow325: We don't have plans for a "get all environments" version of the "environment_classes" endpoint at this point.  I thought it would be better to just purge this method - and the corresponding implementation in `master.rb`.

src/ruby/puppet-server-lib/puppet/server/master.rb
```diff
@@ -126,14 +89,17 @@ def terminate
...
-        manifests = Dir.glob(File.join(env.manifest, '**/*.pp')).sort
-      else
-        manifests = [env.manifest]
+    manifests =
+      case
+        when env.manifest == Puppet::Node::Environment::NO_MANIFEST
+          []
+        when File.directory?(env.manifest)
+          Dir.glob(File.join(env.manifest, '**/*.pp'))
+        when File.exists?(env.manifest)
```
camlow325: In writing a new unit test for the `class-info-test` namespace, "non existent manifest dir for environment", I noticed that if the manifest directory setting pointed to a directory that doesn't actually exist on disk that an empty entry for it - as a "file" - would end up in the payload returned from a call to `getClassInfoForEnvironment`.  This condition eliminates that erroneous element.

src/ruby/puppet-server-lib/puppet/server/master.rb
```diff
@@ -60,45 +60,8 @@ def handleRequest(request)
...
-  end
-
   def getClassInfoForEnvironment(env)
-    # The clear_environment_settings(env) ensures that the calls which happen
-    # later on to enumerate all of the manifests for each environment are
-    # using the latest environment.conf settings for each environment.  Without
-    # this call, cached (and, therefore, possibly stale) environment.conf
-    # settings would be used.  We want to return the latest data for any calls
-    # made to this method.
-    Puppet.settings.clear_environment_settings(env)
```
camlow325: This call was in place to ensure that the latest settings from the environment's "environment.conf" file would be sourced before trying to parse class information.  This call is not necessary so long as the environment is marked as "expired" prior to this call being made.  As we now expect `mark-environments-expired!` calls to be part of the workflow for refreshing the cache, it seemed better to eliminate this call from the code - especially since it is an internal detail from Ruby Puppet which could change out from underneath Puppet Server Ruby code, potentially without warning.

test/integration/puppetlabs/services/jruby/class_info_test.clj
```diff
@@ -167,8 +163,7 @@
                                                [env-2-dir-and-manifests
                                                 env-1-mod-1-dir-and-manifests
                                                 env-1-mod-2-dir-and-manifests])}]
-              (testing "all environments"
-                (is (= expected-envs-info (get-class-info))))
+              (puppet-env/mark-all-environments-expired! env-registry)
```
camlow325: A few of these calls were added to these tests in order to compensate for the `Puppet.settings.clear_environment_settings` call being removed from master.rb's `getClassInfoForEnvironment` method.  See the comment in that method for more context.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -52,7 +105,10 @@
    (comidi/GET "/environments" request
                (request-handler request))
    (comidi/GET ["/status/" [#".*" :rest]] request
-               (request-handler request))))
+               (request-handler request))
+
+   (comidi/GET ["/environment_classes" [#".*" :rest]] request
+               ((environment-class-handler jruby-service) request))))
```
camlow325: Where the functionality for the "environment_classes" endpoint does use a JRuby, another option for this implementation would have been to somehow route it through the `request-handler-service` as is done for the other endpoints under "/puppet/v3".  This endpoint didn't seem to fit that model well, though, in that there's a lot more Clojure-specific functionality planned for it - e.g., use of an internal Clojure/process-wide cache.  Also, at some point, the environment-classes endpoint may not require the use of a JRuby at all if we get to an implementation where the cache is populated asynchronously from an incoming web service request.  The approach in this PR bypasses the `request-handler-service`, while still using the `jruby-puppet-pooled-service` to interact with a JRuby as needed.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_service.clj
```diff
@@ -27,7 +28,9 @@
           master-route-config (master-core/get-master-route-config
                                 master-ns
                                 config)
-          master-route-handler (-> (master-core/root-routes handle-request)
+          jruby-service (tk-services/get-service this :JRubyPuppetService)
+          master-route-handler (-> (master-core/root-routes handle-request
+                                                            jruby-service)
```
cprice404: Yuck :(  The legacy routes never need to use the jruby for anything, right?  Not sure I can think of a better way to do this, nor whether or not it's worth putting much thought into, but it does make me a little :cry:

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -14,13 +18,62 @@
   "v3")
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Private
+
+(defn class-info-from-jruby->class-info-for-json
+  "Convert a class info map received from the jruby service into an
+  appropriate map for use in serializing an environment_classes response to
+  JSON"
+  [info-from-jruby environment]
```
cprice404: Would it be difficult to schematize this?

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -14,13 +18,62 @@
...
+  (->> info-from-jruby
+       (map #(hash-map :path (key %) :classes (val %)))
```
cprice404: Just curious why `hash-map` here instead of a literal; I guess it's just easier to write an anonymous fn and have it work with the threading macro?

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -14,13 +18,62 @@
...
+       (sort-by :path)
+       (into [])
```
cprice404: Is this line just to coerce to a vector?  If so, any reason to prefer `into` to `vec`?

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -14,13 +18,62 @@
...
+       (hash-map :name environment :files)))
+
+(defn get-environment-class-info
```
cprice404: I'd probably name this `get-environment-class-info-fn` or `environment-class-info-fn` to make it more obvious that it returns a fn.  And/or add schema.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -14,13 +18,62 @@
...
+  "Middleware function for constructing a Ring response from an incoming
+  request for environment_classes information"
+  [jruby-service]
+  (fn [request]
+    (if-let [environment (jruby-request/get-environment-from-request request)]
```
cprice404: so we are sticking with a query param for passing in the environment for now?  Makes sense in terms of consistency with the rest of our junk, I guess.

src/clj/puppetlabs/puppetserver/jruby_request.clj
```diff
@@ -0,0 +1,71 @@
...
+
+(defn wrap-with-jruby-instance
+  "Middleware fn that borrows a jruby instance from the `jruby-service` and makes
+  it available in the request as `:jruby-instance`"
+  [f jruby-service]
+  (fn [request]
+    (jruby/with-jruby-puppet
+     jruby-instance
+     jruby-service
+     {:request (dissoc request :ssl-client-cert)}
```
cprice404: Did this fn already exist?

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -14,13 +18,62 @@
...
+              (class-info-from-jruby->class-info-for-json environment)
+              (cheshire/generate-string)
+              (rr/response)
+              (rr/content-type "application/json"))
+          (rr/not-found (str "Could not find environment '" environment "'")))
+        (jruby-request/throw-bad-request!
+         (str
+          "The environment must be purely alphanumeric, not '"
+          environment
+          "'")))
```
cprice404: There are a lot of nested conditions here... I think it would be worth extracting at least one helper function to make this more readable.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -14,13 +18,62 @@
...
+        (if-let [class-info
+                 (jruby-protocol/get-environment-class-info jruby-service
+                                                            (:jruby-instance
+                                                             request)
+                                                            environment)]
+          (-> class-info
+              (class-info-from-jruby->class-info-for-json environment)
+              (cheshire/generate-string)
+              (rr/response)
+              (rr/content-type "application/json"))
```
cprice404: It seems like we have the same logic that these last three lines are implementing in use in some other places in our code... maybe using `ring-api`?  but i may be getting our stuff confused with other git repos.  In any case those three lines seem general enough to extract a fn for if we don't already have one... `json-response` or something.

src/clj/puppetlabs/services/master/master_core.clj
cprice404: Yeah, `request-handler-service` needs to die, it should not have any logic beyond "I'm going to hand this entire request off blindly to Puppet", so what you've done here is better.

Except... it seems like you could create the handler fn in a `let` block at the top of this routes fn.  This implementation seems like it will keep creating a new handler fn for every request.

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
```diff
@@ -259,73 +251,13 @@
...
-  "Middleware fn that borrows a jruby instance from the `jruby-service` and makes
-  it available in the request as `:jruby-instance`"
-  [f jruby-service]
-  (fn [request]
-    (jruby/with-jruby-puppet
-      jruby-instance
-      jruby-service
-      {:request (dissoc request :ssl-client-cert)}
-      
-      (f (assoc request :jruby-instance jruby-instance)))))
```
cprice404: Ah, yep, this fn did indeed exist before.  It seems like kind of a scary function to me, but, I guess that's not relevant to this PR.

src/java/com/puppetlabs/puppetserver/JRubyPuppet.java
cprice404: agree

src/ruby/puppet-server-lib/puppet/server/master.rb
cprice404: definitely agree

src/ruby/puppet-server-lib/puppet/server/master.rb
cprice404: Should we file a PUP ticket for that?

test/unit/puppetlabs/puppetserver/environment_cache_test.clj
```diff
@@ -1,86 +0,0 @@
...
-             (get-all-environments-info @cache)))
-
-      (swap! cache remove-environment-info "test")
-      (is (nil? (get-environment-info @cache "test")))
-      (is (= (get prod-data "production")
-             (get-environment-info @cache "production")))
-
-      (swap! cache reset-cache)
-      (is (nil? (get-environment-info @cache "production")))
-      (is (= {} (get-all-environments-info @cache))))))
```
cprice404: Man I didn't realize how far we'd gotten down this path back then...

test/unit/puppetlabs/services/master/master_core_test.clj
```diff
@@ -51,11 +57,86 @@
...
+                                                  {"name" "twoparam",
+                                                   "type" "Integer",
+                                                   "default_literal" "3",
+                                                   "default_source" "3"}]},
+                                       {
+                                        "name" "twoclass"
+                                        "params" []}]},
+                           {"path" "/two/file"
+                            "classes" []}]}
+                 body-from-json)
```
cprice404: What value do you think these tests add beyond the integration tests?  I know that I've been advocating for having more unit tests, but there's so much setup logic here and so much mocking that I'm not sure it is really worth it.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -14,13 +18,62 @@
...
+(defn class-info-from-jruby->class-info-for-json
+  "Convert a class info map received from the jruby service into an
+  appropriate map for use in serializing an environment_classes response to
+  JSON"
+  [info-from-jruby environment]
+  (->> info-from-jruby
+       (map #(hash-map :path (key %) :classes (val %)))
+       (sort-by :path)
+       (into [])
+       (hash-map :name environment :files)))
```
cprice404: So this fn is the entirety of the munging you need to do in order to go from the Ruby API to the wire format?

src/ruby/puppet-server-lib/puppet/server/master.rb
```diff
@@ -126,14 +89,17 @@ def terminate
...
-      if File.directory?(env.manifest)
-        manifests = Dir.glob(File.join(env.manifest, '**/*.pp')).sort
-      else
-        manifests = [env.manifest]
+    manifests =
+      case
+        when env.manifest == Puppet::Node::Environment::NO_MANIFEST
+          []
+        when File.directory?(env.manifest)
+          Dir.glob(File.join(env.manifest, '**/*.pp'))
```
cprice404: Did we decide that this glob matches the current `spec` of what files Puppet will look at?  Or does the spec that Henrik wrote up as docs have any more specific restrictions than this?

cprice404:
Done with initial review.  Left a handful of nitpicks and clarifying questions, but, looking good!

src/ruby/puppet-server-lib/puppet/server/master.rb
camlow325: I read both Henrik's latest spec -- https://github.com/puppetlabs/puppet-specifications/blob/c6ce99fa77d92ac0db159befa86b94b209721fd4/language/modus-operandi.md#auto-loading-details -- and the current code in the "environment.rb" class -- https://github.com/puppetlabs/puppet/blob/4.3.2/lib/puppet/node/environment.rb#L485-L500 -- as not being more restrictive than this around the individual .pp files in the subpath of the `manifest` setting.  There are restrictions on the name of module directories under an environment's `modulepath` - only alphanumeric characters - but that all is encapsulated behind the `Module.all_manifests` method which is called below for modules.  Apparently, we don't exercise the same restrictions for subdirectories under the `manifest` setting.

Ideally, it would have been nice to be able to call some sort of method on the `Environment` class in Ruby that would provide back this list of files like `Module.all_manifests`does, but nothing appears to exist for that.  If we get back around to PUP-5011 later and decided that we wanted to continue relying upon core Ruby Puppet to help with abstracting the methods for determining the files to be parsed, maybe adding a similar `all_manifests` for the `Environment` class would be useful - although I gathered that we hoped to eventually go the other way and do all of the manifest enumeration in Clojure code instead for performance.

src/ruby/puppet-server-lib/puppet/server/master.rb
camlow325: I think the code change I made here is good anyway in that we wouldn't want to ask the `Puppet::InfoServer::ClassInformationService` to parse a file that doesn't exist.  But I agree that if we do, it might make more sense for it to error out somehow rather than just returning an empty list of classes.  Do you have any preference as to what should happen in that case?  Should the `classes_per_environment` method call on the `ClassInformationService` throw an exception, return an `:error` entry for the erroneous file entry but otherwise good data for the other files that could be parsed, or something else?

src/clj/puppetlabs/services/master/master_core.clj
camlow325: I should be able to do that.  I assume you're thinking that we'd have both a schema for the incoming format and a schema for the translated format - to be written to JSON.  I'd resisted doing that initially because of minor concerns about the performance overhead that might come from having it be on during production given that we could be dealing with maps that have hundreds? of objects in them.

Beyond the potential performance concern, the one part I was a little concerned about schematizing is the `default_literal`.   The format of these objects could be fairly variable - and could change in the future as the underlying Puppet language evolves.  I had been hoping that Puppet Server's knowledge of that could be fairly opaque, leaving it to the `ClassInformationService` in Puppet Ruby code to produce either a more primitive object - Integer, String - or tree of primitive objects that will serialize cleanly to JSON.  If we want a more comprehensive schema at this level, though, arguably we should also add some more tests in this project to exercise the different formats that could be returned.

WDYT?

src/clj/puppetlabs/services/master/master_core.clj
camlow325: Yeah, that's why I did it that way.

src/clj/puppetlabs/services/master/master_core.clj
camlow325: No, good call.  I can change that to a `vec`.

src/clj/puppetlabs/services/master/master_core.clj
camlow325: Yeah.

src/clj/puppetlabs/services/master/master_core.clj
camlow325: I can add an `-fn` on the end of the name.  I could add schema for this too.

src/clj/puppetlabs/services/master/master_core.clj
camlow325: The only other place I saw that this is done in Puppet Server is in the backend of the CA service.  It's implementation is slightly different in that it has to provide the content-type back conditionally as either "text/pson" or "application/json".

If you think it's worth it, though, I could add a `json-response` function to `puppetlabs.puppetserver.ringutils` and call that from here.  Might be good for encouraging that it be called for new endpoints that we would add to Puppet Server later on.

src/clj/puppetlabs/services/master/master_core.clj
camlow325: If I did this with a Clojure `cond` instead of the nested if statements would that seem a little more readable - or, even then, would you suggest still moving some of this into another helper function?

src/clj/puppetlabs/services/master/master_core.clj
camlow325: Yeah, I did think about just having the `environment` be a subpath under the `environment_classes` endpoint.  I agree that would look better.  But I thought that for "consistency" it would be better to just have the environment in a query parameter for now.  It felt really confusing to me to be curling different endpoints under /puppet/v3 and having to remember that the `environment_classes` does it as part of the URL whereas all of the others use a query parameter.

Maybe /puppet/v4 would provide a good opportunity to restructure all of the environment-aware URLs to incorporate the environment into the URL subpath?

src/clj/puppetlabs/services/master/master_core.clj
camlow325: Good call, thanks.  That would be a better way to do it.  I'll make that change.

src/clj/puppetlabs/puppetserver/jruby_request.clj
camlow325: Yes, it did.  Looks like you found it later on in the review.

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
camlow325: Yeah, I didn't change anything here from before - just moved to a new namespace.  Which part did you think was scary?  The idea of adding a `jruby-instance` onto the Ring request map for the next function in the middleware chain to use - or were you talking about the line that removes the `:ssl-client-cert` from the Ring request map, presumably so that isn't surfaced up to the metrics / status APIs?

test/unit/puppetlabs/services/master/master_core_test.clj
camlow325: I had been thinking that the tests at this level would be validating different good request / bad request behaviors whereas the ones at the integration level would be more about exercising the full path - from HTTP request down to Puppet Ruby code and back - for "success" cases but with a little more emphasis on the variations of the return payload to ensure for proper translation across the full path, from Ruby -> Clojure -> Serialized JSON -> Network -> Deserialized JSON.

The "returns good date for environment that exists" test block here is arguably redundant with what gets covered at the integration layer, I agree.  I wrote this test block first before everything else was working so it had value initially.  I suppose I could keep the check for "200 OK" in this test block but not do any of the other checks - JRuby borrowing behaviors and that the actual body of the JSON payload is handled properly - at this level.  Dunno.  WDYT?

src/clj/puppetlabs/services/legacy_routes/legacy_routes_service.clj
camlow325: Yeah, I'm not thrilled about this either.  I suppose we could do this differently by having a layered form of `root-routes` that the "real" master service would call with a jruby-service reference vs. one that doesn't require a jruby-service that the legacy routes service would use.  A little earlier on, we had some differences between the forms of root-routes that the master vs. legacy routes would call and that led to some problems with some aspects of the routing working differently between the two.  Can't remember the details now - maybe was tk-authz related?  In any event, I'd prefer to stick with this approach for now to avoid possibility of further divergence between the two services' uses of the master routes.  Still hoping we can kill this service soon-ish.

camlow325: I guess there have been recent attempts at doing something like this in other PRS, like in #887, where a dummy function was passed in, `(constantly nil)`, instead of the real reference.  Not totally sure how I feel about that, I guess.  Might work for now but if there ever ended up being a reason for the "real" depedency to have to be used in a legacy routes context, I'd be worried that this would be broken without us realizing it.

cprice404: I'm ok with this as-is for now.

src/clj/puppetlabs/services/master/master_core.clj
cprice404: As discussed in meatspace, I'd be fine with a schema as simple as `{schema/Keyword schema/Any}`, or whatever granularity is reasonable.  We don't need to cover all of the variable use cases super-explicitly, but it would make the code more readable to just even know that the input was a map and a string, and the output was a map.

Also agree about adding more tests, and I think maybe we talked about adding that as a line-item in our intern spreadsheeit.

src/clj/puppetlabs/services/master/master_core.clj
cprice404: k

src/clj/puppetlabs/services/master/master_core.clj
cprice404: Not necessarily; I'm happy that it's this straightforward, just was expecting that it might be slightly more munging based on our anecdotal conversations.

src/clj/puppetlabs/services/master/master_core.clj
cprice404: agreed.

src/clj/puppetlabs/services/master/master_core.clj
cprice404: Yeah, +1 on `json-response`.   It's not a huge deal but this feels like a wheel that's being re-invented a bunch across different projects and endpoints.

src/clj/puppetlabs/services/master/master_core.clj
cprice404: Just pulling out `json-response` might be sufficient.  Basically if there is going to be a tree of nested `if` forms, it seems best to try to limit the number of LOC on any of the nested branches to just one or two lines so that it's easier to understand the flow of control at a glance.  I don't know what the cond would look like w/o seeing it, maybe it would be better.

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
cprice404: No, the part about sticking the jruby instance into the ring request map.  Just feels like an opportunity to accidentally lose track of a reference, and/or have someone do something untoward that relies on being able to get the instance out of the map... and those objects are so expensive and dangerous that my ears are always going to perk up when we're not being super-strict about how we expose them.

That said, this is obviously not related to this PR, so you can ignore for now.

src/clj/puppetlabs/services/master/master_core.clj
cprice404: Also, re: schema perf, I still have yet to witness an occasion where that was any kind of bottleneck for us, so I'd prefer to err on the side of correctness and optimize later.  But in this instance my comments were mostly about readability of the fn sigs, so if you wanted to skip the `^:always-validate` part I could live with that.

src/ruby/puppet-server-lib/puppet/server/master.rb
cprice404: Yeah, I think we're on the same page in terms of future enumeration of manifests.  Just wanted to make sure that this glob adheres to the spec for now, and it sounds like it does.

src/ruby/puppet-server-lib/puppet/server/master.rb
cprice404: Not sure.  Mostly just wanted to make sure that Henrik knows what the current behavior is, and that we all agree on whether or not it's the best behavior.

cprice404: And yeah I'm fine with your changes here.

test/unit/puppetlabs/services/master/master_core_test.clj
cprice404: I dunno, yeah, I guess just all of the mock JRuby stuff here seems a little like overkill to me... and definitely the big nested JSON payload if we are going to have "real" coverage of that in the integration tests.  So I'd probably vote for trimming this down a bit, but, it's not a huge deal and not worth burning too much time on, so I'm OK with whatever you think is the appropriate level of coverage here that you think adds value.

cprice404:
Done responding to latest round of comments.

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
camlow325: Yeah, good point.  I suppose by having this wrapped in `with-jruby-puppet` that the instance should get returned to the pool when we bubble back up from the underlying middleware function calls.  But this would be problematic if the underlying middleware were to send that JRuby instance off to a background thread or something and didn't take care to ensure that nothing was using it before returning from that middleware.  In that case, we could have an instance still in use that was returned to the pool.  Good food for thought for the future, though.

src/clj/puppetlabs/services/master/master_core.clj
camlow325: I added some schema for this in 2dd5530.  I only had it go fairly top-level - just looking to see that what we get back from the JRuby layer is a `Map` and that what we translate for parsing to JSON has the intended structure down to a per-file level - but not individual classes within those files.  Hope you think that's reasonable for now.

src/clj/puppetlabs/services/master/master_core.clj
camlow325: Done in 2dd5530.

src/clj/puppetlabs/services/master/master_core.clj
camlow325: I added schema and `-fn` to the name in 2dd5530.

src/clj/puppetlabs/services/master/master_core.clj
camlow325: Did this in 2dd5530.

src/clj/puppetlabs/services/master/master_core.clj
camlow325: I started working up the `cond` approach but wasn't convinced that it looked better than what I was starting with.  Taking your last suggestion, I just moved a few of the lines over to the new `json-response` function - done in 2dd5530.  Hopefully that looks good enough for now.

test/unit/puppetlabs/services/master/master_core_test.clj
camlow325: In 2dd5530, I kept the assertion that a 200 is returned for a request with a good environment.  I had to keep a bit of the mocking for that - at least a JRubyPuppetService that can return data for what it thinks is a good environment.  I did take out some of the assertions and state around the use of JRuby pool instances and removed the large payload that was being translated previously since this is arguably already covered sufficiently in the integration tests.

src/clj/puppetlabs/services/master/master_core.clj
camlow325: Did this in 2dd5530.

src/ruby/puppet-server-lib/puppet/server/master.rb
camlow325: I'll file a PUP ticket with a repro case for this and post the link back here.

camlow325:
@cprice404 I think I've addressed all of your comments with my latest commit.  Anything else that you think needs to be addressed at this point?

cprice404:
+1, Will merge after travis.

@fpringvaldsen @haus dunno if one of you just wants to peek over the parts of this that modify the request-handler fns?  Just to make sure they make sense in the context of what you guys have in flight?

There's a bunch of this PR that you probably won't have context around so don't sweat it too much.  But I am planning on merging this today.

src/ruby/puppet-server-lib/puppet/server/master.rb
camlow325: PUP ticket filed as [PUP-5807](https://tickets.puppetlabs.com/browse/PUP-5807).

PR #896, opened 2016-02-01
(SERVER-1109) restrict code-id char set

jpinsonault:
This PR adds a restriction to what characters can make up a valid code-id
Valid code-ids must now be made of alpha-numerics and `_`, `-`, `:`, or be nil

Adds associated tests


src/clj/puppetlabs/services/versioned_code_service/versioned_code_core.clj
```diff
@@ -63,6 +63,29 @@
       (catch InterruptedException e
         (log-execution-error! e)))))
 
+(schema/defn valid-code-id? :- schema/Bool
+  "Returns false if code-id contains anything but alpha-numerics and
+  '-', '_', or ':'. nil is a valid code-id"
+  [code-id :- (schema/maybe String)]
+  (if (nil? code-id)
```
haus: This could be
```
(or
  (nil? code-id)
  (not (re-find #"[^_\-:a-zA-Z0-9]" code-id)))
```

Not sure if that is clearer or not.

test/unit/puppetlabs/services/versioned_code_service/versioned_code_core_test.clj
```diff
@@ -58,3 +58,24 @@
...
+
+(deftest valid-code-id?-test
+  (testing "valid-code-id? accepts valid code-ids"
+    (is (vc-core/valid-code-id? nil))
+    (is (vc-core/valid-code-id? "dcf16ec"))
+    (is (vc-core/valid-code-id? "master-plan:stage_1:destroy-all-humans")))
+
+  (testing "valid-code-id? rejects invalid code-ids"
+    (is (not (vc-core/valid-code-id? "bad code id")))
+    (is (not (vc-core/valid-code-id? "don't do it!")))
```
fpringvaldsen: Might want to remove the spaces here since those are tested in the last assertion, and this method should return `false` if even one invalid character shows up. Might also be worth testing `'` and `!` in separate tests

fpringvaldsen:
:+1: Other than my small test nitpick

haus:
Also :+1: other than @fpringvaldsen 's nitpick.

src/clj/puppetlabs/services/versioned_code_service/versioned_code_core.clj
jpinsonault: I think it's more clear. A little at least

jpinsonault:
@fpringvaldsen @haus Addressed your feedback

haus:
@jpinsonault something has made travis very angry with you

jpinsonault:
womp womp womp 
Rebased that last commit with an extra paren, should fix

test/integration/puppetlabs/puppetserver/error_handling_int_test.clj
```diff
@@ -54,3 +56,16 @@
...
 
+(deftest ^:integration test-invalid-code-id-error
+  (let [vcs-script (fs/absolute-path (fs/file vcs-scripts "invalid_code_id"))]
+    (testing "Catalog request fails when user provided code-id-command returns invalid code-id"
+      (bootstrap/with-puppetserver-running
+        app
+        {:versioned-code {:code-id-command vcs-script
+                          :code-content-command vcs-script}}
+        (with-test-logging
+          (let [response (testutils/http-get "puppet/v3/catalog/localhost?environment=production")]
```
haus: Testutils has a get-catalog method you might find useful here.

haus: And that only returns the body, so never mind me.

haus:
Will merge once travis is green.

haus:
Alright, my local tests are passing. I'm gonna merge this now.

joshcooper:
Sorry, just noticed this, on earlier discussion we had talked about using semicolon to separate the "value" parts of the code_id, e.g. `urn:puppet:sha=123;environment=production`. Do we need to add semicolons?

KevinCorcoran:
Dangit.  Yes, we need to add semi-colons.  @jpinsonault would you mind opening a follow-up PR for that?  Good catch, @joshcooper ... although we would have noticed that as soon as we promoted this into PE and it broke.

jpinsonault:
Thanks @joshcooper! @KevinCorcoran Can do. I think we would have missed it though since it's baked into the OSS VCS and not puppetserver itself

PR #898, opened 2016-02-02
(MAINT) Suppress ERROR messages from jruby-pool-int test output

camlow325:
This commit uses a `with-test-logging` wrapper to suppress the
ERROR messages that appear in console output when running the
test-503 jruby pool int tests.


haus:
:+1:

fpringvaldsen:
:+1: , will merge once Travis finishes

PR #899, opened 2016-02-03
(MAINT) remove not-found handler for master service

cprice404:
The `not-found` handler for `/puppet` did not appear to be strictly
necessary, and since we are referencing this route tree from other
namespaces (and potentially composing it with other route trees),
getting rid of it makes the code more re-usable.


KevinCorcoran:
seems pretty harmless to me :see_no_evil:

camlow325:
:+1: Travis looks fine so far with no test changes other than the ones that were intentional, so :shipit:

PR #902, opened 2016-02-04
(maint) Schemify puppetlabs.puppetserver.testutils

haus:
Previously these test utilities all did not have schemas applied to
them. This commit starts to change that. This introduces schemas to most
of the functions and defines a schema for puppet catalogs and resources.


nathanielksmith:
:+1: I'm into it. Did you file a bug / find out more about that key thing?

thanks for doing this!

haus:
not yet, planning to still.

jpinsonault:
:+1: thanks for doing this

haus:
for posterity, the schema bug was actually just a special case schema has with keywords in maps.

PR #903, opened 2016-02-06
(SERVER-1111) Add Etag / If-None-Match Processing to environment_classes Endpoint

camlow325:
This commit generates a hash value from the JSON response to an
environment_classes request and adds that hash to the HTTP response in
an Etag header.  The commit includes some work to re-sort maps in the
response body prior to JSON serialization so that the generated hash is
consistent even if the order in which keys had been inserted into the
original sub-maps may differ from one object creation to another.

Requests made by clients with an 'If-None-Match' header matching the
last 'Etag' delivered for an environment_classes request for the corresponding
environment now result in an HTTP 304 (Not Modified) response being delivered,
with an empty response body, instead of re-parsing the environment content and
delivering a response body based on the latest manifests on disk.

This PR also upgrades the dependency on puppetlabs/http-client to
0.5.0.  This was needed in order for test functions to properly handle
an empty response body, which is delivered in cases where the
environment_classes endpoint returns an HTTP 304 (Not Modified) in
conjunction with an empty response body per the HTTP spec.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
@@ -40,7 +40,8 @@
         (-> context
             (assoc :pool-context pool-context)
             (assoc :borrow-timeout (:borrow-timeout config))
-            (assoc :event-callbacks (atom []))))))
+            (assoc :event-callbacks (atom []))
+            (assoc :environment-class-info (atom {}))))))
   (stop
    [this context]
    (let [{:keys [pool-context]} (tk-services/service-context this)
@@ -69,17 +70,35 @@
 
   (mark-environment-expired!
     [this env-name]
-    (let [pool-context (:pool-context (tk-services/service-context this))]
+    (let [environment-class-info (:environment-class-info
+                                  (tk-services/service-context this))
+          pool-context (:pool-context (tk-services/service-context this))]
+      (swap! environment-class-info dissoc env-name)
       (core/mark-environment-expired! pool-context env-name)))
 
   (mark-all-environments-expired!
     [this]
-    (let [pool-context (:pool-context (tk-services/service-context this))]
+    (let [environment-class-info (:environment-class-info
+                                  (tk-services/service-context this))
+          pool-context (:pool-context (tk-services/service-context this))]
+      (reset! environment-class-info {})
       (core/mark-all-environments-expired! pool-context)))
 
   (get-environment-class-info
     [this jruby-instance env-name]
     (.getClassInfoForEnvironment jruby-instance env-name))
+
+  (get-environment-class-info-tag
+   [this env-name]
+   (let [environment-class-info (:environment-class-info
+                                 (tk-services/service-context this))]
+     (get @environment-class-info env-name)))
+
+  (set-environment-class-info-tag!
+   [this env-name tag]
+   (let [environment-class-info (:environment-class-info
+                                 (tk-services/service-context this))]
+     (swap! environment-class-info assoc env-name tag)))
 
   (flush-jruby-pool!
     [this]
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
@@ -41,7 +41,7 @@
             (assoc :pool-context pool-context)
             (assoc :borrow-timeout (:borrow-timeout config))
             (assoc :event-callbacks (atom []))
-            (assoc :environment-class-info (atom {}))))))
+            (assoc :environment-class-info-tags (atom {}))))))
   (stop
    [this context]
    (let [{:keys [pool-context]} (tk-services/service-context this)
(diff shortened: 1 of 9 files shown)


src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -23,52 +24,172 @@
...
-   :classes [Map]})
+   schema/Keyword schema/Any})
 
 (def EnvironmentClassesInfo
-  "Schema for the return payload an environment_classes request"
+  "Schema for the return payload an environment_classes request."
   {:name schema/Str
    :files [EnvironmentClassesFileEntry]})
 
+(defn sort-nested-java-maps
```
camlow325: The approach of sorting keys within sub-maps of the response prior to calculating a UTF-8 byte hash of the JSON-serialized object follows the approach that PuppetDB uses - see https://github.com/puppetlabs/puppetdb/blob/7f7191c883ab138dcdae3e2c06581c6c0cf9f1cc/src/puppetlabs/puppetdb/scf/hash.clj#L14-L21.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -23,52 +24,172 @@
...
+  (rr/get-header request "If-None-Match"))
+
+(schema/defn ^:always-validate
+  basic-manifest-info-from-jruby->basic-manifest-info-for-json
+    :- {schema/Any schema/Any}
+  "Convert the value for a manifest file entry into an appropriate map for
+  use in serializing an environment_classes response to JSON."
+  [file-info :- Map$Entry]
+  (let [file-detail (val file-info)]
+    (if (instance? Map file-detail)
```
camlow325: I expect this logic to go away soon - when [SERVER-1126](https://tickets.puppetlabs.com/browse/SERVER-1126) is done.  After that point, we should just be able to coerce the original RubyHash to a Clojure map rather than conditionally wrapping the content in a Map with a `:classes` key.

test/integration/puppetlabs/services/master/environment_classes_int_test.clj
```diff
@@ -111,4 +188,262 @@
                                                   (str
                                                    "{ foo_1 => 1, "
                                                    "foo_2 => 2 }")}]}]}]}
-                (response->class-info-map response))))))))
+                (response->class-info-map response))
+             "unexpected body following code change")))
+     (testing "environment class cache invalidation for one environment"
```
camlow325: This test ended up being pretty long.  The `testing` blocks prior to this one have order dependencies - one test must run before the result via caching can be validated in the next one.  These last two tests around cache invalidation could be promoted into top-level testing blocks but we'd have to bring the full puppetserver stack up and down two more times in doing so.  I opted for wrapping these all together under one puppetserver service stack instance to reduce the performance overhead a bit.  If others think it would be a better trade-off to separate these tests out, I'd be open to that, though.

test/unit/puppetlabs/services/master/master_core_test.clj
```diff
@@ -80,7 +82,100 @@
                     "non-alphanumeric characters")
         (logging/with-test-logging
          (is (= 400 (:status (request
-                              "/v3/environment_classes?environment=~")))))))))
+                              "/v3/environment_classes?environment=~"))))))
+      (testing "calculates etag properly for response payload"
```
camlow325: I found the serialization order / hash equivalence tests to be easier to do at this level than at the full-stack integration level.

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
```diff
@@ -69,18 +70,36 @@
 
   (mark-environment-expired!
     [this env-name]
-    (let [pool-context (:pool-context (tk-services/service-context this))]
+    (let [environment-class-info (:environment-class-info
+                                  (tk-services/service-context this))
+          pool-context (:pool-context (tk-services/service-context this))]
+      (swap! environment-class-info dissoc env-name)
```
camlow325: I realized after first implementing this that there would be a race condition where if a `mark-*-environment*-expired!` call were to occur while an environment_classes call was right in the middle of parsing info for that environment, the latest environment classes info would be updated based on the result of that environment_classes call - essentially causing this line to be a no-op.  This wouldn't be a problem so long as the environment_classes parsing had started after any code updates on disk had been made, but that's not entirely deterministic without the use of the PE file sync feature.

@cprice404 and I talked about this IRL and we decided, for the sake of getting an initial implementation of this new capability in place, to shore up this race condition as part of a follow-on ticket.  I filed that ticket as SERVER-1130.

cprice404:
@camlow325 fyi, reviewing this by commit, so ignore me if comments are addressed in the latter commit.

cprice404:
Done w/initial review.  Everything looking good, just a handful of nits relating to naming mostly.

camlow325:
@cprice404 - I'm done updating this again for now, pending your follow-up review.

cprice404:
+1
@camlow325 I'm basically ready to merge this now unless you think there is someone else who could give it a quick rundown, but I'm not sure anyone has enough context.

PR #904, opened 2016-02-06
(SERVER-1127) Upgrade to TK version w/HUP support

cprice404:
This commit upgrades us to a new version of TK that includes
HUP support.  The internal API for TK's app-context changed
slightly so a few tests and the repl utils needed to be cleaned
up accordingly.


camlow325:
:+1: once this can be bumped up to a released version of TK - and assuming all of the unit tests will then pass as well.

haus:
this looks good to me. :+1: when tk is available.

cprice404:
SNAPSHOT dep is fixed and this should be ready for final review now.

camlow325:
:+1: Running Travis locally right now as a sanity check.  Will merge if it passes and @haus hasn't objected to anything in the meantime.

camlow325:
Tests pass locally so merging...

PR #905, opened 2016-02-09
(SERVER-882) Restrict static_file_content endpoint

haus:
This PR updates the static-file-content-endpoint to reject any
requests that would have been rejected by the file_content Puppet
endpoint. Tests are also updated to account for the new requirement. Any
files not within `modules/*/files/**` or `environments/*/modules/*/files/**`
will be rejected with a 403 error.

### DIFF
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -2,7 +2,9 @@
   (:import (java.io FileInputStream)
            (clojure.lang IFn)
            (java.util Map Map$Entry)
-           (org.jruby RubySymbol))
+           (org.jruby RubySymbol)
+           (java.util Map)
+           (org.eclipse.jetty.util URIUtil))
   (:require [me.raynes.fs :as fs]
             [puppetlabs.puppetserver.ringutils :as ringutils]
             [puppetlabs.comidi :as comidi]
@@ -308,6 +310,17 @@
    jruby-request/wrap-with-environment-validation
    jruby-request/wrap-with-error-handling
    ring/wrap-params))
+
+(schema/defn ^:always-validate valid-static-file-path? :- schema/Bool
+  "Helper function to decide if a static_file_content path is valid.
+  The access here is designed to mimic Puppet's file_content endpoint."
+  [path :- schema/Str]
+  (if-let [canonicalized-path (URIUtil/canonicalPath path)]
+    (not (nil?
+          (or
+           (bidi.bidi/match-route [["modules/" :module-name "/files/" [#".*" :rest]] :helper] canonicalized-path)
+           (bidi.bidi/match-route [["environments/" :environment-name "/modules/" :module-name "/files/" [#".*" :rest]] :helper] canonicalized-path))))
+    false))
 
 (defn static-file-content-request-handler
   "Returns a function which is the main request handler for the
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -310,16 +310,14 @@
    jruby-request/wrap-with-error-handling
    ring/wrap-params))
 
-(schema/defn ^:always-validate valid-static-file-path? :- schema/Bool
+(schema/defn ^:always-validate valid-static-file-path?
   "Helper function to decide if a static_file_content path is valid.
   The access here is designed to mimic Puppet's file_content endpoint."
   [path :- schema/Str]
-  (if-let [canonicalized-path (URIUtil/canonicalPath path)]
-    (not (nil?
-          (or
-           (bidi.bidi/match-route [["modules/" :module-name "/files/" [#".*" :rest]] :helper] canonicalized-path)
-           (bidi.bidi/match-route [["environments/" :environment-name "/modules/" :module-name "/files/" [#".*" :rest]] :helper] canonicalized-path))))
-    false))
+  (when-let [canonicalized-path (URIUtil/canonicalPath path)]
+    (or
+     (bidi.bidi/match-route [["modules/" :module-name "/files/" [#".*" :rest]] :helper] canonicalized-path)
+     (bidi.bidi/match-route [["environments/" :environment-name "/modules/" :module-name "/files/" [#".*" :rest]] :helper] canonicalized-path))))
 
 (defn static-file-content-request-handler
   "Returns a function which is the main request handler for the
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -316,8 +316,11 @@
   [path :- schema/Str]
   (when-let [canonicalized-path (URIUtil/canonicalPath path)]
     (or
-     (bidi.bidi/match-route [["modules/" :module-name "/files/" [#".*" :rest]] :helper] canonicalized-path)
-     (bidi.bidi/match-route [["environments/" :environment-name "/modules/" :module-name "/files/" [#".*" :rest]] :helper] canonicalized-path))))
+     ;; Here, keywords represent a single element in the path. Anything between two '/' counts.
+     ;; The second vector takes anything else that might be on the end of the path.
+     ;; In the first example, this corresponds to 'modules/*/files/**' in a filesystem glob.
+     (bidi.bidi/match-route [["modules/" :module-name "/files/" [#".*" :rest]] :_] canonicalized-path)
+     (bidi.bidi/match-route [["environments/" :environment-name "/modules/" :module-name "/files/" [#".*" :rest]] :_] canonicalized-path))))
 
 (defn static-file-content-request-handler
   "Returns a function which is the main request handler for the
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -314,7 +314,7 @@
   "Helper function to decide if a static_file_content path is valid.
   The access here is designed to mimic Puppet's file_content endpoint."
   [path :- schema/Str]
-  (when-let [canonicalized-path (URIUtil/canonicalPath path)]
+  (when-let [canonicalized-path (-> path URIUtil/decodePath URIUtil/canonicalPath)]
     (or
      ;; Here, keywords represent a single element in the path. Anything between two '/' counts.
      ;; The second vector takes anything else that might be on the end of the path.
@@ -341,7 +341,7 @@
          "the files directory of a module.")}
         :else
         {:status 200
-         :body (get-code-content environment code-id (URIUtil/canonicalPath file-path))}))))
+         :body (get-code-content environment code-id (-> file-path URIUtil/decodePath URIUtil/canonicalPath))}))))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Routing
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -310,11 +310,16 @@
    jruby-request/wrap-with-error-handling
    ring/wrap-params))
 
+(schema/defn ^:always-validate decode-and-canonicalize-path :- (schema/maybe schema/Str)
+  "Takes a URI path and returns a decoded and canonicalized version."
+  [path :- schema/Str]
+  (-> path URIUtil/decodePath URIUtil/canonicalPath))
+
 (schema/defn ^:always-validate valid-static-file-path?
   "Helper function to decide if a static_file_content path is valid.
   The access here is designed to mimic Puppet's file_content endpoint."
   [path :- schema/Str]
-  (when-let [canonicalized-path (-> path URIUtil/decodePath URIUtil/canonicalPath)]
+  (when-let [canonicalized-path (decode-and-canonicalize-path path)]
      ;; Here, keywords represent a single element in the path. Anything between two '/' counts.
      ;; The second vector takes anything else that might be on the end of the path.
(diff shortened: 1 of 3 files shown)


src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -89,13 +101,17 @@
...
-      (if (or (nil? environment)
-              (nil? code-id)
-              (or (nil? file-path) (= "" file-path)))
+      (cond
+        (some empty? [environment code-id file-path])
         {:status 400
          :body "Error: A /static_file_content request requires an environment, a code-id, and a file-path"}
+        (not (valid-static-file-path? file-path))
+        {:status 403
+         :body (str "Request Denied: A /static_file_content request must be a file within "
```
haus: Would love any wordsmithing ideas on this error message.

KevinCorcoran:
can't be merged, needs a rebase

haus:
i'm rebasing as fast as i can!

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -320,13 +330,17 @@
     (let [environment (get-in req [:params "environment"])
           code-id (get-in req [:params "code_id"])
           file-path (get-in req [:params :rest])]
-      (if (or (nil? environment)
-              (nil? code-id)
-              (or (nil? file-path) (= "" file-path)))
+      (cond
+        (some empty? [environment code-id file-path])
```
KevinCorcoran: nice

test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
```diff
@@ -20,6 +20,12 @@
 
 (def num-jrubies 1)
 
+(defn get-static-file-content
+  [url-end]
+  (http-client/get (str "https://localhost:8140/puppet/v3/static_file_content/" url-end)
+                   (assoc testutils/ssl-request-options
+                     :as :stream)))
```
KevinCorcoran: This can be replaced with `:as :text` which would remove the need to `slurp` the response body everywhere it's used.

test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
```diff
@@ -76,50 +82,68 @@
...
+                           "environments/foo/bar/files"
+                           "environments/../manifests/files/site.pp"
+                           "environments/production/files/~/.bash_profile"
+                           "environments/../modules/foo/files/site.pp"
+                           "environments/../modules/foo/lib/puppet/parser/functions/site.rb"
+                           "environments/production/modules/foo/files/../../../../../../site.pp"]]
+        (testing "valid requests return 200 and the file"
+          (doseq [path valid-paths]
+            (let [response (get-static-file-content (str path "?code_id=foobar&environment=production"))]
+              (is (= 200 (:status response)))
```
KevinCorcoran: We tend to avoid assertions inside a `doseq` like this, instead preferring to duplicate the necessary parts of the test.  It makes the test more verbose and repetitive, but it's much easier to figure out exactly what's going on when an assertion fails.

test/unit/puppetlabs/services/master/master_core_test.clj
```diff
@@ -189,6 +189,29 @@
...
+                       "environments/test/modules/foo/files/bar/baz.txt"
+                       "environments/dev/modules/foo/files/path/to/file/something.txt"]
+          invalid-paths ["modules/foo/manifests/bar.pp"
+                         "manifests/site.pp"
+                         "environments/foo/bar/files"
+                         "environments/../manifests/files/site.pp"
+                         "environments/../modules/foo/lib/puppet/parser/functions/site.rb"
+                         "environments/production/files/~/.bash_profile"
+                         "environments/../modules/foo/files/site.pp"
+                         "environments/production/modules/foo/files/../../../../../../site.pp"]]
```
KevinCorcoran: It seems unnecessary to have the exact same set of paths tested at both the unit and integration levels.

test/unit/puppetlabs/services/master/master_core_test.clj
```diff
@@ -189,6 +189,29 @@
...
+      (testing "valid requests return true"
+      (doseq [path valid-paths]
```
KevinCorcoran: indentation looks off here

test/unit/puppetlabs/services/master/master_core_test.clj
haus: Removing one seems reasonable to me. The unit seems like it should the one to stay then.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -311,6 +312,15 @@
...
 
+(schema/defn ^:always-validate valid-static-file-path?
+  "Helper function to decide if a static_file_content path is valid.
+  The access here is designed to mimic Puppet's file_content endpoint."
+  [path :- schema/Str]
+  (when-let [canonicalized-path (URIUtil/canonicalPath path)]
+    (or
+     (bidi.bidi/match-route [["modules/" :module-name "/files/" [#".*" :rest]] :helper] canonicalized-path)
+     (bidi.bidi/match-route [["environments/" :environment-name "/modules/" :module-name "/files/" [#".*" :rest]] :helper] canonicalized-path))))
+
```
KevinCorcoran: I've been staring at this and not making much sense out of it.  I need to grok `match-route` better, but ... can this be simplified at all and/or should there be a comment added explaining it?  Also these lines are quite long and it'd be nice to shorten them.

haus: I can certainly add a comment explaining it and shorten the lines. I'm not sure how much simpler it can get. I guess extracting part of it into a function would probably work to simplify a little bit.

fpringvaldsen: I think I understand this but I agree a comment would be incredibly helpful.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -311,6 +312,15 @@
...
    ring/wrap-params))
 
+(schema/defn ^:always-validate valid-static-file-path?
+  "Helper function to decide if a static_file_content path is valid.
+  The access here is designed to mimic Puppet's file_content endpoint."
+  [path :- schema/Str]
+  (when-let [canonicalized-path (URIUtil/canonicalPath path)]
+    (or
+     (bidi.bidi/match-route [["modules/" :module-name "/files/" [#".*" :rest]] :helper] canonicalized-path)
+     (bidi.bidi/match-route [["environments/" :environment-name "/modules/" :module-name "/files/" [#".*" :rest]] :helper] canonicalized-path))))
```
fpringvaldsen: Hmm...is this line needed? An agent will always be hitting the `/static_file_content` and passing in a code-id and an environment, so I don't think it should ever be trying to access anything in the `environments` directory directly through the URI path.

src/clj/puppetlabs/services/master/master_core.clj
fpringvaldsen: As above, I think the intention of this endpoint is to only let a user make a request against a specific environment, so we might want to change this to something like `Request Denied: A /static_file_content request must be made for a file within the files directory of a module`

src/clj/puppetlabs/services/master/master_core.clj
haus: I was taking this from both the ticket and from the direct puppet google doc, which both indicate that environment may be in the path. The google doc says "<path> is relative to the master's codedir", which means that environments may be part of it.

test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
```diff
@@ -76,50 +82,68 @@
...
+          (let [response (get-static-file-content "modules/foo/files/bar.txt?code_id=foobar")]
+            (is (= 400 (:status response)))
+            (is (= error-message (:body response)))))
+        (testing "the /static_file_content endpoint returns an error if file-path is not provided"
+          (let [response (get-static-file-content "?code_id=foobar&environment=test")]
+            (is (= 400 (:status response)))
+            (is (= error-message (:body response))))))
+      (let [error-message "Request Denied: A /static_file_content request must be a file within the files directory of a module or a module in an environment"
+            valid-paths ["modules/foo/files/bar.txt"
+                         "modules/foo/files/bar/"
```
fpringvaldsen: Do we allow a user to get the `/static_file_content` of a directory like this?

test/unit/puppetlabs/services/master/master_core_test.clj
fpringvaldsen: :+1: on keeping the unit test

test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
haus: I'm not sure how recursive directory queries will work. I'll ask someone on client.

haus:
@KevinCorcoran pushed out an updated test that doesn't doseq.

KevinCorcoran:
@haus responding to this comment ...
> I was taking this from both the ticket and from the direct puppet google doc, which both indicate that environment may be in the path. The google doc says " is relative to the master's codedir", which means that environments may be part of it.

I don't think that's correct, the doc may just be out of date.  My understanding (and I could be wrong here) is that neither `environments` nor the name of the particular environment will be included in the file path provided by the agent.  @fpringvaldsen I believe we discussed this on one of your recent PRs - does that sound correct to you, or am I misremembering?

test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
haus: So a well behaved agent won't ask for a directory, but I don't think we can reliably know if the path in a request is a directory or not. We could reject requests with the path ending in `/`, but `modules/foo/files/bar` would still be valid and could be a directory.

fpringvaldsen:
@KevinCorcoran What you said sounds correct. To my knowledge an environment should always be specified by the `code-id` and `environment` parameters, and the file-path provided by the agent should be a relative path within the environment directory for the specified environment.

joshcooper:
> the file-path provided by the agent should be a relative path within the environment directory for the specified environment.

:+1: eg `/puppet/v3/static_file_content/modules/ntp/files/ntp.conf?code_id=...`

haus:
@KevinCorcoran @fpringvaldsen I believe I've responded to all of the feedback now.

test/unit/puppetlabs/services/master/master_core_test.clj
```diff
@@ -189,6 +190,31 @@
...
+                       "environments/foo/bar/files"
+                       "environments/../manifests/files/site.pp"
+                       "environments/../modules/foo/lib/puppet/parser/functions/site.rb"
+                       "environments/production/files/~/.bash_profile"
+                       "environments/../modules/foo/files/site.pp"
+                       "environments/production/modules/foo/files/../../../../../../site.pp"
+                       "environments/production/modules/foo/files/bar.txt"
+                       "environments/production/modules/foo/files/..conf..d../..bar..txt.."
+                       "environments/test/modules/foo/files/bar/baz.txt"
+                       "environments/dev/modules/foo/files/path/to/file/something.txt"]
```
joshcooper: Might want to reject `modules/%2E%2E/manifests/site.pp` to make sure there's no confusion about URL decoding. Also `modules/\u002e\u002e/manifests/site.pp`. Or funny business like:

```
irb(main):001:0> puts "h\u0008.i\u0008."
..
```

haus:
REBASELAND

test/unit/puppetlabs/services/master/master_core_test.clj
KevinCorcoran: @haus I just realized that @joshcooper is suggesting here that we reject this kind of path, but the test case you added in c1dee2fd82e42b1aba2db86ab88b6df40a35f997 seems to show that we allow it and pass it along to `code-id-command`.

... or am I still confused?

haus: The following manifest is totally okay in puppet, so I don't think rejecting %2E outright would be great. It would potentially disallow valid requests. I don't have a very good sense of what file names are in use for file resources though.

```
file {"/tmp/foo":
  ensure => present,
  source => 'puppet:///modules/foo/bar..d/foo%2Etxt',
}
```

KevinCorcoran:
I'm :+1: on this aside from the URL decoding issue ... I don't think I understand that issue well enough to have a strong opinion so if @joshcooper is okay with @haus's conclusion that we should not do any deconding ... :shipit:

fpringvaldsen:
I am also :+1: and am willing to defer to @joshcooper on the URL decoding issue.

haus:
I was wrong here. Decoding and encoding is happening and I wasn't looking hard enough.

haus:
I've just pushed a commit that adds URL decoding for the static_file_content endpoint.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -309,6 +310,18 @@
    jruby-request/wrap-with-error-handling
    ring/wrap-params))
 
+(schema/defn ^:always-validate valid-static-file-path?
+  "Helper function to decide if a static_file_content path is valid.
+  The access here is designed to mimic Puppet's file_content endpoint."
+  [path :- schema/Str]
+  (when-let [canonicalized-path (-> path URIUtil/decodePath URIUtil/canonicalPath)]
+    (or
```
jpinsonault: What does the `or` do here? Looks like `match-routes` returns a map or nil, and that `or` just passes that value on unmodified

haus: Oh. the or is leftover from when there were 2 match-route calls below.

haus: Updated to remove the `or`.

jpinsonault:
Piling on my :+1:

PR #906, opened 2016-02-10
(SERVER-1126) Refactoring and tests to support PUP class_per_env changes

camlow325:
This commit contains some changes which validate / support some of
the fixes / changes done to the ClassInformationService in core Ruby
Puppet, including:

* Per PUP-5809, production code and tests updated to expect the
  ClassInformationService to wrap return values for successfully parsed
  classes in a `:classes` hash.

* As an optimization, removed one layer of RubyHash -> map conversion
  during translation of environment_class file info to the JSON
  response per the changes made in PUP-5809.

* Added a couple of tests to the class-info-test suite to validate the
  removal of literal parameter values for non-JSON safe types (PUP-5744)
  and values with expressions (PUP-5713).

* Bump the Puppet submodule to a81f968 and pinned puppet-agent up to
  18ebd9a in order for the above PUP changes to be picked up.

### DIFF
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -179,26 +179,24 @@
   (rr/get-header request "If-None-Match"))
 
 (schema/defn ^:always-validate
-  basic-manifest-info-from-jruby->basic-manifest-info-for-json
-    :- {schema/Any schema/Any}
+  add-path-to-file-entry :- Map
   "Convert the value for a manifest file entry into an appropriate map for
   use in serializing an environment_classes response to JSON."
-  [file-info :- Map$Entry]
-  (let [file-detail (val file-info)]
-    (if (instance? Map file-detail)
-      (into {} file-detail)
-      {:classes file-detail})))
+  [file-detail :- Map
+   file-name :- schema/Str]
+  (.put file-detail "path" file-name)
+  file-detail)
 
 (schema/defn ^:always-validate
-  full-manifest-info-from-jruby->full-manifest-info-for-json
+  manifest-info-from-jruby->manifest-info-for-json
     :- EnvironmentClassesFileEntry
   "Convert the per-manifest file information received from the jruby service
   into an appropriate map for use in serializing an environment_classes
   response to JSON."
   [file-info :- Map$Entry]
   (-> file-info
-      (basic-manifest-info-from-jruby->basic-manifest-info-for-json)
-      (assoc :path (key file-info))
+      val
+      (add-path-to-file-entry (key file-info))
       sort-nested-environment-class-info-maps))
 
 (schema/defn ^:always-validate
@@ -214,7 +212,7 @@
   [info-from-jruby :- Map
    environment :- schema/Str]
   (->> info-from-jruby
-       (map full-manifest-info-from-jruby->full-manifest-info-for-json)
+       (map manifest-info-from-jruby->manifest-info-for-json)
        (sort-by :path)
        (vec)
        (sorted-map :name environment :files)))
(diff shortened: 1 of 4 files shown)


src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -179,26 +179,24 @@
...
   "Convert the value for a manifest file entry into an appropriate map for
   use in serializing an environment_classes response to JSON."
-  [file-info :- Map$Entry]
-  (let [file-detail (val file-info)]
-    (if (instance? Map file-detail)
-      (into {} file-detail)
-      {:classes file-detail})))
+  [file-detail :- Map
+   file-name :- schema/Str]
+  (.put file-detail "path" file-name)
```
camlow325: @cprice404 This is a follow-up to the optimization idea that we talked about in https://github.com/camlow325/puppet-server/commit/f797e3bd664a41f910afee2126d8fa5791971390#commitcomment-15958208.  I'm okay with going back to constructing a Clojure map and `assoc`ing the `:path` key/value pair to it - more like what we had before - if this seems too presumptuous.  WDYT?

test/unit/puppetlabs/services/master/master_core_test.clj
```diff
@@ -91,89 +94,94 @@
...
-                                   "default_source" "literal"},
-                                  {"name" "twoparam",
-                                   "type" "Integer",
-                                   "default_literal" "3",
-                                   "default_source" "3"}]
-                        },
-                       {"name" "twoclass"
-                        "params" []}],
-                      "/two/file" []})
+                      (map-with-classes
```
camlow325: The comparison structures in this test are basically the same as before - just wrapped in a hash map with a `:classes` key now, since that's the format the request down to the JRuby layer will now always return.

src/clj/puppetlabs/services/master/master_core.clj
cprice404: `Map` will always have a `.put`, so this seems OK to me.

test/integration/puppetlabs/services/jruby/class_info_test.clj
```diff
@@ -221,7 +224,105 @@
...
+                            "\"two\" => 3},\n"
+                            "  Hash $some_nested_hash = { \"one\" => 2, "
+                            "\"two\" => { 3 => 4 }},\n"
+                            "  Hash $another_nested_hash = { \"one\" => 2, "
+                            "\"two\" => { \"three\" => 4 }},\n"
+                            "  Array $some_array = [ 1, /^*$/ ],\n"
+                            "  Array $some_nested_array = [ 1, [ 2, "
+                            "default ] ],\n"
+                            "  Array $another_nested_array = [ 1, [ 2, 3 ] ]\n"
+                            "){}"))
```
cprice404: This is a nit that is not worth dealing with right now, but it'd be really nice if we could figure out a better way to represent multi-line string literals w/o having to put all of the `\n` chars in there.

I saw a stackoverflow post where someone wrote a helper fn like this:
```clj
(defn long-str [& strings] (clojure.string/join "\n" strings))
```

Maybe we should put something like that into kitchensink.  Just food for thought, no action necessary for now.

cprice404:
@camlow325 I'm +1 on this, but will hold off on merging until I've touched base w/you about whether you have any concerns about the timing of the submodule bump.

camlow325:
@cprice404 I'm still good with bumping to the submodule version I have up in this PR.  This is compatible with the latest one that was in PE and passing tests AFAIK.

test/integration/puppetlabs/services/jruby/class_info_test.clj
camlow325: Yeah, good idea.  Some of the "\n"s that I had here were just cosmetic formatting in case you wanted to look at the file on disk for troubleshooting.  Puppet would probably parse these files okay without them.

I agree with the idea about the ks function, though.  Might do that the next time I have a need to add on stuff like this if someone else doesn't get to it first.

PR #907, opened 2016-02-11
(SERVER-1134) Fail startup if vcs is partially configured

haus:
Previously if the versioned-code-service were partially configured, code
id would still be added to catalog requests. This commit alters that
behavior so that if only one of the two configuration settings is
present the catalog request returns a 500 error.

### DIFF
--- a/test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
+++ b/test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
@@ -51,7 +51,8 @@
      app {:jruby-puppet
           {:max-active-instances num-jrubies}
           :versioned-code
-          {:code-id-command (script-path "echo")}}
+          {:code-id-command (script-path "echo")
+           :code-content-command (script-path "echo")}}
      (let [catalog (testutils/get-catalog)]
        (is (= "production" (get catalog "code_id"))))))
   (testing "code id is added to the request body for catalog requests"
@@ -62,7 +63,8 @@
      app {:jruby-puppet
           {:max-active-instances num-jrubies}
           :versioned-code
-          {:code-id-command (script-path "echo")}}
+          {:code-id-command (script-path "echo")
+           :code-content-command (script-path "echo")}}
      (let [catalog (testutils/post-catalog)]
        (is (= "production" (get catalog "code_id"))))))
   (testing "code id is added to the request body for catalog requests"
@@ -74,7 +76,8 @@
       app {:jruby-puppet
            {:max-active-instances num-jrubies}
            :versioned-code
-           {:code-id-command (script-path "warn_echo_and_error")}}
+           {:code-id-command (script-path "warn_echo_and_error")
+            :code-content-command (script-path "echo")}}
       (let [catalog (testutils/get-catalog)]
         (is (nil? (get catalog "code_id")))
         (is (logged? #"Non-zero exit code returned while running" :error))
@@ -88,12 +91,12 @@
       {:jruby-puppet
        {:max-active-instances num-jrubies}
        :versioned-code
-       {:code-content-command (script-path "echo")}}
+       {:code-content-command (script-path "echo")
+        :code-id-command (script-path "echo")}}
       (testing "the /static_file_content endpoint successfully streams file content"
         (let [response (get-static-file-content "modules/foo/files/bar.txt?code_id=foobar&environment=test")]
           (is (= 200 (:status response)))
           (is (= "test foobar modules/foo/files/bar.txt\n" (:body response)))))
-
       (let [error-message "Error: A /static_file_content request requires an environment, a code-id, and a file-path"]
         (testing "the /static_file_content endpoint returns an error if code_id is not provided"
           (let [response (get-static-file-content "modules/foo/files/bar.txt?environment=test")]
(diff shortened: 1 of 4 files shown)


test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
```diff
@@ -68,11 +70,32 @@
...
+  (testing "code id is not added and 500 is returned if code-content-command is not set"
+    (logging/with-test-logging
+     (bootstrap/with-puppetserver-running
+      app {:jruby-puppet
+           {:max-active-instances num-jrubies}
+           :versioned-code
+           {:code-id-command (script-path "echo")}}
+      (let [response (testutils/http-get "puppet/v3/catalog/localhost?environment=production")]
+        (is (= 500 (:status response)))
+        (is (logged? #"Error 500 on SERVER"))))))
```
fpringvaldsen: Do we have a test anywhere that also tests if `code-id-command` is absent we get a 500 response?

KevinCorcoran: That is not the correct behavior, `code-id-command` is an optional setting, the result of it being absent is that the catalog will just not have a `code_id`.

fpringvaldsen: @KevinCorcoran What about the case where `code-id-command` is absent, but `code-content-command` is present? The new code returns a 500 in that case.

KevinCorcoran: Ah, yes, that case should be an error.

Also, while we're on the subject ...

How do we feel about HTTP 500 for these cases?  I don't love it, as it always seems to indicate something intrinsically wrong with the server (i.e. "my fault" (as a developer)) as opposed to a misconfiguration.  I looked into this a while ago and didn't come up with a great alternative, the best one being 409 CONFLICT.

haus: I can certainly add a test for the other case.
I'm not sure about 409 though, as the description states "This code is only allowed in situations where it is expected that the user might be able to resolve the conflict and resubmit the request." Here the user might not have any control over the configuration on the server to be able to add code-id-command or code-content-command.

haus: But that brings up a good question. If the puppetserver is not able to serve catalog requests, should it startup at all?

cprice404: The discussion in the meeting the other day mentioned that for any catalog compilation that is not using Direct Puppet, the catalog compilation should still succeed.

IIRC, the determination of whether we're using DP or not will be based on a combination of:
1. did the agent send up a query param indicating that they support static catalogs?
2. is the per-environment setting for static catalogs enabled for this environment?

The first question should be easy to answer.  The second will not be, because it would require us to dig into the environment.conf file, which is complex and something we won't be able to do for this release.  Perhaps we need to discuss this again in the DP meeting today; it may be a situation where we just need to send some value along to the Ruby code and let them make the determination of whether or not the catalog needs to be failed.

Either that, or we just fail to start up if the server is misconfigured.  That'd obviously be much easier so perhaps we should ask at the meeting whether there are any significant downsides to that.

KevinCorcoran: > The discussion in the meeting the other day mentioned that for any catalog compilation that is not using Direct Puppet, the catalog compilation should still succeed.

... even if `code-id-command` is present but `code-content-command` is not?  I don't remember that being stated, although it sounds totally reasonable, especially in a case like when `static_catalogs = false`.  However, if that's true, we definitely need to rethink our approach here.

> IIRC, the determination of whether we're using DP or not will be based on a combination of:
1. did the agent send up a query param indicating that they support static catalogs?
2. is the per-environment setting for static catalogs enabled for this environment?
> The first question should be easy to answer. 

Yes, it would be easy to answer, but AFAIK that's not in the plans for this iteration.

I agree that we should discuss this again.

cprice404: If we're going to fail catalog requests when one command is set but not the other, and we're going to do that regardless of whether the catalog compilation in question involves direct puppet, then i don't see any reason why we shouldn't just fail to start the server.  It's useless if it can't compile *any* catalogs.

fpringvaldsen:
:+1:

test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
```diff
@@ -51,7 +51,8 @@
      app {:jruby-puppet
           {:max-active-instances num-jrubies}
           :versioned-code
-          {:code-id-command (script-path "echo")}}
+          {:code-id-command (script-path "echo")
+           :code-content-command (script-path "echo")}}
```
jpinsonault: I don't know if it's worth doing now (or at all), but there's a lot of duplication in these configs, with stuff like the jrubies which aren't relevant for the test. Maybe just something to think about

jpinsonault:
:+1:

test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
haus: Yea, I'll probably work up a PR with a few improvements like that.

PR #910, opened 2016-02-12
(SERVER-1131) Throw a better error message for commands with arguments

haus:
Previously if a command with arguments were passed to
shell-utils/execute-command, it would throw an error about
'/usr/bin/echo foo' command does not exist. This commit updates the
validation to throw an error if the command does not exist, but the
first part when split on a space does exist.


fpringvaldsen:
:+1:

haus:
I think once this is merged, I or someone could add something to validate-config! in versioned-code-service to make sure that the settings for vcs are valid commands and fail if not.

jpinsonault:
:+1:

PR #911, opened 2016-02-12
(SERVER-1107) Fail fast on non-zero exit code

fpringvaldsen:
Previously, if the `code-id-command` returned a non-zero exit
code, `nil` was returned for the code-id and inserted into the
catalog.

This commit modifies this behavior so that instead of returning
nil, an exception is thrown, which causes the catalog request
to fail.

### DIFF
--- a/src/clj/puppetlabs/services/versioned_code_service/versioned_code_core.clj
+++ b/src/clj/puppetlabs/services/versioned_code_service/versioned_code_core.clj
@@ -31,10 +31,14 @@
                "Command executed: '%s', error generated: '%s'")
               cmd (.getMessage e)))
 
-(schema/defn ^:always-validate execute-code-id-script! :- (schema/maybe schema/Str)
+(schema/defn ^:always-validate execute-code-id-script! :- schema/Str
   [code-id-script :- schema/Str
    environment :- schema/Str]
-  (let [log-execution-error! #(log/error (execution-error-msg code-id-script %))]
+  (let [code-id-error-msg "code-id could not be retrieved"
+        log-execution-error! #(log/error (execution-error-msg code-id-script %))
+        throw-execution-error! (fn []
+                                 (throw (IllegalStateException.
+                                          code-id-error-msg)))]
     (try
       (let [{:keys [exit-code stderr stdout]} (shell-utils/execute-command
                                                code-id-script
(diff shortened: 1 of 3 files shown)


haus:
@fpringvaldsen there are other exceptions that can happen during command execution. currently those get caught and an error message is logged. should we be rethrowing for those cases as well?

fpringvaldsen:
@haus I think we're rethrowing all exceptions caught in `execute-code-content-script!` already

haus:
@fpringvaldsen i was thinking of

```
       (catch IllegalArgumentException e
          (log-execution-error! e))
        (catch IOException e
         (log-execution-error! e))
       (catch InterruptedException e
         (log-execution-error! e)))))
```
which are all in `execute-code-id-script!` Those all just call `log/error` so I imagine they are currently returning nil.

fpringvaldsen:
Oh, I was looking at the wrong function.

I agree with you, if we're trying to fast-fail a catalog request if the code-id command returns a non-zero exit code, then it follows that we would also want to fast-fail the catalog request if an exception occurred while trying to execute the command.

I can go ahead and change those calls so that they re-throw the exception.

jpinsonault:
`execute-code-id-script!`'s schema should probably be updated to not be a `maybe`

fpringvaldsen:
@jpinsonault That's a good idea, I'll go ahead and incorporate that

jpinsonault:
:+1:

fpringvaldsen:
@jpinsonault @haus I've just pushed up a new commit. We will now always fail a catalog request if something goes wrong calculating the code-id.

haus:
@fpringvaldsen looks like this needs a rebase.

haus:
LGTM. There's a little new repetition with the log and throw, but I'm not sure it's worth trying to refactor. :+1:

src/clj/puppetlabs/services/versioned_code_service/versioned_code_core.clj
```diff
@@ -55,13 +59,16 @@
...
+        (log-execution-error! e)
+        (throw-execution-error!))
       (catch IOException e
-        (log-execution-error! e))
+        (log-execution-error! e)
+        (throw-execution-error!))
       (catch InterruptedException e
-        (log-execution-error! e)))))
+        (log-execution-error! e)
+        (throw-execution-error!)))))
```
KevinCorcoran: This looks like it's turned into a bit of a mess.

First of all, why are we catching `IllegalArgumentException` here?  That looks a little suspicious; I doubt that's to catch an error from a call into 3rd party code, and if it's our code that we're expecting to throw an IAE, we should change that to use a slingshot exception if possible.

Second, instead of logging the original exception and then throwing a different one, can we:
a) Just remove this exception handling altogether?
b) Failing that, remove the logging and "wrap and re-throw" i.e. ...
```clj
(catch IOException e
  (throw (IllegalStateException. code-id-error-msg e))
```
(note that the original exception `e` is "wrapped" into the 2nd exception).

haus:
Replaced by https://github.com/puppetlabs/puppet-server/pull/915.

PR #912, opened 2016-02-12
(maint) Return 400 is no environment is in a catalog request

haus:
Previously, because with-code-id simply threw an exception, catalog
requests without an environment would return a 500. A bad request of 400
seems like a better match for the expected return code (as the client
controls whether or not an environment is included in the request), so
this commit updates with-code-id to throw a bad request if no
environment is present in the request.

### DIFF
--- a/test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
+++ b/test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
@@ -81,7 +81,18 @@
       (let [catalog (testutils/get-catalog)]
         (is (nil? (get catalog "code_id")))
         (is (logged? #"Non-zero exit code returned while running" :error))
-        (is (logged? #"Executed an external process which logged to STDERR: production" :warn)))))))
+        (is (logged? #"Executed an external process which logged to STDERR: production" :warn))))))
+  (testing "code id is not added and 400 is returned if environment is not included in request"
+    (logging/with-test-logging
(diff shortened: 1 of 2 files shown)


fpringvaldsen:
:+1:

test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
```diff
@@ -72,7 +72,17 @@
...
         (is (logged? #"Non-zero exit code returned while running" :error))
-        (is (logged? #"Executed an external process which logged to STDERR: production" :warn)))))))
+        (is (logged? #"Executed an external process which logged to STDERR: production" :warn))))))
+  (testing "code id is not added and 400 is returned if environment is not included in request"
+    (logging/with-test-logging
+     (bootstrap/with-puppetserver-running
+      app {:jruby-puppet
+           {:max-active-instances num-jrubies}
+           :versioned-code
+           {:code-id-command (script-path "echo")}}
```
jpinsonault: Will this need a `:code-content-command` after that last config-validation PR was merged?

haus: yes, otherwise the very worst things will happen

haus:
@jpinsonault updated.

jpinsonault:
Tests passed locally (thanks travis), merging

PR #913, opened 2016-02-16
SERVER-1118 Validate code-id-command in FOSS

erikPrime:


erikPrime:
Let me know when I ought to squish commits.

PR #915, opened 2016-02-17
(SERVER-1107) Fail on non-zero exit from code-id-command

haus:
Previously, if the `code-id-command` returned a non-zero exit
code, `nil` was returned for the code-id and inserted into the
catalog.

This commit modifies this behavior so that instead of returning
nil, an exception is thrown, which causes the catalog request
to fail.

### DIFF
--- a/src/clj/puppetlabs/services/versioned_code_service/versioned_code_core.clj
+++ b/src/clj/puppetlabs/services/versioned_code_service/versioned_code_core.clj
@@ -36,10 +36,14 @@
                "Command executed: '%s', error generated: '%s'")
               cmd (.getMessage e)))
 
-(schema/defn ^:always-validate execute-code-id-script! :- (schema/maybe schema/Str)
+(schema/defn ^:always-validate execute-code-id-script! :- schema/Str
   [code-id-script :- schema/Str
    environment :- schema/Str]
-  (let [log-execution-error! #(log/error (execution-error-msg code-id-script %))]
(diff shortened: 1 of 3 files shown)


src/clj/puppetlabs/services/versioned_code_service/versioned_code_core.clj
```diff
@@ -36,37 +36,28 @@
...
+    ;; - exit code is 0 and no stderr generated: groovy. return stdout
+    ;; - exit code is 0 and stderr was generated: that's fine. log an error
+    ;;   about the stderr and return stdout.
+    ;; - exit code is non-zero: oh no! throw an error with all the details
+    (if (zero? exit-code)
+      (do
+        (when-not (string/blank? stderr)
+          (log/error (success-with-stderr-msg code-id-script stderr)))
+        (string/trim-newline stdout))
+      (throw (IllegalStateException. (nonzero-msg code-id-script exit-code stdout stderr))))))
```
jpinsonault: Nice, looks way cleaner

jpinsonault:
Looks like @KevinCorcoran's comment from the previous PR were addressed. I'm :+1:

src/clj/puppetlabs/services/versioned_code_service/versioned_code_core.clj
KevinCorcoran: so much cleaner now!  :fireworks:

KevinCorcoran:
thanks for wrapping this up @haus

PR #916, opened 2016-02-17
(SERVER-1143) Auth rule for static file content

jpinsonault:
This PR adds an auth rule to allow authenticated access to the
static_file_content endpoint, and adds the associated acceptance test

### DIFF
--- a/acceptance/suites/tests/authorization/default_rules.rb
+++ b/acceptance/suites/tests/authorization/default_rules.rb
@@ -135,6 +135,13 @@ with_puppet_running_on(master, {}) do
     assert_allowed
   end
 
+  step 'static file content endpoint' do
+    # We'd actually need to perform a commit and use its code-id in order to
+    # get back a 200, but we know that a 400 means we got past authorization
+    curl_authenticated('/puppet/v3/static_file_content/foo/bar?environment=production')
+    assert_allowed(400)
+  end
+
   step 'certificate_revocation_list endpoint' do
     curl_authenticated('/puppet-ca/v1/certificate_revocation_list/ca?environment=production')
     assert_allowed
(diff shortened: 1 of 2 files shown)


acceptance/suites/tests/authorization/default_rules.rb
```diff
@@ -135,6 +135,13 @@ def report_query(node)
     assert_allowed
   end
 
+  step 'static file content endpoint' do
+    # We'd actually need to perform a commit and use its code-id in order to
+    # get back a 200, but we know that a 400 means we got past authorization
```
KevinCorcoran: Would it be worth, at least, adding a test case which makes a request without a client cert and verifies that the response is a 404?

This is rather weak test, but ... aside from adding the other case I just mentioned, I don't have any other ideas on how to immediately improve this.  I am assume other testing of `/static_file_content` endpoint is coming in the near future from @erikPrime.

KevinCorcoran:
@camlow325 would you mind taking a look at this?  would appreciate a brief review from you as the tk-auth expert.

acceptance/suites/tests/authorization/default_rules.rb
```diff
@@ -135,6 +135,13 @@ def report_query(node)
...
+    curl_authenticated('/puppet/v3/static_file_content/foo/bar?environment=production')
+    assert_allowed(400)
+  end
```
camlow325: It would be good to add a `curl_unauthenticated` test for this - to confirm that you get an HTTP 403 when you try to hit the endpoint with no client certificate.  I'm assuming that's the behavior we want for this endpoint, right, @KevinCorcoran?

KevinCorcoran: :)  yes, exactly what I suggested above.  Although I mistakenly said 404 when I should have said 403.

camlow325: Ah, okay.  :+1: then.

camlow325:
:+1: overall from me other than the one question on adding an "unauthorized" beaker test.

jpinsonault:
@camlow325 Added an unauthenticated test

KevinCorcoran:
LGTM :shipit:

acceptance/suites/tests/authorization/default_rules.rb
haus: To get back a 200, we would also need a valid file path inside of a module (`modules/files/foo/bar` for example).

haus:
:+1:

PR #917, opened 2016-02-18
(SERVER-1153) Handle superfluous --gzip if-none-match header characters

camlow325:
This commit strips the characters "--gzip" from the end of an
'If-None-Match' HTTP header value when comparing the value to that of
the last etag stored for the environment_classes cache.  Previously, no
special compensation for these characters was being done.  This had led
to failures matching up the supplied 'If-None-Match' header value to
that of the last stored value, in cases where the Jetty web server was
appending the "--gzip" characters to the etag value as it delivered a
gzipped response body to a prior environment_classes request.


cprice404:
@camlow325 this looks good to me, but I'm wondering if it would be possible / worthwhile to have a test that explicitly forces a payload that exceeds the gzip size, validates that the response is gzipped, and validates that sending the etag from that response back up to the server will result in a 304?  If we think it's worthwhile to write such a test, it could be targeted at just your middleware thing, so the test would only need to spin up Jetty and a simple ring handler that uses that middleware, as opposed to spinning up the full server stack and doing an environment_classes request.

I guess I'm thinking that even if you do think that is a good idea, we shouldn't block this PR on it, so, I'm +1.

camlow325:
@cprice404 I can try doing that.  Although there are no guarantees that Jetty behavior won't change to have a different size threshold for when it starts gzipping the response in the future.  I'll give it a shot.

cprice404:
@camlow325 ok, but if we get a second review on this ( @rlinehan ? ) before you're done with that, we should go ahead and merge this just to make sure it's not impacting the NC team.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -177,7 +178,22 @@
...
+  ;; value before the Ring request would see it.  The version of Jetty being
+  ;; used at the time this code was written (9.2.10), however, did not have this
+  ;; logic to strip the "--gzip" characters from the incoming header.  This
+  ;; function compensates for that by stripping the characters here - before
+  ;; other Puppet Server code would use it. When/if Puppet Server is upgraded to
+  ;; a version of trapperkeeper-webserver-jetty9 which is based on Jetty 9.3.x
+  ;; or newer, it may be safe to take out the line that removes the '--gzip'
+  ;; characters.
+  (some-> (rr/get-header request "If-None-Match")
+          (str/replace #"--gzip$" "")))
```
rlinehan: Are we always guaranteed that this `--gzip` will be at the end of the header? Is it possible it would ever show up somewhere else, or that something might be added on to the header after the `--gzip`?

camlow325:
@cprice404 I added another integration test in 6bf7f96 which hopefully achieves what you were suggesting.  The test just includes a `wrap-with-etag-check` middleware and some surrounding mocks / backing ring middleware which returns a large response body - 200000 characters - to more explicitly trigger gzip encoding.  It's hard to predict for sure that gzip encoding will be "guaranteed" to happen if we were to upgrade to a newer Jetty web server version which might have different rules than the ones today around when gzipping occurs.  If that were the case, though, hopefully it wouldn't be too tough to adapt this test to whatever new rules we might encounter.

src/clj/puppetlabs/services/master/master_core.clj
camlow325: @rlinehan I don't think we're "guaranteed" that the characters will be exactly these and exactly at the end of the string.  Hopefully, the 'fuller stack' test that I added in 6bf7f96 would start failing / help catch a change in the formatting if Jetty were ever to do that, though.

cprice404:
@camlow325  thanks for adding the test, that seems valuable to me.

+1

test/integration/puppetlabs/services/master/environment_classes_int_test.clj
```diff
@@ -458,3 +480,43 @@
...
+                 (master-core/response-with-etag
+                  (apply str (repeat body-length "a"))
+                  expected-etag))
+               (master-core/wrap-with-etag-check jruby-service))]
+      (jetty9/with-test-webserver
+       app
+       port
+       (let [request-url (str "http://localhost:" port)
+             initial-response (http-client/get request-url {:as :text})
+             response-tag (response-etag initial-response)
```
cprice404: Would it be a terrible idea to add an assertion for now that asserts that the `response-tag` has the `--gzip` suffix?  We know that that is stupid behavior, but having the assertion in the test would at least codify that we are expecting the stupid behavior, and would ensure that we'd notice if Jetty changed anything about that in the future?

camlow325: @cprice404 I'd thought about doing that initially and ultimately didn't since I wasn't sure if we wanted a test to start failing because of a new behavior that didn't "break" anything in our implementation.  That said, though, it's probably a reasonable thing to do.  And maybe if it did start failing, it would prompt us to look into Jetty's new implementation further to see if there's something else we'd need to cover now and/or be able to undo the hacky "--gzip" handling code in Puppet Server.  That seems like it could be valuable.

As one other aside here...

I also debated whether to write this test using a service bootstrapped via `with-app-with-config` rather than the relatively simpler `jetty9/with-test-webserver` to make the test a little more friendly for switching it to run on some non-Jetty based implementation of the `:WebserverService` protocol in the future.  If instead, though, we're acknowledging that this test only really exists to validate Jetty's Etag manipulation behavior - including adding an assertion that checks for "--gzip" in the Etag value - then I don't think it would be worth switching over to `with-app-with-config`.

@cprice404 - you good with just adding the extra "--gzip" Etag value assertion and sticking with `jetty9/with-test-server` for now?

cprice404: yes, the test is fine as it is, just thought maybe the extra assertion would have some value in the future.

src/clj/puppetlabs/services/master/master_core.clj
camlow325: @rlinehan Sorry, I may not have provided enough detail for the question you were asking.

In Jetty 9.2.10 - and the latest Jetty, 9.3.7 - the specific characters "--gzip" will be added to the end of the original tag only when the corresponding response body has been gzipped.  I don't see anything in Jetty's current code that would lead to those extra characters appearing anywhere other than the end of the string.

See the following lines for where Jetty does this:

Jetty v9.2.10:

https://github.com/eclipse/jetty.project/blob/jetty-9.2.10.v20150310/jetty-servlets/src/main/java/org/eclipse/jetty/servlets/gzip/GzipHttpOutput.java#L199
https://github.com/eclipse/jetty.project/blob/jetty-9.2.10.v20150310/jetty-servlets/src/main/java/org/eclipse/jetty/servlets/AsyncGzipFilter.java#L135

Jetty 9.3.7: 

https://github.com/eclipse/jetty.project/blob/jetty-9.3.7.v20160115/jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipHttpOutputInterceptor.java#L218-L219
https://github.com/eclipse/jetty.project/blob/jetty-9.3.7.v20160115/jetty-http/src/main/java/org/eclipse/jetty/http/GzipHttpContent.java#L34

I don't think we're guaranteed that Jetty won't change its implementation later to do something different.  Although, it appears that just by using the GzipHandler, as we already do in latest trapperkeeper-webserver-jetty9 code, that we'll get some of the compensation for Etag / If-None-Match values "for free" as we upgrade to a newer Jetty version:

https://github.com/eclipse/jetty.project/blob/jetty-9.3.7.v20160115/jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipHandler.java#L440-L454

Also, the new integration test that I added for this PR should help catch a change in behavior if there is one in a newer Jetty.

I'd hope that the likelihood of Jetty changing this in a way that would break backward compatibility for upgrading users would be relatively low.  I've come across a fair amount of complaints from users to the Jetty project about the current behaviors and workarounds that they are doing to compensate for it and would hope that the Jetty maintainers would be mindful about this regarding any future changes they may decide to make in this area.

test/integration/puppetlabs/services/master/environment_classes_int_test.clj
camlow325: @cprice404 Sounds good.  I added the extra assertion in df6d6a0.

cprice404:
+1 ; @rlinehan please feel free to merge when you see fit.

src/clj/puppetlabs/services/master/master_core.clj
rlinehan: Cool, thanks for the explanation. I should have expanded my question too - I had noticed that in the place where they're doing the removal of the `--gzip` in the Jetty 3.7 code that they're doing `etag.indexOf(ETAG_GZIP_QUOTE)` rather than just assuming that it was going to be at the end.

But I agree that all the code you've linked to indicates that it should be at the end, and we have tests that should catch if it changes, so :+1:.

rlinehan:
:shipit:

src/clj/puppetlabs/services/master/master_core.clj
camlow325: > I had noticed that in the place where they're doing the removal of the --gzip in the Jetty 3.7 code that they're doing etag.indexOf(ETAG_GZIP_QUOTE) rather than just assuming that it was going to be at the end.

Yep, good point.  I noticed that, too.  Not sure if they were just being extra defensive against the possibility that duplicate sets of "--gzip" characters could somehow be appended onto the tag at the point it is round-tripped or something?  Since I didn't see anything in their response handling code which would produce a tag that looks like that, though, I figured I'd just code the Puppet Server side of this to handle removing the known one at the end for now.  Definitely something to keep an eye on down the road, though...

PR #919, opened 2016-02-19
(MAINT)confine statement failed in PE; replaced with skip

erikPrime:
This test will cause the pe-puppet-server-extensions pipe to file until this PR is merged.


kurtwall:
:see_no_evil:

PR #920, opened 2016-02-19
(MAINT) Bump tk-jetty9 dep to 1.5.1

camlow325:
This commit bumps Puppet Server's dependency on
trapperkeeper-webserver-jetty9 to 1.5.1.  This is being done primarily
in order for Puppet Server to utilize some memory leak fixes related to
JMX beans in a HUP/restart scenario that were added in the latest
release.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,6 +1,6 @@
 (def clj-version "1.7.0")
 (def tk-version "1.3.0")
-(def tk-jetty-version "1.3.1")
+(def tk-jetty-version "1.5.1")
 (def ks-version "1.3.0")
 (def ps-version "2.2.2-master-SNAPSHOT")
 
@@ -100,6 +100,7 @@
 
   :profiles {:dev {:source-paths  ["dev"]
                    :dependencies  [[org.clojure/tools.namespace "0.2.5"]
+                                   [org.clojure/java.jmx "0.2.0"]
                                    [puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty-version]
                                    [puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty-version :classifier "test"]
                                    [puppetlabs/trapperkeeper ~tk-version :classifier "test" :scope "test"]


project.clj
```diff
@@ -100,6 +100,7 @@
 
   :profiles {:dev {:source-paths  ["dev"]
                    :dependencies  [[org.clojure/tools.namespace "0.2.5"]
+                                   [org.clojure/java.jmx "0.2.0"]
```
cprice404: why did you add this as a dev dependency?

camlow325: I needed `org.clojure/java.jmx` to be included during test runs b/c without it, I'd get this failure when I tried to run `lein test`:

~~~~
Exception in thread "main" java.io.FileNotFoundException: Could not locate clojure/java/jmx__init.class or clojure/java/jmx.clj on classpath., compiling:(puppetlabs/trapperkeeper/testutils/webserver.clj:1:1)
...
	at puppetlabs.puppetserver.ruby.http_client_test$eval24317$loading__5340__auto____24318.invoke(http_client_test.clj:1)
	at puppetlabs.puppetserver.ruby.http_client_test$eval24317.invoke(http_client_test.clj:1)
...
~~~~

Looks like jmx is only used in tk-jetty9 as a dev dependency for tests as well.  So apparently we don't get it transitively from either tk-jetty9 or tk-jetty9's testutils jars?

Should I have resolved this in a different way?

cprice404: hrm.  seems like jetty should be forcing it to come in as a transitive dep, and if it's not, we should (ultimately) probably fix that in jetty... but... I don't know if it's worth fixing right this second.

ugh.  i think this means that anyone who is using jetty testutils is going to have to deal with this... so yeah i guess maybe i do think we should fix it in jetty now... wdyt?

camlow325: Fixing it in Jetty sounds like it's probably a good thing to do.  How would you propose doing that - by adding it to the top-level `:dependencies` in the tk-jetty9 "project.clj"?

cprice404: The ideal thing would be to add it as a dep for the test jar, which maybe could be done via the test profile or something?  Failing that, adding it as a top-level dep seems better than where we are now.

camlow325: I tried rebuilding the tk-jetty9 testutils jar with this added to the tk-jetty9 `project.clj`:

~~~~
:testutils {:dependencies [[org.clojure/java.jmx "0.2.0"]]
               ...}
~~~~

The testjar built and installed to my local .m2 repository okay.  When I bumped my Puppet Server project.clj to use that latest snapshot install, though, I got the same "Could not locate clojure/java/jmx" error as before.

Is what I did what you were expecting to be able to do?

cprice404: can you look at the pom.xml in your .m2 and see if it's in there?  if not, maybe google for 15 mins and then punt and make it a regular dep?

camlow325: It does appear in my ".m2/repository/puppetlabs/trapperkeeper-webserver-jetty9/1.5.2-SNAPSHOT/trapperkeeper-webserver-jetty9-1.5.2-SNAPSHOT.pom":

~~~~
...
<dependencies>
  ...
  <dependency>
      <groupId>org.clojure</groupId>
      <artifactId>java.jmx</artifactId>
      <version>0.2.0</version>
      <scope>test</scope>
  </dependency>
  ...
</dependencies>
~~~~

I'll spend a few more minutes seeing if I can figure out something...

cprice404: does `lein deps :tree` show the test jar, and does it have anything at all below it?

camlow325: I don't see any output for the test jar in it, no.

I also tried just putting the `[org.clojure/java.jmx "0.2.0"]` in as a first level dependency in the tk-jetty9 project.clj.  When I `lein install` that to my .m2 repo and try to reference it from a local puppet-server repo `lein test` run, though, I get the same error.  Shouldn't that at least have worked?

rlinehan: When I put jmx as a top-level dependency in tk-j9 and do `lein deps :tree` it shows up as `[org.clojure/java.jmx "0.2.0" :scope "test"]`

does this `:scope "test"` bit maybe have something to do with it?

If I removed jmx from the `dev` profile in tk-j9 (and kept it in as a normal dependency), then it didn't show up with this, and if I `lein install`ed it and used it in puppet-server then it worked.

camlow325: @rlinehan Ah, yeah, that's the step I was missing.  I had added `org.clojure/java.jmx` as a top-level dep but still had it as a `:dev` profile dependency dep as well.  When I remove the reference to it from the `:dev` profile in tk-jetty9's project.clj but have it as a top-level dependency in tk-jetty9's project.clj, then it works like you mentioned.

Still haven't found a way to make it work as a "test" only dependency within tk-jetty9's project.clj, though.

cprice404: just put it as a top-level then so we can keep moving.

camlow325: I did this in https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/pull/132 and released a new tk-jetty9 from it, 1.5.2.  I'll close this PR out and open a new one that bumps the tk-jetty9 dependency to 1.5.2.

camlow325: New PR added as #922.

PR #921, opened 2016-02-19
(SERVER-1142) Sanely handle auth and param parsing for clojure routes

haus:
Previously clojure routes were unauthenticated if legacy auth was
enabled because it was presumed that all routes would be handed down
into puppet for auth via auth.conf. As that is not the case, this commit
separates the jruby request handler from a new clojure request wrapper.
The jruby request handler is applied to all ruby routes, and conditional
authentication is applied to that handler. The new clojure wrapper
unconditionally applies authentication and parses the request for
params.
The legacy routes service also needed a corresponding change to account
for the changes to get-wrapped-handler and root-routes.
Additionally, this commit defines schemas for some of the existing and
all of the new functions in master-core.

### DIFF
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -406,13 +406,12 @@
 (schema/defn ^:always-validate
   v3-clojure-routes :- bidi-schema/RoutePair
   "v3 route tree for the clojure side of the master service."
-  [clojure-request-wrapper :- IFn
-   jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)
+  [jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)
    get-code-content-fn :- IFn]
   (let [environment-class-handler
-        (clojure-request-wrapper (environment-class-handler jruby-service))
+        (environment-class-handler jruby-service)
         static-file-content-handler
-        (clojure-request-wrapper (static-file-content-request-handler get-code-content-fn))]
+        (static-file-content-request-handler get-code-content-fn)]
     (comidi/routes
      (comidi/GET ["/environment_classes" [#".*" :rest]] request
                  (environment-class-handler request))
@@ -430,7 +429,9 @@
    get-code-content-fn :- IFn]
   (comidi/context "/v3"
                   (v3-ruby-routes ruby-request-handler)
-                  (v3-clojure-routes clojure-request-wrapper jruby-service get-code-content-fn)))
+                  (comidi/wrap-routes
+                   (v3-clojure-routes jruby-service get-code-content-fn)
+                   clojure-request-wrapper )))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Lifecycle Helper Functions
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -375,41 +375,41 @@
   (let [environment-class-handler (environment-class-handler jruby-service)
         static-file-content-handler (static-file-content-request-handler get-code-content-fn)]
     (let [ruby-routes (comidi/routes
-     (comidi/GET ["/node/" [#".*" :rest]] request
-                 (request-handler request))
-     (comidi/GET ["/file_content/" [#".*" :rest]] request
-                 (request-handler request))
-     (comidi/GET ["/file_metadatas/" [#".*" :rest]] request
-                 (request-handler request))
-     (comidi/GET ["/file_metadata/" [#".*" :rest]] request
-                 (request-handler request))
-     (comidi/GET ["/file_bucket_file/" [#".*" :rest]] request
-                 (request-handler request))
-     (comidi/PUT ["/file_bucket_file/" [#".*" :rest]] request
-                 (request-handler request))
-     (comidi/HEAD ["/file_bucket_file/" [#".*" :rest]] request
-                  (request-handler request))
+           (comidi/GET ["/node/" [#".*" :rest]] request
+                       (request-handler request))
+           (comidi/GET ["/file_content/" [#".*" :rest]] request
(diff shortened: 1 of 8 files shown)


test/unit/puppetlabs/services/master/master_core_test.clj
```diff
@@ -18,7 +19,7 @@
 
 (defn build-ring-handler
   [request-handler puppet-version jruby-service]
-  (-> (root-routes request-handler jruby-service (constantly nil))
+  (-> (root-routes request-handler #(ring/wrap-params %) jruby-service (constantly nil))
```
jpinsonault: I think this can just be ring/wrap-params without the anonymous fn

jpinsonault:
This looks good to me, I just had one tiny nitpick :+1:

test/unit/puppetlabs/services/master/master_core_test.clj
haus: you're totally right.

test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
```diff
@@ -154,3 +154,18 @@
         (is (= 500 (:status response)))
         (is (re-matches #".*Cannot retrieve code content because the \"versioned-code\.code-content-command\" setting is not present in configuration.*"
                         (:body response))))))))
+
+(deftest ^:integration static-file-content-works-with-legacy-auth
```
camlow325: Might be better to have this test live in the `auth_conf_test.clj` test namespace?

test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
```diff
@@ -154,3 +154,18 @@
...
+  (testing "The static_file_content endpoint works even if legacy-auth is enabled."
+    ;; with-test-logging is used here to supress a warning about running with legacy auth enabled.
```
camlow325: s/supress/suppress

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -367,51 +366,72 @@
...
+               (request-handler request))))
 
+(schema/defn ^:always-validate
+  v3-clojure-routes :- bidi-schema/RoutePair
+  "v3 route tree for the clojure side of the master service."
+  [clojure-request-wrapper :- IFn
+   jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)
+   get-code-content-fn :- IFn]
+  (let [environment-class-handler
+        (clojure-request-wrapper (environment-class-handler jruby-service))
```
camlow325: I think the way @cprice404 was envisioning this would work is that the whole route subtree would somehow be wrapped by the `clojure-request-wrapper` rather than having to individually wrap each sub-handler with the `clojure-request-wrapper`.  TBH, though, I'm not sure if wrapping the whole sub-tree is something you can do with comidi/bidi.

@cprice404 - can you provide a sanity check here?

test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
haus: That seems reasonable.

haus: The only downside to moving it is some duplication of test fixtures like the echo script.

test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
```diff
@@ -154,3 +154,18 @@
...
+    (logging/with-test-logging
+     (bootstrap/with-puppetserver-running
+      app
+      {:jruby-puppet {:use-legacy-auth-conf true}
+       :versioned-code
+       {:code-content-command (script-path "echo")
+        :code-id-command (script-path "echo")}}
+      (testing "for legacy puppet routes"
+        (let [response (get-static-file-content "modules/foo/files/bar?code_id=foobar&environment=test")]
```
camlow325: Do we have a test already that would cover ensuring that the static_file_content endpoint works if legacy-auth is disabled?

Also, is there a way we could write this test where we could ensure that tk-auth is evaluating the route in both cases - as opposed to just automatically always granting access to the endpoint?  Maybe you could do this by bootstrapping Puppet Server with a custom tk-auth rule where you could validate that a 200 (success) corresponds to a case where the auth rule would allow the request vs. a 403 (failure) would correspond to a case where the auth rule would deny the request?

haus: Updated to include a 200 and 403 test.

camlow325:
:+1: with just the one question about whether there's another way to do the wrapping the of `clojure-request-wrapper` logic around the whole route sub-tree w/o having to wrap it around each sub-handler individually.  If there's no sane way to do that, I'm good with just using what you've already done.

haus:
@camlow325 so there is `comidi/routes->handler` that takes a route tree and returns a handler. that would be easy to wrap with the clojure wrapper, but i didn't see an obvious way to take that handler and pass it to `comidi/context` to rejoin the other route subtree.

KevinCorcoran:
Travis failed with:
```
lein test :only puppetlabs.services.jruby.jruby-pool-int-test/test-restart-comes-back
FAIL in (test-restart-comes-back) (jruby_pool_int_test.clj:308)
After a TK restart puppetserver can still handle requests
expected: (= (clojure.core/deref call-seq) [:init-bonus-service :start-bonus-service :stop-bonus-service :init-bonus-service :start-bonus-service])
  actual: (not (= [:init-bonus-service :start-bonus-service :stop-bonus-service] [:init-bonus-service :start-bonus-service :stop-bonus-service :init-bonus-service :start-bonus-service]))
```
... and then a bunch of other errors.

Maybe a race condition in that test?   Doesn't seem related to this PR, and only failed in one cell.

camlow325:
@KevinCorcoran I think @haus said that test has been failing intermittently on other PRs so I doubt the changes on this one are related.  I'm good with this PR being merged whenever you are.  WDYT?

KevinCorcoran:
@camlow325 I haven't looked at the code, but if you and @jpinsonault are :+1: then :shipit:

PR #924, opened 2016-02-23
(SERVER-1007) Print Timezone format with CA correctly on Java 8u60+

justinstoller:
The version of Joda Time used by the previous version of clj-time had
an issue on Java 8u60+ that caused it to improperly print a different
timestamp.

This commit bumps the version of clj-time, which pins to a newer
version of Joda Time, which does not have this bug on Java 8u60 and
above.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -28,7 +28,7 @@
                  [clj-yaml "0.4.0" :exclusions [org.yaml/snakeyaml]]
                  [commons-lang "2.6"]
                  [commons-io "2.4"]
-                 [clj-time "0.10.0"]
+                 [clj-time "0.11.0"]
                  [prismatic/schema "1.0.4"]
                  [me.raynes/fs "1.4.6"]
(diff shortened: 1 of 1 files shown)


justinstoller:
Do I need to update this for ppse as well?

fpringvaldsen:
@justinstoller Yeah, pe-p-s-e pins to a specific version of Joda Time, so that will need to be updated as well.

fpringvaldsen:
@justinstoller It will also need to be updated for pe-puppetserver, as that project also pins to a specific version of both `clj-time` and `joda-time`.

fpringvaldsen:
:see_no_evil: I'll merge this once travis is finished

justinstoller:
Added the commensurate PRs against: https://github.com/puppetlabs/pe-puppetserver/pull/73 and https://github.com/puppetlabs/pe-puppet-server-extensions/pull/555

PR #925, opened 2016-02-23
(MAINT) Add construct-root-routes-handler fn

fpringvaldsen:
Extract the creation of the ruby request handler and the
clojure request wrapper, as well as their use in the creation of
the root-routes handler, into a separate function,
construct-root-routes-handler.

### DIFF
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -14,7 +14,8 @@
             [puppetlabs.kitchensink.core :as ks]
             [cheshire.core :as cheshire]
             [clojure.string :as str]
-            [bidi.schema :as bidi-schema]))
+            [bidi.schema :as bidi-schema]
+            [ring.middleware.params :as ring]))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Constants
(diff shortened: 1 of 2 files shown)


haus:
This looks pretty reasonable to me.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -549,4 +550,28 @@
    (let [handler-maybe-with-authorization (if use-legacy-auth-conf
                                             route-handler
                                             (authorization-fn route-handler))]
-     (wrap-middleware handler-maybe-with-authorization puppet-version))))
\ No newline at end of file
+     (wrap-middleware handler-maybe-with-authorization puppet-version))))
+
+(schema/defn ^:always-validate
+  construct-root-routes-handler :- bidi-schema/RoutePair
```
haus: nitpick: this doesn't really produce a handler. the `comidi/routes->handler` call does that later on.

fpringvaldsen: Ah, okay. Do you know what it does produce?

haus: just a route tree. i think calling it `construct-root-routes` makes sense to me.

fpringvaldsen:
@haus I've changed the name of that function

haus:
:+1:

KevinCorcoran:
What is the reason for this change?  (I assume it's so that this code will work in PE.)  If that's so, It would probably be worthwhile to update the commit message to explain that.  I'm not usually a stickler for that kind of thing, but in this case, it's not at all obvious at a glance why this change is necessary.

KevinCorcoran:
:+1: looks good except for perhaps updating commit message / PR description

KevinCorcoran:
voomed into `2.2.2-master-20160224_010132-g5a62912`

PR #926, opened 2016-02-23
(SERVER-1130) Add a per-environment timestamp into the env-classes cache

camlow325:
This commit adds a "last-updated" timestamp per environment to the
environment classes cache.  The timestamp is used by incoming
environment_classes cache requests to determine if the cache info has
changed between the time the request starts parsing for info and when
the cache is about to be updated, e.g., if a call from a separate
request to mark-environment-expired were to occur during parsing for
class info in a request.  If the timestamp has been updated during that
window, the cache is not updated with the latest parsed data, since that
data may already be stale.

This commit also adds an integration test,
class-info-updated-after-cache-flush-during-prior-request, which
performs an environment cache flush while a class info request is in
progress and validates that the following class info request gets the
latest data.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
@@ -72,14 +72,22 @@
     [this env-name]
     (let [{:keys [environment-class-info-tags pool-context]}
           (tk-services/service-context this)]
-      (swap! environment-class-info-tags dissoc env-name)
+      (swap! environment-class-info-tags
+             assoc
+             env-name
+             {:tag nil
+              :last-updated (System/currentTimeMillis)})
       (core/mark-environment-expired! pool-context env-name)))
 
   (mark-all-environments-expired!
     [this]
     (let [{:keys [environment-class-info-tags pool-context]}
           (tk-services/service-context this)]
-     (reset! environment-class-info-tags {})
+      (swap! environment-class-info-tags
+             #(let [now (System/currentTimeMillis)]
+               (into {} (for [k (keys %)]
+                          [k {:tag nil
+                              :last-updated now}]))))
      (core/mark-all-environments-expired! pool-context)))
 
   (get-environment-class-info
@@ -90,13 +98,26 @@
    [this env-name]
    (let [environment-class-info (:environment-class-info-tags
                                  (tk-services/service-context this))]
-     (get @environment-class-info env-name)))
+     (get-in @environment-class-info [env-name :tag])))
 
-  (set-environment-class-info-tag!
-   [this env-name tag]
+  (get-environment-class-info-tag-last-updated
+   [this env-name]
    (let [environment-class-info (:environment-class-info-tags
                                  (tk-services/service-context this))]
-     (swap! environment-class-info assoc env-name tag)))
+     (get-in @environment-class-info [env-name :last-updated])))
+
+  (set-environment-class-info-tag!
+   [this env-name tag last-update-before-tag-computed]
+   (let [environment-class-info (:environment-class-info-tags
+                                 (tk-services/service-context this))]
+     (swap! environment-class-info
+            #(let [cache-last-updated
+                   (get-in % [env-name :last-updated])]
+              (if (= cache-last-updated last-update-before-tag-computed)
+                (assoc % env-name
+                         {:tag tag
+                          :last-updated (System/currentTimeMillis)})
+                %)))))
 
   (flush-jruby-pool!
     [this]
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
@@ -6,7 +6,8 @@
             [puppetlabs.trapperkeeper.services :as tk-services]
             [puppetlabs.services.protocols.jruby-puppet :as jruby]
             [slingshot.slingshot :as sling]
-            [puppetlabs.services.jruby.jruby-puppet-schemas :as jruby-schemas]))
+            [puppetlabs.services.jruby.jruby-puppet-schemas :as jruby-schemas]
+            [puppetlabs.kitchensink.core :as ks]))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Public
@@ -75,8 +76,7 @@
       (swap! environment-class-info-tags
              assoc
              env-name
-             {:tag nil
-              :last-updated (System/currentTimeMillis)})
+             (core/environment-class-info-entry))
       (core/mark-environment-expired! pool-context env-name)))
 
   (mark-all-environments-expired!
@@ -84,10 +84,7 @@
     (let [{:keys [environment-class-info-tags pool-context]}
           (tk-services/service-context this)]
       (swap! environment-class-info-tags
-             #(let [now (System/currentTimeMillis)]
-               (into {} (for [k (keys %)]
-                          [k {:tag nil
-                              :last-updated now}]))))
+             #(ks/mapvals (fn [_] (core/environment-class-info-entry)) %))
      (core/mark-all-environments-expired! pool-context)))
 
   (get-environment-class-info
@@ -115,8 +112,7 @@
                    (get-in % [env-name :last-updated])]
               (if (= cache-last-updated last-update-before-tag-computed)
                 (assoc % env-name
-                         {:tag tag
-                          :last-updated (System/currentTimeMillis)})
+                         (core/environment-class-info-entry tag))
                 %)))))
 
   (flush-jruby-pool!
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
@@ -108,12 +108,10 @@
    (let [environment-class-info (:environment-class-info-tags
                                  (tk-services/service-context this))]
      (swap! environment-class-info
-            #(let [cache-last-updated
-                   (get-in % [env-name :last-updated])]
-              (if (= cache-last-updated last-update-before-tag-computed)
-                (assoc % env-name
-                         (core/environment-class-info-entry tag))
-                %)))))
+            core/environment-class-info-cache-updated-with-tag
+            env-name
+            tag
+            last-update-before-tag-computed)))
 
   (flush-jruby-pool!
     [this]
--- a/test/integration/puppetlabs/services/master/environment_classes_int_test.clj
+++ b/test/integration/puppetlabs/services/master/environment_classes_int_test.clj
@@ -658,7 +658,7 @@
                          "during a previous class info request")
              (is (= 200 (:status response-after-update))
                  (str
-                  "unexpected status code for response after update"
+                  "unexpected status code for response after update, "
                   "response: "
                   (ks/pprint-to-string response-after-update)))
              (is (not= initial-response-etag
--- a/test/integration/puppetlabs/services/master/environment_classes_int_test.clj
+++ b/test/integration/puppetlabs/services/master/environment_classes_int_test.clj
@@ -645,7 +645,7 @@
                   "unexpected status code for initial response"
                   "response: "
                   (ks/pprint-to-string initial-response)))
-             (is (not (nil? initial-response))
+             (is (not (nil? initial-response-etag))
                  "no etag returned for initial response")
              (is (= {"name" "production",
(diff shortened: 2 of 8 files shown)


camlow325:
Ping @cprice404 for review whenever you get a chance.

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
```diff
@@ -90,13 +95,25 @@
...
    (let [environment-class-info (:environment-class-info-tags
                                  (tk-services/service-context this))]
-     (swap! environment-class-info assoc env-name tag)))
+     (swap! environment-class-info
+            #(let [cache-last-updated
+                   (get-in % [env-name :last-updated])]
+              (if (= cache-last-updated last-update-before-tag-computed)
+                (assoc % env-name
+                         (core/environment-class-info-entry tag))
+                %)))))
```
cprice404: This code is getting big enough to where I'd probably prefer that we try to push it into the core namespace, but, not a blocker for the PR since we are under a time crunch.

test/integration/puppetlabs/services/master/environment_classes_int_test.clj
```diff
@@ -6,9 +6,34 @@
...
+            [puppetlabs.trapperkeeper.services.webserver.jetty9-service :as
+             webserver]
+            [puppetlabs.trapperkeeper.services.webrouting.webrouting-service
+             :as webrouting]
+            [puppetlabs.services.puppet-admin.puppet-admin-service :as admin]
+            [puppetlabs.services.ca.certificate-authority-service :as ca]
+            [puppetlabs.trapperkeeper.services.authorization.authorization-service
+             :as authorization]
+            [puppetlabs.services.versioned-code-service.versioned-code-service
+             :as vcs])
```
cprice404: :crying_cat_face:

test/integration/puppetlabs/services/master/environment_classes_int_test.clj
```diff
@@ -522,3 +550,123 @@
...
+                "classes"
+                @class-info-atom))}]
+        (when-let [promises @wait-atom]
+          (deliver (:wait-promise promises) true)
+          @(:continue-promise promises))
+        class-info))
+    (handleRequest [_ _]
+      (JRubyPuppetResponse. 0 nil nil nil))
+    (terminate [_]
+      (log/info "Terminating Master"))))
```
cprice404: terminated

test/integration/puppetlabs/services/master/environment_classes_int_test.clj
```diff
@@ -522,3 +550,123 @@
...
+        wait-atom (atom nil)]
+    ;; This test uses a mock jruby instance function which can provide mock
+    ;; data for an environment class info query and can suspend a request
+    ;; long enough for the cached environment data to be invalidated.
+    (with-redefs
+     [jruby-internal/create-pool-instance!
+      (partial jruby-testutils/create-mock-pool-instance
+               (partial create-jruby-instance-with-mock-class-info
+                        wait-atom
+                        class-info-atom))]
```
cprice404: (yodawg) i heard you like to partial your partial

test/integration/puppetlabs/services/master/environment_classes_int_test.clj
```diff
@@ -522,3 +550,123 @@
...
+                                                      initial-response-etag)]
+           (testing (str "initial request in progress while environment "
+                         "cache is invalidated contains original class info")
+             (is (= 200 (:status initial-response))
+                 (str
+                  "unexpected status code for initial response"
+                  "response: "
+                  (ks/pprint-to-string initial-response)))
+             (is (not (nil? initial-response))
+                 "no etag returned for initial response")
```
cprice404: Should this be asserting that `initial-response-etag` is not nil?  As opposed to `initial-response`?

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
```diff
@@ -289,3 +289,128 @@
...
+             ;; production-first-updated) - in order to validate that what is
+             ;; already in the cache will just be retained.  This simulates
+             ;; what would happen if the tag had been updated again in
+             ;; between the prior 'get...last-updated' call and the
+             ;; next 'set...tag' call.
+             (jruby-protocol/set-environment-class-info-tag!
+              service
+              "production"
+              "89abprod"
+              (dec production-first-updated))
```
cprice404: Do we expect this to ever happen in prod?  I understand the case where a `mark-expired` causes the tag to be nil'd, but hadnt't really thought through any cases where the timestamp race comes into play with a non-nil tag?

cprice404:
+1 other than the one assertion that i think isn't quite right.

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
camlow325: I'll move the function arg into a new function in the core namespace.  Shouldn't be much work.

test/integration/puppetlabs/services/master/environment_classes_int_test.clj
camlow325: Yeah, I hate doing this too.  We did something like this with one of the tk-auth integration tests a while back, too.  If we're seeing more of a need to do testing where we're using most of the normal service stack but with a mocked version of the jruby-puppet-service, maybe creating a `with-puppetserver-and-mocked-jruby-service` macro would be a good thing to create at some point?  Would be nice to avoid having to proliferate knowledge of what the full service stack is out to lots of different integration test namespaces.

test/integration/puppetlabs/services/master/environment_classes_int_test.clj
camlow325: Yeah, copypasta error I think.  Good catch, thanks.  Will fix.

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
camlow325: No, this was totally contrived.  I'll re-write this bit to do a `mark-expired` call before the second `set-environment-class-info-tag!` call and pass the original `production-first-updated` into that second call.  That's the real case that I'd expect to see in production anyway.

test/integration/puppetlabs/services/master/environment_classes_int_test.clj
cprice404: good idea :put_litter_in_its_place:

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
cprice404: ok.  probably worth timeboxing that, I wouldn't say this should be a blocker, and would like to get this merged in the next hour or so.

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
camlow325: I moved the function in 520fced.

test/integration/puppetlabs/services/master/environment_classes_int_test.clj
camlow325: Fixed this in 7b733ee.

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
camlow325: I reworked this block to do a `mark-expired` call in between the get and set.  The new test is in 7b733ee.

camlow325:
@cprice404 I think I hit all of your feedback so far.  Let me know if you see anything else that you'd like to see changed.

cprice404:
@rlinehan it might be worth having you review this post-facto in case I missed anything and/or just to get yourself a bit more familiarity with this part of the code.  I went ahead and merged it because I wanted to try to start pushing builds up today.

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
```diff
@@ -289,3 +289,128 @@
...
+               (is (= production-third-update
+                      (jruby-protocol/get-environment-class-info-tag-last-updated
+                       service
+                       "production"))))))
+         (testing "when an individual environment is marked expired"
+           (jruby-protocol/mark-environment-expired! service "production")
+           (is (nil? (jruby-protocol/get-environment-class-info-tag
+                      service
+                      "production")))
+           (is (not= production-first-update
```
rlinehan: I find it confusing that this is `(not= production-first-update` rather than `(not = production-third-update` because that's the last update timestamp we have, although I guess that's because `production-third-update` isn't in scope here. It might be less confusing to get another `production-<x>-update` after line 379 here to compare against instead?

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
```diff
@@ -289,3 +289,128 @@
...
+           (is (= test-first-update
+                  (jruby-protocol/get-environment-class-info-tag-last-updated
+                   service
+                   "test"))))
+         (testing "when all environments are marked expired"
+           (jruby-protocol/set-environment-class-info-tag! service
+                                                           "production"
+                                                           "8910prod"
+                                                           nil)
+           (let [production-third-update
```
rlinehan: There's already a `production-third-update` at line 361. It's not in scope here, but this is rather confusing.

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
```diff
@@ -289,3 +289,128 @@
...
+                   (jruby-protocol/get-environment-class-info-tag-last-updated
+                    service
+                    "production")]
+               (is (not= production-first-update production-second-update))
+               (jruby-protocol/set-environment-class-info-tag!
+                service
+                "production"
+                "89abprod"
+                production-second-update)
+               (is (= nil (jruby-protocol/get-environment-class-info-tag
```
rlinehan: Having just reviewed this PR, I understand why this is `nil` but the last-updated timestamp is `production-third-update`. However, I think that if I were to come back to these tests in a few weeks without that context I might be very confused by these assertions. It might help to add some other `(testing` blocks or comments to add description.

rlinehan:
Retroactive :+1:. I left a few comments about the `environment-class-info-tags` unit test, because I found that pretty confusing to read through, and I agree that creating a `with-puppetserver-and-mocked-jruby-service` macro at some point might be really nice, but those are both small/nitpicky things and otherwise this looks good.

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
camlow325: Yeah, thanks, I agree.  I ended up reworking this in #930 to get the last updated tag first before doing the set and used a more generic name, `production-update`.  Hopefully, that will read a little better.

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
camlow325: Yeah, that's a good point.  I'll add something like this onto #930.

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
camlow325: Thanks, I'll add something onto #930 to cover this.  I think I'll just do it as an assertion message rather than a sub-testing block since there are already some nested testing blocks above this one and doing another level might make this harder to read in the output - depending upon your test runner, I suppose.

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
camlow325: I reworked the names and variable assignments a bit in [1bbc78a](https://github.com/camlow325/puppet-server/commit/1bbc78a084de95abbea88be960af250e2c652956) in #930.

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
camlow325: I reworked the names and variable assignments a bit in [1bbc78a](https://github.com/camlow325/puppet-server/commit/1bbc78a084de95abbea88be960af250e2c652956) in #930.

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
camlow325: Added some info to the messages for assertions related to this one in [1bbc78a](https://github.com/camlow325/puppet-server/commit/1bbc78a084de95abbea88be960af250e2c652956) in #930.

PR #927, opened 2016-02-26
(SERVER-1148) Add timed waits and instrumentation to jruby-pool-int-tests

camlow325:
This PR adds some upper bounds on the maximum number of iterations /
wait time for specific operations to complete while running the
jruby-pool-int-tests.  The bounded operations include the maximum wait
time for a new pool to be initialized, instances to be borrowed, pool
flushes to complete, and for the puppetserver service stack to be
stopped.

This commit adds some extra instrumentation to the
`test-restart-comes-back` test to hopefully help with identifying why
the test has failed when it intermittently does.

### DIFF
--- a/test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
+++ b/test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
@@ -341,7 +341,7 @@
        (tk-internal/restart-tk-apps [app])
        (let [start (System/currentTimeMillis)]
          (while (and (not= (count @call-seq) 5)
-                     (< (- (System/currentTimeMillis) start) 90000))
+                     (< (- (System/currentTimeMillis) start) 300000))
            (Thread/yield)))
        (let [shutdown-service (tk-app/get-service app :ShutdownService)]
          (is (nil? (tk-internal/get-shutdown-reason shutdown-service))
--- a/test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
+++ b/test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
@@ -39,6 +39,12 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Utilities
 
+(def default-borrow-timeout 180000)
+
+(defn timed-deref
+  [ref]
+  (deref ref 120000 :timed-out))
+
 (def script-to-check-if-constant-is-defined
   "! $instance_id.nil?")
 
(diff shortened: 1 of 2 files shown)


test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -296,16 +328,27 @@
                           (start [this context] (lc-fn context :start-bonus-service))
                           (stop [this context] (lc-fn context :stop-bonus-service))
                           (bonus-service-fn [this] (lc-fn nil :bonus-service-fn)))]
+      (fs/delete debug-log)
       (bootstrap/with-puppetserver-running-with-services
        app
        (conj (tk-bootstrap/parse-bootstrap-config! bootstrap/dev-bootstrap-file) bonus-service)
-       {:jruby-puppet {:max-active-instances 1}}
+       {:global {:logging-config
```
camlow325: We're speculating that the interesting data that we'd need to see to figure out why the `@call-seq` assertion fails is buried in some of the debug log output in Trapperkeeper but we can't get at that data by default since we normally suppress debug logging for test runs in order to obey the almighty Wolfe Principle™.  The code here overrides the default logback config with one that cranks up the log level to debug but with an appender that dumps everything into a file.  When the `@call-seq` assertion fails, the full contents of the log file are dumped in the assertion message.  This should hopefully keep us in good graces per the Principle but give us a lot more data to look at when the assertion fires.  Hopefully this can be undone later when we figure out the root cause for why this test has periodically been failing.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -368,4 +416,4 @@
          ;; JRubyPuppetService is already stopped.
          (swap! app-context assoc-in [:service-contexts :JRubyPuppetService] {})
          (tk-internal/run-lifecycle-fn! app-context tk-services/init "init" :JRubyPuppetService jruby-service)
-         (tk-internal/run-lifecycle-fn! app-context tk-services/start "start" :JRubyPuppetService jruby-service))))))
\ No newline at end of file
+         (tk-internal/run-lifecycle-fn! app-context tk-services/start "start" :JRubyPuppetService jruby-service))))))
```
camlow325: ~ mccuneprinciple

camlow325:
I'm actually hoping this PR fails Travis since it's trying to get more information about other failures that we've seen intermittently already.  Will try to run this at least a couple of times through Travis before we might try to merge.

src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
```diff
@@ -161,7 +161,8 @@
   (let [{:keys [config pool-state]} pool-context
         new-pool-state (jruby-internal/create-pool-from-config config)
         old-pool @pool-state]
-    (swap-and-drain-pool! pool-context old-pool new-pool-state true)))
+    (swap-and-drain-pool! pool-context old-pool new-pool-state true)
+    (println "*** done flushing and repopulating pool")))
```
rlinehan: It seems not-ideal to have this in our production code?

rlinehan: Nevermind, should have read the commit message before commenting on this. (facepalm)

camlow325: Definitely not.  :smile:   I'm going to prune off this last "println" commit from the PR before we consider merging this.  I just put this in temporarily to see if I could get the tests to fail in Travis with some more insight around where the fail is occurring.

rlinehan: :+1: sounds good

rlinehan:
Travis finally failed! (not something you often get to say on a PR) And looks like it has good output, so I'm :+1: on this PR.

camlow325:
In 5fd6892, I bumped the `test-restart-comes-back` restart timeout from 90 seconds to 5 minutes.  This test failed on two OS cells for the Travis run against the prior commit, 93be143 - https://travis-ci.org/puppetlabs/puppet-server/builds/112915228.  In each case, the test appeared to fail because the initialization of a JRubyPuppet pool instance was started but did not complete before the timeout had fired.  It may be that there is a real problem beyond this just being slow sometimes in Travis.  Hopefully any further failures that we would see with a larger timeout would make it more obvious that there is a deeper problem that we need to look into.

PR #928, opened 2016-02-26
Mergedown master -> stable

nathanielksmith:
This mergedown applied cleanly and the tests are running green locally.


KevinCorcoran:
project.clj has
```clj
(def ps-version "2.2.2-master-SNAPSHOT")
```
I think that should actually be `2.3.0-stable-SNAPSHOT`.  Does that seem right to you, @camlow325 ?

camlow325:
@KevinCorcoran It definitely makes sense to me to have `stable` in the version name instead of `master`.  Keeping it at `2.2.2-stable-SNAPSHOT` seems like the right thing to do in this PR.  The theory was that we wouldn't bump any X/Y/Z version numbers up until after we do the next release, which we wouldn't be doing quite yet.  After we would do the `2.3.0` release off of this branch, I think it would go next go up to `2.3.1-stable-SNAPSHOT`, though.

nathanielksmith:
oops, yes, i forgot to do the version bump before merging down

KevinCorcoran:
@camlow325 That's fine, but what's the reason for keeping it at 2.2.2 instead of 2.3.0 right now?

camlow325:
@KevinCorcoran Mostly I was thinking we were interested in not bumping any version numbers until we were essentially at the very last possible point for the release - like essentially in the PR we're merging just prior to tagging the release itself.  @cprice404, does that sound right to you?

Arguably, we're already at the point of no return for 2.3.0, I suppose, though.  So if you want to call it 2.3.0 right now, I'm okay with that.

KevinCorcoran:
:shipit: I guess, can't really review this diff.  If there we no merge conflicts, must be good, right?  :game_die:

haus:
:+1: this has no diff off 8f2419ec53fcdf68d58dc40158831c6bba34798c except the ps-version and ezbake version. will need to be done again to pick up recent test fixes.

haus:
(looks like an ezbake version bump was in stable and not master)

camlow325:
Yep, I see the same as @haus.  Only diff was the expected version string in the project.clj - "master" vs. "stable" and older ezbake version.

PR #930, opened 2016-02-29
(SERVER-1183) Add entry to env class cache if not present during get

camlow325:
This commit adds some logic behind the
`get-environment-class-info-tag-last-updated!` call in the
jruby-puppet-service to update the environment class info cache with a
new entry for the environment being queried in the event that no prior
entry existed for that environment.  This allows any subsequent
`mark-all-environments-expired!` call that may be made to the service to
update the `last-updated` date in the cache for the environment and,
therefore, for the logic added in SERVER-1130 to avoid updating the
cache with a stale tag to be used for the case where the get precedes a
`mark-all-environments-expired!` call.

This PR also refactors some of the logic in the jruby-puppet-service
namespace related to environment class cache maintenance down into
the jruby-puppet-core namespace.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
@@ -417,8 +417,14 @@
 
 (schema/defn ^:always-validate
   mark-environment-expired!
+  "Mark the environment-registry entry for each JRubyPuppet instance and the
+  environment's entry in the environment class info cache as expired."
   [context :- jruby-schemas/PoolContext
-   env-name :- schema/Str]
+   env-name :- schema/Str
+   environment-class-info-cache :- (schema/atom EnvironmentClassInfoCache)]
+  (swap! environment-class-info-cache
+         environment-class-info-cache-with-invalidated-entry
+         env-name)
   (doseq [jruby-instance (registered-instances context)]
     (-> jruby-instance
         :environment-registry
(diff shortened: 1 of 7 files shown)


camlow325:
@cprice404 Here's the PR for the new race condition that I mentioned in standup this morning.

cprice404:
The final 4 commits look good.  The first commit is confusing to me; it seems like it would be preferable to avoid creating synthetic cache entries and potentially using fabricated timestamps.  It seems like having the `nil` cache entries ought to be able to effectively communicate the state of the cache, even for this race, if we just had a bit more logic that checked for that `nil` case when we were updating the cache with a newly computed value?  Perhaps I'm missing something.

rlinehan:
I'm good on this aside from the issue of the manipulation of timestamps.

camlow325:
There was an exchange after the original commit in this PR with some concerns about the approach of continuing to use a timestamp in the environment class cache to track when the entry was invalidated but with some adjustments to the timestamp for cases where multiple invalidations could occur within a very short period of time.  Unfortunately, that comment thread fell off of the PR when I rebased this PR but the original exchange for posterity is still available here - https://github.com/camlow325/puppet-server/commit/235c0b90f1186961c85da92a33301b75bbfcf3d6#commitcomment-16396929.

Ultimately, I did end up reworking the cache implementation to use an integer that gets incremented on each invalidation - and dropped all of the original references to the timestamp.  The new work is at e2a1749.

camlow325:
In the time since this PR was originally issued, the puppet-server#master branch has been merged down to the puppet-server#stable branch.  In hopes of getting this merged into the stable branch, I'm going to close this PR and put up a replacement PR against the stable branch.

camlow325:
Put up #944 to replace this PR.

PR #931, opened 2016-02-29
(PDB-2329) Shared storage for terminii

mullr:
(This goes with https://github.com/puppetlabs/puppetdb/pull/1875)

In order to support high availability, the PuppetDB terminus needs
to implement sticky failover. That is, after failing over to another
PuppetDB instance, it should keep using that instance for future
requests as long as possible. This is tricky when running in
PuppetServer, since it hosts multiple JRuby instances.

This adds a shared java.util.concurrent.ConcurrentHashMap instance as
the :shared-terminus-state entry of the jruby pool context. It is passed
to each JRuby instance when it is constructed and passed into the ruby
code via the puppet_server_config map (in the shared_terminus_state
key). Finally, the Puppet::Server::Config initialize method probes for
the existence of well-known class from the PuppetDB
terminus (Puppet::Util::Puppetdb::SharedTerminusState). If this class
exists, the map is injected into via its shared_terminus_state property.

In a more perfect world, there would be a uniform terminus API by which
this map could be passed from puppet-server to the terminus.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
@@ -12,7 +12,8 @@
             [clojure.tools.logging :as log])
   (:import (puppetlabs.services.jruby.jruby_puppet_schemas JRubyPuppetInstance)
            (com.puppetlabs.puppetserver PuppetProfiler)
-           (clojure.lang IFn)))
+           (clojure.lang IFn)
+           (java.util.concurrent ConcurrentHashMap)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Constants
@@ -227,7 +228,8 @@
    ;; For an explanation of why we need a separate agent for the `flush-instance`,
    ;; see the comments in jruby-puppet-agents/send-flush-instance
    :flush-instance-agent  (jruby-agents/pool-agent agent-shutdown-fn)
-   :pool-state            (atom (jruby-internal/create-pool-from-config config))})
+   :pool-state            (atom (jruby-internal/create-pool-from-config config))
+   :shared-terminus-state (ConcurrentHashMap.)})
 
 (schema/defn ^:always-validate
   free-instance-count
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
@@ -190,7 +190,8 @@
    id       :- schema/Int
    config   :- jruby-schemas/JRubyPuppetConfig
    flush-instance-fn :- IFn
-   profiler :- (schema/maybe PuppetProfiler)]
+   profiler :- (schema/maybe PuppetProfiler)
+   shared-terminus-state :- (schema/maybe java.util.Map)]
   (let [{:keys [ruby-load-path gem-home compile-mode
                 http-client-ssl-protocols http-client-cipher-suites
                 http-client-connect-timeout-milliseconds
@@ -219,6 +220,7 @@
       (.put puppet-server-config "http_idle_timeout_milliseconds"
         http-client-idle-timeout-milliseconds)
       (.put puppet-server-config "use_legacy_auth_conf" use-legacy-auth-conf)
+      (.put puppet-server-config "shared_terminus_state" shared-terminus-state)
       (let [instance (jruby-schemas/map->JRubyPuppetInstance
                        {:pool                 pool
                         :id                   id
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
@@ -5,7 +5,8 @@
            (com.puppetlabs.puppetserver PuppetProfiler JRubyPuppet EnvironmentRegistry)
            (com.puppetlabs.puppetserver.pool LockablePool)
            (org.jruby Main Main$Status RubyInstanceConfig)
-           (com.puppetlabs.puppetserver.jruby ScriptingContainer)))
+           (com.puppetlabs.puppetserver.jruby ScriptingContainer)
+           (java.util Map)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Schemas
@@ -137,7 +138,8 @@
    :profiler              (schema/maybe PuppetProfiler)
    :pool-agent            JRubyPoolAgent
    :flush-instance-agent  JRubyPoolAgent
-   :pool-state            PoolStateContainer})
+   :pool-state            PoolStateContainer
+   :shared-terminus-state (schema/maybe Map)})
 
 (def JRubyInstanceState
   "State metadata for an individual JRubyPuppet instance"
--- a/src/ruby/puppet-server-lib/puppet/server/config.rb
+++ b/src/ruby/puppet-server-lib/puppet/server/config.rb
@@ -25,6 +25,15 @@ class Puppet::Server::Config
       Puppet::Util::Profiler.add_profiler(@profiler)
     end
 
+    if puppet_server_config["shared_terminus_state"]
+      begin
+        require 'puppet/util/puppetdb/shared_terminus_state'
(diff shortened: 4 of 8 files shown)


cprice404:
NOTE to reviewers: this needs to go into a post-2.3.0 / post-Burnside branch.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -227,7 +227,8 @@
    ;; For an explanation of why we need a separate agent for the `flush-instance`,
    ;; see the comments in jruby-puppet-agents/send-flush-instance
    :flush-instance-agent  (jruby-agents/pool-agent agent-shutdown-fn)
-   :pool-state            (atom (jruby-internal/create-pool-from-config config))})
+   :pool-state            (atom (jruby-internal/create-pool-from-config config))
+   :shared-terminus-state (java.util.concurrent.ConcurrentHashMap.)})
```
cprice404: nitpick: would prefer an import on this class, rather than the fully qualified reference here, just for consistency

src/ruby/puppet-server-lib/puppet/server/config.rb
```diff
@@ -25,6 +25,15 @@ def self.initialize_puppet_server(puppet_server_config)
       Puppet::Util::Profiler.add_profiler(@profiler)
     end
 
+    if puppet_server_config["shared_terminus_state"]
+      begin
+        require 'puppet/util/puppetdb/shared_terminus_state'
+        Puppet::Util::Puppetdb::SharedTerminusState.shared_terminus_state = puppet_server_config["shared_terminus_state"]
```
cprice404: This seems fine for now; if we ever do use this for another terminus, this design would give them all access to one another's state, so it might be nice if we could index into the appropriate terminus key in the map before we pass anything in to the terminus.  But there's a lot of speculation involved in that statement, so I think I'm fine with keeping it like this unless you have any ideas on that front that you think are worth the effort.

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -137,7 +137,8 @@
    :profiler              (schema/maybe PuppetProfiler)
    :pool-agent            JRubyPoolAgent
    :flush-instance-agent  JRubyPoolAgent
-   :pool-state            PoolStateContainer})
+   :pool-state            PoolStateContainer
+   :shared-terminus-state (schema/maybe java.util.Map)})
```
cprice404: same comment re: import here

test/unit/puppetlabs/services/jruby/jruby_testutils.clj
```diff
@@ -113,7 +113,8 @@
     id :- schema/Int
     config :- jruby-schemas/JRubyPuppetConfig
     flush-instance-fn :- IFn
-    _ :- (schema/maybe PuppetProfiler)]
+    _ :- (schema/maybe PuppetProfiler)
+    _ :- (schema/maybe java.util.Map)]
```
cprice404: same comment re: import here

cprice404:
The travis failures look like they are probably legit; there is a place in our testutils where we mock one of those fns that you changed the signatures of and it looks like maybe there are some tests that weren't refactored accordingly?  Let us know if you need help on that, that pattern is unfortunate.

I'm basically +1 on this with minor nitpicks and the one question about isolating the termini.

I would like for @camlow325 to review it before we merge, but it'll probably be a day or two before he has bandwidth.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -190,7 +190,8 @@
    id       :- schema/Int
    config   :- jruby-schemas/JRubyPuppetConfig
    flush-instance-fn :- IFn
-   profiler :- (schema/maybe PuppetProfiler)]
+   profiler :- (schema/maybe PuppetProfiler)
+   shared-terminus-state :- (schema/maybe java.util.Map)]
```
camlow325: Nitpick - could add a `java.util` import and just refer to `Map` here.  I think we mostly have been trying to refer to the most immediate class name in code.

src/ruby/puppet-server-lib/puppet/server/config.rb
camlow325: I like @cprice404's idea about making this a map of state maps.  If we keep it PuppetDB specific for now, I'd be tempted to just call it `:puppetdb-shared-terminus-state` instead elsewhere in the Clojure code.  But either way, fine with it as it is for now.

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -137,7 +138,8 @@
    :profiler              (schema/maybe PuppetProfiler)
    :pool-agent            JRubyPoolAgent
    :flush-instance-agent  JRubyPoolAgent
-   :pool-state            PoolStateContainer})
+   :pool-state            PoolStateContainer
+   :shared-terminus-state (schema/maybe Map)})
```
camlow325: @cprice404 Thinking ahead to the work that we're planning on undertaking soon around the JRuby pool / library refactor, I wonder if you think this is the best place to store this kind of state.  Maybe it would be better to have this as a top-level piece of data in whatever consuming `puppet` Trapperkeeper service rather than putting it right into the pool structure?  Dunno.  Suppose we can move this around later on as we figure out exactly what the new APIs will look like so long as we'd preserve PuppetDB's ability to share state across JRuby instances using the same Ruby-level API.

src/ruby/puppet-server-lib/puppet/server/config.rb
cprice404: @camlow325  that's a good thought, and a good thing to keep in mind, but yeah, I don't think we can answer that question now, and it shouldn't be too big of a deal to move it around later.  At this point I don't think it's worth holding up @mullr any longer for that particular concern.

camlow325:
I'm a little nervous - from a very naive perspective - about whether we might see any problems with constructing an object from within Ruby via one ScriptingContainer instance, stuffing that into the shared state map, and using that object from Ruby in another ScriptingContainer.  Suppose that will be fine so long as the objects being created are primitive Java ones?  But if we were talking about creating objects from custom Ruby-typed classes, I'm not too clear about how well JRuby will handle using those objects across the SINGLETHREAD JRuby containers that we're currently using.

One thing that might be a good test if you haven't tried already is seeing if after creating an object in one ScriptingContainer that the object is still valid / proper when used from other ScriptingContainers after you've let the original ScriptingContainer in which that object was created be destroyed in-process - via a JRuby pool flush or `max-requests-per-instance` be hit.

@mullr, I suppose you already have enough practical experience with this to have rooted out any such problems around this concern?

camlow325:
Other than the concern I mentioned about whether we're going to run into any run-time issues around sharing state across JRuby ScriptingContainers, I'm good with this.  @mullr and @cprice404, if you both say at this point that you're good with letting this be merged as it stands, I'm good pushing the button.

cprice404:
@camlow325 I think the concern that you mentioned is legitimate and that someone should definitely at least do some manual testing on that if we haven't already.  I actually wouldn't mind seeing if we can figure out a way to codify that in a test, because even if it works well right now it could be subject to issues when we upgrade to the next major version of JRuby.

@mullr if you are interested we'd be happy to have one of us pair with you on poking around this concern whenever you have time.

kbarber:
I've manually tested the code combined with the puppetdb patch and it does what it says on the tin it seems, using many jruby instances. I fear some bits aren't codified in tests also. I think, that a test over here seems wiser than doing it in pdb, as the complexity in pdb would force us to write acceptance tests I think. How viable is an integration/clojure.test possible on this side to prove this works? Or are we just being paranoid, and should we just trust jruby to test this functionality?

cprice404:
I think a test is viable, and IMO worthwhile.

camlow325:
I think the key thing from the comment we made earlier is not just that we test with sharing across many JRuby instances but that we also test with having the JRuby instance to which the shared state was initially added be disposed of - e.g., via having the JRuby instance be flushed when the `max-requests-per-instance` for the instance is reached.  Not sure if that was covered in your bench testing.

I'd definitely agree with @cprice404 that we should implement an automated test for this.  I don't have a great deal of comfort about the extent to which the JRuby project has been testing multiple ScriptingContainers within the same process, especially where containers are created / destroyed throughout runtime.

cprice404:
@mullr based on your comments on recent e-mail threads it sounds like this idea is abandoned for now?  Should I just close this?

mullr:
Not completely abandoned, but not using it in the near future at least. closing.

PR #933, opened 2016-03-02
(SERVER-1185) Add enable setting for environment class cache

camlow325:
This commit adds a Trapperkeeper-level configuration setting for use in
enabling/disabling the environment class cache.  When set to "false"
(disabled), an Etag is no longer computed/returned with the payload for
a environment_classes API GET request.  When set to "true" (enabled),
the Etag will be computed and returned with the GET request payload as
was true in prior commits.

### DIFF
--- a/dev/puppet-server.conf.sample
+++ b/dev/puppet-server.conf.sample
@@ -115,3 +115,8 @@ authorization: {
         }
     ]
 }
+
+# general puppetserver settings
+puppetserver: {
+    environment-class-cache-enabled: true
+}
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -262,27 +262,31 @@
    environment :- schema/Str
    jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)
    request-tag :- (schema/maybe String)
-   last-updated :- (schema/maybe schema/Int)]
+   last-updated :- (schema/maybe schema/Int)
+   environment-class-cache-enabled :- schema/Bool]
   (let [info-for-json (class-info-from-jruby->class-info-for-json
                        info-from-jruby
-                       environment)
-        info-as-json (cheshire/generate-string info-for-json)
-        parsed-tag (ks/utf8-string->sha1 info-as-json)]
-    (jruby-protocol/set-environment-class-info-tag!
-     jruby-service
-     environment
-     parsed-tag
-     last-updated)
-    (if (= parsed-tag request-tag)
-      (not-modified-response parsed-tag)
-      (-> (response-with-etag info-as-json parsed-tag)
-          (rr/content-type "application/json")))))
+                       environment)]
+    (if environment-class-cache-enabled
+      (let [info-as-json (cheshire/generate-string info-for-json)
+            parsed-tag (ks/utf8-string->sha1 info-as-json)]
+        (jruby-protocol/set-environment-class-info-tag!
+         jruby-service
+         environment
+         parsed-tag
+         last-updated)
+        (if (= parsed-tag request-tag)
+          (not-modified-response parsed-tag)
+          (-> (response-with-etag info-as-json parsed-tag)
+              (rr/content-type "application/json"))))
+      (ringutils/json-response info-for-json))))
 
 (schema/defn ^:always-validate
   environment-class-info-fn :- IFn
   "Middleware function for constructing a Ring response from an incoming
   request for environment_classes information."
-  [jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)]
+  [jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)
+   environment-class-cache-enabled :- schema/Bool]
   (fn [request]
     (let [environment (jruby-request/get-environment-from-request request)
           last-updated (jruby-protocol/get-environment-class-info-tag-last-updated
@@ -297,7 +301,8 @@
                                      environment
                                      jruby-service
                                      (if-none-match-from-request request)
-                                     last-updated)
+                                     last-updated
+                                     environment-class-cache-enabled)
         (rr/not-found (str "Could not find environment '" environment "'"))))))
 
 (schema/defn ^:always-validate
@@ -324,9 +329,11 @@
 (schema/defn ^:always-validate
   environment-class-handler :- IFn
   "Handler for processing an incoming environment_classes Ring request"
-  [jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)]
+  [jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)
+   environment-class-cache-enabled :- schema/Bool]
   (->
-   (environment-class-info-fn jruby-service)
+   (environment-class-info-fn jruby-service
+                              environment-class-cache-enabled)
    (jruby-request/wrap-with-jruby-instance jruby-service)
    (wrap-with-etag-check jruby-service)
    jruby-request/wrap-with-environment-validation
@@ -414,9 +421,11 @@
   v3-clojure-routes :- bidi-schema/RoutePair
   "v3 route tree for the clojure side of the master service."
   [jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)
-   get-code-content-fn :- IFn]
+   get-code-content-fn :- IFn
+   environment-class-cache-enabled :- schema/Bool]
   (let [environment-class-handler
-        (environment-class-handler jruby-service)
+        (environment-class-handler jruby-service
+                                   environment-class-cache-enabled)
         static-file-content-handler
         (static-file-content-request-handler get-code-content-fn)]
     (comidi/routes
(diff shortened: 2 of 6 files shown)


camlow325:
I'm planning on adding the docs for this new setting as part of the work I already have underway separately for SERVER-1114.  Would like to get this in ahead of those docs being polished up.

cprice404:
+1

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -262,27 +262,31 @@
    environment :- schema/Str
    jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)
    request-tag :- (schema/maybe String)
-   last-updated :- (schema/maybe schema/Int)]
+   last-updated :- (schema/maybe schema/Int)
+   environment-class-cache-enabled :- schema/Bool]
```
rlinehan: It might be good to update the docstring for this function to mention this setting and what it does to the response? (not blocker for merge)

dev/puppet-server.conf.sample
```diff
@@ -115,3 +115,8 @@ authorization: {
         }
     ]
 }
+
+# general puppetserver settings
+puppetserver: {
+    environment-class-cache-enabled: true
+}
```
rlinehan: I'm not a UX person, but I feel like if I were a user I would look at this setting and go "why the hell is that in `puppetserver`? why is there another section than `jruby-puppet`?"

I feel like our users most likely don't understand or care about the distinctions between our trapperkeeper services. I've always kind of looked at the `jruby-puppet` section as the `puppetserver` section, just with a weird name.

I guess from our (developer) perspective, putting this setting in the `jruby-puppet` section makes the code a little bit messier, but I would still be in favor of prioritizing the user experience here.

cprice404: We actually talked through this and chose this section *because* of UX :)  The thinking was:

"Why the hell do I need to know anything about jruby in order to tell you whether I want to cache environments or not?  What does JRuby have to do with caching or environments?"

Also, we're thinking that we may need to bust up that jruby section in some way before long, because of upcoming refactors to make the jruby service shareable for non-puppet use cases...

KevinCorcoran: also worth noting that the new `code-*-command`s are underneath `puppetserver`

rlinehan: Huh, okay. I guess I've just sort of always thought that the name of that section is weird and renamed it in my head to "puppetserver", and so I find having a separate section more confusing than having this setting in a JRuby section. I don't think at all that I would think JRuby had nothing to do with caching of environments.

Have you asked anyone in CS about this? If you have then I'll shut up. :)

rlinehan: Ah, okay - didn't see @KevinCorcoran's comment before I made the above comment. If we do already have settings in this section, then I'm cool with this I guess. Might be worth adding those settings commented out to the example config file.

test/integration/puppetlabs/services/master/environment_classes_int_test.clj
```diff
@@ -92,9 +94,47 @@
...
 
-(deftest ^:integration environment-classes-integration-test
+(deftest ^:integration environment-classes-integration-cache-disabled-test
+  (testing "when environment classes cache is disabled for a class request"
+    (bootstrap/with-puppetserver-running-with-config
+     app
+     (-> {:jruby-puppet {:max-active-instances 1}}
+         (bootstrap/load-dev-config-with-overrides)
+         (ks/dissoc-in [:puppetserver
+                        :environment-class-cache-enabled]))
```
rlinehan: I found this line really confusing and didn't understand why it was here until I traced back what `load-dev-config-with-overrides` did... which is copy over the dev sample config file. Since the default setting for this is `false` I feel like it might be less confusing to have the sample config file set it to false, or not set it at all.

rlinehan:
:+1: left a couple small nitpicks, but those could be addressed later if you agree with them. :ship:

dev/puppet-server.conf.sample
camlow325: @rlinehan Which config file were you thinking we would add the commented out setting to?  Even though it runs counter to what we've done in the past, I think we don't want to add new commented out settings to any of the ezbake config files (ones built into the packages) because doing so has caused formatting issues when used together with Puppet hocon module management (duplicate references to the same setting).  If there's another dev-related config file where it might make sense to have it enabled by default, I could add the new setting to that in a follow-on PR.

test/integration/puppetlabs/services/master/environment_classes_int_test.clj
camlow325: For any test that expects a specific value to be used for a setting - in this case that the `environment-class-cache-enabled` setting not be present in the configuration but the default behavior of "false" be used - I think the setting should be explicitly set within the test itself.  It would seem bad if in the future someone had changed the setting in the dev sample config file without understanding the implicit behavior that this test was relying upon but having the test still pass despite an underlying problem in the implementation being present.  Making the setting explicit in the test that depends upon its value seems like the best safeguard to me.

Note that I followed this line of reasoning with the corresponding `environment-classes-integration-cache-enabled-test` as well where I explicitly set a value of `true` for the setting even though the dev sample config used the same value.

src/clj/puppetlabs/services/master/master_core.clj
camlow325: Yeah, that's a good idea, thanks.  I didn't have anything about the other parameters in the docstring either but that may be helpful to include as well.  I'll keep that in mind as I rework the code for #930.

dev/puppet-server.conf.sample
rlinehan: I was thinking of adding in commented out settings for the `versioned-code` block (which I think is what Kevin was referring to as being in a `puppetserver` block, except they turn out to actually not be) to this sample config file, which is our main "dev" example one. I don't think it's packaged, and we already have a lot of comments in it, so that seems okay?

test/integration/puppetlabs/services/master/environment_classes_int_test.clj
rlinehan: I might have commented in the wrong place/not fully explained what I was thinking. I think that the issue I'm really commenting on is that it seems weird to have this set to true in `dev/puppet-server.conf.sample` when it is by default `false`. I agree with everything you've said above, but it seems in that case that it should be set to `false` and explicitly set to `true` not only in the `environment-classes-integration-cache-enabled-test` but also in the other tests that rely on this setting being set to true, such as the `class-info-updated-after-cache-flush-during-prior-request` test.

rlinehan: I might have commented in the wrong place/not fully explained what I was thinking. I think that the issue I'm really commenting on is that it seems weird to have this set to true in `dev/puppet-server.conf.sample` when it is by default `false`. I agree with everything you've said above, but it seems in that case that it should be set to `false` and explicitly set to `true` not only in the `environment-classes-integration-cache-enabled-test` but also in the other tests that rely on this setting being set to true, such as the `class-info-updated-after-cache-flush-during-prior-request` test.

dev/puppet-server.conf.sample
camlow325: @rlinehan Ah, ok.  That makes sense.  Adding the commented out settings for `code-*-command` to a `versioned-code` section of the dev config seems reasonable to me.

test/integration/puppetlabs/services/master/environment_classes_int_test.clj
camlow325: @rlinehan Yeah, good points.  I should have explicitly set the value in the other tests in that namespace, too.  That was definitely an oversight.  Will include that in a later PR.

As for what the default setting in `dev/puppet-server.conf.sample` should be, I'm kind of mixed.  I generally have found it convenient to have the default value for a setting in the dev config be set to whatever "on" means for a feature that might be enabled/disabled, even if we've chosen to set the feature to "off" in shipping packages for backward compatibility.  But I suppose the argument could also be made that we should just have the dev config be as close as possible to the packaged default config - even if it means one would need to set things to "on" for development.

If you want to reset the value to the `dev` setting to "off" in a follow-up PR, I wouldn't have a problem with that.

rlinehan: I guess thinking about it more that if we have it explicit in all the tests then I don't really care what it is in the default config, and your argument about having it on for development purposes makes sense.

camlow325: Added an extra reference to the setting in [2ab8dd](https://github.com/camlow325/puppet-server/commit/2ab8ddf465e466adbd38069f1ab14855f6af0e20). in #930.

src/clj/puppetlabs/services/master/master_core.clj
camlow325: Added some more info to the docstring in [3317ef](https://github.com/camlow325/puppet-server/commit/3317ef2a5d56a0ff3eab905c7318a6a225511125) in #930.

PR #935, opened 2016-03-02
(SERVER-1084) pre-docs for initial HUP support

cprice404:
@gguillotte here's a rough cut at this stuff.

If you're good with these as 'pre-docs' then we can merge this and I'll assign the ticket back to you so that you can get them into a state you're happy with?


gguillotte:
@cprice404 :+1:

PR #936, opened 2016-03-02
(SERVER-1185) Move environment-class-cache-enabled setting

rlinehan:
Previously, the `environment-class-cache-enabled` setting was in a
`puppetserver` block. This was the only setting in this block. Ultimately, we
would like to separate out the `jruby-puppet` setting block to only contain
the jruby-specific settings and move many of the settings to a `puppetserver`
block. However, we are not at the point of doing that yet, and in the meantime
only having one setting in the `puppetserver` block seems confusing to explain
to users (why is this here when all the other seemingly-puppet server related
settings aren't here?). This commit moves this setting to the `jruby-puppet`
block for now.

I spoke with @loriland and she agreed that this seems easier to explain to users and the best decision for now, until we have a chance to audit our config files.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
@@ -92,7 +92,14 @@
 
     * :use-legacy-auth-conf - Whether to use the legacy core Puppet auth.conf
         (true) or trapperkeeper-authorization (false) to authorize requests
-        being made to core Puppet endpoints."
+        being made to core Puppet endpoints.
+
+    * environment-class-cache-enabled - Whether to use the environment class
+        cache. When disabled (set to 'false') an Etag is not computed/returned
+        with the payload for an environment_classes API GET request. When enabled
+        (set to 'true') the Etag will be computed and returned with the GET request
+        payload as was true in prior commits."
+
   {:ruby-load-path [schema/Str]
    :gem-home schema/Str
    :compile-mode SupportedJRubyCompileModes
@@ -108,7 +115,8 @@
    :borrow-timeout schema/Int
    :max-active-instances schema/Int
    :max-requests-per-instance schema/Int
-   :use-legacy-auth-conf schema/Bool})
+   :use-legacy-auth-conf schema/Bool
+   (schema/optional-key :environment-class-cache-enabled) schema/Bool})
 
 (def JRubyPoolAgent
   "An agent configured for use in managing JRuby pools"
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
@@ -92,14 +92,7 @@
 
     * :use-legacy-auth-conf - Whether to use the legacy core Puppet auth.conf
         (true) or trapperkeeper-authorization (false) to authorize requests
-        being made to core Puppet endpoints.
-
-    * environment-class-cache-enabled - Whether to use the environment class
-        cache. When disabled (set to 'false') an Etag is not computed/returned
-        with the payload for an environment_classes API GET request. When enabled
-        (set to 'true') the Etag will be computed and returned with the GET request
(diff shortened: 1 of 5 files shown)


src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -108,7 +115,8 @@
    :borrow-timeout schema/Int
    :max-active-instances schema/Int
    :max-requests-per-instance schema/Int
-   :use-legacy-auth-conf schema/Bool})
+   :use-legacy-auth-conf schema/Bool
+   (schema/optional-key :environment-class-cache-enabled) schema/Bool})
```
cprice404: Can we try to keep this schema consistent to where all of the keys are required, but the value is optional?

cprice404: And actually it seems like for this setting the value shouldn't be optional either; if the user didn't specify the setting, we just resolve the default before we create an instance of this schema.

cprice404:
+1 other than schema gymnastics

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
camlow325: Seems like the only place we create one of these is here - https://github.com/puppetlabs/puppet-server/blob/1673f849c286d5eaa0a46b0421d63f16de044c56/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj#L173-L196 - for use with the jruby-puppet-pooled-service.  Since the pooled service doesn't use this setting, does it make sense to even add this setting to this schema?

rlinehan: I think it needs to be in the schema, otherwise it will break the jruby-puppet service (because it will still be in that section). I guess we could dissoc it from the config in https://github.com/rlinehan/puppet-server/blob/SERVER-1185-move-setting-to-jruby-puppet/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj#L174 first?

camlow325: Ah, yep, you're right.  Suppose I'd be in favor of dissoc'ing if it doesn't otherwise make sense to pass it around elsewhere in the "run-time" config that the jruby-puppet-pooled-service uses.  @cprice404, what do you think?

camlow325:
:+1: pending the outcome of the schema thing.

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
cprice404: I'd rather dissoc than have an optional key, and I'd rather dissoc than pass it around to places where it doesn't make sense... it seems like the issue here is that we actually need two schemas now, one for what is allowed in the config file and one for what we pass around for things that are actually related to jruby.  That is the pattern that I would prefer that we strive for in the future, personally, but it's probably too much work for this PR since we are on a time crunch.

rlinehan: @cprice404 I definitely agree that that's the best pattern to follow and would be the ideal here if we had time. Do we have tickets anywhere about the work that will be required for splitting out the jruby config - seems like something that would be good to add there, or just create another ticket?

cprice404: I don't know that we have anything about the config yet, I think some of that will probably just come out in the wash as we're doing the initial repo setup.  Creating a ticket and throwing it into that epic seems fine.

As far as this PR goes, dissoc seems OK and ye should feel free to merge when it's ready.

rlinehan:
I think this is going to fail travis because apparently something happened when I rebased onto master, I just can't figure out what.

rlinehan:
Should in theory pass travis now, if only travis weren't taking ages and ages and ages.

camlow325:
Tests passed for me locally so merging.

PR #939, opened 2016-03-03
(SERVER-542) Gracefully handle missing params to /certificate_status PUT

justinstoller:
Prior to this we weren't checking that `(:desired_state json-body)` returned a non-nil value. If it had returned nil our error message formatting will cause an NPE (with `(name nil)`).

We now check for nil and return an appropriate error message and code when the parameter is missing.

### DIFF
--- a/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
+++ b/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
@@ -207,13 +207,16 @@
     (when (= :put (get-in context [:request :request-method]))
       (if-let [body (get-in context [:request :body])]
         (if-let [json-body (try-to-parse body)]
-          (let [desired-state (keyword (:desired_state json-body))]
+          (if-let [desired-state (keyword (:desired_state json-body))]
             (if (schema/check ca/DesiredCertificateState desired-state)
               (malformed
                 (format
                   "State %s invalid; Must specify desired state of 'signed' or 'revoked' for host %s."
                   (name desired-state) subject))
-              [false {::json-body json-body}]))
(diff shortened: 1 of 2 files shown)


jpinsonault:
This looks good to me :+1:

KevinCorcoran:
@justinstoller A minor point of Git hygiene: it's preferable to add the test and the fix in a single commit.  It's nice to have the tests pass on all commits, because it can make things like `git bisect` to find some problem much easier if you know that the tests should pass on every commit.

justinstoller:
Oh yeah, good point. I was trying a couple of things locally to make sure I was repro/testing correctly and wanted a check point while doing so. I'll squash this and similar ones in the future!

justinstoller:
Squashed.

KevinCorcoran:
One of the three Travis test runs failed with
```
FAIL in (admin-api-flush-jruby-pool-test) (jruby_pool_int_test.clj:158)
timed out waiting for the flush to complete
expected: (true? (timed-deref flush-complete))
  actual: (not (true? :timed-out))
```
That looks unrelated to these changes.  @camlow325 is that failure known/expected/ticketed?

camlow325:
@KevinCorcoran Yeah, I've been seeing that Travis failure, too, periodically.  I added some of these assertions recently to try to help isolate cases where we've seen some of these tests hang.  There's some follow-on work that needs to be done for the still-open [SERVER-1148](https://tickets.puppetlabs.com/browse/SERVER-1148) so probably best to keep the conversation going with that one.

KevinCorcoran:
ok, thanks.

PR #941, opened 2016-03-04
(doc) Add static_file_content API and config docs.

gguillotte:
For SERVER-1108.

### DIFF
--- a/documentation/config_file_puppetserver.md
+++ b/documentation/config_file_puppetserver.md
@@ -9,6 +9,9 @@ canonical: "/puppetserver/latest/config_file_puppetserver.html"
 [cache directory]: /puppet/latest/reference/dirs_vardir.html
 [`auth.conf` documentation]: ./config_file_auth.html
 [deprecated]: ./deprecated_features.html
+[static catalogs]: /puppet/latest/reference/static_catalogs.html
+[file resource]: /puppet/latest/reference/type.html#file
+[`static_file_content`]: ./puppet-api/v3/static_file_content.html
 
 The `puppetserver.conf` file contains settings for Puppet Server software. For an overview, see [Puppet Server Configuration](./configuration.html).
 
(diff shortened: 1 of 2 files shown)


documentation/puppet-api/v3/static_file_content.md
```diff
@@ -0,0 +1,87 @@
...
+To retrieve a specific version of a file at a given environment and path, make an HTTP
+request to this endpoint with the required parameters. The `<FILE-PATH>` segment of the
+endpoint corresponds to the requested file's path, relative to the given environment's
+root directory.
+
+### Query parameters
+
+You must pass three parameters in the GET request:
+
+-   `file-path`: the path to the file, relative to the top-level directory in the
```
fpringvaldsen: This is part of the URL path, not a query parameter

fpringvaldsen:
I had one comment about the query parameters list, otherwise :+1:

documentation/config_file_puppetserver.md
```diff
@@ -27,24 +30,30 @@ The `puppetserver.conf` file contains settings for Puppet Server software. For a
...
+        If neither the `authorization-required` nor the `client-whitelist` settings are specified, Puppet Server uses the new authorization methods and [`auth.conf` documentation][] formats to access the admin API endpoints.
+
+* The `versioned-code` settings configure commands required to use [static catalogs][]:
+    * `code-id-command` contains the path to an executable script that provides a `code_id`. When compiling a static catalog, Puppet Server uses the output of this script as the catalog's `code_id` to generate metadata for a specific version of the catalog's file resources. If `code-id-command` is not set and static catalogs are enabled, the catalog's `code_id` will be null.
+    * `code-content-command` contains the path to an executable script that Puppet Server invokes when an agent makes a [`static_file_content`][] API request for the contents of a [file resource][] with a `source` attribute. If `code-content-command` is not set and static catalogs are enabled, Puppet runs on a static catalog that attempt to fetch a file resource's content will fail.
```
KevinCorcoran: might be worth stating that the value of `source` must be a `puppet:///` URL

KevinCorcoran: Puppet Server will refuse to start if `code-id-command` is defined by `code-content-command` is not (and vice-versa).

If `static_catalog = true` but neither of those commands exist, old-style / non-catalogs will be used.  This is the default behavior.  So I don't think this last sentence is appropriate.

documentation/config_file_puppetserver.md
```diff
@@ -90,6 +99,12 @@ http-client: {
...
     enabled: true
 }
-~~~
+
+# Settings related to static catalogs.
+versioned-code: {
+    code-id-command: /opt/puppetlabs/server/apps/puppetserver/code-id-command_script.sh
+    code-content-command: /opt/puppetlabs/server/apps/puppetserver/code-content-command_script.sh
+}
+```
```
KevinCorcoran: Would it be worth noting that these are hypotethical file paths for these user-created scripts?  I worry that, just glancing at this example config, a user might get the idea that those scripts are included with Puppet Server and should be that location.

documentation/puppet-api/v3/static_file_content.md
```diff
@@ -0,0 +1,86 @@
...
+
+[`code-content-command`]: https://docs.puppetlabs.com/puppetserver/latest/config_file_puppetserver.html
+[static catalog]: https://docs.puppetlabs.com/puppet/latest/reference/static_catalogs.html
+[catalog]: https://docs.puppetlabs.com/puppet/latest/reference/subsystem_catalog_compilation.html
+[file resource]: https://docs.puppetlabs.com/puppet/latest/reference/type.html#file
+[environment]: https://docs.puppetlabs.com/puppet/latest/reference/environments.html
+
+The `static_file_content` endpoint returns the standard output of a
+[`code-content-command`][] script, which should output the contents of a specific version
+of a [file resource][] that contains the `source` attribute. That source must be a file
```
KevinCorcoran: same note about `puppet:///` URLs

documentation/puppet-api/v3/static_file_content.md
```diff
@@ -0,0 +1,86 @@
...
+### Notes
+
+When requesting a file from this endpoint, Puppet Server passes the values of the
+`file-path`, `code_id`, and `environment` parameters as arguments to the
+`code-content-command` script. If the script returns an exit code of 0, Puppet Server
+returns the script's standard output, which should be the contents of the requested
+version of the file.
+
+This endpoint returns an error (status 500) if the [`code-content-command`][] setting
+is not configured on Puppet Server.
```
KevinCorcoran: :clap: this is great work, thank you @gguillotte !

documentation/config_file_puppetserver.md
gguillotte: Can do. There's no default location, right?

KevinCorcoran: correct, it's up to the user to choose it.

documentation/config_file_puppetserver.md
gguillotte: @KevinCorcoran That's interesting. There's a [test for the missing `code-content-command` scenario](https://github.com/puppetlabs/puppet-server/blob/9129923a88dd8add56c0cd9da89230aea2a73c7d/test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj#L138), but from your comment it sounds like that scenario won't be possible in a real deployment. I'll revise these.

documentation/config_file_puppetserver.md
joshcooper: Do we ever provide puppet code for managing these using the [puppetlabs-hocon](https://forge.puppetlabs.com/puppetlabs/hocon) module?

```puppet
hocon_setting { 'code-id-command':
  ensure  => present,
  path    => '/etc/puppetlabs/puppetserver/conf.d/puppetserver.conf',
  setting => 'versioned-code.code-id-command',
  value   => '/path/to/code-id.sh',
}

hocon_setting { 'code-content-command':
  ensure  => present,
  path    => '/etc/puppetlabs/puppetserver/conf.d/puppetserver.conf',
  setting => 'versioned-code.code-content-command',
  value   => '/path/to/code-content.sh',
  ]
}
```

gguillotte: We could, though that would add a lot of Puppet code to all the configuration docs that might need to be updated if the module changes how it works. Since we can't assume everyone will use a module (or our module) to manage Server's configuration files, the individual configuration file pages should probably stick to the basics.

We could add a section to the configuration page that notes the module and provides a generic example that could apply to any of the Server configuration files, but that's out of this PR's scope.

gguillotte:
Made more edits to address feedback and squashed commits.

documentation/config_file_puppetserver.md
KevinCorcoran: ah, right.  the difference here is that the test explicitly requests `/static_file_content`, whereas an agent will not do that, since there will be no `code_id`.

documentation/config_file_puppetserver.md
joshcooper: Sounds good. Btw, can we mention that these scripts need to be executable by the user that puppetserver runs as?

gguillotte: That's noted on L52:

> **Note:** The Puppet Server process must be able to execute the `code-id-command` and `code-content-command` scripts

but I can re-emphasize that in the example.

PR #942, opened 2016-03-04
(SERVER-615) Gracefully handle invalid ca_ttl settings

justinstoller:
Previously, if the ca_ttl setting was larger than a Java Integer
(2,147,483,647 -- representing ~68 years) the CA would fail to sign
certs.

This restricts the ca_ttl value to a value equivalent to 50 standard
years and adds validation of the ca_ttl setting to the service's init
phase, invalid ca_ttls will now fail on init, rather than when signing
certificates.

It also refactors existing deprecations warnings for various ca config
settings into core CA with our new validation function and adds tests
for the existing deprecation warnings.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -155,6 +155,10 @@
   "The set of SSL related files that are required on the CA."
   #{:cacert :cacrl :cakey :cert-inventory :serial})
 
+(def max-ca-ttl
+  "The longest valid duration for CA certs, in seconds. 50 standard years."
+  1576800000)
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Internal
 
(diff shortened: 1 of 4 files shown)


src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -214,7 +220,12 @@
   [ca-ttl :- schema/Int]
   (let [now        (time/now)
         not-before (time/minus now (time/days 1))
-        not-after  (time/plus now (time/seconds ca-ttl))]
+        ;; clj-time specific interval classes (eg seconds) do not
+        ;; accept inputs larger than int, so we cast to a long
+        ;; representation of millis and construct our own Period instance
+        ;; for time/plus
```
justinstoller: It seems very non-obvious to me why we're doing this. And so I felt that it needed a comment, but if felt a bit dirty adding the comment to otherwise clean code. Should I put this information in a different form or is it more obvious to folks who are more familiar with the clj ecosystem?

justinstoller:
I asked a question in an outdated diff about how to best capture knowledge about the edge case we're coding around.

I thought the addition of a function would help in that, but I'd like to know what others think. If folks like the content of the last commit I'll squash it down, if not I'll rebase it away.

justinstoller:
Also, in the [ticket](https://tickets.puppetlabs.com/browse/SERVER-615), I raised the question of whether it would be better to have a shorter (but still improbably long in tech years) ttl that doesn't require the joda gymnastics.

fpringvaldsen:
:+1:

KevinCorcoran:
I left a comment on the JIRA ticket.  I don't think this approach is the best solution to this issue.

justinstoller:
Updated to restrict valid ca_ttls to less than 50 years.

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -105,6 +105,45 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Tests
 
+(deftest validate-settings-test
+  (testing "invalid ca-ttl is rejected"
+    (let [settings (assoc
+                     (testutils/ca-settings cadir)
+                     :ca-ttl
+                     (+ max-ca-ttl 1))]
+      (is (thrown? IllegalStateException (validate-settings! settings)))))
```
KevinCorcoran: Since `IllegalStateExecption` is so generic, I think it's preferable here to use `thrown-with-msg?` and test for the specific error message we expect.  I like to avoid that sort of thing when possible, but without it here, there's no guarantee that this `IllegalStateException` is the one we're expecting.

KevinCorcoran:
:+1: This looks good except for my one nitpick about a test case.

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
justinstoller: Cool. I didin't know about `thrown-with-msg?`!

justinstoller:
Updated

PR #944, opened 2016-03-07
(SERVER-1183) Add entry to env class cache if not present during get

camlow325:
This PR is a replacement for #930, which was originally targeted at the
master branch.

This PR replaces the use of a timestamp for synchronization of tags
in the environment class info cache with an incrementing integer, called
the `cache-generation-id`.

This PR adds some logic behind the
`get-environment-class-info-tag-cache-generation-id!` call in the
jruby-puppet-service to update the environment class info cache with a
new entry for the environment being queried in the event that no prior
entry existed for that environment. This allows any subsequent
`mark-all-environments-expired!` calls that may be made to the service to
update the cache-generation-id in the cache for the environment and,
therefore, for the logic added in #926 to avoid updating the
cache with a stale tag to be used for the case where the get precedes a
`mark-all-environments-expired!` call.

This PR also refactors some of the logic in the jruby-puppet-service
namespace related to environment class cache maintenance down into
the jruby-puppet-core namespace and addresses some previously
unaddressed PR comments from #926 and #933.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
@@ -417,8 +417,14 @@
 
 (schema/defn ^:always-validate
   mark-environment-expired!
+  "Mark the environment-registry entry for each JRubyPuppet instance and the
+  environment's entry in the environment class info cache as expired."
   [context :- jruby-schemas/PoolContext
-   env-name :- schema/Str]
+   env-name :- schema/Str
+   environment-class-info-cache :- (schema/atom EnvironmentClassInfoCache)]
+  (swap! environment-class-info-cache
+         environment-class-info-cache-with-invalidated-entry
+         env-name)
   (doseq [jruby-instance (registered-instances context)]
     (-> jruby-instance
         :environment-registry
@@ -426,7 +432,13 @@
 
 (schema/defn ^:always-validate
   mark-all-environments-expired!
-  [context :- jruby-schemas/PoolContext]
+  "Mark all environment-registry entries for each JRubyPuppet instance and
+  all environment entries in the environment class info cache as expired."
+  [context :- jruby-schemas/PoolContext
+   environment-class-info-cache :- (schema/atom EnvironmentClassInfoCache)]
+  (->> invalidated-environment-class-info-entry
+       (partial ks/mapvals)
+       (swap! environment-class-info-cache))
   (doseq [jruby-instance (registered-instances context)]
     (-> jruby-instance
         :environment-registry
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
@@ -451,9 +451,8 @@
   all environment entries in the environment class info cache as expired."
   [context :- jruby-schemas/PoolContext
    environment-class-info-cache :- (schema/atom EnvironmentClassInfoCache)]
-  (->> invalidated-environment-class-info-entry
-       (partial ks/mapvals)
-       (swap! environment-class-info-cache))
+  (swap! environment-class-info-cache
+         (partial ks/mapvals invalidated-environment-class-info-entry))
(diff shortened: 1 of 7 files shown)


src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -356,34 +339,139 @@
...
+  "Return the supplied 'original-environment-class-info-entry' class info entry,
+  only updated with a cache-generation-id value that has been incremented.
+  If the supplied parameter is nil, the return value will have a tag set to
+  nil and cache-generation-id set to 1."
+  [original-environment-class-info-entry :-
+   (schema/maybe EnvironmentClassInfoCacheEntry)]
+  (if original-environment-class-info-entry
+    (update original-environment-class-info-entry
+            :cache-generation-id
+            #(if (= % Long/MAX_VALUE) Long/MIN_VALUE (inc %)))
```
camlow325: This logic may be overkill.  Seems extremely unlikely that the `:cache-generation-id` would roll around the maximum value for a `java.lang.Long` within the lifetime of one puppet-server service lifetime - 9223372036854775807.  I could simplify this back to just `inc` and not bother protecting against buffer overflow if others would prefer.

rlinehan: `9223372036854775807/(1000*60*60*24*365) = 292471208.68`... so it would take 292 million years to reach this if it were being hit every single millisecond for that amount of time. I would say either simplify this back to just `inc` or  add a comment about why this logic is there.

camlow325: I'm good just changing it to an `inc`.  Will do that.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -356,34 +339,139 @@
...
+    (environment-class-info-entry)))
+
+(schema/defn ^:always-validate updated-environment-class-info-entry
```
rlinehan: nitpick: I feel like we usually make our function names verbs? `update-environment-class-info-entry` rather than `updated-environment-class-info-entry`, but I dunno that that's a rule or anything that we all agree on.

camlow325: Made this change in 66022c1.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
camlow325: Replaced the code with `inc` in 66022c1.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -356,34 +339,137 @@
...
+  with the supplied tag and a cache-generation-id value that has been
+  incremented."
+  [original-environment-class-info-entry :-
+   (schema/maybe EnvironmentClassInfoCacheEntry)
+   tag :- (schema/maybe schema/Str)]
+  (-> original-environment-class-info-entry
+      inc-cache-generation-id-for-class-info-entry
+      (assoc :tag tag)))
+
+(schema/defn ^:always-valid invalidated-environment-class-info-entry
```
rlinehan: Nitpick that I should have commented on this before you updated the PR, but this function name also has `invalidated` rather than `invalidate`.

rlinehan:
:+1: looks good to me. Left a couple small comments on some of the tests, but nothing I feel like should block merge.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
camlow325: Changed the name to `invalidate` in 21e74aa.

camlow325:
@rlinehan Thanks again for all of the feedback.  I think I've addressed all of your latest comments but please let me know if you see anything else.

rlinehan:
@camlow325 I'm good! :shipit:

PR #952, opened 2016-03-09
(SERVER-1195) Serve files from any directory

fpringvaldsen:
Modify the path for the `/static_file_content` endpoint so that
it matches any file under the parent directory with the glob
`*/*/files/**`, rather than only matching files under the glob
`modules/*/files/**`.


test/unit/puppetlabs/services/master/master_core_test.clj
```diff
@@ -232,15 +232,18 @@
                      "modules/foo/files/bar"
                      "modules/foo/files/bar/baz.txt"
                      "modules/foo/files/bar/more/path/elements/baz.txt"
-                     "modules/foo/files/bar/%2E%2E/baz.txt"]
+                     "modules/foo/files/bar/%2E%2E/baz.txt"
+                     "environments/production/files/~/.bash_profile"
```
haus: This is awesome. This would be `/etc/puppetlabs/code/environments/production/environments/production/files/~/.bash_profile`. A+

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -352,7 +352,7 @@
      ;; Here, keywords represent a single element in the path. Anything between two '/' counts.
      ;; The second vector takes anything else that might be on the end of the path.
      ;; Below, this corresponds to 'modules/*/files/**' in a filesystem glob.
```
haus: The comment probably needs an update.

test/unit/puppetlabs/services/master/master_core_test.clj
fpringvaldsen: Well we said files from *any* directory

haus: No totally. Absolutely valid. Also totally a path that looks like someone messed up a `cp -r` somewhere.

fpringvaldsen:
@haus I've gone ahead and fixed that comment

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -351,8 +351,8 @@
   (when-let [canonicalized-path (decode-and-canonicalize-path path)]
      ;; Here, keywords represent a single element in the path. Anything between two '/' counts.
      ;; The second vector takes anything else that might be on the end of the path.
-     ;; Below, this corresponds to 'modules/*/files/**' in a filesystem glob.
-     (bidi.bidi/match-route [["modules/" :module-name "/files/" [#".*" :rest]] :_]
+     ;; Below, this corresponds to '*/*/files/**' in a filesystem glob.
+     (bidi.bidi/match-route [[#"[^/]+/" :module-name "/files/" [#".*" :rest]] :_]
```
haus: An alternative here would be `["/" :modules-dir "/" :module-name "/files/" [#".*" :rest]]`, but I'm not sure if it's cleaner.

haus:
:+1:

KevinCorcoran:
I think this needs to target `stable`, not `master`.  Right, @nathanielksmith @camlow325 ?

test/unit/puppetlabs/services/master/master_core_test.clj
KevinCorcoran: :a:+

nathanielksmith:
At this point, I feel like we're going to end up doing another big merge down before we actually want to do a release (which seems like it's not until the 17th). Otherwise, we can do a merge up after the release from stable -> master.

cprice404:
@nathanielksmith I think that all of the 2.3.0 changes have been targeted at stable since the last merge down.  you are correct that we will need to do a merge up after release though.

KevinCorcoran:
> I think that all of the 2.3.0 changes have been targeted at stable since the last merge down.

I think that's correct, and I think it means that this should target `stable`.

fpringvaldsen:
I'll go ahead and target this at stable.

PR #953, opened 2016-03-09
(TK-338) Bump to tk-j9 1.5.4 to handle TimeoutException

cprice404:
This commit bumps us to tk-j9 v1.5.4, which includes a fix for
an issue that would happen if Jetty's stopTimeout was hit during
a HUP restart.  Without the fix, this will be treated as a fatal
error.  With the fix, the exception is handled and the HUP restart
can proceed.


camlow325:
:+1:

PR #950, opened 2016-03-08
(SERVER-1160) Update execution.rb for complex commands

rlinehan:
Previously, Puppet Server failed to correctly parse more complex commands
(such as those with lots of string escaping) passed to it by Puppet for execution.

This PR fixes that problem by updating execution.rb such that when an
array is passed it no longer concatenates that array to be a string, but
instead the first argument is passed to `ShellUtils/executeCommand` as a
string and the rest are passed as an array.

PR #954, opened 2016-03-09
(SERVER-1160) Fix command parsing

rlinehan:
Previously, Puppet Server failed to correctly parse more complex commands
(with lots of string escaping) passed to it by Puppet for execution.

This PR fixes that problem by updating execution.rb such that when an
array is passed it no longer concatenates that array to be a string, but
instead the first argument is passed to `ShellUtils/executeCommand` as a
string and the rest are passed as an array.


camlow325:
:+1:

camlow325:
This already ran successfully in Travis for #950 when targeted at the master branch, so I don't see a need to wait for a full Travis run again against the stable branch (since the two are basically identical at this point).  Merging...

PR #955, opened 2016-03-09
(SERVER-1195) Serve files from any directory

fpringvaldsen:
Modify the path for the `/static_file_content` endpoint so that
it matches any file under the parent directory with the glob
`*/*/files/**`, rather than only matching files under the glob
`modules/*/files/**`.

This is a replacement for https://github.com/puppetlabs/puppet-server/pull/952


src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -368,8 +368,8 @@
   (when-let [canonicalized-path (decode-and-canonicalize-path path)]
      ;; Here, keywords represent a single element in the path. Anything between two '/' counts.
      ;; The second vector takes anything else that might be on the end of the path.
-     ;; Below, this corresponds to 'modules/*/files/**' in a filesystem glob.
-     (bidi.bidi/match-route [["modules/" :module-name "/files/" [#".*" :rest]] :_]
+     ;; Below, this corresponds to '*/*/files/**' in a filesystem glob.
+     (bidi.bidi/match-route [[#"[^/]+/" :module-name "/files/" [#".*" :rest]] :_]
```
haus: For consistency with Puppet's behavior, the `.*` here should be `.+`, so that just the files directory won't be a valid path.

fpringvaldsen:
@haus I've updated the regex to be `.+` instead of `.*`

haus:
:+1:

PR #956, opened 2016-03-09
(SERVER-1194) Validate code-id, env for external scripts

nathanielksmith:
This commit adds validation for inputs to the code id and code content
scripts. While ShellUtils ought to prevent any bash injection attacks,
we still deemed it wise to do validation on this sort of input.

To that end, this commit adds schemas for environment and code-id
strings and uses them at both the HTTP level and the schema level.

An old helper, `valid-code-id?`, was removed in favor of using schema.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/common.clj
+++ b/src/clj/puppetlabs/puppetserver/common.clj
@@ -1,0 +1,18 @@
+(ns puppetlabs.puppetserver.common
+  (:require [schema.core :as schema]))
+
+(def Environment
+  "Schema for environment names. Alphanumeric and _ only."
+  (schema/pred (comp not nil? (partial re-matches #"\w+")) "environment"))
+
+(def CodeId
+  "Validates that code-id contains only alpha-numerics and
+  '-', '_', ';', or ':'."
+  (schema/pred (comp not (partial re-find #"[^_\-:;a-zA-Z0-9]")) "code-id"))
+
+(def environment-validation-error-msg
+  (partial format "The environment must be purely alphanumeric, not '%s'"))
+
+(def code-id-validation-error-msg
+  (partial format "Invalid code-id '%s'. Must contain only alpha-numerics and '-', '_', ';', or ':'"))
+
--- a/src/clj/puppetlabs/puppetserver/jruby_request.clj
+++ b/src/clj/puppetlabs/puppetserver/jruby_request.clj
@@ -2,7 +2,9 @@
   (:require [clojure.tools.logging :as log]
             [ring.util.response :as ring-response]
             [slingshot.slingshot :as sling]
-            [puppetlabs.services.jruby.jruby-puppet-service :as jruby]))
+            [puppetlabs.services.jruby.jruby-puppet-service :as jruby]
+            [schema.core :as schema]
+            [puppetlabs.puppetserver.common :as ps-common]))
 
 (defn throw-bad-request!
   "Throw a ::bad-request type slingshot error with the supplied message"
@@ -82,12 +84,9 @@
         (throw-bad-request!
          "An environment parameter must be specified")
 
-        (not (re-matches #"^\w+$" environment))
+        (not (nil? (schema/check ps-common/Environment environment)))
         (throw-bad-request!
-         (str
-          "The environment must be purely alphanumeric, not '"
-          environment
-          "'"))
+         (ps-common/environment-validation-error-msg environment))
 
         :else
         (handler request)))))
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -6,6 +6,7 @@
            (org.eclipse.jetty.util URIUtil))
   (:require [me.raynes.fs :as fs]
             [puppetlabs.puppetserver.ringutils :as ringutils]
+            [puppetlabs.puppetserver.common :as ps-common]
             [puppetlabs.comidi :as comidi]
             [ring.util.response :as rr]
             [schema.core :as schema]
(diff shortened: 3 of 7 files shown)


src/clj/puppetlabs/puppetserver/common.clj
```diff
@@ -0,0 +1,18 @@
...
+(def CodeId
+  "Validates that code-id contains onlyalpha-numerics and
+  '-', '_', or ':'."
+  (schema/pred (comp not (partial re-find #"[^_\-:;a-zA-Z0-9]")) "code-id"))
+
+(def environment-validation-error-msg
+  (partial format "The environment must be purely alphanumeric, not '%s'"))
+
+(def code-id-validation-error-msg
+  (partial format "Invalid code-id '%s'. Must contain only alpha-numerics and '-', '_', or ':'"))
```
haus: `;` is also valid in code-id

src/clj/puppetlabs/puppetserver/common.clj
```diff
@@ -0,0 +1,18 @@
+(ns puppetlabs.puppetserver.common
+  (:require [schema.core :as schema]))
+
+(def Environment
+  "Schema for environment names. Alphanumeric and _ only."
+  (schema/pred (comp not nil? (partial re-matches #"\w+")) "environment"))
+
+(def CodeId
+  "Validates that code-id contains onlyalpha-numerics and
```
haus: `onlyalpha` => `only alpha`

test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
```diff
@@ -68,32 +69,54 @@
...
+  (testing "when making catalog requests with post"
     (bootstrap/with-puppetserver-running
      app {:jruby-puppet
           {:max-active-instances num-jrubies}
           :versioned-code
           {:code-id-command (script-path "echo")
            :code-content-command (script-path "echo")}}
-     (let [catalog (testutils/post-catalog)]
-       (is (= "production" (get catalog "code_id")))))))
+      (testing "and envrionment is valid code id is injected"
```
haus: typo in environment

src/clj/puppetlabs/puppetserver/jruby_request.clj
```diff
@@ -1,8 +1,10 @@
 (ns puppetlabs.puppetserver.jruby-request
   (:require [clojure.tools.logging :as log]
+            [puppetlabs.puppetserver.common :refer [Environment]]
```
fpringvaldsen: I don't think this `:refer` is necessary

src/clj/puppetlabs/puppetserver/jruby_request.clj
```diff
@@ -1,8 +1,10 @@
...
             [ring.util.response :as ring-response]
             [slingshot.slingshot :as sling]
-            [puppetlabs.services.jruby.jruby-puppet-service :as jruby]))
+            [puppetlabs.services.jruby.jruby-puppet-service :as jruby]
+            [schema.core :as schema]))
```
fpringvaldsen: This `require` is also not necessary from what I can tell

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -15,7 +16,8 @@
             [cheshire.core :as cheshire]
             [clojure.string :as str]
             [bidi.schema :as bidi-schema]
-            [ring.middleware.params :as ring]))
+            [ring.middleware.params :as ring]
+            [puppetlabs.puppetserver.common :as ps-common]))
```
fpringvaldsen: It looks like this has been required twice

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
```diff
@@ -5,10 +5,13 @@
...
             [clojure.string :as string]
             [puppetlabs.ssl-utils.core :as ssl-utils]
+            [puppetlabs.puppetserver.common :as ps-common]
             [ring.util.codec :as ring-codec]
             [puppetlabs.trapperkeeper.authorization.ring :as ring-auth]
             [puppetlabs.puppetserver.ring.middleware.params :as pl-ring-params]
-            [puppetlabs.puppetserver.jruby-request :as jruby-request]))
+            [puppetlabs.puppetserver.jruby-request :as jruby-request]
+            [schema.core :as schema]
+            [puppetlabs.puppetserver.common :as ps-common]))
```
fpringvaldsen: Same thing here, this looks like it was required twice

src/clj/puppetlabs/services/versioned_code_service/versioned_code_core.clj
```diff
@@ -1,6 +1,7 @@
 (ns puppetlabs.services.versioned-code-service.versioned-code-core
   (:require [schema.core :as schema]
             [clojure.tools.logging :as log]
+            [puppetlabs.puppetserver.common :refer [CodeId code-id-validation-error-msg Environment]]
```
fpringvaldsen: Generally I think we try to avoid doing `:refer` outside of tests, so we can more easily tell where things are coming from when reading through code

KevinCorcoran:
@nathanielksmith needs to be rebased :(

nathanielksmith:
; _ ;

nathanielksmith:
@KevinCorcoran rebased

haus:
@nathanielksmith the code changes look totally sane to me.

the one thing i don't know about are the doseq calls in the tests. last time i wrote tests that used doseq, @KevinCorcoran asked me to rework them, i think because doseq made the failure modes pretty hard to interpret. I may be misremembering the reasoning.

rlinehan:
Yeah, I also dislike `doseq` with a whole bunch of assertions. It's a lot better if you can make helper methods for some of it, but keep it so that each thing that you are testing has it's own assertion, so that if you fails you know exactly what is failing.

That said, I am fine with this going in as-is with a follow on PR to refactor the tests, given that we need to get this code promoted ASAP.

fpringvaldsen:
:+1:

I am also in favor of potentially changing the `doseq` calls, but am fine with punting until a later PR.

nathanielksmith:
my last team was in favor of using doseq which is why I updated what I found here. I definitely prefer the doseqs and I think they encourage devs to think of more use cases (simply because it is more convenient to add them). i'll do as the romans do in the future.

it sounds like folks agree it's not worth changing immediately.

rlinehan:
Tests passed locally and this has two :+1:'s, so merging.

PR #957, opened 2016-03-09
(SERVER-1202) Rename run! in gem and irb namespaces 

haus:
clojure 1.7.0 included a new function named run!, which has the same
name as existing functions in the gem and irb namespaces. This produces
unsightly warnings when using the gem subcommand.
This commit adds an exclusion on the run! command from clojure to avoid
these warnings.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/cli/gem.clj
+++ b/src/clj/puppetlabs/puppetserver/cli/gem.clj
@@ -2,10 +2,10 @@
   (:require [puppetlabs.puppetserver.cli.subcommand :as cli]
             [puppetlabs.services.jruby.jruby-puppet-core :as jruby-core]))
 
-(defn run!
+(defn gem-run!
   [config args]
   (jruby-core/cli-run! config "gem" args))
 
 (defn -main
   [& args]
-  (cli/run run! args))
+  (cli/run gem-run! args))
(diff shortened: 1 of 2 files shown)


haus:
On current stable:
```
lein gem
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=256m; support was removed in 8.0
WARNING: run! already refers to: #'clojure.core/run! in namespace: puppetlabs.puppetserver.cli.gem, being replaced by: #'puppetlabs.puppetserver.cli.gem/run!


Missing required argument '--config'!

  -c, --config CONFIG-PATH  Path to a configuration file or directory of configuration files. See the documentation for a list of supported file types.
  -h, --help                Show help
```

With this commit:
```
lein gem
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=256m; support was removed in 8.0


Missing required argument '--config'!

  -c, --config CONFIG-PATH  Path to a configuration file or directory of configuration files. See the documentation for a list of supported file types.
  -h, --help                Show help
```

rlinehan:
:+1: Ran tests locally and they passed.

KevinCorcoran:
It seems like a better solution would be to rename our functions to something other than `run!`.

camlow325:
I'd agree with @KevinCorcoran on that.  I think the `run!` names were totally arbitrary in the first place and could be changed without any external breakage.

haus:
Cool, as long as we don't expect anyone to be calling/using them.

KevinCorcoran:
:+1:

rlinehan:
~travisci

Ran tests locally, everything passed, :see_no_evil:, not gonna wait for travis because that would take way too long, :shipit: .

PR #958, opened 2016-03-10
(SERVER-1196) Add instrumentation to env-class-int-cache-enabled test

camlow325:
This PR adds some instrumentation to the
`environment-classes-integration-cache-enabled-test`, in an attempt to
get more information about why it fails when it does.

This commit adds a timeout to the initial wait that the
`class-info-updated-after-cache-flush-during-prior-request` does to see
that the getClassInfoForEnvironment method in a JRubyPuppet instance has
been invoked.  The current code waits indefinitely for that action,
which may never happen in the case that some higher-level failure has
occurred, e.g., the client fails to connect to the server, in which case
the test could hang.

### DIFF
--- a/test/integration/puppetlabs/services/master/environment_classes_int_test.clj
+++ b/test/integration/puppetlabs/services/master/environment_classes_int_test.clj
@@ -62,15 +62,18 @@
   ([env-name if-none-match]
    (let [opts (if if-none-match
                 {:headers {"If-None-Match" if-none-match}})]
-     (http-client/get
-      (str "https://localhost:8140/puppet/v3/"
-           "environment_classes?"
-           "environment="
-           env-name)
-      (merge
-       testutils/ssl-request-options
-       {:as :text}
-       opts)))))
+     (try
+       (http-client/get
+        (str "https://localhost:8140/puppet/v3/"
+             "environment_classes?"
+             "environment="
+             env-name)
+        (merge
+         testutils/ssl-request-options
+         {:as :text}
+         opts))
+       (catch Exception e
+         (throw (Exception. "environment_classes http get failed" e)))))))
 
 (defn purge-all-env-caches
   []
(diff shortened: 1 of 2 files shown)


test/integration/puppetlabs/services/master/environment_classes_int_test.clj
```diff
@@ -62,15 +62,18 @@
...
+        (str "https://localhost:8140/puppet/v3/"
+             "environment_classes?"
+             "environment="
+             env-name)
+        (merge
+         testutils/ssl-request-options
+         {:as :text}
+         opts))
+       (catch Exception e
+         (throw (Exception. "environment_classes http get failed" e)))))))
```
cprice404: What is this doing

camlow325:
I apologize in advance for how horrible the whitespace diffs look in `environment_classes_int_test` in 7996803.  Most of the line differences are indentation caused by surrounding the pre-existing test body with let and try/catch blocks.  Might be easiest to review that commit with `w=1`.

test/integration/puppetlabs/services/master/environment_classes_int_test.clj
cprice404: oh nm I understand, you just want to know for sure that the thing happened during the get.

cprice404:
+1

test/integration/puppetlabs/services/master/environment_classes_int_test.clj
camlow325: Wrapping the raw exception we get from the HTTP client with an exception that will have the stack trace that this line was called from.  HTTP client exceptions typically occur on Apache worker threads so you can't tell what line of code in the calling test was active on the stack otherwise.

camlow325:
Wouldn't mind letting this run through Travis once first before we merge this.  Would be nice to see it fail in a useful way.

fpringvaldsen:
:+1:

PR #961, opened 2016-03-11
(maint) Bump puppet submodule to stable latest

KevinCorcoran:
@joshcooper @nathanielksmith @camlow325

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -30,7 +30,7 @@ module PuppetServerExtensions
     puppet_build_version = get_option_value(options[:puppet_build_version],
                          nil, "Puppet Agent Development Build Version",
                          "PUPPET_BUILD_VERSION",
-                         "7282f093e3c77b3382938a1c2e22d12b59b7973e", :string)
+                         "7c523b75580210a79e2870ff262e6182d9060c05", :string)
 
     # puppetdb version corresponds to packaged development version located at:
     # http://builds.delivery.puppetlabs.net/puppetdb/


camlow325:
@KevinCorcoran Should we be bumping the puppet-agent version that we're pinning to along with this change - or are you just looking to pick up test changes with this PR?

KevinCorcoran:
just looking to pick up test changes

I don't think there's a new puppet-agent for us to bump to yet ... but maybe you know something I don't?

camlow325:
Another way to ask this...

Do we have high confidence the new test changes we would be picking up are not dependent upon production code changes in underlying Puppet?  If we don't know for sure, it's usually safest to bump both the Puppet submodule and puppet-agent versions together.

KevinCorcoran:
@camlow325 My understanding is that our puppet-agent dependency is currently at `>= 1.3.0`, and I don't know that there's anything reasonable for us to bump that to, yet.  Eventually we'll want `>= 1.4.0` but that does not exist yet.  That said, I defer to @joshcooper.

camlow325:
@KevinCorcoran Ahead of the puppet-agent 1.4.0 release, we've been pinning our puppet-agent for the stable and master branches right now to a specific pre-release build of puppet-agent.  See https://github.com/puppetlabs/puppet-server/blob/a35206571d66c428d0e4602143892cb26770a7d5/acceptance/lib/helper.rb#L33 for where we do that.  Sounds like it's probably best to bump up our puppet-agent to the latest passing puppet-agent build for the stable branch.  Note that we'll need to pick one which is built for the full OS matrix since our Puppet Server pipeline will exercise the fuller matrix in the nightly runs.

joshcooper:
Here are the set of changes in puppet: https://github.com/puppetlabs/puppet/compare/3b363400e0bca44f1762a5e5e7eb5803a320df98...db118077413e6dfe1faf394cd9207e3d603dd190. I think it's good to bump as you'll pick up changes we recently made around percent encoding. There were two changes to acceptance tests, though I think it was only to account for Solaris SPARC, which don't run in the vmpooler and have unusual DNS configuration. So I wouldn't expect it to affect your CI.

The most recent puppet-agent sha from stable is 7c523b75580210a79e2870ff262e6182d9060c05, and include the full suite of platforms.

camlow325:
@joshcooper Yeah, that looks like a good puppet-agent for us to pin to, thanks.  Looks like it included Ruby Puppet from the same git SHA as @KevinCorcoran used for the Puppet submodule on this PR as well.

@KevinCorcoran - you good with updating helper.rb in this PR to pin to that puppet-agent version?

KevinCorcoran:
@camlow325 @joshcooper updated

joshcooper:
:+1: Btw, there's a helpful kerminator command to lookup the puppet SHA given a puppet-agent SHA (has to be the full sha):

```
@kerminator vanagon puppet-agent 7c523b75580210a79e2870ff262e6182d9060c05
...
Components in puppet-agent at ref 7c523b75580210a79e2870ff262e6182d9060c05
----------------------------------------
...
puppet 4.3.2-579-gdb11807
...
```

And db11807 matches what you have in the PR

camlow325:
:+1:

PR #962, opened 2016-03-11
(SERVER-1204) Update execution stub to correctly handle empty args

haus:
Previously, puppet server's execution stub assumed that if it was handed
an array, that the first element would be the command. This turns out
not to be true for all of puppet's uses of the execute method.
Specifically, config_version is often a script with arguments that is
handed as a single string in an array to Puppet::Util::Execution.execute
and then on to puppet server's execution stub.
This is a problem because something like '/bin/echo a b c' is not a valid
command. This commit addresses the problem by using the correct call to
ShellUtils in the case where there is a single element array. The
ShellUtils call will in that case correctly parse '/bin/echo a b c' into
a call to /bin/echo with a b and c as arguments.

### DIFF
--- a/test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
+++ b/test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
@@ -10,6 +10,9 @@
 (def test-resources-dir
   (ks/absolute-path "./dev-resources/puppetlabs/general_puppet/general_puppet_int_test"))
 
+(def environment-dir
+  (ks/absolute-path (str testutils/conf-dir "/environments/production")))
+
 (defn script-path
   [script-name]
   (str test-resources-dir "/" script-name))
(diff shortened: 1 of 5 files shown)


KevinCorcoran:
:see_no_evil:

rlinehan:
Would it also be worth adding an integration test of this (I presume using generate)?

haus:
Couldn't hurt.

haus:
i need to do a touch of cleanup still. please hold.

test/integration/puppetlabs/puppetserver/testutils.clj
```diff
@@ -33,7 +34,7 @@
   {(schema/required-key "name") schema/Str
    (schema/required-key "classes") [schema/Str]
    (schema/required-key "environment") schema/Str
-   (schema/required-key "version") schema/Int
+   (schema/required-key "version") schema/Any
```
rlinehan: sidenote: would be good to update puppet's schema about this

rlinehan:
:+1: Will run tests locally because I'm guessing travis will take hours.

test/integration/puppetlabs/puppetserver/testutils.clj
haus: https://github.com/puppetlabs/puppet/pull/4787

test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
```diff
@@ -179,3 +182,14 @@
...
+(deftest ^:integration test-config-version-execution
+  (testing "config_version is executed correctly"
+    (testutils/create-env-conf
+     environment-dir
+     (format "config_version = \"%s $environment\"" (script-path "echo")))
+    (bootstrap/with-puppetserver-running
+     app {:jruby-puppet
+          {:max-active-instances num-jrubies}}
+     (let [catalog (testutils/get-catalog)]
+       (is (= "production" (get catalog "version")))))))
```
KevinCorcoran: ~mccune principle

KevinCorcoran:
still :see_no_evil:

rlinehan:
Tests passed when run locally, merging.

camlow325:
Post merge :+1: Nice job!

PR #964, opened 2016-03-11
(docs) Update file-path glob details.

gguillotte:
Per @joshcooper in `puppet-docs` PR 622 and SERVER-1195, update the `static_file_content` API docs to specify the paths for files that can have inlined file resource metadata.


joshcooper:
:+1:

PR #965, opened 2016-03-11
(SERVER-557) Remove ezbake timeout override

cprice404:
The timeout has been raised in ezbake itself, and we can't think
of any great reason not to just use the default from there.


camlow325:
:+1:

PR #972, opened 2016-03-16
(ORCH-1235) Add code_id to environment catalog requests

nicklewis:
Similar to agent catalog requests, we now add code id for the given
environment to environment catalog requests.

This extends the logic for determining the request environment to now check
both the :environment URL segment as well as the "environment" query param.
If environment is part of the URL, we favor that over a conflicting query
param.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/jruby_request.clj
+++ b/src/clj/puppetlabs/puppetserver/jruby_request.clj
@@ -67,10 +67,13 @@
      (handler (assoc request :jruby-instance jruby-instance)))))
 
 (defn get-environment-from-request
-  "Gets the environment from a request."
+  "Gets the environment from the URL or query string of a request."
   [req]
-  (-> req
-      (get-in [:params "environment"])))
+  ;; If environment is derived from the path, favor that over a query/form
+  ;; param named environment, since it doesn't make sense to ask about
+  ;; environment production in environment development.
+  (or (get-in req [:route-params :environment])
+      (get-in req [:params "environment"])))
 
 (defn wrap-with-environment-validation
   "Middleware function which validates the presence and syntactical content
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -30,7 +30,7 @@ module PuppetServerExtensions
     puppet_build_version = get_option_value(options[:puppet_build_version],
                          nil, "Puppet Agent Development Build Version",
                          "PUPPET_BUILD_VERSION",
-                         "878f7e7c2084716a483544aeb5ebffb4ec48d9d6", :string)
+                         "2c492d8f1ff5018c171b9f4cef7671f14d92c215", :string)
 
     # puppetdb version corresponds to packaged development version located at:
     # http://builds.delivery.puppetlabs.net/puppetdb/
--- a/dev-resources/puppetlabs/general_puppet/general_puppet_int_test/puppet.conf
+++ b/dev-resources/puppetlabs/general_puppet/general_puppet_int_test/puppet.conf
@@ -3,3 +3,4 @@
 certname = localhost
 environmentpath = $confdir/environments
 environment_timeout = unlimited
+app_management = true
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -437,8 +437,8 @@
                (request-handler request))
    (comidi/GET ["/resource_types/" [#".*" :rest]] request
                (request-handler request))
-   (comidi/GET ["/environment/" [#".*" :rest]] request
-               (request-handler request))
+   (comidi/GET ["/environment/" [#".*" :environment]] request
+               (request-handler (assoc request :include-code-id? true)))
    (comidi/GET "/environments" request
                (request-handler request))
    (comidi/GET ["/status/" [#".*" :rest]] request
--- a/test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
+++ b/test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
@@ -2,6 +2,7 @@
   (:require [clojure.test :refer :all]
             [puppetlabs.puppetserver.bootstrap-testutils :as bootstrap]
             [me.raynes.fs :as fs]
+            [cheshire.core :as json]
             [puppetlabs.http.client.sync :as http-client]
             [puppetlabs.puppetserver.testutils :as testutils]
             [puppetlabs.trapperkeeper.testutils.logging :as logging]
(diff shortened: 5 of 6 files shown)


nicklewis:
I wasn't entirely sure what to do about the submodule. There's a change to Puppet to actually *use* the code id that's added to the request, and that change is necessary for the new integration test to pass, so I pulled it in. I also updated the PUPPET_BUILD_VERSION in the acceptance helper because a) other commits did that and b) it seems like it makes sense to keep the puppet-agent dependency and the puppet submodule versions in sync? The new puppet-agent shouldn't strictly be necessary for the acceptance tests to pass though (since I didn't change them).

fpringvaldsen:
:see_no_evil:

cprice404:
@nicklewis this is for post-burnside, presumably?

nicklewis:
Yep, post-burnside.

src/clj/puppetlabs/puppetserver/jruby_request.clj
```diff
@@ -67,10 +67,10 @@
      (handler (assoc request :jruby-instance jruby-instance)))))
 
 (defn get-environment-from-request
-  "Gets the environment from a request."
+  "Gets the environment from the URL or query string of a request."
   [req]
-  (-> req
-      (get-in [:params "environment"])))
+  (or (get-in req [:params :environment])
+      (get-in req [:params "environment"])))
```
cprice404: Might be worth adding a comment here explaining the difference between the two.

@camlow325 @haus do you have any concerns about this in terms of the other endpoints?  Seems like it should be fine?

cprice404:
@nicklewis what does this endpoint do in OSS?  Can discuss offline if that's preferable.

I'm pretty much +1 on this other than the minor nit/question above.

Also looks like it'll need a rebase.

nicklewis:
@cprice404 This endpoint is available in OSS and behaves the same as in PE. It will compile and return the environment catalog for the specified environment, if the `app_management` setting is enabled in puppet.conf.

The corresponding puppet change happened in puppetlabs/puppet#5099.

Will add comment + rebase.

nicklewis:
Rebased and added a comment explaining the reasoning about path component vs. query string parameter.

src/clj/puppetlabs/puppetserver/jruby_request.clj
haus: I don't have any concerns, except for the symbol vs string shenanigans of puppet past.

camlow325: Giving priority to the parameter we extract from the comidi route over any that might have been found from a query parameter or form parameter seems reasonable to me.

I think it would be good to be consistent between what this helper method does and how the parameter would be translated onto the request map that we construct and send on down to JRuby.  Prior to this line ...

https://github.com/puppetlabs/puppet-server/blob/puppet-server-2.3.0/src/clj/puppetlabs/services/request_handler/request_handler_core.clj#L278

... in the handling of a request, I think the params map could have both an `:environment` and `environment` key in it.  Doesn't seem like `stringify-keys` would ensure that the keyword form of the key would take precedence over the string one always?  Might depend upon the order in which keys were added to the map?  Might be good to replace this call with something which would explicitly give the keywordized keys predence over the string based ones?

cprice404: @camlow325 what would that look like?  Digging into the less generic params maps in the request?  Like... `route-params` and `query-params`?  If that's what you're suggesting that sounds like it might be worthwhile.  Does `stringify-keys` monkey will all of those maps or just the generic `params` map?

camlow325: `stringify-keys just looks at the generic `:params` map, stringifying any keywords found for keys.  I was initially just naiively thinking about instead having it walk the generic `:params` map and choose a keyword representation over a string one when both are present.

I like your idea about merging the params from different sources better, though - stringifying keys for each as you go - with `route-params` winning over `query-params`, winning over `form-params` where there are matching keys.  Maybe we should pull all of this up into some higher level middleware the `params` handling in `wrap-params-for-jruby`?  If we did that, seems like we wouldn't even need the above code because we would have normalized to a single "environment" key being in the map at the point this code is being evaluated?

src/clj/puppetlabs/puppetserver/jruby_request.clj
```diff
@@ -67,10 +67,14 @@
...
+  "Gets the environment from the URL or query string of a request."
   [req]
-  (-> req
-      (get-in [:params "environment"])))
+  ;; :environment is the value of a path component, "environment" is the value
+  ;; of a query string parameter. If environment is an inherent part of the
+  ;; resource being requested (part of the path), we prefer that so it doesn't
+  ;; get overridden by a rogue query string parameter.
+  (or (get-in req [:params :environment])
+      (get-in req [:params "environment"])))
```
cprice404: Picking up this from the comment thread on the old commit so that it's less invisible.

@camlow325 said:

> stringify-keys just looks at the generic :params map, stringifying any keywords found for keys. I was initially just naiively thinking about instead having it walk the generic :params map and choose a keyword representation over a string one when both are present.
>
> I like your idea about merging the params from different sources better, though - stringifying keys for each as you go - with route-params winning over query-params, winning over form-params where there are matching keys. Maybe we should pull all of this up into some higher level middleware replacement for the params handling in wrap-params-for-jruby? If we did that, seems like we wouldn't even need the above code because we would have normalized to a single "environment" key being in the map at the point this code is being evaluated?

@camlow325 that sounds like a reasonable thing to consider but is probably more than what I was imagining for right now.  Right now I'm just thinking about a smaller change that would allow us to get this merged for @nicklewis and we could circle back around later to improve it.

The idea that I'm floating is just to modify these two lines of code so that the first one digs into the 'route-params' map to see if there was a path element for the environment, and then *maybe* also changing the second line here to specifically dig into the `query-params` map.  That last part might not be a good idea though since Puppet may send that "query param" up as part of the form-encoded POST body.  But I think the first part about specifically looking in `route-params` for the value that came out of the URL path should be ok?

camlow325: @cprice404 Yeah, shorter-term fix which allows this to get merged would be fine.  I'm fine with just changing the first line to look for the param in `route-params` else fall back to the `params` map.  The `params` map would have the one from the query-string, if present, else the one from a urlencoded-form parameter, if present, because it will have been culled from the Ring middleware into the `params` map by this point.

cprice404: @nicklewis would you mind making that one minor change for now, then, and we can get this merged and spare you the rest of our musings about disparate sources for request parameters? :)

nicklewis:
Updated to look at `[:route-params :environment]` and then `[:params "environment"]`.

cprice404:
+1, @camlow325 if that looks ok to you feel free to merge.

camlow325:
:shipit:

PR #974, opened 2016-03-16
(doc) Fix JIRA ticket number on known issue.

gguillotte:


gguillotte:
CC @nathanielksmith @haus

nathanielksmith:
:+1:

documentation/release_notes.markdown
```diff
@@ -43,7 +43,7 @@ Puppet Server 2.3.0 and newer support being restarted by sending a hangup signal
 
 Previous versions of Puppet Server could fail when the `config_version` setting is a single string that contains spaces, which is commonly true when that setting points to a script with space-delineated arguments. Puppet Server 2.3.0 resolves this issue.
 
-* [SERVER-1204](https://tickets.puppetlabs.com/browse/SERVER-1204)
+* [SERVER-1160](https://tickets.puppetlabs.com/browse/SERVER-1160)
```
rlinehan: This is the write ticket number to include here, but I think the description isn't accurate. I've updated SERVER-1160 with some release notes. The description here matches SERVER-1204, but that was never a bug in a released version of Puppet Server.

rlinehan: (sorry I forgot to update the release notes on SERVER-1160 to begin with :( )

haus:
@gguillotte i think in addition to the ticket change the wording should be updated.

Something like...

Previously custom functions or calls to generate could exhaust jrubies. Specifically, this is triggered by calls to commands that block waiting on input (the ticket in question opened a python repl in each jruby). In 2.3.0 this is no longer the case.

rlinehan:
:+1:

haus:
:+1:

PR #978, opened 2016-03-18
(maint) Bump project.clj version to 2.3.2-master-SNAPSHOT

KevinCorcoran:
2.3.0 has already been released, so the previous version did not make sense anymore.


rlinehan:
We've already bumped this to 2.4.0-master, so I feel like this ticket can be closed?

PR #980, opened 2016-03-19
(SERVER-864) Allow special characters in certificate_statuses key

justinstoller:
Prior to this the certificate_statuses endpoint would not match calls
when the required key path segment contained anyting that wasn't an
alphanum, hyphen, underscore, or period. eg.
  `puppet-ca/v1/certificate_statuses/*`
would fail to match the specification of:
  `:ca-mount-endpoint/:version/certificate_statuses/:key`
even though there is no further requirements on the key path segment.

This behavior is determined by how we describe our path segment captures
to a third party library (bidi).

This patch updates our route description to allow the key path segment
to match any single path segment of one or more characters in length.

### DIFF
--- a/test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
+++ b/test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
@@ -305,6 +305,20 @@
           (is (= #{localhost-status test-agent-status revoked-agent-status}
                  (set (json/parse-string (:body response) true)))))
 
+        (testing "requires ignored path segment"
+          (let [response (test-app
+                          {:uri "/v1/certificate_statuses/"
+                           :request-method :get})]
+            (is (= 404 (:status response)))))
+
+        (testing "allows special characters in ignored path segment"
+          (let [response (test-app
+                          {:uri "/v1/certificate_statuses/*"
+                           :request-method :get})]
+            (is (= 200 (:status response)))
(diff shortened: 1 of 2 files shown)


fpringvaldsen:
:see_no_evil:

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
```diff
@@ -305,6 +305,20 @@
           (is (= #{localhost-status test-agent-status revoked-agent-status}
                  (set (json/parse-string (:body response) true)))))
 
+        (testing "requires ignored path segment"
+          (let [response (test-app
+                          {:uri "/v1/certificate_statuses/"
+                           :request-method :get})]
+            (is (= 404 (:status response)))))
```
KevinCorcoran: This seems more like a 400 than a 404.  But that [doesn't seem to be documented](https://docs.puppetlabs.com/puppet/latest/reference/http_api/http_certificate_status.html) and I'm not sure if it would be worth the effort to implement that change.  Thoughts?  I'm going to merge this as-is but feel free to submit a follow-up PR if you think it's worth the effort to switch to a 400 here.

justinstoller: It wasn't difficult to implement, and nice way get back in the swing of things after being sick, so I put up a PR at https://github.com/puppetlabs/puppet-server/pull/997 .

PR #988, opened 2016-03-23
(SERVER-1215) Adds a test that validates logback with a syslog config…

erikPrime:
NOTE: 
- Does not validate actual logging
- Will fail unless run against puppet-server 2.3.0+


rlinehan:
I think that this should be targeted at stable, since that's where we fixed SERVER-1215 (and where we released Server 2.3.1 from) and then merged up to master.

erikPrime:
This PR is superseded by https://github.com/puppetlabs/puppet-server/pull/993

PR #989, opened 2016-03-23
(maint) Fix pedantic errors and puppet agent version

rlinehan:
Previously, puppet server packages were failing to be built because `lein
with-profile ezbake` had version conflicts and pedantic errors. This updates to a new tk-ws-j9 that removes these version conflicts.

The previous PUPPET_BUILD_VERSION was set to a puppet-agent build that was
1.3.5, which didn't work with puppet server's current requirement on
puppet-agent >= 1.4.0. This updates the PUPPET_BUILD_VERSION to the latest passing
puppet-agent version off puppet master, and the puppet submodule to the commit
that is in that puppet-agent version.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -30,7 +30,7 @@ module PuppetServerExtensions
     puppet_build_version = get_option_value(options[:puppet_build_version],
                          nil, "Puppet Agent Development Build Version",
                          "PUPPET_BUILD_VERSION",
-                         "2c492d8f1ff5018c171b9f4cef7671f14d92c215", :string)
+                         "adf3cd46f5904ba119567bd4bd9784cc255f1a90", :string)
 
     # puppetdb version corresponds to packaged development version located at:
     # http://builds.delivery.puppetlabs.net/puppetdb/
--- a/project.clj
+++ b/project.clj
@@ -1,6 +1,6 @@
 (def clj-version "1.7.0")
 (def tk-version "1.3.1")
-(def tk-jetty-version "1.5.4")
+(def tk-jetty-version "1.5.5")
 (def ks-version "1.3.0")
 (def ps-version "2.4.0-master-SNAPSHOT")
 
--- a/spec/puppet-server-lib/puppet/jvm/master_spec.rb
+++ b/spec/puppet-server-lib/puppet/jvm/master_spec.rb
@@ -24,7 +24,7 @@ describe 'Puppet::Server::Master' do
     end
 
     it "returns the correct puppet version number" do
-      expect(subject).to eq('4.4.0')
+      expect(subject).to eq('4.4.1')
     end
   end
 


haus:
@rlinehan I think you need to bump the puppet version, as you've pulled in a version of puppet after the 4.4.0 release

rlinehan:
blargh. thanks @haus. I forgot to run the spec tests :/

rlinehan:
@haus updated.

KevinCorcoran:
Should the update to 1.4.1 be done against `stable` and then merged-up instead of directly against `master`?

rlinehan:
@KevinCorcoran the tests failures were caused by a merge-up of stable->master, in which there was a merge conflict in acceptance/lib/helper.rb and the version that was already on master was chosen. I mean, I guess we can do what you're suggesting? but master and stable of puppet have diverged, and it seems like our stable should be tracking stable of puppet, and our master should be tracking master of puppet.

camlow325:
:+1:

KevinCorcoran:
As long as the update to 1.4.1 / 4.4.1 was originally done in `stable` (which, if I understand you correctly, it was), I'm satisfied.

rlinehan:
@KevinCorcoran I don't think that stable has been updated to 1.4.1/4.4.1, I think it's still testing against 1.4.0. I can submit a PR for that if you'd like.

Note however that this PR isn't updating to 1.4.1 and 4.4.1. It's updating to the latest passing puppet-agent building off of puppet master, and the puppet master SHA that is included in it. So we could do a stable change to 1.4.1 and 4.4.1, merge that up, then correct this to be tracking something on master, if we really care that much? But it doesn't really seem like our merge ups should be carrying forward puppet/puppet-agent version changes, because those should actually be tracking different things between branches, they should just be carrying forward our code and test changes.

KevinCorcoran:
> @KevinCorcoran I don't think that stable has been updated to 1.4.1/4.4.1, I think it's still testing against 1.4.0. I can submit a PR for that if you'd like.

Oh.  Yeah, that that seems like a good thing, but less important than I was originally thinking it might be.

> Note however that this PR isn't updating to 1.4.1 and 4.4.1. It's updating to the latest passing puppet-agent building off of puppet master, and the puppet master SHA that is included in it. 

Okay, thanks.  I was obviously very :confused: here.

> So we could do a stable change to 1.4.1 and 4.4.1, merge that up, then correct this to be tracking something on master, if we really care that much? But it doesn't really seem like our merge ups should be carrying forward puppet/puppet-agent version changes, because those should actually be tracking different things between branches, they should just be carrying forward our code and test changes.

Yeah, if we are tracking something from puppet#master, I don't think we need to worry about merging that change from `stable` -> `master`.

camlow325:
Yeah, I definitely think we want to be tracking builds made from puppet#master in puppet-server#master.  We already have some dependencies in puppet-server#master like #972 on work that I don't believe has/will land in the current form of the puppet#stable branch.  @rlinehan's PR here looked good to me in that it was tracking something later that's still on puppet#master and ties to a full OS-matrix puppet-agent build.

PR #993, opened 2016-03-24
(SERVER-1215):Adds test that validates that a syslog config in logbac…

erikPrime:
…k does not prevent puppetserver restart.  Does not validate actual logging.  Supersedes PR#992.

### DIFF
--- a/acceptance/suites/tests/00_smoke/validate-syslog-logback.rb
+++ b/acceptance/suites/tests/00_smoke/validate-syslog-logback.rb
@@ -52,7 +52,7 @@ end
 
 
 step 'Backup logback'
-  on(master, "mv #{logback_path} #{logback_backup}", :acceptable_exit_codes => [0,1])
+  on(master, "mv #{logback_path} #{logback_backup}")
 
 step 'Modify logback configuration'
   create_remote_file(master, logback_path, logback_config)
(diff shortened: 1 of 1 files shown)


acceptance/suites/tests/00_smoke/validate-syslog-logback.rb
```diff
@@ -0,0 +1,67 @@
...
+
+step 'Modify logback configuration'
+  create_remote_file(master, logback_path, logback_config)
+
+step 'Restart puppetserver'
+  on(master, "service #{service_name} restart")
+  # In some cases, service puppetserver restart returns a 0 even if the
+  # service fails to restart and the init script throws an error message.
+
+step 'Validate that the puppetserver service is running'
```
haus: Will step work with a multi line body without a block?

erikPrime: Yes, step works fine without a block.  Would you, and developers in general prefer QA write tests that use steps with bodies?

acceptance/suites/tests/00_smoke/validate-syslog-logback.rb
```diff
@@ -0,0 +1,69 @@
...
+
+teardown do
+  on(master, "mv #{logback_backup} #{logback_path}")
+  on(master, "service #{service_name} restart")
+  on(master, "service #{service_name} status")
+end
+
+
+step 'Backup logback'
+  on(master, "mv #{logback_path} #{logback_backup}", :acceptable_exit_codes => [0,1])
```
haus: Why allow a failed mv here?

haus:
This looks good to me aside from my question about exit codes and `mv`.

PR #996, opened 2016-03-29
(SERVER-1150) Add support for custom oid mappings

fpringvaldsen:
Add support for Puppet's custom trusted oid mapping feature.


src/ruby/puppet-server-lib/puppet/server/oids.rb
```diff
@@ -0,0 +1,31 @@
+module Puppet::Server::Oids
+  def self.load_custom_oid_file(custom_oid_file, map_key='oid_mapping')
+    if File.exists?(custom_oid_file) && File.readable?(custom_oid_file)
+      mapping = nil
+      begin
+        mapping = YAML.load_file(custom_oid_file)
+      rescue => err
```
haus: It's probably worth being specific about what errors we care about rescuing here.

src/ruby/puppet-server-lib/puppet/server/oids.rb
```diff
@@ -0,0 +1,31 @@
+module Puppet::Server::Oids
+  def self.load_custom_oid_file(custom_oid_file, map_key='oid_mapping')
+    if File.exists?(custom_oid_file) && File.readable?(custom_oid_file)
```
haus: Are there any cases where a File will be readable but not exist? Would the `readable?` check on its own be sufficient here?

src/ruby/puppet-server-lib/puppet/server/oids.rb
```diff
@@ -0,0 +1,31 @@
...
+      mapping = nil
+      begin
+        mapping = YAML.load_file(custom_oid_file)
```
haus: This probably needs a `require 'yaml'`.

src/ruby/puppet-server-lib/puppet/server/oids.rb
fpringvaldsen: From what I can tell, the `readable?` check would be sufficient on its own. However, this function was copied straight from the Puppet codebase with minor modifications (it returns `oid_defns` instead of iterating through and registering everything in it with OpenSSL), so I wasn't sure if we should keep parity with the way that function actually performs the file parsing

src/ruby/puppet-server-lib/puppet/server/oids.rb
fpringvaldsen: Same comment here as above

src/ruby/puppet-server-lib/puppet/server/oids.rb
haus: ok that's fair. i don't know how we've done this in the past, but keeping it as close to the source as possible sounds reasonable.

src/ruby/puppet-server-lib/puppet/server/certificate.rb
```diff
@@ -15,6 +16,7 @@ def initialize(java_cert)
     end
 
     @java_cert = java_cert
+    @oid_defns = Puppet::Server::Oids.load_custom_oid_file(Puppet[:trusted_oid_mapping_file]) + Puppet::SSL::Oids::PUPPET_OIDS
```
camlow325: It looks like this line would end up getting run every time a `Certificate` object is created, which happens every time a request is handled by the web server.  I think it would be better to have this be a one-time thing.

Maybe the `load_custom_oid_file` call could just be done from the `puppet_config.rb` file like you have below and that could set a class-level variable that has the concatenation of the mappings from the trusted oid mapping file + `Puppet::SSL::Oids::PUPPET_OIDS`.  From the instance methods in this class, then, we could just retrieve the contents of the class-level variable via a self method on `Puppet::Server::Oids`?

camlow325: Another problem I ran into is if the file pointed to by the `trusted_oid_mapping_file` setting does not exist, I get an error every time I make a request:

```
2016-03-29 17:23:42,249 ERROR [qtp1610975141-41] [puppet-server] Puppet Server Error: undefined method `+' for nil:NilClass
{noformat}

I think the implementation needs to handle just using {{Puppet::SSL::Oids::PUPPET_OIDS}} as a base list but add on the ones from the `trusted_oid_mapping_file` if it exists.

File: src/ruby/puppet-server-lib/puppet/server/oids.rb (hunk 2 above)
Comment (camlow325) [inline src/ruby/puppet-server-lib/puppet/server/oids.rb]: I think it would be good to add a comment in here somewhere which points to the location of the file in the Ruby Puppet repo that this implementation was copied from.

File: test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj [hunk 5]
```diff
@@ -273,4 +279,18 @@
...
      (let [catalog (testutils/get-catalog)]
-       (is (= "production" (get catalog "version")))))))
\ No newline at end of file
+       (is (= "production" (get catalog "version")))))))
+
+(deftest ^:integration custom-trusted-oid-mapping-test
+  (testing "custom_trusted_oid_mapping works properly"
+    (testutils/write-site-pp-file
+      "notify { 'trusted_hash':\n\tmessage => \"${trusted['extensions']}\",\n}")
+    (fs/copy custom-oid-mapping (fs/file testutils/conf-dir "custom_trusted_oid_mapping.yaml"))
```
camlow325: I think it would be good to ensure that these files get cleaned up after the test is done running.

One option might be to just add some variables in the `puppet.conf` in the `general_puppet_int_test` dev-resources directory which point at the specific locations of the `csr-attributes` and `custom-trusted-oid-mapping` files.  Then when the `puppet.conf` file is cleaned out by the text fixture already in place for this file, any tests run outside of this namespace wouldn't use these extra files.

test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
```diff
@@ -273,4 +279,18 @@
...
+    (fs/copy csr-attributes (fs/file testutils/conf-dir "csr_attributes.yaml"))
+    (fs/delete-dir (fs/file testutils/conf-dir "ssl"))
+    (bootstrap/with-puppetserver-running
+      app
+      {}
+      (let [catalog (testutils/get-catalog)]
+        (is (= "{sf => Burning Finger, short => 22}"
+               (get-in (nth (get catalog "resources") 3) ["parameters" "message"])))))))
```
camlow325: Is the order in which this resource appears guaranteed?  Might be better to look for the first resource matching a title of `trusted_hash` and validating the content within it?

src/ruby/puppet-server-lib/puppet/server/certificate.rb
cprice404: Further, if we load the custom OID file from Ruby, doesn't that mean that there will be no clean way for @nathanielksmith to access it for TK auth unless he duplicates the parsing logic?  Maybe we feel like the parsing logic is simple enough to tolerate the duplication?

camlow325:
:+1: on this overall.  From bench testing, I saw that the short names for my custom-named oids were being returned in my catalog requests via `$trusted['extensions']` references placed in the compiled manifest.

The couple of things I'd recommend looking at are ensuring that the trusted oid mapping file only gets parsed no more than once per JRuby initialization and that the server can still handle the absence of a trusted oid mapping file.

This whole code path also reminded me of how the core Ruby Puppet code handles the registration of custom OID mnemonics, basically relying upon the `OpenSSL` gem.

OIDs are registered here...

https://github.com/puppetlabs/puppet/blob/4.4.1/lib/puppet/ssl/oids.rb#L115

... and extracted from here ...

https://github.com/puppetlabs/puppet/blob/4.4.1/lib/puppet/ssl/certificate.rb#L87

We've consciously been trying to avoid having any code running under Ruby via Puppet Server use the OpenSSL gem because of various issues we'd found working with it earlier on in JRuby.  This PR furthers the approach of not having Puppet Server use Ruby's OpenSSL, which seems good.

Since the OpenSSL oid registrations are essentially global, though, it would seem that any arbitrary code running under Ruby might just try to use OpenSSL to handle the OID -> short name translation.  The code above in certificate.rb is essentially replaced under Puppet Server via Puppet Server's own certificate.rb.  I grepped through the Ruby code, though, and only found one other place that does an `OpenSSL::ASN1.decode`, here:

https://github.com/puppetlabs/puppet/blob/4.4.1/lib/puppet/ssl/certificate_request.rb#L123

I don't think the above code runs under Puppet Server since Puppet Server's CA is hopefully the only portion of the stack that deals with certificate request objects.

That said, though, I wonder if we might want to look into filing a follow-on ticket to Ruby Puppet to have a class / module do a one-time stash of the oid mappings with a getter method that could be called to obtain that list when needed at various points and then modifying existing code in Puppet to use that new mapper class / module to do the translation rather than OpenSSL.  I worry at some point about the possibility of some other Ruby-level code being introduced that tries to use OpenSSL directly to do the OID -> name translation and having that unknowingly be broken when run under Puppet Server.

src/ruby/puppet-server-lib/puppet/server/certificate.rb
camlow325: We could make the same argument for this as for the csr attributes yaml file.  We duplicated the logic for parsing it in Puppet Server Clojure code even though Ruby-level Puppet has its own implementation as well.

It seems like there will continue to be a need for Ruby-level Puppet to have to put together an array of the trusted oids - built-in Puppet ones + user-defined ones - until such time as we'd be able to eliminate any of the Ruby-level code paths for certificate requests (even from the agent side).

That said, though, having a single path in the Clojure code to parse the complete list of built-in + custom oid -> name mappings and be able to inject that into both tk-auth and down to the Puppet Server certificate class in Ruby seems like it could be a good thing - even if a shadow Ruby Puppet side implementation continues to exist to support agent / certificate requests for some period of time.  We could also try to consolidate everything into a Ruby Puppet level implementation that the Clojure services mine data from but I'm not sure that's worth it if we're trying to peel more of this code away from Ruby Puppet in the long term.

cprice404: The difference I see here vs. the csr attributes is that that code already lived in core Puppet.  Here we're introducing net new parsing logic in the Puppet Server repo so I feel like we have a lot more control and flexibility over it.

I'm pretty strongly opposed to have the Clojure services reach into the Ruby world to get the data.

I'm not seeing a reason why we couldn't parse the file a single time, in Clojure, at startup, and then just inject the required data into the Ruby containers?

camlow325: What you're saying makes sense.

fpringvaldsen: @cprice404 I think that makes sense, the reason I didn't do it here is I didn't realize that was an option (I haven't done much with our JRuby stuff).

I think it would be worth refactoring all of this so we can parse the file in Clojure (I already have some code locally that's doing this) and injecting it into JRuby.

cprice404: Cool.  To be clear, I'm just doing a drive-by on this PR, so while this is my gut reaction, I don't mind deferring to others who've had more time to think about it if anyone disagrees.

fpringvaldsen:
Closing out this PR, I am planning to refactor this so that we do all the parsing in the clojure code instead.

fpringvaldsen:
Re-opening, as we've decided doing things in the ruby code is the only reasonable approach.

src/ruby/puppet-server-lib/puppet/server/oids.rb
```diff
@@ -0,0 +1,31 @@
+module Puppet::Server::Oids
```
haus: As I mentioned below, it would probably be a good idea to add a `require 'yaml'` here.

fpringvaldsen: @haus I think this whole file is going to go away once https://github.com/puppetlabs/puppet/pull/4827 gets merged (sorry, should have made that clear on the PR)

fpringvaldsen:
This is currently blocked on https://github.com/puppetlabs/puppet/pull/4827. Once that is merged, I'm planning to update our submodule dependency and remove some of the code added in this PR.

src/ruby/puppet-server-lib/puppet/server/oids.rb
haus: ah, okay. cool. i'll wait on more review for now then.

joshcooper:
@fpringvaldsen we merged the puppet PR, you should be good to go

fpringvaldsen:
Okay, all the feedback on this has been addressed, and the Puppet changes have been pulled in, so this is all ready for review/merge.

src/ruby/puppet-server-lib/puppet/server/puppet_config.rb
```diff
@@ -53,5 +53,15 @@ def self.initialize_puppet(puppet_config)
 
     Puppet::ApplicationSupport.configure_indirector_routes("master")
 
+    oid_defns = Puppet::SSL::Oids.parse_custom_oid_file(Puppet[:trusted_oid_mapping_file])
+    if oid_defns.nil?
```
haus: This would be more idiomatic ruby if the conditional was reversed and the if were just `if oid_defns`.

haus: Like this...

```ruby
if oid_defns
  @@oid_defns = Puppet::SSL::Oids::PUPPET_OIDS + oid_defns
else
  @@oid_defns = Puppet::SSL::Oids::PUPPET_OIDS
end
```

haus:
I'm :+1: with one nitpick about reversing a conditional.

fpringvaldsen:
@haus I've addressed that conditional

haus:
:+1: from me.

KevinCorcoran:
This should have been merged to `stable` instead of `master` since we need to ship it in the 2.3.2 release.  @fpringvaldsen would you mind opening a PR against `stable` which just cherry-picks this commit?

fpringvaldsen:
@KevinCorcoran Sure, I'll get to work on that

fpringvaldsen:
@KevinCorcoran https://github.com/puppetlabs/puppet-server/pull/1006

PR #997, opened 2016-03-30
(maint) Return 400 for incomplete URL to certificate_statuses

justinstoller:
Previously, we were falling through to the default handler of our
`web-routes` if the `certificate_statuses` endpoint did not include its
required -- but ignored -- trailing key segment.  The default handler
returns a 404.

This provides a default handler for the `certificate_statuses` endpoint
that will catch malformed URLs and return a 400 instead of a 404.

It also extracts providing a generic 400 response into a method that
is reused where appropriate.

### DIFF
--- a/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
+++ b/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
@@ -28,10 +28,10 @@
 ;;; 'handler' functions for HTTP endpoints
 
 (defn handle-server-error
-  ([message]
-    (-> (rr/response message)
-        (rr/status 400)
-        (rr/content-type "text/plain"))))
+  [message]
+  (-> (rr/response message)
+      (rr/status 400)
+      (rr/content-type "text/plain")))
 
 (defn handle-get-certificate
   [subject {:keys [cacert signeddir]}]
--- a/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
+++ b/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
@@ -27,6 +27,12 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; 'handler' functions for HTTP endpoints
 
+(defn handle-server-error
+  ([message]
+    (-> (rr/response message)
+        (rr/status 400)
+        (rr/content-type "text/plain"))))
+
 (defn handle-get-certificate
   [subject {:keys [cacert signeddir]}]
   (-> (if-let [certificate (ca/get-certificate subject cacert signeddir)]
@@ -51,9 +57,7 @@
     (catch ca/csr-validation-failure? {:keys [message]}
       (log/error message)
       ;; Respond to all CSR validation failures with a 400
-      (-> (rr/response message)
-          (rr/status 400)
-          (rr/content-type "text/plain")))))
+      (handle-server-error message))))
 
 (defn handle-get-certificate-revocation-list
   [{:keys [cacrl]}]
@@ -262,8 +266,10 @@
     (comidi/context ["/v1"]
       (ANY ["/certificate_status/" :subject] [subject]
         (certificate-status subject ca-settings))
-      (ANY ["/certificate_statuses/" [#"[^/]+" :ignored-but-required]] []
-        (certificate-statuses ca-settings))
+      (comidi/context ["/certificate_statuses/"]
+        (ANY [[#"[^/]+" :ignored-but-required]] []
+          (certificate-statuses ca-settings))
+        (ANY [""] [] (handle-server-error "Missing URL Segment")))
       (GET ["/certificate/" :subject] [subject]
         (handle-get-certificate subject ca-settings))
       (comidi/context ["/certificate_request/" :subject]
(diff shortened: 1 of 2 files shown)


justinstoller:
Per suggestion in https://github.com/puppetlabs/puppet-server/pull/980#discussion_r56898773

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -27,6 +27,12 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; 'handler' functions for HTTP endpoints
 
+(defn handle-server-error
+  ([message]
+    (-> (rr/response message)
+        (rr/status 400)
+        (rr/content-type "text/plain"))))
+
```
KevinCorcoran: The syntax here is incorrect.  The `[message]` should be outside of the 2nd set of parens.

KevinCorcoran: How is this not causing a test failure?  :confused:

haus: @KevinCorcoran it's probably as if there were multiple signatures for the function, except only one is defined

justinstoller: Ahh, yeah. I first created this as a fn that took zero arguments, then added a one arg form to use it in `handle-put-certificate-request!`. Then added a message to my usage of it in `web-routes`, deleted the zero arg form, but forgot to remove the surrounding parens from it. I did run the tests for this file at each step though... :\

I'll push an update with the extra parens removed.

fpringvaldsen:
:see_no_evil:

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -27,6 +27,12 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; 'handler' functions for HTTP endpoints
 
+(defn handle-server-error
```
haus: This looks similar in some ways to https://github.com/puppetlabs/puppet-server/blob/0f8cd3602517477a02140bfd961de6a785800a04/src/clj/puppetlabs/puppetserver/jruby_request.clj#L9-L13, which throws a slingshot error that triggers a 400 response. I wonder if we should be consolidating these in some way.

justinstoller: Oooh, good find.

I think the implementation of that is better but since that is also tied to a [jruby specific wrapper](https://github.com/puppetlabs/puppet-server/blob/0f8cd3602517477a02140bfd961de6a785800a04/src/clj/puppetlabs/puppetserver/jruby_request.clj#L43-L51) and would need to refactor out the generally useful part of the wrapper, and that this was intended as a small maint patch, I would lean towards not consolidating them at this time.

I think that same code is also _very_ applicable to [this handler](https://github.com/puppetlabs/puppet-server/pull/998/files#diff-bb1d8f2d842888913c49c1dddc70a5eaR64) that I just wrote as well. Since that's a bit more meaty ticket I _think_ I'd prefer for refactoring as part of that or creating a ticket for the refactoring.

One of the tickets in my working-remote-queue was [TK-214](https://tickets.puppetlabs.com/browse/TK-214) which would consolidate ring middleware/handler helpers. I _think_ my preference would be to either merge this PR and #998 and then work on TK-214 next, or merge this ticket and then do a little bit of refactoring to use the code you found in #998 if we won't get to TK-214 soon.


And of course, when I say preference above I mean more of a musing on your last sentence, which I took as an earnest question -- just let me know if you meant it more as an indirect imperative.  (;

haus: earnest question

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
KevinCorcoran: Oh, phew.  So the syntax is not actually incorrect, just unconventional.

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
```diff
@@ -309,7 +309,7 @@
           (let [response (test-app
                           {:uri "/v1/certificate_statuses/"
                            :request-method :get})]
-            (is (= 404 (:status response)))))
+            (is (= 400 (:status response)))))
```
KevinCorcoran: It's probably a good idea to make this test also check the response body.

KevinCorcoran:
:see_no_evil: just one last comment on adding an assertion to the test

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
justinstoller: Right on, added them in 88bd248

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
justinstoller: Cool, updated to the proper syntax in 336d657

justinstoller:
Awesome, updated according to comments!

PR #998, opened 2016-03-31
(SERVER-977) DELETE handler for certificate_request endpoint

justinstoller:
Prior to this there was no handler for the DELETE method of
certificate_request endpoint.

This adds a basic handler for deleting a pending certificate request.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -1013,7 +1013,7 @@
       (let [msg (format
                   "No certificate request for %s at expected path %s"
                   subject csr-path)]
-        (log/info msg)
+        (log/warn msg)
         {:outcome :not-found
          :message msg }))))
 
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -991,11 +991,11 @@
         (autosign-certificate-request! subject csr settings)
         (fs/delete (path-to-cert-request csrdir subject))))))
 
-(schema/defn ^:always-validate process-csr-deletion!
+(schema/defn ^:always-validate delete-certificate-request!
   :- {:outcome (schema/enum :success :not-found :error) :message schema/Str}
   "Delete pending certificate requests for subject"
-  [subject :- schema/Str
-   {:keys [csrdir]} :- CaSettings]
+  [{:keys [csrdir]} :- CaSettings
+   subject :- schema/Str]
   (let [csr-path (path-to-cert-request csrdir subject)]
     (if (fs/exists? csr-path)
       (if (fs/delete csr-path)
@@ -1008,8 +1008,8 @@
           {:outcome :error
            :message msg}))
       (let [msg (format
-                   "No certificate request for %s at expected path %s"
-                   subject csr-path)]
+                  "No certificate request for %s at expected path %s"
+                  subject csr-path)]
         (log/info msg)
         {:outcome :not-found
          :message msg }))))
@@ -1206,15 +1206,11 @@
         message))))
 
 (schema/defn ^:always-validate delete-certificate!
-  "Delete the certificate and/or CSR for the given subject.
+  "Delete the certificate for the given subject.
    Note this does not revoke the certificate."
   [{:keys [csrdir signeddir]} :- CaSettings
    subject :- schema/Str]
-  (let [cert (path-to-cert signeddir subject)
-        csr  (path-to-cert-request csrdir subject)]
+  (let [cert (path-to-cert signeddir subject)]
     (when (fs/exists? cert)
       (fs/delete cert)
-      (log/debug "Deleted certificate for" subject))
-    (when (fs/exists? csr)
-      (fs/delete csr)
-      (log/debug "Deleted certificate request for" subject))))
+      (log/debug "Deleted certificate for" subject))))
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -991,6 +991,29 @@
         (autosign-certificate-request! subject csr settings)
         (fs/delete (path-to-cert-request csrdir subject))))))
 
+(schema/defn ^:always-validate process-csr-deletion!
+  :- {:outcome (schema/enum :success :not-found :error) :message schema/Str}
+  "Delete pending certificate requests for subject"
+  [subject :- schema/Str
+   {:keys [csrdir] :as settings} :- CaSettings]
+  (let [csr-path (path-to-cert-request csrdir subject)]
+    (if (fs/exists? csr-path)
+      (if (fs/delete csr-path)
+        (let [msg (str "Deleted certificate request for " subject)]
+          (log/debug msg)
+          {:outcome :success
+           :message msg})
+        (let [msg (format "Path %s exists but could not be deleted" csr-path)]
+          (log/error msg)
+          {:outcome :error
+           :message msg}))
+      (let [msg (format
+                   "No certificate request for %s at expected path %s"
+                   subject csr-path)]
+        (log/info msg)
+        {:outcome :not-found
+         :message msg }))))
+
 (schema/defn ^:always-validate
   get-certificate-revocation-list :- schema/Str
   "Given the value of the 'cacrl' setting from Puppet,
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -995,7 +995,7 @@
   :- {:outcome (schema/enum :success :not-found :error) :message schema/Str}
   "Delete pending certificate requests for subject"
   [subject :- schema/Str
-   {:keys [csrdir] :as settings} :- CaSettings]
+   {:keys [csrdir]} :- CaSettings]
   (let [csr-path (path-to-cert-request csrdir subject)]
     (if (fs/exists? csr-path)
       (if (fs/delete csr-path)
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -97,6 +97,10 @@
 (def CertificateRevocationList
   (schema/pred utils/certificate-revocation-list?))
 
+(def OutcomeInfo
+  "Generic map of outcome & message for API consumers"
+  {:outcome (schema/enum :success :not-found :error) :message schema/Str})
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Definitions
 
@@ -991,8 +995,7 @@
         (autosign-certificate-request! subject csr settings)
         (fs/delete (path-to-cert-request csrdir subject))))))
 
-(schema/defn ^:always-validate delete-certificate-request!
-  :- {:outcome (schema/enum :success :not-found :error) :message schema/Str}
+(schema/defn ^:always-validate delete-certificate-request! :- OutcomeInfo
   "Delete pending certificate requests for subject"
   [{:keys [csrdir]} :- CaSettings
    subject :- schema/Str]
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -99,7 +99,8 @@
 
 (def OutcomeInfo
   "Generic map of outcome & message for API consumers"
-  {:outcome (schema/enum :success :not-found :error) :message schema/Str})
+  {:outcome (schema/enum :success :not-found :error)
+   :message schema/Str})
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Definitions
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -1212,7 +1212,7 @@
 (schema/defn ^:always-validate delete-certificate!
   "Delete the certificate for the given subject.
    Note this does not revoke the certificate."
-  [{:keys [csrdir signeddir]} :- CaSettings
+  [{signeddir :signeddir} :- CaSettings
    subject :- schema/Str]
   (let [cert (path-to-cert signeddir subject)]
     (when (fs/exists? cert)
--- a/test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
+++ b/test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
@@ -142,7 +142,8 @@
             (fs/delete expected-path))))
       (testing "Attempted deletion of a non-existant CSR"
         (let [response (handle-delete-certificate-request! subject settings)
-              msg-matcher (re-pattern (str "No cert.* " subject " at .*"))]
+              msg-matcher (re-pattern
+                            (str "No cert.*request for " subject " at.*" expected-path))]
           (is (false? (fs/exists? expected-path)))
           (is (= 404 (:status response)))
           (is (re-matches msg-matcher (:body response)))
--- a/test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
+++ b/test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
@@ -148,35 +148,35 @@
               (is (logged? msg-matcher :debug)))
             (finally
               (fs/delete expected-path))))))
-      (testing "Attempted deletion of a non-existant CSR"
-        (logutils/with-test-logging
-          (let [subject "not-found-agent"
-                response (handle-delete-certificate-request! subject settings)
-                expected-path (ca/path-to-cert-request (:csrdir settings) subject)
-                msg-matcher (re-pattern
-                              (str "No cert.*request for " subject " at.*" expected-path))]
-            (is (false? (fs/exists? expected-path)))
-            (is (= 404 (:status response)))
-            (is (re-matches msg-matcher (:body response)))
-            (is (= "text/plain" (get-in response [:headers "Content-Type"])))
-            (is (logged? msg-matcher :warn)))))
-      (testing "Error during deletion of a CSR"
-        (logutils/with-test-logging
-          (let [subject "err-agent"
-                csr-stream (gen-csr-input-stream! subject)
-                expected-path (ca/path-to-cert-request (:csrdir settings) subject)]
-            (try
-              (handle-put-certificate-request! subject csr-stream settings)
-              (is (true? (fs/exists? expected-path)))
-              (fs/chmod "-w" (fs/parent expected-path))
-              (let [response (handle-delete-certificate-request! subject settings)
-                     msg-matcher (re-pattern (str "Path " expected-path " exists but.*"))]
-                 (is (= 500 (:status response)))
-                 (is (re-matches msg-matcher (:body response)))
-                 (is (= "text/plain" (get-in response [:headers "Content-Type"])))
-                 (is (logged? msg-matcher :error)))
-              (finally
-                (fs/chmod "+w" (fs/parent expected-path)))))))))
+    (testing "Attempted deletion of a non-existant CSR"
+      (logutils/with-test-logging
+        (let [subject "not-found-agent"
+              response (handle-delete-certificate-request! subject settings)
+              expected-path (ca/path-to-cert-request (:csrdir settings) subject)
+              msg-matcher (re-pattern
+                            (str "No cert.*request for " subject " at.*" expected-path))]
+          (is (false? (fs/exists? expected-path)))
+          (is (= 404 (:status response)))
+          (is (re-matches msg-matcher (:body response)))
+          (is (= "text/plain" (get-in response [:headers "Content-Type"])))
+          (is (logged? msg-matcher :warn)))))
+    (testing "Error during deletion of a CSR"
+      (logutils/with-test-logging
+        (let [subject "err-agent"
+              csr-stream (gen-csr-input-stream! subject)
+              expected-path (ca/path-to-cert-request (:csrdir settings) subject)]
+          (try
+            (handle-put-certificate-request! subject csr-stream settings)
+            (is (true? (fs/exists? expected-path)))
+            (fs/chmod "-w" (fs/parent expected-path))
+            (let [response (handle-delete-certificate-request! subject settings)
+                   msg-matcher (re-pattern (str "Path " expected-path " exists but.*"))]
+               (is (= 500 (:status response)))
+               (is (re-matches msg-matcher (:body response)))
+               (is (= "text/plain" (get-in response [:headers "Content-Type"])))
+               (is (logged? msg-matcher :error)))
+            (finally
+              (fs/chmod "+w" (fs/parent expected-path)))))))))
 
 (deftest handle-put-certificate-request!-test
   (let [settings   (assoc (testutils/ca-sandbox! cadir)
--- a/test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
+++ b/test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
@@ -119,6 +119,44 @@
           response (ring-app request)]
       (is (= version-number (get-in response [:headers "X-Puppet-Version"]))))))
 
+(deftest handle-delete-certificate-request!-test
+  (let [settings      (assoc (testutils/ca-sandbox! cadir)
+                             :allow-duplicate-certs true
+                             :autosign false)
+        subject       "foo-agent"
+        csr-stream    (io/input-stream (test-pem-file "foo-agent-csr.pem"))
+        expected-path (ca/path-to-cert-request (:csrdir settings) subject)
+        assert-proper
+          (fn
+            [response code matcher log-level]
+            (do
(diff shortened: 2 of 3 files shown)


justinstoller:
I have a more basic question about the appropriateness of adding this handler as mentioned in [SERVER-977](https://tickets.puppetlabs.com/browse/SERVER-977)

The flow from service core to actual certificate authority library apes the flow of the put method.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -991,6 +991,29 @@
         (autosign-certificate-request! subject csr settings)
         (fs/delete (path-to-cert-request csrdir subject))))))
 
+(schema/defn ^:always-validate process-csr-deletion!
+  :- {:outcome (schema/enum :success :not-found :error) :message schema/Str}
```
justinstoller: I put a schema on the return value since the consumer of this method is fairly tightly coupled to it's return values and this seemed like a great way to enforce and document the expectations between them. I did not separate out the schema definition into a separate `def` because the data that it returns isn't passed through to other methods or used by other consumers, so it seemed like a pre-mature optimization.

That's _why_ I did what I did, I'm making this note because I'm sure if that's what I should be doing and from the lack of seeing similar instances in the code I'm afraid I'm going against the grain?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -991,6 +991,29 @@
...
+  [subject :- schema/Str
+   {:keys [csrdir] :as settings} :- CaSettings]
+  (let [csr-path (path-to-cert-request csrdir subject)]
+    (if (fs/exists? csr-path)
+      (if (fs/delete csr-path)
+        (let [msg (str "Deleted certificate request for " subject)]
+          (log/debug msg)
+          {:outcome :success
+           :message msg})
+        (let [msg (format "Path %s exists but could not be deleted" csr-path)]
```
justinstoller: Is sending the path on disk across the wire a bad idea? It seems important for logging purposes and I didn't like the idea of having two custom messages per outcome.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -991,6 +991,29 @@
...
 
+(schema/defn ^:always-validate process-csr-deletion!
+  :- {:outcome (schema/enum :success :not-found :error) :message schema/Str}
+  "Delete pending certificate requests for subject"
+  [subject :- schema/Str
+   {:keys [csrdir] :as settings} :- CaSettings]
+  (let [csr-path (path-to-cert-request csrdir subject)]
+    (if (fs/exists? csr-path)
+      (if (fs/delete csr-path)
+        (let [msg (str "Deleted certificate request for " subject)]
```
justinstoller: Sending a response body and logging seemed like it may be excessive. I could come up with a few diffirent criteria for when to send to a log, a response, or both but I wanted to hear what others thought, and frankly, if people thought that it was worth the effort to do different things for different outcomes.

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
```diff
@@ -119,6 +119,44 @@
...
+          (finally
+            (fs/delete expected-path))))
+      (testing "Attempted deletion of a non-existant CSR"
+        (let [response (handle-delete-certificate-request! subject settings)
+              msg-matcher (re-pattern (str "No cert.* " subject " at .*"))]
+          (is (false? (fs/exists? expected-path)))
+          (assert-proper response 404 msg-matcher :info)))
+      (testing "Error during deletion of a CSR"
+        (let [response
+              (with-redefs [me.raynes.fs/exists? (constantly true)]
```
justinstoller: Mocking out a function seems less than ideal in Ruby and there it's idiomatic. Here it felt super gross, but I couldn't think of a better way to do this. It did make me question whether or not this outcome branch was even necessary, though. Is it? If not, does anyone know of a better way to test it offhand?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -991,6 +991,29 @@
         (autosign-certificate-request! subject csr settings)
         (fs/delete (path-to-cert-request csrdir subject))))))
 
+(schema/defn ^:always-validate process-csr-deletion!
```
justinstoller: After writing most of this guy I found [`delete-certificate!`](https://github.com/puppetlabs/puppet-server/pull/998/commits/a635f9fd21a04047f25433e21daad4a58601f361#diff-feeac80211305477061ba2a798669e8dR1208). It does a _very_ similar thing, deleting the cert request if it exists, it also deletes a signed cert if it exists, but doesn't return any information about what it did or didn't end up deleting (or why).

It seems like these should share code, but I also was surprised that these two endpoints actions weren't mutually exclusive... Folks' feedback would guide which method I'd collapsed into which -- eg.
1. make this simpler by reusing the existing `delete-certificate!` fn and not returning as much info,
2. remove the csr deletion from `delete-certificate!` and have the certificate_statuses endpoint call both `delete-certificate!` and this fn or
3. make the delete-certificate fn more like this one (though still remove the csr deletion) and return more info from the endpoint (we don't currently return the documented information -- though the [documented response](https://docs.puppetlabs.com/puppet/4.2/reference/http_api/http_certificate_status.html#delete) is a very Ruby specific).



I actually looked at the certificate_status delete endpoint code but didn't go to the library implementation assuming deleting a cert and deleting a cert request would be fundamentally different things if they had different API endpoints. It wasn't until a bad grep for my code that I ran across it and after seeing the implementation went back to the docs for that endpoint...

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -991,6 +991,29 @@
...
+  :- {:outcome (schema/enum :success :not-found :error) :message schema/Str}
+  "Delete pending certificate requests for subject"
+  [subject :- schema/Str
+   {:keys [csrdir] :as settings} :- CaSettings]
```
haus: Since you don't use settings anywhere, you can remove `:as settings` from this.

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
haus: Could you change the ownership or make the directory unwritable? Anything that would make the delete fail seems like it would work.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
justinstoller: Oh, cool. I'll update. I have to admit advanced destructuring is something that I still stumble with.

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
justinstoller: I manually tested it by setting the immutable flag on the file. I didn't get it fail to delete by setting it to be owned by root. 

I wasn't sure what I could expect as far as sudo, chowning, and user availability on other's workstations and test runners or cross platform setting of immutable flags. And I haven't found prior art in the test suite to glean from, either. So I shied away from touching the fs directly. I could go in that direction if my there's a reliable, cross platform way to do that, do you have any recommendations (I'm not seeing how to in me.raynes.fs or java.io.File).

KevinCorcoran: What about just deleting the file here?  That would probably cause the second delete attempt from the production code to fail, I suspect.

haus: changing ownership wouldn't be sufficient unless you also made it only root writable. you should be able to remove write permissions from the file using http://raynes.github.io/fs/me.raynes.fs.html#var-chmod (haven't tried this), but then you would need to give write permissions back and clean it up after the test.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
justinstoller: Updated

cprice404: Just FYI - sometimes I leave the `:as` part even if it's unused, because it can be useful as documentation to the next developer when they are wondering what is passed in.  Not always necessary, but sometimes it is useful.  (In this case I think you are fine either way.)

src/clj/puppetlabs/puppetserver/certificate_authority.clj
nathanielksmith: Not having a ton of familiarity with the code, my gut votes for 2; it sounds like delete-certificate! is doing double duty and having two explicit functions seems better than one function with implied behavior.

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
justinstoller: I'm not sure why clj is still able to delete this file but I can't get changing perms on the file from clj to work:
```clj
user=> (fs/writeable? "dev-resources/puppetlabs/services/ca/certificate_authority_core_test/pems/foo-agent-csr.pem")
true
user=> (fs/chmod "-w" "dev-resources/puppetlabs/services/ca/certificate_authority_core_test/pems/foo-agent-csr.pem")
"dev-resources/puppetlabs/services/ca/certificate_authority_core_test/pems/foo-agent-csr.pem"
user=> (fs/writeable? "dev-resources/puppetlabs/services/ca/certificate_authority_core_test/pems/foo-agent-csr.pem")
false
user=> (fs/delete "dev-resources/puppetlabs/services/ca/certificate_authority_core_test/pems/foo-agent-csr.pem")
true
user=> (fs/exists? "dev-resources/puppetlabs/services/ca/certificate_authority_core_test/pems/foo-agent-csr.pem")
false
```
While in the other window:
```
15:11:25 hobo:pems justin (bug/master/SERVER-977_add_cert_request_delete *%) ‖ ls -l
total 104
...snip...
-rw-r--r--  1 justin  staff   985 Apr  5 15:11 foo-agent-csr.pem

15:11:28 hobo:pems justin (bug/master/SERVER-977_add_cert_request_delete *%) ‖ ls -l
total 104
...snip...
-r--r--r--  1 justin  staff   985 Apr  5 15:11 foo-agent-csr.pem


15:19:40 hobo:pems justin (bug/master/SERVER-977_add_cert_request_delete *%) ‖ ls | grep foo-agent
15:20:06 hobo:pems justin (bug/master/SERVER-977_add_cert_request_delete *%) ‖

```

src/clj/puppetlabs/puppetserver/certificate_authority.clj
justinstoller: Right on, updated in d91d90f

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
justinstoller: So I'm able to reproduce this behavior in the repl bypassing me.raynes.fs:
```
16:00:24 hobo:pems justin (bug/master/SERVER-977_add_cert_request_delete *%) ‖ ls -l foo-agent-csr.pem
-rw-r--r--  1 justin  staff  985 Apr  5 15:24 foo-agent-csr.pem
```

```clj
user=> (java.nio.file.Files/isWritable (java.nio.file.Paths/get (.toURI fp)))
true
user=> (.canWrite fp)
true
user=> (.setWritable fp false)
true
user=> (.canWrite fp)
false
user=> (java.nio.file.Files/isWritable (java.nio.file.Paths/get (.toURI fp)))
false
user=> (java.nio.file.Files/delete (java.nio.file.Paths/get (.toURI fp)))
nil
user=> (.canWrite fp)
false
```
```
16:00:28 hobo:pems justin (bug/master/SERVER-977_add_cert_request_delete *%) ‖ ls -l foo-agent-csr.pem
ls: foo-agent-csr.pem: No such file or directory
```


I don't know if my understanding of the way Java or how filesystems work is broken. Or just my filesystem and Java are having a hard time playing nicely.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
KevinCorcoran: As is, this is certainly out-of-place with our existing code, as far as I'm aware - I've observed we have a strong tendency to introduce a `def` for this schema instead of putting it inline like this.

That said, I don't have actually have a problem with it like this, and we don't have any well-defined best practice that I'm aware of.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
KevinCorcoran: It seems fine to me, especially since it's just a path to a CSR.  If an attacker has root access on the master I doubt that knowing a path to a CSR that couldn't be deleted makes things any worse for us.

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
```diff
@@ -119,6 +119,44 @@
...
 
+(deftest handle-delete-certificate-request!-test
+  (let [settings      (assoc (testutils/ca-sandbox! cadir)
+                             :allow-duplicate-certs true
+                             :autosign false)
+        subject       "foo-agent"
+        csr-stream    (io/input-stream (test-pem-file "foo-agent-csr.pem"))
+        expected-path (ca/path-to-cert-request (:csrdir settings) subject)
+        assert-proper
+          (fn
```
KevinCorcoran: We have a strong tendency to just duplicate the assertions in the test instead of putting them in a helper function like this.  It tends to make the tests longer and more repetitive but much easier to debug when something breaks.

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
KevinCorcoran: In response to your 2nd most recent comment, it appears that `fs/delete` probably just corresponds to `rm -f`: 
```sh
kevin at notbroken in /tmp
$ touch foo

kevin at notbroken in /tmp
$ ls -l foo
-rw-r--r--  1 kevin  wheel  0 Apr  5 17:00 foo

kevin at notbroken in /tmp
$ chmod -w foo

kevin at notbroken in /tmp
$ ls -l foo
-r--r--r--  1 kevin  wheel  0 Apr  5 17:00 foo

kevin at notbroken in /tmp
$ rm -f foo

kevin at notbroken in /tmp
$ ls -l foo
ls: foo: No such file or directory
```

KevinCorcoran: What about returning to either @haus's idea to make the parent directory unwriteable (or unreadable) or something like that, or my suggestion to just delete the file first from the test.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -991,6 +991,29 @@
...
+          {:outcome :success
+           :message msg})
+        (let [msg (format "Path %s exists but could not be deleted" csr-path)]
+          (log/error msg)
+          {:outcome :error
+           :message msg}))
+      (let [msg (format
+                  "No certificate request for %s at expected path %s"
+                  subject csr-path)]
+        (log/info msg)
```
KevinCorcoran: I think the logging in this function is good overall, although I'd log this at `error` or `warn` instead of `info`.

justinstoller: Cool, I've updated it to `warn` in cd45bc0

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
justinstoller: Perfect, I've updated accordingly in 9178e9a

src/clj/puppetlabs/puppetserver/certificate_authority.clj
justinstoller: If there's a strong tendency and no best practice I'll go with flow. I've added a def for this in 	0d40426

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -97,6 +97,10 @@
 (def CertificateRevocationList
   (schema/pred utils/certificate-revocation-list?))
 
+(def OutcomeInfo
+  "Generic map of outcome & message for API consumers"
+  {:outcome (schema/enum :success :not-found :error) :message schema/Str})
```
haus: Generally maps are listed in key val per line.

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
justinstoller: Sorry I spaced responding to your comment about deleting the file. I think the [code branching](https://github.com/puppetlabs/puppet-server/pull/998#discussion-diff-58047866R1000) is such that we only get to the error condition if the file exists but we cannot delete it. In order to trigger it via deletion we'd have to delete the file between lines 1000 and 1001 in pl.ps.certificate-authority and I'm not sure how to do that.

I'm concerned about changing the permissions of a whole directory of static test fixtures for the same reason I'm concerned about `with-redefs`. If I understand it correctly, if the tests are ran in parallel both will cause spurious failures in unrelated tests.

I think if the options are between changing perms on the fixtures directory vs using `with-redefs`, I, naively, have a slight preference to using `with-redefs` since it's in code and I think we have more control over the code than a filesystem the tests run on.

So those are my concerns with the two options mentioned @KevinCorcoran but I'm probably missing something and don't know what I don't know.

justinstoller:
I believe I've updated the PR according to all feedback thus far, save for the testing discussion which is in progress.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
justinstoller: Doh. Updated in 2add0ed.

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
```diff
@@ -119,6 +119,45 @@
           response (ring-app request)]
       (is (= version-number (get-in response [:headers "X-Puppet-Version"]))))))
 
+(deftest handle-delete-certificate-request!-test
+  (let [settings      (assoc (testutils/ca-sandbox! cadir)
+                             :allow-duplicate-certs true
+                             :autosign false)
+        subject       "foo-agent"
+        csr-stream    (io/input-stream (test-pem-file "foo-agent-csr.pem"))
+        expected-path (ca/path-to-cert-request (:csrdir settings) subject)]
```
KevinCorcoran: let bindings [should not be aligned vertically](https://github.com/puppetlabs/pl-clojure-style#aligning-values-in-maps-and-lets)

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
```diff
@@ -119,6 +119,45 @@
...
+        (let [response (handle-delete-certificate-request! subject settings)
+              msg-matcher (re-pattern (str "No cert.* " subject " at .*"))]
+          (is (false? (fs/exists? expected-path)))
+          (is (= 404 (:status response)))
+          (is (re-matches msg-matcher (:body response)))
+          (is (= "text/plain" (get-in response [:headers "Content-Type"])))
+          (is (logged? msg-matcher :warn))))
+      (testing "Error during deletion of a CSR"
+        (let [response
+              (with-redefs [me.raynes.fs/exists? (constantly true)]
```
KevinCorcoran: `s/me.raynes.fs/fs`

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
```diff
@@ -119,6 +119,45 @@
...
+            (is (false? (fs/exists? expected-path)))
+            (is (= 204 (:status response)))
+            (is (re-matches msg-matcher (:body response)))
+            (is (= "text/plain" (get-in response [:headers "Content-Type"])))
+            (is (logged? msg-matcher :debug)))
+          (finally
+            (fs/delete expected-path))))
+      (testing "Attempted deletion of a non-existant CSR"
+        (let [response (handle-delete-certificate-request! subject settings)
+              msg-matcher (re-pattern (str "No cert.* " subject " at .*"))]
```
KevinCorcoran: Seems like this regex could be improved to more closely match the error message `"No certificate request for %s at expected path %s"`

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
KevinCorcoran: oh, i see, I missed the `with-redefs` on this line ...

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1183,15 +1210,11 @@
         message))))
 
 (schema/defn ^:always-validate delete-certificate!
-  "Delete the certificate and/or CSR for the given subject.
+  "Delete the certificate for the given subject.
    Note this does not revoke the certificate."
   [{:keys [csrdir signeddir]} :- CaSettings
```
haus: csrdir is no longer used, so it can be removed from the destructuring.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1183,15 +1210,11 @@
         message))))
 
 (schema/defn ^:always-validate delete-certificate!
```
haus: Would it be worth adding the cases (can't delete, doesn't exist) from the csr deletion to this function in a followup ticket?

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
KevinCorcoran: It looks like the CSR is underneath a directory created by `ca-sandbox!`, so hopefully that's local to this test, and therefore changing permissions on a parent directory wouldn't affect any other tests even if they were run in parallel.

justinstoller: Ahhh... I was confused. I ran into trouble renaming things away from foo-agent because the input csr is staticly shared here: https://github.com/puppetlabs/puppet-server/tree/master/dev-resources/puppetlabs/services/ca/certificate_authority_core_test/pems

But it looks like I was conflating the test inputs and the test's working dir. I'll go ahead and update this to change the perms on parent dir of the expected path.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
justinstoller: I'd be into it. We currently don't follow the documented returns for this interface, though implementing them might be blocked on moving away from liberator for the certificate_status resource (which is SERVER-264 and in my queue).

src/clj/puppetlabs/puppetserver/certificate_authority.clj
justinstoller: Removed in b75ac3c

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
justinstoller: Tightened up in 482bacb

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
justinstoller: Unaligned in 0883e0a

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
justinstoller: Moved to changing the directory permissions in a7a9fc1

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
```diff
@@ -119,6 +127,57 @@
...
+            (let [response (handle-delete-certificate-request! subject settings)
+                  msg-matcher (re-pattern (str "Deleted .* for " subject ".*"))]
+              (is (false? (fs/exists? expected-path)))
+              (is (= 204 (:status response)))
+              (is (re-matches msg-matcher (:body response)))
+              (is (= "text/plain" (get-in response [:headers "Content-Type"])))
+              (is (logged? msg-matcher :debug)))
+            (finally
+              (fs/delete expected-path))))))
+      (testing "Attempted deletion of a non-existant CSR"
```
haus: tabbing seems off here. the first testing call looks correctly indented, while this and the next appear to be indented more.

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
```diff
@@ -119,6 +127,57 @@
...
+            (is (logged? msg-matcher :warn)))))
+      (testing "Error during deletion of a CSR"
+        (logutils/with-test-logging
+          (let [subject "err-agent"
+                csr-stream (gen-csr-input-stream! subject)
+                expected-path (ca/path-to-cert-request (:csrdir settings) subject)]
+            (try
+              (handle-put-certificate-request! subject csr-stream settings)
+              (is (true? (fs/exists? expected-path)))
+              (fs/chmod "-w" (fs/parent expected-path))
```
haus: oh cool. you finally got this working?

haus:
I'm :+1: here. I just had 1 style nitpick about indentation in the tests.

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
justinstoller: yeah, I couldn't get it working by setting the permissions on the individual files and I was worried about setting the perms on a whole directory of test files. But Kevin said in another thread that it wasn't a concern since they were part a testing sandbox and to go for it.

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
justinstoller: My arch nemesis.... Code Formatting. For some reason my editor corrects some issues but misses others. I'm going to be ecstatic when we start using `cljfmt` for this project.

PR #999, opened 2016-03-31
(MAINT) fixed bad service_name variable.  Fixed permissions on test l…

erikPrime:
…ogback.xml file

### DIFF
--- a/acceptance/suites/tests/00_smoke/validate-syslog-logback.rb
+++ b/acceptance/suites/tests/00_smoke/validate-syslog-logback.rb
@@ -1,10 +1,6 @@
 test_name 'SERVER-1215: Validate that logback can be configured to work with syslog'
 
-if options[:type] == 'pe' then
-  service_name   = 'pe-puppetserver'
-else
-  service_name   = 'puppetserver'
-end
+service   = options['puppetservice']
 
 logback_path    = '/etc/puppetlabs/puppetserver/logback.xml'
 logback_backup  = '/etc/puppetlabs/puppetserver/logback.back'
@@ -46,8 +42,8 @@ EOM
 
 teardown do
   on(master, "mv #{logback_backup} #{logback_path}")
-  on(master, "service #{service_name} restart")
-  on(master, "service #{service_name} status")
+  on(master, "service #{service} restart")
+  on(master, "service #{service} status")
 end
 
 step 'Backup logback'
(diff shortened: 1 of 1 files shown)


jpinsonault:
:+1:

acceptance/suites/tests/00_smoke/validate-syslog-logback.rb
```diff
@@ -1,9 +1,9 @@
 test_name 'SERVER-1215: Validate that logback can be configured to work with syslog'
 
 if options[:type] == 'pe' then
-  servicename   = 'pe-puppetserver'
+  service_name   = 'pe-puppetserver'
```
camlow325: Can we handle the service name selection by just using `options['puppetservice']` like we do here?

https://github.com/puppetlabs/puppet-server/blob/puppet-server-2.3.1/acceptance/suites/tests/010-puppetserver-cli/subcommand/foreground.rb#L8

acceptance/suites/tests/00_smoke/validate-syslog-logback.rb
```diff
@@ -50,15 +50,17 @@
...
   on(master, "mv #{logback_path} #{logback_backup}")
 
-step 'Modify logback configuration'
+step 'Modify logback configuration' do
   create_remote_file(master, logback_path, logback_config)
+  on(master, "chmod +r #{logback_path}")
+end
 
 step 'Restart puppetserver'
   on(master, "service #{service_name} restart")
```
camlow325: I'd been thinking we wanted to use `puppet("resource service #{service} ensure=stopped")` ... `ensure=running` to handle the restarts instead of doing the lower-level `service` command because it allows for the tests to be run under a wider swath of OSes in a more compatible way.  Is that not the case?  Not sure if that approach would also help avoiding the extra `sleep` below.  If it does, I'd definitely be more in favor of that.  The `sleep` seems more risky to me.

acceptance/suites/tests/00_smoke/validate-syslog-logback.rb
```diff
@@ -50,15 +50,17 @@
...
 step 'Backup logback'
   on(master, "mv #{logback_path} #{logback_backup}")
 
-step 'Modify logback configuration'
+step 'Modify logback configuration' do
   create_remote_file(master, logback_path, logback_config)
+  on(master, "chmod +r #{logback_path}")
+end
 
 step 'Restart puppetserver'
```
camlow325: Should there be a do / end block around this since there are multiple lines inside of it?

PR #1002, opened 2016-04-06
(PUP-5482) Toggle always_retry_plugins when available

haus:
A performance improvement was just merged into puppet. It is guarded by
a setting called `always_retry_plugins`, which defaults to true. To take
advantage of this improvement, we should set it to false. This commit
makes that change, but checks to see if the setting is available yet by
calling into `Puppet.settings.setting`.

### DIFF
--- a/spec/puppet-server-lib/puppet/jvm/master_spec.rb
+++ b/spec/puppet-server-lib/puppet/jvm/master_spec.rb
@@ -24,7 +24,7 @@ describe 'Puppet::Server::Master' do
     end
 
     it "returns the correct puppet version number" do
-      expect(subject).to eq('4.4.1')
+      expect(subject).to eq('4.4.2')
     end
   end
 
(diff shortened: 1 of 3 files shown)


KevinCorcoran:
Should there be test coverage?  Perhaps a spec test?

haus:
That the setting actually gets set?

haus:
oh hey, we do have such a test for `always_cache_features`. Sure. why not.

KevinCorcoran:
Yep, I guess that's what it would do.  Not sure if there's a great way to test that.

haus:
@KevinCorcoran That sort of test will fail unless I also pull in a newer puppet submodule, which I can absolutely do.

KevinCorcoran:
do it

haus:
i did it.

KevinCorcoran:
👍

fpringvaldsen:
:+1:

PR #1005, opened 2016-04-07
(maint) Conditionally set always_cache_features

haus:
The new always_retry_plugins replaces always_cache_features, which will
be deprecated in the next minor release of puppet. To avoid deprecation
warnings, this commit updates Puppet::Server::PuppetConfig to only set
always_cache_features if always_retry_plugins is unavailable.

### DIFF
--- a/src/ruby/puppet-server-lib/puppet/server/puppet_config.rb
+++ b/src/ruby/puppet-server-lib/puppet/server/puppet_config.rb
@@ -21,6 +21,9 @@ class Puppet::Server::PuppetConfig
     )
     Puppet[:trace] = true
 
+    # This conditional can be replaced with just setting always_retry_plugins
+    # once puppetserver depends on a version of puppet that has that setting
+    # available.
     if Puppet.settings.setting('always_retry_plugins')
       Puppet[:always_retry_plugins] = false
     else
--- a/src/ruby/puppet-server-lib/puppet/server/puppet_config.rb
+++ b/src/ruby/puppet-server-lib/puppet/server/puppet_config.rb
@@ -21,12 +21,12 @@ class Puppet::Server::PuppetConfig
     )
     Puppet[:trace] = true
 
-    # (SERVER-410) Cache features in puppetserver for performance.  Avoiding
-    # the cache is intended for agents to reload features mid-catalog-run.
-    Puppet[:always_cache_features] = true
-
     if Puppet.settings.setting('always_retry_plugins')
       Puppet[:always_retry_plugins] = false
+    else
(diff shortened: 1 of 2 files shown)


haus:
You can see some examples of the deprecation warnings here: https://travis-ci.org/puppetlabs/puppet-server/jobs/121490238

src/ruby/puppet-server-lib/puppet/server/puppet_config.rb
```diff
@@ -21,12 +21,15 @@ def self.initialize_puppet(puppet_config)
...
-
+    # This conditional can be replaced with just setting always_retry_plugins
+    # once puppetserver depends on a version of puppet that has that setting
+    # available.
     if Puppet.settings.setting('always_retry_plugins')
       Puppet[:always_retry_plugins] = false
+    else
+      # (SERVER-410) Cache features in puppetserver for performance.  Avoiding
+      # the cache is intended for agents to reload features mid-catalog-run.
+      Puppet[:always_cache_features] = true
```
cprice404: Why do we need to leave this in?  Once we bump the submodule we'll know we have `always_retry_plugins` so we should just use that, right?

haus: Depends if the server and agent releases are coordinated/coupled or not. Leaving it this way would allow us to ship a Y even if puppet 4.5 hasn't shipped.

cprice404: I'd probably prefer to force the releases to be tightly coupled than to leave behind some tech debt that we have to try to remember to get rid of in the future... thoughts?

haus: That seems fine as long as we think the releases are going to line up.

cprice404: mebbe we just keep this in our pocket until we know for sure, then?  And in the meantime maybe just a patch that deletes the `:always_cache_features` line and bumps the submodule?

fpringvaldsen:
:+1:

src/ruby/puppet-server-lib/puppet/server/puppet_config.rb
haus: That seems pretty reasonable. The submodule has already been bumped.

haus:
I opened https://github.com/puppetlabs/puppet-server/pull/1008 to just remove the setting. If we end up needing to release before the next Puppet Y ships, this PR can be reconsidered.

KevinCorcoran:
Closing this since we don't need it now (after discussion with @haus), we'll re-open if we need it.

PR #1006, opened 2016-04-07
(SERVER-1150) Add support for custom oid mappings

fpringvaldsen:
Add support for Puppet's custom trusted oid mapping feature.

This is a cherry-pick of the commit in https://github.com/puppetlabs/puppet-server/pull/996


ruby/puppet
```diff
@@ -1 +1 @@
-Subproject commit 37d1d457b38d29c39c689ddd484285c7821a7194
+Subproject commit 02328a9b5617c9e0746f9403bab099035ca96492
```
KevinCorcoran: So this commit is only in puppet#master, which doesn't seem good, since I think we are trying to keep puppet-server#stable in-sync with puppet#stable.  I might just set this to be the latest commit in puppet#stable.

haus:
@fpringvaldsen the puppet agent version will need updating to a version that includes in OID work as well

fpringvaldsen:
I've updated the submodule commit and fixed the failing test

fpringvaldsen:
@haus 

> the puppet agent version will need updating to a version that includes in OID work as well

Do you mean for the CI Pipeline to pass? Do you know where I can set this?

haus:
acceptance/lib/helper.rb has a sha listed. the sha will need to be an agent version that includes the puppet changes. you can see #1007 for an example.

haus:
And yea, I meant for acceptance tests to pass.

KevinCorcoran:
submodule SHA looks good, I think this is ready except for the change @haus mentioned

fpringvaldsen:
@KevinCorcoran @haus I've updated the puppet-agent dependency, hopefully that should make CI green

KevinCorcoran:
👍

haus:
:+1: will merge once travis goes green

PR #1007, opened 2016-04-08
(maint) Update puppet-agent version

haus:
This commit updates the puppet-agent version used for acceptance testing
to include the OID work that was recently merged into puppet.


kurtwall:
🙈

PR #1008, opened 2016-04-08
(maint) Don't set always_cache_features in puppet_config

haus:
The new always_retry_plugins setting includes always_cache_features, and
setting always_cache_features now generates an unsightly deprecation
warning, so this commit removes that setting from puppet_config.

### DIFF
--- a/spec/puppet-server-lib/puppet/jvm/puppet_config_spec.rb
+++ b/spec/puppet-server-lib/puppet/jvm/puppet_config_spec.rb
@@ -21,13 +21,6 @@ describe 'Puppet::Server::PuppetConfig' do
       end
     end
 
-    describe '(SERVER-410) Puppet[:always_cache_features]' do
-      subject { Puppet[:always_cache_features] }
-      it 'is true for increased performance in puppet-server' do
-        expect(subject).to eq(true)
-      end
-    end
-
     describe '(PUP-5482) Puppet[:always_retry_plugins]' do
       subject { Puppet[:always_retry_plugins] }
       it 'is false for increased performance in puppet-server' do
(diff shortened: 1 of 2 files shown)


fpringvaldsen:
:+1:

spec/puppet-server-lib/puppet/jvm/puppet_config_spec.rb
```diff
@@ -21,13 +21,6 @@
       end
     end
 
-    describe '(SERVER-410) Puppet[:always_cache_features]' do
-      subject { Puppet[:always_cache_features] }
-      it 'is true for increased performance in puppet-server' do
-        expect(subject).to eq(true)
-      end
-    end
-
```
KevinCorcoran: Would this test still pass?  If so, perhaps it should not be deleted.

haus: This test would fail without it, as always_cache_features defaults to false.

KevinCorcoran: Oh, I see, and we are setting `always_retry_features` to be `false`.  Does that mean we're losing a performance improvement from caching?  Or am I starting to ask questions with answers that I don't really want to know?

... or ...

Is it the case that `always_retry_plugins = false` implies `always_cache_features = true`?  😕

haus: We just did this in person, but I'll jot it down and update the comment.
Having `always_retry_plugins` be false will imply the behavior that `always_cache_features = true` would cause, but doesn't actually change the value of the `always_cache_features` setting.

KevinCorcoran:
👍

PR #1009, opened 2016-04-08
Updates to beaker 2.38.1, and works around BKR-650

erikPrime:
In the recent past I attempted to update puppet-server to beaker 2.38.1.  The initial runs through CI were successful, but the nightly runs failed, because they use a larger set of test hosts, including some that do not have PuppetDB packages (fedora 21 and ubuntu 15).

Trying again.  This time, rather than relying on beakers broken confine_block (which modifies the global hosts list), I'm using a simple if statement to choose when we install PuppetDB packages.

I'll update this comment to list the layout flavors I've tested.  We shouldn't merge until I've tested on Fedora 21, Ubuntu 15, (the previously failing OSes) as well as at least one of the working ones.  Meanwhile I'd appreciate comments/review.

Tested on 
fedora21-64ma.
ubuntu1504-64ma-32a

### DIFF
--- a/acceptance/suites/pre_suite/foss/30_install_dev_repos.rb
+++ b/acceptance/suites/pre_suite/foss/30_install_dev_repos.rb
@@ -26,7 +26,7 @@ if puppet_build_version
   end
 end
 
-confine_block(:to, {:platform => puppetdb_supported_platforms}, master) do
+if puppetdb_supported_platforms.include?(master.platform) then
   step "Install PuppetDB repository" do
     install_puppetlabs_dev_repo(
       master, 'puppetdb', test_config[:puppetdb_build_version],
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ gem 'jira-ruby', :group => :development
 
 group :test do
   gem 'rspec'
-  gem 'beaker', *location_for(ENV['BEAKER_VERSION'] || '~> 2.24.0')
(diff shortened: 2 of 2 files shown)


kurtwall:
✨ but don't want to merge until master, which is red for an entirely unrelated reason, is green.

acceptance/suites/pre_suite/foss/30_install_dev_repos.rb
```diff
@@ -26,7 +26,7 @@
   end
 end
 
-confine_block(:to, {:platform => puppetdb_supported_platforms}, master) do
```
camlow325: @erikPrime Are we also going to have to change this line to do something similar - https://github.com/puppetlabs/puppetserver/blob/7875b01518a8c3c178eba20c46909b2e79ca09e6/acceptance/suites/tests/00_smoke/puppetdb_integration.rb#L13 - or is the problem with Beaker that is forcing us to do this related just to `confine_block` and not `confine`?

I guess if the problem is with `confine_block` and not `confine` and `confine` is a more conventional way to restrict the platforms that the script is run against than the `if` block, we could also just use `confine` and move this into its own Ruby file?  Not sure what's the best way to do this sort of thing...

erikPrime: No, confine seems to work as advertised (without overwriting the array of hosts in the beaker options hash).

kurtwall:
🙈

PR #1011, opened 2016-04-12
(MAINT)Adds the https_request method from pe-p-s-e

erikPrime:
This is a copy pasta from pe-p-s-e, because we want to be able to use the method from inside puppet-server as well as from inside pe-p-s-e.  I've tested that having the method present in both puppet-server and pe-p-s-e did not break file-sync tests.


kurtwall:
🙈

PR #1012, opened 2016-04-12
(maint) Add cljfmt and pl-clojure-style guide

justinstoller:
This has been taken up by trapperkeeper. Thought it might be valuable to start the discussion on puppet-server?

_as in my editor isn't formatting quite correctly, I keep getting dinged for it in PRs and instead of mucking with it, if `cljfmt` is the future, I just want to use it_

This change in and of itself should be innocuous but the application of these styles needs to be somewhat coordinated with a lull in large PRs - it will cause a hellacious rebase for outstanding work.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,3 +7,6 @@
 [submodule "ruby/hiera"]
 	path = ruby/hiera
 	url = https://github.com/puppetlabs/hiera.git
+[submodule "ext/pl-clojure-style"]
+	path = ext/pl-clojure-style
+	url = git://github.com/puppetlabs/pl-clojure-style.git
--- a/project.clj
+++ b/project.clj
@@ -128,6 +128,10 @@
              :uberjar {:aot [puppetlabs.trapperkeeper.main]
                        :dependencies [[puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty-version]]}
              :ci {:plugins [[lein-pprint "1.1.1"]]}
+             :cljfmt {:plugins [[lein-cljfmt "0.5.0"]
+                                [lein-parent "0.2.1"]]
+                      :parent-project {:path "ext/pl-clojure-style/project.clj"
+                                       :inherit [:cljfmt]}}
              :voom {:plugins [[lein-voom "0.1.0-20150115_230705-gd96d771" :exclusions [org.clojure/clojure]]]}}
 
   :test-selectors {:integration :integration
@@ -135,7 +139,8 @@
 
   :aliases {"gem" ["trampoline" "run" "-m" "puppetlabs.puppetserver.cli.gem"]
             "ruby" ["trampoline" "run" "-m" "puppetlabs.puppetserver.cli.ruby"]
-            "irb" ["trampoline" "run" "-m" "puppetlabs.puppetserver.cli.irb"]}
+            "irb" ["trampoline" "run" "-m" "puppetlabs.puppetserver.cli.irb"]
+            "cljfmt" ["with-profile" "+cljfmt" "cljfmt"]}
 
   ; tests use a lot of PermGen (jruby instances)
   :jvm-opts ["-XX:MaxPermSize=256m" "-Xmx2g"]


cprice404:
Might be good to get some mileage on the changes in TK and see how they are working for people before we do this here?

In any case it definitely needs to be coordinated carefully relative to outstanding PRs, as you mentioned.

justinstoller:
That sounds good.

Is there a way I can help us kick the tires on the changes in trapperkeeper? I could add it as a travis check? Or maybe just send a note to the team to try and keep the task in mind when submitting PRs for trapperkeeper?

cprice404:
Mostly what I'd want to know is whether or not IntelliJ (and other editors of choice) are playing nicely with the cljfmt settings.  Maybe you already have some insight into that; @jpinsonault is currently working on a TK PR, maybe after he finishes it you could help him run the cljfmt stuff and see what (if any) things his editor got wrong?

justinstoller:
Oh yeah. I'd love to sit down with you sometime @jpinsonault, just ping me
when you have a minute.

jpinsonault:
Yeah that sounds like a good idea, I'll let you know when I'm around the office

justinstoller:
@jpinsonault and I sat down for a little while yesterday to play with this in TK. We were pretty easily able to get IntelliJ to play nice with these settings for the files in his PR -- after adding a few special cases. After that I went through the repo on my own (with IntelliJ 15 CE + Cursive) and applied the IDE's code formatting to the whole repo. I ran into numerous problems:

1. First, custom macros with args & body forms that start after a newline (eg. [`with-test-logging`](https://github.com/puppetlabs/trapperkeeper/blob/master/test/puppetlabs/trapperkeeper/testutils/logging_test.clj#L118)) are treated as lists/vectors and only indented one space rather than two. Adding a custom macro formatting is a simple change, but one that must be made per macro (ie a lot)
2. There's a number of special cases in cljfmt that... bring it in line with emacs' default formatting? but all also require special casing (eg [`use-fixtures`](https://github.com/weavejester/cljfmt/blob/master/cljfmt/resources/cljfmt/indents/clojure.clj#L54))
3. cljfmt's ability to specify not just the indentation of successive child forms (like parameters to a function) but nested forms (like definition of an implementation of a protocol) is not something I was able to find a direct equivalent in IntelliJ. (eg. [definitions of service](https://github.com/puppetlabs/trapperkeeper/blob/master/test/puppetlabs/trapperkeeper/internal_test.clj#L13-L17) specified in cljfmt [like so](https://github.com/puppetlabs/pl-clojure-style/blob/master/project.clj#L6))

Now, someone more familiar with either IntelliJ or cljfmt may be able to correct any misunderstandings I have here (cc @pcarlisle since he seems to be the local expert in cljfmt). But from the couple of hours I've spent with it, I think getting IntelliJ's code formatter to actually format code the same as cljfmt would require both a large chuck of time manually updating all the macro behaviors as well as changing our style guide (including overwriting some of cljfmt's default formatting) to behave in ways that IntelliJ can mimic (which would then require updates to the indents.el and indents.vim as well).

One thought I had: it's possible to run external tasks via IntelliJ's "Run" feature. Would it be practical to add a Run command in IntelliJ and have IntelliJ users run that instead of IntelliJ's native code format command before commit? Hopefully someone with more experience with IntelliJ could get it to run from a persistent clj repl session similar to how [cljfmt-vim](https://github.com/venantius/vim-cljfmt) works. Luckily, IntelliJ also tries to keep the existing code formatting of code being edited so hopefully, as long as one doesn't run "Ctrl Alt L", code edited in IntelliJ should still be close to the original in case one isn't strict about running cljfmt on their code before making a commit.

cprice404:
> First, custom macros with args & body forms that start after a newline (eg. with-test-logging) are treated as lists/vectors and only indented one space rather than two. Adding a custom macro formatting is a simple change, but one that must be made per macro (ie a lot)

If we decide this is worth pushing forward on, I'm pretty sure there's a way we could export those settings from Intellij and keep a copy of that artifact in @pcarlisle's repo, so that for a new user it'd just be a matter of importing them.  Over time we would presumably collaboratively end up with an artifact that covered the vast majority of cases we care about.

> : it's possible to run external tasks via IntelliJ's "Run" feature. Would it be practical to add a Run command in IntelliJ and have IntelliJ users run that instead of IntelliJ's native code format command before commit?

I'm sure this is possible but tbh I think it's pretty crappy if we're saying we want to standardize on a formatting style that is incompatible with the editor that like half of our devs use.  I think it would make way more sense to either change our guidelines to something that works for everyone, or to file a ticket with Cursive inquiring about supporting whatever thing we're missing.

Or just calling time on this experiment and possibly revisiting it at some point in the future.

pcarlisle:
I was definitely under the impression that we could get it working with intellij. I'm out of the office until monday but I will take a look at it then.

justinstoller:
Cool, I'll wait for you to return and chat with you on Monday, Patrick.

justinstoller:
To close the loop on this:

I chatted with Patrick later in the week when he got back. He believes he's seen a way to configure IntelliJ to meet follow these standards, I completely believe him, but we would have to go spelunking to figure it out and neither of us actually use IntelliJ.

On the Cursive website it says they working on a better way to configure code formatting.

I've also been running lein-cljfmt on my PRs before getting them merged and it seems to update formatting to not-exactly-the-same-as-IntelliJ-but-close-enough-that-no-one-cares.

Since there may be a new way to specify code formatting for Cursive in the near future, I'm not in IntelliJ on a daily basis already to figure out the finer points of its code formatting or contribute to building a jar of settings, and the current distinction between IntelliJ's formatting and cljfmt's seems small enough that most don't recognize it, I'm going to close this PR until one of those three (hopefully one of the first two) change.

PR #1013, opened 2016-04-12
(SERVER-1245) Pass OID shortnames to tk-auth

fpringvaldsen:
Ensure the puppet OID shortname mappings and the custom
oid mappings provided in the custom oid mappings file are passed
to tk-auth.

This will require a release of tk-auth before it can be merged.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -102,6 +102,9 @@
   {:outcome (schema/enum :success :not-found :error)
    :message schema/Str})
 
+(def OIDMappings
+  {schema/Str schema/Keyword})
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Definitions
 
@@ -1218,3 +1221,20 @@
     (when (fs/exists? cert)
       (fs/delete cert)
       (log/debug "Deleted certificate for" subject))))
+
+(schema/defn ^:always-validate get-custom-oid-mappings :- (schema/maybe OIDMappings)
+  "Given a path to a custom OID mappings file, return a map of all oids to
+  shortnames"
+  [custom-oid-mapping-file :- (schema/maybe schema/Str)]
+  (when (fs/file? custom-oid-mapping-file)
+    (try
+      (let [oid-mappings (:oid_mapping (yaml/parse-string (slurp custom-oid-mapping-file)))]
+        (into {} (for [[oid names] oid-mappings] [(name oid) (keyword (:shortname names))])))
+      (catch Exception e
+        (throw (Exception.
+                 (str "There was a problem parsing " custom-oid-mapping-file) e))))))
+
+(schema/defn ^:always-validate get-oid-mappings :- OIDMappings
+  [custom-oid-mapping-file :- (schema/maybe schema/Str)]
+  (merge (get-custom-oid-mappings custom-oid-mapping-file)
+         (clojure.set/map-invert puppet-short-names)))
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -596,19 +596,18 @@
   "Parse the CSR attributes yaml file at the given path and create a list of
   certificate extensions from the `extensions_requests` section."
   [csr-attributes-file :- schema/Str]
-  (when (fs/file? csr-attributes-file)
-    (try
+  (when csr-attributes-file
+    (if (fs/file? csr-attributes-file)
       (let [csr-attrs (yaml/parse-string (slurp csr-attributes-file))
-            ext-req   (:extension_requests csr-attrs)]
+            ext-req (:extension_requests csr-attrs)]
         (map (fn [[oid value]]
                {:oid      (or (get puppet-short-names oid)
                               (name oid))
                 :critical false
                 :value    (str value)})
              ext-req))
-      (catch Exception e
-        (throw (Exception.
-                 (str "There was a problem parsing " csr-attributes-file) e))))))
+      (log/error (format "%s is not a valid file path. CSR Attributes will not be loaded."
+                         csr-attributes-file)))))
 
 (schema/defn create-master-extensions :- (schema/pred utils/extension-list?)
   "Create a list of extensions to be added to the master certificate."
@@ -1226,13 +1225,12 @@
   "Given a path to a custom OID mappings file, return a map of all oids to
   shortnames"
   [custom-oid-mapping-file :- (schema/maybe schema/Str)]
-  (when (fs/file? custom-oid-mapping-file)
-    (try
+  (when custom-oid-mapping-file
+    (if (fs/file? custom-oid-mapping-file)
       (let [oid-mappings (:oid_mapping (yaml/parse-string (slurp custom-oid-mapping-file)))]
         (into {} (for [[oid names] oid-mappings] [(name oid) (keyword (:shortname names))])))
-      (catch Exception e
-        (throw (Exception.
-                 (str "There was a problem parsing " custom-oid-mapping-file) e))))))
+      (log/error (format "%s is not a valid file path. Custom OID Mappings will not be loaded."
+                         custom-oid-mapping-file)))))
 
 (schema/defn ^:always-validate get-oid-mappings :- OIDMappings
   [custom-oid-mapping-file :- (schema/maybe schema/Str)]
--- a/src/clj/puppetlabs/services/ca/certificate_authority_disabled_service.clj
+++ b/src/clj/puppetlabs/services/ca/certificate_authority_disabled_service.clj
@@ -8,7 +8,7 @@
    This is required to maintain service loading order on installations
    which are not running a CA service."
   CaService
-  []
+  [[:AuthorizationService wrap-with-authorization-check]]
   (initialize-master-ssl!
     [this master-settings certname]
     (log/info "CA disabled; ignoring SSL initialization for Master"))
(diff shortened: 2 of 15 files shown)


src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1218,3 +1221,20 @@
     (when (fs/exists? cert)
       (fs/delete cert)
       (log/debug "Deleted certificate for" subject))))
+
+(schema/defn ^:always-validate get-custom-oid-mappings :- (schema/maybe OIDMappings)
+  "Given a path to a custom OID mappings file, return a map of all oids to
+  shortnames"
+  [custom-oid-mapping-file :- (schema/maybe schema/Str)]
+  (when (fs/file? custom-oid-mapping-file)
```
KevinCorcoran: This seems like, if the config has an invalid file path, it'll just be silently ignored.  I think an error should probably be logged in that case.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1218,3 +1221,20 @@
...
+    (try
+      (let [oid-mappings (:oid_mapping (yaml/parse-string (slurp custom-oid-mapping-file)))]
+        (into {} (for [[oid names] oid-mappings] [(name oid) (keyword (:shortname names))])))
+      (catch Exception e
```
KevinCorcoran: `catch Exception` here looks a bit suspicious.  Does `parse-string` throw something more specific that could be caught instead?  And, actually, could we just let that exception bubble-out instead of this `try`/`catch`/`throw`?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -102,6 +102,9 @@
   {:outcome (schema/enum :success :not-found :error)
    :message schema/Str})
 
+(def OIDMappings
+  {schema/Str schema/Keyword})
+
```
KevinCorcoran: It's pretty funky that this map is backwards from both the usual map style of keyword -> string as well as what `puppet-short-names` is already using.

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
```diff
@@ -27,9 +30,9 @@
              comidi/routes->handler)
          settings
          path
-         wrap-with-authorization-check
-         puppet-version)))
-   context)
+         (fn [request] (wrap-with-authorization-check request {:oid-map oid-mappings}))
```
KevinCorcoran: It looks like this is duplicated in 4 different places (here, admin service, legacy service, master service).  Any way we can DRY that up?  Spitballing here, but ... perhaps, instead of adding the `get-oid-mappings` service function, instead we could add a `auth-middleware-with-oid-mappings` service function which would encapsulate this.  Thoughts?  @nathanielksmith?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
fpringvaldsen: I agree. I actually based this on the parsing of the `csr_attributes` file, which we're already doing in clojure code, so I think we should do this there as well. I'll push up a commit that adds logging in both cases.

src/clj/puppetlabs/services/puppet_admin/puppet_admin_service.clj
```diff
@@ -42,5 +44,5 @@
         (core/build-ring-handler route
                                  whitelist-settings
                                  jruby-service
-                                 wrap-with-authorization-check)))
+                                 (fn [request] (wrap-with-authorization-check request {:oid-map oid-map})))))
```
KevinCorcoran: It's worth noting here, I think, that we have a "master service" in PE as well, so any changes made in this namespace will need to applied there too.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
fpringvaldsen: I agree that it's a bit weird. Saying that, it makes sense conceptually to me both ways (i.e., having the OID as the key in this case, or the shortname as the key in the case of the puppet short names map).

This is the format currently accepted by TK-Auth, so we'd have to change it there if we wanted this to be different. I'd be open to changing it so the OID is represented as a keyword, and the shortname is represented as either a string or a keyword. However, I think the mapping of oid -> shortname is the right way to do it, as that most closely matches the format of the `custom_trusted_oid_mapping` configuration file.

Maybe we should change the `puppet-short-names` map to match this format?

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
fpringvaldsen: Hmm..it seems weird to me to be returning authorization middleware, which is the responsibility of the Authorization Service, from the CA service. It seems like that service should only return custom OID mappings. I'd be willing to be overridden on that though, as I think it would be great to DRY this up.

@nathanielksmith what are your thoughts?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
KevinCorcoran: No strong opinion here ... I'd be tempted to leave it as-is, but open to changing that as well.

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
KevinCorcoran: I agree that it would be bit weird but I would probably prefer that weirdness over the current WETness.

I suppose another option would be to introduce a third service which rolled-up the OIDs from the CA and the middleware from the auth service.  Just spitballin' here.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
fpringvaldsen: Cool, I'll leave it as-is for now.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
fpringvaldsen: I have no idea. I took a look at `clj-yaml` but there's no documentation and no indication of whether or not the `parse-string` function will throw any exceptions.

I dug into SnakeYAML, which is what `clj-yaml` uses for all its parsing, and it doesn't look like loading a YAML file should throw any exceptions, although in that case I'm unsure how it handles invalid syntax.

It's possible the `slurp` call could end up throwing an exception?

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
fpringvaldsen: Hmm, that's also a valid option. I think I'd probably go with the first option though. It's a bit weirder, but adding in another service adds in another layer of complexity, and it will make it a bit harder to upgrade as existing users will need to add the new service to their bootstrap configuration files if they've made any changes.

I guess that's probably not a huge deal though.

KevinCorcoran: I hear ya.  Adding a new service just for this feels like an overly big hammer.

I think I'm still leaning towards my initial suggestion of having the CA supply the OID-mapped auth middleware to the other services.  But maybe let's see what @nathanielksmith thinks before proceeding.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
KevinCorcoran: I think it's probably best to just remove this error handling altogether if there's not a specific exception we are trying to catch and report in a better way.  I suspect that if `slurp` or `parse-string` throws an error, that should be good enough for a user to figure out what's up.

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
nathanielksmith: I might be wrong because I'm still pretty new to this codebase but I don't really having a problem with the service function Kevin originally suggested. tk-auth and the ca service seem related enough to not be overly weird in combining the oid mapping and the auth middleware function.

I agree that a whole new service isn't worth the headache and that the moistness is also not worth it, so the `auth-middleware-with-oid-mappings` type idea seems like a reasonable compromise.

fpringvaldsen: Cool, if you too are good with that approach then I'm :+1:. I'll get to work on that.

fpringvaldsen:
I've updated the error handling for YAML parsing. I should have a patch up for DRYing up the code sometime tomorrow.

fpringvaldsen:
@KevinCorcoran @nathanielksmith I've gone ahead and refactored the CA service to DRY up the code.

src/clj/puppetlabs/services/ca/certificate_authority_disabled_service.clj
```diff
@@ -21,6 +21,6 @@
     [this localcacrl]
     (log/info "CA disabled; ignoring retrieval of CA CRL"))
 
-  (get-oid-mappings
+  (get-auth-handler
     [this]
-    {}))
+    (fn [request] (wrap-with-authorization-check request {:oid-map {}}))))
```
KevinCorcoran: I kinda wonder if this should be more of a no-op middleware instead of adding a dependency on the `AuthorizationService`.

fpringvaldsen: I think the issue with that is that the master, legacy routes, and admin api services all use the authorization middleware regardless of whether or not the disabled CA service is used.

KevinCorcoran:
👍  except for comment about legacy + auth service dependency.

src/clj/puppetlabs/services/ca/certificate_authority_disabled_service.clj
KevinCorcoran: Oh, I think I understand.  In that case, I think perhaps this line should change to simply return `wrap-with-authorization-check`.

fpringvaldsen:
@kevin I've updated the disabled CA service

fpringvaldsen:
@nathanielksmith @KevinCorcoran I've released TK-Auth and updated the dep so this should be all ready to merge now

KevinCorcoran:
:shipit:

PR #1015, opened 2016-04-13
(SERVER-1252) Document new tk-auth allow/deny options.

gguillotte:


gguillotte:
@nathanielksmith 👋

documentation/config_file_auth.md
```diff
@@ -133,12 +133,20 @@ After each rule's `match-request` section, it must also have an `allow`, `allow-
...
+* **`allow`**: This parameter can take a single string value, an array of string values, a single map value with either an `extensions` or `certname` key, or an array of string and map values.
+
+    The string values can be:
+
     * An exact domain name, such as `www.example.com`.
     * A glob of names containing a `*` in the first segment, such as `*.example.com` or simply `*`.
     * A regular expression surrounded by `/` characters, such as `/example/`.
     * A backreference to a regular expression's capture group in the `path` value, if the rule also contains a `type` value of `regex`. For example, if the path for the rule were `"^/example/([^/]+)$"`, you can make a backreference to the first capture group using a value like `$1.domain.org`.
 
+    The map values can be:
```
fpringvaldsen: I think this should be `The map values can contain`. Also, can a map value contain both an `extensions` and a `certname` key?

documentation/config_file_auth.md
```diff
@@ -133,12 +133,20 @@ After each rule's `match-request` section, it must also have an `allow`, `allow-
...
+
+    * An `extensions` key that specifies a set of matching X.509 extensions. Puppet Server authenticates the request only if each key in the map appears in the request, and each key's value exactly matches.
```
fpringvaldsen: I think we should use the term `array` here instead of `set`. Also, do we have documentation on what an extension specification looks like?

documentation/config_file_auth.md
```diff
@@ -133,12 +133,20 @@ After each rule's `match-request` section, it must also have an `allow`, `allow-
...
+    * A `certname` key equivalent to a bare string.
+
```
erikPrime: Can we please have an example here?

documentation/config_file_auth.md
gguillotte: I'm not sure. The examples implied so, and I can test if necessary.

> a single string value, an array of string values, a single map value with either an `extensions` or `certname` key, or an array of string and map values

nathanielksmith:
@gguillotte sorry, totally missed this. will review today!

documentation/config_file_auth.md
nathanielksmith: +1 on contain over be. and no, a given map can't have both.

documentation/config_file_auth.md
nathanielksmith: Since we're talking in the abstract, "list" might actually be a better word than set or array.

documentation/config_file_auth.md
```diff
@@ -135,16 +155,16 @@ If a request matches the rule, Puppet Server checks the request's authenticated
...
     * An exact domain name, such as `www.example.com`.
     * A glob of names containing a `*` in the first segment, such as `*.example.com` or simply `*`.
     * A regular expression surrounded by `/` characters, such as `/example/`.
     * A backreference to a regular expression's capture group in the `path` value, if the rule also contains a `type` value of `regex`. For example, if the path for the rule were `"^/example/([^/]+)$"`, you can make a backreference to the first capture group using a value like `$1.domain.org`.
 
-    The map values can be:
+    The map values can contain:
 
-    * An `extensions` key that specifies a set of matching X.509 extensions. Puppet Server authenticates the request only if each key in the map appears in the request, and each key's value exactly matches.
+    * An `extensions` key that specifies an array of matching X.509 extensions. Puppet Server authenticates the request only if each key in the map appears in the request, and each key's value exactly matches.
```
nathanielksmith: Though there is a great example config above, it might still be nice to show the "multiple possible values for an extension" case either here or above (ie something like `{extensions: {gnarlytimes: [yeah, word, sweet]}}`)

documentation/config_file_auth.md
nathanielksmith: oops, this was already updated in the next commit, sorry. array is fine.

rlinehan:
@gguillotte looks like this needs to be rebased :( sorry.

gguillotte:
@rlinehan Rebased, thanks!

KevinCorcoran:
I think this needs to be targeted at `master` since that's where these changes have actually landed.  They're not yet in `stable`.

gguillotte:
@KevinCorcoran Opened PR #1040 targeting these changes at master.

KevinCorcoran:
I'm going to close this since I think these changes should just land in `stable` via #1040 and then eventually merging-down `master` -> `stable`, which we do regularly as releases dictate.

PR #1016, opened 2016-04-13
(SERVER-1187) Adjust log levels for autosign actions

haus:
This PR makes two changes to improve the logging for autosigning.
First, if the autosign command generates error output or a non-zero
exit-code, a warning is generated in the logs. Second, it moves the
signing log message to info from debug. Previously this information
would only be available when logging at the debug level, which would
generate a substantial amount of noise. With these changes many
autosign issues can be diagnosed without needing to change loglevel.


haus:
The before and after logging output can be seen here: https://gist.github.com/haus/d127ae545574c7353002629fbc071eee

fpringvaldsen:
:+1:

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -804,9 +806,16 @@
                   :in csr-stream
                   :env (merge env {"RUBYLIB" rubylib})})]
     (log/debugf "Autosign command '%s %s' exit status: %d"
-                executable subject (:exit results))
+                executable subject (:exit-code results))
     (log/debugf "Autosign command '%s %s' output: %s"
-                executable subject (str (:err results) (:out results)))
+                executable subject (str (:stderr results) (:stdout results)))
```
justinstoller: I'm not clear on the value of having the stderr logged twice?

haus: I could consolidate the two and log at warn if stderr is generated and debug if no stderr is generated, but the value is that if there is stderr, the user won't need to adjust their loglevel to see the log message.

justinstoller: Yeah, I think logging at :warn is a great idea for stderr. But if folks have to turn the logging up to debug this issue then they'll be getting stdout on debug and stderr on both debug and warn. Which I don't know if that's a good thing (really highlights stderr) or a bad thing (duplicated logs).

haus:
@justinstoller updated.

justinstoller:
👍

justinstoller:
Well, at least one of the failures is:
```
lein test puppetlabs.services.master.environment-classes-int-test

2016-04-14 23:33:14,871 ERROR [async-dispatch-41] [p.t.internal] Error during service init!!!
clojure.lang.ExceptionInfo: Input to get-custom-oid-mappings does not match schema: [(named (not (instance? java.lang.String nil)) custom-oid-mapping-file)]
	at puppetlabs.puppetserver.certificate_authority$eval22339$get_custom_oid_mappings__22344.invoke(certificate_authority.clj:1231) ~[na:na]

....snip....

Ran 193 tests containing 2255 assertions.
0 failures, 1 errors.
Tests failed.
```

Which seems to be unrelated to this patch. But I'd rather wait to merge until we get master back green.

fpringvaldsen:
@justinstoller I'm guessing that's the same failure that @camlow325's PR fixes, so we should be good if we merge that first.

fpringvaldsen:
I just merged the other PR and kicked Travis, hopefully it should pass now

justinstoller:
The travis failure only happened on one cell and is a known flaky test. Further improvements to that test are ticketed in SERVER-1269. Not going to block merge on it (let me know if that's the wrong call).

PR #1017, opened 2016-04-13
(SERVER-1258) Update puppet-agent version to greenify tests

haus:
This commit updates the puppet-agent ref to a version that passes the
puppetdb smoke tests and also updates the puppet submodule commit to
match the puppet change.


haus:
acceptance/suites/tests/00_smoke/puppetdb_integration.rb passes for me locally on the agent version.

camlow325:
:+1:

PR #1018, opened 2016-04-14
(MAINT) Remove csr attrs and oid error logging

fpringvaldsen:
Previously, an error was logged if the configured CSR Attributes
file path or the configured Custom Trusted OID Mappings file
path failed an fs/file? check. However, Puppet has a default
value for both of these settings, meaning the error would be
logged on startup unless the user specifically adds those
configuration files.

This removes the logging of that error.

This is what we were originally doing with the `csr-attributes-file`, and I realized why when I added these changes to pe-p-s-e and and it started spewing errors about not loading these files.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -596,18 +596,17 @@
   "Parse the CSR attributes yaml file at the given path and create a list of
   certificate extensions from the `extensions_requests` section."
   [csr-attributes-file :- schema/Str]
-  (when csr-attributes-file
-    (if (fs/file? csr-attributes-file)
-      (let [csr-attrs (yaml/parse-string (slurp csr-attributes-file))
-            ext-req (:extension_requests csr-attrs)]
-        (map (fn [[oid value]]
-               {:oid      (or (get puppet-short-names oid)
-                              (name oid))
-                :critical false
-                :value    (str value)})
-             ext-req))
-      (log/error (format "%s is not a valid file path. CSR Attributes will not be loaded."
-                         csr-attributes-file)))))
+  (if (fs/file? csr-attributes-file)
+    (let [csr-attrs (yaml/parse-string (slurp csr-attributes-file))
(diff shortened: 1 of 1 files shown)


src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1225,12 +1222,9 @@
   "Given a path to a custom OID mappings file, return a map of all oids to
   shortnames"
   [custom-oid-mapping-file :- (schema/maybe schema/Str)]
```
KevinCorcoran: It seems that `schema/maybe` is unnecessary here since Puppet has a default value for the setting.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1224,13 +1221,10 @@
...
-  (when custom-oid-mapping-file
-    (if (fs/file? custom-oid-mapping-file)
-      (let [oid-mappings (:oid_mapping (yaml/parse-string (slurp custom-oid-mapping-file)))]
-        (into {} (for [[oid names] oid-mappings] [(name oid) (keyword (:shortname names))])))
-      (log/error (format "%s is not a valid file path. Custom OID Mappings will not be loaded."
-                         custom-oid-mapping-file)))))
+  [custom-oid-mapping-file :- schema/Str]
+  (when (fs/file? custom-oid-mapping-file)
+    (let [oid-mappings (:oid_mapping (yaml/parse-string (slurp custom-oid-mapping-file)))]
+      (into {} (for [[oid names] oid-mappings] [(name oid) (keyword (:shortname names))])))))
```
KevinCorcoran: I would probably change these two `when` usages to `if` and when `fs/file?` returns falsey log a message at DEBUG (probably?) like `No custom OID mapping configuration file found at $custom-oid-mapping-file, no custom OID mappings loaded.` or something like that (and similar for CSR attributes).

KevinCorcoran:
👍 looks good except comment about adding new log messages

fpringvaldsen:
@KevinCorcoran I've added debug log messages

PR #1019, opened 2016-04-15
(SERVER-1245) Add trusted oid mapping to env classes int test

camlow325:
This commit adds a `trusted-oid-mapping-file` setting value that the
mock pupppetserver config service in the test uses.  This was added in
order for the newly added code which handles the trusted oid mapping
file to work without any problems - although it isn't directly related
to anything that the specific test validates.


camlow325:
@KevinCorcoran @fpringvaldsen This should address the unit test failure introduced in #1013.  /CC @kurtwall.

haus:
:+1:

fpringvaldsen:
:+1:

PR #1021, opened 2016-04-18
(SERVER-1258) Update puppet-agent ref to greenify tests

haus:
This commit updates the puppet-agent ref to fix PUP-6176, which had been
causing puppetdb module failures on ubuntu-14.04.


erikPrime:
:+1:

PR #1014, opened 2016-04-13
(SERVER-1206) Rename occurrences of puppet-server to puppetserver

cprice404:

PR #1022, opened 2016-04-19
(SERVER-1206) Rename occurrences of puppet-server to puppetserver

camlow325:


camlow325:
This replaces #1014.

documentation/config_file_auth.md
```diff
@@ -23,7 +23,7 @@ However, Puppet Server now has its own `auth.conf` file that uses a new HOCON fo
 >
 > Puppet Server follows the following logic when determining whether to use the new or old authorization methods:
 >
-> * Requests to Puppet master service endpoints already manageable through the deprecated authorization methods and [Puppet `auth.conf`][] file --- such as `catalog`, `node`, and `report` --- use Puppet Server's new `auth.conf` rules **only** if the `use-legacy-auth-conf` setting in `puppet-server.conf` is set to `false`. If `use-legacy-auth-conf` is set to true (which is its default), Puppet Server warns you that the legacy authentication method is deprecated.
+> * Requests to Puppet master service endpoints already manageable through the deprecated authorization methods and [Puppet `auth.conf`][] file --- such as `catalog`, `node`, and `report` --- use Puppet Server's new `auth.conf` rules **only** if the `use-legacy-auth-conf` setting in `puppetserver.conf` is set to `false`. If `use-legacy-auth-conf` is set to true (which is its default), Puppet Server warns you that the legacy authentication method is deprecated.
```
rlinehan: Are we changing this file name?

documentation/config_file_auth.md
```diff
@@ -54,7 +54,7 @@ If this setting is `true`, Puppet Server ignores any presented certificate and r
 
 You cannot rename any of the `X-Client` headers when this setting is enabled, and you must specify identity through the `X-Client-Verify`, `X-Client-DN`, and `X-Client-Cert` headers.
 
-For more information, see [External SSL Termination](./external_ssl_termination.html#disable-https-for-puppet-server) in the Puppet Server documentation and [Configuring the Authorization Service](https://github.com/puppetlabs/trapperkeeper-authorization/blob/master/doc/authorization-config.md#allow-header-cert-info) in the `trapperkeeper-authorization` documentation.
+For more information, see [External SSL Termination](./external_ssl_termination.html#disable-https-for-puppetserver) in the Puppet Server documentation and [Configuring the Authorization Service](https://github.com/puppetlabs/trapperkeeper-authorization/blob/master/doc/authorization-config.md#allow-header-cert-info) in the `trapperkeeper-authorization` documentation.
```
rlinehan: I don't think this should be changed.

documentation/restarting.markdown
```diff
@@ -7,15 +7,15 @@ canonical: "/puppetserver/latest/restarting.html"
 [logback.xml]: ./config_file_logbackxml.html
 [Hiera]: /hiera/latest/configuring.html
 [gems]: /puppetserver/latest/gems.html
-[core dependencies]: /puppet/latest/reference/about_agent.html#what-are-puppet-agent-and-puppet-server
+[core dependencies]: /puppet/latest/reference/about_agent.html#what-are-puppet-agent-and-puppetserver
```
rlinehan: I don't think this should be changed.

documentation/restarting.markdown
```diff
@@ -7,15 +7,15 @@ canonical: "/puppetserver/latest/restarting.html"
...
 [environment]: /puppet/latest/reference/environments.html
 [environment caching]: /puppet/latest/reference/configuration.html#environmenttimeout
 
 Starting in version 2.3.0, you can restart Puppet Server by sending a hangup signal, also known as a [HUP signal or SIGHUP](https://en.wikipedia.org/wiki/SIGHUP), to the running Puppet Server process. The HUP signal stops Puppet Server and reloads it gracefully, without terminating the JVM process. This is generally *much* faster than completely stopping and restarting the process. This allows you to quickly load changes to your Puppet Server master, including configuration changes.
 
 There are several ways to send a HUP signal to the Puppet Server process, but the most straightforward is to run the following [`kill`](http://linux.die.net/man/1/kill) command:
 
-    kill -HUP `pgrep -f puppet-server`
+    kill -HUP `pgrep -f puppetserver`
```
rlinehan: Is this changing?

documentation/ssl_server_certificate_change_and_virtual_ips.markdown
```diff
@@ -3,7 +3,7 @@ layout: default
 title: "Puppet Server: Known Issues: SSL Server Certificate Change and Virtual IP Addresses"
 canonical: "/puppetserver/latest/ssl_server_certificate_change_and_virtual_ips.html"
 ---
-[pe_db_instructions]: https://docs.puppetlabs.com/pe/latest/release_notes_known_issues.html#puppetdb-behind-a-load-balancer-causes-puppet-server-errors
+[pe_db_instructions]: https://docs.puppetlabs.com/pe/latest/release_notes_known_issues.html#puppetdb-behind-a-load-balancer-causes-puppetserver-errors
```
rlinehan: This shouldn't be changed.

project.clj
```diff
@@ -76,7 +76,7 @@
 
   :plugins [[lein-release "1.0.5" :exclusions [org.clojure/clojure]]]
 
-  :uberjar-name "puppet-server-release.jar"
+  :uberjar-name "puppetserver-release.jar"
```
rlinehan: Will it cause any problems to change this?

resources/ext/config/conf.d/puppetserver.conf
```diff
@@ -17,7 +17,7 @@ jruby-puppet: {
     # of the corresponding settings, but it is important that any changes made to
     # master-conf-dir and master-code-dir are also made to the corresponding Puppet
     # settings when running the Puppet CLI tools. See
-    # https://docs.puppetlabs.com/puppetserver/latest/puppet_conf_setting_diffs.html#overriding-puppet-settings-in-puppet-server
+    # https://docs.puppetlabs.com/puppetserver/latest/puppet_conf_setting_diffs.html#overriding-puppet-settings-in-puppetserver
```
rlinehan: I don't think this should be changed.

src/clj/puppetlabs/services/config/puppet_server_config_service.clj
```diff
@@ -27,11 +27,11 @@
           (ks/pprint-to-string puppet-config))
         (core/init-webserver! override-webserver-settings!
                               (get-in tk-config
-                                      [:webserver :puppet-server]
+                                      [:webserver :puppetserver]
```
rlinehan: This seems like it will require PE module changes and maybe manual intervention by OSS users if they having changed their webserver.conf (maybe? not sure what exactly is in the OSS webserver.conf).

dev/puppetserver.conf.sample
```diff
@@ -5,7 +5,7 @@ global: {
 product: {
     update-server-url: "http://localhost/"
     name: {group-id: puppetlabs.dev
-           artifact-id: puppet-server}
+           artifact-id: puppetserver}
```
rlinehan: This seems like maybe not a thing we should change without talking to BizDev? I know this file is not the actual thing that gets shipped, but I think we are testing this from it.

rlinehan:
Done reviewing. There are several places where we are changing path fragments in markdown links that I don't think should be changed. I also left comments on anything that seems like it would be user facing and more trouble than it's worth to change.

dev/puppetserver.conf.sample
camlow325: The artifact-id that we've been using to report to dujour has been "puppetserver" rather than "puppet-server" for some time now.  Can't remember when it changed but it was several releases back.  I think the value we use in the dev sample file isn't too relevant anyway.  We use that in conjunction with `puppetlabs.dev` rather than `puppetlabs` (production) to hopefully filter out the `dev` related check-ins somewhat, AFAIR.

src/clj/puppetlabs/services/config/puppet_server_config_service.clj
camlow325: I think it's unlikely this would affect OSS users.  The default 'webserver.conf' file that we ship for OSS has a single nameless server - https://github.com/puppetlabs/puppetserver/blob/puppet-server-2.3.1/ezbake/config/conf.d/webserver.conf.  I suppose some users could have observed that we have special logic in Puppet Server around the "puppet-server" name and taken advantage of that for doing multi-server configurations, but that doesn't seem too likely.

The "puppet-server" name is used in PE, as you mentioned, though - https://github.com/puppetlabs/pe-puppetserver/blob/31613598bec1aad3b7c7c5b7e6b4355ccd7f18c5/config/conf.d/webserver.conf#L2 - and probably would require PE module changes to go along with this.  That wouldn't necessarily make this untenable but where it is possible that users may have made customizations to the webserver.conf in PE outside of what the PE module manages - since I don't think we overwrite that file completely from the PE module anymore - it does seem like we could break some PE users.

For that reason, I'm inclined to agree with you and suggest that we preserve "puppet-server" here instead.  I'll make that change in a new commit.  /CC @cprice404 for awareness on this since I'd just pulled over his change for this originally and he may have a different opinion about it.

documentation/config_file_auth.md
camlow325: The name of this file in OSS - for some number of releases - been "puppetserver.conf".  See https://github.com/puppetlabs/puppetserver/blob/puppet-server-2.3.1/resources/ext/config/conf.d/puppetserver.conf.  So no change here I don't think.

documentation/config_file_auth.md
camlow325: Yep, that looks wrong.  Will fix, thanks.

documentation/restarting.markdown
camlow325: Yep, thanks.  Will fix.

documentation/ssl_server_certificate_change_and_virtual_ips.markdown
camlow325: Looks like this is a dead link anyway.  I don't see this page in the Known Issues for any PE releases past PE 3.8, for some reason.  Maybe this should be linked instead to https://docs.puppet.com/pe/3.8/release_notes_known_issues.html#puppetdb-behind-a-load-balancer-causes-puppet-server-errors for now?  @gguillotte - what do you think?

resources/ext/config/conf.d/puppetserver.conf
camlow325: yep, agreed.

documentation/ssl_server_certificate_change_and_virtual_ips.markdown
gguillotte: @camlow325 The correct link should now be https://docs.puppet.com/pe/latest/release_notes_known_issues_puppetdb.html#puppetdb-behind-a-load-balancer-causes-puppet-server-errors.

project.clj
camlow325: I don't *think* it will.  From a packaging perspective, the code on this PR seems to build a package that runs fine, even with the new uberjar name.  I did a local ezbake build off this branch to here - http://builds.puppetlabs.lan/puppetserver/2.4.0.master.SNAPSHOT.2016.04.20T0815/.  I installed that package on CentOS 7 and was able to start puppetserver and do an agent run against it.

I'm not sure that I'd consider the uberjar file name as "API", per se.  I think an upgrade would remove the old uberjar name and replace with this new one.

@haus - what do you think about this?  Would this change cause you any concern?

documentation/restarting.markdown
camlow325: Yeah, assuming we stick with the new uberjar name of `puppetserver-release.jar` instead of `puppet-server-release.jar`, it would be changing.  With the ezbake package I built from the latest work on the branch, http://builds.puppetlabs.lan/puppetserver/2.4.0.master.SNAPSHOT.2016.04.20T0815/, I did have to do a `pgrep -f puppetserver` instead of `pgrep -f puppet-server` in order to find the id of the running java process for Puppet Server.

project.clj
rlinehan: I think I was mostly thinking about upgrades - if we just change this, will we leave the old jar behind? I don't think I understand packaging enough on the various platforms to know that for sure.

src/clj/puppetlabs/services/config/puppet_server_config_service.clj
camlow325: I put the dash back in for this in 2a4d62e.  Assuming we think this PR can be merged before @cprice404 but decide later on that we still really do want to change it, maybe we could do that in a follow-up PR - and submit separate tickets to track the corresponding pe-puppetserver config and PE module config changes that would presumably need to go along with it.

project.clj
camlow325: Yeah, fair question.  I tried this out on CentOS 7:

1) Installed puppetserver 2.3.1.
2) Started the puppetserver service and did a successful local agent run against the master.
3) Upgraded to the custom puppetserver build - 2.4.0.master.SNAPSHOT.2016.04.20T0815 (via 'yum upgrade puppetserver').
4) Observed that the old puppetserver service automatically stopped and restarted with the new uberjar automatically, near the end of running the 'yum upgrade' command.
5) Did another successful local agent run against the master.
6) Observed that the `/opt/puppetlabs/server/apps/puppetserver/puppet-server-release.jar` file had been removed and that `/opt/puppetlabs/server/apps/puppetserver/puppetserver-release.jar` was now present.

I didn't test packages on other OSes that would use other types of init systems, but I'd hope the behavior would be the same anyway.  @haus, is that what you'd expect as well?

I honestly don't care all that much about the name of the uberjar file in the package.  I just had carried this change forward from @cprice404's original PR.  If we think there's too much risk associated with it, though, I'm open to reverting this part.

camlow325:
I pushed up 2b9fc8f with corrections to various doc links.

I think I've addressed all of your feedback so far, @rlinehan - other than maybe the question about the uberjar file name that were deliberating over.  I'm happy to defer to @haus' judgement on that question as to whether or not there might be some bad consequences to doing that.

project.clj
haus: This should be fine. The ezbake init scripts use :uberjar-name directly. I *think* we may have instructions for using HUP that would need updating though.

camlow325: @haus This PR updates the one reference to "puppet-server" to "puppetserver" in the "restarting.markdown" doc - https://github.com/puppetlabs/puppetserver/pull/1022/files/9d20775a6e446b457bc4827ffd79e6cdb450d6ad#r60331291.  Anything else you can think of besides that which would need updating?

haus: Ah, nope. That should cover it.

dev/puppetserver.conf.sample
rlinehan: Ah cool, good to know.

src/clj/puppetlabs/services/config/puppet_server_config_service.clj
rlinehan: 👍

rlinehan:
👍 looks good to me.

rlinehan:
🎆

PR #1023, opened 2016-04-20
(doc) Add an authorization rule migration document

gguillotte:
Add a document to help Puppet Server users convert from legacy Puppet authorization rules to Server HOCON authorization rules.

I'm making this PR to review the content. If it works, I'll follow up with a commit adding links to it from other documentation.

CC @camlow325


documentation/config_file_auth_migration.md
```diff
@@ -0,0 +1,267 @@
...
+
+## Converting rules directly
+
+Most of the deprecated authorization rules and settings are available in the new format.
+
+> ### Unavailable rules, settings, or values
+>
+> The following rules, settings, and values have no direct equivalent in the new HOCON format. If you require them, you must reimplement them differently in the new format.
+>
+> * **on value of `auth`:** The deprecated `auth` parameter's on value results in a match only when a request provides a client certificate. There is no equivalent behavior in the HOCON format.
```
camlow325: This might be too pedantic to describe here, not sure...

With the Ruby auth.conf format, the "auth" keyword was evaluated as part of the determination of whether the incoming request matches the corresponding rule or not.  In the Puppet Server tk-auth format, though, there is no methodology for expressing that the request matches the rule  based on whether the user was authenticated or not.  Instead, in the Puppet Server tk-auth format, once it is determined that a request matches a rule, the request will then be allowed / denied based on whether the user had authenticated as a matching user name or the user was not authenticated ("allow-unauthenticated: true").

What you have here may cover this sufficiently for what users need to know about this.  Just throwing this out in the event that it might lead you to want to reword this.

documentation/config_file_auth_migration.md
```diff
@@ -0,0 +1,267 @@
...
+4. Next, set the [`allow`](#allow-allow-unauthenticated-and-deny) setting. The legacy rule used a `*` glob, which is also supported in HOCON.
+
+    ``` hocon
+    ...
+            match-request: {
+                path: /puppet/v3/environments
+                type: path
+                method: [get, post]
+            }
+            allow: *
```
camlow325: The asterisk needs to be wrapped inside of double quotes, so like:

```
allow: *
```

Asterisk is a reserved character in HOCON so you can't just have the value be * directly.

documentation/config_file_auth_migration.md
```diff
@@ -0,0 +1,267 @@
...
+```
+path ~ ^/puppet/v3/catalog/([^/]+)$
+```
+
+HOCON:
+
+``` hocon
+...
+        match-request: {
+            path: "/^\/puppet\/v3\/catalog\/([^\/]+)$/"
```
camlow325: I don't think you need the leading and trailing solidus characters, and escaping the forward slash characters is optional in HOCON/JSON, I believe.  We just do this in the default catalog rule embedded in the Puppet Server package:

```
path: "^/puppet/v3/catalog/([^/]+)$"
```

documentation/config_file_auth_migration.md
```diff
@@ -0,0 +1,267 @@
...
+            type: regex
+        }
+        allow: $1
```
camlow325: I'm pretty sure you have to quote this, like ...

```
allow: "$1"
```
... since I think the `$` character is used as a marker for variable substitution in HOCON, which we wouldn't want in this example.  The above is what we do in the rules containing backreferences that we package by default in Puppet Server.

documentation/config_file_auth_migration.md
```diff
@@ -0,0 +1,267 @@
...
+There's no direct equivalent to the search indirector for the deprecated `method` setting. Create the equivalent rule by passing GET and POST to `method` and specifying endpoint paths using the `path` parameter.
+
+Deprecated:
+
+```
+
+``` hocon
+...
+        match-request: {
+            path: "/(s|_search)$/"
```
camlow325: Same comment here as above about not needing the leading and trailing solidus characters.  Might be good for this one to include an example of a full endpoint that has a plural endpoint, like:

```
path: "^/puppet/v3/file_metadatas?/user_files/"
```

camlow325:
Overall, big :+1: on this.  I really like the approach of building up a rule piece by piece.  Seems like it will be really helpful for users doing their rule migrations.

I left a few inline comments, mostly on quoting / syntax.

A few other things that might be good to cover in this doc that I can think of would be...

1) A Ruby auth.conf rule with an `environment` in it.  Users have to express the environment under a `query-params` map with an entry for the `environment` keyword in the HOCON auth.conf.

2) Some more discussion about what values to use for the  `sort-order` field.  In particular, it might help to talk about the reserved "user" ranges for `sort-order` and that users would want to place the rules that they want to supercede the default ones from our Puppet Server packaging in the range from 1-399 and ones they might want to have subordinate to the ones in our default packaging in the range from 601-998.

3) Maybe work in somewhere that if the user was using "auth: any" to grant all users (authenticated or not) access to an endpoint in Ruby Puppet's auth.conf that they would want to use "allow-unauthenticated" in Puppet Server's HOCON auth.conf.

KevinCorcoran:
@gguillotte Similar to this, see https://tickets.puppetlabs.com/browse/SERVER-1270.

gguillotte:
I just realized we already have [compatibility_with_puppet_agent.markdown](https://github.com/puppetlabs/puppetserver/blob/cd1200a07704050a29af429444e7bd563b6fcd2e/documentation/compatibility_with_puppet_agent.markdown), which already covers some of this with examples. I'll rewrite this on feedback from @camlow325 and @KevinCorcoran, remove redundancies, and add links to both documents.

gguillotte:
Revised the guide and added links from other docs. CC @camlow325 @KevinCorcoran

camlow325:
:+1: @KevinCorcoran You good with merging this?

documentation/config_file_auth_migration.markdown
```diff
@@ -0,0 +1,342 @@
...
+canonical: "/puppetserver/latest/config_file_auth_migration.html"
+---
+
+[Puppet `auth.conf`]: /puppet/latest/reference/config_file_auth.markdown
+[deprecated]: ./deprecated_features.markdown
+[Backward Compatibility with Puppet 3 Agents]: ./compatibility_with_puppet_agent.markdown
+
+Puppet Server 2.2.0 introduced a significant change in how it manages authentication to API endpoints. The older [Puppet `auth.conf`][] file and whitelist-based authorization method are [deprecated][]. Puppet Server's new `auth.conf` file, illustrated below in examples, also uses a different format for authorization rules.
+
+Use the following examples and methods to convert your authorization rules when upgrading from Puppet Server 2.2.0 and newer. For detailed information about using the new or deprecated `auth.conf` rules with Puppet Server, see the [Puppet Server `auth.conf` documentation](./config_file_auth.markdown).
```
KevinCorcoran: s/upgrading from/upgrading to

documentation/config_file_auth_migration.markdown
```diff
@@ -0,0 +1,342 @@
...
+
+## Managing rules with Puppet modules
+
+You can reimplement and manage your authorization rules in the new HOCON format and `auth.conf` file by using the [`puppetlabs-puppet_authorization`](https://forge.puppet.com/puppetlabs/puppet_authorization) Puppet module. See the module's documentation for details.
+
+## Converting rules directly
+
+Most of the deprecated authorization rules and settings are available in the new format.
+
+> ### Unavailable rules, settings, or values
```
KevinCorcoran: This is all formatted as a big quote.  (I think that's what it is.)  It looks a bit strange.  Was that intentional?

documentation/config_file_auth_migration.markdown
```diff
@@ -0,0 +1,342 @@
...
+                type: path
+                method: [get, post]
+            }
+            allow: "*"
+            sort-order: 1
+            name: "environments"
+        },
+    ]
+}
+```
```
KevinCorcoran: This walkthrough is really, really good.  💯

documentation/config_file_auth_migration.markdown
```diff
@@ -0,0 +1,342 @@
...
+
+### Converting more complex rules
+
+#### Paths set by regular expressions
+
+To convert a regular expression path, enclose it in double quotation marks and solidus characters (`/`), and set the `type` to regex.
+
+> **Note:** You must escape regular expressions to conform to HOCON standards, which are the same as JSON's and differ from the deprecated Ruby format's regular expressions. For instance, the digit-matching regular expression `\d` must be escaped with a second backslash, as `\\d`.
```
KevinCorcoran: A minor nitpick - I think it's probably preferable to call the old format "legacy" and not "Ruby".  We don't want to hate on Ruby too much and in reality users shouldn't really need to know implementation language, just old vs. new.

documentation/config_file_auth_migration.markdown
gguillotte: It is (we use blockquote styles to break out asides and notes), but it's evolved into a big-enough thing to not need that formatting. I'll remove the blockquotes.

KevinCorcoran:
👍 aside from the few minor comments

documentation/config_file_auth_migration.markdown
gguillotte: Yeah, I'll go with deprecated (also per an earlier discussion with Nick regarding auth in the Puppet docs).

gguillotte:
Adjusted some wording per @KevinCorcoran and caught a missing period. CC @camlow325

PR #1026, opened 2016-04-20
(MAINT) fixes a logic error in PR#1009

erikPrime:
The previous PR treated puppetdb_supported_platforms as an array of strings.  It is actually an array of regexps.  See SERVER-1292

### DIFF
--- a/acceptance/suites/pre_suite/foss/30_install_dev_repos.rb
+++ b/acceptance/suites/pre_suite/foss/30_install_dev_repos.rb
@@ -25,24 +25,3 @@ if puppet_build_version
     end
   end
 end
-
-if puppetdb_supported_platforms.include?(master.platform) then
-  step "Install PuppetDB repository" do
-    install_puppetlabs_dev_repo(
-      master, 'puppetdb', test_config[:puppetdb_build_version],
-      repo_config_dir, install_opts)
-
-    # Internal packages on ubuntu/debian aren't authenticated and thus apt
-    # will fail to install PuppetDB on those platforms.
-    # This hack tells apt that we can trust the PuppetDB packages until RE-6014
-    # is resolved, at which point this [trusted=yes] will already be in the file
-    # and we can delete the on(master, "sed ...") block below.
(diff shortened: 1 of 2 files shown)


acceptance/suites/pre_suite/foss/30_install_dev_repos.rb
```diff
@@ -26,11 +26,12 @@
   end
 end
 
-if puppetdb_supported_platforms.include?(master.platform) then
-  step "Install PuppetDB repository" do
-    install_puppetlabs_dev_repo(
-      master, 'puppetdb', test_config[:puppetdb_build_version],
-      repo_config_dir, install_opts)
+puppetdb_supported_platforms.each do |pdb_platform|
+  if master.platform.include?pdb_platform.source then
```
camlow325: `master.platform` is a string, right?  Feels a little weird to me to do an `include?` on the original string since I think you'd lose the ability to match a regex.  Maybe this would be better?

~~~ruby
if master.platform ~= pdb_platform
~~~

?

Still wonder if it would be better to use beaker's `confine` and move the whole puppetdb install into a different test step rather than doing more seemingly hand-rolled iteration here?

erikPrime: `if master.platform ~= pdb_platform` syntax errors.  

When master.platform = "el-6-x86_64" and pdb_platform.source = "el" , 
`master.platform =~ pdb_platform` returns zero (0).
`master.platform =~ pdb_platform.source` returns type mismatch

camlow325: If I do the following in irb...

~~~ruby
if "el-6-x86_64" =~ /el/ then
  puts "matches"
end
~~~

... "matches" is written out.  If I do ...

~~~ruby
if "debian-7" =~ /el/ then
  puts "matches"
end
~~~

... "matches" is not written out.

Seems like `master.platform =~ pdb_platform` returning zero would be okay here?

PR #1028, opened 2016-04-21
Take up ring-middleware helpers

justinstoller:
Previously we copied and pasted (with slight changes) much of what made
up our ringutils.

As part of TK-214, the re-usable portions have moved to
puppetlabs/ring-middleware. This patch updates puppetserver to use those
common library functions and removes the outdated vendored functions.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/jruby_request.clj
+++ b/src/clj/puppetlabs/puppetserver/jruby_request.clj
@@ -2,29 +2,11 @@
   (:require [clojure.tools.logging :as log]
             [ring.util.response :as ring-response]
             [slingshot.slingshot :as sling]
+            [puppetlabs.ring-middleware.core :as mw]
             [puppetlabs.services.jruby.jruby-puppet-service :as jruby]
             [schema.core :as schema]
             [puppetlabs.puppetserver.common :as ps-common]))
 
-(defn throw-bad-request!
-  "Throw a ::bad-request type slingshot error with the supplied message"
-  [message]
-  (sling/throw+ {:type ::bad-request
-                 :message message}))
-
-(defn bad-request?
-  [x]
-  "Determine if the supplied slingshot message is for a 'bad request'"
-  (when (map? x)
-    (= (:type x)
-       :puppetlabs.puppetserver.jruby-request/bad-request)))
-
-(defn service-unavailable?
-  [x]
-  "Determine if the supplied slingshot message is for a 'service unavailable'"
-  (when (map? x)
-    (= (:type x)
-       :puppetlabs.services.jruby.jruby-puppet-service/service-unavailable)))
 
 (defn jruby-timeout?
   "Determine if the supplied slingshot message is for a JRuby borrow timeout."
@@ -36,9 +18,7 @@
 (defn output-error
   [{:keys [uri]} {:keys [message]} http-status]
   (log/errorf "Error %d on SERVER at %s: %s" http-status uri message)
-  (-> (ring-response/response message)
-      (ring-response/status http-status)
-      (ring-response/content-type "text/plain")))
+  (mw/plain-response http-status message))
 
 (defn wrap-with-error-handling
   "Middleware that wraps a JRuby request with some error handling to return
@@ -47,11 +27,11 @@
   (fn [request]
     (sling/try+
      (handler request)
-     (catch bad-request? e
+     (catch mw/bad-request? e
        (output-error request e 400))
      (catch jruby-timeout? e
        (output-error request e 503))
-     (catch service-unavailable? e
+     (catch mw/service-unavailable? e
        (output-error request e 503)))))
 
 (defn wrap-with-jruby-instance
@@ -84,11 +64,11 @@
     (let [environment (get-environment-from-request request)]
       (cond
         (nil? environment)
-        (throw-bad-request!
+        (mw/throw-bad-request!
          "An environment parameter must be specified")
 
         (not (nil? (schema/check ps-common/Environment environment)))
-        (throw-bad-request!
+        (mw/throw-bad-request!
          (ps-common/environment-validation-error-msg environment))
 
         :else
--- a/src/clj/puppetlabs/puppetserver/jruby_request.clj
+++ b/src/clj/puppetlabs/puppetserver/jruby_request.clj
@@ -57,7 +57,7 @@
 
 (defn wrap-with-environment-validation
   "Middleware function which validates the presence and syntactical content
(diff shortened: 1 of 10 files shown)


justinstoller:
NB: this shouldn't be merged in its current state. It is blocked on merging the ringutils code into ring-middleware. I'm opening this PRs so folks can see the effects of the consolidation. Once the updated ring-middleware is released this PR will need to be updated to pull in the new version of that library this PR will fail travis until then, it does allow those interested to test against a checkout of ring-middleware though.

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
```diff
@@ -148,10 +149,9 @@
                         "jruby-puppet.max-active-instances.")}))
      (when (jruby-schemas/shutdown-poison-pill? pool-instance#)
        (jruby/return-instance ~jruby-service pool-instance# ~reason)
-       (sling/throw+
-        {:type    ::service-unavailable
-         :message (str "Attempted to borrow a JRuby instance from the pool "
-                       "during a shutdown. Please try again.")}))
+       (mw/throw-service-unavailable!
```
KevinCorcoran: I think these helpers are nice but it's unfortunate that they're in a project called `ring-middleware` since it's the helpers themselves don't have anything to do with middleware and without already knowing the connection between the helpers and the middleware, it seems weird out of context.

KevinCorcoran:
🙈

justinstoller:
I think this is ready to go now

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
justinstoller: Yeah, after all that, it would probably have been better to at least break out the helpers into a different namespace so downstream consumers could do something like:
```clj
(ns puppetlabs.example.test
  (:require [puppetlabs.ring-middleware.core :as middleware]
            [puppetlabs.ring-middleware.utils :as ringutils]))

(let [stack (middleware/wrap-bad-request
              (fn [_] (ringutils/throw-bad-request! "Example")))
      response (stack (basic-request)]
  (is (logged? #".*Example.*" :error)))
```

It might also be a good idea because I would guess, outside of tests, consumers would use the different functionality in different namespaces as well.

KevinCorcoran: Yeah, that seems better to me as well.  Perhaps we should make this change along with the other work in the near future to restructure those error maps.

KevinCorcoran:
👍 

@nathanielksmith would you mind doing a 2nd review on this please?

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
justinstoller: I don't know how much ceremony is around z's but if it was a pretty lightweight process I'd be down to put up a maint PR that duplicated these helpers into a utils namespace and deprecated the ones in the core namespace this afternoon.

KevinCorcoran: Given that this project is still pre-1.0 (and likely only used by other puppetlabs projects) (and these functions are brand-spankin'-new) I would like to avoid deprecation and just move them.

The ceremony of a release is pretty much the same regardless of X, Y, or Z:
* update version number (if necessary)
* update changelog
* run jenkins job
* send announcement

src/clj/puppetlabs/puppetserver/jruby_request.clj
```diff
@@ -84,11 +64,11 @@
     (let [environment (get-environment-from-request request)]
       (cond
         (nil? environment)
-        (throw-bad-request!
+        (mw/throw-bad-request!
```
rlinehan: I think the docstring needs to be updated  - `:bad-request slingshot exception is thrown` rather than `::bad-request slingshot exception is thrown`.

rlinehan:
The two changes that this has caused seem to be the charset change in the response to CA errors from `text/plain` to `text/plain; charset=utf-8`. We've had some problems in the past related to this (SERVER-723), but that seems to have been because we were just returning `;charset=utf-8` without the `test/plain` MIME type - @camlow325 if you could verify that, that would be great.

The other change is that `:bad-request` is now thrown in places where `:puppetlabs.puppetserver.jruby-request/bad-request` was previously thrown. The two places this bubbles up are in `jruby-request-handler` in `request-handler-core` and in `wrap-with-environment-validation` in `master-core`. In both places, these errors are caught and don't bubble up to the user, so that seems to be fine.

camlow325:
> The two changes that this has caused seem to be the charset change in the response to CA errors from text/plain to text/plain; charset=utf-8. We've had some problems in the past related to this (SERVER-723), but that seems to have been because we were just returning ;charset=utf-8 without the test/plain MIME type - @camlow325 if you could verify that, that would be great.

Yeah, that all sounds right, @rlinehan.  In the PR for [SERVER-723](https://tickets.puppetlabs.com/browse/SERVER-723), #605, we changed some of the HTTP failure cases for the certificate endpoints that liberator was handling to have a Content-Type of `text/plain; charset=UTF-8` instead of `; charset=UTF-8`.  Assuming this PR would just allow for error cases where we just had `text/plain` to have `text/plain; charset-UTF-8`, I don't see a problem.  I can't think of any cases where the certificate endpoints would return plain text messages error messages formatted using a default encoding of ISO-8859-1.  UTF-8 should be fine for these.

I do think it would be good wherever we have tests that inspect the Content-Type to look for the full string, e.g., `text/plain; charset=UTF-8`, rather than just a regex, e.g., `text/plain.*`.  I haven't looked at this PR in any detail but it sounds like the resulting behavior shouldn't be problematic.

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
```diff
@@ -369,7 +369,7 @@
                           {:uri "/v1/certificate_statuses/"
                            :request-method :get})]
             (is (= 400 (:status response)))
-            (is (= "text/plain" (get-in response [:headers "Content-Type"])))
+            (is (re-matches #"text/plain.*" (get-in response [:headers "Content-Type"])))
```
rlinehan: per @camlow325 it would be good to just assert that this = `text/plain; charset=utf-8`

rlinehan:
I'm going to merge this because it is blocking CI. @justinstoller at some point would be good for you to file a maint PR to fix that one test. 🚢

PR #1032, opened 2016-04-21
DO NOT MERGE: 

erikPrime:
(MAINT) Modifies the confine statement so that... ...the test body is not entered just because a master is present.  
  Tested with ubuntu 15 master.  The puppetdb integration test was skipped as expected.  
  Tested with redhat6 master; the puppetdb_integration test ran as expected.

However, I'm still concerned about this change.  I understand the documentation for "confine" to say that we ought to be able to pass ", master" in, as the host array.  We want the confine to only evaluate the master, not the entire host array.  By deleting ", master", I'm asking the confine to evaluate the entire host array in the beaker options hash.

Read https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/structure.rb#L135 for more...

For our case, I think beaker confine is broken here:
https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/structure.rb#L205

### DIFF
--- a/acceptance/suites/tests/00_smoke/puppetdb_integration.rb
+++ b/acceptance/suites/tests/00_smoke/puppetdb_integration.rb
@@ -10,7 +10,14 @@
 ## PuppetDB successfully received the agent's report sent from Puppet Server.
 ## We can just run the agent that's on the master for this.
 
-confine :to, {:platform => puppetdb_supported_platforms}, master
+# This confine was originally
+# confine :to, {:platform => puppetdb_supported_platforms}, master
+# With , master present, execution unexpectedly enters the body of the
+# test when we expect the test to be skipped.  BKR-650 maybe relavent.
+# Be aware that future modifications to this test might, especially
+# if there are test operations added that run against agents instead of just
+# masters.
+confine :to, {:platform => puppetdb_supported_platforms}
 
 require 'json'
 


acceptance/suites/tests/00_smoke/puppetdb_integration.rb
```diff
@@ -10,7 +10,14 @@
...
 
-confine :to, {:platform => puppetdb_supported_platforms}, master
+# This confine was originally
+# confine :to, {:platform => puppetdb_supported_platforms}, master
+# With , master present, execution unexpectedly enters the body of the
+# test when we expect the test to be skipped.  BKR-650 maybe relavent.
+# Be aware that future modifications to this test might, especially
+# if there are test operations added that run against agents instead of just
+# masters.
+confine :to, {:platform => puppetdb_supported_platforms}
```
camlow325: @nwolfe Could you do a sanity check on this?  I'm thinking you put ", master" on there intentionally but I have no idea how any of this works.

 We found that this blew up with ...

```
<Beaker::DSL::Outcomes::FailTest: There should be one host with master defined
```

... if run against a test configuration matrix where no master from the test matrix matches a `puppetdb_supported_platforms` entry:

https://jenkins.puppetlabs.com/job/platform_puppetserver_integration-system_no-conditional_smoke-master/LAYOUT=ubuntu1510-64ma-64a,LDAP_TYPE=default,PLATFORM=default,label=beaker/18/consoleFull

Although ... it seems like this check has been sufficient to skip platforms previously until so, not sure.  Maybe this is even more broken with the newer beaker version that we bumped up to recently, 2.39.0.

Anyway, does this seem like the best way to work around the problem?

nwolfe: > I'm thinking you put ", master" on there intentionally

I did, because the code in here will install PuppetDB on the master, and if the master isn't a supported platform for PuppetDB, then we can't run this test.  It makes no sense without being tied to the master, but we don't need to necessarily use `confine` to tie it.

I thought I saw @erikPrime has refactored the conditional to not use `confine`, but instead just a `master.platform.include?(puppetdb_supported_platforms)` or something like that?  I think that would work here.

The buggy behavior is within `confine`, so we just want to avoid that method, but we definitely want to tie this to the master's platform somehow.

nwolfe: Maybe we just rewrite this as:
```ruby
skip_test unless puppetdb_supported_platforms.include? master.platform
```

(whatever the syntax for getting the master's platform is)

PR #1034, opened 2016-04-21
(MAINT) Assign default values expansion for all BEAKER_*

kurtwall:
Use Bash's syntax for assigning default values to variables so that values set in the environment can trump defaults set in the script. Update README to reflect these changes.


haus:
:+1:

PR #1037, opened 2016-04-22
(MAINT) Bump puppet agent dep to da1e61e

camlow325:
This commit bumps the puppet-agent dependency up to da1e61e and the
corresponding Ruby puppet submodule up to 57315f.


haus:
:+1:

PR #1040, opened 2016-04-25
(SERVER-1252) Document new tk-auth allow/deny options

gguillotte:
Once more, with ~~feeling~~ a master branch target per @KevinCorcoran.

CC @nathanielksmith, this is the same as PR #1015.


KevinCorcoran:
👍

PR #1045, opened 2016-04-28
(maint) Bump puppet version

melissa:
This bump includes updates to the Xenial puppet-agent build that
compiles facter with JNI support.


camlow325:
:+1:

PR #1046, opened 2016-04-28
adds method to merge rules to tk auth

erikPrime:


erikPrime:
For review only, not yet tested enough for merging.

acceptance/lib/helper.rb
```diff
@@ -239,6 +239,74 @@ def https_request(url, request_method, cert, key, body = nil)
...
+    allow_unauthenticated = 
+                  args[:allow_unauthenticated] || false
+    deny          = args[:deny] || false
+    sort_order    = args[:sort_order] || 77
+    path          = args[:path] || '/'
+    type          = args[:type] || 'path'
+    method        = args[:method] || ['head', 'get', 'put', 'post', 'delete']
+    query_params  = args[:query_params] || {}
+    #TODO: handle TK-293 X509 extensions.
+    authconf_file = args[:authconf_file] || options[:'puppetserver-confdir']+'/auth.conf'
```
kurtwall: I think user-specified values need to be validated before use. We might not be the only people who end up using this code.

Nit: And I *still* don't like this kind of function. The function signature tells me nothing about what it does, but I'm not interested in debating it anymore.

kurtwall:
It's a good start. The input validation is the only needful change. The function signature whing is just that.

acceptance/lib/helper.rb
erikPrime: I want to be able to add INCORRECT rules.  Maybe the thing to do is to add another arg that says "validate args"...

kurtwall: OIC. I think it is fine as it is.

acceptance/lib/helper.rb
```diff
@@ -239,6 +239,74 @@ def https_request(url, request_method, cert, key, body = nil)
...
+    name          = args[:name] || args[:cn]    #The friendly name of the match_request.
+    host          = args[:host] || master
+    allow         = args[:allow]|| args[:cn]
+    allow_unauthenticated = 
+                  args[:allow_unauthenticated] || false
+    deny          = args[:deny] || false
+    sort_order    = args[:sort_order] || 77
+    path          = args[:path] || '/'
+    type          = args[:type] || 'path'
+    method        = args[:method] || ['head', 'get', 'put', 'post', 'delete']
```
kurtwall: If this can store multiple values (in a vector), perhaps it could be named `methods` rather than `method`?

erikPrime:
Somehow I managed to add 90 commits to this PR; closing, destroying branch, etc...

PR #1049, opened 2016-04-29
(SERVER-1302) Merge-up stable -> master

KevinCorcoran:
The biggest source of merge conflicts was the documentation / markdown
files.  @gguillotte if you would please review the docs changes in this PR, that would be much appreciated.  All of the other conflicts were pretty minor and relatively easy
to resolve.

PR #1050, opened 2016-04-29
(doc) Update index.

gguillotte:
- Add static file content API endpoint doc.
- Add new auth and logging docs.
- Convert headings to sentence case per recent changes to company style.
- Structure introduction and usage docs into sections.
- Add note pointing to PE's docs.

### DIFF
--- a/documentation/index.markdown
+++ b/documentation/index.markdown
@@ -6,37 +6,43 @@ canonical: "/puppetserver/latest/"
 
 Puppet Server is the next-generation application for managing Puppet agents.
 
+> **Note:** For information about configuring and tuning settings specific to [Puppet Enterprise](https://docs.puppet.com/pe/), see [its documentation](https://docs.puppet.com/pe/latest/config_puppetserver.html).
 
-* [About Puppet Server](./services_master_puppetserver.markdown)
-* [Release Notes](./release_notes.markdown)
-* [Deprecated Features](./deprecated_features.markdown)
-* [Notable Differences vs. the Apache/Passenger Stack](./puppetserver_vs_passenger.markdown)
-* [Installation](./install_from_packages.markdown)
-* [Configuring Puppet Server](./configuration.markdown)
+* [**About Puppet Server**](./services_master_puppetserver.markdown)
+    * [Release notes](./release_notes.markdown)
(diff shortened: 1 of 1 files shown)


gguillotte:
CC @nfagerlund as most of these changes would need porting over to all the places where Server's sidebar navs live on the Docs site if accepted.

KevinCorcoran:
I'd appreciate it if we could let #1049 land before merging this, since it contains many docs changes, in the interest of avoiding any more conflicts in that already-huge merge-up.

gguillotte:
@KevinCorcoran Totally agreed. This is a very low priority.

documentation/index.markdown
```diff
@@ -6,37 +6,43 @@ canonical: "/puppetserver/latest/"
...
+    * [External SSL termination](./external_ssl_termination.markdown)
+    * [Tuning guide](./tuning_guide.markdown)
+    * [Restarting Puppet Server](./restarting.markdown)
+* **Known issues and workarounds**
+    * [Known issues](./known_issues.markdown)
+    * [SSL problems with load-balanced PuppetDB servers ("Server Certificate Change" error)](./ssl_server_certificate_change_and_virtual_ips.markdown)
+* **Administrative API endpoints**
+    * [Environment cache](./admin-api/v1/environment-cache.markdown)
+    * [JRuby pool](./admin-api/v1/jruby-pool.markdown)
+* **Server-specific Puppet API endpoints**
```
KevinCorcoran: What do you think about combining this section with the one immediately above?

KevinCorcoran:
👍  looks fine to me, just had the one question

documentation/index.markdown
gguillotte: @KevinCorcoran I can do that, though I'd still like to indicate that the admin and Puppet endpoints are at different locations. I'll figure out a way to do this.

rlinehan:
@gguillotte what's the status on this? looks like @KevinCorcoran had a question a while ago, wasn't sure if you were going to update this or if you would prefer we merge it as is. I'm 👍 , fwiw.

gguillotte:
@rlinehan I think we can merge this as-is, and I can follow on once I figure out structural changes to the API docs.

PR #1055, opened 2016-05-04
(SERVER-1303) Adds method that merges TK Auth rules into existing config

erikPrime:
This is a continued effort from PR #1046.  @kurtwall made some comments, and I made some minor updates to those comments.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -239,6 +239,74 @@ module PuppetServerExtensions
     response = http.request(request)
   end
 
+
+  # appends match-requests to TK auth.conf
+  #   Provides many defaults so that users of this method can simply 
+  #   and easily allow a host in TK auth.conf
+  #
+  # NOTE: This method allows the caller to define invalid TK auth rules
+  # by design.  
+  #
+  #   TK Auth is documented here:
+  #   https://github.com/puppetlabs/puppetserver/blob/master
+  #   /documentation/config_file_auth.md
+  #
+  #   Arguments:
+  #   cn:     The cannonical name, usually put in an "allow" or "deny"
+  #   name:   The friendly name of the match-request
+  #   host:   The system under test.  Typcially master.
+  #   allow:  hostname, glob, or regex lookback to allow.
+  #   allow_unauthenticated:
+  #           Boolean value.  Only adds the allow-unauthenticated behavior if
+  #           true.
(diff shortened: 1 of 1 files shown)


acceptance/lib/helper.rb
```diff
@@ -239,6 +239,74 @@ def https_request(url, request_method, cert, key, body = nil)
...
+    #have an invalid match-request.
+    match_request.merge!('allow' => allow) if allow
+    match_request.merge!('allow-unauthenticated' => true) if allow_unauthenticated
+    match_request.merge!('deny' => deny) if deny
+
+    authconf_text = on(master, "cat #{authconf_file}").stdout
+    authconf_hash = Hocon.parse(authconf_text)
+    authconf_hash['authorization']['rules'] << match_request
+ 
+    modify_tk_config(host, authconf_file, authconf_hash, true)
```
erikPrime: Yes, this is a beaker method; if during review, we decide we want to do it our own way with hocon, I can do that.

acceptance/lib/helper.rb
```diff
@@ -239,6 +239,74 @@ def https_request(url, request_method, cert, key, body = nil)
...
+    allow_unauthenticated = args[:allow_unauthenticated] || false
+    deny          	  = args[:deny] || false
+    sort_order   	  = args[:sort_order] || 77
+    path          	  = args[:path] || '/'
+    type     		  = args[:type] || 'path'
+    default_http_methods  = ['head', 'get', 'put', 'post', 'delete']
+    method        	  = args[:method] || default_http_methods
+    query_params  	  = args[:query_params] || {}
+    #TODO: handle TK-293 X509 extensions.
+    authconf_file 	  = args[:authconf_file] ||
```
jpinsonault: I don't know how much you or anyone else cares, but the spaces on the `=`s is off on some of these vars

erikPrime: Fixing...

acceptance/lib/helper.rb
jpinsonault: If it works, it works. Looks like it just dumps json. Im assuming we don't care about any comments that might be in there during an acceptance test. Not that a general hocon file modifying helper wouldn't be good.

jpinsonault:
👍

kurtwall:
👍

PR #1056, opened 2016-05-04
REVIEW ONLY (TK-293/SERVER-1268) Validates server support for CSR attributes

kurtwall:
This test verifies that Puppet Server can use CSR attributes to manage authorization.


kurtwall:
The test is working, but it does not yet actually use CSR attrs to permit/deny access. I'll need some help with that part.

acceptance/suites/tests/authorization/use_csr_auth.rb
```diff
@@ -0,0 +1,141 @@
...
+                  'acceptance/suites/tests/authorization/fixtures/csr_attributes.yaml',
+                  "#{confdir}",
+                  {:acceptable_exit_codes => [0]})
+    end
+  end
+end
+
+# # FIXME
+# step "Update tkauth config file to use CSR attribute to control access" do
+#   agents.each do |a|
```
erikPrime: There are two, maybe three ways of approaching this:
1.  Backup auth.conf, laydown the auth.conf we want to use for this test, hup.  restore backup and hup on teardown.
2.  Create an additional helper method, 'remove_match_request'.
Backup auth.conf, remote_match_request(#some rule set that would allow the agent by cn)
append_match_request(#rule set that allows access by extension), hup.
Reverse on teardown, or just restore backup.

acceptance/suites/tests/authorization/use_csr_auth.rb
```diff
@@ -0,0 +1,141 @@
...
+    if (not_controller(a))
+      rc = on(a,
+              puppet("agent --test --server #{server} --detailed-exitcodes"),
+              {:acceptable_exit_codes => [0,2]})
+    end
+  end
+end
+
+# Not anymore we don't
+step "Revoke and destroy the existing cert on the server" do
```
erikPrime: I don't see a revoke happening here.  This is probably just a nit, and cert destruction is probably sufficient for the needs of this test.  Revocation involves appending the certificate to the Certificate Revocation List, and bouncing the server (probably NOT hup).

erikPrime:
I think we also need to discuss if we are to merge this against master or stable...

KevinCorcoran:
I think it should be `master`, I think the changes tested here are only present in `master`.

erikPrime:
It sounds like @kurtwall is going to be unavailable for a while.  I'm closing this in favor of #1083.

PR #1057, opened 2016-05-05
(SERVER-1273) Add back in info that is missing from log messages

cprice404:
As of this commit in PUP:

https://github.com/puppetlabs/puppet/commit/6debe89bd935d56e2ec424900ad6f573bbf45552

It became critical that log destinations call `to_s` on the `Log` object
in order to ensure that they log all of the extra information (such as
backtraces).  Prior to this commit, the Puppet Server logback log
destination was calling `message` instead of `to_s`, so we were missing
all of the extra detailed information in our log messages.

### DIFF
--- a/src/ruby/puppetserver-lib/puppet/server/logger.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/logger.rb
@@ -5,7 +5,7 @@ java_import org.slf4j.LoggerFactory
 
 Puppet::Util::Log.newdesttype :logback do
   def handle(msg)
-    output = msg.message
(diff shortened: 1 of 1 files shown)


rlinehan:
👍 tested this out in the repl and it works... I'm guessing it's probably not a thing that is worth trying to write a test for?

However, should this be targeted at stable rather than master?

cprice404:
I felt like writing additional tests for testing log messages was probably more pain than benefit.  Targeting at stable is probably a good idea, though, can do.

PR #1058, opened 2016-05-06
(MAINT) Add helper functions encode_{cert,key} to support https_request

kurtwall:
When we ported `https_request()` to puppetserver acceptance, we forgot to bring along a couple of helpful methods for it. This commit remedies that omission.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -196,6 +196,26 @@ module PuppetServerExtensions
     stdout
   end
 
+  # Convert the contents of the certificate file in cert_file on the host
+  # specified by cert_host into an X.509 certificate and return it
+  # cert_host: The host whose cert you want
+  # cert_file: The specific cert file you want
+  # silent   : Suppress Beaker's output; set to false to see it
+  def encode_cert(cert_host, cert_file, silent = true)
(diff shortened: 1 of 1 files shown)


acceptance/lib/helper.rb
```diff
@@ -196,6 +196,26 @@ def get_defaults_var(host, varname)
...
+  # cert_file: The specific cert file you want
+  # silent   : Suppress Beaker's output; set to false to see it
+  def encode_cert(cert_host, cert_file, silent = true)
+    rawcert = on(cert_host, "cat #{cert_file}", {:silent => silent}).stdout.strip
+    OpenSSL::X509::Certificate.new(rawcert)
+  end
+
+  # Convert the contents of the private key file in key_file on the host
+  # specified by key_host into an RSA private key and return it
+  # cert_host: The host whose cert you want
```
camlow325: Should the comment here and the next line say `key_host` and `key_file`?

camlow325:
Seems fine to me.  Doesn't seem like any code was added to call these new methods.  Is that expected?

acceptance/lib/helper.rb
kurtwall: @camlow325 ++ Yeah. I'll push a fix. Misleading comments might be worse than incorrect code.

PR #1062, opened 2016-05-09
(PE-13485) Implement a consistent request logging format

cprice404:
Prior to this commit puppetserver used a different request
logging format than puppetdb and console-services.

After this commit the request logging will be of the suggested
formatting for all trapper-keeper projects.

### DIFF
--- a/resources/ext/config/request-logging.xml
+++ b/resources/ext/config/request-logging.xml
@@ -2,7 +2,7 @@
   <appender name="FILE" class="ch.qos.logback.core.FileAppender">
     <file>/var/log/puppetserver/puppetserver-access.log</file>
     <encoder>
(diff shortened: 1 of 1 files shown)


rlinehan:
👍

It seems a bit concerning that we merged https://github.com/puppetlabs/puppetserver/pull/886 but it isn't in any of our branches... it looks like we last merged up 2.1.x in December, but this was merged in January, and then we deleted that branch in March....

cprice404:
@rlinehan thx; when this gets merged I will try to do some merge-ups across all the branches, and will look into the 2.1.x thing as part of that.

camlow325:
👍 Looks like the prior PR did, so I assume it's good.

PR #1063, opened 2016-05-09
(maint) add hup helper

erikPrime:


acceptance/lib/helper.rb
```diff
@@ -259,7 +285,34 @@ def https_request(url, request_method, cert, key, body = nil)
     response = http.request(request)
   end
 
-
+  def hup_server(host = master, timeout = 30)
+    pid=on(host, 'cat /var/run/puppetlabs/puppetserver/puppetserver').stdout.chomp
```
kurtwall: Not really excited about a hard-coded path. Too bad we don't have a `puppetserver config print pidfile`. The path seems unlikely to change, but...

erikPrime: I couldn't figure out what to query in order to make this a variable.

acceptance/lib/helper.rb
```diff
@@ -259,7 +285,34 @@ def https_request(url, request_method, cert, key, body = nil)
...
+    on(host, "kill -HUP #{pid}")
+    url="https://#{host}:8140/puppet/v3/status"
+    cert=get_cert(master)
+    key=get_key(master)
+    response_code = "0"
+    sleeptime = 1
+    while response_code != "404" && timeout > 0
+      sleep sleeptime
+      begin
+        response = https_request(url, 'GET', cert, key)
```
kurtwall: Not sure it matters here, but I coded `https_request` to use keywords for the method parameter, thus:
`https_request(url, :get, cert, key)`

acceptance/lib/helper.rb
```diff
@@ -259,7 +285,34 @@ def https_request(url, request_method, cert, key, body = nil)
...
+      rescue StandardError => e
+        expected_errors = [ Errno::ECONNREFUSED, Errno::ECONNRESET, 
+                            OpenSSL::SSL::SSLError ]
+        if expected_errors.include?e.class 
+          raise e
+        else
+          # Does this message violate the Wolfe Principle?
```
kurtwall: "Wolfe Principle"?

erikPrime: Wolfe Principle = If it isn't an error, or if it isn't broken, don't log it.

acceptance/lib/helper.rb
rlinehan: you could do `puppet master --configprint rundir` + `puppetserver/puppetserver` I think.

kurtwall:
If you'd put whitespace around operators consistently, I would shake your hand.

acceptance/lib/helper.rb
erikPrime: ? Not in the copy I'm looking at?  Do you mean in pe-puppetserver-extensions?  Or in a later commit?

erikPrime:
White space added...

kurtwall:
🙈

PR #1064, opened 2016-05-09
(SERVER-1273) Add back in info that is missing from log messages

cprice404:
As of this commit in PUP:

https://github.com/puppetlabs/puppet/commit/6debe89bd935d56e2ec424900ad6f573bbf45552

It became critical that log destinations call `to_s` on the `Log` object
in order to ensure that they log all of the extra information (such as
backtraces).  Prior to this commit, the Puppet Server logback log
destination was calling `message` instead of `to_s`, so we were missing
all of the extra detailed information in our log messages.

### DIFF
--- a/src/ruby/puppet-server-lib/puppet/server/logger.rb
+++ b/src/ruby/puppet-server-lib/puppet/server/logger.rb
@@ -5,7 +5,7 @@ java_import org.slf4j.LoggerFactory
 
 Puppet::Util::Log.newdesttype :logback do
   def handle(msg)
-    output = msg.message
+    output = msg.to_s
     if msg.source.size > 0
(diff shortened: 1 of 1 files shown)


cprice404:
This replaces #1057 to target at stable instead of master.

rlinehan:
👍

camlow325:
👍  Functionally tested this by having a parser function raise an exception.  Confirmed that the full stack trace did not appear when testing from code currently on the master branch but did appear when using code from this PR.  Didn't see any other obvious regressions in terms of output that is logged.

rlinehan:
Merging since both @camlow325 and I are good. 🚢

PR #1065, opened 2016-05-09
(doc) Fix and update links in environment_classes doc

gguillotte:
Resolves DOC-2704.

### DIFF
--- a/documentation/puppet-api/v3/environment_classes.markdown
+++ b/documentation/puppet-api/v3/environment_classes.markdown
@@ -4,17 +4,17 @@ title: "Puppet Server: Puppet API: Environment Classes"
 canonical: "/puppetserver/latest/puppet/v3/environment_classes.html"
 ---
 
-[classes]: /puppet/latest/reference/lang_classes.html
-[deprecated WEBrick Puppet master]: /puppet/latest/reference/services_master_webrick.html
-[node definitions]: /puppet/latest/reference/lang_node_definitions.html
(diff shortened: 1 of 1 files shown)


camlow325:
🙈 @gguillotte - are you okay with these changes only being on our master branch or would you like them on our stable branch as well?  I can't say for sure whether or not any new releases will be done from the current incarnation of our stable branch.  Also, do you think anyone else from the docs team needs to look at this before we'd merge it?

gguillotte:
@camlow325 Yeah, I should have probably targeted this at stable. The changes should be on both branches. I can close and refile against stable if that'd be easier.

camlow325:
@gguillotte Yeah, if you want the changes on both branches, it would be better to target them at stable so that they get merged up from their to the master branch.  We don't often do full branch merges in the other direction.

rlinehan:
@gguillotte @camlow325 we're about to do a merge down of master to table for the release, so maybe this is actually okay? Unless we ever might need to pin something to stable at a SHA before the merge down and that would need to include this.

camlow325:
Yeah, assuming it's probably unlikely that we'll resurrect a 2.3.x branch for a special release later and would want these changes to go in the docs for that, I think it might be better to just put this in the master branch.  @rlinehan - if you want to just merge this now and @gguillotte is good with that, I'm 👍 as well.

gguillotte:
:+1:

PR #1066, opened 2016-05-09
(MAINT) merge up from 1.x to stable, 2016 05 09

cprice404:
only conflict was the ezbake version in the project.clj.


camlow325:
👍

PR #1067, opened 2016-05-09
(MAINT) merge old 2.1.x stuff into stable - 2016-05-09

cprice404:
We noticed that Nick's log format changes commit appeared to have landed in a 2.1.x branch and never got merged up from there.  This commit merges up to stable from the commit where his PR originally landed.  Also checked the latest tag from the 2.1.x branch (2.1.2), and there was nothing else to bring in from that tag.

Conflicts here were in:

Gemfile
acceptance/README.md
acceptance/scripts/generic/testrun.sh

In all cases I just kept the content from the current `stable` branch, because I didn't see anything that looked relevant from the 2.1.x versions.


cprice404:
@rlinehan this is my attempt to make sure we get as much of the history from that 2.1.x branch as possible.  :fried_shrimp:

KevinCorcoran:
Is it correct that this PR has an empty diff?

cprice404:
@KevinCorcoran I think so... some of these commits were duplicated, and there's some revert shenanigans going on as well.  This was mostly just about capturing some missing history that @rlinehan discovered this morning.

KevinCorcoran:
ok.  🙈

camlow325:
👍

PR #1069, opened 2016-05-10
Maint/stable/merge down from master for 2.3.3 xenial release

cprice404:


cprice404:
closing because i screwed up the tk-auth version, and because we don't really want to include it in this release

PR #1072, opened 2016-05-10
(TK-360) Fix acceptance tests testing denied response

rlinehan:
Previously, all the checks of requests being denied were matching against the
rule. However, for TK-360 the rule name was removed from the response. This
commit updates those assertions to match against the path and method (which
are in the response) instead.

### DIFF
--- a/acceptance/suites/tests/authorization/default_rules.rb
+++ b/acceptance/suites/tests/authorization/default_rules.rb
@@ -64,7 +64,7 @@ with_puppet_running_on(master, {}) do
     assert_allowed
 
     curl_unauthenticated('/puppet/v3/environments')
-    assert_denied(/denied by rule 'puppetlabs environments'/)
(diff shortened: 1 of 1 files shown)


camlow325:
I'm okay with the changes on the PR.  Although I'm tempted to just change it to ...

```ruby
def assert_denied()
   assert_match(/Forbidden request/, stdout)
   assert_match(/^STATUSCODE=403$/, stdout)
 end
```

... and have the individual calls be ...

```ruby
assert_denied()
```

Previously, the tests could validate that the request was rejected because it matched a specific rule from the configuration.  Where we don't have that context in the error message anymore, just validating that we got a "Forbidden request" / 403 failure seems like it would be good enough now - rather than having to validate that the precise endpoint and method is round-tripped for each request.

@erikPrime - do you have an opinion on this?

erikPrime:
@camlow325  No I don't have an opinion on this...

PR #1073, opened 2016-05-10
Add maintainers section

karenvdv:


erikPrime:
:+1:

rlinehan:
I hate to be that person, but should this be targeted at stable? or is it okay to just have the maintainers section in our README on master? (we'll be merging down soon anyway, so it probably doesn't matter a ton)

erikPrime:
@rlinehan I kind of thought that if we got this in soon, it would get merged down...

rlinehan:
@erikPrime yeah, that seems reasonable. 👍

PR #1070, opened 2016-05-10
(MAINT) update to tk-auth 0.6.1

cprice404:

PR #1075, opened 2016-05-11
(MAINT) Added explicit ring-servlet 1.4.0 pin

camlow325:
This commit adds an explicit ezbake packaging pin to ring/ring-servlet
1.4.0.  In prior commits, the ezbake-time dependency on
puppetlabs/trapperkeeper-webserver-jetty9 1.5.6 would try to bring in
ring/ring-servlet 1.1.8 whereas the recently added dependency on
puppetlabs/trapperkeeper-authorization 0.6.1 would try to bring in
ring/ring-servlet 1.4.0.  This commit resolves the conflict in favor of
the later ring/ring-servlet version.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -123,6 +123,7 @@
                                                [org.clojure/clojure ~clj-version]
                                                [puppetlabs/puppetserver ~ps-version]
                                                [puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty-version]
+                                               [ring/ring-servlet "1.4.0"]
                                                [org.clojure/tools.nrepl "0.2.3"]]
                       :plugins [[puppetlabs/lein-ezbake "0.3.25"]]
                       :name "puppetserver"}


camlow325:
@rlinehan @cprice404 I think this will resolve the current package build issue that was introduced by #1070.

cprice404:
@camlow325 is there not some other library that we could make this change in to resolve the problem?  Making the change in the ezbake section feels a little wonky to me.

camlow325:
@cprice404 The other way I think we could do this is to roll tk-jetty9 to ring-servlet 1.4.0 and then bump puppetserver's dependency up to that later tk-jetty9 version.  Would you rather do that?

cprice404:
Yeah, I think so.  :/  The ezbake profile just seems like a scary place to be resolving dependency issues for some reason.

camlow325:
k, I'll try that instead.  Closing this PR.

PR #1076, opened 2016-05-11
(MAINT) Update to tk-j9 1.5.9

cprice404:


rlinehan:
Needs a release of tk-j9, but tested with `lein with-profile +ezbake` (after doing a `lein install`, which I couldn't remember if was necessary or not) and it seems to fix the problem.

rlinehan:
tk-w-j9 1.5.9 has been released and travis passed, merging.

PR #1077, opened 2016-05-11
(MAINT) add hyphen back to uberjar name

cprice404:
This is currently blocked on a tk-j9 PR and release.

Because of how EZBake finds the pid on Cent6 (involves a pgrep for
the jar file name), upgrades on that distro will fail if we change
the jar name without changing the ezbake logic as well.  It seems
easier to just change the jar name back for now.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -77,7 +77,7 @@
 
   :plugins [[lein-release "1.0.5" :exclusions [org.clojure/clojure]]]
 
-  :uberjar-name "puppetserver-release.jar"
+  :uberjar-name "puppet-server-release.jar"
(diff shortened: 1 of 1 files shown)


camlow325:
The `test-restart-comes-back` test failed in Travis on openjdk7 only.  Appears to be due to a timeout being tripped in the test because of a slow Travis run - unrelated to changes in this PR.  I put up a separate PR to bump up the timeout in #1078.  I think we can ignore this cell's test failure in the context of this PR.

rlinehan:
Closing, this is superseded by #1079.

PR #1079, opened 2016-05-11
(maint) add hyphen back to uberjar name

rlinehan:
Because of how EZBake finds the pid on Cent6 (involves a pgrep for
the jar file name), upgrades on that distro will fail if we change
the jar name without changing the ezbake logic as well.  It seems
easier to just change the jar name back for now.

### DIFF
--- a/acceptance/suites/tests/00_smoke/validate-excluded-bc-jars.rb
+++ b/acceptance/suites/tests/00_smoke/validate-excluded-bc-jars.rb
@@ -1,10 +1,10 @@
 
 test_name "Validate excluded BC jars are not in packaged uberjar."
 
-package_name = options['puppetserver-package'] 
+package_name = options['puppetserver-package']
 install_dir = get_defaults_var(master, "INSTALL_DIR")
 
-jarfile = File.join(install_dir, "puppetserver-release.jar")
+jarfile = File.join(install_dir, "puppet-server-release.jar")
 
 on(master, "test -e \"#{jarfile}\"")
 install_package(master, "unzip")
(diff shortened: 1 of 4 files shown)


cprice404:
+1.  This doesn't seem like it needs a second review, I'd go ahead and merge it now but I'm not sure if we're waiting for current CI runs for anything.  If not, @rlinehan I'd say just push the button whenever you're back at it.

camlow325:
One other change I think we should revert - doesn't absolutely have to be in this PR but it's related - is this reference in the documentation/restarting.markdown file:

```
kill -HUP `pgrep -f puppetserver`
```
I think that should now go back to:

```
kill -HUP `pgrep -f puppet-server`
```

👍 otherwise, though.

rlinehan:
@camlow325 updated

rlinehan:
I'll merge this once the previous CI jobs have gone through, to make sure the dependency issues have been resolved before we change anything else.

camlow325:
:shipit:

rlinehan:
Previous CI that was red is now passing, so going to merge this even though travis isn't done - it's got nothing that should affect/be affected by travis.

KevinCorcoran:
Instead of pgrepping for the PID you can cat the pidfile which is not so brittle.

cprice404:
@KevinCorcoran I assume that the RE team thought of that when they chose the pgrep route, and had a reason for going the way they did.

camlow325:
IIRC, the PID file doesn't get written out at startup across all OSes.  Also, I think there are some race conditions where the content of the file could be written incorrectly.

KevinCorcoran:
brittle

PR #1080, opened 2016-05-11
(TK-360) Fix acceptance tests testing denied response

rlinehan:
Previously, all the checks of requests being denied were matching against the
rule. However, for TK-360 the rule name was removed from the response. This
commit updates those assertions to match against the path and method (which
are in the response) instead.


nathanielksmith:
:+1:

camlow325:
👍

PR #1081, opened 2016-05-11
(SERVER-1314) Corrects a teardown problem

erikPrime:


acceptance/suites/tests/code_commands/code_scripts.rb
```diff
@@ -137,43 +161,28 @@ class { '::ntp': }
...
-      }
-    hocon_setting { 'code-content-command-script' :
-      ensure => present,
-      path => '/etc/puppetlabs/puppetserver/conf.d/puppetserver.conf',
-      setting => 'versioned-code.code-content-command',
-      value => '/opt/puppetlabs/server/apps/puppetserver/code-content-command_script.sh',
-      }
-    CICS
-  create_remote_file(master, '/tmp/config_code_id_command_script.pp', cicsetting)
+  create_remote_file(master, '/tmp/config_code_id_command_script.pp', cicsetting() )
```
justinstoller: I don't think the empty parens for a method invocation with an empty parameter list -- or the space between them and the closing paren in the larger parameter list -- are idiomatic. Though it is clearer, and I know that QA has been using styles that are more explicit than idiomatic Ruby usually is. I'd defer to whatever is idiomatic for QA, but figured I should note that.

justinstoller:
👍 This looks like a good change and important fix.

rlinehan:
This is a nitpick and more a note for the future, but... I think that commit messages are really important for understanding why something changed - especially if you go back and look at it several weeks/months later and are trying to figure out why the hell something was changed in the way it was. The commit message for this doesn't say anything about what was going on previously or what exactly this fixed. Also, we usually try to have the title of the commit message be imperative - i.e. "Correct a teardown problem" rather than "Corrects a teardown problem."

rlinehan:
I think this makes sense to me - seems like we're now making sure the puppetserver config setting isn't there, in addition to removing the files the setting referenced? Seems like a good idea.

rlinehan:
Ran this test locally and it worked, so 👍.

PR #1082, opened 2016-05-11
(MAINT) updates GEM_SOURCE default

erikPrime:
In most test scripts, and all the CI pipelines I've looked at, we set our GEM_SOURCE to http://rubygems.delivery.puppetlabs.net.  Performing bundle operations while inadvertently having to failed EXPORT GEM_SOURCE=http://rubygems.delivery.puppetlabs.net is a constant annoyance.

Let's set the default to the thing that works.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,4 @@
-source ENV['GEM_SOURCE'] || "https://rubygems.org"
+source ENV['GEM_SOURCE'] || "http://rubygems.delivery.puppetlabs.net"
 
 def location_for(place, fake_version = nil)
   if place =~ /^(git[:@][^#]*)#(.*)/


justinstoller:
When/why does using rubygems.org not work?

erikPrime:
Generally, rubygems.delivery.puppetlabs.net has more correct versions.  Scooter is a specific example.

haus:
This means that running specs when not on the VPN will require manual intervention in some cases, which seems less than ideal.

erikPrime:
@Haus I guess the question on if this is a good thing hinges on how often we run spec tests from the puppetserver repo then?

haus:
@erikPrime well travis runs specs from here. I'm a little confused to how it is able to pass with this change in, as it doesn't look like it's exporting GEM_SOURCE.

erikPrime:
@Haus I THINK that Travis uses the gem file from puppet core.  I found this snippet in a travis log...
BUNDLE_GEMFILE='/home/travis/build/puppetlabs/puppetserver/ruby/puppet/Gemfile'

haus:
Ah yes, it does. https://github.com/puppetlabs/puppetserver/blob/master/Rakefile#L54

erikPrime:
So this means that we can run spec tests without being on the VPN then right?  And that this is merge-able?

haus:
Think so.

rlinehan:
seems like we've agreed this is okay? merging.

PR #1083, opened 2016-05-11
(SERVER-1268/TK-293) validates that TK Auth can allow access...

erikPrime:
based on an x509 ssl extension rather than a hostname/cn.

This is %98 percent based on @kurtwall 's work in #1056.  This is an end user style test rather than an API level test.

If possible, it would be nice to get this merged into master before we merge down to stable.  If that isn't possible QA will adapt...

@johnduarte When you have a little time could you please review this?

### DIFF
--- a/acceptance/suites/tests/authorization/x509_auth.rb
+++ b/acceptance/suites/tests/authorization/x509_auth.rb
@@ -1,0 +1,136 @@
+test_name "(SERVER-1268)/(TK-293) TK-AUTH uses certificate extensions for authentication" do
+
+confine :except, :platform => 'windows'
+  
+server = master.puppet['certname']
+ssldir = master.puppet['ssldir']
+confdir = master.puppet['confdir']
+
+teardown do
+  # restore the original tk auth.conf file
+  on master, 'cp /etc/puppetlabs/puppetserver/conf.d/auth.bak /etc/puppetlabs/puppetserver/conf.d/auth.conf'
+end
+
+step "Backup the tk auth.conf file" do
+  on master, 'cp /etc/puppetlabs/puppetserver/conf.d/auth.conf /etc/puppetlabs/puppetserver/conf.d/auth.bak'
+end
+
+# Do we have a functioning cert?
+step "Confirm agent can connect with existing cert" do
+  agents.each do |a|
+    if (not_controller(a))
+      rc = on(a,
+              puppet("agent --test --server #{server} --detailed-exitcodes"),
+              {:acceptable_exit_codes => [0,2]})
+    end
+  end
+end
+
+# Not anymore we don't
+step "Revoke and destroy the existing cert on the server" do
+  agents.each do |a| 
+    if (not_controller(a))
+      rc = on(master, 
+              puppet("cert destroy #{a.hostname}"),
+              {:acceptable_exit_codes => [0,2]})
+    end
+  end
+end
+
+step "HUP the server" do
+  hup_server
+end
+
+# After a server HUP, the agent cert should be rejected
+step "Confirm agent can't connect with existing cert" do
+  agents.each do |a|
+    if (not_controller(a))
+      rc = on(a,
+              puppet("agent --test --server #{server} --detailed-exitcodes"),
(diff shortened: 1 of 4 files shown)


acceptance/suites/tests/authorization/x509_auth.rb
```diff
@@ -0,0 +1,139 @@
...
+              {:acceptable_exit_codes => [0,2]})
+    end
+  end
+end
+
+# Not anymore we don't
+step "Revoke and destroy the existing cert on the server" do
+  agents.each do |a| 
+    if (not_controller(a))
+      nodename = fact_on(a, 'fqdn')
```
johnduarte: Why not use the beaker host's certname value `a.puppet['certname']` rather than performing an additional facter call?

acceptance/suites/tests/authorization/x509_auth.rb
```diff
@@ -0,0 +1,139 @@
...
+              puppet("agent --test --server #{server} --detailed-exitcodes"),
+              {:acceptable_exit_codes => [1]})
+    end
+  end
+end
+
+step "Remove the old certs on the agents so they'll make new ones" do
+  agents.each do |a|
+    if (not_controller(a))
+      nodename = fact_on(a, 'fqdn')
```
johnduarte: Why not use the beaker host's certname value a.puppet['certname'] rather than performing an additional facter call?

erikPrime: Agreed.  I'll fix it.

acceptance/suites/tests/authorization/x509_auth.rb
erikPrime: Agreed. I'll fix it.

acceptance/suites/tests/authorization/x509_auth.rb
```diff
@@ -0,0 +1,139 @@
...
+# a new cert
+# TODO: Make this a here doc with extensions that exist as vars so that they 
+# can be passed into our tk auth.conf rule generator.
+step "Copy the CSR attributes file into place" do
+  agents.each do |a|
+    if (not_controller(a))
+      rc = scp_to(a,
+                  'acceptance/suites/tests/authorization/fixtures/csr_attributes.yaml',
+                  "#{confdir}",
+                  {:acceptable_exit_codes => [0]})
```
johnduarte: Thinking out loud: Are there specifications for ownership/mask for the `csr_attributes` file? If so, they should be adhered to here. Separate test may want to validate that the system enforces them.

acceptance/suites/tests/authorization/x509_auth.rb
erikPrime: On a second look, facter runs MUCH faster.  But, we don't need to go to the agent at all, the certname/hostname is in the options hash.  Lets get it from there

johnduarte: It is true that `hostname` and `certname` are the same *most* of the time, but this is not guaranteed. This is an edge case and I don't know if it is worth guarding against, but I wanted to bring it up.

Example
```
  sol10.delivery.puppetlabs.net (solaris-10-sparc) 13:16:14$ puppet agent --test --server v2u8phod7ts8nak.delivery.puppetlabs.net
    Info: Creating a new SSL key for unknown.delivery.puppetlabs.net
    Info: Caching certificate for ca
    Info: csr_attributes file loading from /etc/puppetlabs/puppet/csr_attributes.yaml
    Info: Creating a new SSL certificate request for unknown.delivery.puppetlabs.net
    Info: Certificate Request fingerprint (SHA256): BE:62:68:35:61:75:A7:49:1C:DE:94:98:73:5D:34:F8:FE:19:D8:04:ED:8D:B4:4E:33:F3:4E:7D:B6:FF:E9:31
    Info: Caching certificate for ca
    Exiting; no certificate found and waitforcert is disabled
```

acceptance/suites/tests/authorization/x509_auth.rb
erikPrime: Not that I know of...

PR #1086, opened 2016-05-12
(maint) Update ezbake to pick up systemd for xenial

underscorgan:


geoffnichols:
👍 , pending Travis

underscorgan:
This resulted in puppetserver packages on trusty that won't start with java 7. DO NOT MERGE.

underscorgan:
This works now that I built it with the correct java.

KevinCorcoran:
Is there a JIRA ticket about adding Xenial support which could be mentioned in the commit message?

👍 otherwise

underscorgan:
@KevinCorcoran there doesn't seem to be one for the foss server stuff, you ok with this staying as a maint since it's a dependency bump?

rlinehan:
I'm cool with this being maint, also would like to get this in before merging down master to stable, which I'd like to do today. So 👍.

PR #1090, opened 2016-05-16
(MAINT) changes the hup helper to use pgrep -f 

erikPrime:
like @camlow325 said in the first place

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -290,8 +290,7 @@ module PuppetServerExtensions
   end
 
   def hup_server(host = master, timeout = 30)
-    pidfile = on(host, 'puppet master --configprint rundir').stdout.chomp + '/puppetserver' 
-    pid = on(host, "cat #{pidfile}").stdout.chomp
+    pid = on(host, 'pgrep -f puppetserver').stdout.chomp
     on(host, "kill -HUP #{pid}")
     url = "https://#{host}:8140/puppet/v3/status"
     cert = get_cert(master)


erikPrime:
Ping @johnduarte or @rlinehan  I think this will make our CI green; but on @rlinehan 's request we are pushing this to stable first before merging up to master...

acceptance/lib/helper.rb
```diff
@@ -290,8 +290,7 @@ def https_request(url, request_method, cert, key, body = nil)
   end
 
   def hup_server(host = master, timeout = 30)
-    pidfile = on(host, 'puppet master --configprint rundir').stdout.chomp + '/puppetserver' 
-    pid = on(host, "cat #{pidfile}").stdout.chomp
+    pid = on(host, "pgrep -f #{options['puppetservice']}").stdout.chomp
```
camlow325: I think we might need to just do 'puppetserver' instead of `options['puppetservice']` here.  I think the name of the jar file in the Java process contains "puppetserver" and not "pe-puppetserver", both for OSS and PE.

rlinehan: Confirmed this on a PE box:

```
[root@pecouch-latest ~]# pgrep -f pe-puppetserver
[root@pecouch-latest ~]# pgrep -f puppetserver
1332
```

haus: @rlinehan didn't the hyphen changes get backed out? shouldn't it be puppet-server if you want to find the jar? (i think puppetserver is in the path to the jar, but isn't the jarfile itself)

camlow325: @haus Works both ways for me - puppet-server and puppetserver - on a PE Puppet Server package build that has "puppet-server-release.jar".  Maybe that works because the path to the jar has "/puppetserver/" in it even though the immediate name of the jar has "/puppet-server-release.jar"?  But, yeah, maybe including the dash here would be better.

johnduarte: @haus the `-f` flag is being used in this PR. So it should be resistance to such changes as long as `puppetserver` is present anywhere in the command.

Current result with `-l`:
```
# pgrep -l -f 'puppetserver'
9355 /usr/bin/java -XX:OnOutOfMemoryError=kill -9 %p -Djava.security.egd=/dev/urandom -Xms2g -Xmx2g -XX:MaxPermSize=256m -cp /opt/puppetlabs/server/apps/puppetserver/puppet-server-release.jar clojure.main -m puppetlabs.trapperkeeper.main --config /etc/puppetlabs/puppetserver/conf.d -b /etc/puppetlabs/puppetserver/bootstrap.cfg
```

haus: @johnduarte sure. that's fine.

i just wanted to clarify if we're looking for the jar itself or the path to the jar.

camlow325:
👍

camlow325:
🚢

PR #1092, opened 2016-05-16
(SERVER-1334) Bump puppet and hiera versions

rlinehan:
to what are included in puppet-agent 1.5.0, and bump PUPPET_BUILD_VERSION in the acceptance helper to match.

Also change the puppetserver version from 2.4.0-master to 2.4.0-stable.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -30,7 +30,7 @@ module PuppetServerExtensions
     puppet_build_version = get_option_value(options[:puppet_build_version],
                          nil, "Puppet Agent Development Build Version",
                          "PUPPET_BUILD_VERSION",
-                         "732e883733fe5e5989afe330e3c5cea00b678d1e", :string)
+                         "1.5.0", :string)
 
     # puppetdb version corresponds to packaged development version located at:
     # http://builds.delivery.puppetlabs.net/puppetdb/
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -30,7 +30,7 @@ module PuppetServerExtensions
     puppet_build_version = get_option_value(options[:puppet_build_version],
                          nil, "Puppet Agent Development Build Version",
                          "PUPPET_BUILD_VERSION",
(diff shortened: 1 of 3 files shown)


erikPrime:
Looks good to me.  I didn't validate the new shas are correct, only read through the diffs.

camlow325:
@rlinehan Should we go ahead and bump the minimum rpm / deb dependencies in the ezbake.conf file in this PR as well?

rlinehan:
@camlow325 I asked @underscorgan that in chat and she said it wasn't necessary - nothing we've done in this release depends on puppet-agent 1.5.0, except for Wily and Xenial, but they don't have packages for 1.4.0 so the >= relationship should work. Do we want to require updating to 1.5.0 for any reason?

camlow325:
@rlinehan @underscorgan Are we sure no puppet-agent 1.4.0 packages for Wily or Xenial have gone out?  This - http://builds.puppetlabs.lan/puppet-agent/1.4.0/shipped/deb/wily/PC1/ - made me think there were some packages that had shipped for Wily at least.  And I was thinking we'd want to bump to at least a 1.5.0 minimum for Deb distros so that people don't try to use the puppet-agent 1.4.0 packages that we know won't work with Puppet Server.

camlow325:
@haus confirmed that we do have puppet-agent packages for 1.4.2 on Xenial already - http://apt.puppetlabs.com/pool/xenial/PC1/p/puppet-agent/.  That package is known to be broken with respect to Facter JNI support, so we'll definitely want to bump our minimum version dependency for Puppet Server 2.4.0 to puppet-agent 1.5.0 so that users don't accidentally end up using the incompatible puppet-agent version.

rlinehan:
@camlow325 @haus @underscorgan updated.

haus:
:+1:

camlow325:
👍 Going to merge.  Assuming Travis will be happy later...

underscorgan:
👍 didn't know there were existing/broken packages out there already, so bumping is good.

PR #1097, opened 2016-05-19
(doc) Add 2.4.0 release notes.

gguillotte:


documentation/release_notes.markdown
```diff
@@ -4,116 +4,104 @@ title: "Puppet Server: Release Notes"
...
+## Puppet Server 2.4
 
-Released April 27, 2016.
+Released May 19, 2016.
 
-This is a bug-fix release that also resolves a security issue in endpoint URL decoding.
+This is a feature and bug-fix release of Puppet Server that also upgrades its included [Trapperkeeper][] framework from version 1.3.1 to 1.4.0.
 
-### **Security fix:** CVE-2016-2785 - Correctly decode endpoint URLs
+This release also adds packages for Ubuntu 16.04 LTS (Xenial Xerus) and no longer includes packages for Fedora 21, which reached its end of life in December.
```
rlinehan: I think it also adds packages for Ubuntu 15.10 (Wily Werewolf)? I think we didn't have those before. But may want to confirm with @mckern or @underscorgan.

documentation/release_notes.markdown
```diff
@@ -4,116 +4,104 @@ title: "Puppet Server: Release Notes"
...
-
-* [All Puppet Server issues targeted at this release](https://tickets.puppet.com/issues/?jql=project%20%3D%20SERVER%20AND%20fixVersion%20%3D%20%22SERVER%202.3.2%22%20ORDER%20BY%20updated%20DESC%2C%20priority%20DESC%2C%20created%20ASC)
-
-## Puppet Server 2.3.1
-
-Released March 21, 2016.
+Trapperkeeper 1.4.0 introduces improvements to its [service bootstrapping][] feature, which Puppet Server uses to manage its certificate authority (CA) service. These improvements include better error tolerance, improved logging of bootstrap-related errors, and the ability to read `bootstrap.cfg` settings from multiple configuration files and directories.
 
-This is a bug-fix release that resolves a disruptive logging configuration issue.
+For more information, see the following [Trapperkeeper project](https://tickets.puppetlabs.com/browse/TK) tickets:
```
rlinehan: I don't know if we actually want to put this into the release notes, but it might be worth noting that we haven't actually changed Puppet Server to use the new split bootstrap.cfg settings that TK 1.4.0 provides out of the box.

KevinCorcoran:
@gguillotte I've only very briefly skimmed these, but they look really awesome!

Also, I think I asked this a year or two ago, but isn't it weird that we don't keep release notes for old versions anywhere?  It seems really weird to just delete them.

documentation/release_notes.markdown
```diff
@@ -4,116 +4,104 @@ title: "Puppet Server: Release Notes"
...
 
-* [SERVER-999](https://tickets.puppet.com/browse/SERVER-999)
+-   [TK-293: tk-auth should support x.509 extensions for authentication instead of just certname](https://tickets.puppetlabs.com/browse/TK-293)
 
-### New feature: `environment_classes` API
+Server 2.4.0 also reads custom OID shortname maps defined in Puppet's [`custom_trusted_oid_mapping.yaml`](https://docs.puppet.com/puppet/latest/reference/config_file_oid_map.html).
 
-The [environment classes API][] in Puppet Server 2.3.0 serves as a replacement for the Puppet [resource type API][] when requesting information about [classes][] available to a Puppet Server.
+-   [SERVER-1150](https://tickets.puppetlabs.com/browse/SERVER-1150)
+-   [SERVER-1245](https://tickets.puppetlabs.com/browse/SERVER-1245)
```
rlinehan: @nathanielksmith might want to chime in here about this and maybe about whether we should note anything about security issues with this because we haven't implemented full support (or something that I am not 100% sure of).

documentation/release_notes.markdown
```diff
@@ -4,116 +4,104 @@ title: "Puppet Server: Release Notes"
...
+Starting with version 2.4.0, Puppet Server [logs](./configuration.markdown#logging) message and warnings when an autosign command generates STDERR output or returns a non-zero exit code. Server 2.4.0 also logs autosigning attempts at the INFO level, rather than DEBUG, to help make autosigning issues easier to diagnose without changing Server's logging level.
 
-* [SERVER-1160](https://tickets.puppet.com/browse/SERVER-1160)
+-   [SERVER-1187](https://tickets.puppetlabs.com/browse/SERVER-1187)
 
-### Known issues
+### Bug fix: Closed memory leak when restarting Server via SIGHUP
 
-#### Modifications to `bootstrap.cfg` might cause problems during upgrades to 2.3.0
+The versions of Trapperkeeper included with Puppet Server 2.3.x leaked a small amount of memory when [restarting Server with a HUP signal](./restarting.markdown). Trapperkeeper 1.4.0, which is included with Puppet Server 2.4.0, resolves this issue.
```
rlinehan: I don't know if we distinguish Trapperkeeper vs. Trapperkeeper Webserver Jetty 9 in these release notes/in our documentation at all, but it was actually the new version of TK-W-J9 (or actually the bump from 1.5.6 to 1.5.9) that fixed this, not the bump in TK itself.

rlinehan:
@cprice404 if you could look this over I would appreciate it.

documentation/release_notes.markdown
cprice404: Agree with @rlinehan , we should just omit this from this round of release notes because we haven't integrated the support for it yet.  We'll include it in the next release.

documentation/release_notes.markdown
underscorgan: Yup, 15.10 is also new in this release.

documentation/release_notes.markdown
cprice404: Agree, would like to get @nathanielksmith or @ahpook to look at this part and see if we want to say anything else, but, may have to wait until after this is merged because @nathanielksmith is out and we probably don't want to block on it.

documentation/release_notes.markdown
cprice404: Yeah, I'd probably just say something like "The versions of the Trapperkeeper components include with...", and then remove the sentence at the end about TK1.4.0.  Presumably if anyone really cares about which component had the issue or what version of the component fixed the issue, they can just click through to the Jira?

documentation/release_notes.markdown
```diff
@@ -4,116 +4,104 @@ title: "Puppet Server: Release Notes"
...
 
-This is a feature release that adds functionality for static catalogs, a new environment classes API, and restarting Puppet Server with a HUP signal.
+-   [TK-360: Error messages returned to client should not include IP or rule blocking the request](https://tickets.puppetlabs.com/browse/TK-360)
 
-> ### New requirements
->
-> Puppet masters running Puppet Server 2.3 depend on [Puppet Agent 1.4.0](https://docs.puppet.com/puppet/4.4/reference/about_agent.html) or newer, which installs [Puppet 4.4](https://docs.puppet.com/puppet/4.4/) and compatible versions of its related tools and dependencies on the server. Puppet agents running older versions of Puppet Agent can connect to Puppet Server 2.3 --- this requirement applies to the Puppet Agent running on the Puppet Server node *only*.
+### New feature: X.509-compliant certificate extensions can match authorization rules
```
cprice404: I would probably move this up to the top of the list, or, put it just after the sections about Ubuntu/systemd, because this is definitely the most compelling thing that lands in this release other than possibly the Ubuntu support.

cprice404:
+1 other than comments I left

documentation/release_notes.markdown
rlinehan: I agree with all this, although I don't think that the tickets in the TK project have versions on them, and I don't know how to add version numbers/I don't think I have permission to add version numbers.

documentation/release_notes.markdown
```diff
@@ -4,116 +4,104 @@ title: "Puppet Server: Release Notes"
...
 
-### Bug fix: Support Ruby master's custom trusted OID mappings
+-   [SERVER-1182](https://tickets.puppetlabs.com/browse/SERVER-1182)
 
-Puppet's Ruby master supports [custom trusted certificate extension object identifier (OID) mappings](https://docs.puppet.com/puppet/latest/reference/config_file_oid_map.html), but previous versions of Puppet Server 2.x did not. This could result in Puppet Server storing trusted extensions with their default numeric OIDs as their keys, instead of the expected custom-mapped short names.
+### New feature: Integrate with systemd services on Debian and Ubuntu
 
-Puppet Server 2.3.2 resolves this issue by supporting the `trusted_oid_mapping_file` and `csr_attributes` settings in [`puppet.conf`](https://docs.puppet.com/puppet/4.4/reference/config_file_main.html), which define paths to YAML files containing the custom mappings and certificate extension request (CSR) attributes, respectively.
+Puppet Server 2.4.0 adds integration with `systemd` on Debian 8 and newer, and Ubuntu 15.04 and newer.
```
camlow325: I think this should say ... "Ubuntu 15.10 and newer" since "15.10" is Wily, which are shipping packages for, and we're not shipping new packages for "15.04", Vivid, anymore.

documentation/release_notes.markdown
cprice404: We could just put it in the comments on the ticket for now.

documentation/release_notes.markdown
```diff
@@ -4,116 +4,104 @@ title: "Puppet Server: Release Notes"
...
 
-For more information about mapping custom trusted OID names, see the Puppet documentation for the  [`custom_trusted_oid_mapping.yaml` file](https://docs.puppet.com/puppet/latest/reference/config_file_oid_map.html).
+-   [EZ-48: Add systemd services for debian 8 and ubuntu >= 15.04](https://tickets.puppetlabs.com/browse/EZ-48)
```
camlow325: Same comment here as above about 15.04 vs. 15.10.

documentation/release_notes.markdown
```diff
@@ -4,116 +4,104 @@ title: "Puppet Server: Release Notes"
...
+-   [TK-372: Jetty JMX Memory Leak in TK WS J9](https://tickets.puppetlabs.com/browse/TK-372)
 
-```
-2016-03-09 11:48:17,460 ERROR [main] [p.t.internal] Error during app buildup!
-java.lang.RuntimeException: Service ':VersionedCodeService' not found
-```
+### Bug fix: Implement DELETE request handling on the `certificate_reqest` endpoint
 
-To resolve this, add this line to your `bootstrap.cfg`:
+Unlike the Ruby Puppet master, previous versions of Puppet Server didn't handle [DELETE requests to the `certificate_request` endpoint](https://docs.puppet.com/puppet/latest/reference/http_api/http_certificate_status.html#delete). Server 2.4.0 resolves this by handling these requests the same way that the Ruby master does.
```
camlow325: Not sure if it's worth mentioning here that one would have to add an auth.conf rule permitting access to the DELETE method for the "certificate_request" endpoint in order to make use of this feature?  Per the rules in the default auth.conf in the Puppet Server package, all DELETE requests made to the "certificate_request" endpoint would be denied.

documentation/release_notes.markdown
```diff
@@ -4,116 +4,104 @@ title: "Puppet Server: Release Notes"
...
 
-If its Logback service is configured to log to syslog, Puppet Server 2.3.0 fails to start. Puppet Server 2.3.1 fixes this regression, which did not affect prior versions of Puppet Server.
+### Other issues
 
-* [SERVER-1215](https://tickets.puppet.com/browse/SERVER-1215)
+-   **Remove hyphens in `puppet-server`:** We've [changed the name of our GitHub repository](https://tickets.puppetlabs.com/browse/SERVER-1206) from `puppet-server` to `puppetserver` and [removed the hyphen from many other references](https://tickets.puppetlabs.com/browse/SERVER-392).
+-   **Log Ruby backtraces ([SERVER-1273](https://tickets.puppetlabs.com/browse/SERVER-1273)):** Previous versions of Server didn't log Ruby backtraces. Server 2.4.0 does, just like a Ruby Puppet master.
+-   **Don't override the service startup timeout ([SERVER-557](https://tickets.puppetlabs.com/browse/SERVER-557)):** Previous versions of Server 2.x overrode the OS-specific service startup timeout with a value of 120 seconds. Server 2.4.0 removes this override.
```
camlow325: Should we also mention what the new timeout would be as a result of removing the override?  EZBake's current default is now 5 minutes.

documentation/release_notes.markdown
haus: Also, systemd support ended up being *just* 16.04. there is follow-up work slated to move jessie and wily to systemd from sysv.

documentation/release_notes.markdown
```diff
@@ -4,116 +4,104 @@ title: "Puppet Server: Release Notes"
...
+Also, see these related tickets:
 
-* [All Puppet Server issues targeted at this release](https://tickets.puppet.com/issues/?jql=project%20%3D%20SERVER%20AND%20fixVersion%20%3D%20%22SERVER%202.3.1%22%20ORDER%20BY%20updated%20DESC%2C%20priority%20DESC%2C%20created%20ASC)
+-   [EZ-72: Implement support for overriding bootstrap-config path in app's project.clj](https://tickets.puppetlabs.com/browse/EZ-72)
 
-## Puppet Server 2.3.0
+### Improvement: Responses to unauthenticated HTTPS requests include less information
 
-Released March 16, 2016.
+When responding to unauthorized HTTPS requests, previous versions of Puppet Server 2.x returned the requester's IP address and [authorization rule][auth.conf] in addition to logging the failed request. Puppet Server 2.4.0 removes this information from the response and directs the responder to consult the server logs for details.
```
rlinehan: It might be worth noting that this is *only* if they have `use-legacy-auth-conf: false` in puppetserver.conf. In the future it will be true whether or not they're using that, once PUP-6150 is done.

documentation/release_notes.markdown
gguillotte: Is there a public ticket for Wily builds, like there was for Xenial?

rlinehan: https://tickets.puppetlabs.com/browse/CPR-323 is the best I could find, but it seems like just a thing we sort of did?

gguillotte: I'll add CPR-323, but if we have a better ticket, I'll swap it out later.

documentation/release_notes.markdown
gguillotte: OK, but in this case that's the title of the ticket in JIRA. I'll just remove the title from the notes.

documentation/release_notes.markdown
gguillotte: I'll omit from the notes going live today and can add it back in once we've confirmed it.

cprice404: Would rather not omit it entirely, as it's the most compelling feature in this release.  I'm OK with this going out with the current wording if we aren't able to get input from the afore-mentioned parties in time.

gguillotte: @cprice404 @rlinehan To be clear, are we talking about the OID shortname mapping part of this section, or the entire section on X.509 extensions?

rlinehan:
👍 I think this looks good to me!

gguillotte:
@KevinCorcoran Because of the server docs are in their own repo, the docs site builds docs for past versions on release tags or commits, so the older notes are all on the docs site. We can try to find a solution for the in-repo docs, but we'd need to find a way to hide them from the docs site.

KevinCorcoran:
@gguillotte Oh, okay.  Well that's not so bad then.  If we cared about keeping them in master/stable after release it seems like we could easily move them somewhere else.  But, I guess if they're all on the docs site, ain't such a big deal.  Thanks for the explanation.

documentation/release_notes.markdown
nathanielksmith: TK-293 doesn't intrinsically pose a security risk; it can just be used in unsafe ways (ie blindly signing certs with certain extensions used for authorizations). I see no problem at all with mentioning it.

PR #1098, opened 2016-05-19
(doc) Update install from packages doc.

gguillotte:
-   Bump version references since 2.4 depends on agent 1.5.
-   Add Wily and Xenial.


rlinehan:
👍

PR #1099, opened 2016-05-20
(SERVER-1247) upgrade to ezbake 0.4.2

cprice404:
This commit upgrades us to the latest version of ezbake, which fixes
a bug in the bootstrap `services.d` packaging on systemd platforms.


justinstoller:
The only thing, AFAICT, travis would test here is that we could still resolve our deps. Which it's already gotten past at this point. No reason to not be on the latest and greatest of ezbake, and it should fix a bug  👍

PR #1100, opened 2016-05-25
(SERVER-1350) Add ubuntu 16.04 to list of PuppetDB supported platforms

rlinehan:
Now that PuppetDB and Puppet Server packages have been released for ubuntu
16.04, we can run PuppetDB tests for that platform.

Bump the PuppetDB version we test against to 4.1.0, since that is what
released Xenial support.


rlinehan:
Ran this the puppetdb test locally against xenial and it passed.

cprice404:
:see_no_evil: +1

PR #1101, opened 2016-05-25
(maint) merge up master to stable may 25 2016

rlinehan:
to bring in the Xenial test helper changes for SERVER-1350 and also some docs changes.

Only conflict was in project.clj around the version number for puppetserver itself - I kept it as `2.5.0-master-SNAPSHOT`.


rlinehan:
Smoke tests for stable that run the puppetdb test passed, so this should hopefully cause no test problems.

PR #1103, opened 2016-06-01
(SERVER-1310) Use standard exception format

justinstoller:
This moves us to use the standard exception format defined in the nogui repo. It also bumps our ring-middleware and tk-auth dependencies as they are also updated to use the new exception format.


KevinCorcoran:
👍

PR #1104, opened 2016-06-02
(SERVER-1375) Adds a pool lock check to pool-lock-supersedes test

camlow325:
This commit adds some logic to the
pool-lock-supersedes-existing-borrows-test to try to wait until
.isLocked() returns true for a requested lock until the test moves on to
validating that borrow requests are held off by the lock.  Without this
change, it was more likely (although still somewhat rare) for the test
to return instances to the pool before the .lock() request was even
made, leading to borrow requests being granted immediately rather than
being held off by the lock.  Test assertions would fail as a result.

### DIFF
--- a/test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
+++ b/test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
@@ -235,6 +235,11 @@
       @blocked-borrow-thread-started?
       @lock-thread-started?
       (is (not (realized? blocked-borrow-thread-borrowed?)))
+      (let [start (System/currentTimeMillis)]
+        (while (and (not (.isLocked pool))
+                    (< (- (System/currentTimeMillis) start) 10000))
+          (Thread/yield)))
+      (is (.isLocked pool))
       (is (not (realized? lock-acquired?)))
 
       (return-instances pool instances)


cprice404:
+1 , but I think we probably need to make this same change in jruby-utils?

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
```diff
@@ -235,6 +235,11 @@
...
       @lock-thread-started?
       (is (not (realized? blocked-borrow-thread-borrowed?)))
+      (let [start (System/currentTimeMillis)]
+        (while (and (not (.isLocked pool))
+                    (< (- (System/currentTimeMillis) start) 10000))
+          (Thread/yield)))
+      (is (.isLocked pool))
       (is (not (realized? lock-acquired?)))
 
       (return-instances pool instances)
```
camlow325: I was able to reproduce the test failure on the following line...

~~~clj
(is (not (realized? blocked-borrow-thread-borrowed?)))
~~~

... by adding a `Thread.sleep(5000)` between the `(deliver lock-thread-started? true)` and `(.lock pool)` lines in the lock-thread future.

I added the `.isLocked pool` check thinking that `.isLocked` is supposed to return true as soon as the request for the lock is granted (even though this can precede when all of the borrowed instances have been checked back into the pool).  With the code above in place, we're much more likely to have gotten past the critical section of code where the requested lock should hold off borrow attempts - which is what this test is primarily about, I think.

I put a 10 second cap on waiting for the lock to be requested because I didn't want to run into the possibility of a wait hang causing the test run to fail mysteriously.  This roughly follows the pattern elsewhere in these tests where we use `timed-deref`s.  Still racy obviously - which isn't great - but I think the above code should help with the test as is.

We had a discussion at one time about changing the name of `.isLocked` to something else because of concerns about whether this is really fulfilling the definition of "locked" in that it returns true between the phases of "caller has been established as the next lock holder" and "currently borrowed instances have been checked back into the pool".  Still open to changing that name at some point but probably outside of the scope of this PR.

camlow325:
@cprice404 - yeah, we should make the same change in `jruby-utils` if others are +1 on it.  Was thinking maybe we could wait until this is merged into master and seems to be passing okay before we put up the same change in `jruby-utils`?

test/unit/puppetlabs/puppetserver/lockable_pool_test.clj
rlinehan: >We had a discussion at one time about changing the name of .isLocked to something else because of concerns about whether this is really fulfilling the definition of "locked" in that it returns true between the phases of "caller has been established as the next lock holder" and "currently borrowed instances have been checked back into the pool". Still open to changing that name at some point but probably outside of the scope of this PR.

In trying to understand these test changes (and having not actually fully read/parsed this comment), I was really confused by this naming. I think I would be +1 on a name change, but agree that it's outside the scope of this PR.

rlinehan:
👍. Agreed that after this goes through master of puppetserver we should bring it into jruby utils, also in favor of (at some point) changing the name of `.isLocked` in jruby utils.

PR #1105, opened 2016-06-08
Add bullet about CA CRL being reloaded from a HUP

npwalker:


npwalker:
@gguillotte @cprice404 This needs to be in the 2.3 docs as well but I couldn't figure out which file to do that in.

cprice404:
@gguillotte let me know if this looks ok to you, and I can create a 2.3 branch if you like.

gguillotte:
:+1: to both this and applying it to a 2.3 branch.

PR #1110, opened 2016-06-10
(PUP-3827) Update tests for changes in puppet error outputs

justinstoller:
This updates our assertion that we are allowed to reach a certain API endpoint to check that we did **not** receive a 403, the proper HTTP response for authentication failure, rather than some other specific non-403 error (defaulting to 200).

This should make the test more robust in the face of changing API error codes, though it could present false positives, if say, our authorization code starts returning a 418 instead of 403 for authorization denied, we won't catch that.

/cc @kurtwall @erikPrime

### DIFF
--- a/acceptance/suites/tests/authorization/default_rules.rb
+++ b/acceptance/suites/tests/authorization/default_rules.rb
@@ -38,15 +38,19 @@ def curl_unauthenticated(path)
   on(master, curl)
 end
 
-def assert_allowed(expected_statuscode = 200)
+def assert_allowed(expected_statuscode = nil)
   assert_no_match(/Forbidden request/, stdout)
-  assert_match(/^STATUSCODE=#{expected_statuscode}$/, stdout)
+  if expected_statuscode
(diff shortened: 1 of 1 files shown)


kurtwall:
+1. I assume removing the regex anchors was intentional.

cprice404:
I *think* this seems ok since it seems super unlikely we'd ever change from a 403 for auth endpoints... but... what if like the SSL handshake failed or something else went wrong before we got the response?  I guess that would still fail in a different way because the regex wouldn't match?  Hmmm... actually... the "assert_no_match" might still pass for that scenario?

So it seems like maybe we'd still need an assertion that "STATUS_CODE" matched... do those assertions support regex group captures?  Would it be possible to do the match against STATUS_CODE and capture the code, and then change the test to make a second assertion about the numeric value?

Or if it's easier we could just change it back to how it was and do the assertion on the 500.  It seems pretty unlikely that we'll be changing the status codes frequently, as doing so should probably technically warrant a major semver bump.

PR #1111, opened 2016-06-10
(PUP-3827) Update default_rules test to search for 500 rather than 400

justinstoller:

PR #1113, opened 2016-06-10
Bump Puppet to latest agent in Davis (743255d)

justinstoller:
Forgot when I tested https://github.com/puppetlabs/puppetserver/pull/1111 that the version we were failing against in PE for the Davis pipeline was not what we are testing in FOSS (so now FOSS is red [here](https://jenkins.puppetlabs.com/job/platform_puppetserver_integration-system_no-conditional_smoke-master/106/LAYOUT=ubuntu1604-64ma-64a,LDAP_TYPE=default,PLATFORM=default,label=beaker/testReport/junit/(root)/acceptance_suites_tests_authorization/default_rules_rb/))

Previously the master branch was pinned to 4.5.0 release of puppet in preparation for Couch hardening. This updates puppetserver#master to what is now promoted into Davis. This includes improvements to indirector REST API errors and the new OID arc for cert extensions.

Still failing to run these locally... (not they're running and failing)

### DIFF
--- a/spec/puppet-server-lib/puppet/jvm/master_spec.rb
+++ b/spec/puppet-server-lib/puppet/jvm/master_spec.rb
@@ -24,7 +24,7 @@ describe 'Puppet::Server::Master' do
     end
 
     it "returns the correct puppet version number" do
-      expect(subject).to eq('4.5.0')
+      expect(subject).to eq('4.6.0')
     end
   end
 


justinstoller:
```
Failures:
  1) Puppet::Server::Master puppet version returns the correct puppet version number
     Failure/Error: expect(subject).to eq('4.5.0')
       expected: "4.5.0"
            got: "4.6.0"
       (compared using ==)
     # ./spec/puppet-server-lib/puppet/jvm/master_spec.rb:27:in `(root)'
     # ./spec/run_specs.rb:5:in `(root)'
     # form-init2136344686080470553.clj:1:in `invoke'
     # main.clj:275:in `invoke'
     # main.clj:280:in `invoke'
     # main.clj:308:in `invoke'
     # main.clj:343:in `invoke'
     # main.clj:421:in `doInvoke'
```

I had no idea that was a thing

~the more you know

cprice404:
@justinstoller so we should hold off on this until you add a commit that patches that test?

justinstoller:
Yes, I should definitely figure out what's going on here.

Sorry that was a rather glib response to finally getting these results right before the show and tell.

cprice404:
@justinstoller I think you just need to change the hard-coded string in that test.  Which will raise the usual string of questions about whether that test is valuable.  And then we'll probably all shrug and say that it's too much work to definitively answer that question for now, and it's easier to just change it and think about the long term strategy independently... and then we will forget to do that until the next time the test fails :)

cprice404:
+1

KevinCorcoran:
👍 

~travisci

Let's see how Jankins rolls with this.

PR #1114, opened 2016-06-11
(SERVER-1380) Validate that the CRL is reloaded after HUPing the server

kurtwall:
After revoking a certificate, HUP the server and see if the revoked certificate is accepted. It shouldn't be.

### DIFF
--- a/acceptance/suites/tests/authorization/hup_reloads_crl.rb
+++ b/acceptance/suites/tests/authorization/hup_reloads_crl.rb
@@ -54,7 +54,7 @@ step 'Revoke the agent certs' do
   end
 end
 
-step 'Demonstrate that the certs work before HUPping the server' do
+step 'Demonstrate that the certs work before HUPing the server' do
   agents.each do |a| 
     if not_controller(a)
       rc = on(a, puppet('agent', '--test', '--server', "#{server}"),
--- a/acceptance/suites/tests/authorization/hup_reloads_crl.rb
+++ b/acceptance/suites/tests/authorization/hup_reloads_crl.rb
@@ -1,0 +1,69 @@
+test_name '(SERVER-1380) CRL should reload after a server HUP' do
+
+server = master.puppet['certname']
+
+teardown do
+  # Leave agents with working certs. Destroy the revoked cert on the master;
+  # remove all traces on the agent; make the agent request a shiny new cert; 
+  # have the master sign it; then use it to make sure it works.
+  agents.each do |a|
+    if not_controller(a)
+      hostcert_file = a.puppet['hostcert']
+      reqdir = a.puppet['requestdir']
+      nodefqdn = a.hostname
+      request_file = File.join(reqdir, nodefqdn + ".pem")
+
+      # FIXME: this is a dirty nasty hack that Beaker should handle seamlessly
+      on(master, puppet('cert', 'destroy', "#{nodefqdn}"),
+         {:accept_all_exit_codes => true})
+      on(a, "rm -fv #{hostcert_file} #{request_file}",
+         {:accept_all_exit_codes => true})
+      on(a, puppet('agent', '--test', '--server', "#{server}"),
+         {:accept_all_exit_codes => true})
+      on(master, puppet('cert', 'sign', "#{nodefqdn}"),
+         {:accept_all_exit_codes => true})
+      on(a, puppet('agent', '--test', '--server', "#{server}"),
+         {:accept_all_exit_codes => true})
+    end
+  end
+end
+
+step 'Demonstrate that the existing certs work' do
+  agents.each do |a| 
+    if not_controller(a)
+      rc = on(a, puppet('agent', '--test', '--server', "#{server}"),
+              {:acceptable_exit_codes => [0,2]})
+      fail_test('Missing expected output') unless \
+        rc.stdout.include?('Applied catalog')
+    end
+  end
+end
+
+step 'Revoke the agent certs' do
+  agents.each do |a| 
+    if not_controller(a)
+      rc = on(master, puppet('cert', 'revoke', "#{a.hostname}"),
+              {:acceptable_exit_codes => [0,2]})
+      fail_test('Missing expected output') unless \
+        rc.stdout.include?('Revoked certificate')
+    end
+  end
+end
+
+step 'HUP the server' do
+  hup_server
+end
+
+step 'Demonstrate that the certs are rejected' do
+  agents.each do |a| 
+    if not_controller(a)
+      rc = on(a, puppet('agent', '--test', '--server', "#{server}"),
+              {:acceptable_exit_codes => [1]})
+      fail_test('Missing expected output') unless \
+        rc.stderr.include?('SSL_connect SYSCALL returned=5')
+
+    end
+  end
+end
+
+end
--- a/acceptance/suites/tests/authorization/hup_reloads_crl.rb
+++ b/acceptance/suites/tests/authorization/hup_reloads_crl.rb
@@ -1,4 +1,9 @@
-test_name '(SERVER-1380) CRL should reload after a server HUP' do
+test_name '(SERVER-1380) Server reloads the CRL after being HUPped' do
+# Note the skip_test statement. This test requires at least one agent that is
+# not also a master.
+
+skip_test 'This test requires a non-master agent' unless \
+  hosts_as('agent').count > hosts_as('master').count
 
 server = master.puppet['certname']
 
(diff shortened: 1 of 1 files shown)


acceptance/suites/tests/authorization/hup_reloads_crl.rb
```diff
@@ -0,0 +1,62 @@
...
+      rc = on(a, puppet('agent', '--test', '--server', "#{server}"),
+              {:acceptable_exit_codes => [0,2]})
+    end
+  end
+end
+
+step 'Revoke the agent certs' do
+  agents.each do |a| 
+    if not_controller(a)
+      agent_fqdn = fact_on(a, 'fqdn')
```
erikPrime: This shouldn't be required.  agent.hostname should return the agent's hostname.

acceptance/suites/tests/authorization/hup_reloads_crl.rb
```diff
@@ -0,0 +1,62 @@
...
+      rc = on(master, puppet('cert', 'revoke', "#{agent_fqdn}"),
```
erikPrime: rc = isn't used.

acceptance/suites/tests/authorization/hup_reloads_crl.rb
```diff
@@ -0,0 +1,62 @@
...
+              {:acceptable_exit_codes => [0,2]})
+    end
+  end
+end
+
+step 'HUP the server' do
```
erikPrime: Prior to hupping the server, we should demonstrate that puppet cert revoke <certname> has not yet taken effect.

acceptance/suites/tests/authorization/hup_reloads_crl.rb
```diff
@@ -0,0 +1,62 @@
...
+  hup_server
+end
+
+step 'Demonstrate that the certs are rejected' do
+  agents.each do |a| 
+    if not_controller(a)
+      rc = on(a, puppet('agent', '--test', '--server', "#{server}"),
```
erikPrime: rc isn't used

acceptance/suites/tests/authorization/hup_reloads_crl.rb
```diff
@@ -0,0 +1,62 @@
...
+              {:accept_all_exit_codes => true})
```
erikPrime: We should be looking for a specific exit code (unless we are going to validate rc in the future?)

acceptance/suites/tests/authorization/hup_reloads_crl.rb
```diff
@@ -0,0 +1,62 @@
...
+    if not_controller(a)
+      rc = on(a, puppet('agent', '--test', '--server', "#{server}"),
+              {:acceptable_exit_codes => [0,2]})
+    end
+  end
+end
+
+step 'Revoke the agent certs' do
+  agents.each do |a| 
+    if not_controller(a)
```
erikPrime: This may cause unexpected behavior in cases where there is only one test system, a host that is both a master and an agent.  In the case where we have a only one test system, I think we can successfully run the test by removing the if block... But we should test and see.

If running this on a single host that has both master and agent beaker roles isn't successful, we probably need to skip this test in the case where there are no agent only hosts available to the test.

acceptance/suites/tests/authorization/hup_reloads_crl.rb
kurtwall: What I need to demonstrate is that HUPping the server reloads the CRL. I don't need to validate that a revoked cert doesn't take effect until after a HUP.

acceptance/suites/tests/authorization/hup_reloads_crl.rb
kurtwall: Fixed wherever I used this pattern.

acceptance/suites/tests/authorization/hup_reloads_crl.rb
kurtwall: Thx for reminding me. Fixed.

acceptance/suites/tests/authorization/hup_reloads_crl.rb
kurtwall: Precisely. But I forgot. You reminded me. :)

acceptance/suites/tests/authorization/hup_reloads_crl.rb
kurtwall: Investigating.

acceptance/suites/tests/authorization/hup_reloads_crl.rb
kurtwall: Yes to both.

acceptance/suites/tests/authorization/hup_reloads_crl.rb
erikPrime: There might be a point in the future in which the puppetserver dynamically reloads the CRL.  It MIGHT be nice if this test fails when we get there so that we can refactor this test accordingly.

On the other hand, this this feature might be so far into the future that the effort of adding a few more lines just isn't worth it.

acceptance/suites/tests/authorization/hup_reloads_crl.rb
kurtwall: Notes in the associated ticket. Short version: with a single combined master and agent, test passes; with a single master and no agents at all, test passes.

acceptance/suites/tests/authorization/hup_reloads_crl.rb
kurtwall: Cross that bridge when we get to it?

erikPrime: Definitely yes.

acceptance/suites/tests/authorization/hup_reloads_crl.rb
erikPrime: My previous objection above doesn't really apply, because there are no cases in our CI pipelines that I can find where we are running with a master only test environment.

acceptance/suites/tests/authorization/hup_reloads_crl.rb
```diff
@@ -0,0 +1,68 @@
...
+  # remove all traces on the agent; make the agent request a shiny new cert; 
+  # have the master sign it; then use it to make sure it works.
+  agents.each do |a|
+    if not_controller(a)
+      hostcert_file = a.puppet['hostcert']
+      reqdir = a.puppet['requestdir']
+      nodefqdn = a.hostname
+      request_file = File.join(reqdir, nodefqdn + ".pem")
+
+      # FIXME: this is a dirty nasty hack that Beaker should handle seamlessly
```
rlinehan: what's it a hack for?

rlinehan: ah, I see, it's a hack for "detroy the revoked cert on the master, remove all traces on the agent, make the agent request a shiny new cert, have the master sign it, then use it to make sure it works"... so I assume you want a helper that actually does this in Beaker? is there a ticket for that? will this ever actually get fixed?

acceptance/suites/tests/authorization/hup_reloads_crl.rb
rlinehan: > What I need to demonstrate is that HUPping the server reloads the CRL. I don't need to validate that a revoked cert doesn't take effect until after a HUP.

I think I might agree with @erikPrime, that it would be good to verify that the revoke hasn't taken affect, in case for some reason this behavior changes without us realizing? I dunno, I guess it's not necessary, it just feels like that would be a test that more accurately demonstrates the scenario?

kurtwall: I think you're both right. I'll push up a change that does the intermediate check.

acceptance/suites/tests/authorization/hup_reloads_crl.rb
```diff
@@ -0,0 +1,68 @@
+test_name '(SERVER-1380) CRL should reload after a server HUP' do
```
rlinehan: It seems like all the steps in this test will only run if there's an agent (not on the master) in addition to a master. Would it be possible/worthwhile to add something to the top of this file saying "skip this test unless there's an additional agent"? It seems like having it be explicit that that's the only scenario in which this test will run might good.

rlinehan:
Left a couple questions, but overall looks good.

acceptance/suites/tests/authorization/hup_reloads_crl.rb
kurtwall: It was mostly a whing that I had to write the code myself. I think I added this comment at the moment I realized I'd spent more time writing the `teardown` block than I spent writing then rest of the test. 

>

acceptance/suites/tests/authorization/hup_reloads_crl.rb
kurtwall: I added a `skip_test` statement at the top of the test to accommodate this. I did write the test to run and "pass" in the no-non-master-agent case, but perhaps skipping it entirely is the smarter approach.

rlinehan:
👍

rlinehan:
Travis won't test anything in this, so merging.

PR #1116, opened 2016-06-13
(maint) Bump version of Puppet Agent used in Acceptance tests

justinstoller:
Set version of puppet-agent (db533e0) in acceptance tests to one
that includes the puppet from our dev env (743255d).


erikPrime:
Looks good to me.

PR #1117, opened 2016-06-14
(maint) Update to new version of puppet-agent

justinstoller:
I don't even... This is a few builds farther ahead than the last one (so should have similar content) but I **think** this link confirms that it has been promoted into PE and has been built against all of our desired platforms: http://agent-downloads.delivery.puppetlabs.net/2016.3/puppet-agent/1.5.1.123/repos/  Though at this point of frustration I wouldn't trust anything I say.


erikPrime:
This looks good to me in terms of syntax.  I'm merging without additional review because it's late in the day, because there isn't any one else online to review, and because CI is currently broken, and this isn't going to break it any worse than it already is.

PR #1112, opened 2016-06-10
(SERVER-1320)Validates that CSRs with CA:TRUE are rejected for signing.

erikPrime:
… for signing

PR #1118, opened 2016-06-14
(SERVER-1163) Remove jruby code and tests

jpinsonault:
This commit removes all of the code that was extracted into the jruby-utils
library, as well as tests for that code.

puppetserver code was then tied back up to the new jruby-utils code. This is
where the majority of the effort has gone, since many small things changed.
Most of the changes are in the jruby-puppet-service and related code.

Some important changes are:
* The jruby pool is no longer puppetserver specific, so the code has been made
  more general to accommodate this
* The work of initializing the jruby-pool is split between puppetserver and
  jruby-utils. There are now lifecycle functions passed to the
  pool-manager-service that are run automatically to finish the initialization.
  These setup the jruby scripting container, the pool instances, as well as
  handle cleanup and shutdown.
* Calls to jruby-puppet namespaces were replaces with jruby-utils calls
* A new pool-manager service from jruby-utils is used to create the
  pool-context in the jruby-puppet-service
* the with-jruby-puppet macro has been replaced by a more general macro from

====

This PR relies on an unreleased version of jruby-utils, so it can't be merged yet.

About half the changes in this PR are removal of code that's duplicated in jruby-utils. Either way, good luck!


src/clj/puppetlabs/services/jruby/jruby_puppet_agents.clj
```diff
@@ -1,217 +0,0 @@
-(ns puppetlabs.services.jruby.jruby-puppet-agents
```
rlinehan: 🎆 goodbye and good riddance

jpinsonault:
travis is failing because  of the jruby-utils dependency

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -61,117 +62,91 @@
...
-(schema/defn lock-acquired :- jruby-schemas/JRubyLockAcquiredEvent
-  [event-callbacks :- [IFn]
-   reason :- jruby-schemas/JRubyEventReason]
-  (notify-event-listeners event-callbacks (create-lock-acquired-event reason)))
-
-(schema/defn lock-released :- jruby-schemas/JRubyLockReleasedEvent
-  [event-callbacks :- [IFn]
-   reason :- jruby-schemas/JRubyEventReason]
-  (notify-event-listeners event-callbacks (create-lock-released-event reason)))
+add-facter-jar-to-system-classloader
```
rlinehan: I think this needs to be indented slightly (2 spaces)? intellij keeps doing this to me and I hate it and haven't figured out the right thing to make it stop.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -61,117 +62,91 @@
...
+                            JRubyPuppet)]
+          (-> jruby-instance
+              (assoc :jruby-puppet jruby-puppet)
+              (assoc :environment-registry env-registry)))))))
+
+(schema/defn initialize-scripting-container-fn
+  [scripting-container
+   config :- jruby-schemas/JRubyConfig]
+  (doto scripting-container
+    (.setEnvironment (jruby-internal/managed-environment (jruby-internal/get-system-env) (:gem-home config)))
```
rlinehan: `jruby-internal` here is still aliased to `jruby-puppet-internal`, but these functions are available in jruby-utils - I think you could use the ones from jruby-utils and delete these from this repo.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
jpinsonault: huh, weird, this is the first time for me

rlinehan: maybe I did it by accident when changing some stuff? if so, sorry!

jpinsonault: nah, my intellij is doing it too, and messing with the indentation settings doesn't seem to fix it

test/integration/puppetlabs/services/jruby/class_info_test.clj
```diff
@@ -63,246 +70,253 @@
...
+                  (jruby-testutils/jruby-puppet-config
                    {:master-code-dir (.getAbsolutePath code-dir)
-                    :master-conf-dir (.getAbsolutePath conf-dir)})
-          instance (jruby-internal/create-pool-instance!
-                     pool 0 config #() nil)
-          jruby-puppet (:jruby-puppet instance)
-          container (:scripting-container instance)
-          env-registry (:environment-registry instance)
+                    :master-conf-dir (.getAbsolutePath conf-dir)}))]
+      ;(clojure.pprint/pprint config)
```
rlinehan: Extra `pprint`.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -159,180 +101,9 @@
...
-          (try
-            (flush-instance-fn pool instance)
-            (finally
-              (.releaseItem pool instance false))))
-        (.releaseItem pool instance)))
-    ;; if we get here, it was from a borrow and we got a Retry, so we just
-    ;; return it to the pool.
-    (.releaseItem (:pool instance) instance)))
-
 (schema/defn ^:always-validate new-main :- jruby-schemas/JRubyMain
```
rlinehan: I think you should be able to use the version of this in jruby-utils (and then you can delete this and the schemas that this needs).

rlinehan: or actually, I'm not sure this is used anywhere anymore?

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -112,39 +87,6 @@
...
-    ;; As of JRuby 1.7.20 (and the associated 'jruby-openssl' it pulls in),
-    ;; we need to explicitly require 'jar-dependencies' so that it is used
-    ;; to manage jar loading.  We do this so that we can instruct
-    ;; 'jar-dependencies' to not actually load any jars.  See the environment
-    ;; variable configuration in 'init-jruby-config' for more
-    ;; information.
-    (.runScriptlet "require 'jar-dependencies'")
-    (.runScriptlet "require 'puppet/server/master'")))
-
 (schema/defn ^:always-validate config->puppet-config :- HashMap
```
rlinehan: I think this might end up being the only function still needed in this namespace, in which case maybe just move it into `jruby-puppet-core`?

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -61,117 +62,91 @@
        (max 1)
        (min 4)))
 
-(schema/defn ^:always-validate
```
rlinehan: I don't think you should need `default-pool-size` above this anymore (there's one in jruby-utils).

rlinehan:
I think there might be some test fixtures left in `dev-resources/puppetlabs/services/jruby/` that can probably be deleted now too.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -192,140 +167,25 @@
       (update-in [:master-run-dir] #(or % default-master-run-dir))
       (update-in [:master-log-dir] #(or % default-master-log-dir))
       (update-in [:max-active-instances] #(or % (default-pool-size (ks/num-cpus))))
-      (update-in [:max-requests-per-instance] #(or % 0))
+      (update-in [:max-borrows-per-instance] #(or % 0))
```
rlinehan: We want to keep this setting in puppet server's config as `max-requests-per-instance` (so that we don't have any breaking changes) - so we're at some point going to need to take it in as `max-requests-per-instance` but then associate the value to the config as `max-borrows-per-instance`.

src/clj/puppetlabs/services/config/puppet_server_config_core.clj
```diff
@@ -65,9 +66,8 @@
   (->> (keyword->setting k)
        (.getSetting jruby-puppet)))
 
-(defn get-puppet-config*
+(schema/defn get-puppet-config*
   [jruby-puppet]
```
rlinehan: this argument should probably be updated to use schema then?

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -192,140 +167,25 @@
       (update-in [:master-run-dir] #(or % default-master-run-dir))
       (update-in [:master-log-dir] #(or % default-master-log-dir))
       (update-in [:max-active-instances] #(or % (default-pool-size (ks/num-cpus))))
```
rlinehan: I think we want to take out the calls to provide default for `compile-mode`, `borrow-timeout`, and `max-active-instances`, and `max-borrows-per-instance` - AFAIK we're using the defaults in jruby-utils and not currently providing our own.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -61,117 +62,91 @@
...
-;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
-;;; Public
+(schema/defn extract-puppet-config
+  [config :- {schema/Keyword schema/Any}]
+  (into {} (for [[key _] jruby-puppet-schemas/JRubyPuppetConfig]
+             {key (get config key)})))
 
 (schema/defn ^:always-validate
-  initialize-config :- jruby-schemas/JRubyPuppetConfig
+  initialize-puppet-config :- jruby-puppet-schemas/JRubyPuppetConfig
```
rlinehan: Right now the `JRubyPuppetConfig` schema has a bunch of overlap with the `JRubyConfig` in jruby-utils. In my opinion, at least, it would be better for this schema to only have the puppetserver-specific settings, and for this function to only initialize defaults for those settings. That might mean that we also want to do the extracting of jruby options and puppet options before we call this, so that from the start we have two separate maps? Not entirely sure about that, but kinda leaning towards the idea that that would make this a bit cleaner.

rlinehan:
I haven't looked at all the tests yet, but everything else looks 📈. I left a couple comments:

1. I think that there's only one function in jruby-puppet-internal that we still need, and so maybe we can just delete that namespace and add that function to jruby-puppet-core.
2. I think that there's a little more cleanup we could do with the config.
3. We need to munge `max-requests-per-instance` to `max-borrows-per-instance` (so that we're reading `max-requests-per-instance` out of the config, but passing `max-borrows-per-instance` to `create-pool` and the CLI functions).
4. I think we can delete some test fixtures.

Nice job!

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
jpinsonault: It's not used, it is now dead

test/unit/puppetlabs/services/master/master_core_test.clj
```diff
@@ -84,148 +87,153 @@
...
-                      "/two/file" (map-with-classes [])})
+    (with-redefs [jruby-core/borrow-from-pool-with-timeout (fn [_ _ _] {:jruby-puppet (Object.)})
+                  jruby-core/return-to-pool (fn [_ _ _] #())]
+      (let [jruby-service (reify jruby/JRubyPuppetService
+                            #_(borrow-instance [_ _]
+                              {:jruby-puppet (Object.)})
+                            #_(return-instance [_ _ _])
+                            (get-pool-context [_] (jruby-core/create-pool-context
+                                                   (jruby-core/initialize-config {:gem-home "bar"
+                                                                                  :ruby-load-path ["foo"]})))
```
rlinehan: Oops, I think this is my fault, but I think that the commented out `borrow-instance` and `return-instance` should be deleted, and also that `get-pool-context` here might be able to be deleted too.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -61,117 +62,91 @@
...
+                            "new"
+                            (into-array Object
+                                        [puppet-config puppetserver-config])
+                            JRubyPuppet)]
+          (-> jruby-instance
+              (assoc :jruby-puppet jruby-puppet)
+              (assoc :environment-registry env-registry)))))))
+
+(schema/defn initialize-scripting-container-fn
+  [scripting-container
```
camlow325: Should be able to apply a schema type to this, I think.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
camlow325: Probably some other stuff that could go away along with this now, too, like `init-jruby-config` and `SupportedJRubyCompileModes`.

test/integration/puppetlabs/services/jruby/jruby_puppet_internal_test.clj
```diff
@@ -1,60 +0,0 @@
...
-      (is (instance? RubyInstanceConfig$CompileMode
-                     (jruby-internal/get-compile-mode mode)))))
-  (testing "throws an exception if mode is nil"
-    (is (thrown? ExceptionInfo
-                 (jruby-internal/get-compile-mode nil))))
-  (testing "throws an exception for values not in enum"
-    (is (thrown? ExceptionInfo
-                 (jruby-internal/get-compile-mode :foo)))))
-
-(deftest ^:integration settings-plumbed-into-jruby-container
```
camlow325: I think we should keep at least the http-client settings portion of this test, in some form.  Not sure we have any coverage elsewhere for confirming that the Puppet Server-specific http-client settings are passed from a higher level down into the Ruby-level HttpClient.settings.

test/unit/puppetlabs/services/jruby/jruby_interpreter_test.clj
```diff
@@ -1,64 +0,0 @@
...
-            [puppetlabs.puppetserver.testutils :as testutils]
-            [puppetlabs.kitchensink.core :as ks]))
-
-(use-fixtures :once
-              (testutils/with-puppet-conf
-                "./dev-resources/puppetlabs/services/jruby/jruby_interpreter_test/puppet.conf"))
-
-(deftest create-jruby-instance-test
-
-  (testing "Var dir is not required (it will be read from puppet.conf)"
```
camlow325: Are the tests in this block which validate settings being honored from puppet.conf vs. from the tk-config which are being removed still covered elsewhere?

test/unit/puppetlabs/services/jruby/jruby_testutils.clj
```diff
@@ -29,6 +28,26 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; JRubyPuppet Test util functions
 
+(schema/defn ^:always-validate
+create-mock-pool-instance :- JRubyInstance
```
camlow325: We might consider changing the use of this to mock the `get-initialize-pool-instance-fn` and `initialize-scripting-container-fn` instead of `create-pool-instance` in jruby-utils so that the mocking is at least limited to functions defined within the immediate library being tested.

test/integration/puppetlabs/services/jruby/jruby_puppet_internal_test.clj
jpinsonault: I think @rlinehan and I decided it was kind of redundant, since tests like `general-puppet-int-test` tickle the code paths that would rely on those settings having been plumbed through (right?), but I think it'd be reasonable to add an integration test that tests this specifically

test/unit/puppetlabs/services/jruby/jruby_testutils.clj
jpinsonault: funny you should mention. I initially did mock `get-initialize-pool-instance-fn` instead of `create-pool-instance`, and it led us down a rabbit hole for 2 days and took 3 developers to figure out why a test was failing. In the test `class-info-updated-after-cache-flush-during-prior-request`, there is a timeout of 10 seconds, which would cause the test to fail, but only when run with a lot of other tests. It would never fail when run by itself. Chris hypothesized it must have something to do with garbage collection, but we never really figured out why. Either way, after that we were so scarred we decided to just fix it the quickest way possible

test/unit/puppetlabs/services/jruby/jruby_interpreter_test.clj
jpinsonault: I don't think so. I'll add a test back in

test/unit/puppetlabs/services/jruby/jruby_testutils.clj
rlinehan: To expand on this a bit, I think it was the timed deref in `class-info-updated-after-cache-flush-during-prior-request` that was failing (https://github.com/jpinsonault/puppetserver/blob/3d23836dcdf8126300b2f2e9016e4a21c646d052/test/integration/puppetlabs/services/master/environment_classes_int_test.clj#L691) - if we bumped it from 1000 ms to 2000 ms it more reliably passed. We hyphothesized that the reason it was taking longer/timing out was because if we just mocked out the `get-initialize-pool-instance-fn` then the instance was still being created, just not loading puppet code, but the instance being created was enough to cause the deref to time out.

camlow325: Huh, weird.  If you've already pored that much time into this, I definitely don't think it's worth pursuing further.  Carry on!

camlow325: The part about it failing more often when run with other tests makes me wonder about whether some of the other tests aren't cleaning up fully before they exit in some cases.  I've definitely seen some weird things (that I can't really explain) with `with-redefs` which affect functions that run on different threads when the lifetime of code running on those threads spans thread boundaries.

I'm good with how this is implemented on this PR, though, so not suggesting we spend any more time diagnosing that at this point.

test/integration/puppetlabs/services/jruby/jruby_puppet_internal_test.clj
rlinehan: I think I might have overzealously deleted this test at some point... I think it would be sufficient to just add back this test (minus the compile-mode test) in a different namespace.

test/unit/puppetlabs/services/jruby/jruby_pool_test.clj
```diff
@@ -1,211 +0,0 @@
...
-            [puppetlabs.services.jruby.jruby-puppet-core :as jruby-core]
-            [puppetlabs.services.jruby.jruby-puppet-internal :as jruby-internal]
-            [puppetlabs.trapperkeeper.testutils.logging :as logutils]))
-
-(use-fixtures :each jruby-testutils/mock-pool-instance-fixture)
-
-;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
-;; Tests
-
-(deftest configuration-validation
```
rlinehan: It might be worthwhile to have a test that the config stuff we've got now/that you are refactoring properly inserts the right defaults or errors when not passed some of these things.

test/unit/puppetlabs/services/jruby/jruby_puppet_agents_test.clj
```diff
@@ -1,134 +0,0 @@
...
-                       jruby-testutils/default-shutdown-fn)]
-    (testing "next instance id should be based on the pool size"
-      (is (= 10 (jruby-agents/next-instance-id 2 pool-context)))
-      (is (= 100 (jruby-agents/next-instance-id 92 pool-context))))
-    (testing "next instance id should wrap after max int"
-      (let [id (- Integer/MAX_VALUE 1)]
-        (is (= (mod id 8) (jruby-agents/next-instance-id id pool-context)))))))
-
-(deftest master-termination-test
-  (testing "Flushing the pool causes masters to be terminated"
```
rlinehan: I wonder if it might still be worthwhile to have a test like this, that validates that the cleanup function we've provided actually works?

test/unit/puppetlabs/services/jruby/jruby_testutils.clj
```diff
@@ -180,6 +130,7 @@
                          tk-service/service-context
                          :pool-context)
         num-jrubies (-> pool-context
+                        :internal
```
rlinehan: It probably would be good to change this function to do

```clojure
(-> pool-context
     jruby-core/get-pool-state
    :size)
```

since the `:internal` key is supposed to be private,so that in case we ever change where they pool state is kept, we won't have to change this here.

test/unit/puppetlabs/services/jruby/jruby_testutils.clj
jpinsonault: During debugging this, we also thought there must be some tests not cleaning up after themselves or dirtying things up in some other way, so we did a little binary search, combining different tests with the failing one. After getting it down to 6 or 7 test namespaces, removing any of them would cause it to stop failing. Even then, it wouldn't fail every time. That makes me suspect it wasn't some dirty side effects from other tests specifically, but just from the amount of them

test/unit/puppetlabs/services/jruby/jruby_testutils.clj
jpinsonault: ha, yes that was the point wasn't it =)

test/unit/puppetlabs/services/master/master_core_test.clj
jpinsonault: I get failures when I remove it, something tries to reify it and errors

jpinsonault:
I think I've addressed all the feedback but the creation/restoration of tests. The latest commit:

* misc fixes, removal of jruby-puppet code
* restore max-requests-per-instance
* add schemas to various functions
* remove some overlap in initialization of configs

jpinsonault:
Need to remove the jruby-puppet-internal namespace too

documentation/config_file_puppetserver.markdown
```diff
@@ -100,7 +100,7 @@ jruby-puppet: {
     master-run-dir: /var/run/puppetlabs/puppetserver
     master-log-dir: /var/log/puppetlabs/puppetserver
     max-active-instances: 1
-    max-requests-per-instance: 0
+    max-borrows-per-instance: 0
```
rlinehan: I think these docs should be changed back to say `max-requests`.

src/clj/puppetlabs/services/jruby/jruby_puppet_internal.clj
```diff
@@ -1,45 +0,0 @@
-(ns puppetlabs.services.jruby.jruby-puppet-internal
```
rlinehan: goddamn it I fail at gifs.

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -79,7 +78,72 @@
...
+   :master-conf-dir (schema/maybe schema/Str)
+   :master-code-dir (schema/maybe schema/Str)
+   :master-var-dir (schema/maybe schema/Str)
+   :master-run-dir (schema/maybe schema/Str)
+   :master-log-dir (schema/maybe schema/Str)
+   :http-client-ssl-protocols [schema/Str]
+   :http-client-cipher-suites [schema/Str]
+   :http-client-connect-timeout-milliseconds schema/Int
+   :http-client-idle-timeout-milliseconds schema/Int
+   :max-requests-per-instance schema/Int
```
rlinehan: I think that this still has some JRuby config settings (`ruby-load-path`, `gem-home`, `max-requests-per-instance`) that it shouldn't have? although I guess `ruby-load-path` and `gem-home` are never initialized/given defaults... still, I think that those should all not be part of this?

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -14,7 +13,7 @@
   "Schema defining the supported values for the JRuby CompileMode setting."
   (apply schema/enum supported-jruby-compile-modes))
 
-(def JRubyPuppetConfig
+(def CombinedJRubyPuppetConfig
```
rlinehan: I think that this could be defined as `(merge JRubyPuppetConfig jruby-schemas/JRubyConfig)`.

jpinsonault: I think that would cause it have both `:max-borrows-per-instance` and `:max-requests-per-instance`

rlinehan: You could `dissoc` one of them? Or maybe `JRubyPuppetConfig` shouldn't have either of them?

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -110,68 +81,70 @@
...
+   :master-run-dir (schema/maybe schema/Str)
+   :master-log-dir (schema/maybe schema/Str)
+   :http-client-ssl-protocols [schema/Str]
+   :http-client-cipher-suites [schema/Str]
+   :http-client-connect-timeout-milliseconds schema/Int
+   :http-client-idle-timeout-milliseconds schema/Int
+   :max-requests-per-instance schema/Int
+   :use-legacy-auth-conf schema/Bool})
 
 (defn jruby-main-instance?
```
rlinehan: I think this is unused.

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -181,125 +154,6 @@
   [x]
   (instance? Main$Status x))
 
-(defn jruby-scripting-container?
-  [x]
-  (instance? ScriptingContainer x))
-
 (defn jruby-instance-config?
```
rlinehan: I don't think this is used anymore.

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -1,48 +1,19 @@
...
-  ;; A sentinel object to put into a pool when we are shutting down.
-  ;; This can be used to build `borrow` functionality that will detect the
-  ;; case where we're trying to borrow during a shutdown, so we can return
-  ;; a sane error code.
-  [pool])
-
 (def supported-jruby-compile-modes
   #{:jit :force :off})
 
 (def SupportedJRubyCompileModes
```
rlinehan: I don't think this or `supported-jruby-compile-modes` should be here... I think if necessary, the `CombinedJRubyPuppetConfig` could use this from `jruby-utils`, but my hope is we can avoid that by merging the configs together rather than enumerating all of the settings.

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
jpinsonault: I think we might need to keep `max-requests-per-instance` in there because we output the initialized jruby-puppet config before extracting `max-requests-per-instance` out and putting it into the jruby-config

jpinsonault: Not sure why I left in the other settings. I think some intermediate frankenstein step I had need them at one point. Don't need them anymore though

jpinsonault:
Pushed some more commits around restoring tests and cleaning things up. I know we'd still like to create a unit test for `initialize-configs-and-create-pool-context`, but I figured these commits could be examined in the mean time

test/unit/puppetlabs/services/jruby/jruby_puppet_agents_test.clj
jpinsonault: Yes, definitely

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -146,14 +58,13 @@
...
-
-(defn jruby-main-status-instance?
-  [x]
-  (instance? Main$Status x))
-
-(defn jruby-instance-config?
-  [x]
-  (instance? RubyInstanceConfig x))
+(def CombinedJRubyPuppetConfig
+  "Schema combining JRubyPuppetConfig and JrubyConfig.
```
rlinehan: Nitpick: s/JrubyConfig/JRubyConfig/.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -206,6 +206,39 @@
                                                                             (:ruby-load-path jruby-config)))]
     (jruby-core/initialize-config (assoc modified-jruby-config :lifecycle lifecycle-fns))))
 
+(schema/defn initialize-configs-and-create-pool-context
+  "Handles initializing the jruby and jruby-puppet configs, and then uses them
+  to create a pool-context"
```
rlinehan: This looks great! I think the name and this docstring should be changed though, because the function isn't creating the pool context anymore.

src/clj/puppetlabs/puppetserver/cli/gem.clj
```diff
@@ -1,10 +1,12 @@
 (ns puppetlabs.puppetserver.cli.gem
   (:require [puppetlabs.puppetserver.cli.subcommand :as cli]
-            [puppetlabs.services.jruby.jruby-puppet-core :as jruby-core]))
+            [puppetlabs.services.jruby.jruby-puppet-core :as jruby-puppet-core]
+            [puppetlabs.services.jruby.jruby-core :as jruby-core]))
 
 (defn gem-run!
   [config args]
-  (jruby-core/cli-run! config "gem" args))
+  (let [jruby-config (jruby-core/initialize-config (jruby-puppet-core/extract-jruby-config (:jruby-puppet config)))]
```
rlinehan: This should now be able to be changed to use the new `initialize-configs` function.

dev/bootstrap.cfg
```diff
@@ -1,5 +1,6 @@
 puppetlabs.services.request-handler.request-handler-service/request-handler-service
 puppetlabs.services.jruby.jruby-puppet-service/jruby-puppet-pooled-service
+puppetlabs.services.jruby.jruby-pool-manager-service/jruby-pool-manager-service
```
rlinehan: I think this also needs to be added to `dev/user_repl.clj`.

rlinehan:
I think some Java files related to the pool can be deleted too (everything under `src/java/com/puppetlabs/puppetserver/pool` maybe?)

jpinsonault:
I pushed some commits. I believe that they cover all the remaining feedback

src/clj/puppetlabs/puppetserver/cli/gem.clj
```diff
@@ -5,7 +5,7 @@
 
 (defn gem-run!
   [config args]
-  (let [jruby-config (jruby-core/initialize-config (jruby-puppet-core/extract-jruby-config (:jruby-puppet config)))]
+  (let [jruby-config (jruby-puppet-core/initialize-and-create-jruby-config config)]
```
rlinehan: I think this needs to be done for the other cli things too (irb and ruby).

jpinsonault: updated

rlinehan:
👍 🙈 I think this looks good to me, nice job!

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -206,126 +93,146 @@
...
+    profiler :- (schema/maybe PuppetProfiler)
+    agent-shutdown-fn :- IFn]
+   (initialize-and-create-jruby-config raw-config profiler agent-shutdown-fn true))
+  ([raw-config :- {schema/Keyword schema/Any}
+    profiler :- (schema/maybe PuppetProfiler)
+    agent-shutdown-fn :- IFn
+    warn-legacy-auth-conf? :- schema/Bool]
+   (let [jruby-puppet-config (initialize-puppet-config
+                              (extract-http-config (:http-client raw-config))
+                              (extract-puppet-config (:jruby-puppet raw-config)))
```
rlinehan: It seems like we could make the schema of `raw-config` at least a little more specific than `{schema/Keyword schema/Any}`, since if it doesn't have an `:http-client` key and a `:jruby-puppet` key with maps underneath each of them, it'll blow up.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -206,126 +93,146 @@
...
+
+(schema/defn initialize-and-create-jruby-config :- jruby-schemas/JRubyConfig
+  "Handles the initialization of the jruby-puppet config for the purpose of returning a
+  jruby config"
+  ([raw-config :- {schema/Keyword schema/Any}]
+   (initialize-and-create-jruby-config raw-config nil (fn [])))
+  ([raw-config :- {schema/Keyword schema/Any}
+    profiler :- (schema/maybe PuppetProfiler)
+    agent-shutdown-fn :- IFn]
+   (initialize-and-create-jruby-config raw-config profiler agent-shutdown-fn true))
```
rlinehan: I think this actually needs to default to `false` in order to not log the warning in the subcommands (try changing `use-legacy-auth-conf` to true in your puppetserver.conf and then running `lein gem`).

rlinehan: Also, it seems like this arity might not actually be used anywhere?

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -206,126 +93,146 @@
...
+        lifecycle-fns {:shutdown-on-error agent-shutdown-fn
+                       :initialize-pool-instance initialize-pool-instance-fn
+                       :cleanup cleanup-fn}
+        modified-jruby-config (assoc jruby-config :ruby-load-path (jruby-puppet-core/managed-load-path
+                                                                            (:ruby-load-path jruby-config)))]
+    (jruby-core/initialize-config (assoc modified-jruby-config :lifecycle lifecycle-fns))))
+
+(schema/defn initialize-and-create-jruby-config :- jruby-schemas/JRubyConfig
+  "Handles the initialization of the jruby-puppet config for the purpose of returning a
+  jruby config"
```
rlinehan: It might be good to update the docstring to say what the different arities do.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
jpinsonault: It shouldn't blow up, since `extract-*-config` can take in nil and return an empty map (since they use `select-keys`, but I agree that the schemas should be more specific. It might not always be that way, and it's easier for the reader

rlinehan: The schema for `extract-*-config` is `{schema/Keyword schema/Any}` which doesn't allow nil. So I guess the other option would be to do `(schema/Maybe {schema/Keyword schema/Any})` for those?

rlinehan: (sidenote, it always confuses me that `{schema/Keyword schema/Any}` doesn't allow nil, but `[schema/Any]` does.)

jpinsonault: ah, yep you're right. I was going to say your first suggestion made the most sense, but I'm thinking that the subcommands don't really need an http-client section. Maybe the schema for `raw-config` should be 
```
{(schema/optional-key :jruby-puppet) {schema/Keyword schema/Any}
 (schema/optional-key :http-client) {schema/Keyword schema/Any}
 schema/Keyword schema/Any}
```

It's just as permissive but better at being self-documenting? Hmm, that might be overkill though. Changing the `extract-` functions to use `maybe` might be better

jpinsonault: and yeah! `(get nil :something)` works fine, maps can be nil everywhere else it seems like

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
jpinsonault: I'd intended the 3-arity version to be the default used by the service, with the 4-arity version available if you wanted to turn off the logging. But having the logging off by default seems fine to me too, I can switch it around

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
rlinehan: I think that later on down the line, there are some things that must be present under `:jruby-puppet` or else things will fail (`:ruby-load-path` and `:gem-home` specifically), so maybe don't make that optional, but `:http-client` maybe should be?

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
jpinsonault: Oh hmm, I see, nevermind

jpinsonault: Yeah I'll switch it around. Probably keep the 1-arity version call the 4 arity version with false, and then have the service's call to this function pass in true, and get rid of the 3-arity

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
jpinsonault: yeah good point. That sounds reasonable

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -77,51 +55,16 @@
...
-
-(def EnvMap
-  "System Environment variables have strings for the keys and values of a map"
-  {schema/Str schema/Str})
-
-(def EnvPersistentMap
-  "Schema for a clojure persistent map for the system environment"
-  (schema/both EnvMap
-    (schema/either PersistentArrayMap PersistentHashMap)))
+(def CombinedJRubyPuppetConfig
```
camlow325: Does this get used anywhere other than from `jruby-puppet-testutils` - in a test namespace?  If not, I'd be inclined not to put this into this namespace, a production code namespace.  Also, it might be helpful to rename this to `JRubyPuppetTkConfig` or something like that to somehow indicate in the name that the config is the TK representation - as opposed a run-time representation that the services use.

jpinsonault: Yeah that sounds reasonable to me. @rlinehan I remember you having some thoughts on this

documentation/config_file_puppetserver.markdown
```diff
@@ -100,7 +100,7 @@ jruby-puppet: {
     master-run-dir: /var/run/puppetlabs/puppetserver
     master-log-dir: /var/log/puppetlabs/puppetserver
     max-active-instances: 1
-    max-requests-per-instance: 0
+    max-request-per-instance: 0
```
camlow325: This should still be `max-requests-per-instance` (plural), right?

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
rlinehan: I feel like I was maybe concerned about what if we at some point want to have a schema for the TK config and forget about what's in the test namespace... but really that's not a major concern and I think I agree with @camlow325 here.

documentation/config_file_puppetserver.markdown
jpinsonault: yes, thanks for finding that

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
jpinsonault: Cool ✔️

jpinsonault:
Addressed the last couple comments

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
```diff
@@ -20,55 +17,42 @@
...
-     (log/debug "Finished flush of JRuby pools for shutdown"))
+   (let [{:keys [pool-context]} (tk-services/service-context this)]
+     (jruby-core/flush-pool-for-shutdown! pool-context))
    context)
 
   (borrow-instance
     [this reason]
-    (let [{:keys [pool-context borrow-timeout event-callbacks]} (tk-services/service-context this)]
-      (core/borrow-from-pool-with-timeout pool-context borrow-timeout reason @event-callbacks)))
+    (let [{:keys [pool-context]} (tk-services/service-context this)
```
camlow325: Based on the current state of the code, it appeared to me that `borrow-instance` and `return-instance` at the service level are now only called by test code - since production code only ever calls `with-jruby-instance` (preceded by the new `get-pool-context` function).  If these are only called by test code, I'd be inclined to just make these functions in the jruby-testutils code and have those helpers use `get-pool-context` and `with-jruby-instance`, as appropriate.  WDYT?

rlinehan: I feel like at some point @cprice404 expressed a desire to keep these in this service. However, I might just have imagined that.

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -67,6 +70,23 @@
...
 
+(schema/defn ^:always-validate
+jruby-config :- jruby-schemas/JRubyConfig
+  "Create a JRubyConfig for testing. The optional map argument `options` may
+  contain a map, which, if present, will be merged into the final JRubyConfig
+  map.  (This function differs from `jruby-tk-config` in that it returns a map
+  that complies with the JRubyConfig schema, which differs slightly from the raw
+  format that would be read from config files on disk.)"
+  ([]
+   (jruby-core/initialize-config
```
camlow325: To cut down on the repetition a bit, seems like this form could be replaced with `(jruby-config {})`.

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
```diff
@@ -114,65 +99,28 @@
...
+        :doc "An alias for the jruby-utils' `with-jruby-instance` macro so
+             that it is accessible from the service namespace along with the
+             rest of the API."}
+  with-jruby-instance #'jruby-core/with-jruby-instance)
+
+(def #^{:macro true
+        :doc "An alias for the jruby-utils' `with-lock` macro so
+             that it is accessible from the service namespace along with the
+             rest of the API."}
+  with-lock #'jruby-core/with-lock)
```
camlow325: Is there any code left in this PR which exercises this?  If we think we might want to have it here just for the convenience of upstream consumers as opposed to just having them call into the  - e.g., PE Puppet Server code related to File Sync - might be good to at least have one test in OSS which confirms that this minimally does what it's supposed to do?

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
```diff
@@ -114,65 +99,28 @@
...
-      (finally
-        (core/unlock-pool pool# ~reason @event-callbacks#)))))
+    (let [pool-context (:pool-context (tk-services/service-context this))]
+      (jruby-core/register-event-handler pool-context callback-fn))))
+
+(def #^{:macro true
+        :doc "An alias for the jruby-utils' `with-jruby-instance` macro so
+             that it is accessible from the service namespace along with the
+             rest of the API."}
+  with-jruby-instance #'jruby-core/with-jruby-instance)
```
camlow325: Throughout the PR now, I see a number of places where we call `get-pool-context` and forward the result on to `with-jruby-instance`.  To cut down a bit on the work typical consumers would need to do, I wonder if it would be better to have this be a unique macro implementation that takes in a service, calls `get-pool-context` on it, and forwards the result down to a `jruby-core/with-jruby-instance` call.  Maybe the use of a nested macro would be problematic, though?

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
camlow325: I suppose also if the idea of a nested macro for `with-jruby-instance` were palatable - from my other comment - maybe doing something similar for `with-lock` would be worthwhile?

rlinehan: This is here just for pe-puppet-server-extensions. @cprice404 wanted to not have many services in pe-puppetserver/puppetserver have to know about the existence of jruby-utils, but instead to just have them use this service, which is now a very stripped-down wrapper on jruby-utils; the solution to that is to alias the namespace names. Since it's just an alias it doesn't really feel like a test is necessary, but that's just my opinion.

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
```diff
@@ -1,82 +1,31 @@
...
-  (merge (jruby-service-test-config 1)
-         {:http-client {:connect-timeout-milliseconds connect-timeout
-                        :idle-timeout-milliseconds    idle-timeout}}))
-
 (def default-services
   [jruby-puppet-pooled-service
+   jruby-pool-manager-service
    profiler/puppet-profiler-service])
 
-(deftest test-error-during-init
```
camlow325: This still seems to me like a useful test to have.  Do we have some other test instead that would cover ensuring that the actual error encountered when trying to create a pool instance percolates all the way out of the app?

rlinehan: There's a test of this in jruby-utils, using a service that calls `create-pool` from its `init`. Not sure whether it's necessary to also have a test in both places, but maybe jruby-utils isn't the right place for it? It just seemed like this behavior was something many users of `create-pool` would want to have, and so having such a test in `jruby-utils` made sense.

test/unit/puppetlabs/services/request_handler/request_handler_core_test.clj
```diff
@@ -326,31 +329,28 @@
...
+                                                            :ruby-load-path ["bar"]}))))]
+      (logutils/with-test-logging
+       (testing "slingshot bad requests translated to ring response"
+         (let [bad-message "it's real bad"]
+
+           (with-redefs [core/as-jruby-request (fn [_ _]
+                                                 (ringutils/throw-bad-request!
+                                                  bad-message))
+                         jruby-core/return-to-pool (fn [_ _ _] #())]
+             (with-redefs [jruby-core/borrow-from-pool-with-timeout (fn [_ _ _] {})]
```
camlow325: Would it be possible to avoid the `borrow-from-pool-with-timeout` with-redefs if this test were to fully initialize the pool with a just single "non-Puppet" instance in it - with a `jruby-pool-manager-core/create-pool` call instead of just a `jruby-pool-manager-core/create-pool-context` one?

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
camlow325: Yeah, on second thought, I think I'm good with this test going away.  Seems like it would have been more worthwhile in the first place as a fuller-blown integration-y style test that includes real Puppet Server services that might encounter an error during JRuby initialization and bubble the error all the way up the stack appropriately.  But this test wasn't that before anyway and may not be worth the effort to change it to be that now.

camlow325:
Looks like one test failed on one OS only, openjdk7, in Travis: 

```
lein test :only puppetlabs.services.certificate-authority.certificate-authority-int-test/empty-whitelist-defined-test

ERROR in (empty-whitelist-defined-test) (HttpAsyncRequestExecutor.java:341)
Uncaught exception, not in assertion.
expected: nil
2016-06-21 22:31:45,340 WARN  [clojure-agent-send-pool-14] [puppetserver] Puppet Comparing Symbols to non-Symbol values is deprecated
   (file & line not available)
2016-06-21 22:31:45,766 WARN  [async-dispatch-98] [o.e.j.s.h.ContextHandler] Empty contextPath
2016-06-21 22:33:47,556 WARN  [clojure-agent-send-pool-16] [puppetserver] Puppet Comparing Symbols to non-Symbol values is deprecated
   (file & line not available)
2016-06-21 22:33:47,990 WARN  [async-dispatch-102] [o.e.j.s.h.ContextHandler] Empty contextPath
  actual: org.apache.http.ConnectionClosedException: Connection closed
 at org.apache.http.nio.protocol.HttpAsyncRequestExecutor.endOfInput (HttpAsyncRequestExecutor.java:341)
    org.apache.http.impl.nio.DefaultNHttpClientConnection.consumeInput (DefaultNHttpClientConnection.java:263)
    org.apache.http.impl.nio.client.InternalIODispatch.onInputReady (InternalIODispatch.java:81)
    org.apache.http.impl.nio.client.InternalIODispatch.onInputReady (InternalIODispatch.java:39)
    org.apache.http.impl.nio.reactor.AbstractIODispatch.inputReady (AbstractIODispatch.java:123)
    org.apache.http.impl.nio.reactor.BaseIOReactor.readable (BaseIOReactor.java:164)
    org.apache.http.impl.nio.reactor.AbstractIOReactor.processEvent (AbstractIOReactor.java:339)
    org.apache.http.impl.nio.reactor.AbstractIOReactor.processEvents (AbstractIOReactor.java:317)
    org.apache.http.impl.nio.reactor.AbstractIOReactor.execute (AbstractIOReactor.java:278)
    org.apache.http.impl.nio.reactor.BaseIOReactor.execute (BaseIOReactor.java:106)
    org.apache.http.impl.nio.reactor.AbstractMultiworkerIOReactor$Worker.run (AbstractMultiworkerIOReactor.java:590)
    java.lang.Thread.run (Thread.java:745)
```

Not sure if this is directly related to the changes in this PR although I can't think of another case in which this test has failed recently.

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
jpinsonault: Agreed on the test

There is a lot of code duplication though, maybe it would be worth having another macro just for grabbing a jruby-puppet instance. But I also think it's probably not something we need to tackle right now

jpinsonault: er, I think I mean the second part to be a reply to the `with-jruby-instance` comment below

jpinsonault:
Hmm, I'd be curious to see if will fail a second time

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -198,13 +198,25 @@
...
+  This arity is indended for uses where a jruby-config is required but will not be used
+  to to create a pool, such as the cli ruby subcommands
+
+  The 4-arity function takes a profiler object which will be placed into the puppetserver
+  config through the :initialize-pool-instance lifecycle function. The agent-shutdown-fn
+  is run when a jruby-instance is terminated. When warn-legacy-auth-conf? is passed in as
+  true, it will log a warning that the use-legacy-auth-conf setting is depreciated if the
+  config setting is set to true as well."
+  ([raw-config :- {:jruby-puppet {schema/Keyword schema/Any}
+                   (schema/optional-key :http-client) {schema/Keyword schema/Any}
```
rlinehan: I think there maybe still needs to be a change the schema for `extract-http-config` to allow nil, because otherwise if this key isn't present then nil will be passed to it, which will make it unhappy.

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
jpinsonault: agreed

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
jpinsonault: dur, yes, thank you

test/unit/puppetlabs/services/request_handler/request_handler_core_test.clj
jpinsonault: I think @rlinehan rewrote this test, but I'm personally torn between not liking redefs, and liking that it doesn't take the time of creating a real pool.

rlinehan: Hmm, yeah, I think I did. I think maybe at the time `jruby-pool-manager-core/create-pool` didn't yet exist as a function. Now that it does, I think I'm more pro- getting rid of with-redefs. @jpinsonault happy to take a stab at rewriting it, if you'd like (I'm currently just waiting on pe-p-s-e tests to run).

camlow325: The overhead of creating a pool that just has one empty (non Ruby Puppet initialized) ScriptingContainer - which wouldn't be needed since this test doesn't need the backend Ruby Puppet code to be functional - should be pretty small, though, right?  I think it's the initialization of Puppet Ruby code into the container that typically takes a lot of time.

Not a big deal either way.  Just hoping to avoid another `with-redef` if we could.

jpinsonault: @camlow325 ah, yeah good point
@rlinehan I would not be opposed, if you have time, but I can take care of it otherwise

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -4,27 +4,31 @@
             [schema.core :as schema]
             [puppetlabs.kitchensink.classpath :as ks-classpath]
             [puppetlabs.kitchensink.core :as ks]
-            [puppetlabs.services.jruby.jruby-puppet-schemas :as jruby-schemas]
+            [puppetlabs.services.jruby.jruby-puppet-schemas :as jruby-puppet-schemas]
+            [puppetlabs.services.jruby-pool-manager.jruby-schemas :as jruby-schemas]
+            [puppetlabs.services.jruby-pool-manager.jruby-core :as jruby-core]
+            [puppetlabs.services.jruby.jruby-puppet-core :as jruby-puppet-core]
```
camlow325: When running `user-repl/reset` in a REPL on this branch, I got the following error:

~~~
Exception Circular dependency between puppetlabs.services.jruby.jruby-puppet-core and puppetlabs.services.jruby.jruby-puppet-core  clojure.tools.namespace.dependency.MapDependencyGraph (dependency.clj:89)
~~~

I found that if I removed this line and changed the references to it to just call the related functions without an alias that the error no longer occurred.  This seems unnecessary in any event, so probably worth removing.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -198,13 +198,25 @@
...
-    agent-shutdown-fn :- IFn]
-   (initialize-and-create-jruby-config raw-config profiler agent-shutdown-fn true))
+  jruby config. This function will only use the :jruby-puppet and :http-client sections
+  from raw-config. If values are not provided, everything in :http-client :jruby-puppet
+  will be given default values, except for :ruby-load-path and :gem-home in the
+  :jruby-puppet subsection.
+
+  The 1-arity function takes only a config and supplies a default of nil for the profiler,
+  an empty fn for the agent-shutdown-fn, and suppresses warnings about legacy auth.conf.
+  This arity is indended for uses where a jruby-config is required but will not be used
```
camlow325: s/indended/intended

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -198,13 +198,25 @@
...
+  to to create a pool, such as the cli ruby subcommands
```
camlow325: s/to to/to

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -198,13 +198,25 @@
...
+
+  The 4-arity function takes a profiler object which will be placed into the puppetserver
+  config through the :initialize-pool-instance lifecycle function. The agent-shutdown-fn
+  is run when a jruby-instance is terminated. When warn-legacy-auth-conf? is passed in as
+  true, it will log a warning that the use-legacy-auth-conf setting is depreciated if the
```
camlow325: s/depreciated/deprecated

camlow325:
Mostly 👍 at this point.  Just a few random nits around tests and the questions about the API for the JRuby service - whether or not the borrow/return functions should still be there and some comments about the implementations for the `with-jruby-instance` and `with-lock` functions, etc.  Happy to defer to @rlinehan and @jpinsonault where any suggestions on these may not make sense to do.

I did some functional testing involving spinning up puppetserver from this branch and concurrently doing various agent runs, full jruby-pool flushes, and max-requests flushes.  All seemed to work okay.  I also took a couple of memory snapshots after all of that had been run in a puppetserver process and didn't see any evidence of `ScriptingContainers` leaking, so all seems good.

test/unit/puppetlabs/services/request_handler/request_handler_core_test.clj
rlinehan: I attempted to do this, and it turned out to not really work. What would work would be to rewrite this test to be a full integration test. However, talking with @camlow325 in person we decided it wasn't worth spending any more time on it.

jpinsonault: Sounds good, thanks for doing that

rlinehan:
The same test looks like it failed again in travis on openjdk7

```
lein test :only puppetlabs.services.certificate-authority.certificate-authority-int-test/empty-whitelist-defined-test

ERROR in (empty-whitelist-defined-test) (HttpAsyncRequestExecutor.java:341)

Uncaught exception, not in assertion.

expected: nil

2016-06-21 22:31:45,340 WARN  [clojure-agent-send-pool-14] [puppetserver] Puppet Comparing Symbols to non-Symbol values is deprecated

   (file & line not available)

2016-06-21 22:31:45,766 WARN  [async-dispatch-98] [o.e.j.s.h.ContextHandler] Empty contextPath

2016-06-21 22:33:47,556 WARN  [clojure-agent-send-pool-16] [puppetserver] Puppet Comparing Symbols to non-Symbol values is deprecated

   (file & line not available)

2016-06-21 22:33:47,990 WARN  [async-dispatch-102] [o.e.j.s.h.ContextHandler] Empty contextPath

  actual: org.apache.http.ConnectionClosedException: Connection closed

 at org.apache.http.nio.protocol.HttpAsyncRequestExecutor.endOfInput (HttpAsyncRequestExecutor.java:341)

    org.apache.http.impl.nio.DefaultNHttpClientConnection.consumeInput (DefaultNHttpClientConnection.java:263)

    org.apache.http.impl.nio.client.InternalIODispatch.onInputReady (InternalIODispatch.java:81)

    org.apache.http.impl.nio.client.InternalIODispatch.onInputReady (InternalIODispatch.java:39)

    org.apache.http.impl.nio.reactor.AbstractIODispatch.inputReady (AbstractIODispatch.java:123)

    org.apache.http.impl.nio.reactor.BaseIOReactor.readable (BaseIOReactor.java:164)

    org.apache.http.impl.nio.reactor.AbstractIOReactor.processEvent (AbstractIOReactor.java:339)

    org.apache.http.impl.nio.reactor.AbstractIOReactor.processEvents (AbstractIOReactor.java:317)

    org.apache.http.impl.nio.reactor.AbstractIOReactor.execute (AbstractIOReactor.java:278)

    org.apache.http.impl.nio.reactor.BaseIOReactor.execute (BaseIOReactor.java:106)

    org.apache.http.impl.nio.reactor.AbstractMultiworkerIOReactor$Worker.run (AbstractMultiworkerIOReactor.java:590)

    java.lang.Thread.run (Thread.java:745)
```

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
jpinsonault: Oh weird. Probably intellij auto-imported it after I copy and pasted something from another namespace

rlinehan:
Right before that test fails, there's an interesting/maybe relevant error message

```
lein test puppetlabs.services.certificate-authority.certificate-authority-int-test

2016-06-21 22:28:40,950 WARN  [async-dispatch-90] [p.p.certificate-authority] The private CA key at '/home/travis/build/puppetlabs/puppetserver/dev-resources/puppetlabs/services/certificate_authority/certificate_authority_int_test/ca_true_test/master/conf/ssl/ca/ca_key.pem' was found to have the wrong permissions set as 'rw-rw-r--'. This has been corrected to 'rw-r-----'.
```

I wonder if this test failure is actually totally unrelated to this PR, it's just that it's the first PR to go in after https://github.com/puppetlabs/puppetserver/pull/1112?

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
rlinehan: Nested macros scare me (I have definitely seen them cause weird errors), but it seems at least worth exploring since it would cut down on some code duplication.

rlinehan:
That warning I mentioned might be a red herring - looks like it's also logged on oraclejdk8.

jpinsonault:
there's also the question of why it doesn't fail on that other PR

jpinsonault:
Tests pass on the latest commit. A race condition?

jpinsonault:
Pushed a couple commits.

The first fixes some typos, schemas, and test code.

The other commit removes the service fns, and adds a macro to wrap `with-jruby-instance`. Surprisingly not-painful to write. I didn't write a similar macro for with-lock, as it didn't seem as useful, but I can if we think it's useful.

I *think* that addresses all the feedback, let me know if I missed something.

jpinsonault:
Pushed a commit that adds a with-lock macro, and removes the macro aliases

rlinehan:
👍 💯

PR #1119, opened 2016-06-15
(SERVER-1354) Updates beaker and host gen pins

erikPrime:
Ran locally without issues.


justinstoller:
Those are valid versions. I trust you in running them locally, and CI is green so it's a great time to merge this kind of thing in case there is issue.

👍

PR #1120, opened 2016-06-17
(PE-16067) (PE-15848) Add i18n libraries and externalize strings

janelu2:
- Externalize strings in src/
- Add i18n libraries
- Add HTTP handlers for translations that are HTTP requests
- Add Makefile and PACKAGES variable for the use of multiple namespaces (see [this](https://github.com/puppetlabs/clj-i18n/blob/master/README.md))


janelu2:
@nathanielksmith @rlinehan can you guys take a look at this sometime? Maybe not this weekend, but eventually :) I mainly need someone to double check that the places where I used trs should be tru or vice versa. (tru is for HTTP requests and trs is all else)

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -432,23 +428,9 @@
       (or (false? (:authorization-required certificate-status-access-control))
           (not-empty certificate-status-whitelist))
       (log/warn
-        "The 'client-whitelist' and 'authorization-required' settings in the"
-        "'certificate-authority.certificate-status' section are deprecated and"
-        "will be removed in a future release.  Remove these settings and create"
-        "an appropriate authorization rule in the"
-        "/etc/puppetlabs/puppetserver/conf.d/auth.conf file.")
+       (trs "The ''client-whitelist'' and ''authorization-required'' settings in the ''certificate-authority.certificate-status'' section are deprecated and will be removed in a future release.  Remove these settings and create an appropriate authorization rule in the /etc/puppetlabs/puppetserver/conf.d/auth.conf file."))
```
libbymo: So 'client-whitelist', etc. Those are not going to be translated strings, I'm guessing.  So I would suggest pulling those out into interpolated variables.  You might also pull this first sentence out as a separate string, as then it could be translated once (once for this line, and one for line 433).  You'd need to join the resulting and following sentences again, with a (str) or something, to insert the space between it and the following sentences... but from a translation standpoint, it's probably worth it.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -984,8 +956,7 @@
     (let [subject (get-csr-subject csr)]
       (sling/throw+
        {:kind :disallowed-extension
-        :msg (format (str "CSR '%s' contains subject alternative names (%s), which are disallowed. "
-                          "Use `puppet cert --allow-dns-alt-names sign %s` to sign this request.")
+        :msg (tru "CSR ''{0}'' contains subject alternative names ({1}), which are disallowed. Use `puppet cert --allow-dns-alt-names sign {2}` to sign this request."
```
libbymo: Also the CLI command `puppet cert --allow-dns-alt-names sign {2}`, could this be a single string that's interpolated? Since I don't think this will be translated. So that all the translator sees is 
`"CSR ''{0}'' contains subject alternative names ({1}), which are disallowed. Use {2} to sign this request."`
@LarissaLane Can you give some guidance on this one? As I type that - does it now need comments for context? Oy vey, am I making a mountain out of a mole hill?

janelu2: Noted. I will change that. I'll write a translator comment so that the translator will understand that it is a command to be used by the user.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
janelu2: Ah ok, I think I was a bit confused about that. I thought that large blocks were to be kept together for the sake of context. In regards to this one it seems like the sentences being taken apart won't really affect the context. I might add some sort of comment for them.

src/clj/puppetlabs/services/puppet_admin/puppet_admin_service.clj
```diff
@@ -22,22 +23,9 @@
       (cond
         (or (false? (:authorization-required whitelist-settings))
             (not-empty client-whitelist))
-          (log/warn
-           "The 'client-whitelist' and 'authorization-required' settings in"
-           "the 'puppet-admin' section are deprecated and will be removed"
-           "in a future release.  Remove these settings and create an"
-           "appropriate authorization rule in the"
-           "/etc/puppetlabs/puppetserver/conf.d/auth.conf file.")
+        (log/warn (trs "The ''client-whitelist'' and ''authorization-required'' settings in the ''puppet-admin'' section are deprecated and will be removed in a future release.  Remove these settings and create an appropriate authorization rule in the /etc/puppetlabs/puppetserver/conf.d/auth.conf file."))
```
libbymo: Let's split these sentences up into separately translated strings...  as this is reused from the previous file, and used on line 28.

libbymo:
Noting here for posterity. Spoke with @janelu2 and @LarissaLane in that Jane will include a few comments around a CI command and a line with config options (basically things that should not be translated) - and Larissa will later include project wide translator notes - given the number of occurrences of config options in this repo. It would be crazy to try and comment them all.

janelu2:
Ok @libbymo and @nathanielksmith I've fixed up that repeated string and changed the "str" calls to "format". I've also added the comment to the CI command. I checked the .pot file and it correctly shows the comment above the message. Let me know if there are any more nitpicks :+1:

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1018,16 +995,15 @@
   (let [csr-path (path-to-cert-request csrdir subject)]
     (if (fs/exists? csr-path)
       (if (fs/delete csr-path)
-        (let [msg (str "Deleted certificate request for " subject)]
```
nathanielksmith: looks like these few msg end up being logged, not returned over an http request or something; i think that makes these `trs` instead of `tru`?

src/clj/puppetlabs/puppetserver/ringutils.clj
```diff
@@ -7,7 +7,8 @@
             [ring.util.response :as ring]
             [schema.core :as schema]
             [ring.util.response :as rr]
-            [cheshire.core :as cheshire]))
+            [cheshire.core :as cheshire]
+            [puppetlabs.i18n.core :as i18n :refer [trs tru]]))
```
nathanielksmith: looks like tru isn't used in this namespace

src/clj/puppetlabs/services/versioned_code_service/versioned_code_core.clj
```diff
@@ -16,26 +17,26 @@
...
+          (trs "exit-code: ''{0}'', stdout: ''{1}'', stderr: ''{2}''" exit-code stdout stderr)))
 
 (schema/defn ^:always-validate
   execution-error-msg :- schema/Str
   [cmd :- schema/Str
    e :- Exception]
-  (format (str "Running script generated an error. "
-               "Command executed: '%s', error generated: '%s'")
-              cmd (.getMessage e)))
+  (format (trs "Running script generated an error. Command executed: ''{0}'', error generated: ''{1}''"
```
nathanielksmith: this format seems vestigial

nathanielksmith:
nearly :+1:, just some minor stuff.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
janelu2: Oops overlooked that. Should be fixed now.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -984,9 +961,9 @@
     (let [subject (get-csr-subject csr)]
       (sling/throw+
        {:kind :disallowed-extension
-        :msg (format (str "CSR '%s' contains subject alternative names (%s), which are disallowed. "
-                          "Use `puppet cert --allow-dns-alt-names sign %s` to sign this request.")
-                     subject (str/join ", " dns-alt-names) subject)}))))
+        :msg (format "%s %s"
+                     (tru "CSR ''{0}'' contains subject alternative names ({1}), which are disallowed." subject (str/join ", " dns-alt-names))
+                     (tru "Use `puppet cert --allow-dns-alt-names sign {0}` to sign this request." subject))}))))
```
cprice404: So, unfortunately for almost all of the places in this file where we are throwing an exception with a message string in it, I fear that we are catching it in a block like this:

https://github.com/janelu2/puppetserver/blob/056c17224fdc93f88f59b2ff8266a391a5fdad4d/src/clj/puppetlabs/services/ca/certificate_authority_core.clj#L52-L57

where it then gets both logged *and* returned in the HTTP response.

I'm not sure that I care about getting those all audited and fixed as part of this PR, because I don't even have a great idea of what the solution is going to end up being.  My best idea at the moment is to put two entries in this exception, like a ':msg-user' and a ':msg-system', and then have the catch block use the appropriate one for the appropriate thing.

If we decide to punt that to another PR then we need to capture a ticket about it.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
cprice404: They also get returned in a map that seems like it might be used by liberator and maybe returned in the response?  I don't really understand the liberator stuff well enough to know for sure that that is true, might be worth just curling the endpoint and seeing what the response looks like, or looking at the API docs?  Or @nathanielksmith maybe you already navigated through the liberator stuff and determined that these maps we return here don't make it all the way to the http response?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1241,10 +1218,10 @@
...
-    (log/debugf "No custom OID mapping configuration file found at %s, custom OID mappings will not be loaded"
-                custom-oid-mapping-file)))
+    (log/debug (trs "No custom OID mapping configuration file found at {0}, custom OID mappings will not be loaded"
+                custom-oid-mapping-file))))
 
 (schema/defn ^:always-validate get-oid-mappings :- OIDMappings
   [custom-oid-mapping-file :- (schema/maybe schema/Str)]
   (merge (get-custom-oid-mappings custom-oid-mapping-file)
-         (clojure.set/map-invert puppet-short-names)))
+(clojure.set/map-invert puppet-short-names)))
```
cprice404: was this change intentional?

janelu2: Hmmm no I do not believe so. It was probably from when I was fixing conflicts merging upstream. Should I add that tab back in?

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -322,4 +320,5 @@
           authorization-fn
           whitelist-path
           (get-in ca-settings [:access-control :certificate-status]))
+        (i18n/locale-negotiator)
```
cprice404: nit: doesn't need to be changed, but just FYI: usually when you are writing a threading macro and there is a single-arg fn call, you won't see it wrapped with parens.  It doesn't hurt anything to wrap it with parens, just slightly less common syntax.

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
```diff
@@ -5,6 +5,7 @@
             [puppetlabs.services.jruby-pool-manager.jruby-core :as jruby-core]
             [puppetlabs.trapperkeeper.core :as trapperkeeper]
             [puppetlabs.trapperkeeper.services :as tk-services]
+            [puppetlabs.i18n.core :as i18n :refer [trs tru]]
```
cprice404: looks like you added the require here but aren't actually using them in this namespace?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
janelu2: I think that it is only being logged. Underneath each line with the let, msg is being passed to log/*. @nathanielksmith correct me if I'm wrong.

src/clj/puppetlabs/services/versioned_code_service/versioned_code_core.clj
```diff
@@ -16,25 +17,25 @@
...
+          (trs "Non-zero exit code returned while running ''{0}''." cmd)
+          (trs "exit-code: ''{0}'', stdout: ''{1}'', stderr: ''{2}''" exit-code stdout stderr)))
 
 (schema/defn ^:always-validate
   execution-error-msg :- schema/Str
   [cmd :- schema/Str
    e :- Exception]
-  (format (str "Running script generated an error. "
-               "Command executed: '%s', error generated: '%s'")
+  (trs "Running script generated an error. Command executed: ''{0}'', error generated: ''{1}''"
```
cprice404: It's possible that these execution error messages might bubble back up through the HTTP API endpoints like `catalog`, etc.; might be worth auditing.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
janelu2: I was told this is how our great jade leader @KevinCorcoran does it

src/clj/puppetlabs/puppetserver/certificate_authority.clj
cprice404: The message is being passed to log, and that one should be `trs` for sure.  The next line after that, though, is returning a map literal as the result of this fn, and that map literal contains the message.  So if that is bubbling up into the HTTP response via liberator then we're in another situation where we need both `trs` and `tru`.

In any case I'm OK with punting that audit to a follow-up ticket.

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
janelu2: Shoot, another upstream merge conflict resolution issue. It was used initially. I'll remove it right now.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
cprice404: If you end up making any other changes (e.g. to bump to the new release of i18n) then it'd be nice to go ahead and put it back.  Not critical though.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
cprice404: ok.  well, he's wrong.  :)  But since you're on his team for the moment you can leave it like this.

janelu2: hahaha ok, I'm going to change it anyway since that's convention.

cprice404:
I'm basically +1 on this with the possible exception of bumping the i18n version to fix the Makefile thing.

I think we should file a follow-up ticket that suggests auditing:

* whether the let blocks in certificate_authority.clj that are returning the maps with 'message' in them end up bubbling up to the http response via liberator
* what to do about the slingshot exceptions also in certificate_authority.clj, where it seems like we need both `tru` and `trs`
* Whether the execution errors in versioned_code_core.clj may bubble up to the http response of catalog or other requests.

janelu2:
@cprice404 i18n version/makefile things have already been updated! And I agree, but I think that maybe we can wait until @nathanielksmith takes a look at it (assuming he understands liberator more than you do) to see if the msg's are bubbling up to the HTTP response before filing a ticket.

cprice404:
oh, nice, didn't notice you already updated those.

We can wait for @nathanielksmith to look if you like, but we are going to have to file a ticket about the slingshot exceptions part regardless, unless you want to take on that work right now.  So if you wanted to create a jira to capture those three things that need auditing I'd be happy to merge this.  That'd reduce the risk of you having to deal with any additional rebasing / merge conflicts :)

janelu2:
@cprice404 ok! I will go ahead and make a ticket for those three. Might as well :)

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
KevinCorcoran: lies, damned dirty lies

cprice404:
For posterity: SERVER-1401 is the ticket describing the follow-up / auditing work described in the comment thread here.

PR #1122, opened 2016-06-23
(SERVER-1400) Update puppet submodule to 0adf62c

KevinCorcoran:
This should pull in fixes to the 'scheduled_task' beaker tests which are
currently failing in Jankins. This is the latest commit on the 'master'
branch of puppetlabs/puppet.


justinstoller:
Should we also update the versions of puppet we're testing against? https://github.com/puppetlabs/puppetserver/blob/master/acceptance/lib/helper.rb#L24-L33

KevinCorcoran:
yes

PR #1127, opened 2016-06-24
(MAINT) Update puppet version to ef8f234

cprice404:
This updates puppet and the corresponding puppet-agent versions;
the hope is that this will fix the current failures in the
master branch, which appear to be related to us having been pinned
to a version of puppet that is no longer available on the build servers.


jpinsonault:
👍

PR #1130, opened 2016-06-27
(SERVER-1402) (MAINT) Modify the pre-suite to ignore PUPPET_VERSION

kurtwall:
Modify the pre-suite to ignore `PUPPET_VERSION`/`puppet_version` and use  `PUPPET_BUILD_VERSION`/`puppet_build_version` consistently.

In 30_install_dev_repos.rb, we install a puppet-agent development repo on the master based on the value of `puppet_build_version`, which maps to the env var `PUPPET_BUILD_VERSION`. Later, in 70_install_puppet.rb, we _attempt_ to install the agent, but the logic ends up pulling in unwanted agent code.

As I read the code, at line 39, we fetch and check the value of `puppet_version`, which maps to the env var `PUPPET_VERSION`; if it is set, we fall through to the `elsif` on line 46 and install the default puppet-agent package, which is the _development_ version in the repo installed in 30_install_dev_repos.rb. But, if `puppet_version` is _not_ set, we fall through to line 49, where we fetch its value again (?) this time _without_ checking the returned value, and then attempt to install the agent using that value, which is empty/`nil`. As a result, we _still_ install the default puppet-agent package available, which is in the the dev repo installed earlier.

`PUPPET_VERSION` is not used in CI, AFAICT, but `PUPPET_BUILD_VERSION` is, so this change modifies the pre-suite to use the value of `PUPPET_BUILD_VERSION`.

### DIFF
--- a/acceptance/suites/pre_suite/foss/70_install_puppet.rb
+++ b/acceptance/suites/pre_suite/foss/70_install_puppet.rb
@@ -36,7 +36,7 @@ step "Install MRI Puppet Agents."
   hosts.each do |host|
     platform = host.platform
 
-    puppet_version = test_config[:puppet_version]
+    puppet_version = test_config[:puppet_build_version]
 
     if /windows/.match(platform)
       arch = host[:ruby_arch] || 'x86'
@@ -46,7 +46,7 @@ step "Install MRI Puppet Agents."
     elsif puppet_version
       install_package host, 'puppet-agent'
     else
-      puppet_version = test_config[:puppet_version]
+      puppet_version = test_config[:puppet_build_version]
 
       variant, _, _, _ = host['platform'].to_array
 


camlow325:
@kurtwall I have to admit that I'm not too familiar with all of the reasons that we ended up with the `puppet_build_version` vs. `puppet_version` split.

It seems like at least part of it is:

`puppet_build_version`, when set, represents a specific version of a repo that should be installed before the puppet-agent package is to be installed (done in `30_install_dev_repos.rb`).

whereas

`puppet_version` just represents a version number that could be passed along to the appropriate yum/apt install command for the puppet-agent package

There may be a place for both of these variables to be used although I don't know how common the need for it would be.

---

One way to have this work would be:

1) Change the logic in `initialize_config` from the acceptance/lib/helper.rb to only derive a default value for `puppet_version` from the “ruby/puppet” submodule if `puppet_build_version` is empty or nil.

~~~
puppet_build_version = get_option_value(options[:puppet_build_version], ...)

puppet_version = ""
if puppet_build.version.nil? || puppet_build_version.empty?
  puppet_version = get_option_value(options[:puppet_version], nil,
      "Puppet Version", "PUPPET_VERSION", nil, :string) || get_puppet_version
end
~~~

2) Use logic like this in `70_install_puppet.rb`:

~~~
  if /windows/.match(platform)
     puppet_build_version = test_config[:puppet_build_version]

     if (puppet_build_version.nil? || puppet_build_version.empty?)
       … error out since we don’t know where to install from?  …
     else
      ... install Windows agent msi from builds.delivery.puppetlabs.net using puppet_build_version variable value ...
    end
  else
    puppet_version = test_config[:puppet_version]

    if (puppet_version.nil? || puppet_version.empty?)
        install_package host, ‘puppet-agent’
    else
        ... do logic to install_package using puppet_agent=#{puppet_version} ...
    end
  end
~~~

---

I think the above would allow `puppet_build_version` to take priority over `puppet_version` by default.  But whenever `puppet_version` is set, it would actually be used in the install command.

Usually, then, we'd only be pinning `puppet_build_version` (and not `puppet_version`) in the `acceptance/lib/helper.rb`.

Happy to defer to others for better ideas that they may have.

cprice404:
ping @waynr in case he has any insight on why we needed/need the split.

waynr:
@cprice404 @kurtwall @camlow325 So I looked at the first commit where I introduced `PUPPET_BUILD_VERSION` and it seems like I was probably attempting not to break existing code for installing the version that matches the SHA of the puppet submodule.

As long as the behavior in the case where `PUPPET_BUILD_VERSION` is not specified remains the same (the puppet version used corresponds to the specified version of the puppet submodule) I think it should be safe to remove interpretation of the `PUPPET_VERSION` environment variable.

PR #1131, opened 2016-06-27
(SERVER-1307) reject CSRs with authorization extensions

nathanielksmith:
Flat out reject any CSRs with authorization extensions. For the
immediate future, such CSRs must be signed using the puppet cert tool.

This PR also adds a small helper to the CA tests to help make assertions about slingshot errors.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -954,6 +954,23 @@
       {:kind :invalid-signature
        :msg (tru "CSR contains a public key that does not correspond to the signing key")})))
 
+(schema/defn ensure-no-authorization-extensions!
+  "Throws an exception if the CSR contains authorization exceptions. These
+  extensions need to be signed using the puppet cert tool. This may change in
+  the future, but for now, it's too risky that certificates with authentication
+  extensions could be signed unintentionally."
+  [csr :- CertificateRequest]
+  (let [extensions (utils/get-extensions csr)]
+    (doseq [extension extensions]
+      (when (utils/subtree-of? ppAuthCertExt (:oid extension))
+        (sling/throw+
+         {:kind :disallowed-extension
+          :msg (format "%s %s"
+                (trs "CSR ''{0}'' contains an authorization extension, which is disallowed." csr)
+                (trs "Use {0} to sign this request."
+                     (format "`puppet cert --allow-authorization-extensions sign %s`"
+                             (get-csr-subject csr))))})))))
+
 (schema/defn ensure-no-dns-alt-names!
   "Throws an exception if the CSR contains DNS alt-names."
   [csr :- CertificateRequest]
@@ -983,6 +1000,7 @@
       (save-certificate-request! subject csr csrdir)
       (when (autosign-csr? autosign subject csr-stream ruby-load-path)
         (ensure-no-dns-alt-names! csr)
+        (ensure-no-authorization-extensions! csr)
         (validate-extensions! (utils/get-extensions csr))
         (validate-csr-signature! csr)
         (autosign-certificate-request! subject csr settings)
@@ -1195,6 +1213,7 @@
       ;; 'process-csr-submission!' when autosigning
       (validate-subject! subject csr-subject)
       (ensure-no-dns-alt-names! csr)
+      (ensure-no-authorization-extensions! csr)
       (validate-extensions! extensions)
       (validate-csr-signature! csr)
 
--- a/test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
+++ b/test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
@@ -154,6 +154,26 @@
                     :actual actual#})
         actual#))))
 
+;; This additional slingshot helper assumes our :kind/:msg maps and is probably
+;; brittle.
+(defmethod assert-expr 'thrown-with-slingshot-msg? [msg form]
+  (let [expected (nth form 1)
+        body     (nthnext form 2)]
+    `(sling/try+
+      ~@body
+      (do-report {:type :fail :message ~msg :expected ~expected :actual nil})
+      (catch map? actual#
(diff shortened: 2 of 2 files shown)


test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -154,6 +154,26 @@
...
+                              (and (= (:kind actual#) (:kind ~expected))
+                                   (re-matches (:msg ~expected) (:msg actual#))))
+                        :pass
+                        :fail)]
+          (do-report {:type result#
+                      :message ~msg
+                      :expected ~expected
+                      :actual actual#}))
+        actual#))))
+
```
KevinCorcoran: If memory serves, `thrown-with-slingshot?` exists because, at the time, we either didn't know about `slingshot.test/thrown+?` or couldn't get it to work.  Since then, I've used it elsewhere successfully.

There's also a `thrown+-with-msg?` in that same namespace, which looks very similar to this with the exception that it doesn't know about our error format.  

Given that, does it make sense to add this?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -954,6 +954,23 @@
...
+  (let [extensions (utils/get-extensions csr)]
+    (doseq [extension extensions]
+      (when (utils/subtree-of? ppAuthCertExt (:oid extension))
+        (sling/throw+
+         {:kind :disallowed-extension
+          :msg (format "%s %s"
+                (trs "CSR ''{0}'' contains an authorization extension, which is disallowed." csr)
+                (trs
+                 "Use `puppet cert --allow-authorization-extensions sign {0}` to sign this request."
+                 (get-csr-subject csr)))})))))
```
KevinCorcoran: This question mainly for my own understanding as I get used to this `trs` stuff ...

Is it necessary to use `format` and two `trs` calls here?  It seems to me like this could just be one `trs` call, but I could be misunderstanding something.

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -1209,6 +1209,33 @@
...
+          subject-dn   (utils/cn subject)
+
+          pp-auth-ext {:oid (:pp_authorization puppet-short-names)
+                       :value "true"
+                       :critical false}
+          pp-auth-role {:oid (:pp_auth_role puppet-short-names)
+                        :value "com"
+                        :critical false}
+          auth-csr (utils/generate-certificate-request subject-keys subject-dn [pp-auth-ext])
+          auth-role-csr (utils/generate-certificate-request subject-keys subject-dn [pp-auth-ext])]
```
KevinCorcoran: looks like you meant to use `pp-auth-role` here instead of `pp-auth-ext`

KevinCorcoran:
🙈

src/clj/puppetlabs/puppetserver/certificate_authority.clj
justinstoller: There's a tension between providing context with long strings vs the readability hit of not being able to break translated strings into multiple lines and saving costs by breaking out repeated strings. Breaking on sentence boundary is the recommended way to solve that tension.

We could do something like: `(trs "CSR ''{0}'' contains an authorization extensions, which is disallowed. Use 'puppet cert --allow-authorization-extensions sign {0} to sign this request." csr)`

But in the code it would all have to be one line. I think Nathaniel's code is more readable, and it also allows us to re-use the `"Use 'puppet cert --allow-authorization-extensions sign {0}' to sign this request."` elsewhere.

IIRC, we want to *not* translate strings that are either beyond our control (eg. certnames) or won't be translated for the user (eg. cli commands). I don't know if breaking out the cli invocation would also make the `Use...` string re-usable between authorization extensions and alt names? If we're not using similar word choice between the two responses,  since they have such similar workflows, should we?

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
nathanielksmith: ugh so i assumed we added our own because there *wasn't* something in slingshot, hence adding another custom thing. I'll put a note in the code so others don't do the same thing.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
nathanielksmith: yeah, all of that ^ and yeah, i should break out the CLI command.

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
nathanielksmith: delicious bowl of copypasta

nathanielksmith:
@justinstoller trs invocation updated
@KevinCorcoran i went ahead and got rid of our helper and switched us to thrown+? in a maint commit.

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -1189,34 +1186,61 @@
...
+          pp-auth-role {:oid (:pp_auth_role puppet-short-names)
+                        :value "com"
+                        :critical false}
+          auth-csr (utils/generate-certificate-request subject-keys subject-dn [pp-auth-ext])
+          auth-role-csr (utils/generate-certificate-request subject-keys subject-dn [pp-auth-role])]
+
+      (testing "pp_authorization is caught"
+        (is (thrown+-with-msg?
+             [:kind :disallowed-extension]
+              #".*contains an authorization extension.*borges.*"
```
justinstoller: I really like regexes that contain the minimum amount of a message that uniquely identifies it - or at least as much as I can bring myself to like a regex.   +1

justinstoller:
This looks great! The travis run failed on a race condition characterized in [SERVER-1269](https://tickets.puppetlabs.com/browse/SERVER-1269). Re-running to be sure and then will merge.

PR #1132, opened 2016-06-28
(MAINT) Replace server restarts with server HUPs

kurtwall:
Now that TK has HUP support, we can replace {some|many|most} server restarts with server HUPs, which will reduce acceptance suite's total runtime. This change makes 3 such changes. There are restarts or starts and stops in the acceptance suite, but they seemed to part of the test in those cases. 

In local testing, these 3 changes reduced the runtime of an ubuntu1604-mda-redhat6-64a-windows2008r2 config from 29:05 to 20:51. Runtime for a redhat6-64mda-ubuntu1404-64a-windows2008r2-64a went from 23:53 to 22:46.  Not highly rigorous testing, but there ya go.


erikPrime:
:+1:

PR #1134, opened 2016-07-07
(MAINT) Change pgrep to include o options (oldest)

erikPrime:
The pgrep -f puppetserver bash command very occasionally returns two
pids; the pid of the puppetserver and the pid of the pgrep command its
self.  When this happens, hup_server tries to kill -HUP two pids.  This
of course causes the test to fail.

This commit resolves this issue by added the -o option to the pgrep
command.  With addition of the -o option, pgrep only returns the 
oldest matching process.


camlow325:
👍 Seems fine to me.  That's what I do locally anyway.

PR #1136, opened 2016-07-09
(maint) Add semantic_puppet gem

kylog:
Note that puppet is planning to add semantic_puppet as a
hard dependency. This commit adds the semantic_puppet gem
as a git submodule under ruby/, so that developers can
work with versions of puppet that have that dependency.

It also updates the ruby load path in several places to
add semantic_puppet.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,3 +7,6 @@
 [submodule "ruby/hiera"]
 	path = ruby/hiera
 	url = https://github.com/puppetlabs/hiera.git
+[submodule "ruby/semantic_puppet"]
+	path = ruby/semantic_puppet
+	url = https://github.com/puppetlabs/semantic_puppet
--- a/dev/puppetserver.conf.sample
+++ b/dev/puppetserver.conf.sample
@@ -39,7 +39,7 @@ web-router-service: {
 jruby-puppet: {
     # Where the puppet-agent dependency places puppet, facter, etc...
     # Puppet server expects to load Puppet from this location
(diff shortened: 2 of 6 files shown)


cprice404:
@kylog isn't this still going to break in production?  Don't we still need to wait until we have a story for RE-7610 before we consider doing this?

kylog:
@cprice404 per the email thread from a few weeks back this is one of three prerequisites for introducing the dependency on semantic_puppet back into puppet:
* this PR, or something akin to it, to enable puppetserver development workflows
* https://github.com/puppetlabs/puppet-agent/pull/755 (I just pushed that up), or something akin to it, to enable puppetserver production workflows
* the work in https://tickets.puppetlabs.com/browse/RE-7392, to enable windows builds to come from vanagon (and thus take advantantage of the PR immediately above)

After all three of those things (or similar) have landed, we could then merge https://github.com/puppetlabs/puppet/pull/5094, which re-introduces the semantic_puppet dependency (originally added a month or two back, then reverted when we discovered the various implications).

That should all be prefaced with a big *I think*, but hope the gist of that makes sense. I'll drop a note in RE-7610 to similar effect.

cprice404:
@kylog ACK, thanks for the clarification.

cprice404:
I think we're going with a different strategy for this.  See:

* #1227 
* [SERVER-1412](https://tickets.puppetlabs.com/browse/SERVER-1412)
* [SERVER-1580](https://tickets.puppetlabs.com/browse/SERVER-1580)

Leaving this PR up for now as an example / reminder, but once the two tickets above have been proven out, we should close this PR.

cprice404:
Closed in favor of #1227

PR #1137, opened 2016-07-11
(maint) Update puppet-agent pin to 1.5.2

rlinehan:
The packages at the SHA for our previous puppet-agent pin were purged (they
get purged every 2 weeks), so this updates the puppet-agent pin to 1.5.2 (last
released) and puppet submodule to 4.5.2 (included in puppet-agent 1.5.2).


rlinehan:
I ran the `testest` smoke test with this and the `ruby/puppet/acceptance/tests/resource/scheduled_task` tests, since those were mentioned in the last commit that bumped these pins as failing.

camlow325:
Works for me.  😃 🎱 🚀

PR #1138, opened 2016-07-11
(maint) Bump puppet-agent version

rlinehan:
The packages for the previous puppet-agent SHA we were pinned too were purged.
This commit updates us to the latest passing puppet-agent version off stable.


kurtwall:
🙈

camlow325:
👍

PR #1139, opened 2016-07-12
(MAINT) Update to use PUPPET_BUILD_VERSION

kurtwall:
Using `PUPPET_BUILD_VERSION` because setting `PUPPET_VERSION` didn't work outside of CI


camlow325:
Sounds fine to me.  👍

PR #1140, opened 2016-07-13
(PUP-6335) Add note explaining duplication in certificate_authority.clj

haus:
The information captured in various vars in certificate_authority.clj is
duplicated in the puppet codebase. Creating a shared oids file isn't
easily achieveable today, so instead this commit adds a comment above
those vars in certificate_authority.clj explaining that any changes to
the oids need to be replicated in the puppet codebase.


haus:
See also: https://github.com/puppetlabs/puppet/pull/5117

PR #1142, opened 2016-07-18
(SERVER-1440) Close up test-503-when-app-shuts-down race condition a bit

camlow325:
This commit adds an initial GET request to the environments endpoint to
the `test-503-when-app-shuts-down` test.  Without this extra call, it
is more likely for the environment ping logic in the test to exhaust up to
the maximum configured 10 second wait if the initialization of the second
JRubyPuppet instance, the one servicing the environments requests, takes
a sufficiently long amount of time.  This would, in turn, cause
assertions related to the service stop to fail because the stop was not
far enough along before the assertions were made.

### DIFF
--- a/test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
+++ b/test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
@@ -221,7 +221,10 @@
            request {:uri "/puppet/v3/environments", :params {}, :headers {},
                     :request-method :GET, :body "", :ssl-client-cert cert, :content-type ""}
            ping-environment #(->> request (handler-core/wrap-params-for-jruby) (handler/handle-request handler-service))
+           ping-before-stop (ping-environment)
            stop-complete? (future (tk-app/stop app))]
+       (is (= 200 (:status ping-before-stop))
+           "environment request before stop failed")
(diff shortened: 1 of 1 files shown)


camlow325:
I was able to more reliably reproduce the error by surgically inserting a couple of sleeps in `jruby-utils`:

1) A > 10 second sleep before the second pool instance is created in `jruby-agents/prime-pool!`.

2) A < 10 second sleep after the second pool instance is created in `jruby-agents/prime-pool!`.

The second sleep needed to be less than 10 seconds to avoid triggering the test to fail for the same condition of an HTTP 200 OK response being returned for the environments request.

This would leave the test still racy but the likelihood of an errant failure should be significantly reduced.  It seems like it would be reasonably common for the initialization of the second JRubyPuppet instance to take 10 seconds or more - corresponding to the first sleep above.  It seems much less likely for the amount of time that elapses between the complete initialization of the pool, services being stopped, and subsequent requests to the environment endpoint to take up to a full 10 seconds.

At some point, we may want to consider making this test a little more white box to reduce the raciness of it - e.g., using the internal state of the `jruby-puppet-service` to validate when the service stop is actually underway before asserting that an HTTP request to the environments endpoint would return a 503.  This PR is a compromise for now to keep the test more black box while hopefully significantly reducing the likelihood of this particular test failure.

PR #1144, opened 2016-07-19
(SERVER-1442) Update FOSS puppetserver checkin to be on interval

briancain:
Prior to this commit, the checkin for puppetserver would only run on
start up. This commit updates that functionality by using
trapperkeeper/scheduler to schedule a regular interval checkin job that
will run at start up, and then again every 24 hours.

### DIFF
--- a/dev/bootstrap.cfg
+++ b/dev/bootstrap.cfg
@@ -10,6 +10,7 @@ puppetlabs.services.legacy-routes.legacy-routes-service/legacy-routes-service
 puppetlabs.services.puppet-admin.puppet-admin-service/puppet-admin-service
 puppetlabs.trapperkeeper.services.authorization.authorization-service/authorization-service
 puppetlabs.services.versioned-code-service.versioned-code-service/versioned-code-service
+puppetlabs.trapperkeeper.services.scheduler.scheduler-service/scheduler-service
 
 # To enable the CA service, leave the following line uncommented
 puppetlabs.services.ca.certificate-authority-service/certificate-authority-service
--- a/project.clj
+++ b/project.clj
@@ -59,6 +59,7 @@
                  [puppetlabs/jruby-utils "0.1.0"]
                  [puppetlabs/trapperkeeper ~tk-version]
                  [puppetlabs/trapperkeeper-authorization "0.7.0"]
+                 [puppetlabs/trapperkeeper-scheduler "0.0.1"]
                  [puppetlabs/kitchensink ~ks-version]
                  [puppetlabs/ssl-utils "0.8.1"]
                  [puppetlabs/ring-middleware "1.0.0"]
--- a/src/clj/puppetlabs/services/master/master_service.clj
+++ b/src/clj/puppetlabs/services/master/master_service.clj
@@ -17,6 +17,7 @@
    [:CaService initialize-master-ssl! retrieve-ca-cert! retrieve-ca-crl! get-auth-handler]
    [:JRubyPuppetService]
    [:AuthorizationService wrap-with-authorization-check]
+   [:SchedulerService interspaced stop-job]
    [:VersionedCodeService get-code-content]]
   (init
    [this context]
@@ -30,6 +31,7 @@
          product-name (or (get-in config [:product :name])
                           {:group-id    "puppetlabs"
                            :artifact-id "puppetserver"})
+         checkin-interval-millis (* 1000 60 60 24) ; once per day
          update-server-url (get-in config [:product :update-server-url])
          use-legacy-auth-conf (get-in config
                                       [:jruby-puppet :use-legacy-auth-conf]
@@ -39,7 +41,8 @@
                                                  [:jruby-puppet
                                                   :environment-class-cache-enabled]
                                                  false)]
-     (version-check/check-for-updates! {:product-name product-name} update-server-url)
+     (interspaced checkin-interval-millis
+                  (fn [] (version-check/check-for-updates! {:product-name product-name} update-server-url)))
 
      (retrieve-ca-cert! localcacert)
      (retrieve-ca-crl! hostcrl)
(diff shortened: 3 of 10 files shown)


src/clj/puppetlabs/services/master/master_service.clj
```diff
@@ -17,6 +17,7 @@
    [:CaService initialize-master-ssl! retrieve-ca-cert! retrieve-ca-crl! get-auth-handler]
    [:JRubyPuppetService]
    [:AuthorizationService wrap-with-authorization-check]
+   [:SchedulerService interspaced stop-job]
```
pcarlisle: stop-job is unused

briancain: Yeah I was wondering about that. We could save the job id, and in the shutdown context for tk also stop the job...but I don't see that for the MasterService? I only see a start context in this file.

pcarlisle: or does tk-scheduler alleviate any concerns about cleaning up at shutdown

briancain: I think if we're worried about leaking threads on shutdown from the job pool, tk-scheduler takes care of that for us by bringing it in as a service here: https://github.com/puppetlabs/trapperkeeper-scheduler/blob/master/src/puppetlabs/trapperkeeper/services/scheduler/scheduler_service.clj#L35-L40

So we probably don't need `stop-job`?

project.clj
```diff
@@ -59,6 +59,7 @@
                  [puppetlabs/jruby-utils "0.1.0"]
                  [puppetlabs/trapperkeeper ~tk-version]
                  [puppetlabs/trapperkeeper-authorization "0.7.0"]
+                 [puppetlabs/trapperkeeper-scheduler "0.0.1"]
```
smcclellan: Just noticed this hasn't been officially released to the github repo?

dev/bootstrap.cfg
```diff
@@ -10,6 +10,7 @@ puppetlabs.services.legacy-routes.legacy-routes-service/legacy-routes-service
 puppetlabs.services.puppet-admin.puppet-admin-service/puppet-admin-service
 puppetlabs.trapperkeeper.services.authorization.authorization-service/authorization-service
 puppetlabs.services.versioned-code-service.versioned-code-service/versioned-code-service
+puppetlabs.trapperkeeper.services.scheduler.scheduler-service/scheduler-service
```
briancain: Hmm, I thought this was the bootstrap.cfg that was used for tests but I have a few failures that say they can't find `:SchedulerService`. Not sure if that's because it isn't being loaded, or if I'm missing adding it into some sort of test utils setup for trapperkeeper for the tests?

project.clj
briancain: Oh hm...how do you mean? I believe the repo is public and it's on clojars.

cprice404: seems like it didn't get tagged in the git repo somehow, but yeah, it is on clojars.

dev/bootstrap.cfg
briancain: Also a question for testing this in the repl...I've got this change, but it doesn't look like it was properly loaded when I attempt to start puppetserver through the repl (it works if I do `lein run -c ~/.puppetserver/puppetserver.conf` though)

```
user=> (go)
2016-07-19 15:56:17,562 DEBUG [nREPL-worker-0] [p.t.internal] Initializing lifecycle worker loop.

RuntimeException Service ':SchedulerService' not found  puppetlabs.trapperkeeper.internal/handle-prismatic-exception! (internal.clj:98)
```

haus: Some tk deps are brought in using defs like this: https://github.com/puppetlabs/puppetserver/blob/8152bc8700009f96717d269958bf2b2fb569e7b1/test/unit/puppetlabs/services/config/puppet_server_config_service_test.clj#L21-L23. There may be more kicking around. That was the first I found in a search. Most have `deps` in the name.

briancain: Ah awesome...I'll fiddle with that and see if I can get an update to get the tests working at least.

I shouldn't have to mess with the tests however for the repl, right? Shouldn't the repl be loading `dev/bootstrap.cfg`? Maybe not...

haus: The repl probably needs a tweak here: https://github.com/puppetlabs/puppetserver/blob/master/dev/user_repl.clj#L47-L63

project.clj
KevinCorcoran: I noticed this the other day as well.  I wouldn't sweat it too much ... if we want to do something about it, I would suggest just doing another release (probably 1.0 since we're using it in production now) of tk-scheduler - and hopefully get a tag out of it this time.

briancain:
Woo thanks @haus ! I believe I've fixed the tests and the repl issue now.

src/clj/puppetlabs/services/master/master_service.clj
briancain: FYI, I don't think we need `stop-job`....so I'm going to remove it. I started up puppet-server in the repl, then used the `(stop)` function to confirm that the stop function is called as well for tk-scheduler and it looks like it is.

```
...
2016-07-20 11:41:18,300 DEBUG [async-dispatch-5] [p.t.s.s.scheduler-service] Shutting down Scheduler Service
2016-07-20 11:41:18,300 DEBUG [async-dispatch-5] [p.t.s.s.scheduler-service] Scheduler Service shutdown complete.
...
```

haus:
@briancain this looks totally sane to me. any chance you can move the jira ticket into the SERVER project (or clone it?) so that it can be referenced publicly?

haus:
i'm :+1:

cprice404:
@briancain note that this is merged into master, which means it will be in our 2.5.1 or 2.6.0 release as opposed to the forthcoming 2.5.0.  It sounds like that shouldn't be an issue since you are going to be doing a PE-specific version of this, but if it is, we can discuss.

briancain:
@cprice404 thanks! That's perfect. We will have a different pull request for the PE changes coming within the week, so no worries!

PR #1147, opened 2016-07-22
(MAINT) Wait for server to return 200 before returning

kurtwall:
After HUPping the server, the `hup_server` method should wait until the server returns a HTTP 200 on the status endpoint. The current behavior, bailing after getting an HTTP 404 on the status endpoint, returns control from `hup_server` before the server has fully restarted. As a result, some file sync commit operations that occur immediately following a `hup_server` call fail with `ECONNREFUSED` because the server has not even finished shutting down. With this change,`ECONNREFUSED` doesn't occur and life is wonderful.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -298,7 +298,7 @@ module PuppetServerExtensions
     key = get_key(master)
     response_code = "0"
     sleeptime = 1
-    while response_code != "404" && timeout > 0
+    while response_code != "200" && timeout > 0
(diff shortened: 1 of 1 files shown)


kurtwall:
Would like @erikPrime and maybe @KevinCorcoran to take a peek at this.

kurtwall:
I asked @erikPrime to go ahead and merge without waiting for @KevinCorcoran because I don't want CI red all weekend.

KevinCorcoran:
👍

PR #1148, opened 2016-07-27
(SERVER-1454) Bump puppet pins to pick up json-pure pin

camlow325:
This commit bumps up the Puppet submodule and puppet package versions in
order to pick up a pin of the json-pure gem to '~> 1.8', which allows
the Ruby spec tests to continue to be run under Ruby 1.9.3.  Previously,
the version of the json-pure gem that was being installed was not
pinned.  The recently released version, 2.0.2, dropped support for
running under versions of Ruby earlier than 2.0.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -28,7 +28,7 @@ module PuppetServerExtensions
 
     puppet_build_version = get_option_value(options[:puppet_build_version],
                          nil, "Puppet Development Build Version",
-                         "PUPPET_BUILD_VERSION", "3.8.3")
+                         "PUPPET_BUILD_VERSION", "e0e68f5a73990ab8767988cf2da503b76ce701de")
 
     @config = {
       :base_dir => base_dir,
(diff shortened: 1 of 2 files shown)


camlow325:
This puppet build is currently still running in the [Puppet 3.x pipeline](https://jenkins.puppetlabs.com/view/legacy_Puppet/view/3.x/job/legacy_puppet_packaging-callback_3.x/) but passed through the prior Puppet specs and packaging jobs so far.

camlow325:
I'll have two more PRs to put up behind this one - one for master and one for stable - so don't need to move the related JIRA card over until they all go through.

camlow325:
Looks like this one was green all the way through the legacy_Puppet Jenkins pipeline - https://jenkins.puppetlabs.com/view/legacy_Puppet/view/3.x/job/legacy_puppet_integration-system_windows_3.x/58/.  So this hopefully should be good to go for merging.

cprice404:
:see_no_evil:

PR #1149, opened 2016-07-28
(SERVER-1454) Bump puppet pins to pick up json-pure pin

camlow325:
This commit bumps up the Puppet submodule and puppet package versions in
order to pick up a pin of the json-pure gem to '~> 1.8', which allows
the Ruby spec test to continue to be run under Ruby 1.9.3.  Previously,
the version of the json-pure gem that was being installed was not
pinned.  The recently released version, 2.0.2, dropped support for
running under versions of Ruby earlier than 2.0.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -22,7 +22,7 @@ module PuppetServerExtensions
                          "PUPPETSERVER_VERSION", nil, :string)
 
     puppet_version = get_option_value(options[:puppet_version],
-                         nil, "Puppet Version", "PUPPET_VERSION", "1.5.2.300.ga3e58f3", :string) ||
+                         nil, "Puppet Version", "PUPPET_VERSION", "1.5.2-232-g53abb67", :string) ||
                          get_puppet_version
 
     # puppet-agent version corresponds to packaged development version located at:
@@ -30,7 +30,7 @@ module PuppetServerExtensions
     puppet_build_version = get_option_value(options[:puppet_build_version],
                          nil, "Puppet Agent Development Build Version",
                          "PUPPET_BUILD_VERSION",
-                         "a3e58f345f2f4dd48ed247ee731508fee3a57ec0", :string)
+                         "53abb674aee774fce29a0a80271cba3bdc4aa03d", :string)
 
     # puppetdb version corresponds to packaged development version located at:
     # http://builds.delivery.puppetlabs.net/puppetdb/
--- a/spec/puppet-server-lib/puppet/jvm/master_spec.rb
+++ b/spec/puppet-server-lib/puppet/jvm/master_spec.rb
@@ -24,7 +24,7 @@ describe 'Puppet::Server::Master' do
     end
 
     it "returns the correct puppet version number" do
-      expect(subject).to eq('4.6.0')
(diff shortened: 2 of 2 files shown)


camlow325:
The corresponding puppet-agent build passed through CI [here](https://jenkins.puppetlabs.com/view/puppet-agent/view/stable/view/puppet-agent/job/platform_puppet-agent_intn-van-sys_suite-daily-puppet-stable/177/).

I also ran the 'rake spec' tests locally to see that they pass with the new 'json_pure' gem version pin.

cprice404:
@camlow325 was this done as a merge-up from 1.x?  If not, perhaps we should do that first; otherwise it seems like next time we do that it will attempt to clobber these?

Also might be worth waiting for the 1.x jruby specs job to pass before merging this?  Just to validate that it's fixed in CI?

camlow325:
@cprice404 No, I didn't do it as a merge-up but that might be better.  Okay with you if I do both of them together in one PR or think I should do them separately?

camlow325:
@cprice404 fyi, those 1.x Ruby spec tests passed - see https://jenkins.puppetlabs.com/job/platform_puppetserver_unit-jruby_1.x/111/.

cprice404:
One PR is fine.

cprice404:
Maybe keep the merge commit separate from the commit where you update to the new puppet 4.x shas though.

camlow325:
@cprice404 Sounds good.  I updated the PR to start with a merge-up commit from the latest 1.x branch, followed by the second commit that bumps the Puppet submodule / Puppet-Agent pins like the original PR had.  Hope this makes sense.

acceptance/lib/helper.rb
```diff
@@ -303,9 +303,9 @@ def hup_server(host = master, timeout = 30)
       begin
         response = https_request(url, 'GET', cert, key)
       rescue StandardError => e
-        expected_errors = [ Errno::ECONNREFUSED, Errno::ECONNRESET, 
+        expected_errors = [ Errno::ECONNREFUSED, Errno::ECONNRESET,
                             OpenSSL::SSL::SSLError ]
-        if !expected_errors.include?e.class 
+        if !expected_errors.include?e.class
```
cprice404: You didn't change this here, so it's not a big deal... but is this actually valid syntax?!

camlow325: No clue but I agree it does look strange.  /CC @erikPrime and @kurtwall since I think they're the most knowledgeable about this stuff.

kurtwall: @cprice404 @camlow325 This is legit syntax -- I've seen it work.

erikPrime: I wrote the above, it ran for me in pry and locally.  Would you like a refactor to make it more readable?

camlow325: If it looked like this (and this still works), it would seem more readable to me:

~~~ruby
if !expected_errors.include?(e.class)
~~~

But not a big deal to me either way.

kurtwall: On Thu, Jul 28, 2016 at 10:00 AM, Jeremy Barlow <notifications@github.com>
wrote:

> if !expected_errors.include?(e.class)
>
>
​+1​

PR #1150, opened 2016-07-28
(SERVER-1453) Create helper for adding/updating TK config files

kurtwall:
The modify_tk_config beaker command corrupts config files. This PR adds a new helper method, `apply_one_hocon_setting`, to the test library. It adds or updates a single hocon setting in a TK config file. PR also adds a helper method to _remove_ a single hocon setting, `delete_one_hocon_setting`.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -317,6 +317,36 @@ module PuppetServerExtensions
     end
   end
 
+  def apply_one_hocon_setting(hocon_host,
+                              hocon_file_path, 
+                              hocon_setting,
+                              hocon_value)
+    hocon_manifest =<<-EOF.gsub(/^ {6}/, '')
+      hocon_setting { "#{hocon_setting}":
+        ensure => present,
+        path => "#{hocon_file_path}",
+        setting => "#{hocon_setting}",
+        value => #{hocon_value},
+      }
+    EOF
+    apply_manifest_on(hocon_host, hocon_manifest,
+                      {:acceptable_exit_codes => [0,2]})    
+  end
+
+  def delete_one_hocon_setting(hocon_host,
+                               hocon_file_path, 
+                               hocon_setting)
+    hocon_manifest =<<-EOF.gsub(/^ {6}/, '')
+      hocon_setting { "#{hocon_setting}":
+        ensure => absent,
+        path => "#{hocon_file_path}",
+        setting => "#{hocon_setting}",
+      }
+    EOF
+    apply_manifest_on(hocon_host, hocon_manifest,
+                      {:acceptable_exit_codes => [0,2]})    
+  end
+
   # appends match-requests to TK auth.conf
   #   Provides many defaults so that users of this method can simply
   #   and easily allow a host in TK auth.conf


acceptance/lib/helper.rb
```diff
@@ -316,7 +316,46 @@ def hup_server(host = master, timeout = 30)
...
+  #
+  # On the server specified in hocon_host, which defaults to master,
+  # update the trapperkeeper config file specified in hocon_file_path,
+  # adding/changing the setting specified in hocon_setting to the value
+  # specified in hocon_value
+  def apply_one_hocon_setting(hocon_host,
+                              hocon_file_path, 
+                              hocon_setting,
+                              hocon_value)
+    hocon_manifest =<<-EOF
```
erikPrime: Your previously deployed pattern where you remove the unneeded white space in the here doc with a gsub is a good one.

acceptance/lib/helper.rb
```diff
@@ -316,7 +316,46 @@ def hup_server(host = master, timeout = 30)
...
+      hocon_setting { "#{hocon_setting}" :
+        ensure  => present,
```
erikPrime: Nit, would like the hash rockets to line up.

acceptance/lib/helper.rb
```diff
@@ -316,7 +316,46 @@ def hup_server(host = master, timeout = 30)
...
+                               hocon_setting)
+    hocon_manifest =<<-EOF
+      hocon_setting { "#{hocon_setting}" :
+        ensure  => absent,
+        path => "#{hocon_file_path}",
+        setting => "#{hocon_setting}",
+      }
+    EOF
+    apply_manifest_on(hocon_host, hocon_manifest,
+                      {:acceptable_exit_codes => [0,2]})
```
erikPrime: Should 0 be an acceptable exit code here?

acceptance/lib/helper.rb
```diff
@@ -316,7 +316,46 @@ def hup_server(host = master, timeout = 30)
...
+    hocon_manifest =<<-EOF
+      hocon_setting { "#{hocon_setting}" :
+        ensure  => present,
+        path => "#{hocon_file_path}",
+        setting => "#{hocon_setting}",
+        value => #{hocon_value},
+      }
+    EOF
+    apply_manifest_on(hocon_host, hocon_manifest,
+                      {:acceptable_exit_codes => [0,2]})
```
erikPrime: Should we allow 0?

acceptance/lib/helper.rb
```diff
@@ -316,7 +316,46 @@ def hup_server(host = master, timeout = 30)
...
+
+  def delete_one_hocon_setting(hocon_host,
+                               hocon_file_path, 
+                               hocon_setting)
+    hocon_manifest =<<-EOF
+      hocon_setting { "#{hocon_setting}" :
+        ensure  => absent,
+        path => "#{hocon_file_path}",
+        setting => "#{hocon_setting}",
+      }
```
erikPrime: There needs to be an additional option here, "value".

haus:
Once merged should we submit this to one of the beaker libraries or pe_acceptance_tests so that others can use the helper?

acceptance/lib/helper.rb
kurtwall: Done. It's kind of brittle, though. Each time code indentation changes, the `gsub` had to be updated.

acceptance/lib/helper.rb
kurtwall: Why _wouldn't_ 0 be acceptable? Puppet is idempotent, so if the `puppet apply` doesn't change anything, it returns 0. If it changes something and there were no other errors, we get 2. Maybe I misunderstand your question.

acceptance/lib/helper.rb
kurtwall: Yes.

kurtwall:
@haus If someone wants to do some copypasta, no objection. I'd really rather pull it into a gem so there aren't promotions to do.

haus:
@kurtwall fair

acceptance/lib/helper.rb
erikPrime: My concern is that if we are expecting / requiring a change, and we get a 0 back, then our desired change wasn't applied for some reason.  OTOH, leaving it with 0 as an acceptable exit code means that if we reapply an existing setting, we don't throw error and die.

erikPrime:
Seems mergable to me.  @haus ?

acceptance/lib/helper.rb
```diff
@@ -316,7 +316,46 @@ def hup_server(host = master, timeout = 30)
...
+  #
+  # On the server specified in hocon_host, which defaults to master,
+  # update the trapperkeeper config file specified in hocon_file_path,
+  # adding/changing the setting specified in hocon_setting to the value
+  # specified in hocon_value
+  def apply_one_hocon_setting(hocon_host,
+                              hocon_file_path, 
+                              hocon_setting,
+                              hocon_value)
+    hocon_manifest =<<-EOF.gsub(/^ {6}/, '')
```
haus: If we do a lot of heredoc undentation, this method may be helpful to pull in as a helper:
https://github.com/puppetlabs/vanagon/blob/master/lib/vanagon/extensions/string.rb

haus:
Seems good to me. I had one comment that shouldn't block merging.

acceptance/lib/helper.rb
kurtwall: Right. I think it is a matter of indifference if it was applied or not, provided the end result is that the setting is set, which both 0 and 2 indicate. It might one day be enabled by default, for example. Or another test may have been sloppy and left state behind. We want to assign a given value to a given key. The output of the `apply` will show what happens, so we're not even losing any information.

acceptance/lib/helper.rb
kurtwall: Nice. Thanks @haus

kurtwall:
Unclosed. Oops.

PR #1151, opened 2016-07-28
(MAINT) Update auth_rules file_bucket test to expect 400 error

camlow325:
This commit updates the test in default_rules which allows an authorized
file_bucket_file request to expect an HTTP status code of 400 (Bad
Request) instead of 500 (Server Error).  Through work related to
PUP-3827 which was done in Ruby Puppet code, 400 (Bad Request) is now
the proper response that the test should see when including a malformed
md5 value.

### DIFF
--- a/acceptance/suites/tests/authorization/default_rules.rb
+++ b/acceptance/suites/tests/authorization/default_rules.rb
@@ -122,9 +122,12 @@ with_puppet_running_on(master, {}) do
 
   step 'file_bucket_file endpoint' do
     # We'd actually need to store a file in the filebucket in order to get
-    # back a 200, but we know that a 500 means we got past authorization
+    # back a 200, but we know that a 400 means we got past authorization.
+    # The 400 in this case is expected because "123" is a malformed md5
+    # checksum, which the file_bucket_file endpoint will reject as a bad
+    # request
     curl_authenticated('/puppet/v3/file_bucket_file/md5/123?environment=production')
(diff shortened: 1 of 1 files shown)


camlow325:
I tested this locally by running the test through beaker a failing SUT from this CI run - https://jenkins.puppetlabs.com/view/puppetserver/view/all/job/platform_puppetserver_integration-system_no-conditional_smoke-stable/LAYOUT=redhat6-64ma-64a,LDAP_TYPE=default,PLATFORM=default,label=beaker/138/testReport/junit/(root)/acceptance_suites_tests_authorization/default_rules_rb/.

cprice404:
This seems sane to me but I'd love for @justinstoller to look at it.

justinstoller:
We discussed this in HC but a summary (as I understand it) for posterity's sake:
This test works against puppet#master but not for puppet#stable currently and this is regressing now because our recent update to the pinned puppet moved to stable *prior* to the core team merging down changes that we require from master.

camlow325:
I'll need to put up a new PR in place of this which bumps us to a version of Ruby Puppet which works around the json_pure gem issue, as #1149 was intended to do.  We don't actually have a puppet-agent build off of the master branch to pin to, though, so it may be a bit before I can do that.  In the meantime, I'm just going to close out this PR.

camlow325:
I submitted #1153 as a follow-up PR to hopefully address this auth_rules test failure.

PR #1158, opened 2016-07-29
(SERVER-1455) Ezbake 0.4.3, empty pidfile fix

cprice404:
This commit upgrades us to ezbake 0.4.3, which contains a fix for
a bug where the packages would create an empty pidfile at install
time, before the service was started.


camlow325:
👍

PR #1160, opened 2016-08-01
(DO NOT MERGE) add restart-file path to global.conf

janelu2:
This is the file that is modified to add a restart file path.


cprice404:
I'm gonna close this until we are ready for this step, because there is a second place where this var needs to be set anyway.

PR #1161, opened 2016-08-03
(SERVER-584) Add support for customizing environment variables for JRuby

janelu2:
This PR adds tests to puppetserver following the changes to the jruby-utils library [in this PR ](https://github.com/puppetlabs/jruby-utils/pull/17). That change adds a whitelisting capability to allow for certain environment variables (in the form of a map) be specified by the user to be visible to any Ruby code. 

This PR adds a test to make sure that given environment variables "FOO" and "BAR" are set correctly and their correct values reachable.


janelu2:
@cprice404 are there any more tests that you think I should do to make this more comprehensive? Possibly add an actual conf file rather than mocking?

janelu2:
Also this build is currently failing because the jruby-utils PR has not yet been merged. This should pass once that one gets merged.

test/unit/puppetlabs/services/jruby/jruby_puppet_testutils.clj
```diff
@@ -17,6 +17,8 @@
 
 (def ruby-load-path ["./ruby/puppet/lib" "./ruby/facter/lib" "./ruby/hiera/lib"])
 (def gem-home "./target/jruby-gem-home")
+(def environment-vars {"FOO" "for_jruby"
+                       "BAR" "also_for_jruby"})
```
rlinehan: I think that the keys should be keywords since that's what's going to be passed in by the HOCON config.

ezbake/config/conf.d/global.conf
```diff
@@ -2,4 +2,5 @@ global: {
     # Path to logback logging configuration file; for more
     # info, see http://logback.qos.ch/manual/configuration.html
     logging-config: /etc/puppetlabs/puppetserver/logback.xml
+    restart-file: /opt/puppetlabs/server/data/puppetserver/restartcounter.txt
```
camlow325: Should this be in this PR?  This seems unrelated to the environment vars work.

test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
```diff
@@ -362,6 +362,35 @@
...
+     (jruby-testutils/jruby-puppet-tk-config
+      (jruby-testutils/jruby-puppet-config
+       {:ruby-load-path  jruby-testutils/ruby-load-path
+        :gem-home        jruby-testutils/gem-home
+        :master-conf-dir jruby-testutils/conf-dir
+        :master-code-dir jruby-testutils/code-dir
+        :master-var-dir  jruby-testutils/var-dir
+        :master-run-dir  jruby-testutils/run-dir
+        :master-log-dir  jruby-testutils/log-dir
+        :environment-vars jruby-testutils/environment-vars}))
```
jpinsonault: Not totally sure but it looks like all these config options are being set to the values `jruby-puppet-config` already spits out. Might not need them here

camlow325:
Just a nit about the title of the PR.  I'd probably call it something like "Add support for customizing environment variables for JRuby" or something like that.  Mainly, I think communicating the point that some production code changes are included in this PR, not just "test" changes.

jpinsonault:
Just had one small comment, looks good

test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
camlow325: Both @cprice404 and I were thinking it would be good to have some sort of integration-level test for this functionality which shows that an environment variable expressed in HOCON can be extracted and delivered all the way through to the running JRuby container.  I think this is a good test as it is but I think we could tweak it a bit to cover the HOCON configuration bit as well.  Maybe we could do something like:

1) Use the Trapperkeeper `load-config` function to parse a HOCON configuration file back into a map.

See https://github.com/puppetlabs/puppetserver/blob/5345de0b27fc34499ab07f04169b3f0ab8435197/test/integration/puppetlabs/puppetserver/bootstrap_testutils.clj#L39-L50 for an example of how this is done behind some of our test helpers.

In this case, I'm thinking that the HOCON file on disk might have a section like:

```hocon
jruby-puppet: {
    ...
    environment-vars: { "FOO": #{PATH} }
}
```

2) Merge the result from the `load-config` with what you're already passing through to `jruby-puppet-config`. 

I'd be fine with just loading up a HOCON file with just that one `environment-vars` setting in it.  Just exercising HOCON to see that the environment variable expansion is happening as expected seems like it would be good enough.

3) Validate that the value for `FOO` stored in the `jruby-env` is equal to the value that you'd get when calling:

```clj
(System/getEnv "PATH")
```

Make sense?

test/unit/puppetlabs/services/jruby/jruby_puppet_testutils.clj
```diff
@@ -126,7 +128,8 @@ create-mock-pool-instance :- JRubyInstance
                   :master-log-dir log-dir
                   :use-legacy-auth-conf false})
                 (jruby-core/initialize-config {:ruby-load-path ruby-load-path
-                                               :gem-home gem-home}))
+                                               :gem-home gem-home
+                                               :environment-vars environment-vars}))
```
camlow325: I'd prefer to leave the `:environment-vars` out of the common test helpers.  The setting of `:FOO` and `:BAR` values seems specific to the new tests that you're creating so I would scope those settings to only be used in the new tests that you're adding for this PR.

test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
janelu2: yeah! that makes sense. thanks for the detail!

camlow325:
I think it would be good to include some docs along with this PR which describe how someone might use the new environment variables configuration, along with some HOCON-specific syntax which would show a user how they would be able to inject an environment variable into JRuby from the shell environment of the Puppet Server process.  Might be best to start with writing this up as a sub-bullet in the `jruby-puppet` section of the [config_file_puppetserver.markdown](https://github.com/puppetlabs/puppetserver/blob/5345de0b27fc34499ab07f04169b3f0ab8435197/documentation/config_file_puppetserver.markdown) file.

camlow325:
I'm pretty much 👍 on this.  I left a few comments about some updates to the tests and some documentation that I think would be good to address before we consider merging this.

I played around a bit with this PR via `lein ruby` and everything seemed to work the way I expected it to.

In my puppetserver HOCON config file, I added:

```hocon
jruby-puppet: {
  environment-vars: { "FOO" : ${FOO} }
}
```

From the command line, I ran...

> FOO=bar lein ruby --config <mypuppetserver.conf> -- -e "puts ENV['FOO']"

... and got `bar` as output.  So looks good to me.

ezbake/config/conf.d/global.conf
janelu2: Accidentally squash from a different branch. I'll have that removed

janelu2:
@camlow325 I've changed the test to read from a HOCON config temp file and removed the hardcoded map that was there before. Please ignore my git mess, I was trying to remove from tracking the testutils file I used for the previous testing approach and accidentally removed them entirely. I think it's good now though! Let me know what you think :)

test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
```diff
@@ -362,6 +363,38 @@
...
+              {:ruby-load-path   jruby-testutils/ruby-load-path
+               :gem-home         jruby-testutils/gem-home
+               :master-conf-dir  jruby-testutils/conf-dir
+               :master-code-dir  jruby-testutils/code-dir
+               :master-var-dir   jruby-testutils/var-dir
+               :master-run-dir   jruby-testutils/run-dir
+               :master-log-dir   jruby-testutils/log-dir}
+              (tk-config/load-config (.getPath tmp-conf))))))
+      (let [jruby-service (tk-app/get-service app :JRubyPuppetService)
+            jruby-instance (jruby-testutils/borrow-instance jruby-service :test)
```
camlow325: Not critical for what this test is doing, but I think it would be good to have the last argument here match the last argument in the corresponding `return-instance` call below - i.e., make them both `:test` or `:settings-plumbed-test`.

camlow325:
With just one minor comment on the test, I'm 👍 on the code.

Are you still planning on including some updates to the documentation about the new setting with this PR?

janelu2:
@camlow325 fixed up that inconsistency, and yes I'm working on it now. Just wanted to get the code finished up first :)

documentation/config_file_puppetserver.markdown
```diff
@@ -26,6 +26,8 @@ The `puppetserver.conf` file contains settings for Puppet Server software. For a
 
     * `gem-home`: The location where JRuby looks for gems. It is also used by the `puppetserver gem` command line tool. If nothing is specified, JRuby uses the Puppet default `/opt/puppetlabs/server/data/puppetserver/jruby-gems`.
 
+    * `environment-vars:` Optional. A map of environment variables to be injected from the shell environment into the the Puppet Server process.
```
janelu2: @camlow325 not sure what else there is to say to end users

documentation/config_file_puppetserver.markdown
```diff
@@ -94,6 +96,8 @@ The `puppetserver.conf` file contains settings for Puppet Server software. For a
 jruby-puppet: {
     ruby-load-path: [/opt/puppetlabs/puppet/lib/ruby/vendor_ruby]
     gem-home: /opt/puppetlabs/server/data/puppetserver/jruby-gems
+    environment-vars: { "HOME" : ${HOME}
+                        "LANG" : "de_DE.UTF-8" }
```
janelu2: @camlow325 used ${HOME} to show how HOCON can use expansion and "LANG" to show that it can be hardcoded, although I'm not sure if it's desirable for users to do that

cprice404: I probably wouldn't use HOME since we happen to have special-cased that one at times.

documentation/config_file_puppetserver.markdown
camlow325: I might say something like the following instead.  What do you think?

Optional.  A map of environment variables which are made visible to Ruby code running within JRuby, for example, via the Ruby `ENV` class.  By default, the only environment variables whose values are set into JRuby from the shell are `HOME` and `PATH`.  The default value for the `GEM_HOME` environment variable in JRuby is set from the value provided for the `jruby-puppet.gem-home` key.  Note that any variable set from the map for the `environment-vars` key will override these defaults; although it would probably be best to not override `HOME`, `PATH`, or `GEM_HOME` here since these values are already configurable.

documentation/config_file_puppetserver.markdown
camlow325: Yeah, I agree with @cprice404 that including `HOME` here might not be something we want to encourage people to do.  I might just include something like `"FOO" : ${FOO}` just to show how it can be applied to arbitrary variable names.

janelu2:
@camlow325 I shamelessly took what you said, I think it's a great explanation. Also changed the example to `FOO`. Any other thoughts?

camlow325:
:shipit:  🚀

PR #1162, opened 2016-08-09
(MAINT) Update to latest green puppet-agent, prep for 1.6.0

cprice404:
This commit updates us to the latest versions of puppet/puppet-agent
that are passing CI.  These should be very close to the final
tagged versions for 1.6.0, so getting them into our CI now will
help make sure we're ready for release.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -22,7 +22,9 @@ module PuppetServerExtensions
                          "PUPPETSERVER_VERSION", nil, :string)
 
     puppet_version = get_option_value(options[:puppet_version],
-                         nil, "Puppet Version", "PUPPET_VERSION", "1.5.2.343.gba39f2f", :string) ||
+                         nil, "Puppet Version", "PUPPET_VERSION",
+                         "1.5.2.411.g5cb5a63",
+                         :string) ||
                          get_puppet_version
 
     # puppet-agent version corresponds to packaged development version located at:
@@ -30,7 +32,8 @@ module PuppetServerExtensions
     puppet_build_version = get_option_value(options[:puppet_build_version],
                          nil, "Puppet Agent Development Build Version",
                          "PUPPET_BUILD_VERSION",
-                         "ba39f2f2fd1b4def9a7f366d469acb1c5f6db824", :string)
+                         "5cb5a63982c01f3ed0182c25b61213d208573d8d",
+                         :string)
 
     # puppetdb version corresponds to packaged development version located at:
     # http://builds.delivery.puppetlabs.net/puppetdb/


acceptance/lib/helper.rb
```diff
@@ -30,7 +30,8 @@ def self.initialize_config(options)
     puppet_build_version = get_option_value(options[:puppet_build_version],
                          nil, "Puppet Agent Development Build Version",
                          "PUPPET_BUILD_VERSION",
-                         "ba39f2f2fd1b4def9a7f366d469acb1c5f6db824", :string)
+                         "1a2a63d86e8f7ee6cc354b988f49b36d007620f4",
```
camlow325: This looks like a Ruby Puppet repo SHA, not a Puppet-Agent package SHA.  Is this right?

camlow325: Here, we might want 5cb5a63982c01f3ed0182c25b61213d208573d8d.  See `SUITE_COMMIT` for https://jenkins.puppetlabs.com/view/puppet-agent/view/master/view/puppet-agent/job/platform_puppet-agent_intn-van-sys_suite-daily-puppet-master/178/.

cprice404: ugh, sorry, you're right, copypaste fail.

cprice404: HIPCHAT

cprice404:
fixed

acceptance/lib/helper.rb
```diff
@@ -30,7 +30,8 @@ def self.initialize_config(options)
     puppet_build_version = get_option_value(options[:puppet_build_version],
                          nil, "Puppet Agent Development Build Version",
                          "PUPPET_BUILD_VERSION",
-                         "ba39f2f2fd1b4def9a7f366d469acb1c5f6db824", :string)
+                         "5cb5a63982c01f3ed0182c25b61213d208573d8d",
```
camlow325: Also, I think we want to update the value for `PUPPET_VERSION` above from`1.5.2.343.gba39f2f` to `1.5.2-411-g5cb5a63`.  TBH, though, I'm not sure why we have to do that part.  I had been thinking at one point that just leaving `PUPPET_VERSION` set to "" was fine - and that the right package to install could be gleaned just from the `PUPPET_BUILD_VERSION`.

cprice404: I was following the confluence page.  It doesn't mention that.  :(  But it seems obviously wrong to have that variable defined and pointing to an incorrect value.

@kurtwall whatever happened when you were looking into getting rid of that?

cprice404: fuggit I'm just going to go ahead and change it for now and we can discuss more later

cprice404: done

camlow325:
:shipit:

acceptance/lib/helper.rb
kurtwall: @cprice404 I talked about it a little with @waynr and it seems like something deep in JJB's innards expect PUPPET_VERSION and are unhappy otherwise. I didn't pursue it beyond that.

PR #1163, opened 2016-08-09
(MAINT) Add long sleep to the hup_server helper

erikPrime:
This PR adds a 30 second sleep to the hup server helper.  If the server
is processing a large catalog when it recieves the hup signal, it will
delay processing the hup signal until it has completed the large
catalog, or until 30 seconds pass.

@camlow325 and I have discussed replacing hup_server with a full
puppetserver stop and start.  We've discussed making the hup_server
helper more robust, perhaps by having a fork run off and hit the status
endpoint every half second until it sees the status endpoint go down and
come back up...

Ultimately, the helper is going to be replaced with 'service puppetserver
restart' because when Chris and Jane's work is merged, the init scripts
ought to block until the server completes processing the HUP signal.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -293,6 +293,7 @@ module PuppetServerExtensions
   def hup_server(host = master, timeout = 30)
     pid = on(host, 'pgrep -fo puppetserver').stdout.chomp
     on(host, "kill -HUP #{pid}")
+    sleep 30
     url = "https://#{host}:8140/puppet/v3/status"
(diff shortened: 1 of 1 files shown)


camlow325:
👍

cprice404:
godspeed @janelu2

camlow325:
Hmm, although could you repoint this at the stable branch instead so we can bump this up into 2016.3.x on the PE side?

erikPrime:
Killing this PR to repoint it at stable...

PR #1170, opened 2016-08-11
(SERVER-1429) Add Server 2.5.0 release notes

gguillotte:


documentation/release_notes.markdown
```diff
@@ -8,85 +8,66 @@ canonical: "/puppetserver/latest/release_notes.html"
...
-This is a feature and bug-fix release of Puppet Server that also upgrades its included [Trapperkeeper][] framework from version 1.3.1 to 1.4.0.
+Released August 11, 2016.
 
-This release also adds packages for Ubuntu 15.10 (Wily Werewolf) and 16.04 LTS (Xenial Xerus), and no longer includes packages for Fedora 21, which reached its end of life in December.
+This is a feature and bug-fix release of Puppet Server.
 
-### New platforms: Ubuntu 15.10 (Wily Werewolf) and 16.04 LTS (Xenial Xerus)
+> **Note:** If you disabled the certificate authority (CA) on Puppet Server by editing the [`bootstrap.cfg`][service bootstrapping] file on older versions of Puppet Server, be aware that Puppet Server 2.5.0 no longer uses the `bootstrap.cfg` file.
+>
+> To ensure that CA services remain disabled after upgrading to Server 2.5.0, create `/etc/puppetlabs/puppetserver/services.d/ca.cfg` with the CA disabled *before* upgrading. The `puppetserver` service restarts on upgrade if the service running when you upgrade, which reloads the `ca.cfg` file.
```
cprice404: missing 'is': "if the service *is* running"

documentation/release_notes.markdown
```diff
@@ -8,85 +8,66 @@ canonical: "/puppetserver/latest/release_notes.html"
...
 
-When responding to unauthorized HTTPS requests, previous versions of Puppet Server 2.x returned the requester's IP address and [authorization rule][auth.conf] in addition to logging the failed request. Puppet Server 2.4.0 removes this information from the response and directs the responder to consult the server logs for details.
+When performing a clean installation of Puppet Server 2.5.0, no PID file is created, and `service puppetserver status` should return the expected `not running` message.
 
--   [TK-360: Error messages returned to client should not include IP or rule blocking the request](https://tickets.puppetlabs.com/browse/TK-360)
+-   [SERVER-1455](https://tickets.puppetlabs.com/browse/SERVER-1455)
+-   [EZ-84](https://tickets.puppetlabs.com/browse/EZ-84)
 
-### New feature: `always_retry_plugins` setting to configure Puppet feature caching
+### New feature: Signing CSRs with OIDs from a new arc
```
cprice404: I think there needs to be some tie-in in this section to the new features of tk-auth that allow you to define authorization rules based on these extensions.  I think we technically released that feature in 2.4.0, but at the very least we should mention it and link to the 2.4.0 docs about that.

@justinstoller do you have any suggestions on possible verbiage or where we should fit in that mention in this section?

cprice404: Also might move this to the top as it's probably the most interesting new feature?

cprice404:
LGTM other than the comments I made; I do think it's important to tie that auth stuff in to the tk-auth extensions work somehow.

documentation/release_notes.markdown
gguillotte: My main concern about bumping the bootstrap stuff from the top is because it represents a potential break on upgrade.

cprice404: oh yeah, sorry, I didn't mean above the bootstrap stuff.  just above the bugfixes.

erikPrime:
I only quickly skimmed.  This LGTM.

documentation/release_notes.markdown
justinstoller: One thing is that the new CLI work is shipped in the `puppet-agent` package. The clojure CA will treat these extensions the same way as it treats DNS alt names (ie it won't sign them).

This is less that a new thing is shipping in Server than a thing that just shipped (in puppet-agent), combined with things we've previously shipped (tk-auth updates in 2.4), combine to now allow for a much more secure policy around accessing Puppet Server's routes and potentially can help users automatically, in a secure way, promote components of their puppet infra (eg. compile masters -> mom)

i'm effing horrible with verbiage...

cprice404: Here's the release notes from 2.4.0 that cover the relevant feature:

https://docs.puppet.com/puppetserver/latest/release_notes.html#new-feature-x509-compliant-certificate-extensions-can-match-authorization-rules

if we have Real Docs relating to that feature, just linking those in here would be great.

gguillotte:
@cprice404 @justinstoller CSR note updated if you want to take another look.

documentation/release_notes.markdown
gguillotte: @cprice404 Added a direct link to the allow/deny rules doc, which includes the extensions map.

justinstoller:
👍

PR #1178, opened 2016-08-15
(maint) Do not fail if journalctl is empty

melissa:
This commit allows the `journal -u` command to return with a failure.
This would happen if a service has no output for some reason. Although
this is not common, this is the case for puppetserver on sles 12. The
start and stop of this process is not recorded in the systemd journal.
As a result, `journalctl -u puppetserver` will fail even though the
puppetserver service is running successfully.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -163,7 +163,7 @@ module PuppetServerExtensions
     FileUtils.mkdir_p(destination)
     scp_from master, "/var/log/puppetlabs/puppetserver/puppetserver.log", destination
     if use_journalctl
-      puppetserver_daemon_log = on(master, "journalctl -u puppetserver").stdout.strip
+      puppetserver_daemon_log = on(master, "journalctl -u puppetserver", :acceptable_exit_codes => [0,1]).stdout.strip
       destination = File.join(destination, "puppetserver-daemon.log")
       File.open(destination, 'w') {|file| file.puts puppetserver_daemon_log }
     else
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -157,6 +157,10 @@ module PuppetServerExtensions
       if version.to_i >= 15
         use_journalctl = true
       end
+    when /^sles$/
+      if version.to_i >= 12
+        use_journalctl = true
+      end
     end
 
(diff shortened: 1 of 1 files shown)


cprice404:
@melissa :see_no_evil: I'm fine with this being merged, but don't have enough context with SLES to validate it or anything.  If you are good with it, I am.  Maybe @erikPrime or @kurtwall wants to take a peek and then :ship:

melissa:
I would be happy if someone with more sles/systemd on sles knowledge looked at this.

It appeared that there was no standard output to the journal (only standard error), even though the puppetserver unit file was set to direct standard output there. So if the service doesn't start up for some reason, we should receive that output with this command and that will be copied over to puppetserver-daemon.log. From what @waynr was saying, the only reason this functionality exists is so we can access the log after tests have run to assess failures. It sounds like standard out would also be nice to have, but standard err should be sufficient

haus:
🍤

mckern:
From commit message:

> Although this is not commit,

Should be "Although this is not common," ?

acceptance/lib/helper.rb
```diff
@@ -154,13 +154,17 @@ def puppet_server_collect_data(host, relative_path)
...
+        use_journalctl = true
+      end
     end
 
     destination = File.join("./log/latest/puppetserver/", relative_path)
     FileUtils.mkdir_p(destination)
     scp_from master, "/var/log/puppetlabs/puppetserver/puppetserver.log", destination
     if use_journalctl
-      puppetserver_daemon_log = on(master, "journalctl -u puppetserver").stdout.strip
+      puppetserver_daemon_log = on(master, "journalctl -u puppetserver",  :acceptable_exit_codes => [0,1]).stdout.strip
```
mckern: (nitpick) Two spaces in `,  :acceptable`; not important, but if you fixup this PR at all maybe grab it?

mckern:
This seems pretty sane to me. I left some nitpick comments, but if there's pressure the get this merged I'm 👍 .

melissa:
@mckern updated

mckern:
@haus does 🍤 count as a 👍 ?

haus:
Yes

kurtwall:
I'll give a posthumous 👍. SLES is...special.

PR #1181, opened 2016-08-19
Fixes documentation about advanced logging

riton:
With the latest version of _puppetserver_, if I copy / paste the documentation you actually provide for converting access logs to JSON, I only get the following output in my log file:

```
{"@version":1}
```

without any information about the request made.

This patch corrects the documentation to obtain correct log using java _LogBack_.

### DIFF
--- a/documentation/config_logging_advanced.markdown
+++ b/documentation/config_logging_advanced.markdown
@@ -82,34 +82,36 @@ To add JSON logging for HTTP requests:
             <providers>
                 <version/>
                 <pattern>
-                    {
-                      "@timestamp":"%date{yyyy-MM-dd'T'HH:mm:ss.SSSXXX}",
-                      "clientip":"%remoteIP",
-                      "auth":"%user",
-                      "verb":"%requestMethod",
-                      "requestprotocol":"%protocol",
-                      "rawrequest":"%requestURL",
-                      "response":"#asLong{%statusCode}",
-                      "bytes":"#asLong{%bytesSent}",
-                      "total_service_time":"#asLong{%elapsedTime}",
-                      "request":"http://%header{Host}%requestURI",
-                      "referrer":"%header{Referer}",
-                      "agent":"%header{User-agent}",
+                    <pattern>
+                        {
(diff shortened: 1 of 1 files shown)


cprice404:
@riton thanks for the submission!  It seems like the key is the missing nested `<pattern>` tag?  If so, happy to fix that.

documentation/config_logging_advanced.markdown
```diff
@@ -69,7 +69,6 @@ To add JSON logging for HTTP requests:
 1.  Add the following Logback appender section to the `request-logging.xml` file:
 
     ``` xml
-    {% raw %}
```
cprice404: I believe that these are required for the tool that we use to convert our docs from the source code repo to the final HTML - @gguillotte can you confirm?

documentation/config_logging_advanced.markdown
```diff
@@ -82,39 +81,40 @@ To add JSON logging for HTTP requests:
...
+                        "response.content-length":"%responseHeader{Content-Length}",
+                        "response.server":"%responseHeader{Server}",
+                        "response.connection":"%responseHeader{Connection}"
+                      }
+                  </pattern>
                 </pattern>
             </providers>
         </encoder>
     </appender>
-    {% endraw %}
```
cprice404: Same with this one - @gguillotte ?

documentation/config_logging_advanced.markdown
gguillotte: Yeah, without these `{% raw %}` liquid tags, the pages will fail during the docs site build. This is due to the xml containing liquid-like structures, such as `{%statusCode}`.

documentation/config_logging_advanced.markdown
gguillotte: Same.

cprice404:
@riton if you want to update the PR to leave in those docs markup tags, we can merge it.  Otherwise we'll clean it up ourselves as soon as we get a chance, and then merge it.  Thanks again for bringing it to our attention!

riton:
Fixed and _pushed force_, this should be okay.

cprice404:
@riton thanks!

PR #1188, opened 2016-08-26
(SERVER-1502) Add the trapperkeeper-status service and web route

camlow325:
This commit adds the trapperkeeper-status service to Puppet Server and
registers an HTTP web-route for it at "/status".

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -259,28 +259,32 @@ module PuppetServerExtensions
   # and slightly modified.
   # url: (String) URL to poke
   # method: (Symbol) :post, :get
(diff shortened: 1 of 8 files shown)


cprice404:
+1

camlow325:
This one should have some sort of test which goes with it to validate that the status service is at least functional.  I think it would be best to do this as a beaker test rather than a Clojure integration test since this PR doesn't involve any production code changes - just packaging.  I'll look into doing the beaker test in a later PR.  Putting this up for now for review and to expedite our ability to use this in our Gatling Automation test setup for the upcoming 2.6.0 release.

camlow325:
I added a beaker acceptance smoke test for the status service in bd1c99d.

PR #1189, opened 2016-08-26
(SERVER-366) Limit log disk usage

jpinsonault:
This commit updates puppetserver's logback file to limit the amount of space
log files can take up.

Since logback is handling log rotation now, `logrotate` is disabled through a new ezbake setting

---

This PR relies on SNAPSHOT builds of ezbake 02.x and trapperkeeper 1.1x, so those two will need to be released before this is merged

The new logback changes will be ported to the master branch as well

### DIFF
--- a/ezbake/config/logback.xml
+++ b/ezbake/config/logback.xml
@@ -5,10 +5,18 @@
         </encoder>
     </appender>
 
-    <appender name="F1" class="ch.qos.logback.core.FileAppender">
+    <appender name="F1" class="ch.qos.logback.core.rolling.RollingFileAppender">
         <!-- TODO: this path should not be hard-coded -->
         <file>/var/log/puppetserver/puppetserver.log</file>
         <append>true</append>
+        <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
+            <!-- rollover daily -->
+            <fileNamePattern>/var/log/puppetserver/puppetserver-%d{yyyy-MM-dd}.%i.log.zip</fileNamePattern>
+            <!-- each file should be at most 10MB, keep 90 days worth of history, but at most 1GB total-->
+            <maxFileSize>10MB</maxFileSize>
+            <maxHistory>90</maxHistory>
+            <totalSizeCap>1GB</totalSizeCap>
+        </rollingPolicy>
         <encoder>
             <pattern>%d %-5p [%c{2}] %m%n</pattern>
         </encoder>
--- a/project.clj
+++ b/project.clj
@@ -1,7 +1,7 @@
-(def tk-version "1.1.4-SNAPSHOT")
+(def tk-version "1.1.4")
 (def tk-jetty-version "1.3.1")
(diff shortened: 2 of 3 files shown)


ezbake/config/logback.xml
```diff
@@ -5,10 +5,18 @@
...
         <file>/var/log/puppetserver/puppetserver.log</file>
         <append>true</append>
+        <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
+            <!-- rollover daily -->
+            <fileNamePattern>/var/log/puppetserver/puppetserver-%d{yyyy-MM-dd}.%i.log.zip</fileNamePattern>
+            <!-- each file should be at most 10MB, keep 90 days worth of history, but at most 1GB total-->
+            <maxFileSize>10MB</maxFileSize>
+            <maxHistory>90</maxHistory>
+            <totalSizeCap>1GB</totalSizeCap>
+        </rollingPolicy>
```
cprice404: @jpinsonault is there a way we can do this for the http access logs too?

jpinsonault: pushed a new commit with it in the access log as well

cprice404:
LGTM.  @jpinsonault do you think we need to do some manual testing on the access log one just to be safe?

Also, we should maybe file a couple of follow-up tickets; one for merging this stuff up to master, and one for merging it up to PE?  (We don't necessarily have to do the PE one any time real soon, and we can probably get away with targeting that at everett or something.)

project.clj
```diff
@@ -83,6 +89,7 @@
 
   :profiles {:dev {:source-paths  ["dev"]
                    :dependencies  [[org.clojure/tools.namespace "0.2.4"]
+                                   [ch.qos.logback/logback-access "1.1.7"]
```
cprice404: Oh and it might be worth adding a comment above these two places where you've added logback-access, to mention that they should hopefully not be necessary in the master branch?

jpinsonault:
I just tested the access log in an ezbake build and confirmed it is at least logging to the correct file still

jpinsonault:
And yeah definitely, I'll create some tickets

project.clj
jpinsonault: good idea

jpinsonault:
Updated with some comments, and bumped to the released version of TK 1.1.4
Now we just need to wait on the ezbake release

camlow325:
👍 Played around with this locally.  Seems to all work as advertised.

cprice404:
+1 from me once we've gotten rid of the ezbake SNAPSHOT version.

rlinehan:
👍 from me too.

jpinsonault:
Updated ezbake dependency to the released version \o/

rlinehan:
👍 I guess we should wait for travis? Although I don't think that the ezbake version will affect it?

cprice404:
Yeah IMHO Travis isn't going to do much for us here so I'd be ok with merging it now.

rlinehan:
:shipit:

PR #1190, opened 2016-08-29
(maint) Make shell-utils/validate-command! public

haus:
In some cases it would be advantageous to log a warning at application
startup (or fail to start entirely) if a command is invalid for some
reason. This is not possible while validate-command! is a private
function, so this commit removes the private annotation and moves it
into the public section of shell-utils.


justinstoller:
👍

PR #1191, opened 2016-08-29
(DO NOT MERGE) (SERVER-858) Bump JRuby to 1.7.26

camlow325:
This commit bumps the JRuby dependency up to 1.7.26, in order to resolve
SERVER-858 and SERVER-1408.


camlow325:
We're going to be doing some memory testing with JRuby 1.7.26 over the next few days so it's probably best to hold off on merging this until we're satisfied that it is going to be stable enough for us to consider shipping.

camlow325:
Just realized I used "master" instead of "1.x" in my base branch name so closed this and put up #1192 with the right name as a replacement.

PR #1192, opened 2016-08-29
(SERVER-858) Bump JRuby to 1.7.26

camlow325:
This commit bumps the JRuby dependency up to 1.7.26, in order to resolve
SERVER-858 and SERVER-1408.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -18,21 +18,20 @@
                  [puppetlabs/kitchensink ~ks-version]
                  [puppetlabs/ssl-utils "0.8.1"]
                  [puppetlabs/http-client "0.4.4"]
-                 [org.jruby/jruby-core "1.7.20.1"
+                 [org.jruby/jruby-core "1.7.26"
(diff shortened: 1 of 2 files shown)


camlow325:
Replaces #1191.  We're going to be doing some memory testing with JRuby 1.7.26 over the next few days so it's probably best to hold off on merging this until we're satisfied that it is going to be stable enough for us to consider shipping.

cprice404:
Just for clarity: I've talked to Jeremy about this IRL, and we're going to wait to merge this until the longer-running perf/memory test has shaken out.

camlow325:
The longer-running perf/memory test against code from the master (since merged to stable) branch has completed and did not uncover any issues.  No HTTP requests failed, agent mean response time as compared to runs including JRuby 1.7.20.1 were comparable, and no evidence of a memory leak was uncovered.

Given all of that, I think we're a go on JRuby 1.7.26 for the Puppet Server 2.6.0 release and we should be good to go for merging this into the 1.x branch as well.

jpinsonault:
👍

PR #1193, opened 2016-08-30
(SERVER-858) Upgrade to jruby-utils 0.2.1 / JRuby 1.7.26

camlow325:
This commit upgrades the jruby-utils dependency to 0.2.1, transitively
updating JRuby dependencies to 1.7.26.  Explicit JRuby dependencies were
removed since they can now be inherited from jruby-utils.


camlow325:
Currently running the tests locally but expecting them to pass.  ~puppet server tests

PR #1194, opened 2016-08-30
(MAINT) Bump puppet-agent pin to 71d8618

camlow325:
This commit bumps the puppet-agent pin forward to 71d8618 and
corresponding puppet submodule to 06fd34be.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -23,7 +23,7 @@ module PuppetServerExtensions
 
     puppet_version = get_option_value(options[:puppet_version],
                          nil, "Puppet Version", "PUPPET_VERSION",
-                         "1.6.0",
+                         "1.6.1-37-g71d8618",
                          :string) ||
                          get_puppet_version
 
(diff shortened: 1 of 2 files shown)


cprice404:
@camlow325 - presumably the lein tests didn't have that same error any more with this SHA?

Also, @pcarlisle mentioned that a recent Puppet commit introduced something that broke with JRuby, dunno if that fix has been merged yet?

pcarlisle:
yeah it was just merged

camlow325:
@cprice404 The Clojure unit test that failed related to [PUP-6650](https://tickets.puppetlabs.com/browse/PUP-6650) no longer fails for me locally.

This wouldn't pick up the recent fix merged for https://github.com/puppetlabs/puppet/pull/5246, though, so maybe we should wait until we have a puppet-agent with that in it?

cprice404:
Yes, I think we should wait for that.

camlow325:
Fair enough.  I'll close this and put up another one after the Puppet fix lands - and assuming its pipeline is green.

camlow325:
Hmm, actually I think we might have been able to merge this since it looks like the code that caused the JRuby failure is only on the puppet#master branch - https://github.com/puppetlabs/puppet/pull/5246/commits/a414067c3fbe811d63a75abe8c300c4ca7f4ae15 - whereas this PR was about bumping the puppet#stable branch pin.  In any event, looks like there's already a later build available so I'll put up a PR to bump up to that.

camlow325:
Put up #1199 as replacement PR.

PR #1195, opened 2016-08-31
(MAINT) Bump puppet pin to 4d95e13

camlow325:
This commit bumps the puppet submodule and package pin up to 4d95e13.


erikPrime:
LGTM.

PR #1198, opened 2016-08-31
(SERVER-1503) merge up logback changes to master from 1.x

jpinsonault:
This PR merges up the changes from 1.x to disable logrotate and let logback
manage puppetserver's log files

---

This PR can't be merged until TK and ezbake have new releases

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -60,6 +60,7 @@
                  [puppetlabs/trapperkeeper ~tk-version]
                  [puppetlabs/trapperkeeper-authorization "0.7.0"]
                  [puppetlabs/trapperkeeper-scheduler "0.0.1"]
+                 [puppetlabs/trapperkeeper-status "0.3.5"]
                  [puppetlabs/kitchensink ~ks-version]
                  [puppetlabs/ssl-utils "0.8.1"]
                  [puppetlabs/ring-middleware "1.0.0"]
--- a/project.clj
+++ b/project.clj
@@ -1,6 +1,6 @@
 (def clj-version "1.7.0")
-(def tk-version "1.4.0")
-(def tk-jetty-version "1.5.9")
+(def tk-version "1.4.2-SNAPSHOT")
+(def tk-jetty-version "1.5.10")
 (def ks-version "1.3.1")
 (def ps-version "2.6.0-master-SNAPSHOT")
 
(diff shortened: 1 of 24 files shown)


cprice404:
LGTM once the SNAPSHenanigans are resolved.

jpinsonault:
Updated tk and ezbake to released versions ❗

cprice404:
+1, @camlow325 any reason not to go ahead and merge this?

project.clj
```diff
@@ -40,8 +41,8 @@
                  ;; we have documentation that shows how users can use it to
                  ;; send their logs to logstash, so we include it in the jar.
                  ;; we may use it directly in the future
-                 [net.logstash.logback/logstash-logback-encoder "4.5.1"]
-
+                 [net.logstash.logback/logstash-logback-encoder "4.5.1" :exclusions [ch.qos.logback/logback-access
+                                                                                     ch.qos.logback/logback-core]]
```
camlow325: Just curious why we did this as an exclusion as opposed to putting an explicit logback version dependency in the "conflict resolution" section.

camlow325:
Looks fine to me - just had the one question about how we want to resolve the logback version conflict.

project.clj
cprice404: I suggested this approach because we are trying to manage the logback version via the versions specified in the Trapperkeeper project.clj.  Otherwise we risk ending up with different versions of some of the 3 logback artifacts, and when that happens, you are subject to weird AbstractMethod errors that aren't always caught by our tests (we shipped a broken release once due to this exact issue).

@jpinsonault maybe it wouldn't hurt to add a comment here, though, saying something like "logback versions should be inherited from trapperkeeper".

jpinsonault: I'm sure @cprice404 can articulate it better, but the gist I came away with is that we'd like to keep the logback dependencies in TK to avoid a whole mess of pedantic errors and version mismatches.

cprice404: @camlow325 what say you

jpinsonault: Didn't see your comment, yeah I'll add a comment

jpinsonault: [](http://s2.quickmeme.com/img/04/043e1508785dd6043fbef737166232d7054e7c942fc6249b0c50b17c2d111daf.jpg)

jpinsonault:
Added a comment about the logback exclusion

camlow325:
👍

PR #1196, opened 2016-08-31
(SERVER-1505) Pre docs for the status endpoint

camlow325:
This commit provides some pre-docs for the status endpoint.

PR #1203, opened 2016-09-06
(SERVER-1505) Pre docs for the status endpoint

camlow325:
This commit provides some pre-docs for the status endpoint.


camlow325:
@jpinsonault This should be identical to what was merged to master in #1196, just bringing it down to stable ahead of the upcoming 2.6.0 release.

jpinsonault:
👍

PR #1206, opened 2016-09-06
(maint) Update https_request test helper to include options

haus:
This commit updates the https_request test helper to include an options
hash, which initially can be used to change the read timeout for https
requests from the default of 60.


haus:
Once this is merged and voomed, there is some work in pe-pse that can be unwound.

justinstoller:
Looks sane. 👍 FWIW, travis won't have anything important to say about this change to an acceptance test helper.

PR #1211, opened 2016-09-07
(SERVER-709) Port test functions from pe-pse

haus:
pe-puppet-server-extensions has some updates to our test helper
functions that mean duplicated code between the two repos. This commit
moves over some of the changes to the catalog related functions so that
they can be removed from pe-pse.

### DIFF
--- a/test/integration/puppetlabs/puppetserver/testutils.clj
+++ b/test/integration/puppetlabs/puppetserver/testutils.clj
@@ -194,10 +194,13 @@
       json/parse-string))
 
 (schema/defn ^:always-validate write-site-pp-file
-  [site-pp-contents :- schema/Str]
-  (let [site-pp-file (fs/file conf-dir "environments" "production" "manifests" "site.pp")]
-    (fs/mkdirs (fs/parent site-pp-file))
-    (spit site-pp-file site-pp-contents)))
+  ([site-pp-contents :- schema/Str]
+   (write-site-pp-file (fs/file conf-dir "environments" "production") site-pp-contents))
+  ([base-dir :- File
+    site-pp-contents :- schema/Str]
+   (let [site-pp-file (fs/file base-dir "manifests" "site.pp")]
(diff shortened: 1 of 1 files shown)


test/integration/puppetlabs/puppetserver/testutils.clj
```diff
@@ -207,21 +210,41 @@
...
+          not)))
+  ([catalog :- PuppetCatalog
+    resource-type :- schema/Str
+    resource-title :- schema/Str
+    param-name :- schema/Str
+    param-value :- schema/Any]
+   (let [resources (get catalog "resources")]
+     (some #(resource-matches? resource-type resource-title % param-name param-value) resources))))
+
+(schema/defn catalog-code-id :- (schema/maybe schema/Str)
```
justinstoller: Where is this used in puppetserver testing? it looks net-new and the PR doesn't contain test updates to use it

haus: It isn't currently used, but it seemed to me like keeping all of the catalog helpers in one repo had benefits. I could also foresee tests being added for direct puppet work that exercised the helper.

justinstoller:
It's unfortunate that we have two similar but very differently implemented workflows around `resource-matches?` and `catalog-contains?` based on whether or not we're matching resource params. I'm tempted to see if we could further refactor the helpers into smaller methods (eg `match-resource-by-name`, `match-resource-by-param`) that we can compose together -- or that we update the tests that call these to test at the same level of specificity (ie pass in a dummy param even if unneeded). But I think those are refactors we might do if we continue with this and probably not worth the same bang for buck that we get with consolidating these helpers now.

👍

jpinsonault:
Agreed on the matching functions, it seems a little wonky, but I'm 👍 if it doesn't seem worth doing right now. Not sure if anyone else wants to vet this or if it can be merged

haus:
I'm fine playing with these a little to decompose them. That will increase the friction a little when taking them up in pe-pse.

PR #1217, opened 2016-09-15
(maint) Pass in a custom tmpdir to install_puppetlabs_dev_repos

haus:
Because the install_puppetlabs_dev_repo uses the same tmpdir under the
hood to prepare repos, if the contents of the repos are not the same,
they will end up causing package install failures, as was seen today
when a new Packages.bz2 file caused the second repo to be corrupted.
This commit manually cleans up the tmpdir to avoid this, until this can
be fixed in beaker itself.

### DIFF
--- a/acceptance/suites/pre_suite/foss/30_install_dev_repos.rb
+++ b/acceptance/suites/pre_suite/foss/30_install_dev_repos.rb
@@ -3,7 +3,7 @@ when :package
   step "Setup Puppet Server repositories." do
     package_build_version = ENV['PACKAGE_BUILD_VERSION']
     if package_build_version
-      install_puppetlabs_dev_repo master, 'puppetserver', package_build_version
+      install_puppetlabs_dev_repo master, 'puppetserver', package_build_version, 'tmp/repos/puppetserver'
     else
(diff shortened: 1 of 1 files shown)


acceptance/suites/pre_suite/foss/30_install_dev_repos.rb
```diff
@@ -4,6 +4,8 @@
     package_build_version = ENV['PACKAGE_BUILD_VERSION']
     if package_build_version
       install_puppetlabs_dev_repo master, 'puppetserver', package_build_version
+      # Cleanup the local repo copy to avoid problems
+      FileUtils.rm_rf('tmp/repo_configs') if File.exists?('tmp/repo_configs')
```
justinstoller: I think we would either want to delete tmp/repo_configs/<variant> (see [beaker/install_utils.rb#969](https://github.com/puppetlabs/beaker/blob/1.x/lib/beaker/dsl/install_utils.rb#L969)) or pass in a custom destination dir.

justinstoller:
👍

camlow325:
👍 Happy to merge if it checks out for @justinstoller.

PR #1218, opened 2016-09-15
(PE-17071) Include data_binding_terminus setting in puppet config keys

briancain:
This commit updates the puppet-config-keys map to include
data_binding_terminus to be able to determine that puppet is configured
to use hiera.

### DIFF
--- a/src/clj/puppetlabs/services/config/puppet_server_config_core.clj
+++ b/src/clj/puppetlabs/services/config/puppet_server_config_core.clj
@@ -34,6 +34,7 @@
     :csrdir
     :csr-attributes
     :dns-alt-names
+    :data-binding-terminus
     :hostcert
     :hostcrl
     :hostprivkey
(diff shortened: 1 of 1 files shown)


briancain:
Ah, I might of messed up by committing `ruby/puppet`. Let me know if I should remove that and rebase the commit...

rlinehan:
@briancain is there a puppet change that needs to be in place in order for this to work? If not, then yeah, you should remove the `ruby/puppet` submodule change. And if you do need a puppet change/new puppet version then there's a couple other things to change too.

briancain:
@rlinehan nope! It should just be a puppetserver change. I'll rebase my commit.

briancain:
Woot! Here's what I see in the config setting when I start up puppetserver from the repl:

```
...
:data-binding-terminus
 #object[org.jruby.RubySymbol 0x2830718a "hiera"],
...
```

cprice404:
:see_no_evil:  +1

cprice404:
@rlinehan this look ok to you since the submodule change was removed?

rlinehan:
yep!

PR #1224, opened 2016-09-28
(SERVER-1570) ExecutionStub should raise error on failonfail

cprice404:
Prior to this commit, Puppet Server's implementation of `ExecutionStub`
would ignore the `:failonfail` option.  The documented behavior is to
raise a `Puppet::ExecutionFailure` if the command returns a non-zero
exit code.

This commit refactors the ExecutionStub code a bit to allow the
options hash to be passed in (which also required some refactoring
of the tests).  Then, if `:failonfail` is set, we raise the appropriate
error when the exit code is non-zero.

### DIFF
--- a/spec/puppet-server-lib/puppet/jvm/execution_spec.rb
+++ b/spec/puppet-server-lib/puppet/jvm/execution_spec.rb
@@ -2,43 +2,63 @@ require 'spec_helper'
 
 require 'puppet/server/execution'
 
+def test_execute(command, args = nil, options = {})
+  if args.nil?
+    orig_command_str = command
+  else
+    orig_command_str = "#{command} #{args.join(" ")}"
+  end
+  # NOTE: the `execute` function is not part of the public API,
+  #  but is still the best level of abstraction to use for this sort
+  #  of test, so we're abusing Ruby's `send` for these tests.
+  Puppet::Server::Execution.send(:execute,
+                                 orig_command_str,
+                                 command, options, args)
+end
+
 describe Puppet::Server::Execution do
   context "execution stub" do
     it "should return an instance of ProcessOutput" do
-      result = Puppet::Server::Execution.execute("echo hi")
+      result = test_execute("echo hi")
       expect(result).to be_a Puppet::Util::Execution::ProcessOutput
     end
 
     it "should return the STDOUT of the process" do
-      result = Puppet::Server::Execution.execute("echo hi")
+      result = test_execute("echo hi")
       expect(result).to eq "hi\n"
     end
 
     it "should return an instance of ProcessOutput for a command with args" do
-      result = Puppet::Server::Execution.execute("echo",  ["hi"])
+      result = test_execute("echo",  ["hi"])
       expect(result).to be_a Puppet::Util::Execution::ProcessOutput
     end
 
     it "should return the STDOUT of the process for a command with args" do
-      result = Puppet::Server::Execution.execute("echo", ["hi"])
+      result = test_execute("echo", ["hi"])
       expect(result).to eq "hi\n"
     end
 
     it "returns an instance of ProcessOutput for a command with an empty array of args" do
-      result = Puppet::Server::Execution.execute("echo hi", [])
+      result = test_execute("echo hi", [])
       expect(result).to be_a Puppet::Util::Execution::ProcessOutput
       expect(result).to eq "hi\n"
     end
 
     it "should return the exit code of the process" do
-      result = Puppet::Server::Execution.execute("echo hi")
+      result = test_execute("echo hi")
       expect(result.exitstatus).to eq 0
 
-      result = Puppet::Server::Execution.execute("echo",  ["hi"])
+      result = test_execute("echo",  ["hi"])
       expect(result.exitstatus).to eq 0
 
-      result = Puppet::Server::Execution.execute("false")
+      result = test_execute("false")
       expect(result.exitstatus).not_to eq 0
+    end
+
+    it "should raise an error if `failonfail` is true and the process returns non-zero" do
+      expect {
+        test_execute("false", nil, {:failonfail => true})
+      }.to raise_error(Puppet::ExecutionFailure, /^Execution of 'false' returned 1/)
     end
   end
 end
--- a/spec/puppet-server-lib/puppet/jvm/execution_spec.rb
+++ b/spec/puppet-server-lib/puppet/jvm/execution_spec.rb
@@ -2,18 +2,12 @@ require 'spec_helper'
 
 require 'puppet/server/execution'
 
-def test_execute(command, args = nil, options = {})
-  if args.nil?
-    orig_command_str = command
-  else
-    orig_command_str = "#{command} #{args.join(" ")}"
-  end
+def test_execute(command, options = {})
   # NOTE: the `execute` function is not part of the public API,
   #  but is still the best level of abstraction to use for this sort
   #  of test, so we're abusing Ruby's `send` for these tests.
   Puppet::Server::Execution.send(:execute,
-                                 orig_command_str,
-                                 command, options, args)
+                                 command, options)
 end
 
 describe Puppet::Server::Execution do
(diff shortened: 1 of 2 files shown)


src/ruby/puppetserver-lib/puppet/server/execution.rb
```diff
@@ -7,31 +7,41 @@ class Puppet::Server::Execution
...
     if args && !args.empty?
       result = ShellUtils.executeCommand(command, args.to_java(:string))
     else
       result = ShellUtils.executeCommand(command)
     end
 
-    Puppet::Util::Execution::ProcessOutput.new(result.getOutput, result.getExitCode)
+    # TODO - not all options from Puppet::Util::Execution are supported yet, see SERVER-74
+
+    if options[:failonfail] and result.exit_code != 0
```
haus: I'm a huge fan of `&&` over `and` even though in this case it won't matter.

haus:
🙈

src/ruby/puppetserver-lib/puppet/server/execution.rb
```diff
@@ -7,31 +7,41 @@ class Puppet::Server::Execution
...
       [stdin, stdout, stderr].each { |io| io.close rescue nil }
 
-      execute(binary, args)
+      execute(orig_command_str, binary, options, args)
     end
   end
 
-  def self.execute(command, args = nil)
+  private
+  def self.execute(orig_command_str, command, options, args)
```
justinstoller: I'm not clear on why having the caller pass in the `orig_command_str` is better than having this method compute this value. It seems like this method has all the information to create that value, that `orig_command_str` is only used in logging, and creating the log message from the other args would be within this method's responsibilities.

I could see a case for passing this argument in to improve the method's testability, without which we would have to have tests understand the implementation of how the method computed the log message. However, since the consumer of the test is `initialize_execution_stub` and we aren't testing that method, our test helper has to know the implementation of how `initialize_execution_stub` will compute the `orig_command_str` value still (ie we're coupling the implementation at line 10 & 14 in this file with lines 6 through 10 in the helper and the regex in the assertion at line 61 in the execution_spec).

spec/puppet-server-lib/puppet/jvm/execution_spec.rb
```diff
@@ -2,43 +2,63 @@
...
+def test_execute(command, args = nil, options = {})
+  if args.nil?
+    orig_command_str = command
+  else
+    orig_command_str = "#{command} #{args.join(" ")}"
+  end
+  # NOTE: the `execute` function is not part of the public API,
+  #  but is still the best level of abstraction to use for this sort
+  #  of test, so we're abusing Ruby's `send` for these tests.
+  Puppet::Server::Execution.send(:execute,
```
KevinCorcoran: I followed this comment until I got to the last part: "... so we're abusing Ruby's `send` for these tests."  Can't this line just call `.execute` directly?

KevinCorcoran:
👍 thanks for finally fixing this

spec/puppet-server-lib/puppet/jvm/execution_spec.rb
justinstoller: `send` bypasses Ruby's private qualifier:
```
07:46:42 hobo:~ justin ‖ irb
irb(main):001:0> class Foo
irb(main):002:1>   def yo
irb(main):003:2>     "yo"
irb(main):004:2>   end
irb(main):005:1>  private
irb(main):006:1>   def b
irb(main):007:2>     "b"
irb(main):008:2>   end
irb(main):009:1> end
=> :b
irb(main):010:0> f = Foo.new
=> #<Foo:0x007ff9728076b8>
irb(main):011:0> f.b
NoMethodError: private method `b' called for #<Foo:0x007ff9728076b8>
        from (irb):11
        from /Users/justin/.rubies/ruby-2.2.2/bin/irb:11:in `<main>'
irb(main):012:0> f.yo
=> "yo"
irb(main):013:0> f.send :b
=> "b"
```

src/ruby/puppetserver-lib/puppet/server/execution.rb
cprice404: I copied this straight out of the puppet code, after discussing with @joshcooper about whether there was a sane way to avoid the duplication.  (There wasn't.)  I'd like to keep it as similar to the puppet code as possible so that it's as easy as possible to tell if something changes upstream that we need to port over.  So, I think the best choices here are:

1. Add a comment explaining that
2. Change it here, and in the puppet code as well (seems like overkill, but I'd be happy to do it)
3. Leave it as-is.

spec/puppet-server-lib/puppet/jvm/execution_spec.rb
cprice404: I added `private` to the production code to try to enforce that this is not public API.

src/ruby/puppetserver-lib/puppet/server/execution.rb
cprice404: I waffled on this too, and mostly did it this way because the logic was already here and was already splitting up the command, and I didn't want to have to re-assemble the command string inside of the private method.  But I think what you're suggesting makes sense, will take a quick pass to see what it looks like if I just push *all* of that logic down into the private method.

cprice404:
@justinstoller the refactor with your suggestion is much cleaner, so I added a commit that contains those changes.

@haus would you mind taking a quick peek at the test changes in the new commit, and make sure that you don't have any concerns about them breaking coverage of puppet's command array stuff that you were trying to make sure we covered the last time you were in this code?

justinstoller:
👍

src/ruby/puppetserver-lib/puppet/server/execution.rb
```diff
@@ -6,32 +6,43 @@
...
+    end
+
+    # TODO - not all options from Puppet::Util::Execution are supported yet, see SERVER-74
+
+    if options[:failonfail] and result.exit_code != 0
+      raise Puppet::ExecutionFailure, "Execution of '#{orig_command_str}' returned #{result.exit_code}: #{result.output.strip}"
     end
 
-    Puppet::Util::Execution::ProcessOutput.new(result.getOutput, result.getExitCode)
+    Puppet::Util::Execution::ProcessOutput.new(result.output, result.exit_code)
```
haus: These changes to what is called on result seem off.

haus: result should be an instance of `ExecutionResult`, which has no exit_code method (or output method).

cprice404: Welcome to the magic of JRuby.  For POJO-style java objects, if you have `getCamelCaseProperty` methods, JRuby supports accessing them via `camel_case_property`.  I thought it would be better to go with the ruby idioms in ruby code, but if people think it's more confusing this way I don't mind changing them back.

FWIW I believe these changes passed travis with no trouble yesterday (and the specs pass with these changes locally).

haus:
@cprice404 tests look reasonable to me. not sure if we have any tests that exercise a custom config_version, but that was the super funky case that led to some of the changes.

src/ruby/puppetserver-lib/puppet/server/execution.rb
haus: spock-smashing-computer

cprice404:
@haus did we add a test for `config_version` when you were making those changes before?  I'm not sure I really understand how that fits in here.

haus:
@cprice404 oh holy shit we did add a test

haus:
:a+:

src/ruby/puppetserver-lib/puppet/server/execution.rb
haus: okay, well as long as intellij/rubymine handles looking up those methods correctly, IDGAF.

cprice404: it does, I just clicky clicked to test.

PR #1225, opened 2016-09-28
(SERVER-1571) Support `:combine` to merge STDOUT/STDERR

cprice404:
This commit modifies the ShellUtils java code to support the ability
to combine STDOUT/STDERR into a single stream, and modifies the
Puppet::Server::Execution implementation of the Puppet ExecutionStub
so that if the `:combine` option is set, STDOUT/STDERR will be
combined.

### DIFF
--- a/spec/puppet-server-lib/puppet/jvm/execution_spec.rb
+++ b/spec/puppet-server-lib/puppet/jvm/execution_spec.rb
@@ -22,6 +22,13 @@ describe Puppet::Server::Execution do
       expect(result).to eq "hi\n"
     end
 
+    it "should combine STDOUT/STDERR of the process when :combine is set" do
+      result = test_execute("./spec/fixtures/puppet-server-lib/puppet/jvm/execution_spec/echo_stdout_and_stderr.sh",
+                            {:combine => true})
+      expect(result).to match(/hello stdout/)
+      expect(result).to match(/hello stderr/)
+    end
+
     it "should return an instance of ProcessOutput for a command with args" do
       result = test_execute(["echo", "hi"])
       expect(result).to be_a Puppet::Util::Execution::ProcessOutput
--- a/src/java/com/puppetlabs/puppetserver/ShellUtils.java
+++ b/src/java/com/puppetlabs/puppetserver/ShellUtils.java
@@ -139,7 +139,6 @@ public class ShellUtils {
     /**
      * Executes the given command in a separate process.
      *
-     *
      * @param command the command [String] to execute. arguments can be
      *                included in the string.
      * @return An ExecutionResult with output[String], error[String], and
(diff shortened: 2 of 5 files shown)


src/java/com/puppetlabs/puppetserver/ShellUtils.java
```diff
@@ -39,7 +56,13 @@ private static ExecutionResult executeExecutor(CommandLine commandLine,
         //    PipedInputStream stdoutInputStream = new PipedInputStream(stdoutOutputStream);
         // but this requires that the input stream be read on a different thread
         // than this one. this is currently out of scope.
-        ByteArrayOutputStream stdoutOutputStream = new ByteArrayOutputStream();
+        ByteArrayOutputStream stdoutOutputStream;
+        if (options.getCombineStdoutStderr()) {
+            log.debug("Combining STDOUT/STDERR for external command '" + commandLine.toString() + "'");
+            stdoutOutputStream = errStream;
```
camlow325: Is passing the same stream object for both `out` and `err` into `PumpStreamHandler` thread-safe?

src/java/com/puppetlabs/puppetserver/ShellUtils.java
```diff
@@ -14,23 +14,56 @@
 
 public class ShellUtils {
 
+    public static class ExecutionOptions {
```
camlow325: Big 👍 on this refactor.  I agree that the method signatures seemed like they might have continued to get out of hand if we had continued down the same road of adding more parameters / method overloads to the base `ShellUtils` class.

jpinsonault:
I think I'm +1 on this, I also wonder about setting stdout and stderr to the same object. I'm just not familiar with doing this kind of thing in java, so I don't know. If that's the right way to do it then I'm +1

spec/puppet-server-lib/puppet/jvm/execution_spec.rb
```diff
@@ -22,6 +22,13 @@ def test_execute(command, options = {})
       expect(result).to eq "hi\n"
     end
 
+    it "should combine STDOUT/STDERR of the process when :combine is set" do
+      result = test_execute("./spec/fixtures/puppet-server-lib/puppet/jvm/execution_spec/echo_stdout_and_stderr.sh",
+                            {:combine => true})
+      expect(result).to match(/hello stdout/)
```
cprice404: It seems non-deterministic at the OS level, because it's about the timing of flushes, not just writes.

src/java/com/puppetlabs/puppetserver/ShellUtils.java
cprice404: Based on this, I think it's safe:

http://svn.apache.org/viewvc/commons/proper/exec/trunk/src/main/java/org/apache/commons/exec/PumpStreamHandler.java?view=markup#l71

jpinsonault: Good enough for me

camlow325: Works for me too.

camlow325:
👍 :shipit:

PR #1226, opened 2016-09-29
(SERVER-1490) Integrate ezbake service reload

camlow325:
This PR bumps the Trapperkeeper dependency from 1.5.0 to 1.5.1 and
the EzBake dependency from 0.5.0 to 1.0 in order for Puppet Server
packages to make use of the new "service reload" functionality.

Most of the turmoil on this PR is from additions/changes to the beaker
acceptance tests related to the reload functionality, including:

1) Added some reload/start/stop/restart and umaks tests to exercise
   the EzBake changes made in conjunction with reload support.

2) Changed tests currently using the "hup" helper to instead use a
   replacement "reload" helper, which uses "service reload" functionality
   to refresh the server.


camlow325:
This PR will fail Travis currently because it is dependent upon a SNAPSHOT version of EzBake.  Once https://github.com/puppetlabs/ezbake/pull/363 is merged and EzBake 1.0 is released, I'll rebase the 1.0 dep into this PR.  At that point, this should be good to go for merging.

camlow325:
Ezbake 1.0 has been released and this PR has been updated to reference it.  This should then be ready for merge.

cprice404:
+1

acceptance/suites/tests/service_framework/reload_updated_crl.rb
```diff
@@ -0,0 +1,101 @@
...
+cert_to_revoke="reload_updated_crl"
+server = master.puppet['certname']
+host_crl_file = master.puppet['hostcrl']
+ca_crl_file = master.puppet['cacrl']
+inventory_file = master.puppet['cert_inventory']
+
+teardown do
+  step 'Remove revoked CRL cert' do
+    on(master, puppet('cert', 'clean', cert_to_revoke),
+       {:acceptable_exit_codes => [0, 24]})
```
rlinehan: what does exiting 24 mean?

camlow325: It's what `cert clean` returns if the certificate that you want to clean doesn't exist.  I had it tolerate either 0 (successful removal) or 24 so that if the test fails at some step before the cert is created, it will still continue on to clean everything else up in the teardown.

camlow325: Why it's 24 - no idea, though.  I didn't bother looking, TBH.

rlinehan: Would it be worth adding a comment that that is what 24 means, or is that somewhat common knowledge?

camlow325: I'll add a comment about why this allows 24.  Thanks.  I doubt it's common knowledge - I didn't know it before I wrote this test.

rlinehan:
👍 lgtm!

cprice404:
I'm not going to wait for travis because it doesn't seem like anything in this PR would affect it.  Sounds like @camlow325 is adding a comment so we can wait for that, but then :ship:

acceptance/suites/tests/service_framework/reload_updated_crl.rb
camlow325: @rlinehan I pushed up another commit, 8c30858, which just adds some comments about why 24 is tolerated as an exist code in a few places in the test.

PR #1227, opened 2016-09-30
(SERVER-1412) package gems

cprice404:
This PR adds the `semantic_puppet` and `hocon` gems to the puppetserver package, by installing the gems at packing time and then including them in the package.  It also adds the ability to configure the `:gem-path`, which by default will now include both the new, vendored jruby gems directory, and the original jruby gems directory.

(I've built packages and run the smoke tests in beaker locally, so I think this is ready to go.)

### DIFF
--- a/dev/puppetserver.conf.sample
+++ b/dev/puppetserver.conf.sample
@@ -44,9 +44,14 @@ jruby-puppet: {
     # Puppet server expects to load Puppet from this location
     ruby-load-path: [./ruby/puppet/lib, ./ruby/facter/lib, ./ruby/hiera/lib]
 
-    # This setting determines where JRuby will look for gems.  It is also
-    # used by the `puppetserver gem` command line tool.
+    # This setting determines where JRuby will install gems.  It is used for loading gems,
+    # and also by the `puppetserver gem` command line tool.
     gem-home: ./target/jruby-gems
+
+    # This setting defines the complete "GEM_PATH" for jruby.  If set, it should include
+    # the gem-home directory as well as any other directories that gems can be loaded
+    # from (including the vendored gems directory for gems that ship with puppetserver)
+    gem-path: ./target/vendored-jruby-gems
 
     # (optional) path to puppet conf dir; if not specified, will use the puppet default
     master-conf-dir: ./target/master-conf-dir
--- a/dev/puppetserver.conf.sample
+++ b/dev/puppetserver.conf.sample
@@ -51,7 +51,7 @@ jruby-puppet: {
     # This setting defines the complete "GEM_PATH" for jruby.  If set, it should include
     # the gem-home directory as well as any other directories that gems can be loaded
     # from (including the vendored gems directory for gems that ship with puppetserver)
-    gem-path: ./target/vendored-jruby-gems
+    gem-path: ${jruby-puppet.gem-home}":./target/jruby-gems"
 
     # (optional) path to puppet conf dir; if not specified, will use the puppet default
     master-conf-dir: ./target/master-conf-dir
--- a/dev/puppetserver.conf.sample
+++ b/dev/puppetserver.conf.sample
@@ -51,7 +51,7 @@ jruby-puppet: {
     # This setting defines the complete "GEM_PATH" for jruby.  If set, it should include
     # the gem-home directory as well as any other directories that gems can be loaded
     # from (including the vendored gems directory for gems that ship with puppetserver)
-    gem-path: ${jruby-puppet.gem-home}":./target/jruby-gems"
+    gem-path: [${jruby-puppet.gem-home}, "./target/jruby-gems"]
 
     # (optional) path to puppet conf dir; if not specified, will use the puppet default
     master-conf-dir: ./target/master-conf-dir
--- a/documentation/config_file_puppetserver.markdown
+++ b/documentation/config_file_puppetserver.markdown
@@ -26,6 +26,8 @@ The `puppetserver.conf` file contains settings for Puppet Server software. For a
 
     * `gem-home`: The location where JRuby looks for gems. It is also used by the `puppetserver gem` command line tool. If nothing is specified, JRuby uses the Puppet default `/opt/puppetlabs/server/data/puppetserver/jruby-gems`.
 
+    * `gem-path`: The complete "GEM_PATH" for jruby.  If set, it should include the `gem-home` directory, as well as any other directories that gems can be loaded from (including the vendored gems directory for gems that ship with puppetserver).  The default value is  `/opt/puppetlabs/server/data/puppetserver/jruby-gems:/opt/puppetlabs/server/data/puppetserver/vendored-jruby-gems`.
+
     * `environment-vars:` Optional. A map of environment variables which are made visible to Ruby code running within JRuby, for example, via the Ruby `ENV` class.
 
         By default, the only environment variables whose values are set into JRuby from the shell are `HOME` and `PATH`.
@@ -102,6 +104,7 @@ The `puppetserver.conf` file contains settings for Puppet Server software. For a
 jruby-puppet: {
     ruby-load-path: [/opt/puppetlabs/puppet/lib/ruby/vendor_ruby]
     gem-home: /opt/puppetlabs/server/data/puppetserver/jruby-gems
+    gem-home: /opt/puppetlabs/server/data/puppetserver/jruby-gems:/opt/puppetlabs/server/data/puppetserver/vendored-jruby-gems
     environment-vars: { "FOO" : ${FOO}
                         "LANG" : "de_DE.UTF-8" }
     master-conf-dir: /etc/puppetlabs/puppet
--- a/resources/ext/build-scripts/install-vendored-gems.sh
+++ b/resources/ext/build-scripts/install-vendored-gems.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+
+set -x
+set -e
+
+SOURCE="${BASH_SOURCE[0]}"
+while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symlink
+  DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
+  SOURCE="$(readlink "$SOURCE")"
+  [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE" # if $SOURCE was a relative symlink, we need to resolve it relative to the path where the symlink file was located
+done
+DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
+
+cat "${DIR}/gem-list.txt"
+
+echo "jruby-puppet: { gem-home: ${DESTDIR}/opt/puppetlabs/server/data/puppetserver/vendored-jruby-gems }" > jruby.conf
+
+while read LINE
+do
+  gem_name=$(echo $LINE |awk '{print $1}')
+  gem_version=$(echo $LINE |awk '{print $2}')
+  java -cp puppet-server-release.jar clojure.main -m puppetlabs.puppetserver.cli.gem --config jruby.conf -- install ${gem_name} --version ${gem_version}
+done < "${DIR}/gem-list.txt"
--- a/resources/ext/config/conf.d/puppetserver.conf
+++ b/resources/ext/config/conf.d/puppetserver.conf
@@ -4,10 +4,14 @@ jruby-puppet: {
     # Puppet server expects to load Puppet from this location
     ruby-load-path: [/opt/puppetlabs/puppet/lib/ruby/vendor_ruby]
 
-    # This setting determines where JRuby will look for gems.  It is also
-    # used by the `puppetserver gem` command line tool.
+    # This setting determines where JRuby will install gems.  It is used for loading gems,
+    # and also by the `puppetserver gem` command line tool.
     gem-home: /opt/puppetlabs/server/data/puppetserver/jruby-gems
 
+    # This setting defines the complete "GEM_PATH" for jruby.  If set, it should include
+    # the gem-home directory as well as any other directories that gems can be loaded
+    # from (including the vendored gems directory for gems that ship with puppetserver)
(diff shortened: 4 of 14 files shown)


resources/ext/config/conf.d/puppetserver.conf
```diff
@@ -8,6 +8,9 @@ jruby-puppet: {
     # used by the `puppetserver gem` command line tool.
     gem-home: /opt/puppetlabs/server/data/puppetserver/jruby-gems
 
+    environment-vars: {
+      "GEM_PATH": "/opt/puppetlabs/server/data/puppetserver/vendored-jruby-gems:/opt/puppetlabs/server/data/puppetserver/jruby-gems"
+    }
```
cprice404: This is just a temporary hack, my intent is to expose this as a first-class `gem-path` setting.

resources/ext/ezbake.conf
```diff
@@ -8,6 +8,19 @@ ezbake: {
...
       redhat: { dependencies: ["puppet-agent >= 1.6.0"],
+                build-dependencies: ["%{open_jdk}"],
+                # Install some gems
+                install: [
+                  "pwd",
+                  "ls -l",
+                  "ls -l ext",
+                  "ls -l ext/build-scripts",
+                  "cat ./ext/build-scripts/foo.sh",
+                  "bash ./ext/build-scripts/foo.sh",
```
cprice404: This is just a bunch of debugging stuff, except for the last line, which calls our build script.  I've run this through ezbake and verified that the build script gets called correctly and produces the appropriate output on both EL and debian platforms.

resources/ext/ezbake.conf
```diff
@@ -8,6 +8,19 @@ ezbake: {
...
+                  "echo \\\"jruby-puppet: { gem-home: ${DESTDIR}/opt/puppetlabs/server/data/puppetserver/vendored-jruby-gems }\\\" > jruby.conf",
+                  "java -cp puppet-server-release.jar clojure.main -m puppetlabs.puppetserver.cli.gem --config jruby.conf -- install semantic_puppet --version 0.1.2",
+                  "java -cp puppet-server-release.jar clojure.main -m puppetlabs.puppetserver.cli.gem --config jruby.conf -- install hocon --version 1.1.2",
```
cprice404: These would go away and be moved into the build script.

documentation/config_file_puppetserver.markdown
```diff
@@ -102,6 +104,7 @@ The `puppetserver.conf` file contains settings for Puppet Server software. For a
 jruby-puppet: {
     ruby-load-path: [/opt/puppetlabs/puppet/lib/ruby/vendor_ruby]
     gem-home: /opt/puppetlabs/server/data/puppetserver/jruby-gems
+    gem-home: /opt/puppetlabs/server/data/puppetserver/jruby-gems:/opt/puppetlabs/server/data/puppetserver/vendored-jruby-gems
```
camlow325: Should this line be `gem-path`, not `gem-home`?

resources/ext/build-scripts/install-vendored-gems.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+set -x
+set -e
+
+SOURCE="${BASH_SOURCE[0]}"
+while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symlink
+  DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
```
camlow325: (whoa) (mindblown)

resources/ext/build-scripts/install-vendored-gems.sh
```diff
@@ -0,0 +1,23 @@
...
+  SOURCE="$(readlink "$SOURCE")"
+  [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE" # if $SOURCE was a relative symlink, we need to resolve it relative to the path where the symlink file was located
+done
+DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
+
+cat "${DIR}/gem-list.txt"
+
+echo "jruby-puppet: { gem-home: ${DESTDIR}/opt/puppetlabs/server/data/puppetserver/vendored-jruby-gems }" > jruby.conf
+
+while read LINE
```
camlow325: Do you not have to pipe the output from the `cat` command above into this in order to read it?  I suck at bash scripting...

documentation/config_file_puppetserver.markdown
```diff
@@ -26,6 +26,8 @@ The `puppetserver.conf` file contains settings for Puppet Server software. For a
 
     * `gem-home`: The location where JRuby looks for gems. It is also used by the `puppetserver gem` command line tool. If nothing is specified, JRuby uses the Puppet default `/opt/puppetlabs/server/data/puppetserver/jruby-gems`.
 
+    * `gem-path`: The complete "GEM_PATH" for jruby.  If set, it should include the `gem-home` directory, as well as any other directories that gems can be loaded from (including the vendored gems directory for gems that ship with puppetserver).  The default value is  `/opt/puppetlabs/server/data/puppetserver/jruby-gems:/opt/puppetlabs/server/data/puppetserver/vendored-jruby-gems`.
```
camlow325: Would it be better to say something like the following?

> The default value includes the value for the `gem-home` setting and `/opt/puppetlabs/server/data/puppetserver/vendored-jruby-gems`.  If the `gem-home` setting is not specified, the default value would be `/opt/puppetlabs/server/data/puppetserver/jruby-gems:/opt/puppetlabs/server/data/puppetserver/vendored-jruby-gems`.

Dunno, just wondered if it would be worth pointing out somehow that the first path derives from the `gem-home` setting.

documentation/config_file_puppetserver.markdown
cprice404: Yep, thanks.

resources/ext/build-scripts/install-vendored-gems.sh
cprice404: This is straight-up copy paste from googling "how to find the directory of the currently executing bash script".

resources/ext/build-scripts/install-vendored-gems.sh
cprice404: The `cat` is there just for debugging.  The `<` after the `done` below this is how the file gets to STDIN.

camlow325: Ah, whoops, missed that.  Makes sense.

cprice404:
This is blocked on all the PRs relating to `lein-parent`, starting with puppetlabs/trapperkeeper#257

dev/puppetserver.conf.sample
```diff
@@ -44,10 +44,15 @@ jruby-puppet: {
...
-    # This setting determines where JRuby will look for gems.  It is also
-    # used by the `puppetserver gem` command line tool.
+    # This setting determines where JRuby will install gems.  It is used for loading gems,
+    # and also by the `puppetserver gem` command line tool.
     gem-home: ./target/jruby-gems
 
+    # This setting defines the complete "GEM_PATH" for jruby.  If set, it should include
+    # the gem-home directory as well as any other directories that gems can be loaded
+    # from (including the vendored gems directory for gems that ship with puppetserver)
+    gem-path: ./target/vendored-jruby-gems
```
justinstoller: Based on the comment I would expect this value to be `./target/jruby-gems` and potentially others.

resources/ext/config/conf.d/puppetserver.conf
```diff
@@ -4,10 +4,14 @@ jruby-puppet: {
...
-    # This setting determines where JRuby will look for gems.  It is also
-    # used by the `puppetserver gem` command line tool.
+    # This setting determines where JRuby will install gems.  It is used for loading gems,
+    # and also by the `puppetserver gem` command line tool.
     gem-home: /opt/puppetlabs/server/data/puppetserver/jruby-gems
 
+    # This setting defines the complete "GEM_PATH" for jruby.  If set, it should include
+    # the gem-home directory as well as any other directories that gems can be loaded
+    # from (including the vendored gems directory for gems that ship with puppetserver)
+    gem-path: "/opt/puppetlabs/server/data/puppetserver/vendored-jruby-gems:/opt/puppetlabs/server/data/puppetserver/jruby-gems"
```
justinstoller: Does our implementations of hocon resolve substitutions? If so should we set this to be "${jruby-puppet.gem-home}:/opt/..../vendored-jruby-gems"?

Also, I think this is the opposite order (vendored in front of gem-home) than what we document elsewhere or what is implemented as defaults in clojure.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -194,15 +203,17 @@
...
+                                                          (:ruby-load-path jruby-config)))
+                                  initialize-gem-path)]
     (jruby-core/initialize-config (assoc modified-jruby-config :lifecycle lifecycle-fns))))
 
 (schema/defn initialize-and-create-jruby-config :- jruby-schemas/JRubyConfig
   "Handles the initialization of the jruby-puppet config for the purpose of returning a
   jruby config. This function will only use the :jruby-puppet and :http-client sections
   from raw-config. If values are not provided, everything in :http-client :jruby-puppet
-  will be given default values, except for :ruby-load-path and :gem-home in the
+  will be given default values, except for :ruby-load-path, :gem-home, and :gem-path in the
```
justinstoller: I'm confused by this line because it seems like we *are* assoc'ing in default values for `:ruby-load-path` and `:gem-path` on lines 206-209 in `create-jruby-config` called by this function (line 250). Am I reading that wrong?

justinstoller:
Seems good save for some defaults/docs updates!

dev/puppetserver.conf.sample
cprice404: yep, I flubbed that, thanks.

resources/ext/config/conf.d/puppetserver.conf
cprice404: Yeah, I wasn't  totally sure I wanted to introduce our first use of HOCON interpolation just for this, but I think you're right, it's probably worth it.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
cprice404: Yeah it may just be that the wording is confusing here.  I think what it's trying to say is that the `jruby-utils` library will merge in defaults for everything except for those.  I'll see if I can't figure out a way to tweak the wording.

cprice404:
Added a commit addressing @justinstoller 's feedback.  I'm currently doing a local ezbake build to make sure the hocon interpolation will work properly in the packages.

resources/ext/config/conf.d/puppetserver.conf
```diff
@@ -4,10 +4,14 @@ jruby-puppet: {
...
-    # This setting determines where JRuby will look for gems.  It is also
-    # used by the `puppetserver gem` command line tool.
+    # This setting determines where JRuby will install gems.  It is used for loading gems,
+    # and also by the `puppetserver gem` command line tool.
     gem-home: /opt/puppetlabs/server/data/puppetserver/jruby-gems
 
+    # This setting defines the complete "GEM_PATH" for jruby.  If set, it should include
+    # the gem-home directory as well as any other directories that gems can be loaded
+    # from (including the vendored gems directory for gems that ship with puppetserver)
+    gem-path: ${jruby-puppet.gem-home}":/opt/puppetlabs/server/data/puppetserver/vendored-jruby-gems"
```
cprice404: So one thing I just realized here is that I set this up as a single string value, because that's what the underlying JRuby setup is going to want.  But that seems kind of dumb and inconsistent since we use an array for ruby-load-path above.  Seems like it would be much better UX for me to change this to an array in the config file, and then munge it into a string programatically before I send it to JRuby.  Anyone disagree?

cprice404:
Added another commit that converts the setting in the config file over to be an array instead of a colon-delimited string.

I've also now done ezbake builds and local beaker smoke tests to make sure this works.  It does require the newly released 1.1.3 version of the hocon gem because the beaker code was tickling a bug having to do with interpolation/variable substitution in hocon strings.

cprice404:
OK I think I've talked this through with all the interested parties, and done ezbake and manual testing, etc.  I think this is clear to merge if the last round of code changes look ok to everyone.

justinstoller:
I've been trying to use the new rake tasks to test locally but ran into issue (I don't believe they're related to this PR though)

👍

cprice404:
@justinstoller cool, thanks.  Would be interested to hear what your issues with the rake tasks are, and hopefully I can fix them up, but we can do that on the other PR.

documentation/config_file_puppetserver.markdown
camlow325: Cosmetic - now that we're using the array syntax, probably good to update this to something like:

> The default value is `[/opt/puppetlabs/server/data/puppetserver/jruby-gems, /opt/puppetlabs/server/data/puppetserver/vendored-jruby-gems]`.

documentation/config_file_puppetserver.markdown
camlow325: Still should update this to `gem-path`, along with the new array-syntax for the value.

test/unit/puppetlabs/services/jruby/jruby_puppet_core_test.clj
```diff
@@ -14,6 +14,7 @@
    {:name "puppetserver", :update-server-url "http://localhost:11111"},
    :jruby-puppet
    {:gem-home "./target/jruby-gem-home",
+    :gem-path "./target/jruby-gem-home:./target/vendored-jruby-gems"
```
camlow325: Hmm, I was going to suggest changing the value for `:gem-path` into an array to match elsewhere.  I can't find any code that actually references this function, though.  So maybe `min-config` should just go away completely?

camlow325:
I left a couple of other cosmetic comments about docs and such but still think this looks good.  I'm going to go ahead and merge it.

PR #1228, opened 2016-10-03
(QENG-4425) Pin docker-api gem version

cprice404:
docker-api version 1.32.0 introduced a dependency on ruby >= 2.0.
This commit pins us to the older docker version in order to avoid
that dependency until we no longer have ruby 1.9.3 in our test
matrices.


camlow325:
👍

PR #1229, opened 2016-10-03
(MAINT) merge from 1.x to stable 2016-10-03

cprice404:
@haus I don't need your custom tmpdir change in `stable` / `master`, do I?


camlow325:
👍

haus:
@cprice404 i don't think so

cprice404:
I'm going to count that as a +1.

PR #1230, opened 2016-10-03
(MAINT) merge stable -> master, 2016-10-03

cprice404:


cprice404:
@camlow325 @gguillotte mostly docs changes here - were we expecting this diff on merge up to master from stable?  e.g. we added some docs directly to stable before the last release?

camlow325:
👍 Yeah, I think we added the docs for 2.6.0 to stable right around the time of the release and I hadn't done a merge back up to master after that, sorry.  In any event, these seem fine to me.

camlow325:
:shipit:

PR #1231, opened 2016-10-04
(SERVER-1571) ensure environment passed to clj shellutils

cprice404:
A recent refactor resulted in a subtle change in behavior in the
clojure shellutils wrapper, wherein if you did not pass an
explicit list of environment variables to the command, you would
end up with an empty map of environment variables for the shell
command.  Prior to that refactor, you would instead have inherited
all of the environment variables from the java process, which is
the desired behavior.  This change broke some acceptance tests which
were relying on having access to environment variables such as PATH.

This commit changes the environment variable behavior back to the
previous setup, where the java environment variables will be passed
through to the shell command unless you explicitly override the
environment.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/shell_utils.clj
+++ b/src/clj/puppetlabs/puppetserver/shell_utils.clj
@@ -39,9 +39,11 @@
 
 (schema/defn ^:always-validate java-exe-options :- ShellUtils$ExecutionOptions
   [{:keys [env in]} :- ExecutionOptions]
-  (doto (ShellUtils$ExecutionOptions.)
-    (.setEnv (if env (ks/mapkeys name env) {}))
-    (.setStdin in)))
+  (let [exe-options (ShellUtils$ExecutionOptions.)]
+    (.setStdin exe-options in)
+    (when env
+      (.setEnv exe-options (ks/mapkeys name env)))
+    exe-options))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Public
--- a/dev-resources/puppetlabs/puppetserver/shell_utils_test/env
+++ b/dev-resources/puppetlabs/puppetserver/shell_utils_test/env
@@ -1,0 +1,3 @@
+#!/usr/bin/env sh
+
+env
--- a/test/unit/puppetlabs/puppetserver/shell_utils_test.clj
+++ b/test/unit/puppetlabs/puppetserver/shell_utils_test.clj
@@ -2,7 +2,8 @@
   (:require [clojure.test :refer :all]
             [puppetlabs.puppetserver.shell-utils :as sh-utils]
             [puppetlabs.kitchensink.core :as ks]
-            [clojure.string :as str])
+            [clojure.string :as str]
+            [clojure.set :as set])
   (:import (java.io ByteArrayInputStream)))
 
 (def test-resources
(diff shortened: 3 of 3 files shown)


cprice404:
I believe that this is the culprit of the current failing acceptance tests on the puppetserver master branch.  Doing a local  ezbake build / beaker run to confirm now.

src/clj/puppetlabs/puppetserver/shell_utils.clj
```diff
@@ -39,9 +39,11 @@
 
 (schema/defn ^:always-validate java-exe-options :- ShellUtils$ExecutionOptions
   [{:keys [env in]} :- ExecutionOptions]
-  (doto (ShellUtils$ExecutionOptions.)
-    (.setEnv (if env (ks/mapkeys name env) {}))
-    (.setStdin in)))
+  (let [exe-options (ShellUtils$ExecutionOptions.)]
+    (.setStdin exe-options in)
+    (when env
+      (.setEnv exe-options (ks/mapkeys name env)))
```
camlow325: Ah, because before this would have been passed through to Java as nil/null when no environment variables were set but now it would be an empty map, which would cause the Apache library to spawn the command with an empty environment?

Just curious, what does the ExecutionStub do when running under an MRI Ruby master?  Did it allow all environment variables from the original process to be passed through, only allow a whitelist of some common ones to be passed through, or something else?

cprice404: Ruby is different because we are running from JRuby and we're intentionally filtering environment variables for the JRubies.  I don't see any reason to mess with the behavior there since we haven't had any issues filed about it.

Yes, the difference in this Clojure code is about passing `nil` vs. passing an empty map.

camlow325: > Ruby is different because we are running from JRuby and we're intentionally filtering environment variables for the JRubies.

That's not what my point was.  I was interested in knowing if MRI Ruby Puppet does any of its own filtering - or has filtering done on its behalf - of environment variables for a process that it launches via its `ExecutionStub`.  I agree we haven't had any reports about the current environment variable behaviors with Puppet Server's current `ExecutionStub`, though, so I'm good with not looking into this further.

cprice404:
Finished local ezbake/beaker runs - the failing `command_scripts.rb` test passed with these changes.

camlow325:
Hmm, this failed Travis in a couple of tests on one cell.  These failures didn't look familiar to me so thought you might want to look at them.  Not clear to me whether these might be related to these tests without looking further although I'm guessing they aren't:

```
lein test :only puppetlabs.services.request-handler.request-handler-core-test/test-jruby-pool-not-full-during-code-id-generation

FAIL in (test-jruby-pool-not-full-during-code-id-generation) (request_handler_core_test.clj:326)
A jruby instance is held while code id is generated
  expected: (deref first-promise 10000 false)
  actual: (not (deref #object[clojure.core$promise$reify__6779 0x62ef502b {:status :pending, :val nil}] 10000 false))

lein test :only puppetlabs.services.request-handler.request-handler-core-test/test-jruby-pool-not-full-during-code-id-generation

FAIL in (test-jruby-pool-not-full-during-code-id-generation) (request_handler_core_test.clj:330)
A jruby instance is held while code id is generated
  expected: (zero? (jruby-protocol/free-instance-count jruby-service))
  actual: (not (zero? 1))
```

rlinehan:
When I run `lein test puppetlabs.puppetserver.shell-utils-test` with this PR I get:

```
lein test :only puppetlabs.puppetserver.shell-utils-test/sets-env-correctly

FAIL in (sets-env-correctly) (shell_utils_test.clj:68)
sets environment variables correctly
expected: (= 2 (count env))
  actual: (not (= 2 4))

lein test :only puppetlabs.puppetserver.shell-utils-test/sets-env-correctly

FAIL in (sets-env-correctly) (shell_utils_test.clj:71)
sets environment variables correctly
expected: (= #{"FOO" "PWD"} (ks/keyset env))
  actual: (not (= #{"FOO" "PWD"} #{"FOO" "PWD" "_" "SHLVL"}))
2016-10-04 14:07:33,877 WARN  [main] [c.p.p.ShellUtils] Executed an external process which logged to STDERR: bar


Ran 11 tests containing 17 assertions.
2 failures, 0 errors.
Tests failed.
```

I'm guessing this is because I usually run my tests/do everything in a tmux session?

cprice404:
@rlinehan hmm, crap.  I don't suppose you would be able to throw an extra commit on top of this that makes the tests tolerate that?  Or I could take a stab at it based on your error output and then have you see if they pass for you?

rlinehan:
Sure, I'll try.

rlinehan:
👍 works on my laptop now! should we wait for travis though?

PR #1233, opened 2016-10-10
(SERVER-1589) Change logback logs to 200MB maxFileSize and gz format

camlow325:
This commit changes the logback log rolling policies to use gz instead
of zip for the archive format and 200MB instead of 10MB as the maximum
size for a log file.

### DIFF
--- a/documentation/config_logging_advanced.markdown
+++ b/documentation/config_logging_advanced.markdown
@@ -28,7 +28,7 @@ The following examples show how to configure Logback for:
 
 Adjust the example configuration settings to suit your needs.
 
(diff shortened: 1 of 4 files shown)


camlow325:
Failed Travis run on OracleJDK 8 only with an apparent test hang:

```
lein test puppetlabs.services.jruby.jruby-puppet-pool-int-test

2016-10-10 15:57:08,286 WARN  [async-dispatch-40] [o.e.j.s.h.ContextHandler] Empty contextPath
2016-10-10 15:58:16,849 WARN  [async-dispatch-48] [o.e.j.s.h.ContextHandler] Empty contextPath

No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
```
Seems highly unlikely for this to be related to the changes on this PR so rekicking the Travis run...

PR #1235, opened 2016-10-10
(SERVER-1591) Work around NPE w/ezbake & managed deps

cprice404:
This commit does the following:

* Updates to lein-ezbake 1.1.1 which fixes bugs preventing managed
  dependencies from working.
* Adds in explicit `nil` versions to the dependencies listed in
  our ezbake profile, because of technomancy/leiningen#2216

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -102,16 +102,28 @@
 
              :testutils {:source-paths ^:replace ["test/unit" "test/integration"]}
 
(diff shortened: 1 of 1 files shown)


camlow325:
👍 I did a smoke test with a build of OSS puppetserver from this PR. The ezbake build was successful and the beaker tests I ran against the package which do a couple of agent runs against the server passed. I think this is good to merge then.

PR #1236, opened 2016-10-12
(MAINT) Bake'n'beak

cprice404:
This commit adds some rake tasks that make it easier to run smoke tests on a specific package version, or to do an ezbake build and then run smoke tests against that.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,3 +1,5 @@
+require 'open3'
+
 PROJECT_ROOT = File.dirname(__FILE__)
 ACCEPTANCE_ROOT = ENV['ACCEPTANCE_ROOT'] ||
   File.join(PROJECT_ROOT, 'acceptance')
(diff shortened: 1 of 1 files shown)


justinstoller:
I'm not really familiar with the ezbake process but when I ran `test:acceptance:bakeNbeak` it failed with:
```
Begin acceptance/suites/pre_suite/foss/30_install_dev_repos.rb

* Setup Puppet Server repositories.
  Fetching: http://builds.delivery.puppetlabs.net/puppetserver/2.6.1.master.SNAPSHOT.2016.10.13T1024/repo_configs/rpm/pl-puppetserver-2.6.1.master.SNAPSHOT.2016.10.13T1024-el-7-x86_64.repo
    and saving to tmp/repo_configs/el/pl-puppetserver-2.6.1.master.SNAPSHOT.2016.10.13T1024-el-7-x86_64.repo
  RuntimeError: Failed to fetch_remote_file 'http://builds.delivery.puppetlabs.net/puppetserver/2.6.1.master.SNAPSHOT.2016.10.13T1024/repo_configs/rpm/pl-puppetserver-2.6.1.master.SNAPSHOT.2016.10.13T1024
-el-7-x86_64.repo' (404 Not Found)
```

Looking at http://builds.delivery.puppetlabs.net/puppetserver/2.6.1.master.SNAPSHOT.2016.10.13T1024/
there's an artifacts directory but not a repo_configs directory. Is that supposed to happen after the ezbake job automatically or is there a step that's missing?

cprice404:
Hmm, not sure; this worked for me when I ran it several times yesterday but I'll try it again today and see if I hit the same issue.

cprice404:
@justinstoller I presume you were running via `bundle exec`, right?

justinstoller:
Yeah, I was running using bundle exec, though I was running on a branch that was master with #1227 merged onto it and this merged on to that. I'll try re-running just this branch. Though the problem seemed that the repo_configs weren't built for the version I ezbaked which didn't seem like something this should affect.

I assumed that if I wanted to understand the problem better I would have to go through exbake and into the the packaging tool. Which sounded like descending several levels into Hell. So I thought I'd ask here first incase my problem rang any bells. I'll go ahead and investigate further since it sounds like it doesn't.

cprice404:
Actually I just tried it on this branch and it doesn't work well at all.  Those rake tasks rely on some subtle changes included in the #1227 PR, so probably best to just ignore this PR until that one is merged and I've rebased this one.

cprice404:
I rebased this, and then tested bakeNbeak locally.  Worked fine for me.   I think this is ready for review now.

Rakefile
```diff
@@ -117,6 +156,48 @@ namespace :test do
...
+      Open3.popen3("lein with-profile ezbake ezbake build 2>&1") do |stdin, stdout, stderr, thread|
+        # sleep 5
+        # puts "STDOUT IS: #{stdout}"
+        stdout.each do |line|
+          if match = line.match(%r|^Your packages will be available at http://builds.delivery.puppetlabs.net/puppetserver/(.*)$|)
+            package_version = match[1]
+          end
+          puts line
+        end
+        puts "PACKAGE VERSION IS #{package_version}"
```
justinstoller: It'd be great to abort the rake task if this job fails. When it has failed for me locally I don't find out until Beaker fails to install the dev repo.

justinstoller:
If it's possible to limit the packages we build with ezbake I think it'd be great to do that. So far every time I've tried to run this patch it has failed building some random Deb even though the smoke test that follows is hardcoded to use centos 7.

Rakefile
cprice404: Good call, but I'm not sure when I will have time to work on this again, so maybe somebody could just hack that in after we merge, next time they run into it?

cprice404:
Re: limiting to certain distros - that would be really nice.  TBH I have no idea how to do that with the underlying packaging tools, or whether or not ezbake has support for interacting with the packaging tools in that way.  Maybe another thing for us to hack on separately?

PR #1238, opened 2016-10-14
(QENG-4310) Upgrade to beaker-hostgenerator.

waynr:


Gemfile
```diff
@@ -6,9 +6,7 @@ gem 'jira-ruby', :group => :development
 group :test do
   gem 'rspec'
   gem 'beaker', '~>1.21.0'
-  if ENV['GEM_SOURCE'] =~ /rubygems\.delivery\.puppetlabs\.net/
-    gem 'sqa-utils', '0.12.1'
-  end
+  gem "beaker-hostgenerator", *location_for(ENV['BEAKER_HOSTGENERATOR_VERSION'] || "~> 0.7")
```
nwolfe: Latest release is `0.8.0`, there shouldn't be any reason not to be on that

waynr: @nwolfe I just pulled the same version that I found from master branch on this repo. I'll leave it up to @camlow325 if it should be upgraded here. Other than the fact that the other puppetserver branches are using `0.7.0` I don't see any reason not to use `0.8.0` here.

nwolfe: Well we'll need `0.8.0`+ for CI.next support

camlow325: Doesn't matter to me either.  @nwolfe - think there's something important in `0.8.0` that we might want / need for Puppet Server eventually anyway?  If not, I'm good sticking with `0.7.~` since we're using it successfully on the other branches already.

camlow325:
:shipit:

PR #1241, opened 2016-10-14
Revert beaker-hostgenerator upgrade

waynr:
@rlinehan @camlow325 @cprice404 @nwolfe 

So here's a revert of the BHG work I submitted earlier. This is necessary because there is an implicit dependency between beaker-hostgenerator and beaker >=2 as @camlow325 suspected earlier today.

We shouldn't need to merge this up into stable because stable branch is on a modern version of beaker.

I'll work on making beaker-hostgenerator compatible with this branch of puppetserver and verify it works on my own fork of the repo before submitting another PR to revert this revert.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ gem 'jira-ruby', :group => :development
 group :test do
   gem 'rspec'
   gem 'beaker', '~>1.21.0'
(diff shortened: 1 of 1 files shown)


waynr:
Oh, just to be clear this should fix the acceptance test failure in the 1.x CI pipeline.

camlow325:
👍

PR #1245, opened 2016-10-18
(SERVER-1551) i18n string updates

camlow325:
This commit updates additional messages to use i18n, along with
including some of the i18n scaffolding such as the Makefile,
Makefile.i18n, and messages.pot files.


src/clj/puppetlabs/services/puppet_admin/puppet_admin_core.clj
```diff
@@ -132,9 +132,9 @@
       ;; of the 'wrap-with-trapperkeeper-or-client-whitelist-authorization'
       ;; function and replace with a line chaining the handler into a call to
       ;; 'authorization-fn'.
-      i18n/locale-negotiator
       (ringutils/wrap-with-trapperkeeper-or-client-whitelist-authorization
         authorization-fn
         path
         settings)
+      i18n/locale-negotiator
```
camlow325: I moved this so that any failure messages from tk-auth could be localized.

PR #1246, opened 2016-10-18
(SERVER-1583) Add missing cloud OID shortnames

justinstoller:
This adds eight shortnames for Puppet OIDs that have been added in the last year to the Ruby CA. They are all intended to be used with cloud infrastructure.


justinstoller:
I'm down to improve it, but I haven't been able to reproduce it yet. See comments in the associated ticket.

justinstoller:
I've been able to reproduce an error, but not the unfriendly error that needed improving. I'm off hand inclined to believe the error that we were seeing is an issue with kitchensink that this is only tangentially tickling. Unless @camlow325 disagrees I'd say we should merge it.

camlow325:
Good enough for me, :shipit:

PR #1248, opened 2016-10-19
(SERVER-1599) Enable Telemetry/Update check opt-out

briancain:
Prior to this commit, there was no way to opt-out of update checking and
telemetry data collection. This commit updates puppetserver to look for
an opt-out file on disk before checking for updates and sending
telemetry data.

### DIFF
--- a/src/clj/puppetlabs/services/master/master_service.clj
+++ b/src/clj/puppetlabs/services/master/master_service.clj
@@ -40,9 +40,13 @@
          environment-class-cache-enabled (get-in config
                                                  [:jruby-puppet
                                                   :environment-class-cache-enabled]
-                                                 false)]
-     (interspaced checkin-interval-millis
-                  (fn [] (version-check/check-for-updates! {:product-name product-name} update-server-url)))
+                                                 false)
+         check-for-updates (get-in config [:product :check-for-updates])]
+     (if (or (nil? check-for-updates) check-for-updates)
+       (interspaced checkin-interval-millis
+                    (fn [] (version-check/check-for-updates!
+                             {:product-name product-name} update-server-url)))
+       (log/info (i18n/trs "Not checking for updates - opt-out setting exists")))
 
      (retrieve-ca-cert! localcacert)
      (retrieve-ca-crl! hostcrl)
--- a/locales/messages.pot
+++ b/locales/messages.pot
@@ -450,6 +450,10 @@ msgid "Route not found for service {0}"
 msgstr ""
 
 #: src/clj/puppetlabs/services/master/master_service.clj
+msgid "Not checking for updates - opt-out setting exists"
+msgstr ""
+
+#: src/clj/puppetlabs/services/master/master_service.clj
 msgid "Master Service adding ring handlers"
 msgstr ""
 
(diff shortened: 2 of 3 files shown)


briancain:
Oops, will push out a fix for the tests here soon.

briancain:
Hmm, I'm not entirely sure how to do that without writing a beaker test of some sort. The opt-out file is in a place that requires root to be created, which means lein would have to run as root.

haus:
@briancain you could push the opt-out check into a function that takes a file as a param, or make the opt-out file config driven with a sane default. both of those would make it testable.

briancain:
@haus - I think I'd probably prefer a config setting. That's currently how FOSS puppetdb works. Only reason why I went with the file approach is because that's how pe-puppetserver does it. Any objections to a config file setting? And if so, probably under `:puppetserver` as `:opt-out`, defaulting to false if not present in the config?

camlow325:
@briancain I like the config setting approach.  Are there any plans to have additional analytics-specific configuration in the future?  If so, would creating an `analytics` section or some such be worthwhile?  That might also be good in the event that, in the future, we were to create an OSS-specific, reusable analytics service that were included in both the puppetdb and puppetserver stacks.

briancain:
@camlow325 - nothing in our backlog. I think for now it'll just be an on/off approach.

briancain:
I wonder if it makes sense to put it in the `:product` key, since the other values in that are all related to analytics? @camlow325

briancain:
I think placing it under `:product` makes sense since it fits with the idea of having our own analytics config map. Also pushed up a test that sets opt-out to false and ensures that nothing is actually sent.

camlow325:
Ah, good call.  `:product` does sound like a good place for that since the other items in that map are already related.

src/clj/puppetlabs/services/master/master_service.clj
```diff
@@ -7,6 +7,7 @@
             [puppetlabs.comidi :as comidi]
             [puppetlabs.dujour.version-check :as version-check]
             [puppetlabs.services.protocols.master :as master]
+            [me.raynes.fs :as fs]
```
haus: i don't think `fs` is required anymore.

briancain:
@haus removed the unneeded require!

src/clj/puppetlabs/services/master/master_service.clj
```diff
@@ -40,9 +40,13 @@
...
-                                                 false)]
-     (interspaced checkin-interval-millis
-                  (fn [] (version-check/check-for-updates! {:product-name product-name} update-server-url)))
+                                                 false)
+         check-for-updates (get-in config [:product :check-for-updates])]
+     (if (or (nil? check-for-updates) check-for-updates)
+       (interspaced checkin-interval-millis
+                    (fn [] (version-check/check-for-updates!
+                             {:product-name product-name} update-server-url)))
+       (log/info (trs "Not checking for updates - opt-out setting exists")))
```
briancain: here's the changes @pcarlisle

pcarlisle:
cool

briancain:
This should be good to go now once travis finishes.

briancain:
@camlow325  - No immediate plans for updating how PE users opt out, however I would really prefer this method over our current file method.

camlow325:
Travis tests passed and sounds like others are good with this so :shipit:

PR #1250, opened 2016-10-20
(SERVER-1585) Expose JRuby `compat-version` as an experimental option

camlow325:
This commit exposes the JRuby MRI `compat-version` as an experimental
configurable option.  The underlying work which enables this is all in
a new version of jruby-utils, 0.4.0, and this commit bumps the
dependency to this new version.

### DIFF
--- a/documentation/config_file_puppetserver.markdown
+++ b/documentation/config_file_puppetserver.markdown
@@ -70,6 +70,8 @@ The `puppetserver.conf` file contains settings for Puppet Server software. For a
 
         For more information, see the [`environment_classes` API documentation][].
 
+    * `compat-version`: Optional, experimental.  Used to control the MRI compatibility version under which JRuby runs.  The allowed values are "1.9" and "2.0".  The default value is "1.9".
+
     * `compile-mode`: Optional, experimental. Used to control JRuby's "CompileMode", which may improve performance. The default value is `off`, which is the most conservative value. A value of `jit` enables JRuby's "just-in-time" compilation of Ruby code. A value of `force` causes JRuby to attempt to pre-compile all Ruby code.
 
 * The `profiler` settings configure profiling:
--- a/project.clj
+++ b/project.clj
@@ -38,7 +38,7 @@
                  ;; in different versions of the three different logback artifacts
                  [net.logstash.logback/logstash-logback-encoder]
 
-                 [puppetlabs/jruby-utils "0.3.0"]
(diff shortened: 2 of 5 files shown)


camlow325:
I did some manual functional testing of this via `lein ruby` and `lein run` to see that the `compat-version` from the puppetserver.conf file was being honored.  For testing via `lein run` of the whole stack, I wrote a simple Puppet parser function which returns the value of `RUBY_VERSION` and echoed that back via a `notify` resource in a catalog for an agent run.  I saw that `RUBY_VERSION` returned `2.0` vs. `1.9.3` per the `compat-version` configured in my puppetserver.conf file, using `1.9.3` by default.

cprice404:
@camlow325 would it be worth doing a `lein gem install` (or an ezbake build and then a `puppetserver gem install`) on the gem in the original ticket that this came up in, and confirm that that gem can at least be installed?  I don't know that we need to do much vetting of whether the gem works 100% properly since we are labeling this as experimental, but at least checking to confirm it will get past the install phase?

camlow325:
@cprice404 I have been able to install the `google-api-client` gem (even with the Ruby > 2.0) dependency via lein gem with Puppet Server when running under JRuby 1.7 with compat mode 2.0.  I didn't try to do anything with it but it at least installs.  Given that the google-api-ruby-client MIGRATING doc also indicates that 2.0 compat mode is supported with JRuby 1.7, it seems promising that this will work.

cprice404:
@camlow325 awesome, thanks.

camlow325:
Yeah, it does get validated by schema in jruby-utils at a lower level.  In puppetserver, we just handle it as a raw map.  I suppose it might be good at some point to do a second-level of schema validation in puppetserver directly.  @rlinehan and @jpinsonault may remember more about the reasons why we ended up with the design that we did for this.

jpinsonault:
Off the top of my head I want to say that it's because this setting goes straight to the jruby pool manager service from the jruby-utils repo and is used directly in that service, not by any services in the puppetserver services.

jpinsonault:
Figure I'll wait for travis, unless there's a hurry

camlow325:
I also ran both the Puppet Server smoke tests and the full Ruby Puppet acceptance tests on a CentOS 7 VM (master/agent on the same node) with `compat-version` set to "2.0" in the puppetserver.conf file.  All of the tests passed.

PR #1251, opened 2016-10-21
(SERVER-528) Generate the master cert if keys already present at startup

camlow325:
Previously, if a master public key were present but the master cert were
not at startup, Puppet Server would error out at startup with a "partial
state error" message.  This was problematic if an agent were run on the
master before the master had been started for the first time since the
initial agent run would create a public and private key for the master.

This PR changes the master cert initialization logic to generate the
cert from a public key already on disk, if one exists.  If the public
and private keys do not exist at startup, Puppet Server will still
create a public and private key for the master before generating the
cert, as has been done in the past.


camlow325:
Rebased this on latest HEAD of the master branch.

PR #1252, opened 2016-10-21
(SERVER-1601) Set umask for gem subcommand to 0022

camlow325:
The bump to EZBake 1.0+ causes cli subcommands to run with a default
umask of 027.  Previously, gems installed via the puppetserver gem
subcommand would have the world read bits for their gemspec files masked
out by this umask even though the gemspec files could be owned by the
root user.  This could cause puppetserver service, running as user
'puppet', to fail to load gems properly.

This commit sets the umask for the gem subcommand to 0022.  This should
allow gemspec files to be world-readable and, therefore, loadable by the
puppetserver service.

### DIFF
--- a/acceptance/suites/tests/010-puppetserver-cli/subcommand/gem.rb
+++ b/acceptance/suites/tests/010-puppetserver-cli/subcommand/gem.rb
@@ -46,6 +46,22 @@ gems.each do |gem_name, gem_version|
     assert(/^#{gem_name}/.match(stdout), "#{gem_name} was not found after installation.")
   end
 
+  if gem_name == 'excon'
+    step "SERVER-1601: Validate use of excon gem successful as puppet user" do
+      if options[:type] == 'pe'
+        runuser = 'pe-puppet'
+      else
+        runuser = 'puppet'
+      end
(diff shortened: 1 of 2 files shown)


acceptance/suites/tests/010-puppetserver-cli/subcommand/gem.rb
```diff
@@ -46,6 +46,22 @@
...
+        runuser = 'puppet'
+      end
+      on(master, "su #{runuser} -s /bin/bash -c "\
+        "'/opt/puppetlabs/bin/puppetserver ruby "\
+        "-rexcon -e \"puts Excon::VERSION\"'") do
+        assert_equal(gems['excon'], stdout.strip,
+                     "Unexpected output for excon version")
+      end
+    end
+  end
```
cprice404: Would this test have failed w/o the umask change?

camlow325: Yes, it would.  You'd see an exception/failure to load the "excon" gem.  Thought this test would be useful for future regression testing along with the fix.

PR #1253, opened 2016-10-21
(SERVER-1600) add gettext-related gems required for ruby i18n

cprice404:
This PR has two commits:

* First one just adds a fail fast to bakeNBeak so that it won't try to run the beaker tests if the packaging job fails.
* Second one adds the gems that @branan suggested that we will need for the puppet-agent ruby code to be able to handle i18n in the server.

I was able to build one set of packages and manually check the gem list via `puppetserver gem list` - it looked fine.  Most of my package runs are failing due to mirror/network issues, though.


branan:
These look right to me :+1:

camlow325:
:shipit:

PR #1254, opened 2016-10-24
(SERVER-1545) Generate multi-RDN issuer correctly for new certificate

camlow325:
Previously, if the CA certificate's subject had multiple RDNs in it --
e.g., O=Puppet, CN=mycn -- the order in which the RDNs would be inserted
into a client certificate generated from the CA would be reversed --
e.g., CN=mycn, O=Puppet.  This would cause subsequent attempts to
validate the certificate against the CA during an SSL handshake to fail.

This commit, when generating client certificates,  uses a new helper in
jvm-ssl-utils, get-subject-from-x509-certificate, which extracts the
CA subject from an X509Certificate with the RDNs in the exact order as
they are encoded into the PEM.  This allows the generated client
certificate to be validated properly against the CA certificate during
SSL handshakes.


camlow325:
This has a snapshot reference to jvm-ssl-utils and so cannot be merged yet.  It is dependent upon a new helper function defined in https://github.com/puppetlabs/jvm-ssl-utils/pull/68.  When the related PR is merged and new jvm-ssl-utils released, I'll update this PR to use the new jvm-ssl-utils version in place of the snapshot.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -308,13 +308,7 @@
            (i18n/trs "Missing:")
            (str/join "\n" missing-files))))
 
-; TODO - PE-5529 - these should be moved to jvm-c-a.
-(schema/defn get-subject :- schema/Str
-  [cert :- Certificate]
-  (-> cert
-      (.getSubjectX500Principal)
-      (.getName)))
```
camlow325: The problem with using `X500Principal.getName()` for this is that the format in which the method returns the subject does not preserve the original ordering of the RDNs as encoded into DER/ASN.1 in the certificate.  The `X500Principal` object does, however, retain a copy of the original encoded bytes from the certificate, available via a call to `X500Principal.getEncoded()`.  BouncyCastle has a helper for converting the DER/ASN.1 encoded bytes back into a string.  It didn't seem appropriate to call into the BouncyCastle helper directly from Puppet Server code.  Instead, then, I wrote a helper in jvm-ssl-utils which uses `X500Principal.getEncoded` and BouncyCastle in combination to return a subject from the `X509Certificate` which has the order of the RDNs preserved from how they were encoded into the source PEM file for the `X509Certificate`.  Throughout the rest of this PR, I changed Puppet Server code to call into the new jvm-ssl-utils helper.

project.clj
```diff
@@ -44,7 +44,7 @@
                  [puppetlabs/trapperkeeper-scheduler]
                  [puppetlabs/trapperkeeper-status]
                  [puppetlabs/kitchensink]
-                 [puppetlabs/ssl-utils]
+                 [puppetlabs/ssl-utils "0.8.2-SNAPSHOT"]
```
cprice404: When we have a non-SNAPSHOT version ready, do you want to override the value from lein-parent here, or update the parent?  Jon updated a few deps in the parent recently so it might be nice to go ahead and rev it?

camlow325: Updating the clj-parent dep sounds good to me.  jvm-ssl-utils doesn't get revved that often.

camlow325:
I rebased this at `e20f22c` and bumped the `clj-parent` dependency to 0.1.7 to pick up the new `jvm-ssl-utils` dependency, 0.8.2.  Assuming this passes Travis now, I think this should be good to go for merging.

jpinsonault:
👍

PR #1255, opened 2016-10-24
(SERVER-1582) Add some service reload docs

camlow325:
This commit includes some documentation updates related to the service
reload feature.

### DIFF
--- a/documentation/restarting.markdown
+++ b/documentation/restarting.markdown
@@ -19,7 +19,7 @@ There are several ways to send a HUP signal to the Puppet Server process, but th
 
 Starting in version 2.7.0, you can also reload Puppet Server by running the "reload" action via the operating system's service framework.  This is analogous to sending a hangup signal but with the benefit of having the "reload" command pause until the server has been completely reloaded, similar to how the "restart" command pauses until the service process has been fully restarted.
 
-The following commands could be used to trigger a service reload.
+Use the following commands to reload Puppet Server.
 
 All current OS distributions:
 
@@ -39,7 +39,7 @@ There are three ways to trigger your Puppet Server environment to refresh and pi
 
 > **Note:** Changes to Puppet Server's [logging configuration in `logback.xml`][logback.xml] don't require a server reload or restart. Puppet Server recognizes and applies them automatically, though it can take a minute or so for this to happen.
 
-### Changes applied after a JRuby pool flush, service reload, or full Server restart
+### Changes applied after a JRuby pool flush, HUP signal, service reload, or full Server restart
 
 * Changes to your `hiera.yaml` file to change your [Hiera][] configuration.
 * [Installation or removal of gems][gems] for Puppet Server by `puppetserver gem`.
@@ -49,7 +49,7 @@ There are three ways to trigger your Puppet Server environment to refresh and pi
   [Admin API endpoint for flushing the environment cache](./admin-api/v1/environment-cache.markdown).
 * Changes to the CA CRL file.  For example, a `puppet cert clean`
 
-### Changes applied after a service reload or full Server restart
+### Changes applied after a HUP signal, service reload, or full Server restart
 
 * Changes to Puppet Server [configuration files](./configuration.markdown) in its `conf.d` directory.
 * Changes to the CA CRL file.  For example, a `puppet cert clean`
--- a/documentation/restarting.markdown
+++ b/documentation/restarting.markdown
@@ -17,9 +17,15 @@ There are several ways to send a HUP signal to the Puppet Server process, but th
 
     kill -HUP `pgrep -f puppet-server`
 
-Starting in version 2.7.0, you can also reload Puppet Server by running the "reload" action via the operating system's service framework.  This is analogous to sending a hangup signal but with the benefit of having the "reload" command pause until the server has been completely reloaded, similar to how the "restart" command pauses until the service process has been fully restarted.
+Starting in version 2.7.0, you can also reload Puppet Server by running the "reload" action via the operating system's service framework.  This is analogous to sending a hangup signal but with the benefit of having the "reload" command pause until the server has been completely reloaded, similar to how the "restart" command pauses until the service process has been fully restarted.  Advantages to using the "reload" action as opposed to just sending a "HUP" signal include:
 
-Use the following commands to reload Puppet Server.
+1) Unlike with the "HUP signal" approach, you do not have to determine the process ID of the puppetserver process to be reloaded.
+
+2) When using the "HUP signal" with an automated script (or Puppet code), it is possible that any additional commands in the script might behave improperly if performed while the server is still reloading.  With the "reload" command, though, the server should be up and using its latest configuration before any subsequent script commands are performed.
+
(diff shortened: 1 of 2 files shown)


camlow325:
@nfagerlund, @gguillotte - ping for your review if you have time.  If you're good with this as a first pass, happy to review a follow-up PR with your edits.  Thanks!

documentation/restarting.markdown
```diff
@@ -17,13 +17,29 @@ There are several ways to send a HUP signal to the Puppet Server process, but th
...
+    systemctl reload puppetserver
+
 ## Restarting Puppet Server to pick up changes
 
-There are three ways to trigger your Puppet Server environment to refresh and pick up changes you've made. A request to the [HTTP Admin API to flush the JRuby pool](./admin-api/v1/jruby-pool.markdown) is the quickest, but picks up only certain types of changes. A HUP signal restart is also quick, and applies additional changes. Other changes require a full Puppet Server restart.
+There are three ways to trigger your Puppet Server environment to refresh and pick up changes you've made. A request to the [HTTP Admin API to flush the JRuby pool](./admin-api/v1/jruby-pool.markdown) is the quickest, but picks up only certain types of changes. A HUP signal or service reload is also quick, and applies additional changes. Other changes require a full Puppet Server restart.
 
 > **Note:** Changes to Puppet Server's [logging configuration in `logback.xml`][logback.xml] don't require a server reload or restart. Puppet Server recognizes and applies them automatically, though it can take a minute or so for this to happen.
 
-### Changes applied after a JRuby pool flush, HUP signal, or full Server restart
```
gguillotte: Why remove "HUP signal" here? Did something change regarding sending a manual HUP signal?

documentation/restarting.markdown
```diff
@@ -17,13 +17,29 @@ There are several ways to send a HUP signal to the Puppet Server process, but th
 
     kill -HUP `pgrep -f puppet-server`
 
+Starting in version 2.7.0, you can also reload Puppet Server by running the "reload" action via the operating system's service framework.  This is analogous to sending a hangup signal but with the benefit of having the "reload" command pause until the server has been completely reloaded, similar to how the "restart" command pauses until the service process has been fully restarted.
+
+The following commands could be used to trigger a service reload.
```
gguillotte: Or more simply:

> Use the following commands to reload Puppet Server.

documentation/restarting.markdown
```diff
@@ -17,13 +17,29 @@ There are several ways to send a HUP signal to the Puppet Server process, but th
 
     kill -HUP `pgrep -f puppet-server`
 
+Starting in version 2.7.0, you can also reload Puppet Server by running the "reload" action via the operating system's service framework.  This is analogous to sending a hangup signal but with the benefit of having the "reload" command pause until the server has been completely reloaded, similar to how the "restart" command pauses until the service process has been fully restarted.
```
gguillotte: > This is analogous to sending a hangup signal but with the benefit of having the "reload" command pause until the server has been completely reloaded

Why is this a benefit? I think I know why--it's safer in terms of stability and behaves more consistently--but I'm not sure that the benefit to users is clear here.

Also, if this is the preferred way of doing it, we should either bump the `reload` command process before the `kill -HUP` command, or just remove `kill -HUP` if we really don't want people to use it.

camlow325: > Why is this a benefit? I think I know why--it's safer in terms of stability and behaves more consistently--but I'm not sure that the benefit to users is clear here.

I see at least three clear advantages / cases where you'd want to use the `service reload` instead of `kill -HUP`:

1) In the context of automated scripting/tooling, `reload` more robustly helps you ensure when the server is back up and using any configuration changes that were just applied.

For example, you might have a script that does something like this:

- Change a setting in one of the puppetserver config files
- Reload the server
- Perform an agent run against the master

If the reload step were done via HUP, the server might not be up and running with the new configuration by the time the agent run is performed, possibly leading the agent run to fail or unintentionally perform actions based on the old configuration of the server.  If the reload were done via a `service reload` instead, though, the agent run should be able to contact the server and the server would assuredly be using the new configuration - leading to a more reliable, predictable result for the agent.

2) If the configuration change would cause the server to fail to reload properly - e.g., because the value for a setting were unintentionally fat-fingered - the `service reload` command would fail with an error exit code whereas the `kill -HUP` would give no indication of a failure.

3) It's arguably a bit easier to use the `service reload` than the `kill -HUP` approach because, with the `service reload` approach, the PID of the running puppetserver process does not need to be known in advance.

Granted, the `pgrep -f puppet-server` part would still allow you to dynamically determine the PID.  The whole `kill -HUP $(pgrep -f puppet-server)` command just seems a bit more onerous to me to remember than is `service puppetserver reload`.

Do you think any of the above would be useful to call out here?  I didn't want to be unnecessarily wordy here if some of these benefits would be obvious to most users.

> Also, if this is the preferred way of doing it, we should either bump the reload command process before the kill -HUP command, or just remove kill -HUP if we really don't want people to use it.

I wouldn't say that we don't want users to use the `kill -HUP` approach.  In most references that I've seen users make around the desire for service configuration changes to be reloaded without requiring a process restart, the request was for the ability to send a HUP signal to cause the reload, e.g., [SERVER-15](https://tickets.puppetlabs.com/browse/SERVER-15).  I suspect for many of our users, `kill -HUP` may still be more intuitive to use than `service puppetserver reload` - at least for one-off configuration reloads during development.  I don't think we should remove the `kill -HUP` documentation outright.  I'd be good with saying something about how we recommend the use of `service reload` in automated tooling because it more reliably determines when/whether the reload is successful and/or put the `service reload` section before the `HUP` section if you think that makes more sense, though.

@gguillotte - WDYT?  Happy to reword this however you think would be best for users.

documentation/restarting.markdown
camlow325: No, this section still applies to the HUP signal as well.  I'd removed that because I tend to think of "reload" as the more generic term that encompasses HUP signal, service reload, and systemd/init script reload.  But, thinking about it again, it would probably be confusing to remove the reference to HUP signal here so I added it back in with d98b55a.   Thanks!

documentation/restarting.markdown
gguillotte: > Do you think any of the above would be useful to call out here? I didn't want to be unnecessarily wordy here if some of these benefits would be obvious to most users.

For the users where it isn't immediately obvious, I think we can at least note that `reload` doesn't require users to track the `puppetserver` process's PID, has more consistent and predictable results, and returns a more informative exit code.

> I'd be good with saying something about how we recommend the use of service reload in automated tooling because it more reliably determines when/whether the reload is successful

👍  to that.

camlow325: @gguillotte I added some more verbiage about the advantages of the "reload" command vs. SIGHUP in 5eb3e88.  Let me know if there are any additional changes that you'd like to see me make before merging this.  Thanks!

jpinsonault:
@gguillotte Looks like your feedback has been addressed

camlow325:
Sweet, thanks, @gguillotte.  @jpinsonault - care to do the honors merging this?

PR #1256, opened 2016-10-26
(SERVER-797) Allow jvm heap size to be configurable for dev/test

camlow325:
This PR adds options for configuring the min/max heap size which is
used when running lein project commands.  This allows for the heap size
to be bumped up to a significantly higher value for local development
purposes, with the benefit of significantly reducing the total run time
for the test suite.  An alternate min/max heap size can be configured
via either an environment variable, PUPPETSERVER_HEAP_SIZE, or a
profiles.clj setting, user.puppetserver-heap-size.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -7,6 +7,26 @@
     :password :env/nexus_jenkins_password
     :sign-releases false })
 
+(def heap-size-from-profile-clj
+  (let [profile-clj (io/file (System/getenv "HOME") ".lein" "profiles.clj")]
+    (if (.exists profile-clj)
+      (-> profile-clj
+        slurp
+        read-string
+        (get-in [:user :puppetserver-heap-size])))))
+
+(defn heap-size
+  [default-heap-size heap-size-type]
+  (or
+    (System/getenv "PUPPETSERVER_HEAP_SIZE")
+    heap-size-from-profile-clj
+    (do
+      (println "Using" default-heap-size heap-size-type
+        "heap since not set via PUPPETSERVER_HEAP_SIZE environment variable or"
+        "user.puppetserver-heap-size in ~/.lein/profiles.clj file. Set to at"
+        "least 5G for best performance during test runs.")
+      default-heap-size)))
+
 (defproject puppetlabs/puppetserver ps-version
   :description "Puppet Server"
 
@@ -138,7 +158,9 @@
             "irb" ["trampoline" "run" "-m" "puppetlabs.puppetserver.cli.irb"]}
 
   ; tests use a lot of PermGen (jruby instances)
-  :jvm-opts ["-XX:MaxPermSize=256m" "-Xmx2g"]
+  :jvm-opts ["-XX:MaxPermSize=256m"
+             ~(str "-Xms" (heap-size "1G" "min"))
+             ~(str "-Xmx" (heap-size "2G" "max"))]
 
   :repl-options {:init-ns user}
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ jdk:
 - openjdk7
 - oraclejdk8
 script: "./ext/travisci/test.sh"
-sudo: false
+sudo: required
 notifications:
   email: false
   hipchat:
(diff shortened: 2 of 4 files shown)


camlow325:
I'll be playing around with Travis test runs against these changes to see if I can optimize the run time a bit.  Please don't merge at least until I've been able to clean out some of the Travis instrumentation.  Will add a comment on the PR when I'm done tinkering.

camlow325:
I fixed up the Travis config in `033ecde` with a heap size value, 6 GB, that I think will be workable for now.  This is under the 7.5 GB that Travis currently uses for VMs on its [Standard infrastructure](https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments).

With the new heap size value, I saw [one run of the PR](https://travis-ci.org/puppetlabs/puppetserver/builds/170872441) go through Travis in 37 minutes total (including the time waiting for the job to start running).  The run time on the fastest cell was 10 minutes vs. about 12 minutes for the slowest one.

Comparatively, we have been seeing total run times on recent PRs frequently be in the 1 - 2 hr range total, with per-cell times in the 30-40 minute range.

Seems like the Travis run times might be considerably improved with these changes although we'd have to monitor this to see how it goes.  If the Standard infrastructure VMs don't appear to hold up well, we could go back to the Container infrastructure and drop the heap size back down to something under the 4 GB range that Travis currently advertises.

cprice404:
yay

project.clj
```diff
@@ -7,6 +7,26 @@
...
+      (-> profile-clj
+        slurp
+        read-string
+        (get-in [:user :puppetserver-heap-size])))))
+
+(defn heap-size
+  [default-heap-size heap-size-type]
+  (or
+    (System/getenv "PUPPETSERVER_HEAP_SIZE")
+    heap-size-from-profile-clj
```
rlinehan: Would it be useful to have a message print out in this case too? So that users are reminded of what settings they are using? I feel like I could go either way on that being actually useful.

project.clj
```diff
@@ -7,6 +7,26 @@
...
+    (do
+      (println "Using" default-heap-size heap-size-type
```
rlinehan: It seems slightly odd that this gets printed out if you do anything with lein in this project and you don't have something set - e.g.

```
$ lein version
Using 1G min heap since not set via PUPPETSERVER_HEAP_SIZE environment variable or user.puppetserver-heap-size in ~/.lein/profiles.clj file. Set to at least 5G for best performance during test runs.
Using 2G max heap since not set via PUPPETSERVER_HEAP_SIZE environment variable or user.puppetserver-heap-size in ~/.lein/profiles.clj file. Set to at least 5G for best performance during test runs.
Leiningen 2.7.1 on Java 1.8.0_51 Java HotSpot(TM) 64-Bit Server VM
```

This also shows up in the travis output, since it's doing `lein version` and `lein deps` before the `lein test` with the env variable sent (e.g. https://travis-ci.org/puppetlabs/puppetserver/jobs/170895358).

rlinehan:
👍 aside from my question about whether it's possible to only print out that message when running the tests.

project.clj
camlow325: I'm not sure there's a way to avoid it.  I see those messages if I don't have the settings defined even if I move the `:jvm-opts` section down into the `:test` profile and just run `lein version`.  @cprice404 - do you have any ideas on that?

cprice404: @camlow325 I think lein has a `:test` profile as one of it's default profiles, and that it automatically adds that in when you run `lein test`.  You could experiment with just putting the jvm args explicitly into a `:test` profile in our `project.clj`, and only calling your macro expansions from that version of the jvm args definition?

camlow325: Hmm, yeah, that's what I tried.  And still `lein version` displays the messages for me.  It seems like just having it in place at all - even if the profile isn't being used - is enough to have the messages displayed.  I see them if I do:

```clj
:profiles {:blah {
                    :jvm-opts [
                           ~(str "-Xms" (heap-size "1G" "min"))
                           ~(str "-Xmx" (heap-size "2G" "max"))]}
```

... and just run `lein version`.  Seems like the code is resolved even if for a profile that you don't have in use?

cprice404: Oh yeah, duh, it's a macro expansion.  Hmmm... then yeah I can't necessarily think of a way to lazily execute that :/

PR #1257, opened 2016-10-27
Remove known issue about config reloads

jvperrin:
Reloads do work successfully now and [SERVER-15](https://tickets.puppetlabs.com/browse/SERVER-15) is closed, so I believe this section of the documentation about a known issue with config reloading can be removed.


jvperrin:
Not sure why this failed on a documentation change, looks like the first test just timed out for some reason. I'll try the hacky way of retriggering a build by closing and opening the PR.

cprice404:
@jvperrin Thanks for the PR!  Had forgotten about that section in the docs.

Also note that we should be doing another release soon that  includes support for a `reload` action built into the service frameworks (systemd/sysv) so that it's a bit easier to take advantage of this.

PR #1259, opened 2016-10-28
(SERVER-797) Use more jruby mocking in various Clojure tests

camlow325:
This PR includes a new mock `jruby-pool-manager-service` for
testing, with a `create-pool` implementation which creates a pool of
fake JRuby ScriptingContainers.  The commit also includes variations of
the `with-puppetserver-running...` functions which utilize the mock
`jruby-pool-manager-service`.

The fake `ScriptingContainer`s are reified from a new
`ScriptingContainer` interface exposed from the jruby-utils library,
which requires bumping the `jruby-utils` dependency to a newer version,
0.4.1.

This PR also updates many of the existing Clojure tests to use the new
mock `jruby-pool-manager-service`.  Locally, I saw this result in a
reduction from just over 7 minutes for the full Clojure test suite to about
4 minutes, 45 seconds.


camlow325:
This PR is dependent upon changes that would be delivered in a new jruby-utils release - currently pending https://github.com/puppetlabs/jruby-utils/pull/30 - and so will not currently pass Travis tests.  Once that is merged and released, I'll update this PR with the new dependency.

camlow325:
I added another commit to the PR, 5f2cce6, which allows the http-client tests to reuse a single ScriptingContainer.  With this change, the total runtime for the suite dropped to just under 4 minutes locally.

camlow325:
I added one more commit, 92b86a8, which refactors the environment-classes-int-tests to use the new `mock-jruby-pool-manager-service` in place of the prior logic it had been doing to `with-redef` the `create-pool-instance` function for mocking.  This commit doesn't have much of an impact on test run time but it seemed worthwhile to eliminate the `with-redef` along with some other test-specific helper code.

camlow325:
Added one more commit, 38edc02, based on feedback from the related jruby-util PR.  This commit eliminates the use of a method that we would prefer not to expose through the new `ScriptingContainer` interface in jruby-utils for now.

test/integration/puppetlabs/puppetserver/bootstrap_testutils.clj
```diff
@@ -49,6 +51,12 @@
         (assoc-in [:jruby-puppet :master-log-dir] master-log-dir)
         (ks/deep-merge overrides))))
 
+(defn services-from-dev-bootstrap-plus-mock-jruby-pool-manager-service
+  [config]
+  (conj (tk-bootstrap/parse-bootstrap-config!
+         jruby-puppet-testutils/dev-bootstrap-file-without-jruby-pool-service)
```
cprice404: Sucks that we have to write a temp file for this, but I think I can see why you did it that way. 

Would it not be possible, though, to just call `parse-bootstrap-config!` directly on the "real" file, and then do a `filter` on the resulting vector to remove the real jruby pool service?

test/unit/puppetlabs/services/jruby/jruby_puppet_testutils.clj
```diff
@@ -181,3 +189,114 @@ create-mock-pool-instance :- JRubyInstance
...
+     "localcacert" (str certdir "/ca.pem")
+     "manage_internal_file_permissions" true
+     "privatekeydir" private-key-dir
+     "requestdir" (str ssldir "/certificate_requests")
+     "serial" (str cadir "/serial")
+     "signeddir" (str cadir "/signed")
+     "ssl_client_header" "HTTP_X_CLIENT_DN"
+     "ssl_client_verify_header" "HTTP_X_CLIENT_VERIFY"
+     "trusted_oid_mapping_file" (str confdir
+                                     "/custom_trusted_oid_mapping.yaml")}))
```
cprice404: Hmmm, this is a bummer.  It's probably better than any alternatives but it seems like it might be a bit of a bear to maintain.  I can definitely imagine people from feature teams coming in and adding new settings to the whitelist of stuff we pull from puppet, and not knowing that they need to update this - not sure if any tests would fail unless they happened to be using mock jrubies *and* exercising their new things in the same test?  Which seems like it wouldn't always be the case?

Do you think it might be worth having a memoized helper function that just loads up a scriptingcontainer and scrapes this stuff once?  Or loads up the puppet-server-config-service or whatever it's called, grabs the config from that, and memoizes it?  I know that'd probably add 10 seconds or something back to the overall test runtime but maybe that's worth it in exchange for the maintenance overhead of this chunk?

Not sure, just a thought.

test/unit/puppetlabs/services/jruby/jruby_puppet_testutils.clj
```diff
@@ -181,3 +189,114 @@ create-mock-pool-instance :- JRubyInstance
...
+    (reify JRubyPuppet
+      (getSetting [_ setting]
+        (get puppet-config setting))
+      (puppetVersion [_]
+        "1.2.3")
+      (handleRequest [_ _]
+        (JRubyPuppetResponse. (Integer. 200)
+                              "looks good to me"
+                              "text/plain"
+                              "1.2.3"))
```
cprice404: Are there tests that are going to call this?  If not I might vote for having this throw an exception?

test/unit/puppetlabs/services/jruby/jruby_puppet_testutils.clj
```diff
@@ -181,3 +189,114 @@ create-mock-pool-instance :- JRubyInstance
...
+        count (.remainingCapacity pool)]
+    (dotimes [i count]
+      (let [id (inc i)]
+        (create-mock-pool-instance
+         (partial create-mock-jruby-puppet config)
+         pool
+         id
+         jruby-config
+         (constantly nil))))
+    pool-context))
```
cprice404: I presume this is largely lifted from a function in `jruby-utils`?  If so, maybe worth adding a comment mentioning that, just to try to help make sure we keep them roughly in sync with one another?

test/unit/puppetlabs/services/jruby/jruby_puppet_testutils.clj
camlow325: I think one of the problems with the memoizing approach is that many of the settings are calculated based on some base settings whose values change - especially the `:master-conf-dir`.  Many tests use a "tmp" `:master-conf-dir`.  If we had to memoize the settings per `:master-conf-dir` used, might negate a lot of the savings we get by avoiding JRuby entirely.

One thing I could do to make it more obvious that a new setting needs to be stubbed into the test is have the `getSetting` method in the `create-mock-jruby-puppet` helper explicitly throw an `Exception` if the requested setting isn't found in the map.  If we did that, a bunch of tests would fail if a new setting were plumbed but no value were stubbed into this map.  I could also add a comment in the `puppet-server-config-service` which indicates that if a value is added to the whitelist that an entry needs to be inserted into this map.

Do you think that would work?

test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
```diff
@@ -259,7 +259,7 @@
 (deftest ^:integration static-file-content-endpoint-test-no-code-content-command
   (logging/with-test-logging
    (testing "the /static_file_content endpoint errors if :code-content-command is not set"
-     (bootstrap/with-puppetserver-running
+     (bootstrap/with-puppetserver-running-with-mock-jrubies
```
cprice404: So I think I'm good with using mocks in all the places that you did so in this namespace.  However, there are a couple of them that are confusing because they appear to be making a call to the catalog endpoint, which in my head would *definitely* need the jrubies to be real.  It works here though because the code-id check stuff happens in clojure before we would hand off the request, and fails.

I'm wondering if it would be worthwhile to add a comment at the beginning of this namespace explaining why it's OK?  Or if maybe we should have some general pattern around explaining why it's safe to use mocks on the various tests where we do it?

This may sound crazy (and it probably is) but I almost wonder if we should have something like a docstring argument to these macros, that just explains why mocking is safe for the test at hand.  Might save some confusion in the future?  Might also be total overkill?

cprice404: Yeah, basically the process of reviewing this PR for me is going to be looking at each one of the calls to this macro one by one and reasoning about why it is/isn't sane.  So maybe a docstring-ish approach isn't totally bananapants?

camlow325: Are you asking for a docstring in the defmacro itself or a comment at each point where the macro is used to say why it's safe to use a mock instead of the real thing - or both?

test/unit/puppetlabs/services/jruby/jruby_puppet_testutils.clj
camlow325: Yeah, there are a few.  I don't remember which ones offhand, though.  In each of those cases, it's just looking for a 200 OK (as opposed to some other error that might be thrown at a higher level in the Clojure request handler stack - nothing in the response body.  I figured "200" would be a safe vanilla response code for asserting that the call down to the response handler was successful.

test/integration/puppetlabs/puppetserver/auth_conf_test.clj
```diff
@@ -304,7 +304,10 @@
                            :sort-order 1
                            :name "all endpoints"}]}
          :webserver {:host "localhost"
-                     :port 8080}}
+                     :port 8080}
+         :puppet {"csr_attributes" (str test-resources-dir "/csr_attributes.yaml")
+                  "trusted_oid_mapping_file" (str test-resources-dir
+                                                  "/custom_trusted_oid_mapping.yaml")}}
```
cprice404: Interesting.  Hadn't thought about cases where the puppet settings were the only things missing.  In general that seems like probably an ok thing to add to the tests to work around that dependency... I think?

Do you know if there is another test that exercises these settings "for realz" so that we can validate that they are plumbed through?  Also these don't seem like the default values for these settings, so how were they getting plumbed through before?  Via puppet.conf?

test/unit/puppetlabs/services/jruby/jruby_puppet_testutils.clj
camlow325: Yeah, it's similar to one of the `jruby-utils` functions - although simpler with respect to the construction of instances for the pool.  I can add a comment referencing the corresponding `jruby-utils` function.

test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj
cprice404: I'm saying that we could add a fake docstring arg to the macro definition, so that consumers of the macro would be forced to pass in a string as the first argument.  The string could be ignored by the implementation of the macro but would enforce that we had to put a little snippet of text in each use case explaining why it's safe to circumvent the real rubies.  Does that seem like overkill?

camlow325: Sure, if you want to do that, that'd be great.  Or if you want me to do it to justify why I thought it was okay to use mocks in various places, I don't have a problem with that.  Whatever you think is best.

test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
```diff
@@ -175,17 +174,21 @@
                           (init [this context] (lc-fn context :init-bonus-service))
                           (start [this context] (lc-fn context :start-bonus-service))
                           (stop [this context] (lc-fn context :stop-bonus-service))
-                          (bonus-service-fn [this] (lc-fn nil :bonus-service-fn)))]
+                          (bonus-service-fn [this] (lc-fn nil :bonus-service-fn)))
+          config-overrides {:global {:logging-config
+                                     (str "./dev-resources/puppetlabs/services/"
+                                          "jruby/jruby_pool_int_test/"
+                                          "logback-test-restart-comes-back.xml")}
```
cprice404: Is this just for Wolfe Principle?

test/unit/puppetlabs/services/jruby/jruby_puppet_testutils.clj
cprice404: Might be worth talking this one through a little more once we get closer to a final version of this PR.  My gut is that any test that is actually making a call down into `handleRequest` should probably not be getting mocked, and if there are only a few tests that are relying on this, it might be a worthwhile tradeoff in my mind to just let those tests be a little slower for now.  But it might also depend on which tests we're talking about... we can circle back around to this.

test/integration/puppetlabs/services/jruby/service_macros_test.clj
```diff
@@ -4,21 +4,21 @@
...
-        ; Just make sure the macro doesn't fail
-        (is (true? true)))
-       (is (true? true))))))
+    (let [config (jruby-testutils/jruby-puppet-tk-config
+                  (jruby-testutils/jruby-puppet-config))]
+      (tk-testutils/with-app-with-config
+       app
+       [profiler/puppet-profiler-service
+        jruby-puppet/jruby-puppet-pooled-service
+        (jruby-testutils/mock-jruby-pool-manager-service config)]
```
cprice404: Haven't gone all the way through the stack to fully grok how locking the mocked pool manager service would relate to locking the "real" service.  The locking is important enough to where I might vote for not mocking here, but we can circle back around on this one later as well.

test/integration/puppetlabs/services/legacy_routes/legacy_routes_test.clj
```diff
@@ -31,20 +31,22 @@
...
-      {}
-      (is (= 200 (:status (http-get "/v2.0/environments"))))
-      (is (= 200 (:status (http-get "/production/node/localhost"))))
-      (is (= 200 (:status (http-get "/production/certificate_statuses/all")))))))
+    (bootstrap/with-puppetserver-running-with-mock-jrubies
+     app
+     {}
+     (is (= 200 (:status (http-get "/v2.0/environments"))))
+     (is (= 200 (:status (http-get "/production/node/localhost"))))
+     (is (= 200 (:status (http-get "/production/certificate_statuses/all")))))))
```
cprice404: This seems like one of the tests that is probably relying on the mocked `handleRequest`.  It kind of feels like we are losing some of the value of this test by mocking that - now we're really only validating that `handleRequest` gets called, not that it gets called in a manner that would actually yield a valid result from the ruby code... right?

I could see a case being made here that these tests should potentially be making some additional assertions on the body of the response, and that the lack of that coverage is what made it seem acceptable to just mock 200s coming back, but I'm iffy on this one.

test/integration/puppetlabs/services/jruby/service_macros_test.clj
camlow325: The locking controls whether borrows can be done against a `JRubyPool` object.  Even in the case of mocking, a real `JRubyPool` object is still created and exercised when the lock is performed.  The thing that's different when the mock is in place is what the objects in the pool are - mock `JRubyPuppet` objects.  If you're concerned enough about this, I could convert it back to using the real pool manager.  I don't think it's necessary, though, to exercise the code that we're trying to test here.

test/integration/puppetlabs/services/legacy_routes/legacy_routes_test.clj
camlow325: For the `environments` and `node` endpoints, yeah, that would be true.  If you're concerned enough about it, I'll convert it back to using a real JRuby - and add some more assertions around the body of the response as well.

test/integration/puppetlabs/services/legacy_routes/legacy_routes_test.clj
```diff
@@ -31,20 +31,22 @@
...
-        {:master-routes "/puppet"
-         :invalid-in-puppet-4 "/"}}}
-      (is (= 200 (:status (http-get "/puppet/v3/node/localhost?environment=production"))))))
+    (bootstrap/with-puppetserver-running-with-mock-jrubies
+     app
+     {:web-router-service
+      {:puppetlabs.services.master.master-service/master-service
+       {:master-routes "/puppet"
+        :invalid-in-puppet-4 "/"}}}
+     (is (= 200 (:status (http-get "/puppet/v3/node/localhost?environment=production"))))))
```
cprice404: This one I think I'd be OK with the mocking because there is nothing about the request handler that is relevant to this test, but it'd be better if we could just make a request that doesn't trigger the request handler, so that the mocked request handler could still throw an exception.  I don't recall if there are any routes under `/puppet` that are clojure-only though :/

camlow325: I guess maybe `static_file_content` could be used here instead but then we'd probably need to add a lot of other configuration to support it as well.

Sounds like what I need to do is change `handleRequest` in the mock JRubyPuppet to throw an exception and, for every test that fails as a result, change the test back to using a real JRuby instead.  Kind of a bummer, but I'll do it.

test/unit/puppetlabs/services/request_handler/request_handler_core_test.clj
```diff
@@ -292,45 +290,41 @@
...
+          (deref in-catalog-request-future)
+          (is (deref first-promise 10000 false))
+          ; Because we are blocking inside current-code-id, which happens to be
+          ; used during a jruby request, we can assert that there will be no
+          ; jruby instances left in the pool.
+          (is (zero? (jruby-protocol/free-instance-count jruby-service)))
+          (deliver second-promise true))))))
 
 (deftest request-handler-test
     (let [dummy-service (reify jruby-protocol/JRubyPuppetService
```
cprice404: Done with this commit, nothing here other than calls to the macro that I will evaluate when I'm adding docstrings.

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -98,19 +97,14 @@ jruby-config :- jruby-schemas/JRubyConfig
...
-
+         scripting-container (-> (jruby-impl-core/create-pool jruby-config)
+                                 (jruby-core/borrow-from-pool :http-client-test [])
+                                 :scripting-container)]
+     (.runScriptlet scripting-container "require 'puppet/server/http_client'")
      (let [http-client-class (.runScriptlet scripting-container "Puppet::Server::HttpClient")]
        (.callMethod scripting-container http-client-class "initialize_settings" http-client-settings Object))
      (.runScriptlet scripting-container (str "require 'puppet/server/http_client';"
-                                             (format "c = Puppet::Server::HttpClient.new('localhost', %d, {:use_ssl => %s});"
+                                             (format "$c = Puppet::Server::HttpClient.new('localhost', %d, {:use_ssl => %s});"
```
cprice404: state

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -325,28 +319,28 @@ jruby-config :- jruby-schemas/JRubyConfig
...
+            (is (= "200" (.runScriptlet sc "$response.code")))
+            (is (= "bye" (.runScriptlet sc "$response.body")))
+            (.runScriptlet sc (str (format "$c = Puppet::Server::HttpClient.new('localhost', %d, {:use_ssl => %s});"
                                            10080 true)
-                                   "response = c.get('/', {})"))
-            (is (= "200" (.runScriptlet sc "response.code")))
-            (is (= "hi" (.runScriptlet sc "response.body")))
+                                   "$response = $c.get('/', {})"))
+            (is (= "200" (.runScriptlet sc "$response.code")))
+            (is (= "hi" (.runScriptlet sc "$response.body")))
```
cprice404: This commit seems fine.

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
```diff
@@ -21,222 +21,216 @@
...
-                    (jruby-protocol/get-environment-class-info-cache-generation-id!
-                     service
-                     "staging"))
-                 (str "Cache id should not have been changed by the prior "
-                      "set since the environment was marked expired after the "
-                      "`staging-cache-id-before-marked-expired` was "
-                      "read")))))))))
+                        "set since the environment was marked expired after the "
+                        "`staging-cache-id-before-marked-expired` was "
+                        "read"))))))))))
```
cprice404: I don't really have my head wrapped around the implications of testing the etag stuff w/o real jrubies but I will trust your judgment on this commit.

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -76,82 +76,120 @@ jruby-config :- jruby-schemas/JRubyConfig
...
+
+(defmacro with-scripting-container
+  [scripting-container & body]
+  `(let [~scripting-container (or *scripting-container*
+                                  (create-scripting-container))]
+     (try
+       ~@body
+       (finally
+         (when-not *scripting-container*
+           (terminate-scripting-container ~scripting-container))))))
```
cprice404: Just curious if there is a reason why using a macro and a dynamic var is better here, as opposed to just def'ing up a single scripting container as a variable, or making a memoized function that returns one?

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -181,166 +219,164 @@ jruby-config :- jruby-schemas/JRubyConfig
...
+          (with-scripting-container sc
+            (with-http-client sc port {:use-ssl false}
+              (.runScriptlet sc "$response = $c.get('/', {})")
+              (is (= "200" (.runScriptlet sc "$response.code")))
+              (is (= "bye" (.runScriptlet sc "$response.body")))
+              (.runScriptlet sc (str (format "$c = Puppet::Server::HttpClient.new('localhost', %d, {:use_ssl => %s});"
+                                             10080 true)
+                                    "$response = $c.get('/', {})"))
+              (is (= "200" (.runScriptlet sc "$response.code")))
+              (is (= "hi" (.runScriptlet sc "$response.body"))))))))))
```
cprice404: This commit seems fine, except possibly exploring doing something a little simpler than the macro/dynamic var/binding stuff, but it's not a blocker or anything.

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
camlow325: May not be a big deal, I guess, but I was thinking it would be nice to have the `ScriptingContainer` created when the test suite starts and get freed up when the tests in the http-client suite are all finished (before moving on to other namespaces in the context of a full test run).  It kind of made sense to me to do this via a test fixture.  The pattern I used for these basically follows this: https://stuartsierra.com/2016/05/19/fixtures-as-caches.  If you think this is too onerous, though, I could go with the def instead.

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -109,8 +109,11 @@ jruby-config :- jruby-schemas/JRubyConfig
                                                     :cipher-suites]))
         http-client-class (.runScriptlet scripting-container
                                          "Puppet::Server::HttpClient")]
-    (.callMethod scripting-container http-client-class
-                 "initialize_settings" http-client-settings Object)
+    (.callMethodWithArgArray scripting-container
+                             http-client-class
+                             "initialize_settings"
+                             (into-array Object [http-client-settings])
+                             Object)
```
cprice404: yep

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
cprice404: Hmm.  You could still terminate it if it was memoized but I guess it wouldn't be freed because there'd still be a reference to it.

meh.  i guess it's fine this way.

cprice404:
OK, I'm done with a first pass at this.  I skipped most of the diffs where it was just a one line change from the old with-puppetserver macros to the new mock ones.  Next step for me is to add a commit that puts docstrings into all of those.  Then we can lump together anyones that seem questionable that come out of that, plus the rando stuff I commented on in the first pass, and get a checklist going on how to finish up.

test/integration/puppetlabs/puppetserver/bootstrap_testutils.clj
camlow325: Sounds reasonable.   I did this in d144308.  Hope that works for you.

test/unit/puppetlabs/services/jruby/jruby_puppet_testutils.clj
cprice404: We talked about this IRL, and I think that where we landed is that we will add a test that does the following:

* Instantiates the puppet server config service with a real JRuby and grabs all the values from it
* Makes an assertion about each value that we expect to see there (probably a copy of everything from our whitelist, with some hard-coded definitions of expected values)
* Makes an assertion that there are no other values in the map returned by the config service beyond the ones that we made assertions about in the previous step
* Repeats both of the previous two bullet points using the hard-coded mock settings map as input

With a test like this in place, it shouldn't be possible for people to add new settings to the whitelist without both a) adding an assertion about the expected value for the setting, from a "real" JRuby code path, to this test, and b) adding an appropriate default value for the setting to the mock data structure.

test/integration/puppetlabs/services/jruby/service_macros_test.clj
cprice404: okay - if you're convinced this is safe that's fine w/me.

test/integration/puppetlabs/services/legacy_routes/legacy_routes_test.clj
cprice404: Don't do that yet.  I just want to talk through them.

test/integration/puppetlabs/puppetserver/bootstrap_testutils.clj
```diff
@@ -53,9 +54,11 @@
...
 (defn services-from-dev-bootstrap-plus-mock-jruby-pool-manager-service
   [config]
-  (conj (tk-bootstrap/parse-bootstrap-config!
-         jruby-puppet-testutils/dev-bootstrap-file-without-jruby-pool-service)
-        (jruby-puppet-testutils/mock-jruby-pool-manager-service config)))
+  (->> dev-bootstrap-file
+       tk-bootstrap/parse-bootstrap-config!
+       (remove #(= :PoolManagerService (tk-services/service-def-id %)))
+       vec
+       (cons (jruby-puppet-testutils/mock-jruby-pool-manager-service config))))
```
cprice404: thanks, this seems better to me.

cprice404:
At this point I think we are waiting for:

- [ ] extended test coverage of puppet settings
- [ ] discussion on test-jruby-pool-not-full-during-code-id-generation in request_handler_core_test.clj
- [ ] discussion on test-503-when-jruby-is-first-to-shutdown in jruby_puppet_pool_int_test.clj
- [ ] discussion on legacy-routes in legacy_routes_test.clj
- [ ] discussion on old-master-route-config in legacy_routes_test.clj (3 separate tests)

test/integration/puppetlabs/services/legacy_routes/legacy_routes_test.clj
```diff
@@ -32,6 +32,9 @@
 (deftest ^:integration legacy-routes
   (testing "The legacy web routing service properly handles old routes."
     (bootstrap/with-puppetserver-running-with-mock-jrubies
+     "MOCK.TODO: discuss this one.  Currently all this is testing is that the
```
camlow325: Yeah, that's a good point.  Definitely a good argument to be made that the original tests for this were already insufficient.

I think the primary thing we need to cover is ensuring that the request made to the old route is eventually funneled down as the new route into the Ruby Puppet layer.  To that end, we *could* still handle this via mocking if we had the mock `.handleRequest` method just echo back the request map into the response body and have these tests inspect the response body to see that each request has been mapped to the appropriate new indirector route by the legacy routes service.  I suspect you might still think that's insufficient, though.

If you think we need to do an end-to-end test down through the indirector terminii, I can go ahead and convert this back to a non-mock `with-puppetserver-running` and add some extra assertions against the response body content for each.

test/integration/puppetlabs/services/legacy_routes/legacy_routes_test.clj
```diff
@@ -41,6 +44,11 @@
 (deftest ^:integration old-master-route-config
   (testing "The old map-style route configuration map still works."
     (bootstrap/with-puppetserver-running-with-mock-jrubies
+     "MOCK.TODO: discuss - this is *probably* ok because all we really care about
+     is that the webserver is accepting connections and handing them to the correct
+     ring handler.  Would be better if we could implement this to hit a clojure
+     endpoint instead of a Ruby one but I don't remember if there are any pure-clojure
+     endpoints under /puppet."
```
camlow325: static-file-content is a pure Clojure one, I think, although it would seem to come with a fair amount of extra overhead to setup with the versioned-code-service config, et al.  TBH, though, I think it's fine to just have this one only go so far as to ensure that the request would get handed off to the Ruby layer.  The only reason this test exists, if I recall correctly, is for backward compatibility with the one early release of 2.0 where the `:invalid-in-puppet-4` key was present in the config.  Probably pretty unlikely that many users will even have this format in their configs after upgrading to later Puppet Server versions.

test/integration/puppetlabs/services/legacy_routes/legacy_routes_test.clj
```diff
@@ -57,6 +65,11 @@
           webserver-config (:webserver default-config)]
       ;; use `-with-config` variant, so that we can pass in the entire config map
       (bootstrap/with-puppetserver-running-with-mock-jrubies
+       "MOCK.TODO: discuss - this is *probably* ok because all we really care about
+        is that the webserver is accepting connections and handing them to the correct
+        ring handler.  Would be better if we could implement this to hit a clojure
+        endpoint instead of a Ruby one but I don't remember if there are any pure-clojure
+        endpoints under /puppet."
```
camlow325: static-file-content might work here, too, although it seems onerous to have to set that up for this test.  As with an earlier comment, I think if we provided a way to echo the request map from `.handleRequest()` back out though the response and could use that to just validate that the request was handed off correctly that I'd be fine with leaving this as a mock and using the "node" endpoint.  If you're opposed to that whole approach, though, maybe it'd be better to convert this back to using a real JRubyPuppet and add some more assertions to see that we got a response body which is appropriate for a "node" request.

test/unit/puppetlabs/services/request_handler/request_handler_core_test.clj
```diff
@@ -311,6 +311,17 @@
...
        (jruby-bootstrap/with-puppetserver-running-with-services-and-mock-jrubies
+        "MOCK.TODO: This test is making a catalog request, in order
+        to trigger the code-id handling.  With the current jruby mocking that
+        means that we are artificially returning a 200 for that catalog request,
+        but this test passes because it doesn't make any assertions on the body.
+        The test description probably needs updating too... it seems like the
+        goal of this test is to assert that the jruby is borrowed before the
+        code id command is run, so that we know that a lock can't be acquired
+        (and new code deployed) while we're running the code id command.  If that's
+        the case, we should explain it more clearly, and it's probably OK to
```
camlow325: That all makes sense to me.  I agree that we're validating that the code id command involves reserving a JRuby specifically so that the pool cannot be locked (and therefore possibly code being updated) while the code id is being computed.  I'll make an attempt to add an extra comment about the purpose of the test in the comment under the `testing` block above.

test/integration/puppetlabs/services/legacy_routes/legacy_routes_test.clj
camlow325: @cprice404 I don't think any other acceptance tests cover it.  I agree that we should add some more assertions beyond just the status code - some that validate the content of the response body.  The part I'm still a little on the fence about is whether we think we could use mock JRubies / maybe in combination with something that echoes the request map supplied to `.handleRequest` back out through the response -- so we can assert that the route was translated correctly -- or if we think we need a "real" JRubyPuppet so we can get back "real" indirector endpoint responses.  Probably just need to make a call one way or the other on that.  I can go either way.  I'm comfortable with the mock approach for speed but understand if you're concerned enough about insufficient real test coverage that we should instead use real JRubies.  What do you think?

test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
```diff
@@ -249,6 +249,8 @@
 (deftest ^:integration test-503-when-jruby-is-first-to-shutdown
   (testing "During a shutdown requests result in 503 http responses"
     (bootstrap/with-puppetserver-running-with-mock-jrubies
+     "MOCK.TODO: discuss this one; it's probably safe because we don't really care
```
camlow325: With this test, I think we're just interested in seeing that after the jruby-puppet-pooled-service is shutdown, any requests to borrow an instance from the pool return a 'service-unavailable' - translated to an HTTP 503 at the HTTP request layer.  All of the logic involved in emptying the pool and dumping a shutdown pill into it so that a borrow gets the pill and throws the 'service-unavailable' in response is still run even if the mock-jruby-pool-manager-service is being used.

So long as the mock requests would return a successful response before the JRuby service is stopped - i.e., anything other than a 503 error - I think the rest of the production code that this test is trying to validate is all still being exercised.

So, yeah, I'm basically okay with this as a mock for the reason you mentioned.  If we decide we want to change `.handleRequest` for the mock JRubyPuppet to throw an exception, though, this test would probably no longer work as written because the pre-shutdown environments requests would start failing.

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
camlow325: I have some other tests in the `class-info-test` namespace which use real JRubies to get environment class responses based on the content of manifests on disk.  Those are a layer below the etag-related code.  I've figured that for the purpose of calculating etag values that just having a representative string payload to be hashed would be sufficient at the unit test level.  For additional coverage, I also wrote some "integration" level tests in the "environment-classes-int-test" namespace.  Those ones still do end-to-end requests using real JRubies and exercise the etag validation in the process.  The only reason the unit tests are here is to do some more extensive validation of different etag scenarios than the integration tests get into - but can do so much more cheaply without any backing "real" JRubies.

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
camlow325: Yeah, probably would be better to change these tests to just get the `HttpClient` object back from the container and make calls directly on that.  Just took the lazy way out to avoid thinking about how to do all of the right `.callMethodWithArgArray` calls and interop for the different arguments.

test/unit/puppetlabs/services/jruby/jruby_puppet_testutils.clj
camlow325: Reviewing this further, we found that the prior work done in PE-17071 to add the `data_binding_terminus` setting was no longer needed.  I removed references to the `data_binding_terminus` setting from the code, along with the work to coerce the setting from a Symbol to a String which would no longer be needed.  This was done in dd5a10d.

cprice404: This all sounds good.

test/integration/puppetlabs/services/legacy_routes/legacy_routes_test.clj
cprice404: If we don't have coverage in acceptance then I'm probably in the "don't mock it" camp.  What would that add, like 10 seconds?

But I swear we have something in our rakefile that defines a Puppet 3 build version, and there was all that simmons stuff that Nate worked on... and if we're not exercising the legacy routes in acceptance then I don't totally understand what that puppet3 version is for, nor what happened with all the Simmons stuff.

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
cprice404: OK, sounds good.  Maybe add a comment pointing out where those other tests live, but otherwise, sounds fine.

test/unit/puppetlabs/services/jruby/jruby_puppet_testutils.clj
```diff
@@ -246,7 +246,13 @@ create-mock-pool-instance :- JRubyInstance
...
     (reify JRubyPuppet
       (getSetting [_ setting]
-        (get puppet-config setting))
+        (let [value (get puppet-config setting :not-found)]
+          (if (= value :not-found)
+            (throw (IllegalArgumentException.
+                    (str "Setting not in mock-puppet-config-settings "
+                         "requested: " setting ". Add an appropriate value "
+                         "to the map to correct this problem.")))
+            value)))
```
cprice404: This commit seems good

test/integration/puppetlabs/services/legacy_routes/legacy_routes_test.clj
cprice404: This is the same one we're  discussing in the thread above, right?

test/integration/puppetlabs/services/legacy_routes/legacy_routes_test.clj
camlow325: Oh, duh.  You're right.  The whole Puppet 3 compatibility suite would definitely exercise the legacy routes - using real Puppet 3 agent runs - against Puppet Server 2.x.  So, does that change your opinion about mock / don't mock - along with extra assertions for these tests.

camlow325: At least, simmons would cover the endpoints that a Puppet 3 agent would hit during its run.  I don't think an agent would hit the "environments" or "certificate_statuses" endpoints here.  For that matter, I guess "certificate_statuses" is pure Clojure anyway so that one's not even relevant to this whole discussion.

cprice404: It might.

I haven't finished reading the rest of your comments yet.  I'd still like to know how many tests would go red if we have that requesthandler thing throw an exception, and which ones they are.  Maybe you've talked about that elsewhere.

test/integration/puppetlabs/services/legacy_routes/legacy_routes_test.clj
cprice404: I think I'm fine with that rationale as it regards this particular test, but less comfortable with the idea that that justifies having the default mock behavior return a 200, because not all tests that might be exercising that mocked code path are as inconsequential as this one.

cprice404: Do you think it might be possible to do something like add another (optional?) argument to the macro that was just a function that would be called as the implementation of that request handler method in the reify?

Because, for this specific test, if you passed in a function there that returned a 200, I wouldn't have a problem with that.  And I'm sure there are other tests like that.  It just makes me nervous for it to be the default behavior rather than something we have to think about on a test-by-test basis.

camlow325: Yeah, I think I could do that.  At a lower-level, I already have an optional argument to the `mock-jruby-pool-manager-service` function which takes in a whole `mock-jruby-puppet-fn` for creating the JRubyPuppet instance.  For a little more flexibility, I could add an optional parameter to the `create-mock-jruby-puppet` function which is the `request-handler` function that you want it to call back to.  Then maybe I could plumb an optional `mock-jruby-puppet-fn` from `with-puppetserver-running-with-mock-jrubies` on through to the `mock-jruby-pool-manager-service`.  The advantage of that vs. what you're suggesting is that individual tests could also mock the other JRubyPuppet functions as well, if needed.  WDYT?

cprice404: :fried_shrimp: :books: doesn't matter much to me if you want to take it that far.  I probably wouldn't do that until we had a use case, doesn't seem like it would be hard to  add in later?  But if we could just plumb in a fn for 'request-handler', then several of the tests on our current list I think I'd be satisfied with just passing in a 200 handler and keeping the mocking turned on.

test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
```diff
@@ -240,32 +241,55 @@
...
+             (jruby-testutils/return-instance jruby-service
+                                              instance
+                                              :i-want-this-instance))
+           ;; Stop the tk app.  In success cases, this will end up being done
+           ;; twice - which is benign.  This is done in the finally block so
+           ;; that it is done even if the test errors out prematurely.  Not
+           ;; doing so could leave a Jetty webserver running indefinitely -
+           ;; causing many tests downstream from this one that try to create
+           ;; new Jetty servers to fail.
+           (tk-app/stop app)))))))
```
cprice404: this commit seems fine

test/unit/puppetlabs/services/jruby/jruby_puppet_testutils.clj
camlow325: I added a comment about this in eea0275.

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
camlow325: Added a comment about this in 12adb62.

test/integration/puppetlabs/services/legacy_routes/legacy_routes_test.clj
camlow325: Yep, we've got a couple of threads where we talked about this one.  I think this is addressed now in f476043.

test/integration/puppetlabs/services/legacy_routes/legacy_routes_test.clj
camlow325: I worked this up in f476043.

camlow325:
@cprice404 I *think* I've hit all of your comments so far on this PR now.  All of the "MOCK: TODO"s should be replaced with comments on why we're doing the mocking, along with some test-specific `.handleRequest` implementations and related assertions, where appropriate.

test/integration/puppetlabs/puppetserver/bootstrap_testutils.clj
```diff
@@ -58,6 +75,41 @@
...
+    `(let [services# (conj ~services
+                           (jruby-puppet-testutils/mock-jruby-pool-manager-service
+                            ~config))]
+       (tk-testutils/with-app-with-config
+        ~app
+        services#
+        ~config
+        ~@body))))
+
+(defmacro with-puppetserver-running-with-services-and-mock-jruby-puppet-fn
```
KevinCorcoran: Could this simply be a 2nd arity of the macro immediately above?  (i.e. optional `mock-jruby-puppet-fn` parameter)

KevinCorcoran: Same question below about `with-puppetserver-running-with-mock-jruby-puppet-fn`

test/integration/puppetlabs/services/legacy_routes/legacy_routes_test.clj
```diff
@@ -31,20 +33,57 @@
...
+     translation is expected to occur at the Clojure layer, not in Ruby."
+     app
+     {}
+     (partial jruby-testutils/create-mock-jruby-puppet
+              (fn [request]
+                (JRubyPuppetResponse.
+                 (Integer. 200)
+                 (str "request routed to: " (get request "uri"))
+                 "text/plain"
+                 "1.2.3")))
```
KevinCorcoran: This function (or something very similar) is repeated several places.  Would it be worthwhile to extract it into a helper?

camlow325:
> More importantly, though, what do you think about trying to get a code coverage report before/after these changes? 

@KevinCorcoran, I definitely understand the motivation for that although I'm not sure what the best tools for doing that are.  I did at least use [cloverage](https://github.com/cloverage/cloverage) to produce reports comparing code coverage for the tests at the Clojure-level in OSS Puppet Server.  I've uploaded links with zip files of the report content for each below:

[coverage-pr.zip](https://github.com/puppetlabs/puppetserver/files/572851/coverage-pr.zip)
[coverage-master.zip](https://github.com/puppetlabs/puppetserver/files/572852/coverage-master.zip)

Basically, the code coverage is exactly the same for master vs. code on this PR.  The coverage percentage looks slightly lower for this PR in the `puppetlabs.services.config.puppet-server-config-core` namespace but that appears to be due to this PR having removed one setting, `:data-binding-terminus`.  The actual lines executed appears to be the same between the two.

Does this help or were you looking for something else?

test/integration/puppetlabs/services/legacy_routes/legacy_routes_test.clj
camlow325: In a4cc0e7, I created a new helper function and changed some of the existing tests to use it.  The helper only applies when a canned set of parameters is supplied in the `JRubyPuppetResponse` for a `.handleRequest`.  For cases where a test needs more special logic to compute a response for a `.handleRequest` - e.g., echoing back some of the content from the request - I just left the code as-is.  It seemed that in those cases that the more verbose method for constructing the mock would be good enough.

camlow325:
@cprice404, @KevinCorcoran - okay, I think I've hit all of the feedback so far.  Let me know if you see anything else.

cprice404:
I'm still +1 but now something in travis is hitting that `public_suffix` issue.

camlow325:
@cprice404 Yeah, we'll need to merge #1267 in order for this PR to pass Travis again.  Although we may want to hold back on doing that because I think we'll need to bump that one to an even newer - yet to be determined - Puppet submodule version in order for Ruby Puppet acceptance tests to pass fully.

KevinCorcoran:
Nice work on the code coverage reports.  Was there anything about the process of creating those reports that would be worth adding to the README and/or project.clj?  Or is cloverage sufficiently self-explanatory?

cprice404:
k, sounds like we're good to merge pending travis/ruby shenans.

camlow325:
> Was there anything about the process of creating those reports that would be worth adding to the README and/or project.clj? Or is cloverage sufficiently self-explanatory?

It seemed pretty self-explanatory to me.  I just added it as a :plugin like the cloverage README described and then ran "lein test".  In the test output, it gives you a text summary of the coverage statistics - and a pointer to the HTML reports.  Probably not worth mentioning in the README, I think.

camlow325:
We should be good to go for merging this now.  The public_suffix issue was resolved by bumping the Puppet submodule up in a separate PR, #1267.  Tests are passing again via Travis on this PR.

PR #1260, opened 2016-11-01
(PE-17713) Document check_for_updates flag/analytics data collected.

mindymo:
New documentation for the code the Product Analytics team added in SERVER-1599 (allowing FOSS users to opt out of analytics data collection). Please let me know if you have any questions. Thanks! 

Attn: @gguillotte

### DIFF
--- a/documentation/config_file_puppetserver.markdown
+++ b/documentation/config_file_puppetserver.markdown
@@ -73,6 +73,19 @@ The `puppetserver.conf` file contains settings for Puppet Server software. For a
 * The `profiler` settings configure profiling:
 
     * `enabled`: If this is set to `true`, Puppet Server enables profiling for the Puppet Ruby code. The default is `false`.
+    
+* The `product` settings configure update checking and analytics data collection:
+
+    * `check_for_updates`: If set to `false`, Puppet Server will not automatically check for updates, and will not send analytics data to Puppet. 
+    
+        If this setting is unspecified (default) or set to `true`, Puppet Server checks for updates upon start or restart, and every 24 hours thereafter, by sending the following data to Puppet:
+        
+        * Product name
+        * Puppet Server version
+        * IP address
+        * Data collection timestamp 
+        
+        The data Puppet collects provides just one of many methods we use for learning about our community of users. The more we know about how you use Puppet, the better we can address your needs. No personally identifiable information is collected, and the data we collect is never used or shared outside Puppet. 
 
 * The `puppet-admin` section configures Puppet Server's administrative API. (This API is unavailable with Rack or WEBrick Puppet masters.)
 
--- a/documentation/config_file_puppetserver.markdown
+++ b/documentation/config_file_puppetserver.markdown
@@ -76,7 +76,7 @@ The `puppetserver.conf` file contains settings for Puppet Server software. For a
     
 * The `product` settings configure update checking and analytics data collection:
 
-    * `check_for_updates`: If set to `false`, Puppet Server will not automatically check for updates, and will not send analytics data to Puppet. 
+    * `check-for-updates`: If set to `false`, Puppet Server will not automatically check for updates, and will not send analytics data to Puppet. 
     
         If this setting is unspecified (default) or set to `true`, Puppet Server checks for updates upon start or restart, and every 24 hours thereafter, by sending the following data to Puppet:
         
@@ -85,7 +85,7 @@ The `puppetserver.conf` file contains settings for Puppet Server software. For a
         * IP address
         * Data collection timestamp 
         
-        The data Puppet collects provides just one of many methods we use for learning about our community of users. The more we know about how you use Puppet, the better we can address your needs. No personally identifiable information is collected, and the data we collect is never used or shared outside Puppet. 
+        Puppet requests this data as one of the many ways we learn about and work with our community. The more we know about how you use Puppet, the better we can address your needs. No personally identifiable information is collected, and the data we collect is never used or shared outside Puppet. 
 
 * The `puppet-admin` section configures Puppet Server's administrative API. (This API is unavailable with Rack or WEBrick Puppet masters.)
 


documentation/config_file_puppetserver.markdown
```diff
@@ -73,6 +73,19 @@ The `puppetserver.conf` file contains settings for Puppet Server software. For a
 * The `profiler` settings configure profiling:
 
     * `enabled`: If this is set to `true`, Puppet Server enables profiling for the Puppet Ruby code. The default is `false`.
+    
+* The `product` settings configure update checking and analytics data collection:
+
+    * `check_for_updates`: If set to `false`, Puppet Server will not automatically check for updates, and will not send analytics data to Puppet.
```
cprice404: @pcarlisle I have a hazy recollection of you maybe splitting this up into two different flags?  But maybe only one is worth documenting?

cprice404: OK - so probably we are just waiting for @gguillotte then.

documentation/config_file_puppetserver.markdown
```diff
@@ -73,6 +73,19 @@ The `puppetserver.conf` file contains settings for Puppet Server software. For a
...
+    * `check_for_updates`: If set to `false`, Puppet Server will not automatically check for updates, and will not send analytics data to Puppet. 
+    
+        If this setting is unspecified (default) or set to `true`, Puppet Server checks for updates upon start or restart, and every 24 hours thereafter, by sending the following data to Puppet:
+        
+        * Product name
+        * Puppet Server version
+        * IP address
+        * Data collection timestamp 
+        
+        The data Puppet collects provides just one of many methods we use for learning about our community of users. The more we know about how you use Puppet, the better we can address your needs. No personally identifiable information is collected, and the data we collect is never used or shared outside Puppet.
```
gguillotte: > The data Puppet collects provides just one of many methods we use for learning about our community of users

The phrase "data Puppet collects provides" is a little confusing, and I think we can also phrase this more actively while also clarifying it:

> Puppet requests this data as one of the many ways we learn about and work with our community.

documentation/config_file_puppetserver.markdown
gguillotte: @mindymo @cprice404 One comment left and I'm otherwise :+1:.

rlinehan: should this actually be `check-for-updates`? that's what https://github.com/puppetlabs/puppetserver/pull/1248/files seems to indicate at least.

documentation/config_file_puppetserver.markdown
```diff
@@ -73,6 +73,19 @@ The `puppetserver.conf` file contains settings for Puppet Server software. For a
 * The `profiler` settings configure profiling:
 
     * `enabled`: If this is set to `true`, Puppet Server enables profiling for the Puppet Ruby code. The default is `false`.
+    
+* The `product` settings configure update checking and analytics data collection:
```
camlow325: Would it be better to document that the `product` settings go into a `product.conf` file rather than listing it under the `puppetserver` conf section.  Doesn't seem like a puppetserver-specific thing anyway, like `webserver` and others.  Seems like for users that want to opt-out of metrics collection but are otherwise good with using the defaults for everything else that goes into the `puppetserver.conf` file that they'd be better protected against a breaking change on upgrade if they put the `product.check-for-updates` setting in a new file that won't conflict with the file in the package.  Note that in today's Puppet Server, the `product` section doesn't exist in any conf files by default.

gguillotte: Is that how it works? I wasn't aware that a separate `product.conf` file was an option, and the notes on the ticket suggested it was a `puppetserver.conf` setting. CC @briancain

camlow325: It is an option, yeah.  You can put a section in any *.conf file under the service's confdir and the Trapperkeeper app will be able to use it.  Just thought I'd bring that up because we have some flexibility on where it could go, seems like we have a lot already in `puppetserver.conf`, and that there may be some benefit to reducing upgrade pain later if it lives in a different file.

rlinehan: @gguillotte the actual file/file name that the setting is in doesn't actually matter for puppetserver. So long as the top level `product` key is somewhere in the directory, it will be picked up.

That said, I don't think we have any other files other than puppetserver.conf that we currently create by default, and IMO if we do want to separate out the `product` key then why not separate out `webserver` and `jruby-puppet` etc. etc. I think the arguments you're making for `product` could apply to the rest of those too.

camlow325: `webserver` is already separated out into a `webserver.conf` file in the packages.  Only `jruby-puppet`, `http-client`, and `profiler` are in the puppetserver.conf file by default.  It would be a bit harder to separate the latter ones out from `puppetserver.conf` now because of compatibility / upgrade implications but that may have been a better way to have done it if we could do it over again.

documentation/config_file_puppetserver.markdown
briancain: Yep, this should be `check-for-updates`.

documentation/config_file_puppetserver.markdown
gguillotte: I'm fine with breaking this out if it's an option, especially if we're not releasing today. It's just been a surprise that a separate `product.conf` file _is_ an option.

To answer question 2, a separate config file is going to bring the docs folks into it regardless. A new conf file = a new page in the docs = a new link in the nav = probably new redirects since the docs site version selector will point people to versions where that file and doc don't exist.

briancain: It seems like the headache that users will encounter from upgrades if the `product` key exists in puppetserver.conf would be pretty frustrating. I'm 👍 to having users place this setting in its own file `product.conf` within puppetservers config directory.

documentation/config_file_puppetserver.markdown
mindymo: Will fix.

documentation/config_file_puppetserver.markdown
mindymo: Good edit. Will update while I fix the underscores.

documentation/config_file_puppetserver.markdown
gguillotte: > With the amount of work you're describing would be required for docs, that doesn't seem realistic.

It's realistic if we decide today. It's maybe 1, conservatively 2 hours of docs work. It's not realistic, however, if we don't decide until tomorrow. I'm in favor of splitting it, just pick a side.

This is, uh, a weird time to say something like this, but the indecision is a bigger issue than just holding noses and picking one of the two options.

gguillotte:
@cprice404 @camlow325 @mindymo @briancain I'm :+1: to merge this, then will follow on with another PR to move the `product` section to its own doc. If we want to go that direction, great, if not, we'll have this already in the repo.

gguillotte:
Filed #1273 and a supporting puppet-docs PR that splits this PR's content into a separate doc file, adds it to the nav, and adds useful redirects from other version URLs.

If we want to split `product.conf` out into its own file, please close this PR and review #1273.

If we want to document a `product` section in `puppetserver.conf`, please close #1273 and merge this PR.

Either way, let me know what's decided and I'll pull the appropriate levers in puppet-docs land.

gguillotte:
Per discussion in these PRs and chat, I'm closing this PR and merging #1273. I am unable to merge #1273 because doing so requires an approved GH review.

PR #1262, opened 2016-11-04
puppetserver gem should honor JAVA_ARGS

m4ce:
Hi,

I bumped into the SSL issue related to SNI.

After adding -Djsse.enableSNIExtension=false to /etc/sysconfig/puppetserver, puppetserver gem install kept failing with 'handshake alert:  unrecognized_name'.

It looks like the gem wrapper script is not honoring the JAVA_ARGS defined under /etc/sysconfig/puppetserver.

### DIFF
--- a/resources/ext/cli/gem.erb
+++ b/resources/ext/cli/gem.erb
@@ -2,6 +2,6 @@
 
 umask 0022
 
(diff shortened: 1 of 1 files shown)


claytononeill:
This has caused a bug in 2.7.0, more details here: https://tickets.puppetlabs.com/browse/SERVER-1644

PR #1265, opened 2016-11-04
(MAINT) Explicitly set GEM_HOME/PATH and avoid extra/dev gem install

camlow325:
Previously, the GEM_HOME and GEM_PATH variables for installing the
bundler gem from the Rakefile for JRuby spec tests were not set.
This could cause gem content from the default system Ruby to be
pulled in during installation, leading to some odd error messages
at best or possibly failure at worst.  This commit sets the GEM_HOME and
GEM_PATH variables to the local ./vendor directory to avoid potential
collisions with a system MRI Ruby, just as is done for the later bundle
install of gems from the Puppet submodule's Gemfile for the spec run.

This commit also avoids installing the 'extra' and 'development' gems
from the Puppet Gemfile since these are unnecessary for running the
JRuby spec tests.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -48,6 +48,7 @@ namespace :spec do
       ## Line 1 launches the JRuby that we depend on via leiningen
       ## Line 2 programmatically runs 'gem install bundler' via the gem command that comes with JRuby
       gem_install_bundler = <<-CMD
+      GEM_HOME='#{TEST_GEMS_DIR}' GEM_PATH='#{TEST_GEMS_DIR}' \
       lein run -m org.jruby.Main \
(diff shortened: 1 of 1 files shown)


cprice404:
@camlow325 I'm good on this one but not sure if there is any kind of ordering necessary for the various things we have in flight now due to other CI failures, so, not clicking the button.  Feel free to merge it whenever you think it the best time.

camlow325:
@cprice404 - this one's somewhat independent of the other shenanigans going on right now.  I haven't seen any job failures related to it - just the ugly-looking warnings/errors in the output from a Travis run.  I think we could go ahead and merge this for 1.x and confirm that it doesn't cause any problems before merging it up.  I don't think it needs to get in for the run where we do the Puppet Server 2.7.0 build off of the stable branch.

PR #1266, opened 2016-11-05
(MAINT) Bump puppet submodule to 1ea537

camlow325:
Previously, JRuby spec tests were failing because of an inability to
install the public_suffix gem.  When the Puppet Gemfile was bundle
installed, version 2.5.0 of the addressable gem tried to pull in the
public_suffix gem, which requires Ruby 2.  JRuby spec tests, however,
run under MRI 1.9.3 compatibility mode, causing the install to fail.

This commit bumps the Puppet submodule up to 1ea537 and corresponding
puppet-agent package pin to 549d028.  This is being done in order to
pick up a change to the Puppet Gemfile, which pins the version of the
addressable gem to < 2.5.0.  Version 2.4.0 of the addressable gem can be
installed and used properly under MRI 1.9.3 compatibility mode.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -23,7 +23,7 @@ module PuppetServerExtensions
 
     puppet_version = get_option_value(options[:puppet_version],
                          nil, "Puppet Version", "PUPPET_VERSION",
-                         "1.6.2",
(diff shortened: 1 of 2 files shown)


camlow325:
Merging to hopefully unblock upcoming release activities.

PR #1270, opened 2016-11-08
(docs) Add Puppet Server 2.7 release notes

gguillotte:
Hold until questions about documenting `check-for-updates` are resolved in #1260.

### DIFF
--- a/documentation/_puppetserver_nav.html
+++ b/documentation/_puppetserver_nav.html
@@ -19,6 +19,7 @@
           <li><a href="/puppetserver/{{ server_version }}/config_file_global.html">global.conf: Trapperkeeper Settings</a></li>
           <li><a href="/puppetserver/{{ server_version }}/config_file_ca.html">ca.conf: CA Service Access Control (deprecated)</a></li>
           <li><a href="/puppetserver/{{ server_version }}/config_file_master.html">master.conf: Authorization by HTTP Header (deprecated)</a></li>
+          <li><a href="/puppetserver/{{ server_version }}/config_file_product.html">product.conf: Configuring Product-level Interactions (optional)</a></li>
           <li><a href="/puppetserver/{{ server_version }}/config_file_logbackxml.html">logback.xml: Logging Level and Location</a></li>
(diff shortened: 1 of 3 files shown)


documentation/release_notes.markdown
```diff
@@ -7,9 +7,126 @@ canonical: "/puppetserver/latest/release_notes.html"
...
+Exception in thread "main" org.jruby.exceptions.RaiseException: (Errno::EACCES) /opt/puppetlabs/server/data/puppetserver/jruby-gems/specifications/trollop-2.1.2.gemspec
+```
+
+Puppet Server 2.7 resolves this issue by explicitly setting a umask of 0022 when running any `puppetserver gem` subcommand. This ensures that `puppetserver` can use any gems installed by the `gem` subcommand at run-time.
+
+-   [SERVER-1601](https://tickets.puppetlabs.com/browse/SERVER-1601)
+
+### Other new features
+
+-   [SERVER-1585](https://tickets.puppetlabs.com/browse/SERVER-1585): Set JRuby `compat-version` in [`puppetserver.conf`][puppetserver.conf].
```
camlow325: Might be worth wordsmithing this a little more.  This adds the ability for users to configure Puppet Server to experimentally run under an MRI 2.0 compatibility mode, rather than just the default 1.9.3 mode.  I think it's worth calling out the "2.0" version in this description somewhere.  Since several users have been asking about it, I'd also probably move this up toward the top of the list of new stuff for better visibility

documentation/release_notes.markdown
```diff
@@ -7,9 +7,126 @@ canonical: "/puppetserver/latest/release_notes.html"
...
+-   [SERVER-1589](https://tickets.puppetlabs.com/browse/SERVER-1589): * Use `.gz` extensions for Puppet Server log file archives, and rotate them when they reach 200MB in size instead of 10MB.
```
camlow325: Maybe remove the asterisk from before "Use '.gz'" here?

camlow325:
Latest changes look good to me.  @jpinsonault - WDYT?  Suppose we should wait to merge until someone makes the final call on the "product.conf vs puppetserver.conf" decision for the `product.check-for-updates` setting.

PR #1271, opened 2016-11-08
(MAINT) Drop heap size for Travis testing to 5 GB

camlow325:
This commit drops the heap size used for testing in Travis from 6 GB to
5 GB.  This is being done because JDK 8 builds under Travis are
frequently failing with a 'Cannot allocate memory' error, which may be
related to the heap size being too close to the total RAM available on
Travis CI nodes.

### DIFF
--- a/ext/travisci/test.sh
+++ b/ext/travisci/test.sh
@@ -2,7 +2,7 @@
 
 set -e
 
-export PUPPETSERVER_HEAP_SIZE=6G
+export PUPPETSERVER_HEAP_SIZE=5G
 
 echo "Using heap size: $PUPPETSERVER_HEAP_SIZE"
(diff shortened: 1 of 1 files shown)


camlow325:
Please don't merge this until I've had a chance to see how a few Travis runs will work with it.

camlow325:
I've done a few Travis runs with this change and the tests have passed for all JDKs for each.  I'm good with this being merged in whenever now.

cprice404:
@camlow325 since you have this targeted at `stable`, do you think there's any reason we should wait until the release is out the door before merging?  Probably not, and if not, we can go ahead and push the button, but just wanted to check.

camlow325:
@cprice404 I think we're good merging it now but happy to wait until 2.7.0 goes live if you're worried.

PR #1273, opened 2016-11-09
(docs) Add product.conf doc

gguillotte:
-   Adds a new `config_file_product.markdown` doc for a `product.conf` file to configure/disable update checks and analytics (see discussion on #1260)
-   Adds a link to this new doc on the `configuration.markdown` doc and aux nav.

If we merge this, also merge updates to puppet-docs and either revert or don't merge #1260.


briancain:
I'm good with this. 👍

PR #1275, opened 2016-11-09
(SERVER-1580) Introduce `dev-setup`

cprice404:
This PR brings us into alignment with the Clojure dev environment best practices discussed in recent e-mail threads.  Specifically:

* Switch the default location for all of the config/state directories to be under `~/.puppetlabs` and match the defaults used by puppet agent
* Switch the default REPL namespace to `dev-tools`
* Add ability to override REPL config settings via `get-config` in `user.clj`
* Add a `dev-setup` script to initialize the dev environment, including:
  * Install vendored gems
  * Create necessary directories under `~/.puppetlabs`
  * Create a default `puppet.conf` file that sets up `certname` and `server` settings, honoring `MASTERHOST` environment variable if present.


cprice404:
Hmm, probably should update the docs for this stuff.  Will add a commit.

cprice404:
Added a commit with docs.  This should be good to go.

dev/dev_tools.clj
```diff
@@ -21,25 +21,25 @@
...
-      (println "Copying puppetserver.conf.sample to" conf-dir)
-      (fs/copy "./dev/puppetserver.conf.sample" default-conf-file-dest))
-    (ks/absolute-path default-conf-file-dest)))
+  (config/load-config "./dev/puppetserver.conf"))
+
+(defn get-config
+  "This function is required by the Puppet clojure repo best practices guide.  It
+  will be called during the bootstrapping of the application via `go`, and will
+  be passed as an argument a copy of the default configuration map for the dev
+  environment for the app; users may override the implementation function in their
```
camlow325: The ability to override selective settings from the default is pretty cool.  👍

documentation/dev_running_from_source.markdown
```diff
@@ -37,17 +19,22 @@ Step 0: Quick Start for Developers
...
     # Run Puppet Server
-    user=> (go)
-    user=> (help)
+    dev-tools=> (go)
+    dev-tools=> (help)
 
-You should now have a running server.  Your confdir is set to `./target/master-conf`;
-see ~/.puppetserver/puppetserver.conf to examine this and other default paths.
+You should now have a running server.  All relevant paths (`$confdir`, `$codedir`,
+etc.) are configured by default to point to directories underneath `~.puppetlabs`.
```
camlow325: nit, could say `~/.puppetlabs`.

dev/puppetserver.conf
```diff
@@ -3,7 +3,7 @@ global: {
 }
 
 product: {
-    update-server-url: "http://localhost/"
+    update-server-url: "http://some.update.server.that.does.not.exist/"
```
cprice404: Mostly what I was going for here is for the error message to contain something that made it obvious what the stack trace was about.

dev/puppetserver.conf
```diff
@@ -46,27 +46,27 @@ jruby-puppet: {
 
     # This setting determines where JRuby will install gems.  It is used for loading gems,
     # and also by the `puppetserver gem` command line tool.
-    gem-home: ./target/jruby-gems
+    gem-home: ${HOME}"/.puppetlabs/opt/server/data/puppetserver/jruby-gems"
```
cprice404: https://github.com/typesafehub/config/blob/master/HOCON.md#substitutions

Yeah, this is just how their syntax works.  It won't matter if there are special chars in their HOME var though; the interpolation happens in memory way after this file is parsed, and this is effectively just a string concatenation operation.

project.clj
```diff
@@ -32,7 +32,7 @@
 
   :min-lein-version "2.7.1"
 
-  :parent-project {:coords [puppetlabs/clj-parent "0.1.7"]
+  :parent-project {:coords [puppetlabs/clj-parent "0.2.0"]
```
cprice404: Yeah.

dev/dev_tools.clj
```diff
@@ -1,4 +1,4 @@
-(ns user-repl
+(ns dev-tools
```
cprice404: ugh, that sucks.  separate ticket seems right, though.

documentation/dev_running_from_source.markdown
cprice404: good catch.

cprice404:
Added commits that I think should address the current PR feedback.

dev/dev_tools.clj
camlow325: Sounds good.  I filed [SERVER-1643](https://tickets.puppetlabs.com/browse/SERVER-1643).

cprice404:
@puppetlabs/discuss-team-puppetserver-maint this could use a second review (probably worth checking it out locally and playing with it) if anyone gets a chance.

camlow325:
Since we have a couple of approvals now, I'm going to merge this.

PR #1277, opened 2016-11-11
(maint) Bump EZBake to latest version

melissa:


camlow325:
@melissa - any reason you can think of why we shouldn't go ahead and merge this?

melissa:
None that I can think of. I know puppetserver builds successfully with this change. I have not run tests though

PR #1278, opened 2016-11-14
Correcting comments to use HTTPS rather than HTTP

ivernus:
Puppet clients use HTTPS rather than HTTP, fixing comments.


cprice404:
Thanks!

PR #1280, opened 2016-11-15
(MAINT) Skip agent_run_before_ssl_files_inited test on PE

camlow325:
Previously, the `master_starts_if_agent_run_before_ssl_files_inited`
acceptance test was configured to run both for OSS and PE Puppet Server.
The test nukes the ssl directory so that an agent run can reconfigure
just the public and private key before the server is started up.  For
PE, the `ssl_crl_path` is configured at installation time.  The absence
of the CRL file after the test nukes it causes the server to fail to
startup.  Work to be done for SERVER-346 and SERVER-911 would allow for
the CRL file to properly be replaced at startup time during this
scenario.  Since that work has not yet been done, though, this commit
just skips the test when run under PE for now.

### DIFF
--- a/acceptance/suites/tests/00_smoke/master_starts_if_agent_run_before_ssl_files_inited.rb
+++ b/acceptance/suites/tests/00_smoke/master_starts_if_agent_run_before_ssl_files_inited.rb
@@ -1,5 +1,9 @@
 test_name "Master can startup even if agent run before ssl files inited"
 
+skip_test 'Skipping for PE since it pre-configures a CRL file and Puppet ' \
+  'Server does not yet update it at startup - SERVER-346 and SERVER-911' \
+  if options[:type] == 'pe'
+
 puppetservice=options['puppetservice']
 ssldir = master.puppet['ssldir']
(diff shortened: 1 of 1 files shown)


jpinsonault:
If I understand this correctly, you're saying it's ok to skip this because this isn't a situation PE usually finds itself because of the installer managing the SSL stuff?

camlow325:
Yeah, PE will pre-configure all of the cert artifacts before the server is ever started so the "agent run before master first started" situation isn't something that should ever come up on PE.  It is possible in PE that someone might nuke their SSL directory in order to have everything regenerated the next time the server starts up.  In that case, the server would fail to startup in the way this test is failing.  To fix that problem, though, I think we'd need to do the work in SERVER-346 and SERVER-911, which I keep finding a hard time justifying to others that we need to do.  So for now, I'm proposing we just disable this test for PE.

camlow325:
Assuming this is approved, I can at least add a comment to SERVER-346 and SERVER-911 about looking into re-enabling this test for PE if those ever do get fixed.

jpinsonault:
Cool, thanks for the details. I'm 👍 on this

PR #1281, opened 2016-11-15
(SERVER-1629) Bump beaker gem dependency to custom 3.3.0+reload

camlow325:
This commit bumps the beaker gem dependency in Puppet Server to a custom
version of beaker which is based on 3.3.0 but adds on support for
restarting the service via a `reload` action, when possible.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ gem 'jira-ruby', :group => :development
 
 group :test do
   gem 'rspec'
-  gem 'beaker', *location_for(ENV['BEAKER_VERSION'] || '~> 2.50.0')
+  gem 'beaker', :git => 'https://github.com/camlow325/beaker', :ref => '3a9f3c'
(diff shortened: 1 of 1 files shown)


camlow325:
This should not be merged until after https://github.com/puppetlabs/ci-job-configs/pull/2030 is merged since we need to have the Jenkins slave node which runs beaker be on MRI Ruby 2.2.5 or later whereas it is on 2.1.1 today.

camlow325:
I've done local runs of the full Puppet acceptance test suite against a Cent 7 VM and have seen all of the tests pass.  I'd like to get some experience using it in the full matrix Puppet Server acceptance pipeline to see if any issues might come up.  I'm hoping this would give the Q* teams more confidence about pulling this change into Beaker.  If that all goes well, I'd put up a PR with changes to the beaker project and revert our Puppet Server beaker dependency back to a released 3.x version.  We could then upgrade our beaker dependency to a newer 3.x version later that adds on the reload support whenever that might become available.

camlow325:
ci-job-configs PR has been merged and the pipeline is passing on MRI Ruby 2.3.1 now, so this should be ready for merge.

PR #1283, opened 2016-11-15
(MAINT) Remove 'test/unit/assertions' requires from subcommand tests

camlow325:
Previously, many of the subcommand acceptance tests explicitly required
the 'test/unit/assertions' path.  When running under Ruby 2.3.1 with
only the minitest gem (and not the test-unit gem) installed, this
path does not exist.  This causes the require and, therefore, the tests
to fail.

The minitest gem is pulled in by the beaker gem and defines the
assert_* functions that the tests use.  Rather than either adding the
'test-unit' gem to the Gemfile or changing the tests to require a valid
path in minitest instead, this commit just removes the
'test/unit/assertions' requires since the path should have already been
loaded when running under beaker anyway.  No other tests include require
statements to locate the assert_* functions, so this commit allows the
subcommand tests to follow the same pattern.

### DIFF
--- a/acceptance/suites/tests/010-puppetserver-cli/subcommand/SERVER-297_common_behavior.rb
+++ b/acceptance/suites/tests/010-puppetserver-cli/subcommand/SERVER-297_common_behavior.rb
@@ -1,4 +1,3 @@
-require 'test/unit/assertions'
 require 'json'
 
 test_name "Puppetserver subcommand consolidated ENV handling tests."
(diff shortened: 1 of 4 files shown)


camlow325:
I ran all of the subcommand tests locally with beaker running under Ruby 2.3.1 and the latest Puppet Server package build off the stable branch installed on a CentOS 7 VM.  All of the tests passed.

justinstoller:
@camlow325 did you try running this with Ruby 1.9? Is that still a supported Ruby version for running acceptance tests?

camlow325:
@justinstoller We only run the acceptance tests with beaker on a node that is using Ruby 2.3.1 now.  1.9 already hasn't been working for a while.  This allows us to pull in some newer stuff that CI.next, etc. needs that has a minimum Ruby 2.x dependency.

jpinsonault:
👍

PR #1285, opened 2016-11-16
(SERVER-1644) Use JAVA_ARGS_CLI env var for Java args in subcommands

camlow325:
Previously, the value of the JAVA_ARGS environment variable, as a result
of work done for the Puppet Server 2.7.0 release, was passed along as
arguments to the Java process running the `gem` subcommand.  This was
problematic for cases where the JAVA_ARGS had been specifically tuned
with arguments for use when running the puppetserver service but
inappropriate for the `gem` command.  For example, Java args that enable
debug logging when included in the `gem` command could break the ability
to perform basic `gem` installs.

This commit changes the `gem` subcommand to source custom Java arguments
from a separate environment variable, `JAVA_ARGS_CLI`.  This commit also
adds the use of the `JAVA_ARGS_CLI` args to the ruby and irb subcommands,
which previously provided no mechanism for specifying custom Java args.

### DIFF
--- a/acceptance/suites/tests/010-puppetserver-cli/subcommand/gem.rb
+++ b/acceptance/suites/tests/010-puppetserver-cli/subcommand/gem.rb
@@ -84,3 +84,9 @@ end
 
 step "Verify that gem env operates"
 on(master, "#{cli} gem env", :acceptable_exit_codes => [0])
+
+step "Verify that Java cli args passed through to gem command"
+on(master, "JAVA_ARGS_CLI=-Djruby.cli.version=true #{cli} gem help") do
+  assert_match(/jruby \d\.\d\.\d.*$/, stdout,
+               'jruby version not included in gem command output')
+end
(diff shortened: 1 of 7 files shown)


cprice404:
@camlow325 if someone set this variable in the defaults file, that would work too, right?  If so, should we update the docs to reflect that?  And also maybe include a commented-out example line in the packaged defaults file?  I'm guessing that most users who have an issue that warrants this would prefer to set the value once, in that file?

camlow325:
Yeah, should work from the sysconfig/defaults file, too.  I don't think I could put a commented-out example line in the packaged file w/o changing ezbake, though, right?  Not sure we should do that part.

cprice404:
Ah, crud, was thinking that file was local to puppetserver.  Agree that we shouldn't put that into ezbake.  So I'm good with just mentioning it in your docs here.

camlow325:
@cprice404 I added a commit, 3b936ec, which talks about setting the option via the sysconfig or default file.  I also tried it out on a VM and confirmed that it works.

cprice404:
@gguillotte if you have time/interest in taking a peek at the docs changes in here before we merge it, I can wait - otherwise we can handle that in subsequent PRs leading up to the release?

gguillotte:
@cprice404 LGTM :+1:

PR #1289, opened 2016-11-17
(SERVER-377) Add tests for proxy-related environment variables

rlinehan:
Changes to jruby-utils make it so that the environment variables `HTTP_PROXY`,
`http_proxy`, `HTTPS_PROXY`, `https_proxy`, `NO_PROXY`, and `no_proxy` are now
passed through to the jruby scripting container, so that gems can be installed
in scenarios where a proxy is set.

This commit adds acceptance tests ensuring that when set those variables are
available when running `puppetserver ruby` and `puppetserver irb` (and thus
will also be available when running `puppetserver gem`).

Note that this cannot be merged until https://github.com/puppetlabs/jruby-utils/pull/32 has been merged and a new version of jruby-utils released.

### DIFF
--- a/acceptance/suites/tests/010-puppetserver-cli/subcommand/SERVER-297_common_behavior.rb
+++ b/acceptance/suites/tests/010-puppetserver-cli/subcommand/SERVER-297_common_behavior.rb
@@ -3,6 +3,8 @@ require 'json'
 
 test_name "Puppetserver subcommand consolidated ENV handling tests."
 
+proxy_env_vars = "HTTP_PROXY=foo http_proxy=foo HTTPS_PROXY=foo https_proxy=foo NO_PROXY=foo no_proxy=foo"
+
 step "ruby: Check that PATH, HOME, GEM_HOME JARS_REQUIRE and JARS_NO_REQUIRE are present"
 on(master, "puppetserver ruby -rjson -e 'puts JSON.pretty_generate(ENV.to_hash)'") do
   env = JSON.parse(stdout)
@@ -13,6 +15,18 @@ on(master, "puppetserver ruby -rjson -e 'puts JSON.pretty_generate(ENV.to_hash)'
   assert(env['JARS_NO_REQUIRE'], "JARS_NO_REQUIRE missing")
 end
 
+step "ruby: Check that proxy env-variables are present"
+on(master, "#{proxy_env_vars} puppetserver ruby -rjson -e 'puts JSON.pretty_generate(ENV.to_hash)'") do
+  env = JSON.parse(stdout)
+  assert(env['HTTP_PROXY'], "HTTP_PROXY is missing")
+  assert(env['http_proxy'], "http_proxy is missing")
+  assert(env['HTTPS_PROXY'], "HTTPS_PROXY is missing")
+  assert(env['https_proxy'], "https_proxy is missing")
+  assert(env['NO_PROXY'], "NO_PROXY is missing")
+  assert(env['no_proxy'], "no_proxy is missing")
+  assert_nil(env['RANDOM_VARIABLE'], "RANDOM_VARIABLE is present and should not be")
+end
+
 step "irb: Check that PATH, HOME, GEM_HOME JARS_REQUIRE and JARS_NO_REQUIRE are present"
 on(master, "echo 'puts JSON.pretty_generate(ENV.to_hash)' | puppetserver irb -f -rjson") do
   assert_match(/\bPATH\b/, stdout, "PATH missing")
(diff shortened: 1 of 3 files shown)


acceptance/suites/tests/010-puppetserver-cli/subcommand/SERVER-297_common_behavior.rb
```diff
@@ -13,6 +15,18 @@
   assert(env['JARS_NO_REQUIRE'], "JARS_NO_REQUIRE missing")
 end
 
+step "ruby: Check that proxy env-variables are present"
+on(master, "#{proxy_env_vars} puppetserver ruby -rjson -e 'puts JSON.pretty_generate(ENV.to_hash)'") do
+  env = JSON.parse(stdout)
+  assert(env['HTTP_PROXY'], "HTTP_PROXY is missing")
```
camlow325: I think it would be good to have each of these assertions validate that the actual value of the environment variable was passed through properly, e.g., that `env['HTTP_PROXY']` is equal to "foo".  WDYT?

acceptance/suites/tests/010-puppetserver-cli/subcommand/SERVER-297_common_behavior.rb
```diff
@@ -13,6 +15,18 @@
...
+  assert(env['http_proxy'], "http_proxy is missing")
+  assert(env['HTTPS_PROXY'], "HTTPS_PROXY is missing")
+  assert(env['https_proxy'], "https_proxy is missing")
+  assert(env['NO_PROXY'], "NO_PROXY is missing")
+  assert(env['no_proxy'], "no_proxy is missing")
+  assert_nil(env['RANDOM_VARIABLE'], "RANDOM_VARIABLE is present and should not be")
```
camlow325: Not a big deal here, but not sure if it's worth even having the `RANDOM_VARIABLE` assertion here.

test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
```diff
@@ -450,6 +450,9 @@
             jruby-scripting-container (:scripting-container jruby-instance)
             jruby-env (.runScriptlet jruby-scripting-container "ENV")]
         (try
+         ;; Note that if this test is run in an environment where the
+         ;; variables HTTP_PROXY, HTTPS_PROXY, NO_PROXY, or their lowercase
+         ;; variants are set, this will fail.
           (is (= #{"HOME" "PATH" "GEM_HOME" "GEM_PATH"
```
camlow325: Would it make sense to just code this assertion so it would work either way, e.g., by filtering out any of the http/no proxy keys from `jruby-env` that may be present first before doing the equality comparison for the other keys we had expected to be passed through?

acceptance/suites/tests/010-puppetserver-cli/subcommand/SERVER-297_common_behavior.rb
```diff
@@ -21,3 +35,13 @@
   assert_match(/\bJARS_REQUIRE\b/, stdout, "JARS_REQUIRE missing")
   assert_match(/\bJARS_NO_REQUIRE\b/, stdout, "JARS_NO_REQUIRE missing")
 end
+
+step "irb: Check that proxy env-variables are present"
+on(master, "echo 'puts JSON.pretty_generate(ENV.to_hash)' | #{proxy_env_vars} puppetserver irb -f -rjson") do
+  assert_match(/\bHTTP_PROXY\b/, stdout, "HTTP_PROXY missing")
```
camlow325: Same comment here on how I think it would be good to assert that the expected value was passed through for each variable.

jpinsonault:
Pretty much what jeremy said. Looks good, maybe just tweak a few things, make the clojure test more robust

test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
camlow325: I think we still want to fail if other environment variables that we're not explicitly intending to filter out end up get passed through, though.  The equality comparison is still useful.  I just think it would be good to ignore the http proxy-related ones since they may or may not exist in whatever environment the tests are run from.

jpinsonault: oh good point 👍

acceptance/suites/tests/010-puppetserver-cli/subcommand/SERVER-297_common_behavior.rb
rlinehan: I think my plan had been to add that variable as an environment variable in the list and then assert that it wasn't present, since it doesn't seem like we have anything that tests that we're not just allowing all variables through. But I didn't include half of the behavior necessary to make such a test useful. Also I don't think that that should necessarily be part of this test?

camlow325: There are some tests in the other subcommand specs that look at this, for example in the ruby.rb specs.  If you also want to include one in the 'common_behavior' tests for that, I'd be good with it.  I'd agree, though, that maybe having it as a separate test step that sets `RANDOM_VARIABLE` without whitelisting it and confirming that it gets filtered out would be better.  Or if you just want to remove this line and not replace it with anything else, I'm good with that, too.

rlinehan:
Updated. This should now be ready for merge.

camlow325:
Looks like @jpinsonault was already basically good with this one so I'm going to merge it.

PR #1292, opened 2016-11-23
(MAINT) Pin tests to puppet-agent 1.8.1

jpinsonault:
CI has begun failing because the sha we were previously pinned to no longer
has artifacts on the builds server. 1.8.1 contains the changes we needed from the old sha

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -23,7 +23,7 @@ module PuppetServerExtensions
 
     puppet_version = get_option_value(options[:puppet_version],
                          nil, "Puppet Version", "PUPPET_VERSION",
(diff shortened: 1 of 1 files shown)


acceptance/lib/helper.rb
```diff
@@ -32,7 +32,7 @@ def self.initialize_config(options)
     puppet_build_version = get_option_value(options[:puppet_build_version],
                          nil, "Puppet Agent Development Build Version",
                          "PUPPET_BUILD_VERSION",
-                         "549d0283c85bc100205e9c5de251219db64ee698",
+                         "1.8.1",
```
cprice404: I think maybe we should also update the "PUPPET_VERSION" above?

PR #1296, opened 2016-11-30
(SERVER-1670) Bump lein-ezbake to fix dependency issues

justinstoller:
This should pull in the resolution to SERVER-1670 (autorequires causing system ruby to be an rpm dependency).


camlow325:
Since this is just an ezbake bump, shouldn't affect any Travis tests.  I'm going to go ahead and merge this.

PR #1298, opened 2016-12-01
(SERVER-1629) Bump beaker pin to 3.5.0

camlow325:
This commit bumps the beaker pin for puppetserver acceptance testing to
~>3.5.0.  This allows us to utilize the 'reload' capability in
puppetserver for performance improvements when running acceptance tests.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ gem 'jira-ruby', :group => :development
 
 group :test do
   gem 'rspec'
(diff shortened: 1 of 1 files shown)


camlow325:
@justinstoller - fyi, I did a local run with the Puppet acceptance tests against a vmpooler instance with this change and the tests all pass.  So this should be good.

justinstoller:
Kay, I was going to wait until CI greened up, but I suppose this should help green it up sooner.

PR #1299, opened 2016-12-01
Merge branch 'stable' into master at 8564f98

camlow325:
* stable:
      (SERVER-1593) Update gettext-setup gem to 0.8
      (SERVER-1629) Bump beaker pin to 3.5.0
      (SERVER-1670) Bump lein-ezbake to fix dependency issues

    Conflicts:
    - project.clj - Kept most dependency versions already on master since
      they were newer than the ones on stable.  Only exception was bumping
      to the newer lein-ezbake from stable (1.1.5) vs. the one previously in
      master (1.1.4).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ gem 'jira-ruby', :group => :development
 
 group :test do
   gem 'rspec'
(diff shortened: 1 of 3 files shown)


jpinsonault:
👍

PR #1306, opened 2016-12-16
(SERVER-1475) Configure an slf4j logger for JRuby

camlow325:
Previously, any messages being logged by JRuby under Puppet Server would
be written to JRuby's default stderr logger, which is most commonly
routed to syslog / journal for the running service.  Also, debug-level
JRuby log messages would always be suppressed by the stderr logger.

Starting with this commit, Puppet Server configures an slf4j logger,
provided by jruby-utils, which allows JRuby log messages to be handled
by the logback framework, as is most other logging done under Puppet
Server (via Trapperkeeper).  The slf4j logger also enables debug
messages to be logged, depending upon the log level configured in the
logback.xml file.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -58,7 +58,7 @@
                  ;; in different versions of the three different logback artifacts
                  [net.logstash.logback/logstash-logback-encoder]
 
-                 [puppetlabs/jruby-utils "0.5.0"]
+                 [puppetlabs/jruby-utils "0.6.0"]
                  [puppetlabs/trapperkeeper]
                  [puppetlabs/trapperkeeper-authorization]
                  [puppetlabs/trapperkeeper-scheduler]
@@ -90,7 +90,8 @@
   :lein-ezbake {:vars {:user "puppet"
                        :group "puppet"
                        :build-type "foss"
-                       :java-args "-Xms2g -Xmx2g -XX:MaxPermSize=256m"
+                       :java-args ~(str "-Xms2g -Xmx2g -XX:MaxPermSize=256m "
+                                     "-Djruby.logger.class=com.puppetlabs.jruby_utils.jruby.Slf4jLogger")
                        :repo-target "PC1"
                        :bootstrap-source :services-d
                        :logrotate-enabled false}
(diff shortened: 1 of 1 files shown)


camlow325:
This PR will not pass Travis currently.  It is dependent upon https://github.com/puppetlabs/jruby-utils/pull/35 and a subsequent jruby-utils release.  Putting it up just for review for now.

camlow325:
I did some manual testing of this with an ezbake package of Puppet Server just to see that some JRuby log messages would be handled by logback.

For this, I did the following:

1) Set `jruby-puppet.compile-mode` to `jit` in the "/etc/puppetlabs/puppetserver/conf.d/puppetserver.conf" file.

2) Restarted the puppetserver service.

3) Observed that a number of messages like this were written to the "/var/log/puppetlabs/puppetserver.log" file:

> 2016-12-16 12:37:43,036 INFO  [Ruby-0-JIT-1] [j.JITCompiler] done jitting:<anon class>.Puppet::Util::Storage.cache at /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/util/storage.rb:2

Without the `-Djruby.logger.class` option being configured in the "/etc/sysconfig/puppetserver" file, the above messages would have appeared in the journal instead of the "puppetserver.log" file.

camlow325:
I "could" write a beaker test which does something like the manual testing that I did but it seems like it could be pretty fragile since we don't control the formatting / nature of any log messages being written by JRuby.  We could also do something more artificial with the one of the CLI commands.  This works, for example:

> $ JAVA_ARGS_CLI="-Djruby.logger.class=com.puppetlabs.jruby_utils.jruby.Slf4jLogger" puppetserver ruby -rjava -e "org.jruby.util.log.LoggerFactory.getLogger('mylogger').debug('debug this')
13:03:53.926 [main] DEBUG jruby.mylogger - debug this

The CLI args, though, may not be in sync with the `JAVA_ARGS` for the running process, though, so this probably wouldn't be very robust either.

camlow325:
@cprice404, @rlinehan - this should be good to go for merging now since it's using a jruby-utils release, 0.6.0, which has the new slf4j logger class in it.

PR #1307, opened 2016-12-16
(MAINT) Upgrade to clj-parent 0.3.0.

cprice404:


project.clj
```diff
@@ -84,7 +84,7 @@
                  ["snapshots" "http://nexus.delivery.puppetlabs.net/content/repositories/snapshots/"]]
 
   :plugins [[lein-parent "0.3.1"]
-            [puppetlabs/i18n "0.4.3"]]
+            [puppetlabs/i18n "0.5.1"]]
```
rlinehan: Do we need to reinitialize the i18n scripts or anything?

cprice404: I don't think so... and I think that, going forward, that kind of thing is hopefully supposed to be handled by the CI workflow.  But we can always update in a subsequent PR if we need to.

camlow325: @rlinehan Are you good with this change?  Wasn't sure if you had more insight here than the rest of us do.

PR #940, opened 2016-03-04
(SERVER-1114) environment_classes API documentation

camlow325:
This PR adds a markdown page and schema with API documentation for
the environment_classes endpoint.

This PR also includes new documentation for the
`environment-class-cache-enabled` setting.

PR #1310, opened 2017-01-05
(SERVER-1696) Add new environment_modules v3 route

briancain:
This commit adds new endpoint into puppet-server: environment_modules
When a user makes a request against this endpoint with a given
envirnoment, puppet-server will use JRuby to determine what modules are
installed within puppet.

### DIFF
--- a/documentation/puppet-api/v3/environment_modules.json
+++ b/documentation/puppet-api/v3/environment_modules.json
@@ -1,0 +1,33 @@
+{
+  "$schema":     "http://json-schema.org/draft-04/schema#",
+  "title":       "Environment Modules",
+  "description": "Information about the modules in a Puppet code environment",
+  "type":        "object",
+  "properties": {
+    "modules": {
+      "description": "The array of modules which exist in an environment.",
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "name": {
+            "description": "The name of the puppet module",
+            "type": "string"
+          },
+          "version": {
+            "description": "The version of the puppet module",
+            "type": "string"
+          }
+        },
+        "required": ["name", "version"],
+        "additionalProperties": false
+      }
+    },
+    "name": {
+      "description": "Name of the environment",
+      "type": "string"
+    }
+  },
+  "required": ["modules", "name"],
+  "additionalProperties": false
+}
--- a/documentation/puppet-api/v3/environment_modules.markdown
+++ b/documentation/puppet-api/v3/environment_modules.markdown
@@ -1,0 +1,127 @@
+---
+layout: default
+title: "Puppet Server: Puppet API: Environment Modules"
+canonical: "/puppetserver/latest/puppet-api/v3/environment_modules.html"
+---
+
+[deprecated WEBrick Puppet master]: https://docs.puppet.com/puppet/latest/reference/services_master_webrick.html
+[`auth.conf`]: ../../config_file_auth.markdown
+[`puppetserver.conf`]: ../../config_file_puppetserver.markdown
+
+The environment modules API will return information about what modules are
+installed for the requested environment.
+
+This endpoint is available only when the Puppet master is running Puppet Server, not
+on Ruby Puppet masters, such as the [deprecated WEBrick Puppet master][]. It also ignores
+the Ruby-based Puppet master's authorization methods and configuration. See the
+[Authorization section](#authorization) for more information.
+
+## `GET /puppet/v3/environment_modules?environment=:environment`
+
+Making a request with no query parameters is not supported and returns an HTTP 400 (Bad
+Request) response.
+
+### Supported HTTP Methods
+
+GET
+
+### Supported Formats
+
+JSON
+
+### Query Parameters
+
+Provide one parameter to the GET request:
+
+* `environment`: Request information about modules pertaining to the specified
+environment only.
+
+### Responses
+
+#### GET request with results
+
+```
+GET /puppet/v3/environment_modules?environment=env
+
+HTTP/1.1 200 OK
+Content-Type: application/json
+
+{
+    "modules": [
+        {
+            "name": "puppetlabs/ntp",
+            "version": "6.0.0"
+        },
+        {
+            "name": "puppetlabs/stdlib",
+            "version": "4.14.0"
+        }
+    ],
+    "name": "env"
+}
+```
+
+#### Environment does not exist
+
+If you send a request with an environment parameter that doesn't correspond to the name of a
+directory environment on the server, the server returns an HTTP 404 (Not Found) error:
+
+```
+GET /puppet/v3/environment_modules?environment=doesnotexist
+
+HTTP/1.1 404 Not Found
+
+Could not find environment 'doesnotexist'
+```
+
+#### No environment given
+
+```
+GET /puppet/v3/environment_modules
+
+HTTP/1.1 400 Bad Request
+
+An environment parameter must be specified
+```
+
+#### Environment parameter specified with no value
+
+```
+GET /puppet/v3/environment_modules?environment=
+
+HTTP/1.1 400 Bad Request
+
+The environment must be purely alphanumeric, not ''
+```
+
+#### Environment includes non-alphanumeric characters
+
+If the environment parameter in your request includes any characters that are
+not `A-Z`, `a-z`, `0-9`, or `_` (underscore), the server returns an HTTP 400 (Bad Request) error:
+
+```
+GET /puppet/v3/environment_modules?environment=bog|us
+
+HTTP/1.1 400 Bad Request
+
+The environment must be purely alphanumeric, not 'bog|us'
+```
+
+### Schema
+
+An environment modules response body conforms to the
+[environment modules schema](./environment_modules.json).
+
+### Authorization
+
+Unlike other Puppet master service-based API endpoints, the environment modules API is
+provided exclusively by Puppet Server. All requests made to the environment
+modules API are authorized using the Trapperkeeper-based [`auth.conf`][] feature
+introduced in Puppet Server 2.2.0, and ignores the older Ruby-based authorization process
+and configuration. The value of the `use-legacy-auth-conf` setting in the `jruby-puppet`
+configuration section of [`puppetserver.conf`][] is ignored for requests
+to the environment modules API, because the Ruby-based authorization process is not equipped to
+authorize these requests.
+
+For more information about the Puppet Server authorization process and configuration
+settings, see the [`auth.conf` documentation][`auth.conf`].
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -1,7 +1,7 @@
 (ns puppetlabs.services.master.master-core
   (:import (java.io FileInputStream)
            (clojure.lang IFn)
-           (java.util Map Map$Entry)
+           (java.util List Map Map$Entry)
            (org.jruby RubySymbol)
            (org.eclipse.jetty.util URIUtil))
   (:require [me.raynes.fs :as fs]
@@ -124,6 +124,17 @@
   {:name schema/Str
    :files [EnvironmentClassesFileEntry]})
 
+(def EnvironmentModuleInfo
+  "Schema for a given module that can be returned within the
+  :modules key in EnvironmentModulesInfo."
+  {:name schema/Str
+   :version schema/Str})
+
+(def EnvironmentModulesInfo
+  "Schema for the return payload for an environment_classes request."
+  {:name schema/Str
+   :modules [EnvironmentModuleInfo]})
+
 (defn obj-or-ruby-symbol-as-string
   "If the supplied object is of type RubySymbol, returns a string
   representation of the RubySymbol.  Otherwise, just returns the original
@@ -142,7 +153,7 @@
     (throw (IllegalArgumentException.
             (i18n/trs "Object cannot be coerced to a keyword: {0}" obj)))))
 
-(defn sort-nested-environment-class-info-maps
+(defn sort-nested-info-maps
   "For a data structure, recursively sort any nested maps and sets descending
   into map values, lists, vectors and set members as well. The result should be
   that all maps in the data structure become explicitly sorted with natural
@@ -171,10 +182,10 @@
     (instance? Map data)
     (into (sorted-map) (for [[k v] data]
                          [(obj->keyword k)
-                          (sort-nested-environment-class-info-maps v)]))
+                          (sort-nested-info-maps v)]))
 
     (instance? Iterable data)
-    (map sort-nested-environment-class-info-maps data)
+    (map sort-nested-info-maps data)
 
     :else data))
 
@@ -219,7 +230,7 @@
   (-> file-info
       val
       (add-path-to-file-entry (key file-info))
-      sort-nested-environment-class-info-maps))
+      sort-nested-info-maps))
 
 (schema/defn ^:always-validate
   class-info-from-jruby->class-info-for-json :- EnvironmentClassesInfo
@@ -327,6 +338,43 @@
         (rr/not-found (i18n/tru "Could not find environment ''{0}''" environment))))))
 
 (schema/defn ^:always-validate
+  module-info-from-jruby->module-info-for-json  :- EnvironmentModulesInfo
+  "Creates a new map with a top level key `name` that corresponds to the
+  requested environment and a top level key `modules` which contains the module
+  information obtained from JRuby."
+  [info-from-jruby :- List
+   environment :- schema/Str]
+  (->> info-from-jruby
+       sort-nested-info-maps
+       vec
+       (sorted-map :name environment :modules)))
+
+(schema/defn ^:always-validate
+  environment-module-response! :- ringutils/RingResponse
+  "Process the environment module information, returning a ring response to be
+  propagated back up to the caller of the environment_modules endpoint."
+  [info-from-jruby :- List
+   environment :- schema/Str]
+  (let [info-as-json (module-info-from-jruby->module-info-for-json
+                       info-from-jruby environment)]
+    (middleware-utils/json-response 200 info-as-json)))
+
+(schema/defn ^:always-validate
+  environment-module-info-fn :- IFn
+  "Middleware function for constructing a Ring response from an incoming
+  request for environment_modules information."
+  [jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)]
+  (fn [request]
+    (let [environment (jruby-request/get-environment-from-request request)]
+      (if-let [module-info
+               (jruby-protocol/get-environment-module-info jruby-service
+                                                           (:jruby-instance request)
+                                                           environment)]
+        (environment-module-response! module-info
+                                      environment)
+        (rr/not-found (i18n/tru "Could not find environment ''{0}''" environment))))))
+
+(schema/defn ^:always-validate
   wrap-with-etag-check :- IFn
   "Middleware function which validates whether or not the If-None-Match
   header on an incoming environment_classes request matches the last Etag
@@ -357,6 +405,17 @@
                               environment-class-cache-enabled)
    (jruby-request/wrap-with-jruby-instance jruby-service)
    (wrap-with-etag-check jruby-service)
+   jruby-request/wrap-with-environment-validation
+   jruby-request/wrap-with-error-handling))
+
+
+(schema/defn ^:always-validate
+  environment-module-handler :- IFn
+  "Handler for processing an incoming environment_modules Ring request"
+  [jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)]
+  (->
+   (environment-module-info-fn jruby-service)
+   (jruby-request/wrap-with-jruby-instance jruby-service)
    jruby-request/wrap-with-environment-validation
    jruby-request/wrap-with-error-handling))
 
@@ -448,13 +507,17 @@
   (let [environment-class-handler
         (environment-class-handler jruby-service
                                    environment-class-cache-enabled)
+        environment-module-handler
+        (environment-module-handler jruby-service)
         static-file-content-handler
         (static-file-content-request-handler get-code-content-fn)]
     (comidi/routes
-     (comidi/GET ["/environment_classes" [#".*" :rest]] request
-                 (environment-class-handler request))
-     (comidi/GET ["/static_file_content/" [#".*" :rest]] request
-                 (static-file-content-handler request)))))
+      (comidi/GET ["/environment_classes" [#".*" :rest]] request
+                  (environment-class-handler request))
+      (comidi/GET ["/environment_modules" [#".*" :rest]] request
+                  (environment-module-handler request))
+      (comidi/GET ["/static_file_content/" [#".*" :rest]] request
+                  (static-file-content-handler request)))))
 
 (schema/defn ^:always-validate
   v3-routes :- bidi-schema/RoutePair
--- a/src/ruby/puppetserver-lib/puppet/server/master.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/master.rb
@@ -70,6 +70,13 @@ class Puppet::Server::Master
     end
   end
 
+  def getModuleInfoForEnvironment(env)
+    environment = @env_loader.get(env)
+    unless environment.nil?
+      self.class.getModules(environment)
+    end
+  end
+
   def getSetting(setting)
     Puppet[setting]
   end
@@ -87,6 +94,20 @@ class Puppet::Server::Master
   end
 
   private
+
+  def self.getModules(env)
+    env.modules.collect do |mod|
+      module_data = {}
+
+      # NOTE: If in the future we want to collect more
+      #       Module settings, this should be more programatic
+      #       rather than getting these settings one by one
+      module_data[:name] = mod.forge_name
+      module_data[:version] = mod.version
+
+      module_data
+    end
+  end
 
   def self.getManifests(env)
     manifests =
--- a/test/integration/puppetlabs/services/jruby/module_info_test.clj
+++ b/test/integration/puppetlabs/services/jruby/module_info_test.clj
@@ -1,0 +1,85 @@
+(ns puppetlabs.services.jruby.module-info-test
+  (:require [clojure.test :refer :all]
+            [puppetlabs.kitchensink.core :as ks]
+            [puppetlabs.services.jruby.jruby-puppet-testutils :as jruby-testutils]
+            [puppetlabs.services.jruby.jruby-puppet-service :refer [jruby-puppet-pooled-service]]
+            [puppetlabs.services.puppet-profiler.puppet-profiler-service :refer [puppet-profiler-service]]
+            [puppetlabs.services.jruby-pool-manager.jruby-pool-manager-service :refer [jruby-pool-manager-service]]
+            [me.raynes.fs :as fs]
+            [cheshire.core :as cheshire]
+            [puppetlabs.puppetserver.testutils :as testutils]
+            [puppetlabs.trapperkeeper.app :as tk-app]
+            [puppetlabs.trapperkeeper.testutils.bootstrap :as tk-bootstrap]))
+
+(defn gen-modules
+  [[mod-dir manifests]]
+  (let [modules-dir (fs/file mod-dir "modules")]
+    (ks/mkdirs! modules-dir)
+    (doseq [manifest manifests]
+      (let [module-dir (fs/file modules-dir manifest)
+            metadata-json {"name" manifest
+                           "version" "1.0.0"
+                           "author" "Puppet"
+                           "license" "apache"
+                           "dependencies" []
+                           "source" "https://github.com/puppetlabs"}]
+        (ks/mkdirs! (fs/file module-dir))
+        (spit (fs/file module-dir "metadata.json")
+              (cheshire/generate-string metadata-json))))))
+
+(defn create-env
+  [[env-dir manifests]]
+  (testutils/create-env-conf env-dir "")
+  (gen-modules [env-dir manifests]))
+
+(defn roundtrip-via-json
+  [obj]
+  (-> obj
+      (cheshire/generate-string)
+      (cheshire/parse-string)))
+
+(deftest ^:integration module-info-test
+  (testing "module info properly enumerated for"
+    (let [code-dir (ks/temp-dir)
+          conf-dir (ks/temp-dir)
+          config (jruby-testutils/jruby-puppet-tk-config
+                  (jruby-testutils/jruby-puppet-config
+                   {:master-code-dir (.getAbsolutePath code-dir)
+                    :master-conf-dir (.getAbsolutePath conf-dir)}))]
+
+      (tk-bootstrap/with-app-with-config
+       app
+       [jruby-puppet-pooled-service
+        puppet-profiler-service
+        jruby-pool-manager-service]
+       config
+       (let [jruby-service (tk-app/get-service app :JRubyPuppetService)
+             instance (jruby-testutils/borrow-instance jruby-service :test)
+             jruby-puppet (:jruby-puppet instance)
+             env-registry (:environment-registry instance)
+
+             _ (testutils/create-file (fs/file conf-dir "puppet.conf")
+                                      "[main]\nenvironment_timeout=unlimited\nbasemodulepath=$codedir/modules\n")
+
+             env-dir (fn [env-name]
+                       (fs/file code-dir "environments" env-name))
+             env-1-dir (env-dir "env1")
+             env-1-dir-and-manifests [env-1-dir ["foo" "bar"]]
+             _ (create-env env-1-dir-and-manifests)
+
+             get-module-info-for-env (fn [env]
+                                       (-> (.getModuleInfoForEnvironment jruby-puppet
+                                                                         env)
+                                           (roundtrip-via-json)))]
+         (try
+           (testing "retrieves basic module information about installed modules"
+             (let [env-1-module-info #{{"name" "foo", "version" "1.0.0"}
+                                       {"name" "bar", "version" "1.0.0"}}]
+               (is (= env-1-module-info
+                      (set (get-module-info-for-env "env1")))
+                   "Unexpected info retrieved for 'env1'")))
+
+           (testing "returns nothing if a fake env is given"
+             (is (= nil (get-module-info-for-env "bogus_env"))))
+           (finally
+             (jruby-testutils/return-instance jruby-service instance :test))))))))
--- a/test/integration/puppetlabs/services/master/environment_modules_int_test.clj
+++ b/test/integration/puppetlabs/services/master/environment_modules_int_test.clj
@@ -1,0 +1,83 @@
+(ns puppetlabs.services.master.environment-modules-int-test
+  (:require [clojure.test :refer :all]
+            [puppetlabs.http.client.sync :as http-client]
+            [puppetlabs.kitchensink.core :as ks]
+            [puppetlabs.puppetserver.bootstrap-testutils :as bootstrap]
+            [puppetlabs.trapperkeeper.testutils.logging :as logutils]
+            [puppetlabs.puppetserver.testutils :as testutils]
+            [cheshire.core :as cheshire]
+            [me.raynes.fs :as fs]))
+
+(def test-resources-dir
+  "./dev-resources/puppetlabs/services/master/environment_modules_int_test")
+
+(defn purge-env-dir
+  []
+  (-> testutils/conf-dir
+      (fs/file "environments")
+      fs/delete-dir))
+
+(use-fixtures :once
+              (testutils/with-puppet-conf
+               (fs/file test-resources-dir "puppet.conf")))
+
+(use-fixtures :each
+              (fn [f]
+                (purge-env-dir)
+                (try
+                  (f)
+                  (finally
+                    (purge-env-dir)))))
+
+(defn get-env-modules
+  ([env-name]
+   (try
+       (http-client/get
+        (str "https://localhost:8140/puppet/v3/"
+             "environment_modules?"
+             "environment="
+             env-name)
+        (merge
+         testutils/ssl-request-options
+         {:as :text}))
+       (catch Exception e
+         (throw (Exception. "environment_modules http get failed" e))))))
+
+(defn response->module-info-map
+  [response]
+  (-> response :body cheshire/parse-string))
+
+(deftest ^:integration environment-modules-integration-cache-disabled-test
+  (testing "when environment modules cache is disabled for a module request"
+    (bootstrap/with-puppetserver-running-with-config
+     app
+     (-> {:jruby-puppet {:max-active-instances 1}}
+         (bootstrap/load-dev-config-with-overrides)
+         (ks/dissoc-in [:jruby-puppet
+                        :environment-class-cache-enabled]))
+     (logutils/with-test-logging
+       (let [foo-file (testutils/write-foo-pp-file
+                        "class foo (String $foo_1 = \"is foo\"){}")
+             expected-response {"modules" [{"name" "foo", "version" "1.0.0"}]
+                                "name" "production"}
+             response (get-env-modules "production")
+             emptyenv-response (get-env-modules "")
+             notreal-response (get-env-modules "notreal")]
+         (testing "a successful status code is returned"
+           (is (= 200 (:status response))
+               (str
+                 "unexpected status code for response, response: "
+                 (ks/pprint-to-string response))))
+         (testing "a failed status code is returned"
+           (is (= 404 (:status notreal-response))
+               (str
+                 "unexpected status code for response, response: "
+                 (ks/pprint-to-string response))))
+         (testing "a failed status code is returned"
+           (is (= 400 (:status emptyenv-response))
+               (str
+                 "unexpected status code for response, response: "
+                 (ks/pprint-to-string response))))
+         (testing "the expected response body is returned"
+           (is (= expected-response
+                  (response->module-info-map response)))))))))
--- a/dev-resources/puppetlabs/services/master/environment_modules_int_test/puppet.conf
+++ b/dev-resources/puppetlabs/services/master/environment_modules_int_test/puppet.conf
@@ -1,0 +1,5 @@
+[main]
+
+certname = localhost
+environmentpath = $confdir/environments
+environment_timeout = unlimited
--- a/documentation/index.markdown
+++ b/documentation/index.markdown
@@ -41,6 +41,7 @@ Puppet Server is the next-generation application for managing Puppet agents.
     * [JRuby pool](./admin-api/v1/jruby-pool.markdown)
 * **Server-specific Puppet API endpoints**
     * [Environment classes](./puppet-api/v3/environment_classes.markdown)
+    * [Environment modules](./puppet-api/v3/environment_modules.markdown)
     * [Static file content](./puppet-api/v3/static_file_content.markdown)
 * **Status API endpoints**
     * [Status services](./status-api/v1/services.markdown)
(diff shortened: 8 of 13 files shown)


documentation/puppet-api/v3/environment_modules.json
```diff
@@ -0,0 +1,19 @@
...
+  "title":       "Environment Modules",
+  "description": "Information about the modules in a Puppet code environment",
+  "type":        "object",
+  "properties": {
+    "modules": {
+      "description": "The array of modules which exist in an environment.",
+      "type": "array",
+      "properties": {
+
+      }
```
briancain: I stopped making this because it feels like this was probably auto-generated when it was made for environment_classes.json? If not I'll continue, but I didn't want to make it all by hand if it could be auto-generated.

documentation/puppet-api/v3/environment_modules.markdown
```diff
@@ -0,0 +1,150 @@
...
+[deprecated WEBrick Puppet master]: https://docs.puppet.com/puppet/latest/reference/services_master_webrick.html
+[node definitions]: https://docs.puppet.com/puppet/latest/reference/lang_node_definitions.html
+[defined types]: https://docs.puppet.com/puppet/latest/reference/lang_defined_types.html
+[`environment_timeout`]: https://docs.puppet.com/puppet/latest/reference/config_file_environment.html#environmenttimeout
+[resource type API]: https://docs.puppet.com/puppet/latest/reference/http_api/http_resource_type.html
+[`manifest` setting]: https://docs.puppet.com/puppet/latest/reference/config_file_environment.html#manifest
+
+[`auth.conf`]: ../../config_file_auth.markdown
+[`puppetserver.conf`]: ../../config_file_puppetserver.markdown
+[environment cache API]: ../../admin-api/v1/environment-cache.markdown
```
briancain: A lot of these links should be deleted...

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -124,6 +124,11 @@
   {:name schema/Str
    :files [EnvironmentClassesFileEntry]})
 
+(def EnvironmentModulesInfo
+  "Schema for the return payload an environment_classes request."
+  {:name schema/Str
+   :modules schema/Any}) ; TODO: fixme
```
briancain: As far as I can there, there's no `schema/Array` ?

documentation/puppet-api/v3/environment_modules.markdown
```diff
@@ -0,0 +1,137 @@
...
+```
+GET /puppet/v3/environment_modules?environment=env
+
+HTTP/1.1 200 OK
+Content-Type: application/json
+
+{
+    "modules": [
+        {
+            "forge_name": "puppetlabs/ntp",
```
briancain: This reminds me, I need to double check what happens when a module is installed that isn't form the forge. Not sure if the key is just missing from the `Puppet::Modules` object or if it's null.

briancain:
I think I still have some work here to get the tests to pass....

briancain:
Also don't trust travis ci I have no idea how it thinks those checks passed :) I'm getting failures in the `puppetlabs.services.jruby.class-info-test` namespace so I probably broke something.

briancain:
OH interesting, it looks like the unit tests are actually grabbing the modules I have installed on disk? I see ntp and stdlib in the failure diff here (both of which are modules I have installed)

```clojure
AIL in (class-info-test) (class_info_test.clj:136)
class info properly enumerated for initial parse
Unexpected info retrieved for 'env1'
expected: (= (expected-envs-info "env1") (get-class-info-for-env "env1"))
  actual: (not (= {"/var/folders/50/pp8gf2qj7rvcwgzq8ytnxf6w0000gn/T/null1483658212806-3918597310/environments/env1/manifests/foo.pp" {"classes" [{"name" "foo", "params" [{"name" "foo_a"} {"name" "foo_b", "type" "Integer"} {"name" "foo_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]} {"name" "foo2", "params" [{"name" "foo2_a"} {"name" "foo2_b", "type" "Integer"} {"name" "foo2_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]}]}, "/var/folders/50/pp8gf2qj7rvcwgzq8ytnxf6w0000gn/T/null1483658212806-3918597310/environments/env1/manifests/bar.pp" {"classes" [{"name" "bar", "params" [{"name" "bar_a"} {"name" "bar_b", "type" "Integer"} {"name" "bar_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]} {"name" "bar2", "params" [{"name" "bar2_a"} {"name" "bar2_b", "type" "Integer"} {"name" "bar2_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]}]}, "/var/folders/50/pp8gf2qj7rvcwgzq8ytnxf6w0000gn/T/null1483658212806-3918597310/environments/env1/modules/envmod1/manifests/envmod1baz.pp" {"classes" [{"name" "envmod1baz", "params" [{"name" "envmod1baz_a"} {"name" "envmod1baz_b", "type" "Integer"} {"name" "envmod1baz_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]} {"name" "envmod1baz2", "params" [{"name" "envmod1baz2_a"} {"name" "envmod1baz2_b", "type" "Integer"} {"name" "envmod1baz2_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]}]}, "/var/folders/50/pp8gf2qj7rvcwgzq8ytnxf6w0000gn/T/null1483658212806-3918597310/environments/env1/modules/envmod1/manifests/envmod1bim.pp" {"classes" [{"name" "envmod1bim", "params" [{"name" "envmod1bim_a"} {"name" "envmod1bim_b", "type" "Integer"} {"name" "envmod1bim_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]} {"name" "envmod1bim2", "params" [{"name" "envmod1bim2_a"} {"name" "envmod1bim2_b", "type" "Integer"} {"name" "envmod1bim2_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]}]}, "/var/folders/50/pp8gf2qj7rvcwgzq8ytnxf6w0000gn/T/null1483658212806-3918597310/environments/env1/modules/envmod2/manifests/envmod2baz.pp" {"classes" [{"name" "envmod2baz", "params" [{"name" "envmod2baz_a"} {"name" "envmod2baz_b", "type" "Integer"} {"name" "envmod2baz_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]} {"name" "envmod2baz2", "params" [{"name" "envmod2baz2_a"} {"name" "envmod2baz2_b", "type" "Integer"} {"name" "envmod2baz2_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]}]}, "/var/folders/50/pp8gf2qj7rvcwgzq8ytnxf6w0000gn/T/null1483658212806-3918597310/environments/env1/modules/envmod2/manifests/envmod2bim.pp" {"classes" [{"name" "envmod2bim", "params" [{"name" "envmod2bim_a"} {"name" "envmod2bim_b", "type" "Integer"} {"name" "envmod2bim_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]} {"name" "envmod2bim2", "params" [{"name" "envmod2bim2_a"} {"name" "envmod2bim2_b", "type" "Integer"} {"name" "envmod2bim2_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]}]}, "/var/folders/50/pp8gf2qj7rvcwgzq8ytnxf6w0000gn/T/null1483658212806-3918597310/modules/basemod1/manifests/basemod1bap.pp" {"classes" [{"name" "basemod1bap", "params" [{"name" "basemod1bap_a"} {"name" "basemod1bap_b", "type" "Integer"} {"name" "basemod1bap_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]} {"name" "basemod1bap2", "params" [{"name" "basemod1bap2_a"} {"name" "basemod1bap2_b", "type" "Integer"} {"name" "basemod1bap2_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]}]}} {"/opt/puppetlabs/puppet/modules/ntp/manifests/service.pp" {"classes" [{"name" "ntp::service", "params" []}]}, "/opt/puppetlabs/puppet/modules/ntp/manifests/config.pp" {"classes" [{"name" "ntp::config", "params" []}]}, "/var/folders/50/pp8gf2qj7rvcwgzq8ytnxf6w0000gn/T/null1483658212806-3918597310/environments/env1/manifests/bar.pp" {"classes" [{"name" "bar", "params" [{"name" "bar_a"} {"name" "bar_b", "type" "Integer"} {"name" "bar_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]} {"name" "bar2", "params" [{"name" "bar2_a"} {"name" "bar2_b", "type" "Integer"} {"name" "bar2_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]}]}, "/opt/puppetlabs/puppet/modules/ntp/manifests/init.pp" {"classes" [{"name" "ntp", "params" [{"name" "broadcastclient", "type" "Boolean"} {"name" "config", "type" "TypeReference['Stdlib::Absolutepath']"} {"name" "config_dir", "type" "Optional[TypeReference['Stdlib::Absolutepath']]"} {"name" "config_file_mode", "type" "String"} {"name" "config_epp", "type" "Optional[String]"} {"name" "config_template", "type" "Optional[String]"} {"name" "disable_auth", "type" "Boolean"} {"name" "disable_dhclient", "type" "Boolean"} {"name" "disable_kernel", "type" "Boolean"} {"name" "disable_monitor", "type" "Boolean"} {"name" "fudge", "type" "Optional[Array[String]]"} {"name" "driftfile", "type" "TypeReference['Stdlib::Absolutepath']"} {"name" "leapfile", "type" "Optional[TypeReference['Stdlib::Absolutepath']]"} {"name" "logfile", "type" "Optional[TypeReference['Stdlib::Absolutepath']]"} {"name" "iburst_enable", "type" "Boolean"} {"name" "keys", "type" "Array[String]"} {"name" "keys_enable", "type" "Boolean"} {"name" "keys_file", "type" "TypeReference['Stdlib::Absolutepath']"} {"name" "keys_controlkey", "type" "Optional[TypeReference['Ntp::Key_id']]"} {"name" "keys_requestkey", "type" "Optional[TypeReference['Ntp::Key_id']]"} {"name" "keys_trusted", "type" "Optional[Array[TypeReference['Ntp::Key_id']]]"} {"name" "minpoll", "type" "Optional[TypeReference['Ntp::Poll_interval']]"} {"name" "maxpoll", "type" "Optional[TypeReference['Ntp::Poll_interval']]"} {"name" "package_ensure", "type" "String"} {"name" "package_manage", "type" "Boolean"} {"name" "package_name", "type" "Array[String]"} {"name" "panic", "type" "Optional[Integer[0, default]]"} {"name" "peers", "type" "Array[String]"} {"name" "preferred_servers", "type" "Array[String]"} {"name" "restrict", "type" "Array[String]"} {"name" "interfaces", "type" "Array[String]"} {"name" "interfaces_ignore", "type" "Array[String]"} {"name" "servers", "type" "Array[String]"} {"name" "service_enable", "type" "Boolean"} {"name" "service_ensure", "type" "String"} {"name" "service_manage", "type" "Boolean"} {"name" "service_name", "type" "String"} {"name" "service_provider", "type" "Optional[String]"} {"name" "stepout", "type" "Optional[Integer[0, 65535]]"} {"name" "step_tickers_file", "type" "Optional[TypeReference['Stdlib::Absolutepath']]"} {"name" "step_tickers_epp", "type" "Optional[String]"} {"name" "step_tickers_template", "type" "Optional[String]"} {"name" "tinker", "type" "Optional[Boolean]"} {"name" "tos", "type" "Boolean"} {"name" "tos_minclock", "type" "Optional[Integer[1, default]]"} {"name" "tos_minsane", "type" "Optional[Integer[1, default]]"} {"name" "tos_floor", "type" "Optional[Integer[1, default]]"} {"name" "tos_ceiling", "type" "Optional[Integer[1, default]]"} {"name" "tos_cohort", "type" "Variant[Boolean, Integer[0, 1]]"} {"name" "udlc", "type" "Boolean"} {"name" "udlc_stratum", "type" "Optional[Integer[1, 15]]"} {"name" "ntpsigndsocket", "type" "Optional[TypeReference['Stdlib::Absolutepath']]"} {"name" "authprov", "type" "Optional[String]"}]}]}, "/opt/puppetlabs/puppet/modules/ntp/manifests/install.pp" {"classes" [{"name" "ntp::install", "params" []}]}, "/var/folders/50/pp8gf2qj7rvcwgzq8ytnxf6w0000gn/T/null1483658212806-3918597310/modules/basemod1/manifests/basemod1bap.pp" {"classes" [{"name" "basemod1bap", "params" [{"name" "basemod1bap_a"} {"name" "basemod1bap_b", "type" "Integer"} {"name" "basemod1bap_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]} {"name" "basemod1bap2", "params" [{"name" "basemod1bap2_a"} {"name" "basemod1bap2_b", "type" "Integer"} {"name" "basemod1bap2_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]}]}, "/var/folders/50/pp8gf2qj7rvcwgzq8ytnxf6w0000gn/T/null1483658212806-3918597310/environments/env1/manifests/foo.pp" {"classes" [{"name" "foo", "params" [{"name" "foo_a"} {"name" "foo_b", "type" "Integer"} {"name" "foo_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]} {"name" "foo2", "params" [{"name" "foo2_a"} {"name" "foo2_b", "type" "Integer"} {"name" "foo2_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]}]}, "/opt/puppetlabs/puppet/modules/stdlib/manifests/init.pp" {"classes" [{"name" "stdlib", "params" []}]}, "/opt/puppetlabs/puppet/modules/stdlib/manifests/stages.pp" {"classes" [{"name" "stdlib::stages", "params" []}]}, "/var/folders/50/pp8gf2qj7rvcwgzq8ytnxf6w0000gn/T/null1483658212806-3918597310/environments/env1/modules/envmod1/manifests/envmod1bim.pp" {"classes" [{"name" "envmod1bim", "params" [{"name" "envmod1bim_a"} {"name" "envmod1bim_b", "type" "Integer"} {"name" "envmod1bim_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]} {"name" "envmod1bim2", "params" [{"name" "envmod1bim2_a"} {"name" "envmod1bim2_b", "type" "Integer"} {"name" "envmod1bim2_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]}]}, "/var/folders/50/pp8gf2qj7rvcwgzq8ytnxf6w0000gn/T/null1483658212806-3918597310/environments/env1/modules/envmod2/manifests/envmod2bim.pp" {"classes" [{"name" "envmod2bim", "params" [{"name" "envmod2bim_a"} {"name" "envmod2bim_b", "type" "Integer"} {"name" "envmod2bim_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]} {"name" "envmod2bim2", "params" [{"name" "envmod2bim2_a"} {"name" "envmod2bim2_b", "type" "Integer"} {"name" "envmod2bim2_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]}]}, "/var/folders/50/pp8gf2qj7rvcwgzq8ytnxf6w0000gn/T/null1483658212806-3918597310/environments/env1/modules/envmod1/manifests/envmod1baz.pp" {"classes" [{"name" "envmod1baz", "params" [{"name" "envmod1baz_a"} {"name" "envmod1baz_b", "type" "Integer"} {"name" "envmod1baz_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]} {"name" "envmod1baz2", "params" [{"name" "envmod1baz2_a"} {"name" "envmod1baz2_b", "type" "Integer"} {"name" "envmod1baz2_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]}]}, "/var/folders/50/pp8gf2qj7rvcwgzq8ytnxf6w0000gn/T/null1483658212806-3918597310/environments/env1/modules/envmod2/manifests/envmod2baz.pp" {"classes" [{"name" "envmod2baz", "params" [{"name" "envmod2baz_a"} {"name" "envmod2baz_b", "type" "Integer"} {"name" "envmod2baz_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]} {"name" "envmod2baz2", "params" [{"name" "envmod2baz2_a"} {"name" "envmod2baz2_b", "type" "Integer"} {"name" "envmod2baz2_c", "type" "String", "default_literal" "c default value", "default_source" "'c default value'"}]}]}}))
```

src/clj/puppetlabs/services/master/master_core.clj
briancain: Pretty much. The array looks like this:

```json
[
        {
         "forge_name": "puppetlabs/ntp",
            "name": "ntp",
            "version": "6.0.0"
        },
        {
            "forge_name": "puppetlabs/stdlib",
            "name": "stdlib",
            "version": "4.14.0"
        }
]
```

briancain: Maybe `schema/Map` would work?

haus: Okay, then how about something like...

```
(def ModuleInfo
{ "forge_name" (schema/Str)
  "name" (schema/Str)
  "version" (schema/Str)
}
```

`:modules [ModuleInfo]`

briancain: Oh that's perfect!

documentation/puppet-api/v3/environment_modules.json
camlow325: Nah, unfortunately, I did hand create the whole thing for the environment_classes endpoint.  In retrospect, though, I think there was value in writing up what I thought the payload should look like independent from what was actually generated in a json payload - the first time anyway.  I used `JSON::Validator` in the Ruby json-schema gem (https://github.com/ruby-json-schema/json-schema) to validate the schema against the response data from some sample queries that I made to the environment_classes endpoint.  For reference, the JSON schema for that came in originally on #940.

src/clj/puppetlabs/services/master/master_core.clj
camlow325: `forge_name` can be null, though, right?  Should the value for it be `(schema/maybe schema/Str)`?

documentation/puppet-api/v3/environment_modules.markdown
haus: if the key is missing then you'll want some `schema/optional-key` calls.

src/clj/puppetlabs/services/master/master_core.clj
haus: if the value can be null, then yea, `schema/maybe` sounds good. if the key can be missing, then `(schema/optional-key "forge_name")` probably?

briancain: @camlow325 yeah, I'll have to make it that. I also still need to confirm what shows up in the Puppet::Modules data structure if it is a module not installed from forge.

briancain: Hmm, this is a bit tricker than I thought. At this point, they key `:modules` is still a RubyHash. I'm relying on Ring to convert that map for me when I send it back to the user. Maybe that's a bad idea?

```
Output of module-info-from-jruby->module-info-for-json does not match schema: {:modules [(not (map? a-org.jruby.RubyHash)) (not (map? a-org.jruby.RubyHash))]}
```

camlow325: I think the problem there is that while `RubyHash` implements `java.util.Map`, schema only seems to work well with Clojure map classes, like `clojure.lang.PersistentArrayMap`.

For the environment_classes endpoint, I wrote a function in the `master_core` namespace called `sort-nested-environment-class-info-maps`.  I used that to transform the Java class types coming from JRuby into Clojure-friendly types that work with Clojure's `sorted-map`, which I used to preserve element order for serialization purposes.  This has the benefit of schema maps being usable to validate the output as well.  Serialization order of map elements is more crucial for the environment_classes endpoint since we calculate etags for use with its caching layer and need those to be consistent for the whole payload.  The environment_modules endpoint doesn't need to deal with etags currently, so the extra ordering benefit for serialization isn't critical.  Assuming a little extra memory usage and a little poorer performance isn't a big deal - which I believe to be true at this point - I think using this function could help for schema validation here.

For example, I got this to work:

```clj
(def EnvironmentModuleInfo
  {:forge_name (schema/maybe schema/Str)
   :name schema/Str
   :version schema/Str})

(def EnvironmentModulesInfo
   "Schema for the return payload an environment_classes request."
   {:name schema/Str
    :modules [EnvironmentModuleInfo]})

  module-info-from-jruby->module-info-for-json  :- EnvironmentModulesInfo
   [info-from-jruby :- List
    environment :- schema/Str]
   (->> info-from-jruby
        sort-nested-environment-class-info-maps
        vec
        (sorted-map :name environment :modules)))
```

Assuming you want to go this way, I think it would be good to rename `sort-nested-environment-class-info-maps` to be something more generic like `sort-nested-java-maps` or something.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -327,6 +332,40 @@
         (rr/not-found (i18n/tru "Could not find environment ''{0}''" environment))))))
 
 (schema/defn ^:always-validate
+  module-info-from-jruby->module-info-for-json  :- EnvironmentModulesInfo
+  "Creates a new map with a top level name that corresponds to the requested environment
+  and a top level key modules which contains the module information obtained from JRuby."
+  [info-from-jruby :- schema/Any
```
camlow325: I think `List` would be a little better than `schema/Any` and should work here.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -327,6 +332,40 @@
...
+   environment :- schema/Str]
+  (->> info-from-jruby
+       (sorted-map :name environment :modules)))
+
+(schema/defn ^:always-validate
+  environment-module-response! :- ringutils/RingResponse
+  "Process the environment module information, returning a ring response to be
+  propagated back up to the caller of the environment_modules endpoint."
+  [info-from-jruby :- schema/Any
```
camlow325: Same comment here with `List` instead of `schema/Any`.

src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -88,6 +95,20 @@ def terminate
 
   private
 
+  def self.getModules(env)
```
camlow325: Might be able to cut this down to just:

```ruby
def self.getModules(env)
    env.modules.collect do |mod|
      module_data = {}
      # There might be other module metadata that we care about,
      # but for now just return these.
      [:name, :version, :forge_name].each do |attribute|
        module_data[attribute.to_s] = mod.send(attribute)
      end
      module_data
    end
end
```

documentation/puppet-api/v3/environment_modules.markdown
```diff
@@ -0,0 +1,137 @@
...
+The environment must be purely alphanumeric, not ''
+```
+
+#### Environment includes non-alphanumeric characters
+
+If the environment parameter in your request includes any characters that are
+not `A-Z`, `a-z`, `0-9`, or `_` (underscore), the server returns an HTTP 400 (Bad Request) error:
+
+```
+GET /puppet/v3/environment_classes?environment=bog|us
```
camlow325: s/classes/modules

documentation/puppet-api/v3/environment_modules.markdown
```diff
@@ -0,0 +1,137 @@
...
+
+HTTP/1.1 400 Bad Request
+
+The environment must be purely alphanumeric, not 'bog|us'
+```
+
+### Schema
+
+An environment classes response body conforms to the
```
camlow325: s/classes/modules

documentation/puppet-api/v3/environment_modules.markdown
```diff
@@ -0,0 +1,137 @@
...
+[environment modules schema](./environment_modules.json).
+
+### Authorization
+
+Unlike other Puppet master service-based API endpoints, the environment classes API is
```
camlow325: s/classes/modules

documentation/puppet-api/v3/environment_modules.markdown
```diff
@@ -0,0 +1,137 @@
...
+provided exclusively by Puppet Server. All requests made to the environment
+classes API are authorized using the Trapperkeeper-based [`auth.conf`][] feature
```
camlow325: s/classes/modules

documentation/puppet-api/v3/environment_modules.markdown
```diff
@@ -0,0 +1,137 @@
...
+introduced in Puppet Server 2.2.0, and ignores the older Ruby-based authorization process
+and configuration. The value of the `use-legacy-auth-conf` setting in the `jruby-puppet`
+configuration section of [`puppetserver.conf`][] is ignored for requests
+to the environment classes API, because the Ruby-based authorization process is not equipped to
```
camlow325: s/classes/modules

camlow325:
There are a couple of docs links that we should probably add to the new `environment_modules.markdown` doc - from `index.markdown` and `services_master_puppetserver.markdown` like we have with the `environment_classes.markdown` doc.

camlow325:
Were you able to figure out why the `class-info-tests` were failing at one point?  The last Travis run seems to have passed.

briancain:
@camlow325 - from what I can tell, it looks like the unit tests are pulling in module/class information from my local modulepath and that's why the diffs are failing. I only have ntp and stdlib installed and they showed up in the diff. I'm guessing that's why it passes on Jenkins?

briancain:
Also I'm 👍 on all this feedback. Planning on spending the afternoon addressing all of it and starting to look at getting some tests too.

src/clj/puppetlabs/services/master/master_core.clj
briancain: Cool, I think this approach makes sense to me, especially since it's already being used for `environment_classes`. I'll go ahead and rename the function too.

briancain:
@camlow325 FYI: I figured out where `forge_name` comes from. In [puppet](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/module.rb#L156), the modules class will generate that name itself based on the `name` key in the `metadata.json` file so custom modules not from the forge will still have a `forge_name` in the Modules class. I wonder then if it makes sense to only capture `forge_name` and `version`?

camlow325:
@briancain Given what you found about the `forge_name`, maybe it would make sense to capture the `forge_name` from Ruby Puppet but just report that as `name` from the HTTP endpoint (and then not have a `forge_name` on the HTTP endpoint)?  Seems like it would make more sense to consumers for the `name` from the 'metadata.json' file to match the `name` returned from the endpoint?

briancain:
@camlow325 that makes sense to me too. I'll plan on that. Then I think the only main thing left to do is to create some test for this assuming there's no more feedback on the code itself!

test/integration/puppetlabs/services/jruby/module_info_test.clj
```diff
@@ -0,0 +1,86 @@
...
+                                       (-> (.getModuleInfoForEnvironment jruby-puppet
+                                                                         env)
+                                           (roundtrip-via-json)))]
+         (try
+           (testing "retrieves basic module information about installed modules"
+             (let [env-1-module-info '({"name" "bar", "version" "1.0.0"}
+                                       {"name" "foo", "version" "1.0.0"})]
+               (is (= env-1-module-info
+                      (get-module-info-for-env "env1"))
+                   "Unexpected info retrieved for 'env1'")))
```
briancain: @camlow325 - Hmm, this seems to come out on a different order on travis. Is there a good way to test that the map I defined up in the let will be equal to what gets returned in `get-module-info-for-env` regardless of the order without having to manipulate what gets returned by the function?

haus: Generally in other tests like this we convert into sets on both sides.

documentation/puppet-api/v3/environment_modules.json
```diff
@@ -0,0 +1,26 @@
...
+          "type": "string"
+        }
+      }
+    },
+    "name": {
+      "description": "Name of the environment",
+      "type": "string"
+    }
+  }
+}
```
camlow325: I think this schema isn't strict enough to catch malformed json.  I used the ruby `json-schema` gem to validate JSON responses that I captured from the endpoint.  Here's the command line I ran for that:

```
ruby -rjson-schema -e "puts JSON::Validator.validate!('./documentation/puppet-api/v3/environment_modules_fixed.json','response.json')"
```
'response.json' is the file I captured from a curl request to the endpoint.  Even if I intentionally messed up some of the key names, e.g., replacing "name" with "bogus-name", the `validate!` call would still return `true`.

I added a couple of `required` elements with specific names expected in the output and added an `items` wrapper underneath the `array` element, which I think is required for JSON schema validation of an array.  The result is below.  With this, I saw validation failures for any intentionally malformed json vs. success for the "good" json that the endpoint returned.

Does this look okay to you?

```json
{
  "$schema":     "http://json-schema.org/draft-04/schema#",
  "title":       "Environment Modules",
  "description": "Information about the modules in a Puppet code environment",
  "type":        "object",
  "properties": {
    "modules": {
      "description": "The array of modules which exist in an environment.",
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "name": {
            "description": "The name of the puppet module",
            "type": "string"
          },
          "version": {
            "description": "The version of the puppet module",
            "type": "string"
          }
        },
        "required": ["name", "version"],
        "additionalProperties": false
      }
    },
    "name": {
      "description": "Name of the environment",
      "type": "string"
    }
  },
  "required": ["modules", "name"],
  "additionalProperties": false
}
```

briancain: @camlow325 that seems pretty solid to me. I can update the PR with this new schema.

src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -88,6 +95,20 @@ def terminate
...
+    env.modules.collect do |mod|
+      module_data = {}
+
+      # NOTE: If in the future we want to collect more
+      #       Module settings, this should be more programatic
+      #       rather than getting these settings one by one
+      module_data[:name] = mod.send(:forge_name)
```
camlow325: I think these could be shortened to just this:

```ruby
module_data[:name] = mod.forge_name
module_data[:version] = mod.version
```

test/integration/puppetlabs/services/jruby/module_info_test.clj
```diff
@@ -0,0 +1,86 @@
+(ns puppetlabs.services.jruby.module-info-test
+  (:require [clojure.test :refer :all]
+            [puppetlabs.kitchensink.core :as ks]
+            [puppetlabs.services.jruby.jruby-puppet-testutils :as jruby-testutils]
+            [puppetlabs.services.jruby.jruby-puppet-service :refer [jruby-puppet-pooled-service]]
+            [puppetlabs.services.puppet-profiler.puppet-profiler-service :refer [puppet-profiler-service]]
+            [puppetlabs.services.jruby-pool-manager.jruby-pool-manager-service :refer [jruby-pool-manager-service]]
+            [puppetlabs.services.jruby.puppet-environments :as puppet-env]
```
camlow325: Looks like this is unused.

test/integration/puppetlabs/services/jruby/module_info_test.clj
```diff
@@ -0,0 +1,86 @@
...
+            [puppetlabs.trapperkeeper.testutils.bootstrap :as tk-bootstrap]))
+
+(defn gen-modules
+  [[mod-dir manifests]]
+  (let [modules-dir (fs/file mod-dir "modules")]
+    (ks/mkdirs! modules-dir)
+    (doseq [manifest manifests]
+      (let [module-dir (fs/file modules-dir manifest)
+            metadata-json {"name" manifest
+                           "version" "1.0.0"
```
camlow325: Might be nice to vary the version between the modules?  Not a big deal.

camlow325:
I like the new `module_info_test` for validating the response from the `.getModuleInfoForEnvironment` closer to the JRuby layer.  I still think it would be good to have some sort of test which validates the use of the HTTP `environment_modules` endpoint as well so that the extra middleware functions that you added in the `master_core` namespace can be validated.

For the `environment_classes` endpoint, we added the following HTTP-layer tests:

https://github.com/puppetlabs/puppetserver/blob/2.7.2/test/integration/puppetlabs/services/master/environment_classes_int_test.clj#L103-L138

Could you look into adding something analogous to that for the `environment_modules` endpoint?  It could obviously be a lot simpler than the `environment_classes` ones since you shouldn't have to deal with caching, etc.

camlow325:
@gguillotte Whenever you have a minute, would be cool if you could take a peek at some of the docs being added for this new endpoint.  Thanks!

documentation/puppet-api/v3/environment_modules.markdown
```diff
@@ -0,0 +1,127 @@
...
+### Supported Formats
+
+JSON
+
+### Query Parameters
+
+Provide one parameter to the GET request:
+
+* `environment`: Only the modules information pertaining to the specified
+environment will be returned for the call.
```
gguillotte: I'd rewrite this:

```
* `environment`: Request information about modules pertaining to the specified environment only.
```

Also, is this parameter mandatory? If so, we should note that (L34, "Provide one *mandatory* parameter ...).

If it's not required, is there a default value or response?

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -327,6 +338,42 @@
         (rr/not-found (i18n/tru "Could not find environment ''{0}''" environment))))))
 
 (schema/defn ^:always-validate
+  module-info-from-jruby->module-info-for-json  :- EnvironmentModulesInfo
+  "Creates a new map with a top level name that corresponds to the requested environment
+  and a top level key modules which contains the module information obtained from JRuby."
```
gguillotte: > top level key modules

What is this referring to?

gguillotte:
@briancain One docs recommendation and a docstring question, then :+1:.

src/clj/puppetlabs/services/master/master_core.clj
briancain: @gguillotte I think maybe wrapping modules in \` would make that clearer. (i.e. a top level key `modules` which contains....)

briancain:
Thanks for the feedback @gguillotte , I left a comment and updated the two doc strings.

camlow325:
@briancain Are you still working on updating this based on feedback?  I think the biggest item left from my review is the addition of a higher-level test which would exercise the new middleware functions that this PR adds.

briancain:
@camlow325 - ah, oops I forgot to add that kind of test. I'll plan on doing that today. Thanks for the reminder!

briancain:
@camlow325 I've got a new test ready but I have some failures that I'm unsure about. I might go ahead and push it so we can talk about it on the PR if that works

test/integration/puppetlabs/services/master/environment_modules_int_test.clj
```diff
@@ -0,0 +1,392 @@
...
+                                      "2.0.0")
+
+             (let [production-response-before-flush (get-env-modules
+                                                     "production")
+                   test-response-before-flush (get-env-modules
+                                               "test")]
+               (is (= 304 (:status production-response-before-flush))
+                   (str
+                    "unexpected status code for prod response after code change "
+                    "but before flush"))
```
briancain: It seems like the following tests are failing. I kind of expected caching to work the same way as environment_classes with no changes on my end, so perhaps something it set up wrong here? For example:

```
FAIL in (environment-modules-integration-cache-enabled-test) (environment_modules_int_test.clj:258)
environment class cache invalidation for one environment
unexpected status code for prod response after code change but before flush
expected: (= 304 (:status production-response-before-flush))
  actual: (not (= 304 200))

lein test :only puppetlabs.services.master.environment-modules-int-test/environment-modules-integration-cache-enabled-test

FAIL in (environment-modules-integration-cache-enabled-test) (environment_modules_int_test.clj:262)
environment class cache invalidation for one environment
unexpected body for production response
expected: (empty? (:body production-response-before-flush))
  actual: (not (empty? "{\"modules\":[{\"name\":\"somemod\",\"version\":\"1.0.0\"}],\"name\":\"production\"}"))
```

I thought maybe the `purge-env-dir` needed to be updated but it looks like it completely deletes the environment folder which should include modules?

camlow325: @briancain I wouldn't have expected the caching to work with the environment_modules endpoint as this PR currently stands.

For the environment_classes endpoint, the middleware functions hash the response body and include an http etag, like [here](https://github.com/puppetlabs/puppetserver/blob/2.7.2/src/clj/puppetlabs/services/master/master_core.clj#L292-L301).  Since you don't have any of that in the environment_modules middleware functions, I wouldn't expect the endpoint to return a 304 when the module content doesn't change from one query to the next.

I don't necessarily think you need to bother cache-enabling the environment_modules endpoint at this point.  Assuming you aren't planning on having the analytics service query for this info super frequently, the performance without caching may be fine.  Enumerating module info should be a lot less heavy than what the environment_classes endpoint has to do to parse manifest info anyway and the caching layer adds a fair bit of complexity.

Setting aside the caching, I think you could thin these extra tests out considerably.  I'd be fine with just a single test which validates that a 200 OK is returned with a valid json payload containing modules and version info for a couple of different environments.

briancain: Nah, also the analytics endpoint will be going through my JRuby wrapper function, and not making an HTTP request. And the request isn't so frequent that it needs to be cached. I'll remove these, then. Thanks!

briancain:
@camlow325 - Ok! I think this PR is ready to go. Just did some cleanup of that middleware integration test.

test/integration/puppetlabs/services/jruby/module_info_test.clj
briancain: @camlow325 if you want, I can do what I did with my middleware integration test and make the module version a parameter that can be passed in.

dev-resources/puppetlabs/services/master/environment_modules_int_test/logback-environment-modules-integration-cache-enabled-test.xml
```diff
@@ -0,0 +1,15 @@
+<configuration scan="true">
```
camlow325: Is this file used anywhere?

briancain: Ah! Not anymore...I'll remove it.

test/integration/puppetlabs/services/master/environment_modules_int_test.clj
```diff
@@ -0,0 +1,78 @@
...
+  (:require [clojure.test :refer :all]
+            [puppetlabs.http.client.sync :as http-client]
+            [puppetlabs.kitchensink.core :as ks]
+            [puppetlabs.puppetserver.bootstrap-testutils :as bootstrap]
+            [puppetlabs.trapperkeeper.testutils.logging :as logutils]
+            [puppetlabs.puppetserver.testutils :as testutils]
+            [cheshire.core :as cheshire]
+            [me.raynes.fs :as fs]
+            [clojure.tools.logging :as log]
+            [puppetlabs.services.jruby.jruby-puppet-testutils :as jruby-puppet-testutils])
```
camlow325: It doesn't appear that the `log` or `jruby-puppet-testutils` requires or any of the `import` references are used.

test/integration/puppetlabs/services/master/environment_modules_int_test.clj
```diff
@@ -0,0 +1,78 @@
...
+  "./dev-resources/puppetlabs/services/master/environment_modules_int_test")
+
+(use-fixtures :once
+              (testutils/with-puppet-conf
+               (fs/file test-resources-dir "puppet.conf")))
+
+(defn get-env-modules
+  ([env-name]
+   (get-env-modules env-name nil))
+  ([env-name if-none-match]
```
camlow325: Looks like you could get rid of the overload which takes an `if-none-match` argument since the deftest doesn't use it.

test/integration/puppetlabs/services/master/environment_modules_int_test.clj
```diff
@@ -0,0 +1,78 @@
...
+(deftest ^:integration environment-modules-integration-cache-disabled-test
+  (testing "when environment modules cache is disabled for a module request"
+    (bootstrap/with-puppetserver-running-with-config
+     app
+     (-> {:jruby-puppet {:max-active-instances 1}}
+         (bootstrap/load-dev-config-with-overrides)
+         (ks/dissoc-in [:jruby-puppet
+                        :environment-class-cache-enabled]))
+     (logutils/with-test-logging
+       (let [foo-file (testutils/write-foo-pp-file
```
camlow325: May not need to write this file for this test to work?  Not sure...

camlow325: Ah, but I guess the metadata.json file is written behind this call so nevermind.

briancain: @camlow325 it probably doesn't need to be as complicated but I modified the `write-pp-file` function to lay down a metadata.json file as well based on the name passed in.

camlow325: I think it's fine as you have it.  Just misunderstood that earlier.

briancain: Aha! Makes sense :) 👍

test/integration/puppetlabs/services/master/environment_modules_int_test.clj
```diff
@@ -0,0 +1,78 @@
...
+         {:as :text}
+         opts))
+       (catch Exception e
+         (throw (Exception. "environment_modules http get failed" e)))))))
+
+(defn response->module-info-map
+  [response]
+  (-> response :body cheshire/parse-string))
+
+(deftest ^:integration environment-modules-integration-cache-disabled-test
```
camlow325: It doesn't look like there's anything in this test which cleans out the environments directory after the test is done running.  I think we should do that so that the files setup for this test don't accidentally break some other test downstream.  In the similar environment-classes-int-tests, we used a [test fixture](https://github.com/puppetlabs/puppetserver/blob/2.7.2/test/integration/puppetlabs/services/master/environment_classes_int_test.clj#L44-L60) which purges the directory before / after the test is run.  Could do something like that with this test, too.

briancain:
@camlow325 woo, thanks! Pushed up changes to clean up the testing dir like what you linked to. I'm assuming the `(use-fixtures :each ...)` function will run for each test automatically?

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -124,6 +124,17 @@
...
    :files [EnvironmentClassesFileEntry]})
 
+(def EnvironmentModuleInfo
+  "Schema for a given module that can be returned within the
+  :modules key in EnvironmentModulesInfo."
+  {:name schema/Str
+   :version schema/Str})
+
+(def EnvironmentModulesInfo
+  "Schema for the return payload an environment_classes request."
```
jpinsonault: Looks like a word went missing here somewhere

briancain: Thanks joe! Pushed a fix.

briancain:
@jpinsonault @camlow325 I think this is good to go now!

PR #1311, opened 2017-01-09
(SERVER-1671) Don't log stdout from shell utils

camlow325:
Previously, when ShellUtils was configured to combine both stderr and
stdout from the executed command, the combination of the two was
logged as a warn level message.  With the changes in this commit, none
of the info written to stdout is logged.  Only the info written to
stderr, if applicable, is logged.

### DIFF
--- a/src/java/com/puppetlabs/puppetserver/ShellUtils.java
+++ b/src/java/com/puppetlabs/puppetserver/ShellUtils.java
@@ -3,6 +3,7 @@ package com.puppetlabs.puppetserver;
 import org.apache.commons.exec.DefaultExecutor;
 import org.apache.commons.exec.PumpStreamHandler;
 import org.apache.commons.exec.CommandLine;
+import org.apache.commons.io.output.TeeOutputStream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


test/unit/puppetlabs/puppetserver/shell_utils_test.clj
```diff
@@ -32,9 +34,26 @@
 
 (deftest returns-stderr-correctly
   (testing "echo can add content to stderr as well"
-    (is (= "bar\n" (:stderr (sh-utils/execute-command
-                             (script-path "warn")
-                             {:args ["bar"]}))))))
+    (logging/with-test-logging
```
camlow325: This is just for Wolfe Principle correctness.

src/java/com/puppetlabs/puppetserver/ShellUtils.java
```diff
@@ -72,15 +73,18 @@ private static ExecutionResult executeExecutor(CommandLine commandLine,
         //    PipedInputStream stdoutInputStream = new PipedInputStream(stdoutOutputStream);
         // but this requires that the input stream be read on a different thread
         // than this one. this is currently out of scope.
-        ByteArrayOutputStream stdoutOutputStream;
+        ByteArrayOutputStream stdoutOutputStream = new ByteArrayOutputStream();
```
haus: @camlow325 do the comments above this section still apply?

camlow325: Yeah, we could still do something like the `PipedOutputStream` rather than writing to a `ByteArrayOutputStream` and having to read the bytes back from that.  Might save a bit of memory if we did it that way instead.  I don't think my latest changes helps or hurts with this.

haus:
@camlow325 you want a second set of eyes on this or should i just merge?

camlow325:
@jpinsonault You have time to take a quick peek at this?  Not much code.

jpinsonault:
👍

PR #1312, opened 2017-01-10
(PE-19130) Allow puppet conf file to be resources.

aperiodic:
Change the puppetserver testutils to allow the sources of the temporary
puppet conf files set up by with-puppet-conf-files to be java.net.URLs
in addition to strings or java.io.Files, and to copy the sources to the
destination files by wrapping the sources in an input stream, which
works even when the source is a URL that points to a classpath resource
contained in a jar.

Together, these changes allow the conf file sources passed to
with-puppet-conf-files to be classpath resources distributed in a jar,
instead of requiring them to be files.

### DIFF
--- a/test/integration/puppetlabs/puppetserver/testutils.clj
+++ b/test/integration/puppetlabs/puppetserver/testutils.clj
@@ -1,17 +1,19 @@
 (ns puppetlabs.puppetserver.testutils
-  (:require [me.raynes.fs :as fs]
+  (:require [clojure.java.io :as io]
+            [me.raynes.fs :as fs]
             [schema.core :as schema]
             [cheshire.core :as json]
             [puppetlabs.http.client.sync :as http-client]
             [puppetlabs.kitchensink.core :as ks]
             [puppetlabs.puppetserver.ringutils :as ringutils])
-  (:import (java.io File)))
+  (:import (java.io File)
+           (java.net URL)))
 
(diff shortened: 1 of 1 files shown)


test/integration/puppetlabs/puppetserver/testutils.clj
```diff
@@ -97,8 +99,12 @@
   [puppet-conf-files :- PuppetConfFiles
    target-paths :- PuppetConfFiles]
   (doseq [filename (keys puppet-conf-files)]
-    (fs/copy+ (get puppet-conf-files filename)
-              (get target-paths filename))))
+    (let [dest (io/file (get target-paths filename))
+          dest-dir (.getParentFile dest)]
+      (when (and dest-dir (not (.exists dest-dir)))
+        (.mkdirs dest-dir))
```
camlow325: Could do this instead to eliminate a couple of Java method calls and have the function blow up if the directory creation fails for some reason:

```clj
(when dest-dir
  (ks/mkdirs! dest-dir)
```

aperiodic: Seems reasonable.

aperiodic:
@camlow325 updated to use `ks/mkdirs!`

camlow325:
Thanks, @aperiodic.  This all seems fine to me.  Since this is a test-only change and the existing tests pass with it, going to merge it.

PR #1313, opened 2017-01-11
(PE-13552) Update tests for new jruby-utils-version

jpinsonault:
This PR is in response to https://github.com/puppetlabs/jruby-utils/pull/34
It updates the jruby-utils dep to an unreleased version and updates the tests to work with the new changes 

The "503 during shutdown" tests have been modified to test for 503 after
shutdown, because it's no longer possible to pause the shutdown in the same
way

It can't be merged until a new jruby-utils is released with the new changes from https://github.com/puppetlabs/jruby-utils/pull/34

### DIFF
--- a/test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
+++ b/test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
@@ -91,7 +91,7 @@
 (defn add-watch-for-flush-complete
   [pool-context]
   (let [flush-complete (promise)]
-    (add-watch (get-in pool-context [:internal :pool-agent]) :flush-callback
+    (add-watch (get-in pool-context [:internal :modify-instance-agent]) :flush-callback
                (fn [k a old-state new-state]
                  (when (= k :flush-callback)
                    (remove-watch a :flush-callback)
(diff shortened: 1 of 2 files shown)


jpinsonault:
More detail on the 503 tests

Even though we can't figure out a way to pause the shutdown and make sure a 503 is returned, we can still test that once the shutdown has occurred, a 503 is returned. Technically any requests that come in during a shutdown will block until the shutdown is complete, and then return a 503. So it kind of amounts to the same thing

camlow325:
I think I'm good with the reasoning on the test changes given the latest JRuby pool behavior.  I definitely still think there's value in validating 503 errors post-JRuby service shutdown and agree that being able to test for 503 errors while the shutdown is in progress would be difficult unless we were to mock out the lock behavior to wait for the pool to fill back up, which doesn't seem valuable since it wouldn't happen in the real world anyway.

jpinsonault:
Pushed a commit to update jruby-utils to the released version

jpinsonault:
ping @rlinehan for review

haus:
@jpinsonault seems like it would probably be worth squashing these two commits together, as tests won't pass on the first commit, which will make a git bisect very sad.

jpinsonault:
@haus good call, will do

jpinsonault:
squashed

haus:
sweet

test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
```diff
@@ -284,7 +267,7 @@
            (tk-app/stop app)))))))
 
 (deftest ^:integration test-503-when-jruby-is-first-to-shutdown
-  (testing "During a shutdown requests result in 503 http responses"
+  (testing "After a shutdown, requests result in 503 http responses"
```
rlinehan: I'm having a hard time understanding what exactly the difference is between this test and the test above (`test-503-when-app-shuts-down`). It seems like the difference now is that in the first we shut down the whole app, but in this one we just shut down the `JRubyService`? Maybe it would be good to add a little bit more to the testing strings or comments to distinguish them? (I guess this isn't really a comment entirely about this PR...)

rlinehan:
Had one comment about the 503/shutdown tests, otherwise LGTM.

test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
jpinsonault: I had to go look at those again for a while, but I think you're right. I can add some detail, because it was really confusing

jpinsonault:
Pushed a commit that rewords the test strings and adds a comment to clarify the difference

PR #1320, opened 2017-01-20
(MAINT) Add PuppetDB info to the running from source docs

senior:
This commit adds some docs on running a source based PuppetDB instance
with a source based Puppet Server instances. I also added some
clarification around invoking the Puppet Agent that I found helpful
when setting up a dev environment.

### DIFF
--- a/documentation/dev_running_from_source.markdown
+++ b/documentation/dev_running_from_source.markdown
@@ -159,6 +159,21 @@ run the following:
 
     $ lein run -c /path/to/puppetserver.conf
 
+Running the Agent
+-----
+
+Use a command like the one below to run an agent against your running puppetserver:
+
+        puppet agent --confdir ~/.puppetlabs/etc/puppet \
+                     --debug -t
+
+Note that a system installed Puppet Agent is ok for use with
+source-based PuppetDB and Puppet Server. The `--confdir` above
+specifies the same confdir that Puppet Server is using. Since the
+Puppet Agent and Puppet Server instances are both using the same
+confdir, they're both using the same certificates as well. This
+alleviates the need to sign certificates as a separate step.
+
 Running tests
 -----
 
@@ -176,13 +191,13 @@ environment via either of the following methods:
 
 1) An environment variable named `PUPPETSERVER_HEAP_SIZE`.  For example, to
   use a heap size of 6 GiB for a `lein test` run, you could run the following:
-  
+
     $ PUPPETSERVER_HEAP_SIZE=6G lein test
-  
+
 2) A lein `profiles.clj` setting in the `:user` profile under the
   `:puppetserver-heap-size` key.  For example, to use a heap size of 6 GiB, you
   could add the following key to your `~/.lein/profiles.clj` file:
-  
+
 ```clj
 {:user {:puppetserver-heap-size "6G"
         ...}}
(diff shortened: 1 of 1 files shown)


documentation/dev_running_from_source.markdown
```diff
@@ -159,6 +159,25 @@ run the following:
 
     $ lein run -c /path/to/puppetserver.conf
 
+Running the Agent
+-----
+
+Use a command like the one below to run an agent against your running puppetserver:
+
+        puppet agent --confdir ~/.puppetlabs/etc/puppet \
```
camlow325: I think if you specify the --confdir setting that you wouldn't need the --server setting since it should read the value for the server setting that was populated into the ~/.puppetlabs/etc/puppet/puppet.conf file, which should be equal to whatever you used for <MASTERHOST>.  Ideally, you wouldn't need to specify the --confdir setting either for a Puppet v4.x agent since ~/.puppetlabs/etc/puppet should be the default.  Doesn't hurt anything either way, though.

documentation/dev_running_from_source.markdown
```diff
@@ -214,3 +233,50 @@ Debugging
...
+for connecting to PuppetDB. Instructions on this configuration are
+below, but the official docs for this can be found
+[here](https://docs.puppet.com/puppetdb/4.3/connect_puppet_master.html).
+
+Update `~/.puppetlabs/etc/puppet/puppet.conf` to include:
+
+    [master]
+    storeconfigs = true
+    storeconfigs_backend = puppetdb
+    reports = store,puppetdb
```
camlow325: Should we bother suggesting to configure the `store` report terminus here or just use `reports = puppetdb`?

documentation/dev_running_from_source.markdown
senior: On Debian, not sure where else, it's not looking there, it wants to find it in ~/.puppet for whatever reason. I'm happy to put caveats or something else above, but I know I've fumbled around trying to remember the right config options to make this work and they aren't documented.

camlow325: Works for me to include the `--confdir` flag.  Still think we might be able to avoid using the `--server` flag in that case, but I'm good either way.

haus: @senior are you using the distro package and not ours?

documentation/dev_running_from_source.markdown
camlow325: I guess we list `store` in the puppetdb docs, so probably fine to keep this here for consistency.

camlow325:
Might be nice to make an explicit reference in these docs to where you could go to find help on running PuppetDB from source.  Not sure if this is the best current reference - https://docs.puppet.com/puppetdb/4.3/install_from_source.html?

camlow325:
One specific thing I had to do to get this to work which might be useful here is the SSL certificate setup.  When PuppetDB's terminus, running inside of Puppet Server, makes requests to a PuppetDB instance, it will expect the server certificate that PuppetDB serves up be issued from the same CA that issued Puppet Server's own server certificate.  For development, I've found it most convenient to just use the same certs for the agent, Puppet Server, and Puppet DB.

For PuppetDB, then, I've added these settings to PuppetDB's config.ini:

```ini
[jetty]

port=8080
ssl-port = 8081
ssl-cert = <home dir>/.puppetlabs/etc/puppet/ssl/certs/<MASTERHOST>.pem
ssl-key = <home dir>/.puppetlabs/etc/puppet/ssl/private_keys/<MASTERHOST>.pem
ssl-ca-cert = <home dir>/.puppetlabs/etc/puppet/ssl/certs/ca.pem
```

I restart PuppetDB with these settings *after* starting up Puppet Server.  That way, Puppet Server will have created the certs and keys before PuppetDB tries to use them.

Do you think it would be helpful to work something about the above into these docs?

documentation/dev_running_from_source.markdown
senior: @haus Yes, 4.8.1 from stretch

documentation/dev_running_from_source.markdown
senior: Yeah, that's why I had it in there, just a direct copy and paste from the setup instructions.

documentation/dev_running_from_source.markdown
senior: @camlow325 just tested it out locally, you're right about the server getting pulled from the confdir and server no longer being necessary. I'll remove that part

senior:
@camlow325 I use the same PDB SSL setup you do, that seems like useful info to add, I'll let you know when it's updated

senior:
@camlow325 updated

documentation/dev_running_from_source.markdown
```diff
@@ -159,6 +159,24 @@ run the following:
...
+        puppet agent --confdir ~/.puppetlabs/etc/puppet \
+                     --debug -t
+
+Note that a system installed Puppet Agent is ok for use with
+source-based PuppetDB and Puppet Server. The `--confdir` above
+specifies the same confdir that Puppet Server is using. Since the
+Puppet Agent and Puppet Server instances are both using the same
+confdir, they're both using the same certificates as well. This
+alleviates the need to sign certificates as a separate step. The
+`--server` flag will also indicate the host puppetserver is on. If you
```
camlow325: Probably should take out the line about the `--server` flag now that the example command line doesn't have that in it.

camlow325: @senior Ping.  You have time to remove this comment?  I can merge it after that.

senior: Good catch, I remove that from the description.

PR #1322, opened 2017-01-24
(PE-18344) Set http client metric registry if metrics service is present

rlinehan:
Use the Trapperkeeper optional service dependency functionality to get a
metric registry if the metrics service is present, and pass it and the
`server-id` through to the http client that puppet server uses, so that the
http client will register metrics. If no metrics service is present, the http
client will not get passed a metric registry or server-id and will not
register metrics.


camlow325:
Do you think it would be hard to add a test which does an actual request through the http client and validates that some reasonable-looking metrics were stored for the request?

camlow325:
Oh, probably not since we haven't plumbed metric-ids on any of the individual requests yet.

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
```diff
@@ -15,18 +16,25 @@
...
                           jruby/JRubyPuppetService
-                          [[:ConfigService get-config]
-                           [:ShutdownService shutdown-on-error]
-                           [:PuppetProfilerService get-profiler]
-                           [:PoolManagerService create-pool]]
+                          {:required [[:ConfigService get-config]
+                                      [:ShutdownService shutdown-on-error]
+                                      [:PuppetProfilerService get-profiler]
+                                      [:PoolManagerService create-pool]]
+                           :optional [MetricsService]}
```
camlow325: If we went ahead and made the metrics service a required dependency at this point (including putting it in the bootstrap.cfg) -- without pulling any of the PE-specific stuff down into OSS yet -- could that simplify this code a bit?  Can you think of any problems that might cause?

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -232,11 +242,12 @@
   ([raw-config :- {:jruby-puppet {schema/Keyword schema/Any}
                    (schema/optional-key :http-client) {schema/Keyword schema/Any}
                    schema/Keyword schema/Any}]
-   (initialize-and-create-jruby-config raw-config nil (fn []) false))
+   (initialize-and-create-jruby-config raw-config nil (fn []) false nil))
```
rlinehan: Yes to both of you. @camlow325 this arity is only used by the CLI tools, and I think it's fine that they won't get a metric registry. @jpinsonault yeah, these will always be set to nil. I could change it so that they're only set if they are set to something in the map? Either way will work - that might be a little less confusing?

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
```diff
@@ -15,18 +16,25 @@
...
+
   (init
     [this context]
     (log/info (i18n/trs "Initializing the JRuby service"))
-    (let [jruby-config (core/initialize-and-create-jruby-config
-                        (get-config)
+    (let [config (get-config)
+          metrics-registry (when-let [metrics-service (tk-services/maybe-get-service this :MetricsService)]
+                             {:metric-registry (metrics/get-metrics-registry metrics-service :puppetserver)
+                              :server-id (get-in config [:metrics :server-id])})
```
rlinehan: I guess so. I don't really feel like it's that big a deal to be getting it out of the config, but it might be cleaner to have a service function.

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
rlinehan: Yeah, I think that can simplify the code, and probably makes sense now that we have permission to move all the rest of the metrics to open source. The only thing is that we'll need to add a metrics service to the tk bootstrap, but that shouldn't really be a problem.

My only question is should we have a setting about whether to use http metrics or not? We haven't done any performance testing with the http metrics, and there might be a possibility it could affect performance. I guess I'm thinking it wouldn't be a bad thing to have a way to turn it off.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
jpinsonault: I don't have a strong opinion, it's fine how it is

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
jpinsonault: Right now we only have the ability to enable or disable the entire metrics system right? Would adding a setting just for the http metrics be weird?

rlinehan: @jpinsonault You can toggle whether to report metrics to jmx or not, as well as whether to report metrics to graphite or not.. You can also enable/disable the profiler. I was thinking that this was maybe kind of akin to the profiler - a couple folks I've talked to have reported that they were maybe seeing some performance issues with that, so they turned it off. I would be concerned that if we might run into some performance issues with these metrics and so we would want a way to turn them off in case someone upgraded and ran into performance issues.

camlow325: @rlinehan As we talked about IRL yesterday, I'm good with the idea of creating a new setting which would allow a user to turn off the `http-client` specific metrics in the event that they may cause performance problems, etc.  Where do you think such a setting should live?  I was thinking maybe having a `metrics-enabled` setting in the `http-client` area of the TK config, where we already have client-specific settings like `ssl-protocols` and `cipher-suites`, might make sense.  Could default the setting if not specified to `true` just to make it easier for users to use them for the majority cases where hopefully they won't cause any problems.  WDYT?

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
camlow325: I went ahead and created a separate ticket for this, [TK-433](https://tickets.puppetlabs.com/browse/TK-433), and threw it into the OSS Puppet Server metrics epic, [SERVER-1259](https://tickets.puppetlabs.com/browse/SERVER-1259) for now.

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
camlow325: @rlinehan - were you hoping to do the work to make tk-metrics a required dependency and add a new toggle setting in this PR or later on?

rlinehan: @camlow325 I was thinking in this PR, I just haven't managed to get to it yet.

rlinehan: Putting the setting under `http-client` and defaulting it to `true` makes sense to me.

rlinehan:
@camlow325 @jpinsonault updated!

project.clj
```diff
@@ -61,6 +61,7 @@
                  [puppetlabs/jruby-utils "0.7.0"]
                  [puppetlabs/trapperkeeper]
                  [puppetlabs/trapperkeeper-authorization]
+                 [puppetlabs/trapperkeeper-metrics "0.6.0"]
```
camlow325: We don't have to hold up the PR for this change, but might be nice to bump this version in clj-parent and update to a new clj-parent from puppetserver.  Nice to avoid pinning common version dependencies when we can.

test/integration/puppetlabs/puppetserver/bootstrap_testutils.clj
```diff
@@ -52,19 +52,19 @@
         (assoc-in [:jruby-puppet :master-log-dir] master-log-dir)
         (ks/deep-merge overrides))))
 
+(def services-from-dev-boostrap
```
camlow325: s/boostrap/bootstrap, although boozetrap works too.  :)

test/integration/puppetlabs/services/jruby/class_info_test.clj
```diff
@@ -75,9 +72,7 @@
 
       (tk-bootstrap/with-app-with-config
        app
-       [jruby-puppet-pooled-service
```
camlow325: These were really nice to cleanup.  Thanks!

jpinsonault:
Wow, nice cleanup in the tests. I'm 👍. Looks like just the one typo left? If we care

project.clj
rlinehan: I think I had been thinking that it would be bad to bump this is tk-metrics because then pe-tk-metrics would be incompatible... but if we release pe-tk-metrics too and then bump both in clj-parent simultaneously, that will work. I'll do that.

rlinehan:
@camlow325 @jpinsonault updated!

PR #1324, opened 2017-01-24
(MAINT) Bump to clj-parent 0.3.3

camlow325:
This commit bumps the clj-parent version from 0.3.2 to 0.3.3 in order to
pick up the `request-body-max-size` feature in
trapperkeeper-webserver-jetty9 1.7.0.  This commit also updates the
clj-i18n plugin dependency version to 0.6.0 for consistency with the
version of clj-i18n pulled in from the new clj-parent version.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -32,7 +32,7 @@
 
   :min-lein-version "2.7.1"
 
-  :parent-project {:coords [puppetlabs/clj-parent "0.3.2"]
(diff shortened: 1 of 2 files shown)


camlow325:
Travis tests failed on the initial bump to clj-parent 0.3.3 due to the tk-jetty9 fix for gzip encoding causing a couple of the environment-classes tests to fail.  In 9ac3325, I fixed up the tests to properly allow for the presence of gzip encoding in HTTP responses.

PR #1325, opened 2017-01-28
(PDB-2640) Add `:compress` option to Ruby http_client post

camlow325:
This commit adds an option to the http_client Ruby post method for
specifying compression that should be applied to the request payload.
The only supported compression mode is `gzip`.  The new option, if
specified, is passed on through to the underlying http-client library.

### DIFF
--- a/test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
+++ b/test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
@@ -3,7 +3,8 @@
            (org.apache.http ConnectionClosedException)
            (javax.net.ssl SSLHandshakeException)
            (java.util HashMap)
-           (java.io IOException))
+           (java.io IOException)
+           (java.util.zip GZIPInputStream))
   (:require [clojure.test :refer :all]
             [puppetlabs.trapperkeeper.testutils.logging :as logutils]
             [puppetlabs.trapperkeeper.testutils.webserver :as jetty9]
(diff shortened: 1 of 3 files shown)


camlow325:
This PR can't be merged until a corresponding PR in clj-http-client, https://github.com/puppetlabs/clj-http-client/pull/67, is merged and a new http-client library is released.  At that time, I'll replace the clj-http-client SNAPSHOT dependency (via a clj-parent bump).  At that point, this should be ready to merge.  Just putting this up for initial review for now.

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -191,6 +200,33 @@ jruby-config :- jruby-schemas/JRubyConfig
...
+                  (.runScriptlet
+                   sc
+                   (str "begin;"
+                        "  $response = $c.post('/', 'foo', {}, "
+                        compress
+                        ");"
+                        "  'No error raised from post';"
+                        "rescue ArgumentError => e;"
+                        "  e.message;"
+                        "end"))))))))))
```
rlinehan: yuck, this is pretty gross. but also totally works and there's probably not a better way to do it.

camlow325: Yeah, I had something earlier that tried to let the exception / error bubble up from JRuby into the Java layer but that created a pretty gross Wolfe Principle violation that I couldn't figure out how to get rid of easily. JRuby wants to log a full stack trace for any uncaught error that bubbles up. This is definitely gross but seemed a bit less gross.

camlow325:
@rlinehan Yep, should be good to go now since I updated dependencies following the clj-http-client release with the new compression feature and Travis tests appear to be passing.

camlow325:
@jpinsonault Whenever you have a minute, would you mind doing a second review on this?  Thanks!

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
jpinsonault: huh, I agree though

jpinsonault:
👍

PR #1327, opened 2017-02-04
(SERVER-1723) Use /dev/urandom for gem, ruby, and irb subcommands

camlow325:
This commit adds -Djava.security.egd=/dev/urandom as an argument to the
Java command line for the gem, ruby, and irb subcommands.  This argument
was already being used in the Java command line for when puppetserver
runs as a service.  Performance for some operations like gem installs
appears to be significantly faster with this option in place.

### DIFF
--- a/resources/ext/cli/gem.erb
+++ b/resources/ext/cli/gem.erb
@@ -2,7 +2,7 @@
 
 umask 0022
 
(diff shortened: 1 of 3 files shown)


rlinehan:
This seems simple enough that I'm just gonna merge it.

PR #1330, opened 2017-02-10
(SERVER-1704) Support flush timeout setting for jruby pool

jpinsonault:
**This PR goes with** a jruby-utils PR which implements the timeout functionality: https://github.com/puppetlabs/jruby-utils/pull/40

---

This commit adds support for a timeout during a jruby pool flush through the
admin api.

Modifies the admin api routes to support returning a 503 if the timeout is
exceeded.

As a bonus it nukes a liberator resource too!

This shouldn't be merged until the corresponding jruby-utils PR is merged and a new release happens

### DIFF
--- a/src/clj/puppetlabs/services/puppet_admin/puppet_admin_core.clj
+++ b/src/clj/puppetlabs/services/puppet_admin/puppet_admin_core.clj
@@ -10,7 +10,8 @@
             [puppetlabs.puppetserver.ring.middleware.params :as pl-ring-params]
             [puppetlabs.i18n.core :as i18n]
             [slingshot.slingshot :as sling]
-            [ring.util.response :as rr]))
+            [ring.util.response :as rr]
+            [clojure.tools.logging :as log]))
 
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
@@ -58,6 +59,7 @@
     {:status 204}
     ; If a lock timeout occurs return a 503 Service Unavailable
     (catch [:kind :puppetlabs.services.jruby-pool-manager.impl.jruby-agents/jruby-lock-timeout] e
+      (log/error (:msg e))
       (-> (rr/response (:msg e))
           (rr/status 503)
           (rr/content-type "text/plain")))))
--- a/src/clj/puppetlabs/services/puppet_admin/puppet_admin_core.clj
+++ b/src/clj/puppetlabs/services/puppet_admin/puppet_admin_core.clj
@@ -6,10 +6,11 @@
             [puppetlabs.puppetserver.liberator-utils :as liberator-utils]
             [schema.core :as schema]
             [liberator.core :refer [defresource]]
-    ;[liberator.dev :as liberator-dev]
             [puppetlabs.comidi :as comidi]
             [puppetlabs.puppetserver.ring.middleware.params :as pl-ring-params]
-            [puppetlabs.i18n.core :as i18n]))
+            [puppetlabs.i18n.core :as i18n]
+            [slingshot.slingshot :as sling]
+            [ring.util.response :as rr]))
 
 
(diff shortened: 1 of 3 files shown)


src/clj/puppetlabs/services/puppet_admin/puppet_admin_core.clj
```diff
@@ -49,33 +50,17 @@
...
-  :delete!
-  (fn [context]
-    (jruby-puppet/flush-jruby-pool! jruby-service)))
-
+  (sling/try+
+    (jruby-puppet/flush-jruby-pool! jruby-service)
+    ; 204 No Content on success
+    {:status 204}
+    ; If a lock timeout occurs return a 503 Service Unavailable
+    (catch [:kind :puppetlabs.services.jruby-pool-manager.impl.jruby-agents/jruby-lock-timeout] e
```
camlow325: When I run this locally, a timeout doesn't cause anything to be written to the Puppet Server log beyond just the "Draining and refilling" message.  I think it would be good to at least log some sort of error in the case of a flush timeout so it's easy to see from looking at the log when the flush failed.  WDYT?

jpinsonault: Well if it's not a concern then we might as well leave it with the whole namespace to be explicit. Ah yeah good catch, it should log an error

test/integration/puppetlabs/services/puppet_admin/puppet_admin_int_test.clj
```diff
@@ -178,5 +181,26 @@
...
+       (let [jruby-service (tk-app/get-service app :JRubyPuppetService)
+             pool-context (jruby-protocol/get-pool-context jruby-service)
+             pool (jruby-core/get-pool pool-context)
+             borrowed-instance (.borrowItem pool)]
+         (try
+           (let [response (http-client/delete
+                           (str "https://localhost:8140/puppet-admin-api/v1/jruby-pool")
+                           ssl-request-options)]
+             (is (= 503 (:status response)))
+             (is (= "Timeout limit reached before lock could be granted"
```
rlinehan: I think this is going to need to be changed once you update to a released jruby-utils, cuz the string there has changed.... But you'll probably realize that cuz the test will fail :P.

src/clj/puppetlabs/services/puppet_admin/puppet_admin_core.clj
```diff
@@ -6,10 +6,12 @@
             [puppetlabs.puppetserver.liberator-utils :as liberator-utils]
             [schema.core :as schema]
             [liberator.core :refer [defresource]]
-    ;[liberator.dev :as liberator-dev]
```
jpinsonault: Ah, ok, well I'm fine with leaving it

test/integration/puppetlabs/services/puppet_admin/puppet_admin_int_test.clj
jpinsonault: Thanks for pointing it out either way!

test/integration/puppetlabs/services/puppet_admin/puppet_admin_int_test.clj
```diff
@@ -178,5 +181,26 @@
...
+       app
+       {:jruby-puppet {:flush-timeout 0}}
+       ; Borrow a jruby instance so that the timeout of 0 is reached immediately
+       (let [jruby-service (tk-app/get-service app :JRubyPuppetService)
+             pool-context (jruby-protocol/get-pool-context jruby-service)
+             pool (jruby-core/get-pool pool-context)
+             borrowed-instance (.borrowItem pool)]
+         (try
+           (let [response (http-client/delete
+                           (str "https://localhost:8140/puppet-admin-api/v1/jruby-pool")
```
jpinsonault: derp, copy paste oversight

jpinsonault:
Once https://github.com/puppetlabs/jruby-utils/pull/41 is merged I'll do a release and push another set of commits

jpinsonault:
Addressed the last set of feedback, updated jruby-utils to the released version, and rebased to change the commit message to explain why the liberator resource was removed

rlinehan:
🎈 🎉

PR #1331, opened 2017-02-14
(PE-18345) Pass metric-id setting through to http client

rlinehan:
This makes it so that if a `metric-id` request option is set for a GET
or POST request (the only two request types we support), it is passed through
to the http client's request options. Doing this also meant a small refactor
of some common logic between the `get` and `post` methods, as they both shared
most of the same logic for creating `RequestOptions`.

### DIFF
--- a/test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
+++ b/test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
@@ -21,13 +21,15 @@
             [puppetlabs.puppetserver.testutils :as testutils :refer
              [ca-cert localhost-cert localhost-key ssl-request-options]]
             [puppetlabs.trapperkeeper.testutils.logging :as logging]
+            [puppetlabs.trapperkeeper.services.protocols.metrics :as metrics-protocol]
             [clojure.tools.logging :as log]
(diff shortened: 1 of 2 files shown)


test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
```diff
@@ -384,6 +386,90 @@
...
+                            (add-metric-ns "with-metric-id.foo.post")}
+                          (set (map :metric-name metric-id-data)))))
+                 (is (= 2 (:count (first (metrics/get-client-metrics-data-by-metric-id
+                                          metric-registry ["foo"])))))
+                 (is (= 1 (:count (first (metrics/get-client-metrics-data-by-metric-id
+                                          metric-registry ["foo" "get"])))))
+                 (is (= 1 (:count (first (metrics/get-client-metrics-data-by-metric-id
+                                          metric-registry ["foo" "post"]))))))))
+           (finally
+             (jruby-testutils/return-instance jruby-service jruby-instance :no-metrics-test))))))))
```
camlow325: Seems like the last argument in this call should be `:http-client-metrics-test` - although it probably doesn't really matter.

jpinsonault:
Looks good to me 👍

rlinehan:
@jb fixed the metric id names!

PR #1333, opened 2017-03-01
(SERVER-1262) Wrap http metrics for comidi routes

camlow325:
This PR imports work done in pe-puppet-server-extensions for
wrapping the comidi routes for the master and legacy-routes services
with http-metrics.

### DIFF
--- a/src/clj/puppetlabs/services/master/master_service.clj
+++ b/src/clj/puppetlabs/services/master/master_service.clj
@@ -6,18 +6,22 @@
             [puppetlabs.trapperkeeper.services :as tk-services]
             [puppetlabs.comidi :as comidi]
             [puppetlabs.dujour.version-check :as version-check]
+            [puppetlabs.metrics.http :as http-metrics]
             [puppetlabs.services.protocols.master :as master]
-            [puppetlabs.i18n.core :as i18n]))
+            [puppetlabs.i18n.core :as i18n]
+            [puppetlabs.trapperkeeper.services.status.status-core :as status-core]))
 
 (defservice master-service
   master/MasterService
   [[:WebroutingService add-ring-handler get-route]
    [:PuppetServerConfigService get-config]
    [:RequestHandlerService handle-request]
+   [:MetricsService get-metrics-registry initialize-registry-settings]
    [:CaService initialize-master-ssl! retrieve-ca-cert! retrieve-ca-crl! get-auth-handler]
    [:JRubyPuppetService]
    [:AuthorizationService wrap-with-authorization-check]
    [:SchedulerService interspaced]
+   [:StatusService register-status]
    [:VersionedCodeService get-code-content]]
   (init
    [this context]
@@ -63,8 +67,18 @@
                                                 (get-auth-handler)
                                                 environment-class-cache-enabled)
            routes (comidi/context path ring-app)
+           route-metadata (comidi/route-metadata routes)
            comidi-handler (comidi/routes->handler routes)
-           ring-handler (comidi/wrap-with-route-metadata comidi-handler routes)]
+           registry (get-metrics-registry :puppetserver)
+           metrics-server-id (get-in config [:metrics :server-id])
+           metrics (http-metrics/initialize-http-metrics!
+                    registry
+                    metrics-server-id
+                    route-metadata)
+           ring-handler (-> comidi-handler
+                            (http-metrics/wrap-with-request-metrics metrics)
+                            (comidi/wrap-with-route-metadata routes))]
+
        ;; if the webrouting config uses the old-style config where
        ;; there is a single key with a route-id, we need to deal with that
        ;; for backward compat.  We have a hard-coded assumption that this route-id
@@ -81,8 +95,15 @@
                             :normalize-request-uri true})
          (add-ring-handler this
                            ring-handler
-                           {:normalize-request-uri true}))))
-   context)
+                           {:normalize-request-uri true}))
+
+       (register-status
+        "master"
+        (status-core/get-artifact-version "puppetlabs" "puppetserver")
+        1
+        (partial core/v1-status metrics))
+
+       (assoc context :http-metrics metrics))))
   (start
     [this context]
     (log/info (i18n/trs "Puppet Server has successfully started and is now ready to handle requests"))
--- a/src/clj/puppetlabs/services/master/master_service.clj
+++ b/src/clj/puppetlabs/services/master/master_service.clj
@@ -27,25 +27,48 @@
    [this context]
    (core/validate-memory-requirements!)
    (let [config (get-config)
+         route-config (core/get-master-route-config ::master-service config)
+         path (core/get-master-mount ::master-service route-config)
          certname (get-in config [:puppetserver :certname])
          localcacert (get-in config [:puppetserver :localcacert])
          puppet-version (get-in config [:puppetserver :puppet-version])
-         hostcrl (get-in config [:puppetserver :hostcrl])
          settings (ca/config->master-settings config)
+         metrics-server-id (get-in config [:metrics :server-id])
+         jruby-service (tk-services/get-service this :JRubyPuppetService)
+         use-legacy-auth-conf (get-in config
+                                      [:jruby-puppet :use-legacy-auth-conf]
+                                      true)
+         environment-class-cache-enabled (get-in config
+                                                 [:jruby-puppet
+                                                  :environment-class-cache-enabled]
+                                                 false)
+         ring-app (core/construct-root-routes puppet-version
+                                              use-legacy-auth-conf
+                                              jruby-service
+                                              get-code-content
+                                              handle-request
+                                              (get-auth-handler)
+                                              environment-class-cache-enabled)
+         routes (comidi/context path ring-app)
+         route-metadata (comidi/route-metadata routes)
+         comidi-handler (comidi/routes->handler routes)
+         registry (get-metrics-registry :puppetserver)
+         metrics (http-metrics/initialize-http-metrics!
+                  registry
+                  metrics-server-id
+                  route-metadata)
+         ring-handler (-> comidi-handler
+                          (http-metrics/wrap-with-request-metrics metrics)
+                          (comidi/wrap-with-route-metadata routes))
          product-name (or (get-in config [:product :name])
                           {:group-id    "puppetlabs"
                            :artifact-id "puppetserver"})
          checkin-interval-millis (* 1000 60 60 24) ; once per day
          update-server-url (get-in config [:product :update-server-url])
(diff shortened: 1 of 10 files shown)


src/clj/puppetlabs/services/master/master_service.clj
```diff
@@ -81,8 +95,15 @@
...
          (add-ring-handler this
                            ring-handler
-                           {:normalize-request-uri true}))))
-   context)
+                           {:normalize-request-uri true}))
+
+       (register-status
+        "master"
+        (status-core/get-artifact-version "puppetlabs" "puppetserver")
+        1
```
jpinsonault: Should the version number be put into a variable somewhere?

test/integration/puppetlabs/services/legacy_routes/legacy_routes_test.clj
```diff
@@ -25,39 +28,126 @@
...
+               (is (= 4 (:count (find-route http-metrics "total"))))
+               (is (= 1 (:count (find-route http-metrics "puppet-v3-environments"))))
+               (is (= 1 (:count (find-route http-metrics "puppet-v3-node-/*/"))))
+               (is (= 1 (:count (find-route http-metrics "puppet-v3-catalog-/*/"))))
+               (is (= 1 (:count (find-route http-metrics "other"))))))
+           (testing "The counts should be zero for endpoints that we didn't hit"
+             (is (every? #(= 0 %) (map :count
+                                       (filter
+                                        #(not (hit-routes (:route-id %)))
+                                        http-metrics)))))))))
```
rlinehan: So this is now like a 200 line test... it seems like it would be good to break it up a bit, and this seems like a reasonable point to do it. Have a test for the metrics in `master -> experimental -> http-metrics` and then have a test that these request show up in the `jruby-metrics` when they have borrowed a jruby.

test/integration/puppetlabs/services/master/master_service_test.clj
```diff
@@ -51,6 +53,77 @@
...
+                 (is (= 3 (:count (find-route http-metrics "total"))))
+                 (is (= 1 (:count (find-route http-metrics "puppet-v3-node-/*/"))))
+                 (is (= 1 (:count (find-route http-metrics "puppet-v3-catalog-/*/"))))
+                 (is (= 1 (:count (find-route http-metrics "other"))))))
+             (testing "The counts should be zero for endpoints that we didn't hit"
+               (is (every? #(= 0 %) (map :count
+                                         (filter
+                                          #(not (hit-routes (:route-id %)))
+                                          http-metrics)))))))))
+
```
rlinehan: Same comment here about this being a really long test and maybe it would be good to split it up.

src/clj/puppetlabs/services/master/master_service.clj
```diff
@@ -27,25 +27,48 @@
...
          localcacert (get-in config [:puppetserver :localcacert])
          puppet-version (get-in config [:puppetserver :puppet-version])
-         hostcrl (get-in config [:puppetserver :hostcrl])
          settings (ca/config->master-settings config)
-         product-name (or (get-in config [:product :name])
-                          {:group-id    "puppetlabs"
-                           :artifact-id "puppetserver"})
-         checkin-interval-millis (* 1000 60 60 24) ; once per day
-         update-server-url (get-in config [:product :update-server-url])
+         metrics-server-id (get-in config [:metrics :server-id])
```
rlinehan: This could use the metrics service `get-server-id` function now.

src/clj/puppetlabs/services/master/master_service.clj
```diff
@@ -53,36 +80,31 @@
...
+                          :normalize-request-uri true})
+       (add-ring-handler this
+                         ring-handler
+                         {:normalize-request-uri true}))
+
+     (register-status
+      "master"
+      (status-core/get-artifact-version "puppetlabs" "puppetserver")
+      1
+      (partial core/v1-status metrics))
```
rlinehan: Hm... what if rather than duplicating this in the master service in pe-p-s-e, we just registered another status here with `"pe-master"`. Could have registering it dependent on whether the pe-p-s-e artifact is being used. Something like:

```clojure
    (let [version (versioneer/get-version "puppetlabs" "pe-puppet-server-extensions")]
      (when-not (empty? version)
        (register-status
         "pe-master"
         version
         1
         (fn [level]
           (log/warn "The pe-master status is deprecated")
           (core/v1-status metrics-registry level)))))
```

(`status-core/get-artifact-version` throws an error if it doesn't get back a version, which is why I put `versioneer/get-version` there.)

camlow325: This is an interesting idea, thanks, @rlinehan.  I think, though, that I'd prefer to not embed any PE-specific/aware code in OSS puppetserver.  I'd like to be able to control how/if we deprecate (or eventually remove) the registration of status info for the pe-master (as well as pe-jruby-metrics and pe-puppet-profiler) services via code that sits in pe-puppet-server-extensions.  Having to make a change in OSS Puppet Server and a release/rollup of that change into pe-puppet-server-extensions in order to remove the registration of the "pe-master" service name from status info later wouldn't feel right to me.

Given that it sounds like we're probably going to need to keep providing status info under both the PE and OSS names of these services for some period of time to preserve compatibility, I'm leaning toward having pe-puppetserver bootstrap both OSS and PE versions of the services during the transition period.  The OSS versions of the services would do most of the real work.  The PE services would basically just do a `register-status` call with the appropriate `pe-*` service names for compatibility, reusing the actual metrics objects / functions from the related OSS service implementations.  Then when/if we want to remove the PE services references, we can just delete the service implementations (and bootstrap config references) entirely from the PE codebase - on whatever cadence would most make sense for PE.

Would that approach make sense to you?

camlow325: I like the idea of using a warn message to output a message to the log about the pe-* service name being deprecated.  Seems like we could do that no matter what way we go with this.

test/integration/puppetlabs/services/legacy_routes/legacy_routes_test.clj
camlow325: Yeah, I think that makes sense.  Doesn't look like the borrowed/requested tests depend upon any prior state (other than the webserver being started) so seems like that should work.  I'll see if the borrowed/requested tests might be able to easily use mock jrubies under the new server instance in order to not slow down the test much more.

src/clj/puppetlabs/services/master/master_service.clj
rlinehan: That approach sounds good to me, @camlow325.

One thought is that we may need to have the OSS master service know about some PE-specific metric names in order to add things like file sync metrics to the "default metrics allowed," since we have a restriction of only calling `initialize-registry-settings!` once per-domain. Another option would be to make a tk-service solely for managing that registry, and then having a different implementation of that service for PE? Maybe that's overkill though.

camlow325: Hmm, good point about `initialize-registry-settings!`.  I hadn't really thought that through.  It does feel a little weird to be updating info for PE-specific metrics, like the file-sync ones, from OSS Puppet Server.  I wonder if it might be better to come up with a way to distribute the list of registry settings across the services that manage them.  Like, maybe we could introduce an `add-to-or-initialize-registry-settings` function on the tk-metrics protocol which could be called multiple times, concatenating the stored list as it goes, before the registry is locked.  If we did that, we could let OSS puppetserver manage the Puppet profiler and http related defaults and maybe move the file-sync related ones over to pe-file-sync if that makes sense.  Maybe we could pick this conversation up further as we do [SERVER-1721](https://tickets.puppetlabs.com/browse/SERVER-1721)?

test/integration/puppetlabs/services/legacy_routes/legacy_routes_test.clj
```diff
@@ -201,19 +291,29 @@
...
-                 "are still being forwarded to the core CA functions."))
-
-        (is (not (= 200 (:status (http-get "/production/certificate_statuses/all"))))
-            (str "A 200 was returned from a request made to a legacy CA endpoint "
-                 "indicating the disabled CA service was not detected."))))))
+    ;; Startup server once with real CA service so that SSL certs/keys/etc.
+    ;; are generated properly before we start again with the disabled CA
+    ;; service bootstrapped.  When starting up with disabled CA service,
+    ;; the certs/keys/etc. are assumed to already be in place at startup.  The
+    ;; server would fail to start on being unable to find any of these files.
```
jpinsonault: This seems less than ideal, but I don't have a better idea. Also why does a disable CA service need cert files? Honest question, probably not worth getting into here though

camlow325: Any time you start a webserver with SSL enabled in the configuration, the webserver will need cert files.  When the "real" (non-disabled) CA is configured, it will create its own cert files for the webserver to use.  When the "disabled" CA is configured, it is assumed that you have already dropped the cert files that you want to use in place before starting up the server - so it doesn't need to create its own.

I could have also handled this by having the test create cert files without doing a full service startup or checking some pre-generated certs into the repo for the test to use.  I think we already have too many certs checked into the repo as it is though.  This seemed like a simple enough change and the performance overhead is probably pretty minimal overall.

src/clj/puppetlabs/services/master/master_service.clj
camlow325: Good idea.  Did this in f8d105b.

test/integration/puppetlabs/services/legacy_routes/legacy_routes_test.clj
camlow325: Did this in 2869ce2.

src/clj/puppetlabs/services/master/master_service.clj
```diff
@@ -53,36 +80,31 @@
...
+                         {:route-id :master-routes
+                          :normalize-request-uri true})
+       (add-ring-handler this
+                         ring-handler
+                         {:normalize-request-uri true}))
+
+     (register-status
+      "master"
+      (status-core/get-artifact-version "puppetlabs" "puppetserver")
+      1
```
camlow325: That makes sense to me.  Did this in f8d105b.

test/integration/puppetlabs/services/legacy_routes/legacy_routes_test.clj
jpinsonault: Ahh, ok, thanks for the explanation! Yeah, and this isn't a common pattern, so cool 👍

jpinsonault:
I'm 👍 
I guess I'll wait for you to look at the new commits if you want @rlinehan

PR #1335, opened 2017-03-07
(SERVER-1261) Import advanced puppet-profiler service from PE

camlow325:
This PR pulls in the advanced puppet-profiler service from PE Puppet Server.
This includes the registration of various profiled Puppet objects as metrics with
dropwizard and returning profiler info from the status endpoint.


src/clj/puppetlabs/services/puppet_profiler/puppet_profiler_core.clj
```diff
@@ -1,24 +1,132 @@
...
+  (let [level>= (partial status-core/compare-levels >= level)]
+    {:state :running
+     :status (cond->
+               ;; no status info at ':critical' level
+               {}
+               ;; no extra status at ':info' level yet
+               (level>= :info) identity
+               (level>= :debug) (assoc-metrics-data metrics-profiler))}))
+
+
```
jpinsonault: extra new line

camlow325: Thanks, @jpinsonault.  Where this was a small change, I just amended the commit where I brought it in at ff16d42.

src/clj/puppetlabs/services/puppet_profiler/puppet_profiler_core.clj
```diff
@@ -1,24 +1,131 @@
...
+(schema/defn ^:always-validate initialize :- MetricsProfilerServiceContext
+  [config :- MetricsProfilerServiceConfig
+   hostname :- schema/Str
+   registry :- (schema/maybe MetricRegistry)]
+  (let [enabled (ks/to-bool (:enabled config))]
+    (if-not enabled
+      {:profiler nil}
+      (if registry
+        {:profiler (metrics-profiler hostname registry)}
+        (do
```
rlinehan: Is it actually possible for this to be false? I don't think it is anymore... it used to be because you could disable metrics globally, but we removed that ability - at some point we started ignoring the `:enabled` setting in the `metrics` section (https://github.com/puppetlabs/trapperkeeper-metrics/blob/0.x/src/puppetlabs/trapperkeeper/services/metrics/metrics_core.clj#L69), and then we removed it all together for (what will be) tk-metrics 1.0.0 (https://github.com/puppetlabs/trapperkeeper-metrics/blob/master/src/clj/puppetlabs/trapperkeeper/services/metrics/metrics_core.clj#L65 - no longer allowed in the config). As far as I can tell, that setting was the only way you would not have a registry at this point.

test/integration/puppetlabs/puppetserver/testutils.clj
```diff
@@ -331,3 +331,30 @@
...
+      (doseq [f files] (fs/copy (fs/file src f) (fs/file target f))))))
+
+(defn remove-config-files
+  [dirs-to-copy]
+  (doseq [[src target] dirs-to-copy]
+    (let [[_ dirs files] (first (fs/iterate-dir src))]
+      (doseq [d dirs] (fs/delete-dir (fs/file target d)))
+      (doseq [f files] (fs/delete (fs/file target f))))))
+
+(defmacro with-puppet-config-files
```
rlinehan: So I guess is the difference between this and the `with-puppet-conf-files` macro is that this works for not just the confdir but also the codedir? Would be nice at some point to clean these helper functions/macros up because it feels like we have a lot of functions doing similar things. But probably outside the scope of this PR?

src/clj/puppetlabs/enterprise/services/puppet_profiler/puppet_profiler_core.clj
```diff
@@ -0,0 +1,132 @@
...
+                       :count schema/Int
+                       :mean schema/Num
+                       :aggregate schema/Num}]
+    :inline-metrics [{:metric schema/Str
+                      :count schema/Int}]}})
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Private
+
+(schema/defn metrics-profiler :- PuppetProfiler
```
adrienthebo: Newb question: since these functions are private can we use `defn-` to limit their accessibility outside of this namespace? (And would that work for functions with schemas?)

src/clj/puppetlabs/services/puppet_profiler/puppet_profiler_core.clj
```diff
@@ -1,24 +1,131 @@
...
+                                (sort-by :aggregate)
+                                reverse)
+          catalog-metrics (->> (map catalog-metric (.getCatalogTimers metrics-profiler))
+                               (sort-by :aggregate)
+                               reverse)
+          inline-metrics (->> (map inline-metric (.getInliningTimers metrics-profiler))
+                               (sort-by :count)
+                               reverse)]
+      (-> status
+          ;; TODO: decide whether 50 is a reasonable max here
```
camlow325: I'll take it out.  Suppose we'll find out over time what's most reasonable.

project.clj
```diff
@@ -48,6 +48,7 @@
                  [me.raynes/fs]
                  [liberator]
                  [org.apache.commons/commons-exec]
+                 [io.dropwizard.metrics/metrics-core "3.1.2"]
```
camlow325: Even though it's not in clj-parent currently, metrics-core probably should be added to clj-parent since it's being pinned to version 3.1.2 from multiple projects now, including tk-metrics, clj-http-client, jdbc-util, and (pe-)puppetserver.

We talked offline about this and agreed that it'd probably be easiest to just do that along with a bump to tk-metrics 1.0 once that is released.  I think @jpinsonault can take care of that as part of the work he'll be doing for [SERVER-1721](https://tickets.puppetlabs.com/browse/SERVER-1721).

test/integration/puppetlabs/puppetserver/testutils.clj
camlow325: @rlinehan I agree, the differerences in names here are confusing.

I think the biggest difference between the two is that `with-puppet-conf-files` allows you to provide a list of files (potentially from different source directories) that you want to have copied over to a target directory, like:

`{"puppet.conf" "/my/source/dir1",
   "puppetdb.conf" "/my/source/dir2"} "/my/target/dir"`

... whereas `with-puppet-config-files` (which I just lifted verbatim from PE Puppet Server along with the tests which used it) allows you to copy multiple source directories to corresponding target directories, like:

`{"/my/source/dir1" "/my/target/dir1",
   "/my/source/dir2" "/my/target/dir2"}`

I suppose both capabilities might be useful.  For the new tests being imported, the ability to have multiple target directories - confdir and codedir - is a useful thing that you can't get out of `with-puppet-conf-files`.

Would just a name change for now make it a little better - like calling the new function this PR would bring in `with-puppet-config-dirs` - or even just `with-config-dirs` since there isn't anything specifically "Puppet"-specific about them?

src/clj/puppetlabs/enterprise/services/puppet_profiler/puppet_profiler_core.clj
camlow325: @adrienthebo Good question.  I swear we had documented better at some point why we do this but am not finding any references to this now.  I believe the reasons we originally started doing this in the Puppet Server-related projects instead of using `defn-` or a `^:private` modifier are:

1) Makes it easier to reference the function directly from test code when you need to.  You can work around the access modifier in tests by using the [@#'some.ns/var](https://github.com/bbatsov/clojure-style-guide#access-private-var) modifier but that is a bit uglier.

2) We were enamored with the idea of using "pragma" blocks like the ";;; ... Private " above for readability, following the lead of [Prismatic](https://github.com/plumatic/eng-practices/blob/4cfaa9099841c7e041a1eb4f41e66fa460774386/clojure/20130927-ns-organization.md#separate-code-into-pragma-blocks).

In the above case, I was just copying in the code basically as-is from PE, pragma blocks and all.  You'll see this style for the most part in Puppet Server code although it certainly isn't uncommon to see the `defn-` style in other projects.  Mostly a question of style, I guess.

camlow325: oh, to answer your question about whether this would work with schema, I don't think there's a `schema/defn-` although I think you would get privacy with `schema/defn ^:private`.

src/clj/puppetlabs/services/puppet_profiler/puppet_profiler_core.clj
camlow325: Took this out in 00064de.

src/clj/puppetlabs/services/puppet_profiler/puppet_profiler_core.clj
camlow325: Removed the unreachable code path and related test in 00064de.

src/clj/puppetlabs/services/puppet_profiler/puppet_profiler_service.clj
```diff
@@ -1,13 +1,28 @@
...
-    (core/create-profiler (get-in-config [:profiler]))))
+                          profiler/PuppetProfilerService
+                          [[:MetricsService get-metrics-registry]
+                           [:ConfigService get-in-config]
+                           [:StatusService register-status]]
 
+  (init [this context]
+    (let [context (puppet-profiler-core/initialize
+                    (get-in-config [:profiler] {})
+                    (get-in-config [:metrics :server-id])
```
camlow325: Yeah, that makes sense.  I did that in 00064de.

test/integration/puppetlabs/puppetserver/testutils.clj
camlow325: I went ahead and changed the name of the function to `with-config-dirs` in e450ca4.  Happy to do something different if this doesn't seem to make sense.

camlow325:
@rlinehan I think I've hit all of your comments so far.  Let me know if you see something I missed.  Thanks!

test/integration/puppetlabs/puppetserver/testutils.clj
rlinehan: That sounds good to me - I think the new name is better. I still question exactly how useful it is to have both of these functions - mostly I think in the past we've had a tendency to create test helper functions all over the place, and it gets a little out of control. I just want to make sure we don't recreate functionality we already have. Anyway, I'm good with this as-is.

project.clj
rlinehan: Sounds good to me.

PR #1337, opened 2017-03-09
(SERVER-1743) Move check-for-updates call into analytics service

camlow325:
This commit moves the check-for-updates call to dujour out into a
separate analytics service.  This would hopefully allow for the OSS
master service to be reusable directly in PE Puppet Server once all of
the metrics-related work is ported down into OSS.

### DIFF
--- a/src/clj/puppetlabs/services/analytics/analytics_service.clj
+++ b/src/clj/puppetlabs/services/analytics/analytics_service.clj
@@ -26,5 +26,5 @@
                     (fn [] (version-check/check-for-updates!
                             {:product-name product-name} update-server-url)))
        (log/info (i18n/trs "Not checking for updates - opt-out setting exists"))))
-   (log/info (i18n/trs "Puppet Server Analytics has successfully started and will run in the background"))
+   (log/info (i18n/trs "Puppet Server Update Service has successfully started and will run in the background"))
    context))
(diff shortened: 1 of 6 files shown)


camlow325:
The `check-for-updates` tests previously failed in Travis CI on one cell because of timing issues, I think.  I reworked the timing in the tests a bit in 1b760de.  Hopefully the tests will pass this time.

briancain:
@camlow325 - Thanks for doing this! I can test this later today.

src/clj/puppetlabs/services/analytics/analytics_service.clj
```diff
@@ -0,0 +1,30 @@
...
+                           :artifact-id "puppetserver"})
+         checkin-interval-millis (* 1000 60 60 24) ; once per day
+         update-server-url (get-in config [:product :update-server-url])
+         check-for-updates (get-in config [:product :check-for-updates] true)]
+     (if check-for-updates
+       (interspaced checkin-interval-millis
+                    (fn [] (version-check/check-for-updates!
+                            {:product-name product-name} update-server-url)))
+       (log/info (i18n/trs "Not checking for updates - opt-out setting exists"))))
+   (log/info (i18n/trs "Puppet Server Analytics has successfully started and will run in the background"))
```
briancain: I wonder if it makes more sense right now to have this log say something more like "Puppet Server Update Service has..." since it is just checking for updates right now and does not send back any telemetry.

camlow325:
@briancain Sweet - thanks for reviewing before I pinged you about it.  😄

camlow325:
@pcarlisle - would appreciate it if you could skim over this too if you have time.  Thanks!

jpinsonault:
Didn't notice that there were two reviewers already but I'm also 👍

src/clj/puppetlabs/services/analytics/analytics_service.clj
camlow325: @briancain I updated this as you suggested in 1e6fe6a.

PR #1339, opened 2017-03-13
(SERVER-1721) Add allowed metrics to master service

jpinsonault:
This commits adds a list of default allowed metrics to puppetserver's master
service.

tk-metrics 1.0.0 added the ability for graphite exporting. The metrics service
can take a list of whitelisted metrics through a call to the
update-registry-settings service function during the init phase

Can't be merged until a new clj-parent release is made

### DIFF
--- a/ezbake/config/conf.d/metrics.conf
+++ b/ezbake/config/conf.d/metrics.conf
@@ -2,5 +2,36 @@
 metrics: {
     # a server id that will be used as part of the namespace for metrics produced
     # by this server
-    server-id: "localhost"
+    server-id: localhost
+    registries: {
+        puppetserver: {
+            # specify metrics to allow in addition to those in the default list
+            #metrics-allowed: ["compiler.compile.production"]
+
+            reporters: {
+                # enable or disable JMX metrics reporter
+                jmx: {
+                    enabled: true
+                }
+                # enable or disable Graphite metrics reporter
+                #graphite: {
+                #    enabled: true
+                #}
+            }
+
+        }
+    }
+
+    # this section is used to configure settings for reporters that will send
+    # the metrics to various destinations for external viewing
+    reporters: {
+        #graphite: {
+        #    # graphite host
+        #    host: "127.0.0.1"
+        #    # graphite metrics port
+        #    port: 2003
+        #    # how often to send metrics to graphite
+        #    update-interval-seconds: 5
+        #}
+    }
 }
--- a/project.clj
+++ b/project.clj
@@ -32,7 +32,7 @@
 
   :min-lein-version "2.7.1"
 
-  :parent-project {:coords [puppetlabs/clj-parent "0.5.0-SNAPSHOT"]
+  :parent-project {:coords [puppetlabs/clj-parent "0.5.0"]
                    :inherit [:managed-dependencies]}
 
   :dependencies [[org.clojure/clojure]
--- a/project.clj
+++ b/project.clj
@@ -32,7 +32,7 @@
 
   :min-lein-version "2.7.1"
 
-  :parent-project {:coords [puppetlabs/clj-parent "0.4.1"]
+  :parent-project {:coords [puppetlabs/clj-parent "0.5.0-SNAPSHOT"]
                    :inherit [:managed-dependencies]}
 
   :dependencies [[org.clojure/clojure]
--- a/project.clj
+++ b/project.clj
@@ -48,7 +48,7 @@
                  [me.raynes/fs]
                  [liberator]
                  [org.apache.commons/commons-exec]
-                 [io.dropwizard.metrics/metrics-core "3.1.2"]
+                 [io.dropwizard.metrics/metrics-core]
 
                  ;; We do not currently use this dependency directly, but
                  ;; we have documentation that shows how users can use it to
@@ -116,6 +116,7 @@
                                    [puppetlabs/trapperkeeper-webserver-jetty9]
                                    [puppetlabs/trapperkeeper-webserver-jetty9 nil :classifier "test"]
                                    [puppetlabs/trapperkeeper nil :classifier "test" :scope "test"]
+                                   [puppetlabs/trapperkeeper-metrics :classifier "test" :scope "test"]
                                    [puppetlabs/kitchensink nil :classifier "test" :scope "test"]
                                    [ring-basic-authentication]
                                    [ring-mock]
--- a/src/clj/puppetlabs/services/master/master_service.clj
+++ b/src/clj/puppetlabs/services/master/master_service.clj
@@ -15,15 +15,7 @@
 
 ;; Default list of allowed histograms/timers
 (def default-metrics-allowed-hists
-  ["compiler"
-   "compiler.compile"
-   "compiler.find_facts"
-   "compiler.find_node"
-   "compiler.static_compile"
-   "compiler.static_compile_inlining"
-   "compiler.static_compile_postprocessing"
-   "functions"
-   "http.active-histo"
+  ["http.active-histo"
    "http.puppet-v3-catalog-/*/-requests"
    "http.puppet-v3-environment-/*/-requests"
    "http.puppet-v3-environment_classes-/*/-requests"
@@ -34,19 +26,7 @@
    "http.puppet-v3-file_metadatas-/*/-requests"
    "http.puppet-v3-node-/*/-requests"
    "http.puppet-v3-report-/*/-requests"
-   "http.puppet-v3-static_file_content-/*/-requests"
-   "jruby.borrow-timer"
-   "jruby.free-jrubies-histo"
-   "jruby.lock-held-timer"
-   "jruby.lock-wait-timer"
-   "jruby.requested-jrubies-histo"
-   "jruby.wait-timer"
-   "puppetdb.catalog.save"
-   "puppetdb.command.submit"
-   "puppetdb.facts.find"
-   "puppetdb.facts.search"
-   "puppetdb.report.process"
-   "puppetdb.resource.search"])
+   "http.puppet-v3-static_file_content-/*/-requests"])
 
 ;; Default list of allowed values/counts
 (def default-metrics-allowed-vals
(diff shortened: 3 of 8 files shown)


jpinsonault:
ping @camlow325 @rlinehan for review

I think this list of metrics is appropriate for puppetserver

camlow325:
@jpinsonault I think we should pull some tests over from PE as well - or create new ones, where you think it may be appropriate - to validate the allowed metric registration.  This test in particular would seem good to port.

camlow325:
Now that we'll be pulling in a tk-metrics version which uses the new config format, I think it would be good to add in enabling the jmx reporter for the `puppetserver` registry by default, like we do for PE.  Maybe we could just change the metrics sections in the [ezbake metrics.conf](https://github.com/puppetlabs/puppetserver/blob/master/ezbake/config/conf.d/metrics.conf) and [dev puppetserver.conf](https://github.com/puppetlabs/puppetserver/blob/master/dev/puppetserver.conf#L115-L117) files to look more like [the ones we're using in PE](https://github.com/puppetlabs/pe-puppet-server-extensions/blob/2017.2.0.1/resources/ext/config/conf.d/metrics.conf)?  @rlinehan - WDYT?

project.clj
```diff
@@ -32,7 +32,7 @@
 
   :min-lein-version "2.7.1"
 
-  :parent-project {:coords [puppetlabs/clj-parent "0.4.1"]
+  :parent-project {:coords [puppetlabs/clj-parent "0.5.0-SNAPSHOT"]
```
camlow325: Seems like this may need to be rebased since the diff isn't showing the `io.dropwizard.metrics/metrics-core "3.1.2"` dependency that I see on the [current HEAD of the master branch](https://github.com/puppetlabs/puppetserver/blob/ff50c9cd40647ace5a94cc25499b1e20c7cddc62/project.clj#L51).  In any event, I think we should be able to remove the "3.1.2" and inherit the version from the forthcoming clj-parent release.

src/clj/puppetlabs/services/master/master_service.clj
```diff
@@ -13,12 +13,96 @@
 
 (def master-service-status-version 1)
 
+;; Default list of allowed histograms/timers
+(def default-metrics-allowed-hists
+  ["compiler"
```
camlow325: Arguably, it seems like we could move the `jruby.` ones over to the `jruby-metrics` service and the `compiler`, `function`, and `puppetdb` ones over to the `puppet-profiler` service.  I'd be inclined to go that way just to keep the lists a little closer to the code paths that register them but could understand the counter argument that it creates more places to go look for all of the defaults and that the list of services within OSS Puppet Server should for the most part be a black box.

src/clj/puppetlabs/services/master/master_service.clj
```diff
@@ -13,12 +13,96 @@
...
+   "http.puppet-v3-status-/*/-percentage"
+   "http.total-requests"
+   "jruby.borrow-count"
+   "jruby.borrow-retry-count"
+   "jruby.borrow-timeout-count"
+   "jruby.num-free-jrubies"
+   "jruby.num-jrubies"
+   "jruby.request-count"
+   "jruby.return-count"
+   "num-cpus"])
```
camlow325: Although I see [this one](https://github.com/puppetlabs/pe-puppet-server-extensions/blob/2017.2.0.1/src/clj/puppetlabs/enterprise/services/master/master_service.clj#L81) in the default list for PE Puppet Server as well, I haven't been able to find where this metric is actually registered.  Do you know?

project.clj
jpinsonault: Ah, yep

src/clj/puppetlabs/services/master/master_service.clj
```diff
@@ -13,12 +13,96 @@
...
+
+; List of allowed jvm gauges/values
+(def default-jvm-metrics-allowed
```
camlow325: Note that the actual code which registers these metrics wouldn't come into OSS Puppet Server code until we do [SERVER-1752](https://tickets.puppetlabs.com/browse/SERVER-1752).  May be best in terms of version history to have this block come in with that ticket but not a big deal either way.

src/clj/puppetlabs/services/master/master_service.clj
jpinsonault: I'm not sure, I can try to track it down though. Do you know @rlinehan?

src/clj/puppetlabs/services/master/master_service.clj
jpinsonault: Ah ok, I hadn't realized those metrics weren't ported yet. I can pull them out

src/clj/puppetlabs/services/master/master_service.clj
camlow325: Ah, think I found it.  Apparently it comes from trapperkeeper-comidi-metrics.  See [here](https://github.com/puppetlabs/trapperkeeper-comidi-metrics/blob/0.1.1/src/puppetlabs/metrics/http.clj#L117-L120).  Probably fine to keep that one here, then.  Might be good to add a comment here about where it comes from for future reference, though?

camlow325:
I did some testing with a Graphite server with a local checkout of this PR.  All seemed to be working fine from what I could tell.  I was able to see changes appear from a Grafana dashboard view as I did puppet agent runs against the server.

src/clj/puppetlabs/services/master/master_service.clj
jpinsonault: I'm on the fence about that. Both arguments seem reasonable to me. I don't disagree that they should be close to the code that registers them, so if you think that's preferable I'm happy to move them there

jpinsonault:
I pushed some commits,
* Fixup project.clj dependencies
* Fix a misbehaving test macro
* Port some graphite tests from pe-ps-e
* Move the default allowed metrics lists to services that make more sense

I'm not sure that more graphite testing at this layer would be helpful/economical, since we have SERVER-1264 to handle the higher level testing

I think the only unaddressed question is if we should enable jmx by default. @camlow325 any argument against doing that?

test/unit/puppetlabs/services/jruby/jruby_puppet_testutils.clj
```diff
@@ -378,3 +379,30 @@ create-mock-pool-instance :- JRubyInstance
   (add-mock-jruby-pool-manager-service
    jruby-service-and-dependencies
    config))
+
+(defn copy-config-files
```
camlow325: I added pretty much this same code in a recent PR.  That's now [here|https://github.com/puppetlabs/puppetserver/blob/09c359c6343af06e5b7e9b667e94f61d54b82ce5/test/integration/puppetlabs/puppetserver/testutils.clj#L335-L368].  Would be good to rebase this PR and use the `with-config-dirs` instead of `with-puppet-config-files` macro.

jpinsonault: oh nice, will do

jpinsonault:
Rebased

test/integration/puppetlabs/services/master/master_service_test.clj
```diff
@@ -281,3 +283,76 @@
...
+      (:graphite-reporter (get-puppetserver-registry-context app)))
+
+(deftest graphite-filtering-works
+  (testing "default filter works"
+    (let [reported-metrics-atom (atom {})]
+      (with-redefs [metrics-core/build-graphite-sender
+                    (fn [_ domain]
+                      (metrics-testutils/make-graphite-sender reported-metrics-atom domain))]
+        (logutils/with-test-logging
+          (testutils/with-config-dirs
```
camlow325: We may need not the extra `with-puppet-config-files` wrapper here and in the next test.  The `with-config-dirs` wrapper in the [namespace-wide test fixture](https://github.com/puppetlabs/puppetserver/blob/09c359c6343af06e5b7e9b667e94f61d54b82ce5/test/integration/puppetlabs/services/master/master_service_test.clj#L29-L34) would seem to setup the proper files for the tests to run.  At least, it worked for me locally - tests passed - without these extra wrappers.

test/unit/puppetlabs/services/jruby/jruby_puppet_testutils.clj
```diff
@@ -18,7 +18,8 @@
             [puppetlabs.trapperkeeper.services.scheduler.scheduler-service :as scheduler-service]
             [puppetlabs.trapperkeeper.services.status.status-service :as status-service]
             [puppetlabs.trapperkeeper.services.webrouting.webrouting-service :as webrouting-service]
-            [puppetlabs.trapperkeeper.services :as tk-services])
+            [puppetlabs.trapperkeeper.services :as tk-services]
+            [me.raynes.fs :as fs])
```
camlow325: Probably don't need to add this now?

camlow325:
> I think the only unaddressed question is if we should enable jmx by default. @camlow325 any argument against doing that?

I'd still be in favor of doing that.  Doesn't seem to be a downside to me and one less thing for users to have to setup in order to use the new metrics.  @rlinehan - WDYT?

test/integration/puppetlabs/services/master/master_service_test.clj
jpinsonault: Ha, I wish I'd noticed this fixture earlier, would have saved me some effort

jpinsonault:
Removed the extra with-conf-dirs calls and the ks import

Went ahead an included the updated metrics.conf optimistically, but can remove it if we decided not to go that route

test/unit/puppetlabs/services/jruby/jruby_puppet_testutils.clj
```diff
@@ -378,3 +378,4 @@ create-mock-pool-instance :- JRubyInstance
   (add-mock-jruby-pool-manager-service
    jruby-service-and-dependencies
    config))
+
```
camlow325: Could remove this line since it's irrelevant to this PR now, no big deal.

jpinsonault: my own fault for not running git diff before committing

ezbake/config/conf.d/metrics.conf
```diff
@@ -2,5 +2,36 @@
 metrics: {
     # a server id that will be used as part of the namespace for metrics produced
     # by this server
-    server-id: "localhost"
+    server-id: localhost
+    registries: {
```
camlow325: This looks fine to me.  Could we add the same to the [dev/puppetserver.conf](https://github.com/puppetlabs/puppetserver/blob/09c359c6343af06e5b7e9b667e94f61d54b82ce5/dev/puppetserver.conf#L115-L117) file for local development?

jpinsonault:
Added the new metrics settings to dev/puppetserver.conf

PR #1345, opened 2017-03-16
(SERVER-1752) Port memory metrics from PE

jpinsonault:
This commit ports the jvm memory metrics and related tests from
pe-puppet-server-extensions

These metrics are a little redundant with the jvm metrics in the status
service, but they are documented and so we're porting them over to maintain
backwards compatibility

### DIFF
--- a/test/integration/puppetlabs/services/master/master_service_test.clj
+++ b/test/integration/puppetlabs/services/master/master_service_test.clj
@@ -355,76 +355,75 @@
     (with-redefs [metrics-core/build-graphite-sender
                   (fn [_ domain]
                     (metrics-testutils/make-graphite-sender reported-metrics-atom domain))]
-      (logutils/with-test-logging
-       (bootstrap-testutils/with-puppetserver-running
-        app
-        graphite-enabled-config
-        (let [registry (:registry (get-puppetserver-registry-context app))
-              get-memory-map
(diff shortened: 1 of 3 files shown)


jpinsonault:
ping @camlow325 @rlinehan for review when you have a chance

test/integration/puppetlabs/services/master/master_service_test.clj
```diff
@@ -349,3 +349,82 @@
                     @reported-metrics-atom
                     :puppetserver
                     "puppetlabs.localhost.compiler.compile.production.mean")))))))))
+
+(deftest jvm-metrics-sent-to-graphite-test
+  (let [reported-metrics-atom (atom {})]
+    (with-redefs [metrics-core/build-graphite-sender
+                  (fn [_ domain]
+                    (metrics-testutils/make-graphite-sender reported-metrics-atom domain))]
+      (logutils/with-test-logging
```
camlow325: Do you know if the `with-test-logging` wrapper here is essential?  Would be nice to be able to run these tests without it if we can - to avoid hiding errors in the output that might be useful to see.  If it's going to be unnecessarily spammy if we take it out, though, I'm good with this.

jpinsonault:
Removed the with-test-logging wrapper, it didn't produce any spam without it

PR #1347, opened 2017-03-17
(SERVER-1264) Add beaker graphite metrics tests

camlow325:
This commit adds a beaker acceptance test, collect_default_metrics,
which sets up a Graphite server, configures Puppet Server to export
metrics to it, and confirms that the default set of metrics that the
server should export are actually exported.  This test is configured to
run on el-6 only because the underlying Puppet module used to setup the
Graphite server, cprice404/grafanadash, only supports el-6 at present.

I copied the bulk of this over from the Graphite beaker tests that we've
been running in PE Puppet Server.  See:

* collect_default_metrics test
* metrics helpers
* graphite setup

I left out the verification of the `puppetdb.*` metrics which was being
done in the PE version of the test.  I did this because, even though we
have a separate [beaker test which sets up puppetdb](https://github.com/puppetlabs/puppetserver/blob/2.7.2/acceptance/suites/tests/00_smoke/puppetdb_integration.rb), I didn't
want to require a specific execution order among the tests.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -1,4 +1,6 @@
 require 'beaker/dsl/install_utils'
+require 'net/http'
+require 'json'
 
 module PuppetServerExtensions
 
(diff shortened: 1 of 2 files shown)


acceptance/lib/helper.rb
```diff
@@ -413,7 +415,177 @@ def append_match_request(args)
...
+        url += '&target=' << metrics_target(master, m)
+      end
+    end
+    url += '&format=json'
+  end
+
+  # returns a hash of metrics results
+  def query_metrics(master_hostname, graphite_hostname, metrics, minutes=10)
+    sleep_between_queries = 5
+    seconds_remaining = minutes * 60
```
camlow325: Unfortunately, this is probably going to be a slow test to run in CI.  😢

Testing this with pooler VMs, I've seen it take up to 3 minutes for all of the metrics that Puppet Server exports to have been created / available to be queried back from the Graphite API, even with Puppet Server configured to report to Graphite on a 1 second interval.  I added the looping logic in here to wait up to 10 minutes for all of the metrics to be delivered - although hopefully it won't take that long typically.

camlow325: Ah, I think the bottleneck was coming from Graphite, not Puppet Server.  If I configure the `MAX_CACHE_SIZE` setting in `/opt/graphite/conf/carbon.conf` to "inf" instead of the default, "50", the whole metric set is created in under 2 seconds.  Will see how best to work that into the Graphite configuration for this test.

jpinsonault:
I'm 👍 after the graphite config stuff is worked out

acceptance/lib/helper.rb
camlow325: In 2a39f30, I added a step which configures the `MAX_CREATES_PER_MINUTE` setting to "inf(inite)".  This should shave around 3 minutes off the test run time.

acceptance/suites/tests/metrics/collect_default_metrics.rb
```diff
@@ -0,0 +1,108 @@
...
+teardown do
+  on master, "mv -f #{puppetserver_conf_file}.bak #{puppetserver_conf_file}"
+  on master, "mv #{metrics_conf_file}.bak #{metrics_conf_file}"
+  if !master.is_pe?
+    on master, "rm -f #{sitepp}"
+  end
+  bounce_service(master, puppetservice)
+end
+
+if !master.is_pe?
```
rlinehan: Curious - why are we not doing this if the master is a pe node?

camlow325: Yeah, good question.  Basically, I had cased this out for PE because it didn't seem to be necessary - because for PE, we should have some function metrics exercised as part of a typical PE agent run anyway - and because I didn't want to have the test dirty up the code dir if run when file-sync is turned on, potentially causing some downstream test failures as a result.  This isn't obvious, though, so I added a comment in 993f5ed to hopefully explain this better.

camlow325:
I noticed a problem when running this test in PE.  The conf file permissions weren't being preserved after restoration at the end of the test, which caused the pe-puppetserver service to not start back up properly.  Not sure exactly why this didn't happen when I ran the test for OSS.  In any event, though, I added another commit which preserves the permissions, 9d91450.

rlinehan:
👍

PR #1348, opened 2017-03-17
(SERVER-1758) Add environment_modules endpoint

briancain:
This commit adds new functionality to the environment_modules endpoint.
It now allows users to not specify a query param for the endpoint. If
one is not provided, puppetserver will return all modules for all
environments in a single request.

__TODO__:

- [x] update the integration tests for the new endpoint, hopefully testing that multiple modules are returned for multiple environments
- [x] update the docs with information on the new endpoint

### DIFF
--- a/documentation/puppet-api/v3/environment_modules.markdown
+++ b/documentation/puppet-api/v3/environment_modules.markdown
@@ -16,10 +16,55 @@ on Ruby Puppet masters, such as the [deprecated WEBrick Puppet master][]. It als
 the Ruby-based Puppet master's authorization methods and configuration. See the
 [Authorization section](#authorization) for more information.
 
-## `GET /puppet/v3/environment_modules?environment=:environment`
+## `GET /puppet/v3/environment_modules`
 
-Making a request with no query parameters is not supported and returns an HTTP 400 (Bad
-Request) response.
+### Supported HTTP Methods
+
+GET
+
+### Supported Formats
+
+JSON
+
+### Responses
+
+#### GET request with results
+
+```
+GET /puppet/v3/environment_modules
+
+HTTP/1.1 200 OK
+Content-Type: application/json
+
+[{
+    "modules": [
+        {
+            "name": "puppetlabs/ntp",
+            "version": "6.0.0"
+        },
+        {
+            "name": "puppetlabs/stdlib",
+            "version": "4.14.0"
+        }
+    ],
+    "name": "env"
+},
+{
+    "modules": [
+        {
+            "name": "puppetlabs/stdlib",
+            "version": "4.14.0"
+        },
+        {
+            "name": "puppetlabs/azure",
+            "version": "1.1.0"
+        }
+    ],
+    "name": "production"
+}]
+```
+
+## `GET /puppet/v3/environment_modules?environment=:environment`
 
 ### Supported HTTP Methods
 
@@ -107,10 +152,68 @@ HTTP/1.1 400 Bad Request
 The environment must be purely alphanumeric, not 'bog|us'
 ```
 
+### No metadata.json file
+
+If your modules do not have a [metadata.json](https://docs.puppet.com/puppet/latest/modules_metadata.html)
+file, puppetserver will not be able to determine the version of your module. In
+this case, puppetserver will return a null value for `version` in the response
+body.
+
 ### Schema
 
 An environment modules response body conforms to the
 [environment modules schema](./environment_modules.json).
+
+#### Validating your json
+
+If you have a response body that you'd like to validate against the
+[environment_modules.json](./environment_modules.json) schema, you can do so using the ruby library
+[json-schema](https://github.com/ruby-json-schema/json-schema).
+
+First, install the ruby gem to be used:
+
+```bash
+gem install json-schema
+```
+
+Next, given a json file, you can validate its schema.
+
+Here is a basic json file called _example.json_:
+
+```json
+{
+    "modules": [
+        {
+            "name": "puppetlabs/ntp",
+            "version": "6.0.0"
+        },
+        {
+            "name": "puppetlabs/stdlib",
+            "version": "4.16.0"
+        }
+    ],
+    "name": "production"
+}
+```
+
+Run this command from the root dir of the puppetserver project (or update the
+path to the json schema file in the command below):
+
+```bash
+ruby -rjson-schema -e "puts JSON::Validator.validate!('./documentation/puppet-api/v3/environment_modules.json','example.json')"
+```
+
+If the json is a valid schema, the command should output `true`. Otherwise, the
+library will print a schema validation error detailing which key or keys validate
+the schema.
+
+If you have a response that is the entire list of environment modules (i.e. the
+environment_modules endpoint), you will need to use this command to validate
+the json schema:
+
+```bash
+ruby -rjson-schema -e "puts JSON::Validator.validate!('./documentation/puppet-api/v3/environment_modules.json','all.json', :list=>true)"
+```
 
 ### Authorization
 
--- a/src/clj/puppetlabs/puppetserver/jruby_request.clj
+++ b/src/clj/puppetlabs/puppetserver/jruby_request.clj
@@ -57,18 +57,27 @@
 (defn wrap-with-environment-validation
   "Middleware function which validates the presence and syntactical content
   of an environment in a ring request.  If validation fails, a :bad-request
-  slingshot exception is thrown."
-  [handler]
-  (fn [request]
-    (let [environment (get-environment-from-request request)]
-      (cond
-        (nil? environment)
-        (ringutils/throw-bad-request!
-         (i18n/tru "An environment parameter must be specified"))
+  slingshot exception is thrown. If the optional-environment-param is
+  provided as true, it allows for the endpoint to have its query param
+  environment to be optional. However if the param is provided, it will
+  still validate that param for the given request."
+  ([handler]
+   (wrap-with-environment-validation handler false))
+  ([handler optional-environment-param]
+   (fn [request]
+     (let [environment (get-environment-from-request request)]
+       (cond
+         (and (true? optional-environment-param)
+              (nil? environment))
+         (handler request)
 
-        (not (nil? (schema/check ps-common/Environment environment)))
-        (ringutils/throw-bad-request!
-         (ps-common/environment-validation-error-msg environment))
+         (nil? environment)
+         (ringutils/throw-bad-request!
+           (i18n/tru "An environment parameter must be specified"))
 
-        :else
-        (handler request)))))
+         (not (nil? (schema/check ps-common/Environment environment)))
+         (ringutils/throw-bad-request!
+           (ps-common/environment-validation-error-msg environment))
+
+         :else
+         (handler request))))))
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -134,7 +134,7 @@
   "Schema for a given module that can be returned within the
   :modules key in EnvironmentModulesInfo."
   {:name schema/Str
-   :version schema/Str})
+   :version (schema/maybe schema/Str)})
 
 (def EnvironmentModulesInfo
   "Schema for the return payload for an environment_classes request."
@@ -359,11 +359,17 @@
   environment-module-response! :- ringutils/RingResponse
   "Process the environment module information, returning a ring response to be
   propagated back up to the caller of the environment_modules endpoint."
-  [info-from-jruby :- List
-   environment :- schema/Str]
-  (let [info-as-json (module-info-from-jruby->module-info-for-json
-                       info-from-jruby environment)]
-    (middleware-utils/json-response 200 info-as-json)))
+  ([info-from-jruby :- Map]
+   (let [all-info-as-json (map #(module-info-from-jruby->module-info-for-json
+                                  (val %)
+                                  (name (key %)))
+                               (sort-nested-info-maps info-from-jruby))]
+     (middleware-utils/json-response 200 all-info-as-json)))
+  ([info-from-jruby :- List
+    environment :- schema/Str]
+   (let [info-as-json (module-info-from-jruby->module-info-for-json
+                        info-from-jruby environment)]
+     (middleware-utils/json-response 200 info-as-json))))
 
 (schema/defn ^:always-validate
   environment-module-info-fn :- IFn
@@ -371,14 +377,18 @@
   request for environment_modules information."
   [jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)]
   (fn [request]
-    (let [environment (jruby-request/get-environment-from-request request)]
+    (if-let [environment (jruby-request/get-environment-from-request request)]
       (if-let [module-info
                (jruby-protocol/get-environment-module-info jruby-service
                                                            (:jruby-instance request)
                                                            environment)]
         (environment-module-response! module-info
                                       environment)
-        (rr/not-found (i18n/tru "Could not find environment ''{0}''" environment))))))
+        (rr/not-found (i18n/tru "Could not find environment ''{0}''" environment)))
+      (let [module-info
+            (jruby-protocol/get-all-environment-module-info jruby-service
+                                                            (:jruby-instance request))]
+        (environment-module-response! module-info)))))
 
 (schema/defn ^:always-validate
   wrap-with-etag-check :- IFn
@@ -422,7 +432,7 @@
   (->
    (environment-module-info-fn jruby-service)
    (jruby-request/wrap-with-jruby-instance jruby-service)
-   jruby-request/wrap-with-environment-validation
+   (jruby-request/wrap-with-environment-validation true)
    jruby-request/wrap-with-error-handling))
 
 (schema/defn ^:always-validate valid-static-file-path?
--- a/src/java/com/puppetlabs/puppetserver/JRubyPuppet.java
+++ b/src/java/com/puppetlabs/puppetserver/JRubyPuppet.java
@@ -18,6 +18,7 @@ import java.util.List;
 public interface JRubyPuppet {
     Map getClassInfoForEnvironment(String environment);
     List getModuleInfoForEnvironment(String environment);
+    Map getModuleInfoForAllEnvironments();
     JRubyPuppetResponse handleRequest(Map request);
     Object getSetting(String setting);
     String puppetVersion();
--- a/src/ruby/puppetserver-lib/puppet/server/master.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/master.rb
@@ -77,6 +77,15 @@ class Puppet::Server::Master
     end
   end
 
+  def getModuleInfoForAllEnvironments()
+    all_envs = @env_loader.list
+    all_mod_data = {}
+    all_envs.each { |env|
+      all_mod_data[env.name] = self.class.getModules(env)
+    }
+    all_mod_data
+  end
+
   def getSetting(setting)
     Puppet[setting]
   end
@@ -97,15 +106,11 @@ class Puppet::Server::Master
 
   def self.getModules(env)
     env.modules.collect do |mod|
-      module_data = {}
-
       # NOTE: If in the future we want to collect more
       #       Module settings, this should be more programatic
       #       rather than getting these settings one by one
-      module_data[:name] = mod.forge_name
-      module_data[:version] = mod.version
-
-      module_data
+      {:name    => mod.forge_name ||= mod.name,
+       :version => mod.version}
     end
   end
 
--- a/documentation/puppet-api/v3/environment_modules.json
+++ b/documentation/puppet-api/v3/environment_modules.json
@@ -16,7 +16,7 @@
           },
           "version": {
             "description": "The version of the puppet module",
-            "type": "string"
+            "type": ["string","null"]
           }
         },
         "required": ["name", "version"],
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
@@ -73,6 +73,10 @@
     [this jruby-instance env-name]
     (.getModuleInfoForEnvironment jruby-instance env-name))
 
+  (get-all-environment-module-info
+    [this jruby-instance]
+    (.getModuleInfoForAllEnvironments jruby-instance))
+
   (get-environment-class-info-tag
    [this env-name]
    (let [environment-class-info (:environment-class-info-tags
--- a/src/clj/puppetlabs/services/protocols/jruby_puppet.clj
+++ b/src/clj/puppetlabs/services/protocols/jruby_puppet.clj
@@ -54,6 +54,10 @@
     [this jruby-instance env-name]
     "Get module information for a specific environment")
 
+  (get-all-environment-module-info
+    [this jruby-instance]
+    "Get all module information for all environments")
+
   (flush-jruby-pool!
     [this]
     "Flush all the current JRuby instances and repopulate the pool.")
--- a/test/integration/puppetlabs/services/master/environment_modules_int_test.clj
+++ b/test/integration/puppetlabs/services/master/environment_modules_int_test.clj
@@ -30,18 +30,30 @@
                     (purge-env-dir)))))
 
 (defn get-env-modules
-  ([env-name]
-   (try
-       (http-client/get
-        (str "https://localhost:8140/puppet/v3/"
-             "environment_modules?"
-             "environment="
-             env-name)
-        (merge
-         testutils/ssl-request-options
-         {:as :text}))
-       (catch Exception e
-         (throw (Exception. "environment_modules http get failed" e))))))
+  [env-name]
+  (try
+    (http-client/get
+      (str "https://localhost:8140/puppet/v3/"
+           "environment_modules?"
+           "environment="
+           env-name)
+      (merge
+        testutils/ssl-request-options
+        {:as :text}))
+    (catch Exception e
+      (throw (Exception. "environment_modules http get failed" e)))))
+
+(defn get-all-env-modules
+  []
+  (try
+    (http-client/get
+      (str "https://localhost:8140/puppet/v3/"
+           "environment_modules")
+      (merge
+        testutils/ssl-request-options
+        {:as :text}))
+    (catch Exception e
+      (throw (Exception. "environment_modules http get failed" e)))))
 
 (defn response->module-info-map
   [response]
@@ -58,9 +70,22 @@
      (logutils/with-test-logging
        (let [foo-file (testutils/write-foo-pp-file
                         "class foo (String $foo_1 = \"is foo\"){}")
+             all-foo-file (testutils/write-pp-file
+                            "class foo (String $foo_1 = \"is foo\"){}"
+                            "foo"
+                            "woo"
+                            "hoo"
+                            "1.0.0")
              expected-response {"modules" [{"name" "foo", "version" "1.0.0"}]
                                 "name" "production"}
+             expected-all-response '({"modules"
+                                      [{"name" "foo" "version" "1.0.0"}]
+                                      "name" "hoo"}
+                                     {"modules"
+                                      [{"name" "foo" "version" "1.0.0"}]
+                                      "name" "production"})
              response (get-env-modules "production")
+             all-response (get-all-env-modules)
              emptyenv-response (get-env-modules "")
              notreal-response (get-env-modules "notreal")]
          (testing "a successful status code is returned"
@@ -80,4 +105,12 @@
                  (ks/pprint-to-string response))))
          (testing "the expected response body is returned"
            (is (= expected-response
-                  (response->module-info-map response)))))))))
+                  (response->module-info-map response))))
+         (testing "a successful status code is returned"
+           (is (= 200 (:status all-response))
+               (str
+                 "unexpected status code for response, response: "
+                 (ks/pprint-to-string response))))
+         (testing "the expected response body is returned"
+           (is (= expected-all-response
+                  (response->module-info-map all-response)))))))))


src/clj/puppetlabs/puppetserver/jruby_request.clj
```diff
@@ -62,6 +62,10 @@
   (fn [request]
     (let [environment (get-environment-from-request request)]
       (cond
+        (and (nil? environment)
+             (= "puppet-v3-environment_modules-/*/" (get-in request [:route-info :route-id])))
+        (handler request)
```
briancain: This is probably terrible, but I wasn't sure how to get around the endpoint wanting validation for an `environment` param without making separate wrapper functions for `/v3/environment_modules` and `/v3/environment_modules?environment=:environment`. Went about it this way initially just to make some forward progress...but I'll think about a better way to do this if there is one.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -356,26 +356,37 @@
...
-  [info-from-jruby :- List
-   environment :- schema/Str]
-  (let [info-as-json (module-info-from-jruby->module-info-for-json
-                       info-from-jruby environment)]
-    (middleware-utils/json-response 200 info-as-json)))
+  ([info-from-jruby :- List]
+   (let [return-data (map #(module-info-from-jruby->module-info-for-json
+                             (first (vals %))
+                             (name (first (keys %))))
+                          (sort-nested-info-maps info-from-jruby))]
```
briancain: This is also a tricky part of this change. This gives me want I want, but surely there's a better way to handle this data and use these formatting functions in a cleaner way...? Maybe not.

briancain: Also note to self: change `return-data` to `all-info-as-json` to be consistent.

briancain: And change the response from 404 to 200...

src/clj/puppetlabs/puppetserver/jruby_request.clj
camlow325: Might be better to have an overload of `wrap-with-environment-validation` which takes in an extra parameter that can be used to determine whether or not an environment is required.  That parameter could default to `false` to support other users of this endpoint for now but be set as `true` when called from the `environment-module-handler`.  The `cond` could then just thread through to the `handler` if no environment is provided, else throw the "environment parameter must be specified" error.  This way, we're not baking any special knowledge about endpoints into this implementation but allowing users of the function to vary the behavior, as desired.  WDYT?

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -356,26 +356,37 @@
...
                                                            environment)]
         (environment-module-response! module-info
                                       environment)
-        (rr/not-found (i18n/tru "Could not find environment ''{0}''" environment))))))
+        (rr/not-found (i18n/tru "Could not find environment ''{0}''" environment)))
+      (if-let [module-info
+               (jruby-protocol/get-all-environment-module-info jruby-service
+                                                           (:jruby-instance request))]
+        (environment-module-response! module-info)
+        (rr/not-found (i18n/tru "Could not find any puppet module data"))))))
```
camlow325: Would you expect this line to ever be hit?  If the expected behavior for when there are no modules that an empty hash be returned from the Ruby layer, I'd probably just avoid doing the extra `if-let` with "else" conditional here.

src/clj/puppetlabs/puppetserver/jruby_request.clj
briancain: Ah yeah, I like that idea. I definitely did not want to go with this approach since like you said it bakes in special knowledge in the wrapper function. I'll go ahead with that.

src/clj/puppetlabs/services/master/master_core.clj
briancain: That's a fair point....I'm assuming it'll at least return an empty array with the production environment so this probably won't ever happen.

src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -77,6 +77,16 @@ def getModuleInfoForEnvironment(env)
     end
   end
 
+  def getAllModuleInfoForEnvironment()
+    all_envs = @env_loader.list
+    all_envs.collect do |env|
```
camlow325: Seems like we'd end up with an array of hashes that each have one element in them, the environment with associated modules.  Would it be better maybe to just have a single hash with the environment (key) to module info (values) as each entry in that hash?

src/java/com/puppetlabs/puppetserver/JRubyPuppet.java
```diff
@@ -18,6 +18,7 @@
 public interface JRubyPuppet {
     Map getClassInfoForEnvironment(String environment);
     List getModuleInfoForEnvironment(String environment);
+    List getAllModuleInfoForEnvironment();
```
camlow325: Would a better name for this maybe be `getModuleInfoForAllEnvironments`?

src/ruby/puppetserver-lib/puppet/server/master.rb
camlow325: I wonder if you did it as a single hash if it might make the code in Clojure which walks back over the map a little simpler?

documentation/puppet-api/v3/environment_modules.markdown
```diff
@@ -16,10 +16,55 @@ on Ruby Puppet masters, such as the [deprecated WEBrick Puppet master][]. It als
...
+            "version": "6.0.0"
+        },
+        {
+            "name": "puppetlabs/stdlib",
+            "version": "4.14.0"
+        }
+    ],
+    "name": "env"
+},
+}
```
camlow325: This seems backwards?

documentation/puppet-api/v3/environment_modules.markdown
```diff
@@ -16,10 +16,55 @@ on Ruby Puppet masters, such as the [deprecated WEBrick Puppet master][]. It als
 the Ruby-based Puppet master's authorization methods and configuration. See the
 [Authorization section](#authorization) for more information.
 
-## `GET /puppet/v3/environment_modules?environment=:environment`
+## `GET /puppet/v3/environment_modules`
```
camlow325: I assume we'll also need to make some updates to the `environment_classes.json` schema to allow for the module info either to be for a single environment entry vs. a list of environment entries?

src/java/com/puppetlabs/puppetserver/JRubyPuppet.java
briancain: Ah yeah, reading what I have there now doesn't really make sense. I'll update it.

documentation/puppet-api/v3/environment_modules.markdown
briancain: Hmm...what do you mean?

documentation/puppet-api/v3/environment_modules.markdown
briancain: Ah yeah....I need to do that. Was going to save it in case the structure some how changed since it's a manual process to update that file.

documentation/puppet-api/v3/environment_modules.markdown
camlow325: Shouldn't this be a `{` - left curly brace instead of a right curly brace?

briancain: Oh yep, good catch!

src/ruby/puppetserver-lib/puppet/server/master.rb
briancain: That data structure does seem easier to work with...then hopefully the function I use to process it on the clojure side might not have to be so complicated.

documentation/puppet-api/v3/all_environment_modules.json
```diff
@@ -0,0 +1,8 @@
+{
+  "$schema":     "http://json-schema.org/draft-04/schema#",
+  "title":       "All Environment Modules",
+  "description": "An array of information about all the modules in all Puppet code environments",
+  "type":        "array",
+  "properties": ["Environment Modules"],
+  "additionalProperties": false
+}
```
briancain: Does something like this....make sense?

camlow325: Does this work if you were to run it through a tool that does schema validation?

src/clj/puppetlabs/puppetserver/jruby_request.clj
```diff
@@ -57,18 +57,28 @@
...
-        (ringutils/throw-bad-request!
-         (i18n/tru "An environment parameter must be specified"))
+  slingshot exception is thrown. If the optional-environment-param is
+  provided as true, it allows for the endpoint to have its query param
+  environment to be optional. However if the param is provided, it will
+  still validate that param for the given request."
+  ([handler optional-environment-param]
+   (fn [request]
+     (let [environment (get-environment-from-request request)]
+       (if (and optional-environment-param
```
camlow325: If `optional-environment-param` is specified as `false` and the environment is `nil`, won't this `if` statement evaluate to false and pass the request through to the handler.  That doesn't seem right.

One alternative here might be to just have the 1 argument version have no other logic in it other than to call through to the 2 argument version, which would now have the `cond` in it.  And the first `cond` statement, on detecting a `nil` environment, could throw the error vs. threading through to the handler based on the value of the `optional-environment-param`.  WDYT?

camlow325:
This will need to be rebased now, it seems.  Sorry.  😞

documentation/puppet-api/v3/all_environment_modules.json
camlow325: If not, then, I guess I'd either prefer to see a single JSON schema which represents both the single and array formats - reusing the single item format to avoid duplication - or two separate full docs which have the duplication.  Either way, I think it will be most helpful to people to have the schema(e) capable of validating the different resource content which can come from the endpoint.

documentation/puppet-api/v3/all_environment_modules.json
```diff
@@ -0,0 +1,10 @@
+{
+  "$schema":     "http://json-schema.org/draft-04/schema#",
+  "title":       "All Environment Modules",
+  "description": "An array of information about all the modules in all Puppet code environments",
+  "type":        "array",
+  "properties": {
+    "$ref": "#/environment_modules"
+  },
+  "additionalProperties": false
+}
```
briancain: I think that comment got deleted when I rebased...but I think this passes the validator according to http://www.jsonschemavalidator.net/ using the draft 04 validation.

src/clj/puppetlabs/puppetserver/jruby_request.clj
briancain: Hmm, yeah that looks like a problem to me too. I think your suggestion makes sense. I'll go head and update it to work that way.

camlow325:
Doing some functional testing, I ran into an issue which may be pre-existing.

I have a few modules installed which don't have a `metadata.json` file in them - just some one-off code I'd written locally.  For that case, I'm getting a schema error, probably due to the name and version being computed incorrectly for the module.  For that case, I think it might be better to use the `name` instead of the `forge_name` for the module and some placeholder, e.g., <empty string>, for the `version` number.  Some example output below.  WDYT?

curl request for one environment:

````
curl -k -X GET --ipv4 "https://localhost:8140/puppet/v3/environment_modules?environment=production"
<html>
<head>
<meta http-equiv="Content-Type" content="text/html;charset=ISO-8859-1"/>
<title>Error 500 </title>
</head>
<body>
<h2>HTTP ERROR: 500</h2>
<p>Problem accessing /puppet/v3/environment_modules. Reason:
<pre>    Output of module-info-from-jruby-&gt;module-info-for-json does not match schema: {:modules [nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil {:name (not (instance? java.lang.String nil)), :version (not (instance? java.lang.String nil))} {:name (not (instance? java.lang.String nil)), :version (not (instance? java.lang.String nil))} nil nil nil nil nil nil nil nil nil nil nil]}</pre></p>
<hr /><i><small>Powered by Jetty://</small></i>
</body>
</html>
````

Stack trace:

````
2017-03-23 15:06:36,766 WARN  [qtp2058058362-57] [o.e.j.s.HttpChannel] /puppet/v3/environment_modules?environment=production
clojure.lang.ExceptionInfo: Output of module-info-from-jruby->module-info-for-json does not match schema: {:modules [nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil {:name (not (instance? java.lang.String nil)), :version (not (instance? java.lang.String nil))} {:name (not (instance? java.lang.String nil)), :version (not (instance? java.lang.String nil))} nil nil nil nil nil nil nil nil nil nil nil]}
	at puppetlabs.services.master.master_core$eval31346$module_info_from_jruby__GT_module_info_for_json__31351.invoke(master_core.clj:346)
	at puppetlabs.services.master.master_core$eval31375$environment_module_response_BANG___31384$fn__31389.invoke(master_core.clj:370)
	at puppetlabs.services.master.master_core$eval31375$environment_module_response_BANG___31384.invoke(master_core.clj:358)
	at puppetlabs.services.master.master_core$eval31413$environment_module_info_fn__31420$fn__31421$fn__31422.invoke(master_core.clj:385)
	at puppetlabs.puppetserver.jruby_request$wrap_with_jruby_instance$fn__22267.invoke(jruby_request.clj:46)
	at puppetlabs.puppetserver.jruby_request$wrap_with_environment_validation$fn__22279.invoke(jruby_request.clj:84)
	at compojure.response$eval24114$fn__24115.invoke(response.clj:33)
	at compojure.response$eval24069$fn__24070$G__24060__24077.invoke(response.clj:6)
	at puppetlabs.services.master.master_core$eval31630$v3_clojure_routes__31637$fn__31638$fn__31641.invoke(master_core.clj:545)
````

curl request for all environments:

````
curl -k -X GET --ipv4 "https://localhost:8140/puppet/v3/environment_modules"
<html>
<head>
<meta http-equiv="Content-Type" content="text/html;charset=ISO-8859-1"/>
<title>Error 500 </title>
</head>
<body>
<h2>HTTP ERROR: 500</h2>
<p>Problem accessing /puppet/v3/environment_modules. Reason:
<pre>    Output of module-info-from-jruby-&gt;module-info-for-json does not match schema: {:modules [{:name (not (instance? java.lang.String nil)), :version (not (instance? java.lang.String nil))}]}</pre></p>
<hr /><i><small>Powered by Jetty://</small></i>
</body>
</html>
````

Stack trace:

````
2017-03-23 15:08:05,953 WARN  [qtp2058058362-56] [o.e.j.s.HttpChannel] /puppet/v3/environment_modules
clojure.lang.ExceptionInfo: Output of module-info-from-jruby->module-info-for-json does not match schema: {:modules [{:name (not (instance? java.lang.String nil)), :version (not (instance? java.lang.String nil))}]}
	at puppetlabs.services.master.master_core$eval31346$module_info_from_jruby__GT_module_info_for_json__31351.invoke(master_core.clj:346)
	at puppetlabs.services.master.master_core$eval31375$environment_module_response_BANG___31384$fn__31385$fn__31386.invoke(master_core.clj:363)
````

**Comment by briancain** [comment]
Yeah...this won't work if the module(s) doesn't have a metadata.json file. I believe it's required now for modules on the forge. Perhaps if one does not exist (i.e. custom one-off modules), we fail over to just the name? The only downside I think is `name` isn't namespaced like `forge_name` is...so I'm not sure how to represent those in the response body without duplicate key entries.

File: documentation/puppet-api/v3/all_environment_modules.json (hunk 10 above)
Comment (camlow325) [inline documentation/puppet-api/v3/all_environment_modules.json]: It isn't clear to me from the spec that "#/environment_modules" is something that a tool would be expected to resolve in an external file.  Following [this example](http://json-schema.org/latest/json-schema-core.html#rfc.section.7), I had been thinking that something more like this might work:

```json
{
  "$schema":     "http://json-schema.org/draft-04/schema#",
  "title":       "All Environment Modules",
  "description": "An array of information about all the modules in all Puppet code environments",
  "type":        "array",
  "definitions": {
    "environment": { "id": "environment_modules.json"}
  },
  "items": {
    "$ref": "#/definitions/environment"
  },
  "additionalProperties": false
}
```
I was hoping to be able to successfully use this with a schema validator tool but have not so far.  I chose to use the Ruby [json-schema library](https://github.com/ruby-json-schema/json-schema).  In my case, I've been running this command:

```
ruby -rjson-schema -e "puts JSON::Validator.validate!('./all_environment_modules.json','response.json')"
```

... where the first arg is the base schema to validate and the second arg is the JSON file to be validated against the schema.  I expect the printed result to be "true" if schema validation succeeds, else an error if validation failed.

For an "all environments" payload from the endpoint, though, the above schema - as well as the one you currently have on the PR - will return "true" even if content on one or more of the array elements is invalid.  I haven't been able to get the tool to appear to successfully navigate into the "environment_modules.json" as part of the validation.

I have, however, been able to create a single schema that appears to properly validate both the single element and array form of the environment_modules payload:

```json
{
  "$schema":     "http://json-schema.org/draft-04/schema#",
  "title":       "Environment Modules",
  "description": "Information about the modules in a Puppet code environment",
  "oneOf": [
    { "$ref": "#/definitions/environment" },
    { "type": "array",
      "items": { "$ref": "#/definitions/environment"}
    }
  ],
  "definitions": {
    "environment": {
      "type": "object",
      "properties": {
        "modules": {
          "description": "The array of modules which exist in an environment.",
          "type": "array",
          "items": {
            "type": "object",
            "properties": {
              "name": {
                "description": "The name of the puppet module",
                "type": "string"
              },
              "version": {
                "description": "The version of the puppet module",
                "type": "string"
              }
            },
            "required": ["name", "version"],
            "additionalProperties": false
          }
        },
        "name": {
          "description": "Name of the environment",
          "type": "string"
        }
      },
      "required": ["modules", "name"],
      "additionalProperties": false
    }
  }
}
```

While this appeared to work fine, the error I get when schema validation fails is pretty generic.  In my case, to force a schema validation failure, I changed the data for one of the module names from a string to a boolean, true.  Here's the error I saw when validating with the "oneOf" schema:

```
.../attribute.rb:18:in `validation_error': The property '#/' of type Array did not match any of the required schemas (JSON::Schema::ValidationError)
```

Another option which worked for me was to just use the original single environment schema, environment_modules.json, but pass an option into the validator indicating that the source JSON has a "list" of the items represented in the schema:

```
ruby -rjson-schema -e "puts JSON::Validator.validate!('./documentation/puppet-api/v3/environment_modules.json','resource.json', :list=>true)"
```

The validation error for the same bad JSON as in my example above was much more useful in this case:

```
attribute.rb:18:in `validation_error': The property '#/0/name' of type TrueClass did not match the following type: string (JSON::Schema::ValidationError)
```

I'm not sure how prevalent a "list" capability is various schema validators.

---

Backing up a level, there doesn't seem to be an option here which is clearly best in my mind.  I'd probably lean at this point toward just leaving the original JSON schema doc, which has just the single environment representation in it, but just add a comment to the [Schema section](https://github.com/puppetlabs/puppetserver/blob/46c32ecfd5ba80a1342e9ed1728677e3d4aff1c7/documentation/puppet-api/v3/environment_modules.markdown#schema) of the "environment_modules.markdown doc which indicates that the linked schema applies to the response for a single environment but that the format would include an array of environments when no query parameter is specified.  Could also give an example of how to do the list form with the Ruby json-schema library, maybe?  I think I'd be good leaving it there and adding to / modifying this later based on any user feedback we might get.  WDYT?

briancain: @camlow325  - Ahhh ok, yeah in this case it probably makes sense do go ahead with what you listed at the bottom of your comment. I think it makes sense to just leave a note there as well as mention the ruby library to validate the schema. I'll go ahead and update the markdown file and delete the all environment json schema.

camlow325:
@briancain Yeah, wrt the duplicate key entries, seems like that could also happen if someone had duplicate names in the "metadata.json" files for different modules.  I suppose the code would currently just end up picking one of the name entries during the coercion to a `Hash`.  I think I'm good with that behavior for now since it seems like it wouldn't be very likely to happen.  I think it's okay for the modules that don't have a metadata.json to have a non-namespaced name in the result as well.  I'd mostly just be concerned about the endpoint producing a garbage response when one or more modules don't have a "metadata.json" file and am open to however you think might be best to head off that problem.

briancain:
@camlow325  - Ok, I think I've responded to all of the comments. The only thing I think I might want to do is put a little note on the documentation about what happens when modules don't have a metadata.json. I'm not sure where that might go, perhaps near the bottom? Or maybe it makes sense to call it out on the `Responses` header for each endpoint.

briancain:
Pushed up a commit leaving a note about what happens to the response body if no metadata.json file exists in modules.

src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -102,8 +111,14 @@ def self.getModules(env)
       # NOTE: If in the future we want to collect more
       #       Module settings, this should be more programatic
       #       rather than getting these settings one by one
-      module_data[:name] = mod.forge_name
-      module_data[:version] = mod.version
+      if mod.forge_name
+        module_data[:name] = mod.forge_name
+        module_data[:version] = mod.version
```
camlow325: If someone were to, probably accidentally I guess, have a "metadata.json" file in a module but leave the `version` field out of it, would this end up being assigned / returned as `nil` - and potentially fail schema validation somewhere in the Clojure code?  If so, maybe it would be better here to have the logic to assign a default value for the field based on a `nil` value for the field in the Ruby object be done separately for `name` vs. `version`.  WDYT?

camlow325:
👍 with just the one minor question / comment about what might happen if the `version` field is missing from a "metadata.json" file for a module.

src/ruby/puppetserver-lib/puppet/server/master.rb
briancain: It looks like what happens is if a user leaves off the version key in the metadata.json file, puppet will throw a `MissingMetadata` ruby exception from puppets module loader.

```
2017-03-27 13:28:29,217 ERROR [qtp1400259539-33] [puppetserver] Puppet No version module metadata provided for ntp
... ruby stacktrace ...
```

And just leaves that module out of the response.

I'm trying to write some code to handle this, but it doesn't seem like ruby here can catch the exception that gets thrown. At first I tried catching `Puppet::Module::MissingMetadata` since that's what gets thrown, but then I simplified it to just be an `Exception` when I try to access `mod.version` but the same thing happens. Version itself is an instance variable...so I can't do the mod.version.nil? check because even if it's left out of the metadata.json it looks like it gets initialized.

[It looks like](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/module.rb#L62-L71) at the very least we can rely on name always being a variable we can have per module so this could read something like:

```ruby
```

briancain: @camlow325 - yeah, so far as I can tell if one of [these](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/module.rb#L165) keys are missing from the metadata.json file, you can't operate at all on the modules object. So the entire iteration gets skipped and we can't fail over to a sensible default like empty string. It's definetly possible that someone could have a metadata.json file and have some of those keys missing, but I'm not sure how likely it is or what else will also be broken if they try to use those modules.

camlow325: Thanks, @briancain.  Given that in this scenario we'd probably see failures outside of the use of this API anyway, I'm inclined not to worry about this part further for now, then.  Thanks for checking into it!

briancain: Yeah I'm guessing a lot breaks because of this. I just tried using the module tool to uninstall those modules I was using to test this PR with, and it gave me the same exception/stacktrace:

```
brian@localhorse:ntp % sudo puppet module uninstall puppetlabs-ntp;sudo puppet module uninstall puppetlabs-stdlib
Password:
Error: No version module metadata provided for ntp
Notice: Preparing to uninstall 'puppetlabs-ntp' ...
Error: No version module metadata provided for ntp
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/module.rb:160:in `block in load_metadata'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/module.rb:158:in `each'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/module.rb:158:in `load_metadata'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/module.rb:69:in `initialize'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/node/environment.rb:358:in `new'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/node/environment.rb:358:in `block (3 levels) in modules_by_path'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/node/environment.rb:357:in `map'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/node/environment.rb:357:in `block (2 levels) in modules_by_path'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/node/environment.rb:353:in `chdir'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/node/environment.rb:353:in `block in modules_by_path'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/node/environment.rb:351:in `each'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/node/environment.rb:351:in `modules_by_path'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/module_tool/applications/uninstaller.rb:51:in `find_installed_module'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/module_tool/applications/uninstaller.rb:24:in `run'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/module_tool/applications/application.rb:12:in `run'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/face/module/uninstall.rb:63:in `block (3 levels) in <top (required)>'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/interface/action.rb+eval[wrapper]:242:in `uninstall'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/application/face_base.rb:248:in `main'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/application.rb:352:in `run_command'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/application.rb:344:in `block in run'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/util.rb:540:in `exit_on_fail'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/application.rb:344:in `run'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/util/command_line.rb:132:in `run'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/util/command_line.rb:72:in `execute'
/usr/local/bin/puppet:5:in `<main>'
Error: No version module metadata provided for ntp
Notice: Preparing to uninstall 'puppetlabs-stdlib' ...
Error: No version module metadata provided for ntp
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/module.rb:160:in `block in load_metadata'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/module.rb:158:in `each'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/module.rb:158:in `load_metadata'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/module.rb:69:in `initialize'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/node/environment.rb:358:in `new'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/node/environment.rb:358:in `block (3 levels) in modules_by_path'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/node/environment.rb:357:in `map'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/node/environment.rb:357:in `block (2 levels) in modules_by_path'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/node/environment.rb:353:in `chdir'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/node/environment.rb:353:in `block in modules_by_path'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/node/environment.rb:351:in `each'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/node/environment.rb:351:in `modules_by_path'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/module_tool/applications/uninstaller.rb:51:in `find_installed_module'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/module_tool/applications/uninstaller.rb:24:in `run'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/module_tool/applications/application.rb:12:in `run'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/face/module/uninstall.rb:63:in `block (3 levels) in <top (required)>'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/interface/action.rb+eval[wrapper]:242:in `uninstall'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/application/face_base.rb:248:in `main'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/application.rb:352:in `run_command'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/application.rb:344:in `block in run'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/util.rb:540:in `exit_on_fail'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/application.rb:344:in `run'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/util/command_line.rb:132:in `run'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/util/command_line.rb:72:in `execute'
/usr/local/bin/puppet:5:in `<main>'
```

src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -102,8 +111,14 @@ def self.getModules(env)
...
+      else
+        # no metadata.json file, so all we have is `name`
+        module_data[:name] = mod.name
+        module_data[:version] = nil
```
camlow325: Not a big deal but could do something shorter like the following to avoid having any conditional logic for the `:version` field and still get the behavior of a `nil` being returned when no version can be derived from a metadata.json file for the module:

```ruby
  def self.getModules(env)
    env.modules.collect do |mod|
      {:name => mod.forge_name ||= mod.name,
       :version => mod.version}
    end
  end
```

That said, though, I'm good with what you have already if you want to go with that instead.

briancain: I was actually thinking of cleaning up the code here, so I'll go ahead and do that!

camlow325:
@gguillotte Since this has some docs changes in it, you may want to take a look at this - or pass it over to whomever might be picking that up from you if you know who that is.

@briancain Should someone else from your team also do a review of this - like @pcarlisle?

documentation/puppet-api/v3/environment_modules.markdown
```diff
@@ -107,11 +152,69 @@ HTTP/1.1 400 Bad Request
 The environment must be purely alphanumeric, not 'bog|us'
 ```
 
+### No metadata.json file
+
+If your modules do not have a [metadata.json](https://docs.puppet.com/puppet/4.9/modules_metadata.html)
```
gguillotte: This is fine for now, just remember to bump this to a newer version of the Puppet docs should a newer version of Puppet come out before this officially ships in Server.

documentation/puppet-api/v3/environment_modules.markdown
```diff
@@ -107,11 +152,69 @@ HTTP/1.1 400 Bad Request
...
+
 ### Schema
 
 An environment modules response body conforms to the
 [environment modules schema](./environment_modules.json).
 
+#### Validating your json
+
+If you have a response body that you'd like to validate against the
+environment_modules.json schema, you can do so using the ruby library
```
gguillotte: Can we link to the schema here? Where is it?

gguillotte:
One docs suggestion but no blockers. 👍 from me.

documentation/puppet-api/v3/environment_modules.markdown
briancain: Is there a "latest" URL that would work better?

briancain: Ah looks like there is. I'll just update it with that instead.

gguillotte: The `/latest/` URL should be fine as long as Puppet and Server are released together.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -355,30 +355,54 @@
...
-   environment :- schema/Str]
-  (let [info-as-json (module-info-from-jruby->module-info-for-json
-                       info-from-jruby environment)]
-    (middleware-utils/json-response 200 info-as-json)))
+  ([info-from-jruby :- Map]
+   (let [all-info-as-json (map #(module-info-from-jruby->module-info-for-json
+                                  (val %)
+                                  (name (key %)))
+                               (sort-nested-info-maps info-from-jruby))]
+     (println (extract-module-data all-info-as-json))
```
camlow325: Since this just appears to be for debugging, I think we should remove this line - and the corresponding `extract-module-data` function since nothing else appears to use it - before merging.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -355,30 +355,54 @@
        vec
        (sorted-map :name environment :modules)))
 
+(def ModuleList
```
camlow325: Seems like this should be removed since nothing is referencing it.

briancain: Yep apologizes....I had some extra debugging functions that I accidentally committed. They should be removed now!

camlow325:
Now that we're going to allow for the `version` field in a module entry to be `null`, I think we should update the "environment_modules.json" schema to allow for that.  If I change [this line](https://github.com/puppetlabs/puppetserver/blob/0d001b2b154544aa451e7feae21d3d72543bd75a/documentation/puppet-api/v3/environment_modules.json#L19) to:

```json
"type": ["string", "null"]
```

... the behavior seems to be what we want.

briancain:
@camlow325 - updated with the schema changes!

camlow325:
:shipit: 🚀

PR #1349, opened 2017-03-17
(maint) Test puppet 3 compat against Puppet 4/5

adrienthebo:
/cc @rlinehan


rlinehan:
travis isn't gonna do anything for it, so 🚢 .

PR #1350, opened 2017-03-21
(SERVER-1737) Add http client metrics to status

rlinehan:
Pull out specific http client metrics and add them to the master service
status endpoint under an `http-client-metrics` key.

Since other services (specifically, the pe-master service) may have http
client metrics that they would like to add to this list to be included in the
status, the metric ids to be included in the status are registered in an atom
in the service context, and there is a new service function
`add-metric-ids-to-http-client-metrics-list!` that will append metric ids to
this list.

Add these metrics to the list of `default-metrics-allowed` that will be exported to graphite.

### DIFF
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -464,11 +464,13 @@
         {:status 200
          :body (get-code-content environment code-id file-path)}))))
 
+(def MetricIdsForStatus (schema/atom [[schema/Str]]))
+
 (schema/defn http-client-metrics-summary
   :- {:metrics-data [http-client-common/MetricIdMetricData]
       :sorted-metrics-data [http-client-common/MetricIdMetricData]}
   [metric-registry :- MetricRegistry
-   metric-ids-to-select :- (schema/atom [[schema/Str]])]
+   metric-ids-to-select :- MetricIdsForStatus]
   (let [metrics-data (map #(http-client-metrics/get-client-metrics-data-by-metric-id metric-registry %)
                           @metric-ids-to-select)
         flattened-data (flatten metrics-data)]
@@ -757,13 +759,13 @@
    ["puppetdb" "resource" "search"]])
 
 (schema/defn ^:always-validate add-metric-ids-to-http-client-metrics-list!
-  [metric-id-atom :- (schema/atom [[schema/Str]])
+  [metric-id-atom :- MetricIdsForStatus
    metric-ids-to-add :- [[schema/Str]]]
   (swap! metric-id-atom concat metric-ids-to-add))
 
 (schema/defn ^:always-validate v1-status :- status-core/StatusCallbackResponse
   [http-metrics :- http-metrics/HttpMetrics
-   http-client-metric-ids-for-status :- (schema/atom [[schema/Str]])
+   http-client-metric-ids-for-status :- MetricIdsForStatus
    metric-registry :- MetricRegistry
    level :- status-core/ServiceStatusDetailLevel]
   (let [level>= (partial status-core/compare-levels >= level)]
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -3,7 +3,6 @@
            (clojure.lang IFn)
            (java.util List Map Map$Entry)
            (org.jruby RubySymbol)
-           (org.eclipse.jetty.util URIUtil)
            (com.codahale.metrics MetricRegistry Gauge)
            (java.lang.management ManagementFactory))
   (:require [me.raynes.fs :as fs]
@@ -24,6 +23,8 @@
             [puppetlabs.i18n.core :as i18n]
             [puppetlabs.metrics :as metrics]
             [puppetlabs.metrics.http :as http-metrics]
+            [puppetlabs.http.client.metrics :as http-client-metrics]
+            [puppetlabs.http.client.common :as http-client-common]
             [puppetlabs.trapperkeeper.services.status.status-core :as status-core]))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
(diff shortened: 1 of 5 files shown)


rlinehan:
The ticket mentions adding these metrics to the developer dashboard. I wasn't sure how useful that would actually be (and also didn't really want to mess around with box sizes and such), so I didn't end up doing this. What do other folks think about adding this to the dashboard or not?

test/integration/puppetlabs/services/master/master_service_test.clj
```diff
@@ -48,6 +53,22 @@
...
+    (http-client/put (str "https://localhost:8140" url)
+                     {:ssl-cert (str master-service-test-runtime-ssl-dir
+                                     "/certs/localhost.pem")
+                      :ssl-key (str master-service-test-runtime-ssl-dir
+                                    "/private_keys/localhost.pem")
+                      :ssl-ca-cert (str master-service-test-runtime-ssl-dir
+                                        "/certs/ca.pem")
+                      :body body
+                      :headers {"Accept" "pson"
+                                "Content-type" "text/pson"}
```
rlinehan: Can we switch to using JSON instead of PSON now? or should we add a TODO for that?

test/integration/puppetlabs/services/master/master_service_test.clj
```diff
@@ -440,3 +461,212 @@
...
+     {:jruby-puppet {:max-active-instances 2 ; we need 2 jruby-instances since processing the report uses an instance
+                     :master-code-dir test-resources-code-dir
+                     :master-conf-dir master-service-test-runtime-dir}
+      :metrics {:server-id "localhost"}}
+     (let [jruby-service (tk-app/get-service app :JRubyPuppetService)
+           jruby-instance (jruby-testutils/borrow-instance jruby-service :http-report-processor-metrics-test)
+           container (:scripting-container jruby-instance)]
+       (try
+         ;; we don't care at all about the content of the report, just that it is valid
+         (let [report (.runScriptlet container "Puppet::Transaction::Report.new('apply').to_pson")]
```
rlinehan: Can we switch to using JSON instead of PSON now? or should we add a TODO for that?

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -463,6 +464,16 @@
         {:status 200
          :body (get-code-content environment code-id file-path)}))))
 
+(schema/defn http-client-metrics-summary
+  :- {:metrics-data [http-client-common/MetricIdMetricData]
+      :sorted-metrics-data [http-client-common/MetricIdMetricData]}
+  [metric-registry :- MetricRegistry
+   metric-ids-to-select :- (schema/atom [[schema/Str]])]
```
camlow325: Since we'd reference this in a few places, might be nice to factor the `(schema/atom [[schema/Str]])` out into a separate def.

src/clj/puppetlabs/services/master/master_service.clj
```diff
@@ -165,9 +173,18 @@
...
     [this context]
     (log/info (i18n/trs "Puppet Server has successfully started and is now ready to handle requests"))
-    context))
+    context)
+
+  (add-metric-ids-to-http-client-metrics-list!
+   [this metric-ids-to-add]
+   (let [metric-ids-from-context (:http-client-metric-ids-for-status
+                                  (tk-services/service-context this))]
+     (master-core/add-metric-ids-to-http-client-metrics-list! metric-ids-from-context
```
camlow325: Do you think it's worth "locking" the list so that nothing can append to it after the service has started, like we do with registries in tk-metrics?  I think I'm good with not adding that on - just curious if you'd thought about it.

test/integration/puppetlabs/services/master/master_service_test.clj
camlow325: Does the version of the puppet submodule that we're pinned to right now understand json yet?  I was thinking we'd need to at least wait until [PUP-7259](https://tickets.puppetlabs.com/browse/PUP-7259) lands.  Maybe we should create a ticket to clean up tests to use json instead of pson after that work has all wrapped up though?  Seems reasonable to me.

test/integration/puppetlabs/services/master/master_service_test.clj
```diff
@@ -440,3 +461,212 @@
...
+                   (make-request-with-metric-id "['puppetdb', 'query']")
+                   (is (= [["puppetdb" "query"]] (map :metric-id (get-http-client-metrics-status)))))
+
+                 (testing "only includes exact metric-ids it's supposed to have"
+                   (make-request-with-metric-id "['puppetdb', 'command', 'replace_catalog', 'foonode']")
+                   (let [metric-ids (map :metric-id (get-http-client-metrics-status))]
+                     (is (some #(= ["puppetdb" "command" "replace_catalog"] %) metric-ids))
+                     (is (not-any? #(= ["puppetdb" "command" "replace_catalog" "foodnode"] %)
+                                   metric-ids))
+                     (is (= #{["puppetdb" "query"] ["puppetdb" "command" "replace_catalog"]}
```
camlow325: Do we need the prior two assertions if this one is checking the entire set?

test/integration/puppetlabs/services/master/master_service_test.clj
camlow325: Seems like to_json for the request payload might be fine here already.  Since I don't think this would produce any funky character encoding that would require it to be rendered as "pson", maybe using the "Content-Type" of "text/pson" for the request but a request body that's actually json would be okay?  If we need to wait on the json "Content-Type" to work until after [PUP-7259](https://tickets.puppetlabs.com/browse/PUP-7259) comes in, though, maybe it would be better to just use to_pson here for now and change it later?

src/clj/puppetlabs/services/master/master_service.clj
rlinehan: Hm. I think there the problem is that after we create a `GraphiteReporter` with a filter, we can't modify it. Here, though, if we did add metric ids to a list some other time, then the status endpoint would have an updated list the next time you called it. So I kind of am inclined to not add that, unless we feel like it's weird that these two things have different behavior.

test/integration/puppetlabs/services/master/master_service_test.clj
rlinehan: Nah, I was just being super explicit, but you're right it's not necessary.

test/integration/puppetlabs/services/master/master_service_test.clj
rlinehan: Looks like it'll pass right now if we make it `to_json` so I'll change to that.

test/integration/puppetlabs/services/master/master_service_test.clj
rlinehan: Yeah, I tried switching to json here and it didn't work. I'll put in a ticket about updating all our tests to use json in headers rather than pson.

rlinehan: Filed SERVER-1762

rlinehan:
@camlow325 pushed a commit addressing your comments.

rlinehan:
@jpinsonault can you review this when you get a chance?

PR #1355, opened 2017-03-23
(maint) Abort dev-setup script on error

adrienthebo:
The dev-setup script doesn't do any error checking but a casual glance
indicates that any command failing should be a critical error. This
commit adds `set -e` which will cause the script to abort if any command
exits with a non-zero exit status.


adrienthebo:
This might have been the fastest PR merge I've ever seen. 💯

PR #1356, opened 2017-03-23
(SERVER-1732) default use-legacy-auth-conf to false

adrienthebo:


camlow325:
I think we should also change [this comment](https://github.com/puppetlabs/puppetserver/blob/2.7.2/dev/puppetserver.conf.sample#L74-L76) to reflect that the default would now be false.

camlow325:
There are also some notes about the default for the `use-legacy-auth-conf` setting in a few of the [docs](https://github.com/puppetlabs/puppetserver/tree/2.7.2/documentation) that would be good to update for this as well.  We usually like to make updates to the docs along with code changes in the same PR, where possible, and ping someone from the docs team, like @gguillotte, for their awareness/review on those pieces.

camlow325:
It would also be good to update where we reference the `use-legacy-auth-conf` setting in the conf file which goes into the Puppet Server package to reflect the new default.  That would be [here](https://github.com/puppetlabs/puppetserver/blob/2.7.2/resources/ext/config/conf.d/puppetserver.conf#L50-L53).

camlow325:
In [this comment](https://tickets.puppetlabs.com/browse/SERVER-1732?focusedCommentId=411060&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-411060) on the ticket, I had suggested that we might go a step further with this ticket and effectively eliminate the ability for a user to be able to set the `use_legacy_auth_conf` setting.  This PR only changes the default but allows a user the ability to still configure the setting.

While I still think that may be a good forcing function for users to migrate, I'm having second thoughts about my suggestion now, though.  The main problem, I fear, would be around how we would be able to continue to run the Ruby Puppet acceptance test suite with no ability to have Ruby Puppet honor the legacy auth.conf rules anymore.  Unless we don't have any auth.conf test coverage in the Ruby Puppet acceptance suite, I fear that not having this ability could lead to a number of test failures.  There would probably be ways to hide direct user configurability of the setting and still work through the test suite logistics - e.g., skipping some tests when running in Puppet Server CI.  Ultimately, I'm not sure if it's worth it to do that, though.  It might be better to have a second round where we rip out `use-legacy-auth-conf` setting configurability in a later major release where we know WEBrick / Rack / auth.conf support would be gone from the underlying Ruby Puppet code as well.

Assuming we go with the approach you have on this PR, though, I think we may still need to do some extra work in the Puppet Server acceptance test pre-suite to ensure that `use-legacy-auth-conf` is set to `true` before the Ruby Puppet acceptance test suite is run - to enable legacy tests to continue passing.  Unfortunately, then, we may also need to have something similar done in the pre-suite for the Puppet agent/platform pipeline as well.  I think that before we would merge this PR that we should do some extra work to see what would happen if the existing Ruby Puppet acceptance tests are run with Puppet Server implicitly using a value of `false` for `use-legacy-auth-conf`.  Can you look into that?  @rlinehan or I could definitely help you with that where needed.

@rlinehan - WDYT about all of this?

adrienthebo:
I'll look into running the acceptance tests with this change applied and see if anything exciting catches fire. While I do that would you like me to throw in a deprecation warning if the setting is given?

camlow325:
@adrienthebo I think we're already good on a deprecation warning - see [this code](https://github.com/puppetlabs/puppetserver/blob/2.7.2/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj#L247-L251).  If you think there's some language there which is worth updating, though, open to whatever you think makes sense.

adrienthebo:
I ran the Puppet acceptance tests against a Puppetserver containing these changes, and while there was one failure it might be unrelated because it deals with server side facts.

adrienthebo:
Documentation and config file comments for `use-legacy-auth-conf` updated in 573c6085.

camlow325:
Looks like there are a couple of Travis test failures off of the latest changes:

```
lein test :only puppetlabs.puppetserver.auth-conf-test/custom-oids-passed-to-tk-auth

FAIL in (custom-oids-passed-to-tk-auth) (auth_conf_test.clj:371)
puppet server successfully utilizes custom oid mappings and puppet short names for authorization master endpoints use oid shortnames
expected: (= 403 (:status (http-get-no-ssl path url-encoded-deniable-cert)))
  actual: (not (= 403 200))

lein test :only puppetlabs.puppetserver.auth-conf-test/custom-oids-passed-to-tk-auth

FAIL in (custom-oids-passed-to-tk-auth) (auth_conf_test.clj:378)
puppet server successfully utilizes custom oid mappings and puppet short names for authorization legacy endpoints support oid shortnames
expected: (= 403 (:status (http-get-no-ssl path url-encoded-deniable-cert)))
  actual: (not (= 403 200))
```

adrienthebo:
(╯°□°）╯︵ ┻━┻

Fixing.

dev/puppetserver.conf
```diff
@@ -74,10 +74,11 @@ jruby-puppet: {
...
     max-active-instances: 1
 
-    # (optional) Authorize access to Puppet master endpoints via rules specified
-    # in the legacy Puppet auth.conf file (if true or not specified) or via rules
-    # specified in the Puppet Server HOCON-formatted auth.conf (if false).
-    use-legacy-auth-conf: false
+    # (optional) Authorize access to Puppet master endpoints via rules
+    # specified in the legacy Puppet auth.conf file (if true) or via rules
+    # specified in the Puppet Server HOCON-formatted auth.conf (if false or not
+    # specified.).
```
camlow325: Nit - could get rid of the extra '.' - from (if false or not specified.).

adrienthebo: De-nitted in e7a63df.

resources/ext/config/conf.d/puppetserver.conf
```diff
@@ -47,10 +47,11 @@ jruby-puppet: {
...
 
-    # (optional) Authorize access to Puppet master endpoints via rules specified
-    # in the legacy Puppet auth.conf file (if true or not specified) or via rules
-    # specified in the Puppet Server HOCON-formatted auth.conf (if false).
-    #use-legacy-auth-conf: false
+    # (optional) Authorize access to Puppet master endpoints via rules
+    # specified in the legacy Puppet auth.conf file (if true) or via rules
+    # specified in the Puppet Server HOCON-formatted auth.conf (if false or not
+    # specified.).
+    #use-legacy-auth-conf: true
```
camlow325: Same nit here with the extraneous period as with the dev version of this file.

resources/ext/config/conf.d/puppetserver.conf
```diff
@@ -47,10 +47,11 @@ jruby-puppet: {
...
     #max-active-instances: 1
 
-    # (optional) Authorize access to Puppet master endpoints via rules specified
-    # in the legacy Puppet auth.conf file (if true or not specified) or via rules
-    # specified in the Puppet Server HOCON-formatted auth.conf (if false).
-    #use-legacy-auth-conf: false
+    # (optional) Authorize access to Puppet master endpoints via rules
+    # specified in the legacy Puppet auth.conf file (if true) or via rules
+    # specified in the Puppet Server HOCON-formatted auth.conf (if false or not
+    # specified.).
```
camlow325: Same nit here with the extraneous period as with the dev version of this file.

documentation/deprecated_features.markdown
```diff
@@ -21,11 +21,11 @@ endpoints:
 * [Puppet's HTTPS API (current)](https://docs.puppet.com/puppet/latest/reference/http_api/http_api_index.html)
 * [Puppet's HTTPS API (3.x)](https://github.com/puppetlabs/puppet/blob/3.8.0/api/docs/http_api_index.md)
 
-For a value of `true` (also the default if not specified), the core
+For a value of `true` , the core
```
gguillotte: Extraneous space after ``true``.

gguillotte:
👍 One nitpick but no blockers on docs.

documentation/deprecated_features.markdown
adrienthebo: Fixed in f8a0f9d.

camlow325:
It looks like we have a acceptance test fails after this went in, both in the [Ruby Puppet tests](https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppetserver/view/all/job/platform_puppetserver_integration-system_no-conditional_full-master/190/) and the [Puppet 3 Compatibility tests](https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppetserver/view/all/job/platform_puppetserver_integration-system_no-conditional_puppet3-compat_master/183/).

For each of the cases, it appears that failures are due to with HTTP 403/unauthorized errors, presumably because they try to open up custom auth rules in the legacy auth.conf that we no longer honor in Puppet Server when `use_legacy_auth_conf` is set to `true`.

We'll probably want to hash out with the Agent/Platform team as to how we should work around this in both of our test suites for the Ruby Puppet acceptance tests.  Since we only have a few affected tests, maybe it would be best to modify just those tests to ensure that a puppetserver auth conf rule which opens up the extra access that those specific tests need is created - allowing the rest of the suite to be run with the new Puppet Server default auth setting?  For tests that modify the auth settings via Puppet code runs, we could make use of the [puppet_authorization](https://forge.puppet.com/puppetlabs/puppet_authorization) module.

The [Puppet 3 Compatibility tests](https://github.com/puppetlabs/puppetserver/tree/2.7.2/acceptance/suites/puppet3_tests) are only run in the Puppet Server pipeline, so we should be able to modify those tests in the Puppet Server repo such that they can run with the new `use_legacy_auth_conf` setting.

PR #1357, opened 2017-03-24
(SERVER-1761) Enable profiler by default

adrienthebo:
This commit defaults the Puppet profiler to true, and removes some
extraneous logging.

I've dropped a set of tests that seemed to be exercising the debug log statement in `MetricsPuppetProfiler.finish()`, if those need to be replaced instead of dropped let me know.

### DIFF
--- a/test/unit/puppetlabs/services/puppet_profiler/puppet_profiler_core_test.clj
+++ b/test/unit/puppetlabs/services/puppet_profiler/puppet_profiler_core_test.clj
@@ -36,39 +36,6 @@
                          (= level (.getLevel event))))
         @logs))
 
-(defn profile
-  [profiler message metric-id]
-  (let [metric-id (into-array String metric-id)
-        context (.start profiler message metric-id)]
-    (.finish profiler context message metric-id)))
-
-(deftest test-metrics-profiler
-  (testing "metrics profiler"
-    (logutils/with-test-logging
-     (with-test-logs logs
-       (let [registry (MetricRegistry.)
-             profiler (metrics-profiler "localhost" registry)]
-         (profile profiler "hi" ["function" "hiera-lookup"])
-         (profile profiler "bye" ["compile" "init-environment"])
-         (testing "keeps timers for all metrics"
-           (let [metrics-map (.getMetrics registry)
-                 expected-metrics ["puppetlabs.localhost.function"
-                                   "puppetlabs.localhost.function.hiera-lookup"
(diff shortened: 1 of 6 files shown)


test/unit/puppetlabs/services/puppet_profiler/puppet_profiler_core_test.clj
```diff
@@ -36,39 +36,6 @@
...
-         (profile profiler "bye" ["compile" "init-environment"])
-         (testing "keeps timers for all metrics"
-           (let [metrics-map (.getMetrics registry)
-                 expected-metrics ["puppetlabs.localhost.function"
-                                   "puppetlabs.localhost.function.hiera-lookup"
-                                   "puppetlabs.localhost.compile"
-                                   "puppetlabs.localhost.compile.init-environment"]]
-             (is (= (set expected-metrics)
-                    (.keySet metrics-map)))
-             (is (every? #(instance? Timer (.get metrics-map %)) expected-metrics))))
```
camlow325: Getting rid of the tests below around logging messages seems okay but I think we should leave these tests which validate that timers were added to the registry as expected.

camlow325:
There are some docs about the profile default in [this file](https://github.com/puppetlabs/puppetserver/blob/2.7.2/documentation/config_file_puppetserver.markdown) that I think need to be updated to cover the new default.

camlow325:
Looks like there were a couple of test failures from Travis:

```
lein test puppetlabs.services.puppet-profiler.puppet-profiler-service-test

lein test :only puppetlabs.services.puppet-profiler.puppet-profiler-service-test/test-profiler-service

FAIL in (test-profiler-service) (puppet_profiler_service_test.clj:31)
get-profiler returns nil if profiling is not enabled
expected: (pred? (get-profiler service))
  actual: false

lein test :only puppetlabs.services.puppet-profiler.puppet-profiler-service-test/test-profiler-service

FAIL in (test-profiler-service) (puppet_profiler_service_test.clj:31)
get-profiler returns nil if profiling is not enabled
expected: (pred? (get-profiler service))
  actual: false
```

test/unit/puppetlabs/services/puppet_profiler/puppet_profiler_core_test.clj
adrienthebo: Re-added in 0446d91.

adrienthebo:
Tests should be passing and documentation updated.

rlinehan:
💯

PR #1359, opened 2017-03-27
Fix version numbers in doc.

gguillotte:
Didn't bump from 2.6 to 2.7.

### DIFF
--- a/documentation/install_from_packages.markdown
+++ b/documentation/install_from_packages.markdown
@@ -11,7 +11,7 @@ canonical: "/puppetserver/latest/install_from_packages.html"
 
 Puppet Server is configured to use 2 GB of RAM by default. If you'd like to just play around with an installation on a Virtual Machine, this much memory is not necessary. To change the memory allocation, see [Memory Allocation](#memory-allocation).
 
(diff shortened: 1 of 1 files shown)


documentation/install_from_packages.markdown
```diff
@@ -11,7 +11,7 @@ canonical: "/puppetserver/latest/install_from_packages.html"
...
 
-> **Note:** Puppet masters running Puppet Server 2.6 depend on [Puppet Agent 1.6.0](https://docs.puppet.com/puppet/4.6/reference/about_agent.html) or newer, which installs [Puppet 4.6](https://docs.puppet.com/puppet/4.6/) and compatible versions of its related tools and dependencies on the server. Puppet agents running older versions of Puppet Agent can connect to Puppet Server 2.6 --- this requirement applies to the Puppet Agent running on the Puppet Server node *only*.
+> **Note:** Puppet masters running Puppet Server 2.7 depend on [Puppet Agent 1.7.0](https://docs.puppet.com/puppet/4.7/reference/about_agent.html) or newer, which installs [Puppet 4.7](https://docs.puppet.com/puppet/4.7/) and compatible versions of its related tools and dependencies on the server. Puppet agents running older versions of Puppet Agent can connect to Puppet Server 2.7 --- this requirement applies to the Puppet Agent running on the Puppet Server node *only*.
```
camlow325: FWIW, we would support Puppet Server 2.7+ being installed with Puppet Agent 1.6+.  See [our package dependency](https://github.com/puppetlabs/puppetserver/blob/2.7.2/resources/ext/ezbake.conf#L10) for this.

Practically speaking, though, we probably had been testing Puppet Agent 1.7 or newer most often with Puppet Server 2.7+, so I'm okay with this text if that's what you'd like to do.

Might be good to target this at the stable branch instead of the master branch if we want to republish it right away / have it available for a Puppet Server 2.7.3 or later release, if we end up doing one.  Current work going into the master branch is intended to roll into the Puppet Server 5.0 release.

PR #1361, opened 2017-03-29
(SERVER-1759) Forward JRuby exit status for subcommands

adrienthebo:


src/clj/puppetlabs/puppetserver/cli/subcommand.clj
```diff
@@ -52,3 +52,11 @@
       (throw+))
     (finally
       (shutdown-agents))))
+
+(defn jruby-run
```
adrienthebo: Would it be better to push this behavior into https://github.com/puppetlabs/jruby-utils/blob/master/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj#L279 ? Or is it better to keep the invocation of `System/exit` as far down the stack as possible?

camlow325: I'm okay with having this be done here as you've done it.  The `System/exit` for non-JRuby based subcommands already lives in this namespace, so it seems reasonable to do the same here.

We've talked in the past about maybe pushing more of the JRuby CLI functions into `jruby-utils`.  If we had other consumers of that besides just Puppet Server, that might still make sense to do.  Doing that now, though, would just lead us to need to do a release of jruby-utils everytime we want to expose a change to Puppet Server, with no code sharing benefits.  That wouldn't seem worthwhile.

camlow325:
We have a few beaker-level tests for the different subcommands [here](https://github.com/puppetlabs/puppetserver/tree/2.7.2/acceptance/suites/tests/010-puppetserver-cli/subcommand).  I think it would be good for some test coverage of these changes to add a few extra tests for `gem`, `irb`, and `ruby` which validate that the shell receives expected success / failure exit codes.

src/clj/puppetlabs/puppetserver/cli/subcommand.clj
rlinehan: Agreed on this - if we get other consumers of jruby-utils then I think it'll be worth revisiting this, but for the time being it doesn't seem worth the effort.

rlinehan:
👍, agree with Jeremy that it would be good to expand test coverage for this.

adrienthebo:
Just added acceptance tests for ruby, irb, and gem.

PR #1362, opened 2017-03-29
(maint) Bump puppet and puppet-agent versions 29 march 2017

rlinehan:


camlow325:
Ain't going to get any redder in CI right now since we're pinned to builds that don't exist anymore, so I'm just going to merge this.

PR #1364, opened 2017-03-30
(maint) Use legacy auth.conf for Puppet acceptance tests

adrienthebo:


adrienthebo:
Verified with the following:

```
export PACKAGE_BUILD_VERSION="5.0.0.master.SNAPSHOT.2017.03.30T1010"
export PUPPET_BUILD_VERSION="1.9.3"
export PUPPET_LEGACY_VERSION="3.8.1"

bundle exec beaker \
  --debug \
  --root-keys \
  --no-color \
  --repo-proxy \
  --preserve-hosts never \
  --type aio \
  --helper acceptance/lib/helper.rb \
  --options-file acceptance/config/beaker/options.rb \
  --load-path acceptance/lib,ruby/puppet/acceptance/lib \
  --pre-suite acceptance/suites/pre_suite/foss \
  --config acceptance/config/beaker/jenkins/redhat-7-x86_64-mdca-a.cfg \
  --keyfile ~/.ssh/id_rsa-acceptance \
  --tests ruby/puppet/acceptance/tests/allow_arbitrary_node_name_fact_for_agent.rb,ruby/puppet/acceptance/tests/allow_arbitrary_node_name_for_agent.rb,ruby/puppet/acceptance/tests/node/check_woy_cache_works.rb
```

rlinehan:
Alright, so what are these tests trying to do, and rather than using the legacy auth.conf for all the tests, are there just certain tests that we should be trying to modify? It seems a little heavy handed to run all the puppet acceptance tests in a different way than we run puppet in a typical setup.

adrienthebo:
It is a heavy handed solution, and I think it's possible to reduce the scope of this. I didn't want to leave CI jammed up though, since I'm used to an acceptance failure causing the commit to be reverted. What do you think of merging this PR to get us green and then I'll throw a ticket in the next sprint to clean this up?

I prefer the latter option because reducing the scope of the legacy auth.conf changes (or replacing them with tk-auth when puppetserver is running) will mean changes to the Puppet codebase which means that getting those changes landed could take a while.

rlinehan:
Okay, that's probably sensible so long as we don't think it's hiding any problems. Probably going to be easier than trying to revert various stuff etc.

rlinehan:
Okay, that's probably sensible so long as we don't think it's hiding any problems. Probably going to be easier than trying to revert various stuff etc.

adrienthebo:
Ticket added for fine tuning auth.conf settings in https://tickets.puppetlabs.com/browse/SERVER-1766.

rlinehan:
Travis failed because rubygems--. Which is stupid because travis doesn't test acceptance things. Merging despite failure.

camlow325:
Looks like @rlinehan was good with this too so I'll go ahead and merge it.

rlinehan:
Oh wait what I meant to merge this yesterday. Thanks @camlow325. I think I got distracted because my wifi was terrible.

I believe Agent is currently pinned to an older version of Puppet Server, so hopefully we can get SERVER-1766 in and not have to do this update for them too.

PR #1365, opened 2017-04-04
(SERVER-1767) Limit concurrent requests to Puppet API

Sharpie:
This patch implements a ring handler that rejects incoming requests to the
Puppet API with a 503 status code when the backlog of outstanding requests
exceeds a given limit. The handler can also be configured with a splay limit,
a random fraction of which is set as the value of a Retry-After header that is
returned with the 503 error.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/jruby_request.clj
+++ b/src/clj/puppetlabs/puppetserver/jruby_request.clj
@@ -1,10 +1,12 @@
 (ns puppetlabs.puppetserver.jruby-request
+  (:import [clojure.lang IFn])
   (:require [clojure.tools.logging :as log]
             [ring.util.response :as ring-response]
             [slingshot.slingshot :as sling]
             [puppetlabs.ring-middleware.core :as middleware]
             [puppetlabs.ring-middleware.utils :as ringutils]
             [puppetlabs.services.protocols.jruby-puppet :as jruby-puppet]
+            [puppetlabs.services.protocols.jruby-metrics :as jruby-metrics]
             [puppetlabs.services.jruby.jruby-puppet-service :as jruby]
             [schema.core :as schema]
             [puppetlabs.puppetserver.common :as ps-common]
@@ -44,6 +46,30 @@
     (jruby/with-jruby-puppet
      jruby-puppet jruby-service {:request (dissoc request :ssl-client-cert)}
      (handler (assoc request :jruby-instance jruby-puppet)))))
+
+(schema/defn ^:always-validate
+  wrap-with-request-queue-limit :- IFn
+  "Middleware fn that short-circuits incoming requests with a 503 'Service
+  Unavailable' response if the queue of outstanding requests for JRuby
+  instances exceeds the limit given by the max-queued-requests argument.
+  The response will include a Retry-After header set to a random fraction
+  of the value given by the max-retry-delay argument if set to a positive
+  non-zero number."
+  [handler :- IFn
+   metrics-service :- (schema/protocol jruby-metrics/JRubyMetricsService)
+   max-queued-requests :- schema/Int
+   max-retry-delay :- schema/Int]
+  (fn [request]
+    (let [metrics-map (jruby-metrics/get-metrics metrics-service)
+          requested-count (-> metrics-map :requested-instances deref count)]
+      (if (>= requested-count max-queued-requests)
+        (let [err-msg (i18n/trs "The number of requests waiting for a JRuby instance has exceeded the limit allowed by the max-queued-requests setting.")
+              response (output-error request {:msg err-msg} 503)]
+          (if (pos? max-retry-delay)
+            (assoc-in response [:headers "Retry-After"]
+                      (-> (rand) (* max-retry-delay) int str))
+            response))
+        (handler request)))))
 
 (defn get-environment-from-request
   "Gets the environment from the URL or query string of a request."
--- a/src/clj/puppetlabs/services/request_handler/request_handler_service.clj
+++ b/src/clj/puppetlabs/services/request_handler/request_handler_service.clj
@@ -2,6 +2,7 @@
   (:require [puppetlabs.trapperkeeper.core :as tk]
             [puppetlabs.services.protocols.request-handler :as handler]
             [puppetlabs.services.request-handler.request-handler-core :as request-handler-core]
+            [puppetlabs.puppetserver.jruby-request :as jruby-request]
             [puppetlabs.trapperkeeper.services :as tk-services]
             [clojure.tools.logging :as log]
             [puppetlabs.i18n.core :as i18n]))
@@ -9,11 +10,21 @@
 (tk/defservice request-handler-service
   handler/RequestHandlerService
   [[:PuppetServerConfigService get-config]
+   [:ConfigService get-in-config]
    [:VersionedCodeService current-code-id]
-   [:JRubyPuppetService]]
+   [:JRubyPuppetService]
+   [:JRubyMetricsService]]
   (init [this context]
-    (let [jruby-service (tk-services/get-service this :JRubyPuppetService)
-          config (get-config)]
+    (let [config (get-config)
+          max-queued-requests (get-in-config [:jruby-puppet :max-queued-requests] 0)
+          max-retry-delay (get-in-config [:jruby-puppet :max-retry-delay] 1800)
+          jruby-service (tk-services/get-service this :JRubyPuppetService)
+          metrics-service (tk-services/get-service this :JRubyMetricsService)
+          request-handler (request-handler-core/build-request-handler
+                            jruby-service
+                            (request-handler-core/config->request-handler-settings
+                              config)
+                            current-code-id)]
       (when (contains? (:master config) :allow-header-cert-info)
         (if (true? (get-in config [:jruby-puppet :use-legacy-auth-conf]))
           (log/warn (format "%s %s"
@@ -22,12 +33,14 @@
           (log/warn (format "%s %s"
                             (i18n/trs "The ''master.allow-header-cert-info'' setting is deprecated and will be ignored in favor of the ''authorization.allow-header-cert-info'' setting because the ''jruby-puppet.use-legacy-auth-conf'' setting is ''false''.")
                             (i18n/trs "Remove the ''master.allow-header-cert-info'' setting.")))))
-      (assoc context :request-handler
-                     (request-handler-core/build-request-handler
-                      jruby-service
-                      (request-handler-core/config->request-handler-settings
-                       config)
-                      current-code-id))))
+      (assoc context :request-handler (if (pos? max-queued-requests)
+                                        (jruby-request/wrap-with-request-queue-limit
+                                          request-handler
+                                          metrics-service
+                                          max-queued-requests
+                                          max-retry-delay)
+                                        request-handler))))
+
   (handle-request
     [this request]
     (let [handler (:request-handler (tk-services/service-context this))]
--- a/documentation/config_file_puppetserver.markdown
+++ b/documentation/config_file_puppetserver.markdown
@@ -54,6 +54,10 @@ The `puppetserver.conf` file contains settings for Puppet Server software. For a
 
         JRuby flushing can be useful for working around buggy module code that would otherwise cause memory leaks, but it slightly reduces performance whenever a new JRuby instance reloads all of the Puppet Ruby code. If memory leaks from module code are not an issue in your deployment, the default value of 0 performs best.
 
+    * `max-queued-requests`: Optional. The maximum number of requests that may be queued waiting to borrow a JRuby from the pool. Once this limit is exceeded, a 503 "Service Unavailable" response will be returned for all new requests until the queue drops below the limit. If `max-retry-delay` is set to a positive value, then the 503 responses will include a `Retry-After` header indicating a random sleep time after which the client may retry the request. The default is 0, which disables the queue limit.
+
+    * `max-retry-delay`: Optional. Sets the upper limit for the random sleep set as a `Retry-After` header on 503 responses returned when `max-queued-requests` is enabled. A value of 0 will cause the `Retry-After` header to be omitted. Default is 1800 seconds which corresponds to the default run interval of the Puppet daemon.
+
     * `borrow-timeout`: Optional. The timeout in milliseconds, when attempting to borrow an instance from the JRuby pool. The default is 1200000.
 
     * `use-legacy-auth-conf`: Optional. The method to be used for authorizing access to the HTTP endpoints served by the master service. The applicable endpoints are listed in [Puppet v3 HTTP API](https://docs.puppet.com/puppet/latest/reference/http_api/http_api_index.html#puppet-v3-http-api). As of Puppet Server 5.0, this setting's default value is false.
--- a/test/integration/puppetlabs/services/jruby/jruby_metrics_service_test.clj
+++ b/test/integration/puppetlabs/services/jruby/jruby_metrics_service_test.clj
@@ -147,6 +147,10 @@
                                       resp)))
             routes (comidi/context "/foo"
                      (comidi/routes
+                       ;; Route handler that isn't wrapped with the request
+                       ;; coordinator.
+                       (comidi/GET ["/uncoord/" :uncoord] request
+                         (handle-request request))
                        (comidi/GET ["/bar/" :bar] request
                          (app-request-handler request))
                        (comidi/GET ["/baz/" :baz] request
@@ -874,3 +878,36 @@
                                  :return-count 2
                                  :current-borrowed-instances -2})
         (is (= (expected-metrics-values) (current-metrics-values)))))))
+
+(deftest request-queue-limit
+  (with-metrics-test-env
+    test-env
+    (assoc-in default-test-config
+             [:jruby-puppet :max-queued-requests] 2)
+    (let [{:keys [current-metrics-values coordinator]} test-env
+          {:keys [requested-count requested-instances borrowed-instances]} (:metrics test-env)]
+      (testing "denies requests when rate limit hit"
+        ;; Block up two JRuby instances
+        (async-request coordinator 1 "/foo/bar/async1" :returning-jruby)
+        (async-request coordinator 2 "/foo/bar/async2" :returning-jruby)
+
+        ;; Create two pending requests
+        (async-request coordinator 3 "/foo/bar/async3")
+        (async-request coordinator 4 "/foo/bar/async4")
+        (coordinator/unblock-task-to coordinator 3 :borrowed-jruby)
+        (coordinator/unblock-task-to coordinator 4 :borrowed-jruby)
+
+        ; Wait for async requests to hit metrics.
+        (while (> 4 (.getCount requested-count)))
+
+        (let [resp        (http-get "http://127.0.0.1:8140/foo/uncoord/sync1")
+              status-code (:status resp)
+              retry-after (-> resp
+                              (get-in [:headers "retry-after"])
+                              Integer/parseInt)]
+          (is (= 503 status-code))
+          (is (<= 0 retry-after 1800)))
+
+        ;; unblock all requests
+        (doseq [i (range 1 5)]
+          (coordinator/final-result coordinator i))))))
--- a/test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
+++ b/test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
@@ -11,6 +11,7 @@
             [me.raynes.fs :as fs]
             [puppetlabs.trapperkeeper.internal :as tk-internal]
             [puppetlabs.trapperkeeper.core :as tk]
+            [puppetlabs.services.jruby.jruby-metrics-service :as jruby-metrics-service]
             [puppetlabs.services.request-handler.request-handler-service :as handler-service]
             [puppetlabs.services.versioned-code-service.versioned-code-service :as vcs]
             [puppetlabs.services.config.puppet-server-config-service :as ps-config]
@@ -228,6 +229,7 @@
                                       "test knows the webserver is still running"))
            services (jruby-testutils/add-mock-jruby-pool-manager-service
                      (conj jruby-testutils/jruby-service-and-dependencies
+                           jruby-metrics-service/jruby-metrics-service
                            handler-service/request-handler-service
                            ps-config/puppet-server-config-service
                            vcs/versioned-code-service)
--- a/test/unit/puppetlabs/services/request_handler/request_handler_core_test.clj
+++ b/test/unit/puppetlabs/services/request_handler/request_handler_core_test.clj
@@ -25,6 +25,7 @@
             [puppetlabs.trapperkeeper.services.webrouting.webrouting-service :as routing-service]
             [puppetlabs.trapperkeeper.services.authorization.authorization-service :as authorization-service]
             [puppetlabs.trapperkeeper.services.metrics.metrics-service :as metrics]
+            [puppetlabs.services.jruby.jruby-metrics-service :as jruby-metrics-service]
             [puppetlabs.trapperkeeper.services.status.status-service :as status-service]
             [puppetlabs.services.protocols.jruby-puppet :as jruby-protocol]
             [puppetlabs.puppetserver.testutils :as testutils]
@@ -326,6 +327,7 @@
                      custom-vcs
                      tk-scheduler/scheduler-service
                      metrics/metrics-service
+                     jruby-metrics-service/jruby-metrics-service
                      status-service/status-service]]
        (jruby-bootstrap/with-puppetserver-running-with-services-and-mock-jruby-puppet-fn
         "For this test we're basically just validating that a JRuby is reserved


Sharpie:
Tossing this up to kick it around a bit. If the idea looks good, I'll finish the PR out with docs, tests, and error handling.

Sharpie:
The second piece of this implementation would be a change to the Puppet::HTTP client to handle 503 errors with `Retry-After` headers by sleeping and then re-trying the request.

src/clj/puppetlabs/services/request_handler/request_handler_service.clj
```diff
@@ -22,12 +32,16 @@
...
                             (i18n/trs "Remove the ''master.allow-header-cert-info'' setting.")))))
-      (assoc context :request-handler
-                     (request-handler-core/build-request-handler
-                      jruby-service
-                      (request-handler-core/config->request-handler-settings
-                       config)
-                      current-code-id))))
+      ;; TODO: Throw error when max-active-requests > 0, but metrics-service is nil
+      ;; because :JRubyMetricsService isn't in bootstrap.cfg.
+      (assoc context :request-handler (if (> max-active-requests 0)
```
camlow325: Why would we want to tie this to the `max-active-requests` specifically?  Where `max-active-requests` indicates the number of requests which can be made to a JRuby instance before it is recycled, it seems like the values for it could range wildly.  Ideally, users would be able to set this to 0 (or to a very high number) for best performance - i.e., for the case where there are no memory leaks.  It seems like in either of these cases that there wouldn't be much of a benefit to the extra code added here.

Sharpie: Ah. I think you're thinking of `max-requests-per-instance` which is the number of requests an individual JRuby can handle before being recycled. The `max-active-requests` here is a new setting that gives the limit on outstanding requests for JRuby instances.

Would love to hear suggestions for better or less confusing names for this setting.

camlow325: Ah, you're right.  😄  That was the setting I was thinking of.  `max-active-requests` sounds like a reasonable name to me for a new setting - can't think of a better one.

src/clj/puppetlabs/services/request_handler/request_handler_service.clj
```diff
@@ -22,12 +32,16 @@
           (log/warn (format "%s %s"
                             (i18n/trs "The ''master.allow-header-cert-info'' setting is deprecated and will be ignored in favor of the ''authorization.allow-header-cert-info'' setting because the ''jruby-puppet.use-legacy-auth-conf'' setting is ''false''.")
                             (i18n/trs "Remove the ''master.allow-header-cert-info'' setting.")))))
-      (assoc context :request-handler
-                     (request-handler-core/build-request-handler
-                      jruby-service
-                      (request-handler-core/config->request-handler-settings
-                       config)
-                      current-code-id))))
+      ;; TODO: Throw error when max-active-requests > 0, but metrics-service is nil
```
camlow325: I think it would be better to separate out the dependency from the metrics service in order to cut down the complexity of the error handling a bit.  One possibility might be to declare a `request-counter` atom inside of the `wrap-with-concurrent-request-limit` helper which is incremented/decremented as requests are created/finished.  That way, we'd have a local way in this new code path to determine the number of active requests.

src/clj/puppetlabs/puppetserver/jruby_request.clj
```diff
@@ -45,6 +46,24 @@
...
+  to a random fraction of the limit."
+  [handler metrics-service request-limit splay-limit]
+  ;; TODO: Log warnings when limit is hit. Add i18n to user-facing strings.
+  (fn [request]
+    (let [metrics-map (jruby-metrics/get-metrics metrics-service)
+          requested-count (-> metrics-map :requested-count .getCount)]
+      (if (> requested-count request-limit)
+        (let [response (ringutils/plain-response 503 "JRuby capacity exceeded.")]
+          (if (> splay-limit 0)
+            (assoc-in response [:headers "Retry-After"]
```
camlow325: This is an interesting idea - controlling the splay from the server side.  Does the Puppet agent not already have any concept of a splay / randomization built into it to retry a request based on the receipt of specific HTTP error codes, like a 503?  This could get more interesting in the future if we could tune the server to more heuristically determine a good retry threshold to use based on load, rather than just always using a random value.  I think what you have would be a good approach for now, though.

src/clj/puppetlabs/services/request_handler/request_handler_service.clj
Sharpie: Hmm. That is an interesting idea. I suppose the trade-off is that the helper would then duplicate a lot of logic and error handling to ensure the counter is properly incremented and decremented --- logic which is already present (and tested!) in the metrics service. Also, if we ended up using heuristics to tune the retry behavior then we'd probably want to drive that using other bits of data from the JRuby metrics service.

So, I'm not sure which is better: keeping the retry handler decoupled from the metrics or not having to re-implement something that the metrics are already tracking.

src/clj/puppetlabs/puppetserver/jruby_request.clj
Sharpie: > Does the Puppet agent not already have any concept of a splay / randomization built into it to retry a request based on the receipt of specific HTTP error codes, like a 503?

Not yet. The agent's HTTP client code would have to be updated to respond to 503 error codes. However, the `Retry-After` header on a 503 is well-defined by standards, so it should be a very sensible change.

> This could get more interesting in the future if we could tune the server to more heuristically determine a good retry threshold to use based on load, rather than just always using a random value.

I was thinking that `average-wait-time` might be another metric that folks would be interested in setting a threshold on. Also, using a random multiple of `average-wait-time` to splay the agents might break up herds without requiring the agents to wait longer than necessary.

In the end, a limit based on `requested-count` and splay based on a static `splay-limit` seemed much simpler and probably "good enough" for a first iteration.

src/clj/puppetlabs/puppetserver/jruby_request.clj
```diff
@@ -45,6 +46,24 @@
...
+  "Middleware fn that returns a 503 'Service Unavailable' response if the
+  number of outstanding requests for a JRuby instance exceed the given limit.
+  If splay-limit is set, the response will include a Retry-After header set
+  to a random fraction of the limit."
+  [handler metrics-service request-limit splay-limit]
+  ;; TODO: Log warnings when limit is hit. Add i18n to user-facing strings.
+  (fn [request]
+    (let [metrics-map (jruby-metrics/get-metrics metrics-service)
+          requested-count (-> metrics-map :requested-count .getCount)]
+      (if (> requested-count request-limit)
```
camlow325: Hmm, not sure that `requested-count` is what you'd want to use for this.  I think this is an aggregate counter of all requests that have been made since service startup, not a count of the requests currently in flight.  Maybe getting the vector of `requested-instances` and doing a count of those would get you what you're looking for here, though.

src/clj/puppetlabs/puppetserver/jruby_request.clj
camlow325: `average-wait-time` might be difficult threshold to use for when to start rejecting requests with a 503, assuming that once you start rejecting you wouldn't be allowing any more borrow attempts through and so the `average-wait-time` would stay persistently above the threshold.  Using it as a factor in computing what value to return for the `Retry-After` value which accompanies a 503 may make sense, though.

In any event, I agree that starting with an implementation based on the active request count for the threshold to start returning 503 responses and that using the `Retry-After` header randomized from a more static value seems reasonable for now.

camlow325:
Thanks much, @Sharpie, for working this up.  TBH, my preference for handling the limiting behavior would be to come up with a more general-purpose request-limiting solution that we could apply across endpoints hosted under the webserver - something that could be made extensible to any consumers of the tk-webserver service, including PuppetDB and PE Console Services.  I'd been kicking around the idea in the back of my head for a while now on making use of Jetty's [QoSFilter](https://wiki.eclipse.org/Jetty/Reference/QoSFilter), allowing for controls which could cause requests to wait/suspend/fail per configured thresholds, prioritizing requests by attribute - most commonly, by URL endpoint.  That would probably involve considerably more work than what you've come up with here, though - and is probably not something we'd be able to get to for a while yet.  So, that said, I think the approach on this PR would be a significant improvement over what we have today.  Are you comfortable with introducing a request limiting solution which is more targeted at JRuby even if something broader were to come along later which might eventually replace it?

src/clj/puppetlabs/puppetserver/jruby_request.clj
Sharpie: Ahh. Good catch! `requested-instances` is definitely what I want since it looks like we don't `.dec` the counter when an instance is returned.

src/clj/puppetlabs/services/request_handler/request_handler_service.clj
camlow325: Good points - I'm a bit torn on this myself.

At the point we might want to use other metrics that the jruby-metrics service is already tracking to decide how to enforce request limits, I'd agree with you that it would seem better to avoid having to duplicate any of the metrics that the jruby-metrics service tracks.

As a minor point, the jruby-metrics service currently does not have a tidy counter for tracking the number of "active requests".  You could probably capture that by summing a count of `requested-instances` and `borrowed-instances` from the metrics registry together.  It might be a bit less expensive, though, to just have a running active request counter local to the `wrap-with-concurrent-request-limit` which is checked.  It seems like the logic wouldn't have to be too involved to do that.  I'm thinking of something like this:

```clojure
(defn wrap-with-concurrent-request-limit
  [handler metrics-service request-limit splay-limit]
  (let [request-count (atom 0)]
    (fn [request]
      (swap! request-count inc)
      (try
        ...
        (finally
          (swap! request-count dec))))))
```

For now, I'd probably still lean toward doing something like the above for now to avoid having to tie this functionality to the presence / enablement of the metrics service - and the associated error handling that we'd have to write around that as a result.

Would be interested to get another opinion on this, though.  Maybe @jpinsonault or @rlinehan could weigh in here.

Sharpie:
I took a look at the QoS filter and it does provide URL matching and priority, but didn't look like it supported configuring a 503 responses with randomized Retry-After headers. So, we'd probably have to extend the implementation in order to get something that would be useful for mitigating thundering herds.

The other issue I noticed is that the QoS filter assumes a ServletRequest is being used --- especially for the suspend and resume functionality. So, we'd have to make sure filters would play nice with whatever `ring.util.servlet` is returning and could be passed to the tk-jetty `ring-handler`.

In comparison, a Ring handler was much more straight-forward to implement.

camlow325:
Assuming we decide to go with something like this, I do think it would be good to get some "UX" opinions as to what some good names for the additional `jruby-puppet` configuration settings would be.  Specifically:

* `max-active-requests`- Maximum number of requests which use JRuby instances can be active at one time.  For any additional requests which exceed the threshold, an HTTP 503 response would be returned.
* `splay-limit` - When an HTTP 503 response is returned for a request due to the `max-active-requests` limit being reached, a `Retry-After` header is provided in the response.  The value for the `Retry-After` header is computed based on a randomized value less than or equal to the value for `splay-limit`.  A client can use this value to decide the amount of time it will wait before retrying the failed request.

`max-active-requests` seems like a reasonable name to me.  There may be a more obvious name for `splay-limit`, not sure.  @nfagerlund - do you have any opinions on this?

camlow325:
> I took a look at the QoS filter and it does provide URL matching and priority, but didn't look like it supported configuring a 503 responses with randomized Retry-After headers. So, we'd probably have to extend the implementation in order to get something that would be useful for mitigating thundering herds.

Yeah, I don't think it has anything for that.  I suppose that wouldn't preclude us from putting some more generic behavior in the agent to randomize retry intervals even in the absence of any formal value being returned from the server.  I agree, though, that we'd probably need to look into extending the default behavior if we want to have the server make a recommendation.

> The other issue I noticed is that the QoS filter assumes a ServletRequest is being used --- especially for the suspend and resume functionality. So, we'd have to make sure filters would play nice with whatever ring.util.servlet is returning and could be passed to the tk-jetty ring-handler.  In comparison, a Ring handler was much more straight-forward to implement.

Yeah, the issue of how the ServletFilter would apply in the context of a Ring handler, especially with respect to the suspend/resume functionality, would probably be the most involved piece of this - and I agree that your Ring middleware-based approach is simpler.

Sharpie:
> I suppose that wouldn't preclude us from putting some more generic behavior in the agent to randomize retry intervals even in the absence of any formal value being returned from the server.

One wrinkle here is that the `Puppet::Network::Http` library is used for more than just requests to the Puppet Server REST API. So, having the agent re-splay its self for any 503 response could be unexpected if some provider code is making a request to a 3rd-party service. Having the agent respect the `Retry-After` header on a 503 seems better as it matches the standards around HTTP --- but this pushes the splay duration decision up to Puppet Server.

src/clj/puppetlabs/services/request_handler/request_handler_service.clj
jpinsonault: I don't have a strong opinion yet, but I have a couple questions. I'm curious why the metrics service isn't a TK dependency here, since that would take care of `metrics-service` potentially being nil.

I don't think it would be terrible to couple the metrics-service to this code, but i'd lean towards @jb's idea if right now all we care about is the number we could produce locally without having to bring in the metrics service.

camlow325: Yeah, I think we could just make `:JRubyMetricsService` a required dependency to avoid the need to check for `nil`.  I'd been holding onto the thought, though, that it might still be useful -- I guess just if memory usage / perf to collect measurements were too excessive for some users? -- to have a way to disable JRuby metrics functionality by removing the `jruby-metrics-service` from the bootstrap config.  But maybe we're far enough down the road on this now with evidence of this not having needed to be done for any PE users that it's not worth preserving that capability.

camlow325: Hmm, actually I guess the `:JRubyMetricsService` essentially was already required via the `JRubyPuppetService` for the setup work we did around the http-client metrics in [this PR](https://github.com/puppetlabs/puppetserver/pull/1322).  I think it's probably fine to just go with making `:JRubyMetricsService` a required dependency, allowing us to do away with the check for `nil` handling as @jpinsonault had suggested.  I'm also fine with just trying to mine the active request count out of the metrics service info.

Sharpie: > Hmm, actually I guess the :JRubyMetricsService essentially was already required via the JRubyPuppetService for the setup work we did around the http-client metrics 

Ah. If the JRubyMetricsService is already a hard dependency of JRubyPuppetService then that makes it much simpler as there isn't a way for folks to leave it out of bootstrap.cfg and still have a bootable Puppet Server.

> One other thing to consider is that not every use of a JRuby instance involves going through the request-handler-service. 

I thought about this a bit as well and I think the handler separation is a good thing at the moment --- as stampeding puppet agents are the problem we want to target and rejecting other traffic such as console or API requests would be a problem. It might be worth enforcing this by adding a check for the Puppet `User-Agent` header.

camlow325: > I thought about this a bit as well and I think the handler separation is a good thing at the moment --- as stampeding puppet agents are the problem we want to target and rejecting other traffic such as console or API requests would be a problem. It might be worth enforcing this by adding a check for the Puppet User-Agent header.

Yeah, I suppose with the QoSFilter-style approach, we could prioritize those ahead of the other agent-based ones to make it much less likely that those would be dropped but what you have for now in lieu of that probably makes sense.  I wouldn't object to an extra `User-Agent` header check here if you want to add that but not sure it's essential at this point.  Maybe simpler is better for now.

Sharpie: > Yeah, I suppose with the QoSFilter-style approach, we could prioritize those ahead of the other agent-based ones to make it much less likely that those would be dropped 

The problem here is that prioritization alone doesn't solve the issue of the thundering herd. In order to break a herd up, we need to tell agents to apply a randomized backoff so that their runtimes de-sync. Suspending requests but otherwise processing them as fast as possible may maintain quality of service for routes like `/status` an `/environment_classes` but doesn't seem like it will be able to break a herd up.

camlow325: > The problem here is that prioritization alone doesn't solve the issue of the thundering herd.  In order to break a herd up, we need to tell agents to apply a randomized backoff so that their runtimes de-sync.

Sure, I wasn't questioning that aspect of this.  I think you're right about that.  Was just thinking if we had the ability to suspend requests with prioritization that we could prioritize ones like `/environment_classes` and `/status` ahead of ones that are typically agent-driven like `../catalog` - hopefully making it a lot less likely that we'd ever need to drop the higher priority ones.  If we end up pursuing a QoSFilter approach for prioritization later, I think it would be good to consider how to factor in a randomized splay for requests rejected with a 503 as your PR does.

Sharpie:
I've updated this PR to use an explicit dependency on the `:JRubyMetricsService` along with using the size of the `:requested-instances` map as the metric for determining when `max-active-requests` will be exceeded.

Now I'm looking into testing and I'm trying to find the best place to add a testcase. So far test/integration/puppetlabs/services/jruby/jruby_metrics_service_test.clj looks like the best spot as it has all the mocking for performing concurrent requests againt a stubbed JRuby pool, but I'm wondering if there is a better place to put it.

nfagerlund:
Somehow I missed this when I was first pinged on it. Sorry @camlow325!

* I agree, `max-active-requests` seems fine. 
* For the other one, maybe `max-retry-delay`?

camlow325:
@Sharpie Using the `coordinated-mock-jruby-instance` and related helpers from the `jruby-metrics-service-test` probably would be pretty helpful for these kinds of tests.  I'd be okay with moving the test helpers into a more shareable lower-level test utils namespace, like `jruby_puppet_testutils` so that you can write the new tests in a namespace that is more directly related to the code you're changing, like `request_handler_test` maybe, if you think that makes sense.

src/clj/puppetlabs/puppetserver/jruby_request.clj
```diff
@@ -45,6 +47,29 @@
...
+  If splay-limit is set, the response will include a Retry-After header set
+  to a random fraction of the limit."
+  [handler :- IFn
+   metrics-service :- (schema/protocol jruby-metrics/JRubyMetricsService)
+   request-limit :- schema/Int
+   splay-limit :- schema/Int]
+  ;; TODO: Log warnings when limit is hit. Add i18n to user-facing strings.
+  (fn [request]
+    (let [metrics-map (jruby-metrics/get-metrics metrics-service)
+          requested-count (-> metrics-map :requested-instances deref count)]
```
camlow325: Do you think the `max-active-requests` limit should be checked against the sum of `requested-instances` and `borrowed-instances`, the latter of which could range up to whatever value is configured for `max-active-instances` in the pool, or just the `requested-instances`?  I think I'm good either way, just curious if you had thought about it.

camlow325:
No rush on resolving these, but just noticed that Travis had two failures for this PR:

```
lein test puppetlabs.services.request-handler.request-handler-core-test

2017-04-26 18:56:11,315 ERROR [main] [p.t.internal] Error during app buildup!
java.lang.RuntimeException: Service ':JRubyMetricsService' not found
	at puppetlabs.trapperkeeper.internal$handle_prismatic_exception_BANG_.invokeStatic(internal.clj:121)
	at puppetlabs.trapperkeeper.internal$handle_prismatic_exception_BANG_.invoke(internal.clj:111)
	at puppetlabs.trapperkeeper.internal$instantiate$fn__13983.invoke(internal.clj:162)
	at puppetlabs.trapperkeeper.internal$instantiate.invokeStatic(internal.clj:159)
	at puppetlabs.trapperkeeper.internal$instantiate.invoke(internal.clj:152)
	at puppetlabs.trapperkeeper.internal$eval14517$build_app_STAR___14526$fn__14527.invoke(internal.clj:569)
	at puppetlabs.trapperkeeper.internal$eval14517$build_app_STAR___14526.invoke(internal.clj:537)
	at puppetlabs.trapperkeeper.internal$eval14619$boot_services_STAR___14628$fn__14629$fn__14630.invoke(internal.clj:639)
	at puppetlabs.trapperkeeper.internal$eval14619$boot_services_STAR___14628$fn__14629.invoke(internal.clj:638)
	at puppetlabs.trapperkeeper.internal$eval14619$boot_services_STAR___14628.invoke(internal.clj:633)
	at puppetlabs.trapperkeeper.core$boot_services_with_config_fn.invokeStatic(core.clj:77)
	at puppetlabs.trapperkeeper.core$boot_services_with_config_fn.invoke(core.clj:61)
	at puppetlabs.trapperkeeper.core$boot_services_with_config.invokeStatic(core.clj:93)
	at puppetlabs.trapperkeeper.core$boot_services_with_config.invoke(core.clj:79)
	at puppetlabs.trapperkeeper.testutils.bootstrap$bootstrap_services_with_config.invokeStatic(bootstrap.clj:16)
	at puppetlabs.trapperkeeper.testutils.bootstrap$bootstrap_services_with_config.invoke(bootstrap.clj:13)
	at puppetlabs.services.request_handler.request_handler_core_test$fn__53462$fn__53501.invoke(request_handler_core_test.clj:330)
	at clojure.core$with_redefs_fn.invokeStatic(core.clj:7216)
	at clojure.core$with_redefs_fn.invoke(core.clj:7200)
	at puppetlabs.services.request_handler.request_handler_core_test$fn__53462.invokeStatic(request_handler_core_test.clj:330)
	at puppetlabs.services.request_handler.request_handler_core_test$fn__53462.invoke(request_handler_core_test.clj:287)

lein test :only puppetlabs.services.request-handler.request-handler-core-test/test-jruby-pool-not-full-during-code-id-generation

ERROR in (test-jruby-pool-not-full-during-code-id-generation) (internal.clj:121)
Uncaught exception, not in assertion.
expected: nil

  actual: java.lang.RuntimeException: Service ':JRubyMetricsService' not found
 at puppetlabs.trapperkeeper.internal$handle_prismatic_exception_BANG_.invokeStatic (internal.clj:121)
    puppetlabs.trapperkeeper.internal$handle_prismatic_exception_BANG_.invoke (internal.clj:111)
    puppetlabs.trapperkeeper.internal$instantiate$fn__13983.invoke (internal.clj:162)
    puppetlabs.trapperkeeper.internal$instantiate.invokeStatic (internal.clj:159)
    puppetlabs.trapperkeeper.internal$instantiate.invoke (internal.clj:152)
    puppetlabs.trapperkeeper.internal$eval14517$build_app_STAR___14526$fn__14527.invoke (internal.clj:569)
    puppetlabs.trapperkeeper.internal$eval14517$build_app_STAR___14526.invoke (internal.clj:537)
 puppetlabs.trapperkeeper.internal$eval14619$boot_services_STAR___14628$fn__14629$fn__14630.invoke (internal.clj:639)
    puppetlabs.trapperkeeper.internal$eval14619$boot_services_STAR___14628$fn__14629.invoke (internal.clj:638)
    puppetlabs.trapperkeeper.internal$eval14619$boot_services_STAR___14628.invoke (internal.clj:633)
    puppetlabs.trapperkeeper.core$boot_services_with_config_fn.invokeStatic (core.clj:77)
    puppetlabs.trapperkeeper.core$boot_services_with_config_fn.invoke (core.clj:61)
    puppetlabs.trapperkeeper.core$boot_services_with_config.invokeStatic (core.clj:93)
    puppetlabs.trapperkeeper.core$boot_services_with_config.invoke (core.clj:79)
    puppetlabs.trapperkeeper.testutils.bootstrap$bootstrap_services_with_config.invokeStatic (bootstrap.clj:16)
    puppetlabs.trapperkeeper.testutils.bootstrap$bootstrap_services_with_config.invoke (bootstrap.clj:13)
    puppetlabs.services.request_handler.request_handler_core_test$fn__53462$fn__53501.invoke (request_handler_core_test.clj:330)
    clojure.core$with_redefs_fn.invokeStatic (core.clj:7216)
    clojure.core$with_redefs_fn.invoke (core.clj:7200)
    puppetlabs.services.request_handler.request_handler_core_test$fn__53462.invokeStatic (request_handler_core_test.clj:330)
    puppetlabs.services.request_handler.request_handler_core_test/fn (request_handler_core_test.clj:287)
```

Sharpie:
@camlow325 Hmm. Looks like the `:JRubyMetricsService` dependency isn't actually a dependency of `:JRubyPuppetService` yet as there are a few tests that don't include it in their service bootstrap lists. Guess I'll clean that up first and then swing back to factoring out a test for the 503 response.

camlow325:
@Sharpie I haven't looked back at this PR recently but it appears it is passing Travis now.  Do you consider it to be complete to review for merging at this point or did you still have more work that you wanted to do for it?

camlow325:
@Sharpie Ping.  Are you still hoping for this to get in ahead of Puppet 5 (for which we need to have all new code merged in by mid next week)?  Do you think it's ready to review for merge at this point?

Sharpie:
At this point I'm still blocked on getting a test in --- I could probably get it done pretty quickly if I just tacked it onto the existing `jruby-metrics-service-test` suite instead of trying to also factor the helper code out into a separate library.

src/clj/puppetlabs/puppetserver/jruby_request.clj
Sharpie: I think the important bit is the size of the "request backlog" which is captured by `requested-instances`. Also, it's simpler to pull one metric compared to two.

Sharpie:
@camlow325 Alright, finally got around to polishing this one up:

  - Settings have been renamed to `max-queued-requests` and `max-retry-delay`.

  - A test has been added to `puppetlabs.services.jruby.jruby-metrics-service-test` that asserts a 503 is returned when `max-queued-requests` is exceeded.

  - Docs and proper error logging have been added.

src/clj/puppetlabs/puppetserver/jruby_request.clj
```diff
@@ -45,6 +47,28 @@
...
+   metrics-service :- (schema/protocol jruby-metrics/JRubyMetricsService)
+   max-queued-requests :- schema/Int
+   max-retry-delay :- schema/Int]
+  (fn [request]
+    (let [metrics-map (jruby-metrics/get-metrics metrics-service)
+          requested-count (-> metrics-map :requested-instances deref count)]
+      (if (>= requested-count max-queued-requests)
+        (let [err-msg (i18n/trs "The number of requests waiting for a JRuby instance has exceeded max-queued-requests.")
+              response (output-error request {:msg err-msg} 503)]
+          (if (> max-retry-delay 0)
```
camlow325: Not a big deal, but I think `(pos? max-retry-delay)` would work here too.

src/clj/puppetlabs/services/request_handler/request_handler_service.clj
```diff
@@ -22,12 +33,14 @@
           (log/warn (format "%s %s"
                             (i18n/trs "The ''master.allow-header-cert-info'' setting is deprecated and will be ignored in favor of the ''authorization.allow-header-cert-info'' setting because the ''jruby-puppet.use-legacy-auth-conf'' setting is ''false''.")
                             (i18n/trs "Remove the ''master.allow-header-cert-info'' setting.")))))
-      (assoc context :request-handler
-                     (request-handler-core/build-request-handler
-                      jruby-service
-                      (request-handler-core/config->request-handler-settings
-                       config)
-                      current-code-id))))
+      (assoc context :request-handler (if (> max-queued-requests 0)
```
camlow325: Could use `pos?` here too, I think.

Sharpie:
Replaced comparison operators with `pos?`.

Sharpie:
🎉

PR #1370, opened 2017-04-17
(SERVER-1765) Drop support for jdk7

adrienthebo:


.travis.yml
```diff
@@ -1,8 +1,9 @@
 language: clojure
 lein: 2.7.1
+dist: trusty
+sudo: false
```
camlow325: Why do we have both `false` and `required` (below) for `sudo` here?  Would it work to just keep the `sudo` that we already have, `required`?  That appears to be working for our jdk8-only Travis testing in pe-puppet-server-extensions.

adrienthebo: I didn't see that earlier when I was testing; the `sudo` change was done so that we could use docker containers for faster testing time. Since this key was defined twice I suspect the latter value overwrote the former. I'm going to drop the `sudo: required` bit and see if CI passes.

camlow325:
I suppose we should wait to merge this until around the same time the corresponding [ci-job-configs PR](https://github.com/puppetlabs/ci-job-configs/pull/2730) is ready to be merged?

camlow325:
Corresponding ci-job-configs PR has now been merged so I'm going to merge this one too.

PR #1371, opened 2017-04-20
(SERVER-1251) remove resource types endpoint

rlinehan:
The `/puppet/v3/resource_types` endpoint was deprecated in Server 2.3.0 and is being
removed in Server 5. This commit removes it from the route tree, and removes
associated tests.

Also update the `PuppetResource` test schema which was out of date.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -476,8 +476,6 @@ module PuppetServerExtensions
       http.puppet-v3-file_metadatas-.*.-percentage
       http.puppet-v3-node-.*.-percentage
       http.puppet-v3-report-.*.-percentage
-      http.puppet-v3-resource_type-.*.-percentage
-      http.puppet-v3-resource_types-.*.-percentage
       http.puppet-v3-static_file_content-.*.-percentage
       http.puppet-v3-status-.*.-percentage
       num-cpus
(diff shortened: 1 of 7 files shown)


rlinehan:
We should hold off on merging this until https://github.com/puppetlabs/puppet/pull/5813 goes in and we include the SHA from that, since that removes an acceptance test that would otherwise fail.

Also, I wasn't sure what exactly to do with regards to references to the `resource_types` endpoint in our docs - should we remove them? Keep them in? Most of our doc on the `environment_classes` endpoint is written in regards to its differences from the `resource_types` endpoint.

camlow325:
> Also, I wasn't sure what exactly to do with regards to references to the resource_types endpoint in our docs - should we remove them? Keep them in? Most of our doc on the environment_classes endpoint is written in regards to its differences from the resource_types endpoint.

Hmm, good question.  Off the cuff, I'm inclined to leave in the references to `resource_types` in the `environment_classes` docs in place for Puppet/Server 5 to help orient any users migrating up from Puppet 4 / Puppet Server 2 who might have relied upon it.  Maybe dropping the references to `resource_types` in the docs altogether in the Puppet/Server 6 timeframe might make sense?  I suppose that even if we leave `resource_types` related references in place that we might need to make some minor updates to the version references?

@gguillotte - what do you think about this?

rlinehan:
Rebased on master and updated - there's a new submodule pin so this should be good to go now.

rlinehan:
@adrienthebo this has passed in travis and I think it's good to go - can you review?

PR #1374, opened 2017-04-24
(MAINT) Bump beaker version to 3.15

camlow325:
This commit bumps the beaker version used in the Gemfile from 3.5 to
3.15.  This is being done in order to pick up the updated platform code
name -> version matrix used in various beaker helpers, needed for use
with running tests for yakkety.


adrienthebo:
Merging this PR since a beaker version bump is pretty small.

PR #1375, opened 2017-04-24
(maint) drop -XX:MaxPermSize

adrienthebo:
MaxPermSize was made a noop JVM argument in JDK8; since we no longer
support JDK7 we can omit this option entirely.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -102,7 +102,7 @@
   :lein-ezbake {:vars {:user "puppet"
                        :group "puppet"
                        :build-type "foss"
-                       :java-args ~(str "-Xms2g -Xmx2g -XX:MaxPermSize=256m "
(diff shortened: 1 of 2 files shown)


project.clj
```diff
@@ -102,7 +102,7 @@
   :lein-ezbake {:vars {:user "puppet"
                        :group "puppet"
                        :build-type "foss"
-                       :java-args ~(str "-Xms2g -Xmx2g -XX:MaxPermSize=256m "
+                       :java-args ~(str "-Xms2g -Xmx2g "
```
camlow325: Makes sense.  Should we also drop it from the higher-level `:jvm-opts` key as well - so we don't have to see the 'perm size not supported' warnings when running from source anymore?

adrienthebo: Yep, I've done that and dropped references to this setting in the documentation as well.

PR #1377, opened 2017-04-24
(SERVER-1788) Use different graphite port in collect_default_metrics

camlow325:
Previously, the collect_default_metrics test would, via the grafanadash
module, try to setup a graphite webserver on port 80.  When run in the
PE Puppet Server Extensions pipeline, this test would fail because
Apache would be unable to listen on port 80 since the PE console already
would have claimed the port.

This commit changes the test to have the graphite web server hosted on
port 50000 in order to avoid the conflict with the PE console.


camlow325:
I tested this both against an OSS and a PE monolithic master install and confirmed that the test passes for both.

jpinsonault:
merging, ignoring travis since this is a beaker code change

PR #1380, opened 2017-04-25
(maint) Don't run puppetdb acceptance tests on yakkety

rlinehan:
Update `puppetdb_supported_platforms` acceptance test helper to exclude
yakkety from the platforms we run PuppetDB smoke tests agains, since PuppetDB
doesn't support yakkety.


rlinehan:
I haven't actually tried this out because my wifi + vpn situation right now is really flaky.

rlinehan:
Ran locally against 1604 and 1610 - both passed.

PR #1382, opened 2017-04-26
(MAINT) Bump puppet and puppet-agent versions 26-apr-2017

adrienthebo:


acceptance/lib/helper.rb
```diff
@@ -34,7 +34,7 @@ def self.initialize_config(options)
     puppet_build_version = get_option_value(options[:puppet_build_version],
                          nil, "Puppet Agent Development Build Version",
                          "PUPPET_BUILD_VERSION",
-                         "b04c9f4a8d4eace424a718073315a7da0a34f3b5",
+                         "2aa05621ebb25ce724760ab4e65425b8a55f3f60",
```
adrienthebo: This is supposed to be the puppet-agent commit matching `1.10.0.337.g2aa0562` right?

camlow325:
Looks like we had a rubygems / network failure related to the spec tests on one of the Travis cells:

```
Connection reset by peer (https://api.rubygems.org/gems/bundler-1.14.6.gem)
```

I'll reload that and see if it recurs.

acceptance/lib/helper.rb
camlow325: Yep, that's what we've been doing.  Looks right to me.

rlinehan:
Looks like travis passed? That seems fast. But I'm 👍

PR #1383, opened 2017-04-28
(maint) install jdk8 for debian 8 for puppet3 compat

adrienthebo:
This is kinda gross because it's copy/paste, but I've done many worse things.

/cc @camlow325

### DIFF
--- a/acceptance/suites/pre_suite/puppet3_compat/70_install_puppet.rb
+++ b/acceptance/suites/pre_suite/puppet3_compat/70_install_puppet.rb
@@ -9,13 +9,5 @@ end
 
 install_puppet_on(nonmaster_agents, {:version => puppet_version})
 
-variant = master['platform'].variant
-version = master['platform'].version
-if variant == 'debian' && version == "8"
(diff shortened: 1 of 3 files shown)


acceptance/suites/pre_suite/puppet3_compat/70_install_puppet.rb
```diff
@@ -9,5 +9,13 @@
 
 install_puppet_on(nonmaster_agents, {:version => puppet_version})
 
+variant = master['platform'].variant
```
camlow325: We could also put this in a helper in [helper.rb](https://github.com/puppetlabs/puppetserver/blob/ff0e5e84860ea9b035839e77e9b27d4c3a923e36/acceptance/lib/helper.rb) and have the different pre-suites call into it.  The helper.rb is included in each of the beaker command lines for our different suites.  WDYT?

adrienthebo: This is a lot better. I've updated the code, lemme know if you like the new approach.

camlow325:
This looks simple enough to me so I'm going to merge it.

PR #1384, opened 2017-05-02
(SERVER-1762) Use json in tests

jpinsonault:
Many of the puppetserver tests used pson as the content type in accept headers.
This commit changes them to json.

Some tests specifically test that using pson (and json) in an accept header will
work, and these tests have been left in place

### DIFF
--- a/test/integration/puppetlabs/puppetserver/auth_conf_test.clj
+++ b/test/integration/puppetlabs/puppetserver/auth_conf_test.clj
@@ -208,7 +208,7 @@
           http-get-no-ssl (fn [path]
                             (http-client/get
                              (str "http://localhost:8080/" path)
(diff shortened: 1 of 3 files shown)


joshcooper:
Since puppetserver needs to continue supporting 4.x agents, do we want to test that the server accepts both pson and json, where json will be the default in puppet 5 (sometimes between now and when we ship)?

rlinehan:
We have a ticket in to set up a pipeline for Puppet 4 compatibility (similar to what we have for the Puppet 3 compatibility tests). I kind of feel like having that kind of smoke test would be more beneficial than trying to keep both pson and json in all of our unit/clojure tests (mostly because I don't feel like keeping pson in those would be sufficient to ensure that there is compatibility there).

PR #1385, opened 2017-05-04
(maint) Update packaged logback for http client logging

rlinehan:
org.apache.http logs a ton of unhelpful things at DEBUG level that
should really be at TRACE. This commit updates the packaged logback.xml so tat
if the general log level is changed to `debug`, http client will remain at
`info` so as to not spam the logs.


rlinehan:
rebased on master, so this should be good to go now.

PR #1386, opened 2017-05-04
(SERVER-1800) Remove openjdk-7-jre-headless as debian build dependency

camlow325:
Previously, both openjdk-7-jre-headless and openjdk-8-jre-headless were
listed as build dependencies for Debian.  Since we'll no longer need to
be building for JDK 7 for Puppet 5+, this commit removes the
openjdk-7-jre-headless dependency.

### DIFF
--- a/resources/ext/ezbake.conf
+++ b/resources/ext/ezbake.conf
@@ -36,7 +36,7 @@ ezbake: {
              }
 
       debian: { dependencies: ["puppet-agent (>= 1.6.0)"],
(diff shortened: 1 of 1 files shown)


camlow325:
There's some work needed to enable packaging jobs for Jessie to work for openjdk8, tracked in [RE-8816](https://tickets.puppetlabs.com/browse/RE-8816).  We should probably wait until that gets sorted out before merging this PR.

camlow325:
@shrug said the work to install JDK 8 build dependencies during ezbaking for Jessie should all be in place now so I think we can merge this whenever.

PR #1387, opened 2017-05-05
Remove openjdk7 build dep

shrug:


camlow325:
@shrug fyi, I already submitted a PR for the same thing as #1386.  We were thinking we should wait on merging it until we were ready to use openjdk8 on the Jessie builders.  Should that all be ready now?

shrug:
Hah, nice! Yes, the builder should have jessie-backports now, I was able to run some successful builds

PR #1388, opened 2017-05-05
(SERVER-1630) Exclude jruby-deps only for ezbake

camlow325:
Previously, Puppet Server's JRuby dependencies were excluded from
jruby-utils and moved into profiles only (provided and jruby9k) in order
to avoid having the default dependencies from jruby-utils be included in
the puppet-server-release uberjar for ezbake packages.  This, however,
would be problematic for any consumer of the immediate puppetserver
artifact since those consumers would have to explicitly include/manage
the JRuby dependencies themselves.

In this commit, the jruby-deps for JRuby 1.7 are re-added as a standard
dependency but excluded in the ezbake profile.  This should allow the
standard puppetserver artifact to have reasonable JRuby 1.7 dependencies
by default again while still preserving the ability during an ezbake
build to separate JRuby dependencies for 1.7 and 9k into separate
uberjars from the puppet-server-release.jar.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -70,6 +70,8 @@
                                 org.jruby/jruby-stdlib
                                 com.github.jnr/jffi
                                 com.github.jnr/jnr-x86asm]]
+                 [puppetlabs/jruby-deps ~jruby-1_7-version]
+
                  [puppetlabs/trapperkeeper]
                  [puppetlabs/trapperkeeper-authorization]
                  [puppetlabs/trapperkeeper-comidi-metrics]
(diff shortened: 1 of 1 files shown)


camlow325:
I did an ezbake build of this at http://builds.puppetlabs.lan/puppetserver/5.0.0.master.SNAPSHOT.2017.05.05T1134/.  I opened up the puppet-server-release.jar and compared it to one built from the code currently on the master branch and the two were basically identical other than that the builds currently on master were including the immediate jruby-deps artifact without including the upstream jruby-core, jruby-stdlib, or other JRuby dependencies.  From an ezbake perspective, then, this seems like an improvement anyway in that the jruby-deps would no longer be part of the puppet-server-release.jar at all.

camlow325:
I also validated with this one that where a PuppetDB unit test job which [was failing because of the missing jruby dependencies](https://cinext-jenkinsmaster-enterprise-prod-1.delivery.puppetlabs.net/job/enterprise_puppetdb_unit-clj-puppetdb_master/33/) that the failure, which is reproducible locally against the current puppetserver master branch, is no longer reproducible with the changes on this PR (because the JRuby 1.7 dependencies are now being pulled in again.  fyi, @rbrw.

PR #1390, opened 2017-05-08
(SERVER-1630) Remove jruby exclusions from jruby-utils dep

camlow325:
Previously, for the jruby-utils dependency, explicit exclusions were
included for various jruby-related dependencies - jruby-core,
jruby-stdlib, jffi, and jnr.  jruby-utils itself, though, no longer has
any explicit dependencies for any of these.  Instead, the jruby-deps
dependency controls which/whether any of the jruby-related dependencies
are pulled in.  This commit removes the exclusions for these
dependencies from the jruby-utils dependency since they are no longer
needed.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -65,11 +65,7 @@
                  ;; in different versions of the three different logback artifacts
                  [net.logstash.logback/logstash-logback-encoder]
 
-                 [puppetlabs/jruby-utils "0.9.0"
-                   :exclusions [org.jruby/jruby-core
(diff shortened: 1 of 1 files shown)


camlow325:
I did an ezbake build to confirm that no changes to the puppet-server-release.jar file resulted from the changes on this PR vs. what is currently on master.  Also, Clojure tests passed locally for me under both JRuby 1.7 and 9k.

PR #1392, opened 2017-05-09
(SERVER-1775) Add support for running JRuby 9k in CI

camlow325:
This commit contains two changes for use in running JRuby 9k in CI:

1) Add Jenkins scripts for running Clojure unit and JRuby spec tests
   under 9k.

2) Add a beaker pre-suite step for configuring the Puppet Server service
   to run with JRuby 9k based on the configuration of a new environment
   variable, USE_JRUBY_9K.  A value of "y" or "Y" for the variable would
   enable 9K whereas any other value would lead to the default JRuby at
   installation, 1.7.x, being used.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -194,6 +194,7 @@ module PuppetServerExtensions
     else
       abort("Invalid install type: " + test_config[:puppetserver_install_type])
     end
+    maybe_configure_jruby9k
   end
 
   def install_puppet_server_deps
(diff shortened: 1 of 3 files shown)


camlow325:
We'll need changes in a linked ci-job-configs PR, https://github.com/puppetlabs/ci-job-configs/pull/2846, to land before this would be exercised in CI.  This PR should go in first, though.

jpinsonault:
👍

PR #1394, opened 2017-05-10
(maint) Clean up acceptance puppet repo setup

adrienthebo:
The puppetserver and puppet agent setup for acceptance was starting to pick up a lot of organic growth; this adds some cleanup to those files.

### DIFF
--- a/acceptance/suites/pre_suite/foss/30_install_dev_repos.rb
+++ b/acceptance/suites/pre_suite/foss/30_install_dev_repos.rb
@@ -1,23 +1,19 @@
-install_opts = options.merge( { :dev_builds_repos => ["PC1"] })
+if test_config[:puppetserver_install_type] == :package
+  package_build_version = ENV['PACKAGE_BUILD_VERSION']
+  if package_build_version.nil?
+    abort("Environment variable PACKAGE_BUILD_VERSION required for package installs!")
+  end
 
-case test_config[:puppetserver_install_type]
-when :package
-  step "Setup Puppet Server repositories." do
-    package_build_version = ENV['PACKAGE_BUILD_VERSION']
-    if package_build_version
-      install_puppetlabs_dev_repo(master, 'puppetserver', package_build_version,
-                                  nil, install_opts)
-    else
-      abort("Environment variable PACKAGE_BUILD_VERSION required for package installs!")
-    end
+  step "Setup Puppet Server dev repository on the master." do
+    install_opts = options.merge({ :dev_builds_repos => ["PC1"]})
+    install_puppetlabs_dev_repo(master, 'puppetserver', package_build_version, nil, install_opts)
   end
 end
 
 confine_block :except, :platform => ['windows'] do
   step "Setup Puppet Labs Dev Repositories." do
     hosts.each do |host|
-      install_puppetlabs_dev_repo(host, 'puppet-agent', test_config[:puppet_build_version],
-                                    nil, install_opts)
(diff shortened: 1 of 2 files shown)


acceptance/suites/pre_suite/foss/30_install_dev_repos.rb
```diff
@@ -1,27 +1,19 @@
...
-  confine_block :except, :platform => ['windows'] do
-    step "Setup Puppet Labs Dev Repositories." do
-      hosts.each do |host|
-        install_puppetlabs_dev_repo host, 'puppet-agent', puppet_build_version,
-                                    repo_config_dir, install_opts
-      end
+confine_block :except, :platform => ['windows'] do
+  step "Setup Puppet Labs Dev Repositories." do
+    hosts.each do |host|
+      install_puppet_agent_dev_repo_on(host, {:puppet_agent_version => test_config[:puppet_build_version]})
```
camlow325: Before it looked like `:dev_builds_repos => ["PC1"]` would have been passed in here along with the options but not now.  Will this automatically install from PC1 then?

camlow325: This effectively makes `puppet_build_version` required to be non-empty, right?  I can't remember if there was ever a reason / scenario where we wanted to have it empty before, e.g., with no pin - like to just use whatever agent is installed along with the puppetserver package on the master.  @rlinehan - does this ring a bell for you at all?  Drawing blanks...

adrienthebo: This will automatically pull from PC1, yes. Looking at https://github.com/puppetlabs/puppetserver/blob/master/acceptance/lib/helper.rb#L34-L38 indicates that `puppet_build_version` will always be set, and I think that's a reasonable assumption to make. Why not always have a sane version of the agent installed for FOSS installations?

camlow325: Yeah, it's probably fine this way.  I honestly can't think of a good reason not to always have a version there - even if in the future we need to do something different to derive the "latest good version" from some other source rather than an explicit pin like we do today.

camlow325: Looks like the puppetserver nightlies failed CI on SLES 12 only after this change went in.  Appears to be trying to access the wrong URL to download the rpm which gets installed:

```
Setup Puppet Labs Dev Repositories.
Fetching: http://builds.delivery.puppetlabs.net/puppet-agent/e749206bc4be4c81b604c305ce7550bf36f54524/repos/sles/12/PC1/x86_64/puppet-agent-e749206bc4be4c81b604c305ce7550bf36f54524-1.sles12.x86_64.rpm
```

I think the problem is due to the `install_puppet_agent_dev_repo_on` function falling back to trying to download the rpm directly instead of using the repo_config, as it was prior to that change, and failing.  See [this code](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/install_utils/foss_utils.rb#L1114-L1135).  We might be able to make the .rpm file resolve correctly if we change this to pass in both a `puppet_agent_sha` and `puppet_agent_version`.  Having to maintain both in our helper config would seem onerous, though.  We might also be able to change beaker to use a repo_config instead of an rpm like is done for other EL variants but I'm not sure if that would have broader implications.  Maybe it would be best to just revert this back to using `install_puppetlabs_dev_repo` as it was previously?

PR #1395, opened 2017-05-11
(SERVER-1805) Enabling static code scans of clojure for security issues

jsane:
Enabling code scans for some basic security checks in Clojure code. Currently the script only looks for read-string instances and will be updated for additional patterns as needed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,9 @@ dist: trusty
 jdk:
 - openjdk8
 - oraclejdk8
-script: "./ext/travisci/test.sh"
+script:
+  - "./ext/travisci/test.sh"
+  - $TRAVIS_BUILD_DIR/ext/travisci/secscan.sh "$TRAVIS_BUILD_DIR/src/clj" "clj"
 sudo: required
 notifications:
   email: false
--- a/ext/travisci/secscan.sh
+++ b/ext/travisci/secscan.sh
@@ -1,0 +1,59 @@
+#! /usr/bin/env bash
+
+#set -e
+
+
+# TODO::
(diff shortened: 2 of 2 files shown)


camlow325:
Rather than going with a bash script, have you considered writing a Leiningen plugin - or seeing if you can repurpose an existing plugin which does static analysis - to see if it would do what you need?  If we did this via a plugin, seems like it would be easier to update it across different Clojure projects as needed.

jsane:
My approach was largely due to not being aware of these things; I don't understand the environment so this is primitive. But yes I can research and see if there is a better way to do it. Ok to not merge it as it is not very pressing at this stage.

camlow325:
@jsane Yeah, I'm not sure what the best approach is exactly.  You might look at something like [lein-bikeshed](https://github.com/dakrone/lein-bikeshed) as an example of how a plugin which inspects Clojure code might be written.  Something more like [eastwood](https://github.com/jonase/eastwood) could provide some examples for how to do a lot deeper / more complex analysis of Clojure namespaces, depending upon how thorough you are thinking this might end up being over time.  I was just thinking that a simple plugin might be easier to start with and give you the ability to avoid having to deal with bash scripting while having some support for versioning as you might want to roll this out across different projects.  That said, though, I'm happy to merge this as-is if you don't have time to look into this further at this point.  It appears to work fine with our project currently.

jsane:
If it is working lets enable it temporarily till I figure a more better solution per some of your suggestions. It might take me a bit but I will certainly look into it. thanks,

adrienthebo:
I'm ok with merging this as-is while we work on an improved static code analyzer. This is better than what we have now, solves our immediate concerns, and a more improved version might take a little while.

PR #1396, opened 2017-05-11
(SERVER-1786) Add puppet4 compatibility acceptance tests

adrienthebo:
This pull request refactors the Puppet compatibility pre-suites and tests so that we can share logic between the Puppet 3/4 tests and pre-suites, and pushes the remaining logic into separate suites and test directories where needed. Invoking these tests will require adding two directories apiece to the `--pre-suite` and `--tests` options.

Because this pull request changes the directories used in testing we'll need to update the appropriate CI job configs.


camlow325:
Looks like this will need to be rebased now based on changes from #1401 and #1402 that went in.

acceptance/suites/pre_suite/puppet4_compat/70_install_puppet.rb
```diff
@@ -0,0 +1,11 @@
+require 'puppetserver/acceptance/compat_utils'
+
+step "Install Legacy Puppet Agents."
+
+puppet_agent_version = '1.9.3'
```
camlow325: Do we want to have this hardcoded / use of the environment variable commented out below.  That would make the value not possible to override from ci-job-configs, right?  Not sure if that's a big deal but different than what we've done for this in the past anyway.

acceptance/suites/pre_suite/puppet4_compat/80_configure_agents.rb
```diff
@@ -0,0 +1,13 @@
+require 'puppetserver/acceptance/compat_utils'
```
camlow325: Do we need to have any of these steps be run for a Puppet 4 agent?  I was thinking these were only needed for a Puppet 3 agent.

acceptance/suites/pre_suite/puppet4_compat/70_install_puppet.rb
adrienthebo: Oh drat, that was a test change that I left in. Will fix.

acceptance/suites/pre_suite/puppet4_compat/80_configure_agents.rb
adrienthebo: We're using the `nonmaster_agent` function. It's pretty complex.

```
def nonmaster_agents()
  agents.reject { |agent| agent == master }
end
```

Would you prefer to inline that function?

camlow325: I'd be okay with inlining it if you are.  My point here was I wasn't sure if we needed to do any of the steps in this file for Puppet 4 agents - i.e., could this entire file just be deleted?  I was thinking that with the default values for settings that Puppet 4 agents use that these might be irrelevant - but maybe I'm wrong about that.

rlinehan: Agreed that we should delete this file if we can.

rlinehan:
I think it would be good to check out whether the file Jeremy mentioned can be deleted, otherwise I am good with this.

acceptance/suites/pre_suite/puppet4_compat/70_install_puppet.rb
```diff
@@ -0,0 +1,10 @@
+require 'puppetserver/acceptance/compat_utils'
+
+step "Install Legacy Puppet Agents."
+
+puppet_version = ENV['PUPPET_LEGACY_VERSION']
+if not puppet_version
+  fail "PUPPET_LEGACY_VERSION is not set, e.g. '1.9.3'"
+end
+
+install_puppet_agent_on(nonmaster_agents, {:puppet_agent_version => puppet_agent_version})
```
rlinehan: when I get to here I get

```
  Test Case acceptance/suites/pre_suite/puppet4_compat/70_install_puppet.rb reported: #<NameError: undefined local variable or method `puppet_agent_version' for #<Beaker::TestCase:0x007fad4045c870>
```

I think it needs to be `puppet_version` or the variable at line 5 needs to be renamed to `puppet_agent_version`.

adrienthebo: Oi. Fixed this, and ran the Puppet 3 and Puppet 4 compat tests. We should be good this time.

rlinehan:
Gonna hold off on merging though until https://github.com/puppetlabs/puppetserver/pull/1408 goes through and pipeline is hopefully green.

PR #1404, opened 2017-05-18
(SERVER-1819) don't set node_cache_terminus

MosesMendoza:
This commit stops overriding the default `node_cache_terminus` for puppet
server's master application as `write_only_yaml` and removes the setting of
Puppet::Node's indirection cache_class. With this, Puppet::Node's indirection
never gets its cache class set and won't cache by default. Per PUP-6060, with
the advent of PuppetDB, workflows using lists of nodes are better off querying
it than these yaml files on the master, and since this was a `write_only` cache
we were always querying the node terminus for nodes anyway.

### DIFF
--- a/src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
@@ -35,7 +35,6 @@ class Puppet::Server::PuppetConfig
     master_run_mode = Puppet::Util::RunMode[:master]
     app_defaults = Puppet::Settings.app_defaults_for_run_mode(master_run_mode).
         merge({:name => "master",
-               :node_cache_terminus => :write_only_yaml,
                :facts_terminus => 'yaml'})
     Puppet.settings.initialize_app_defaults(app_defaults)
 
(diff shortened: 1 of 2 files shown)


MosesMendoza:
Relates to/review with https://github.com/puppetlabs/puppet/pull/5889

joshcooper:
nit, I think we want a SERVER ticket for this PR

rlinehan:
I mean, we could have one, but I don't think we care as much as you do about tickets matching up.

joshcooper:
Oh I don't care, but I assume you had jira ticket reconciliation. Also not sure if you care about tracking the fix version associated with this change, as we can only put puppet versions on PUP tickets

MosesMendoza:
No worries! Happy to open a server ticket. Opened SERVER-1819, updated commit message, and linked tickets.

MosesMendoza:
This introduced CI failures, resulting in this PR https://github.com/puppetlabs/puppet/pull/5898 to set the terminus for a specific test

src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
```diff
@@ -49,8 +48,6 @@ def self.initialize_puppet(puppet_config)
     Puppet::FileServing::Metadata.indirection.terminus_class = :file_server
     Puppet::FileBucket::File.indirection.terminus_class = :file
 
-    Puppet::Node.indirection.cache_class = Puppet[:node_cache_terminus]
```
MosesMendoza: Discussing with @joshcooper it would be better to have left this in place - honoring the setting value would make it easier for the user to keep the 4.x behavior if they wanted (as opposed to having to configure this with a `routes.yaml` entry for node. I'm going to open a separate PR.

PR #1405, opened 2017-05-19
(SERVER-1803) Bump jruby deps and resolve org.ow2.asm conflict

camlow325:
This commit resolves a dependency conflict between JRuby and
org.ow2.asm via several related packaging changes.  Several dependencies
are bumped in this commit:

- jruby-deps (for 1.7) - 1.7.26-1 => 1.7.26-2
- jruby-deps (for 9k) - 9.1.8.0-1 => 9.1.9.0-1
- jruby-utils - 0.9.0 => 0.9.1
- clj-parent - 0.8.0 => 1.1.0
- ezbake - 1.3.0 => 1.4.0

The ezbake changes are needed to build the jruby-deps uberjars properly
along with the changes to resolve the org.ow2.asm conflict.

The new clj-parent version brings in the org.ow2.asm version, 5.0.3, which
resolves the dependency conflict.

The tk-webserver-jetty9 pin is removed because the latest version,
2.0.1, can now be picked up from the newer clj-parent version.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,7 +1,6 @@
 (def ps-version "5.0.0-master-SNAPSHOT")
-(def tk-ws-jetty9-version "2.0.1")
-(def jruby-1_7-version "1.7.26-1")
-(def jruby-9k-version "9.1.8.0-1")
+(def jruby-1_7-version "1.7.26-2")
+(def jruby-9k-version "9.1.9.0-1")
 
 (defn deploy-info
   [url]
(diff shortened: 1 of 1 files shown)


camlow325:
This will fail Travis currently because it needs a jruby-utils bump to go along with it.  That would be pending a merge of https://github.com/puppetlabs/jruby-utils/pull/49 and a subsequent jruby-utils release.

camlow325:
I think this should be ready to review for merging now.  It should be able to pass tests in Travis and be ezbakeable since all of the dependency versions it refers to should exist now.

PR #1406, opened 2017-05-22
(SERVER-1819) honor node_cache_terminus setting

MosesMendoza:
38e0a0c stopped setting the node_cache_terminus setting value, and stopped
setting the cache_class to its result. This commit undoes the latter half of
that change - we return to setting the cache_class. This will still default to
nil since we aren't specifying the setting value, which will disable caching,
but it will enable users to easily revert to the 4.x behavior of caching by
specifying :node_cache_terminus => :write_only_yaml. Without this commit users
would have to augment their routes.yaml file with a `:node` key specifying the
cache_class.


camlow325:
Merging since we have a couple of approvals...

PR #1407, opened 2017-05-22
(MAINT) Cleanup scripting container in test-profiler-via-ruby test

camlow325:
This commit reworks the test-profiler-via-ruby test to call .terminate()
when finished using the ScriptingContainer that it creates.  The test
also calls `create-scripting-container` instead of
`empty-scripting-container` to create the container, ensuring that more
of the common initialization logic from jruby-utils is utilized.

### DIFF
--- a/test/unit/puppetlabs/services/puppet_profiler/puppet_profiler_core_test.clj
+++ b/test/unit/puppetlabs/services/puppet_profiler/puppet_profiler_core_test.clj
@@ -61,32 +61,35 @@
               (is (every? #(instance? Timer (.get metrics-map %)) expected-metrics)))))))))
 
 (deftest test-profiler-via-ruby
-  (let [sc      (jruby-internal/empty-scripting-container
+  (let [sc      (jruby-internal/create-scripting-container
                  (jruby-core/initialize-config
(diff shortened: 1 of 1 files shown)


camlow325:
This test has failed for me a couple of times locally when running under JRuby 9k with an error related to an inability to find a class from the jopenssl gem when running the test's Ruby scriptlet.  I haven't been able to reproduce the failure at will, though.  Also, I have yet to see this test fail when run in CI - either for JRuby 1.7 or 9k.

In prior cases where I've seen a failure like this, the problem had to do with the process running out of file descriptors, see [SERVER-858](https://tickets.puppetlabs.com/browse/SERVER-858).  I don't believe that is the case here, though.  I ran the whole test suite, monitoring file descriptor usage with `lsof` as the tests were running.  The count never got above 600, so there did not appear to be a file descriptor leak.  Also, I didn't see any evidence of leaking ScriptingContainers (outside of this one which wasn't explicitly terminated).  I'm inclined not to try to follow up on the JRuby 9k failure further, though, until it becomes more reproducible.

jpinsonault:
👍

PR #1409, opened 2017-05-24
(MAINT) Bump puppet-agent to f9a1b77, puppet submodule to 82e65ec

camlow325:
This commit bumps the puppet-agent package version used for testing to
f9a1b77 and the corresponding Puppet submodule to 82e65ec.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -25,7 +25,7 @@ module PuppetServerExtensions
 
     puppet_version = get_option_value(options[:puppet_version],
                          nil, "Puppet Version", "PUPPET_VERSION",
-                         "4.99.0.31.g17aa7d8",
+                         "4.99.0.125.gf9a1b77",
                          :string) ||
                          get_puppet_version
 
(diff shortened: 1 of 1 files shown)


camlow325:
@rlinehan - I think this one should be good - corresponding puppet-agent [puppet test job](https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/master/view/Suite/job/platform_puppet-agent_intn-van-sys_suite-daily-puppet-master/87/) appeared to be green.

acceptance/lib/helper.rb
```diff
@@ -25,7 +25,7 @@ def self.initialize_config(options)
 
     puppet_version = get_option_value(options[:puppet_version],
                          nil, "Puppet Version", "PUPPET_VERSION",
-                         "4.99.0.31.g17aa7d8",
+                         "4.99.0-125-gf9a1b77",
```
rlinehan: I'm not sure where exactly this is used, but the output when I run `! passing puppet-agent master` gives it with `.` rather than `-`....

camlow325: Ah, I probably should use the dots instead good catch.  I think I got that from running `git describe` on the commit.  Must be some logic which subs hyphens for dots somewhere.  The `SUITE_VERSION` on the puppet-agent job uses the dot form, as does the name of the rpm file on the build server - http://builds.delivery.puppetlabs.net/puppet-agent/f9a1b779d71164f7325bd5cdd63077aa20e4cd29/artifacts/el/7/PC1/x86_64/puppet-agent-4.99.0.125.gf9a1b77-1.el7.x86_64.rpm - so that seems like what we should use.  I'll fix that up.

camlow325: I just modified the existing commit to include the dots instead, now a23a2f3.

PR #1410, opened 2017-05-25
(SERVER-1661) Bump JRuby dep to 1.7.27 and remove compat-version setting

camlow325:
This commit bumps the jruby-deps dependency for 1.7 from 1.7.26-2 to
1.7.27-1 and jruby-utils dep from 0.9.0 to 0.10.0.

This commit removes support for the jruby-puppet.compat-version setting.
The compat-version setting was introduced in Puppet Server 2.7.0 to
allow for JRuby to run under Ruby language version 2.0 instead of the
default version, 1.9.  Setting compat-version to "2.0" for JRuby 1.7.27,
however, was broken due to a regression - see:
https://github.com/jruby/jruby/issues/4613 - and is unlikely to be fixed
in the future.  For Puppet Server 5+, users will be able to, via the
JRUBY_JAR environment variable in packaging, configure JRuby 9k to be
used instead of JRuby 1.7 for Ruby language version 2+ support.

In conjunction with a corresponding jruby-utils bump, this commit causes
the presence of the jruby-puppet.compat-version setting to throw an
error at startup, with a message indicating how users can configure the
use of JRuby 9k for Ruby language version 2+ support.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
@@ -270,9 +270,9 @@
     metrics-service]
    (when (get-in raw-config [:jruby-puppet :compat-version])
      (log/errorf "%s %s %s"
-                 "The jruby-puppet.compat-version setting is no longer supported."
-                 "To use Ruby language version 1.9.3, remove the jruby-puppet.compat-version setting."
-                 "To use a Ruby language version later than 2.0, remove the jruby-puppet.compat-version setting and add the line 'JRUBY_JAR=/opt/puppetlabs/server/apps/puppetserver/jruby-9k.jar' to the /etc/sysconfig/(pe-)puppet(-)server or /etc/default/(pe-)puppet(-)server file.")
+                 (i18n/trs "The jruby-puppet.compat-version setting is no longer supported.")
+                 (i18n/trs "To use Ruby language version 1.9.3, remove the jruby-puppet.compat-version setting.")
+                 (i18n/trs "To use a Ruby language version later than 2.0, remove the jruby-puppet.compat-version setting and add the line 'JRUBY_JAR=/opt/puppetlabs/server/apps/puppetserver/jruby-9k.jar' to the /etc/sysconfig/(pe-)puppet(-)server or /etc/default/(pe-)puppet(-)server file."))
      (throw (IllegalArgumentException.
              (i18n/trs "jruby-puppet.compat-version setting no longer supported"))))
    (let [jruby-puppet-config (initialize-puppet-config
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
@@ -268,6 +268,13 @@
     agent-shutdown-fn :- IFn
     warn-legacy-auth-conf? :- schema/Bool
     metrics-service]
+   (when (get-in raw-config [:jruby-puppet :compat-version])
+     (log/errorf "%s %s %s"
+                 "The jruby-puppet.compat-version setting is no longer supported."
+                 "To use Ruby language version 1.9.3, remove the jruby-puppet.compat-version setting."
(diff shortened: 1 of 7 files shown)


camlow325:
This will fail Travis currently because the jruby-deps 1.7.26-1 and jruby-utils 0.10.0 releases don't exist yet and so shouldn't be merged. Putting it up for review only at this point.

For the jruby-utils bump, https://github.com/puppetlabs/jruby-utils/pull/52 will need to land first and a subsequent release done.

For the jruby-deps bump, the dependency work has been merged, https://github.com/puppetlabs/jruby-deps/pull/6, but we're currently in a holding pattern on doing the release because of build server issues.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -268,6 +268,13 @@
     agent-shutdown-fn :- IFn
     warn-legacy-auth-conf? :- schema/Bool
     metrics-service]
+   (when (get-in raw-config [:jruby-puppet :compat-version])
+     (log/errorf "%s %s %s"
```
rlinehan: I know we're not doing i18n for logs yet, but to future-proof this should we wrap it in and i18n call?

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -268,6 +268,13 @@
...
+                 "The jruby-puppet.compat-version setting is no longer supported."
+                 "To use Ruby language version 1.9.3, remove the jruby-puppet.compat-version setting."
+                 "To use a Ruby language version later than 2.0, remove the jruby-puppet.compat-version setting and add the line 'JRUBY_JAR=/opt/puppetlabs/server/apps/puppetserver/jruby-9k.jar' to the /etc/sysconfig/(pe-)puppet(-)server or /etc/default/(pe-)puppet(-)server file.")
```
rlinehan: Nitpick, would it be possible to break this up a little more so the line isn't so long?

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
camlow325: ah, woops, yeah - I'll do that, thanks.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
camlow325: I don't really know how to i18n but I was thinking that full sentences need to be wrapped in an `i18n/trs` form because of how the i18n tooling works, so like...:

```clj
(i18n/trs "To use a Ruby language ... or /etc/default/(pe-)puppet(-)server file.")
```

If that's true, then I think I'd have to break this into separate sentences to make it work for i18n.  Assuming we went that way, can you suggest a good way to break them up?

rlinehan: Oh right, I forgot about that about i18n... uh....... in that case maybe just forget I said anything.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
camlow325: Added the `i18n/trs` wrapper in afdcbc4.

camlow325:
@rlinehan, I removed the info about the `compat-version` setting from the config_file_puppetserver.markdown file in db8c55b.

@gguillotte, fyi, this is a setting that we'll need to somehow document as "dropped" / no longer valid in Puppet 5.  We have a "deprecated settings" doc already but this one seems a bit different since it will now have been ripped out completely, not just deprecated.  I'll put some info in the release notes on the JIRA ticket as a reminder that we need to do something for this one.

camlow325:
@rlinehan I think I've hit all of your feedback so far and this should be able to be merged now since the jruby-utils and jruby-deps versions that the project.clj references now exist.  Should wait to see if it passes Travis tests.

gguillotte:
@camlow325 Would it be appropriate to document the JRUBY_JAR variable since we mention it in the variable's warning message? It seems appropriate for the `configuration.markdown` doc with a mention and link on the deprecated features page.

camlow325:
@gguillotte Yeah, I definitely think we should document the use of the `JRUBY_JAR` variable.  I was hoping to do that as part of the work for [SERVER-1779](https://tickets.puppetlabs.com/browse/SERVER-1779), which we were hoping to pick up as a follow-on to this ticket.

jpinsonault:
Since Ruth's feedback looks like it's been addressed I'll just merge this

PR #1412, opened 2017-05-31
(SERVER-1786) Remove site.pp file between Puppet compat tests

camlow325:
Previously, each of the Puppet compat tests uploaded a site.pp file but
did not remove the file as each test completed.  This would lead to any
subsequent tests which did not replace the site.pp file having problems
when an agent run reused the file from the prior test - in particular
the puppetX_version_test.rb.  In this commit, the site.pp file is
cleaned up as part of each test's teardown.

### DIFF
--- a/acceptance/lib/puppetserver/acceptance/compat_utils.rb
+++ b/acceptance/lib/puppetserver/acceptance/compat_utils.rb
@@ -16,7 +16,7 @@ SITEPP
   end
 end
 
-def rm_vardirs()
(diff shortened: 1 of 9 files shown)


camlow325:
I ran the puppet3 compat test suite for a centos-7 master / windows-2008-r2 agent beaker host config and all of the tests passed.

adrienthebo:
Merging promptly since this doesn't affect tests that will be run in Travis.

PR #1414, opened 2017-06-01
(SERVER-1779) Pre-docs for configuring JRuby version

camlow325:
This commit adds a section to the `configuration.markdown` file about
configuring the version of JRuby that Puppet Server uses and referencing
the removal of support for the `jruby-puppet.compat-version` setting.


camlow325:
@gguillotte Here's my first cut at this.  I suspect that we may want to sprinkle some of this information into other docs - like adding the `compat-version` setting into a `removed-settings` doc or somesuch, if it makes sense to create one.

PR #1415, opened 2017-06-01
(SERVER-1786) Set default versions of puppet when testing legacy compat

adrienthebo:
This adds a default value to puppet versions that we will use when
testing legacy compatability.

Previously we failed when no value was given, with this change we will
now log an info message.

Amended to apply to Puppetserver 2.7.x.


camlow325:
Going to merge it since it looks like what we have on master already and shouldn't affect the output of a Travis run anyway.

PR #1416, opened 2017-06-01
(SERVER-1787) Puppet 3 agents must use foss install type

adrienthebo:
If we're running Puppet 3 compatibility tests the agent install type
doesn't matter, as we _must_ use the 'foss' install type. This commit
manually reconfigures the Puppet agent hosts to use the 'foss' install
type and ignore the '--aio' command line flag and config option as that
option can only meaningfully apply to the master node, and never the
agents under test.

### DIFF
--- a/acceptance/suites/pre_suite/puppet3_compat/70_install_puppet.rb
+++ b/acceptance/suites/pre_suite/puppet3_compat/70_install_puppet.rb
@@ -10,4 +10,6 @@ if not puppet_version
   puppet_version = default_puppet_version
 end
 
+# Always force 3.x agents to use FOSS puppet install type
+nonmaster_agents.each { |agent| agent[:type] = 'foss' }
 install_puppet_on(nonmaster_agents, {:version => puppet_version})
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ gem 'jira-ruby', :group => :development
 
 group :test do
   gem 'rspec'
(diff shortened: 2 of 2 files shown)


acceptance/suites/pre_suite/puppet3_compat/70_install_puppet.rb
```diff
@@ -10,4 +10,6 @@
   puppet_version = default_puppet_version
 end
 
+# Always force 3.x agents to use FOSS puppet install type
+nonmaster_agents.each { |agent| agent[:type] == 'foss' }
```
camlow325: Should this be "agent[:type] = 'foss'" rather than ==?

adrienthebo: Yeeep. Initially I fixed this issue via pry, ran the tests, and when that worked I did the initial faulty PR and ran the tests again. When they failed I flagged this with the DO NOT MERGE commit tag while I verified. Running the tests now with the fixes; assuming that works I'm going to force push on top of this whoopsie.

adrienthebo: Okay, got this fixed. IN SOVIET RUSSIA COMMIT FORCE PUSH YOU!

rlinehan:
Tested it out, seems to fix the problem. LGTM.

PR #1419, opened 2017-06-06
(SERVER-1808) Adjust file endpoint auth.conf rules for valid methods

camlow325:
This commit splits the /puppet/v3/file auth.conf rule into separate
rules for file_bucket_file, file_content, and file_metadata.  The
discrete rules reflect the minimal set of request methods which are
valid for each endpoint.

### DIFF
--- a/ezbake/config/conf.d/auth.conf
+++ b/ezbake/config/conf.d/auth.conf
@@ -56,18 +56,43 @@ authorization: {
             name: "puppetlabs environments"
         },
         {
-            # Allow nodes to access all file services; this is necessary for
-            # pluginsync, file serving from modules, and file serving from
-            # custom mount points (see fileserver.conf). Note that the `/file`
-            # prefix matches requests to file_metadata, file_content, and
-            # file_bucket_file paths.
+            # Allow nodes to access all file_bucket_files.  Note that access for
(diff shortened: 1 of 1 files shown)


camlow325:
I ran the beaker Puppet Server smoke and full Puppet acceptance tests with these changes and all tests were still passing.

camlow325:
@joshcooper, do you care to look at this or should we just merge it?

ezbake/config/conf.d/auth.conf
```diff
@@ -56,18 +56,43 @@ authorization: {
...
+            allow: "*"
+            sort-order: 500
+            name: "puppetlabs file content"
+        },
+        {
+            # Allow nodes to access all file_metadata.  Note that access for the
+            # 'delete' method is forbidden by Puppet regardless of the
+            # configuration of this rule.
+            match-request: {
+                path: "/puppet/v3/file_metadata"
```
joshcooper: I assume this will match the plural version (file_metadatas) too?

camlow325: It will, yeah.  With the `type: path` qualifier, the value for `path` is basically treated as a "starts-with" for matching purposes.

PR #1423, opened 2017-06-12
(SERVER-1826) Send mime type with static file serving response

justinstoller:
I'm curious what folks think of this. Since it was just for one response of one endpoint I thought that simply updating the map returned would be the simplest solution, and since we're already directly returning a map with `:body` and `:status` it seemed weird to use the ring helpers to set the header.

### DIFF
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -472,6 +472,7 @@
 
         :else
         {:status 200
+         :headers {"Content-Type" "application/octet-stream"}
          :body (get-code-content environment code-id file-path)}))))
 
 (def MetricIdsForStatus (schema/atom [[schema/Str]]))
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -457,17 +457,22 @@
       (cond
         (some empty? [environment code-id file-path])
         {:status 400
+         :headers {"Content-Type" "text/plain"}
          :body (i18n/tru "Error: A /static_file_content request requires an environment, a code-id, and a file-path")}
+
         (not (nil? (schema/check ps-common/Environment environment)))
         {:status 400
+         :headers {"Content-Type" "text/plain"}
          :body (ps-common/environment-validation-error-msg environment)}
 
         (not (nil? (schema/check ps-common/CodeId code-id)))
         {:status 400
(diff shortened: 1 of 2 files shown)


src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -472,6 +472,7 @@
 
         :else
         {:status 200
+         :headers {"Content-Type" "application/octet-stream"}
```
adrienthebo: This solves the immediate problem of setting the content type for a successful request. However, there are a lot of other cases where we don't return a content type - such as the error cases prior to this line. Should we be setting a default content-type of `text/plain` if no content-type was given?

justinstoller: I'd be into doing that, though I interpreted the ticket as pretty directed toward just this response. If @camlow325 is cool with that being the scope of this work, I'll go ahead and update this accordingly.

camlow325: Yeah, I think adding `text/plain` would be reasonable to do for the error responses - the 400 and 403s response maps above and would be good with adding that to this PR while we're in here.

> Should we be setting a default content-type of text/plain if no content-type was given?

@finch For this one, do you mean for the success case or just the fail cases?  I had been thinking for the success case that we'd always return `application/octet-stream`.

adrienthebo: The success case should always be `application/octet-stream`; for the above error cases it seems reasonable to return `text/plain`. I was musing about the slightly wider case - if a given response doesn't have a content-type at all should we add one?

camlow325: > I was musing about the slightly wider case - if a given response doesn't have a content-type at all should we add one?

Oh, you mean like at a higher level in the middleware stack we might default in a Content-Type if none had been set in the response map at the lower-level?  I tend to think it's best to avoid a higher-level default since you may not have as much context there to know what the body represents.

If we do `text/plain` for each of the error clauses and `application/octet-stream` for success for this endpoint, I think we have all of the `cond` clauses covered for this endpoint anyway.

adrienthebo: Sounds good to me.

justinstoller: updated

camlow325:
@adrienthebo This looks good to me.  Are you good with merging it?

PR #1424, opened 2017-06-13
[dnm] (TK-442) Bring in clj-parent with new version of tk-metrics

justinstoller:
I think we'll probably want to wait on this and instead pull in a new version after we've bumped tk-metrics within clj-parent to take in this PR https://github.com/puppetlabs/trapperkeeper-metrics/pull/52 as well.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -37,7 +37,7 @@
 
   :min-lein-version "2.7.1"
 
-  :parent-project {:coords [puppetlabs/clj-parent "1.1.0"]
+  :parent-project {:coords [puppetlabs/clj-parent "1.2.1"]
                    :inherit [:managed-dependencies]}
 
   :dependencies [[org.clojure/clojure]


rlinehan:
Do we think that it's likely that we're gonna get all the things bumped/into clj-parent before tomorrow? I'm kind of inclined to just merge this so we get the change in, and then bump the other things later.

justinstoller:
This is now blocked on merging (and releasing) [this update to clj-parent](https://github.com/puppetlabs/clj-parent/pull/70).

justinstoller:
Thinking about this after our sprint planning, I'm going to close this PR and reopen it after clj-parent has been updated and I've actually configured puppetserver to disable the url metrics. This work will be under the [SERVER-1841](https://tickets.puppetlabs.com/browse/SERVER-1841) banner.

PR #1425, opened 2017-06-14
(SERVER-1827) Change default cert-status content-type to json

camlow325:
This commit changes the default content-type which is returned when no
'accept' header is given for a certificate-status request from
'text/pson' to 'application/json'.  This is being done for consistency
with defaulting to 'application/json' in place of 'text/pson' elsewhere
for server endpoints in Puppet 5.  Support for roundtripping 'accept:
pson' or 'accept: text/pson' requests in responses as 'content-type:
pson' or 'content-type: text/pson', respectively, has been preserved.

### DIFF
--- a/test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
+++ b/test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
@@ -356,6 +356,15 @@
                 response (test-app request)]
             (is (= 404 (:status response)))))
 
+        (testing "returns json when no accept header specified"
+          (let [request {:uri "/v1/certificate_status/localhost"
+                         :request-method :get
+                         :headers {}}
+                response (test-app request)]
+            (is (= 200 (:status response))
+                (ks/pprint-to-string response))
+            (is (.startsWith (get-in response [:headers "Content-Type"]) "application/json"))))
+
         (testing "honors 'Accept: pson' header"
           (let [request {:uri "/v1/certificate_status/localhost"
                          :request-method :get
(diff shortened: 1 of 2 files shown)


test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
```diff
@@ -407,13 +416,23 @@
...
+                   (set (json/parse-string (:body response) true))))))
+
         (testing "with 'Accept: pson'"
           (let [response (test-app
                           {:uri "/v1/certificate_statuses/thisisirrelevant"
                            :request-method :get
                            :headers {"accept" "pson"}})]
           (is (= 200 (:status response)))
-          (is (.startsWith (get-in response [:headers "Content-Type"]) "text/pson"))
+          (is (.startsWith (get-in response [:headers "Content-Type"]) "pson"))
```
adrienthebo: Why are we changing from `"text/pson"` to `"pson"`? Is this a change that's coming from ruby Puppet?

camlow325: No, this would be a result of the changes on my PR.  Previously, it was defaulting to `text/pson` because liberator was failing to pass through the value from the `accept` header as the `media-type` for the representation - probably because it thinks that `pson` looks like a bogus content-type - and so we were falling back to delivering our hard-coded internal default.  I think the original implementation was actually buggy but `text/pson` and `pson` are somewhat interchangeable so we probably just accepted this at the time.

When I initially tried just changing that default to `application/json`, it resulted in an `accept` header of `pson` for the request returning a `content-type` of `application/json` in the response, which seemed wrong.  For the changes on the PR, then, I just changed the implementation to try to roundtrip the value from the `accept` header in the event that there is one and only fallback to `application/json` in the event that the client didn't provide an `accept` header.  That seems like the best thing to do from a standards perspective.  It would seem that a legacy client which sent an `accept` header of `pson` should know how to deal with a `Content-Type` of `pson` in the response.

Does this make sense?

adrienthebo: Sounds sensible to me.

PR #1426, opened 2017-06-14
(MAINT) Bump puppet-agent to 98208e4

camlow325:
This PR bumps the puppet submodule and puppet-agent versions for
testing to 27fdad2 and 4.99.0.269.g98208e4, respectively.

Previously, the Clojure unit tests contained a test for the value of the
Puppet archive_files setting.  Per changes in PUP-7260 which just landed
in Ruby Puppet, this setting no longer exists.  This commit changes the
test to instead validate the value of the autoflush setting.  The
specific setting being tested isn't important.  The test just proves
that a known expected default value for a setting can be retrieved from
Ruby Puppet.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -25,7 +25,7 @@ module PuppetServerExtensions
 
     puppet_version = get_option_value(options[:puppet_version],
                          nil, "Puppet Version", "PUPPET_VERSION",
-                         "4.99.0.202.ge36cade",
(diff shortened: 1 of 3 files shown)


camlow325:
Updating this to pick up a fix for an acceptance test which has frequently been failing of late when running the Puppet acceptance tests under JRuby 9k for Puppet Server, [PUP-7652](https://tickets.puppetlabs.com/browse/PUP-7652).

PR #1430, opened 2017-06-15
(TK-443) Disable http url based metrics

justinstoller:
Previously http-client would create url and url+method metrics for every
call which could become expensive in terms of memory on a production
system.

With the 0.9.0 release (brought in by clj-parent 1.2.1) we can disable this
behavior, which we do with this patch.

### DIFF
--- a/test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
+++ b/test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
@@ -419,20 +419,14 @@
                (testing "GET request"
                  (.runScriptlet container "$c.get('/hello', {})")
                  (let [metrics-data (metrics/get-client-metrics-data metric-registry)]
-                   (is (= #{(add-metric-ns "with-url.http://localhost:8080/hello")}
-                          (set (map :metric-name (:url metrics-data)))))
-                   (is (= #{(add-metric-ns "with-url-and-method.http://localhost:8080/hello.GET")}
-                          (set (map :metric-name (:url-and-method metrics-data)))))))
+                   (is (= [] (:url metrics-data)))
+                   (is (= [] (:url-and-method metrics-data)))))
                (testing "POST request"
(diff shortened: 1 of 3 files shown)


justinstoller:
One interesting thing is that puppetserver doesn't seem to use the clj api except as a testing utility. In that case our test clients will automatically generate url based metrics. To do so otherwise would require creating a new client, in tests, or to create a helper namespace that largely copies the content of of clj-http-client's puppetlabs.http.client.sync namespace, but with a different default value for `:enable-url-metrics?`. Both of which seem tedious from a user's perspective. It has me wondering if we want to change the default behavior to have url based metrics disabled by default?

camlow325:
Our test clients would only generate the URL metrics wherever they would construct and configure a `MetricRegistry`, right?  I think I'm okay for those cases with the default behavior to be to have the URL metrics created.  It still seems reasonable to me that if you've gone to the trouble of configuring metrics that the default behavior of having URL metrics posted into that registry is reasonable.  For more "advanced" cases like Puppet Server where we've done the additional work to ensure that all client calls pass in a metric-id, making an explicit call to opt out of the URL-based metrics because we don't have a specific use for them and they consume extra memory also seems reasonable to me.  Does that make sense or am I missing the point?

test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
```diff
@@ -419,20 +419,14 @@
                (testing "GET request"
```
camlow325: Might be good to change the testing string above a bit to indicate that we're expecting the url and url-and-method metrics to be absent because we intentionally disabled them when setting up the HTTP client.

justinstoller:
> Does that make sense or am I missing the point?

No, Im still figuring out what's what and the various use cases with this. That makes a lot of sense.

camlow325:
@rlinehan - I'm +1 on this and would be good with you merging whenever you're satisfied.

rlinehan:
Merging despite travis not being done since the only change since the last time it ran was to a test string.

PR #1431, opened 2017-06-16
(MAINT) Bump minimum puppet-agent for packaging to 4.99.0

camlow325:
This commit bumps the minimum puppet-agent package version from 1.6.0 to
4.99.0 in preparation for the Puppet 5 release.

### DIFF
--- a/resources/ext/ezbake.conf
+++ b/resources/ext/ezbake.conf
@@ -7,7 +7,7 @@
 ezbake: {
    pe: {}
    foss: {
-      redhat: { dependencies: ["puppet-agent >= 1.6.0"],
+      redhat: { dependencies: ["puppet-agent >= 4.99.0"],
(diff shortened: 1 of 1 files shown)


camlow325:
For reference, I did an ezbake build of this [here](http://builds.puppetlabs.lan/puppetserver/5.0.0.master.SNAPSHOT.2017.06.16T0927/).  I ran the testtest.rb smoke test to ensure that the puppetserver packages install with this dependency version change and that a basic agent run works after the fact.  I ran this against a representative set of our supported OSes - Centos 6 & 7, Debian 8, Ubuntu 16.04, and SLES 12.  All passed.

camlow325:
@justinstoller Nah, I'm good just merging this without you having to repeat the tests I did, if you are.

PR #1432, opened 2017-06-16
(SERVER-1798) Support testing against puppetdb and puppet-agent nightly

adrienthebo:
This commit allows the PUPPET_BUILD_VERSION and PUPPETDB_BUILD_VERSION variables to be set to 'LATEST'. When those environment variables are set to 'LATEST' the version will be set based on the last successful build of puppetdb or puppet-agent.

### DIFF
--- a/acceptance/suites/pre_suite/foss/00_setup_environment.rb
+++ b/acceptance/suites/pre_suite/foss/00_setup_environment.rb
@@ -1,2 +1,13 @@
-step "Initialize Test Config"
+step "Initialize Test Config" do
   PuppetServerExtensions.initialize_config options
+
+  if PuppetServerExtensions.config[:puppet_build_version] == 'LATEST'
+    PuppetServerExtensions.config[:puppet_build_version] = latest_agent_build
+    logger.info "Using latest Puppet agent build version (#{PuppetServerExtensions.config[:puppet_build_version]})"
+  end
+
+  if PuppetServerExtensions.config[:puppetdb_build_version] == 'LATEST'
+    PuppetServerExtensions.config[:puppetdb_build_version] = latest_pdb_build
+    logger.info "Using latest PuppetDB build version (#{PuppetServerExtensions.config[:puppetdb_build_version]})"
+  end
+end
--- a/acceptance/suites/pre_suite/foss/00_setup_environment.rb
+++ b/acceptance/suites/pre_suite/foss/00_setup_environment.rb
@@ -1,13 +1,16 @@
 step "Initialize Test Config" do
   PuppetServerExtensions.initialize_config options
 
-  if PuppetServerExtensions.config[:puppet_build_version] == 'LATEST'
-    PuppetServerExtensions.config[:puppet_build_version] = latest_agent_build
-    logger.info "Using latest Puppet agent build version (#{PuppetServerExtensions.config[:puppet_build_version]})"
-  end
+  latest_options = [
+    [:puppet_build_version, lambda { latest_agent_build }],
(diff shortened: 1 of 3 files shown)


camlow325:
Another thing I think we'll need at a minimum for this is some logic which checks out the puppet submodule at the Puppet repo commit which corresponds to the puppet-agent build that we'll be testing.  That way, we're assured that we're running tests from the Puppet repo that should be compatible with the corresponding puppet-agent package version.  Does that make sense?

acceptance/suites/pre_suite/foss/35_install_pdb.rb
```diff
@@ -6,7 +6,7 @@
 
 step "Install PuppetDB repository" do
   install_puppetlabs_dev_repo(
-    master, 'puppetdb', test_config[:puppetdb_build_version],
+    master, 'puppetdb', test_config[:pdb_build_version],
```
camlow325: Why did this change?  I didn't see anything else in the PR which referenced it.

acceptance/suites/pre_suite/foss/00_setup_environment.rb
```diff
@@ -1,2 +1,13 @@
-step "Initialize Test Config"
+step "Initialize Test Config" do
   PuppetServerExtensions.initialize_config options
+
+  if PuppetServerExtensions.config[:puppet_build_version] == 'LATEST'
```
camlow325: In our current [helper.rb file](https://github.com/puppetlabs/puppetserver/blob/bd05e6ef2ca0229c82526928843085bd50f0e58f/acceptance/lib/helper.rb#L28), we've been setting both the `puppet_version` and `puppet_build_version`.  Should we have something similar here for `puppet_version`?

camlow325: In order for these new settings to take effect in place of the ones that we're currently using in the [helper.rb file](https://github.com/puppetlabs/puppetserver/blob/bd05e6ef2ca0229c82526928843085bd50f0e58f/acceptance/lib/helper.rb#L26-L44), should this PR also set the corresponding helper.rb versions to "LATEST"?  Maybe it would also be worthwhile to have a value like `nil` be treated like `LATEST` for each setting?

adrienthebo: What do we want our default behavior to be? If we expect our acceptance tests to default to running against the latest version we can remove the default values and use the latest on empty values. I'm not strongly opinionated either way.

adrienthebo: Done.

acceptance/suites/pre_suite/foss/35_install_pdb.rb
adrienthebo: Thank you for catching this, fixed.

acceptance/suites/pre_suite/foss/00_setup_environment.rb
camlow325: I think we should move our default behavior to "latest" but that it would be good to leave the code in place which sets the different variables in the helper.rb file as a "template" for what we need to set in the event that we need to pin to a specific version temporarily.  We may need to temporarily pin in the future if we have a prolonged period of puppet-agent failures on latest but need to release puppet-server based on whatever the last puppet-agent release version was, for example.

adrienthebo: I've replaced the default values with `nil` when getting the version option values, could you take a look at the latest changes?

camlow325:
Looks like this will need a rebase.  😞

PR #1433, opened 2017-06-20
(SERVER-1829) Pre-docs for Metrics APIs

justinstoller:
Previous, to Puppet Server 5 the metrics v1 api was provided by a custom
PE only extension. That functionality has been open sourced and moved
into `trapperkeeper-metrics`.

In addition to the v1 metrics, a v2 metrics endpoint was recenetly added
that provided additional features built upon the Jolokia library.

This commit ports the PE v1 metrics documentation to Puppet Server and
adds an introduction to the v2 Jolokia-based metrics.

------

It dawns on me as I'm making this PR that my initial idea of what would be valuable to @gguillotte might be waaayyy off. If so, please let me know and I'll do whatever is best for getting it in that state (or filed against whatever repo it needs to be)


documentation/index.markdown
```diff
@@ -45,6 +45,8 @@ Puppet Server is the next-generation application for managing Puppet agents.
     * [Static file content](./puppet-api/v3/static_file_content.markdown)
 * **Status API endpoints**
     * [Status services](./status-api/v1/services.markdown)
+* **Metrics API endpoints**
+    * [Metrics API versions](./metrics_api.markdown)
```
camlow325: Maybe could just call this `Metrics service` instead of `Metrics API versions`, for consistency with the "status" service reference above?

documentation/metrics_api.markdown
```diff
@@ -0,0 +1,267 @@
...
+metrics, namely
+[managed beans (MBeans)](https://docs.oracle.com/javase/tutorial/jmx/mbeans/).
+
+## Metrics v2
+
+The first uses the [Jolokia](https://jolokia.org) library and is accessible at:
+
+    /metrics/v2/
+
+Since the v2 API is implimented using a large open source metrics library with
```
camlow325: s/implimented/implemented

gguillotte:
FYI, I'm revising this in a PR, and I'll break up the API into versions.

documentation/metrics_api.markdown
```diff
@@ -0,0 +1,267 @@
...
+---
+
+# The Metrics API
+
+Puppet Enterprise includes two optional, enabled-by-default web APIs for
+[Java Management Extension (JMX)](https://docs.oracle.com/javase/tutorial/jmx/index.html)
+metrics, namely
+[managed beans (MBeans)](https://docs.oracle.com/javase/tutorial/jmx/mbeans/).
+
+## Metrics v2
```
camlow325: Hmm, so one thing I think we should talk about here is how to enable the Jolokia version of the Metrics endpoint - since it isn't enabled by default, from what I remember.  This [doc](https://github.com/puppetlabs/trapperkeeper-metrics/blob/7e6d2bf6a1c61f1c652405492c402ee1186ea860/documentation/configuration.md) in the trapperkeeper-metrics repo has a section about the `metrics-webservice.jolokia` configuration in it.  It doesn't appear that we updated the default [metrics.conf](https://github.com/puppetlabs/puppetserver/blob/bd05e6ef2ca0229c82526928843085bd50f0e58f/ezbake/config/conf.d/metrics.conf) file in Puppet Server with the skeleton of this configuration section, unfortunately.  I actually think it might be good to do that as part of this work - adding in a config where jolokia is disabled by default.  Could be done as a separate PR from this one if @gguillotte is working on his own revision to these docs and doesn't want us to merge this.

documentation/metrics_api.markdown
```diff
@@ -0,0 +1,267 @@
...
+For security reasons only the read operations of the Jolokia interface are
+enabled by default. This includes:
+
+* `read`
+* `list`
+* `version`
+* `search`
+
+Users may change the security access policy by providing their own
+`/etc/puppetlabs/puppetserver/jolokia-access.xml` file that follows the
```
camlow325: I *think* this would only work if the user had added a `metrics-webserver.jolokia.servlet-init-params.policyLocation` setting which was set to `file:///etc/puppetlabs/puppetserver/jolokia-access.xml`.  We could set that value in the config by default but if we do that without providing our own default one in the puppetserver package, I think users would see a nasty-looking ERROR message in the puppetserver log at startup, something like:

> 2017-06-21 09:15:42,029 ERROR [p.t.s.m.jolokia] Error while accessing access restrictor at file:///etc/puppetlabs/puppetserver/jolokia-access.xml. Denying all access to MBeans for security reasons. Exception: java.io.FileNotFoundException: /etc/puppetlabs/puppetserver/jolokia-access.xml (No such file or directory)

I'm thinking it might be best to put that setting in our default config but have it be commented out by default and document here that the user would need to uncomment it, set to the desired location, and restart/reload the service.

@Sharpie - what do you think on this?

documentation/metrics_api.markdown
camlow325: d'oh, I think I'm wrong about the default for Jolokia.  I think it actually is enabled by default.  Might still be good to add a boilerplate `metrics-webservice` section into the config with the `enabled` and `servlet-init-params` sections commented out, so they would be using their default values.  @justinstoller, WDYT?

justinstoller: Its definitely enabled by default. The example responses are edited actual responses that I got from a running instance.

I may have conflated the content in the metrics.conf by default with some of the structure in the docs (I can't resolve the host I set up to help with this right now to confirm). I definitely agree that having the metrics.conf with the default values for the main settings in there and commented out is valuable and should be done now (I guess I just thought that was already done).

camlow325: > I definitely agree that having the metrics.conf with the default values for the main settings in there and commented out is valuable and should be done now (I guess I just thought that was already done).

Yeah, that was probably an oversight on our part.  I honestly had thought so too until I looked at it again today.

justinstoller:
Hot Damn. @gguillotte whipped the docs into shape. I've merged his changes. I've also confirmed the behaviors and defaults that @camlow325 and I discussed in the comments. I'll push a quick update for that before we should merge.

documentation/metrics-api/v1/metrics_api.markdown
```diff
@@ -0,0 +1,92 @@
...
+#### Response keys
+
+The response format, though always JSON, depends on the request format:
+
+-   Requests with a JSON object return a JSON object where the values of the
+    original object are transformed into the Mbeans' attributes for the metric
+    names.
+-   Requests with a JSON array return a JSON array where the items of the original
+    array are transformed into the Mbeans' attributes for the metric names.
+-   Requests with a JSON string return the a JSON object of the Mbean's attributes
```
camlow325: s/the a/a

justinstoller: 😝

documentation/metrics-api/v2/metrics_api.markdown
```diff
@@ -0,0 +1,182 @@
...
+-   `list`
+-   `version`
+-   `search`
+
+### Configuring Jolokia
+
+To change the security access policy, create the `/etc/puppetlabs/puppetserver/jolokia-access.xml` file with contents that follow the [Jolokia access policy](https://jolokia.org/reference/html/security.html) and uncomment the `metrics.metrics-webservice.jolokia.servlet-init-params.policyLocation` parameter before restarting puppetserver.
+
+The `metrics.metrics-webservice.jolokia.servlet-init-params` table
+within the [`/etc/puppetlabs/puppetserver/conf.d/metrics.conf`](../../config_file_metrics.markdown) file provides more configuration options. See Jolokia's [agent initialization documentation](https://jolokia.org/reference/html/agents.html#agent-war-init-params) for all of the available options.
```
justinstoller: oy vey, it looks like this file doesn't exist? And that the metrics.conf file isn't documented anywhere (PE [links to how to configure graphite via hiera in the console](https://docs.puppet.com/pe/latest/config_puppetserver.html#puppet-server-configuration-files) and tk-metrics isn't [much more than the default config file](https://github.com/puppetlabs/trapperkeeper-metrics/blob/master/documentation/configuration.md))...

I assume I should probably pull in the tk-metrics configuration file and annotate it some with the equivalent file based recommendations for configuring the graphite reporter from the PE docs. Would that be sufficient? Sorry, about that.

gguillotte: Yeah, I'm working on the metrics.conf docs right now, actually.

justinstoller: 🎆 🎊 ✨ 🎉

camlow325: @justinstoller Do you mean that the docs for this file don't exist yet or that the metrics.conf file itself doesn't exist?  I would have expected a file at `/etc/puppetlabs/puppetserver/conf.d/metrics.conf` which looks like [this](https://github.com/puppetlabs/puppetserver/blob/bd05e6ef2ca0229c82526928843085bd50f0e58f/ezbake/config/conf.d/metrics.conf) to be present after installation of a Puppet Server-5 based OSS package.  Are you seeing that?

justinstoller: The doc, the file exists.

documentation/metrics_api.markdown
Sharpie: Yeah, the default policy file is `classpath:/jolokia-access.xml` which uses the locked-down policy packaged in the JAR. Putting a commented example in the default configuration sounds like a good idea.

PR #1435, opened 2017-06-21
(SERVER-1830) Add metrics docs.

gguillotte:
-   Pull and revise PE metrics docs for open source metrics.
-   Add Grafana dashboard JSON.
-   Document `metrics.conf`.

### DIFF
--- a/documentation/config_file_metrics.markdown
+++ b/documentation/config_file_metrics.markdown
@@ -13,7 +13,7 @@ All settings in the file are contained in a HOCON `metrics` section.
 -   `server-id`: A unique identifier to be used as part of the namespace for metrics that this server produces.
 
 -   `registries`: A section that contains settings to control which metrics are reported, and how they're reported.
-    -   `puppetserver`: A section that wraps registries settings.
+    -   `<REGISTRY NAME>`: A section named for a registry that contains its settings. In Puppet Server's case, this section should be `puppetserver`.
         -   `metrics-allowed`: An array of metrics to report. See the [metrics documentation](./puppet_server_metrics.markdown) for details about individual metrics.
         -   `reporters`: Can contain `jmx` and `graphite` sections with a single Boolean `enabled` setting to enable or disable each reporter type.
 -   `reporters`: Configures reporters that distribute metrics to external services or viewers.
--- a/documentation/config_file_metrics.markdown
+++ b/documentation/config_file_metrics.markdown
@@ -1,0 +1,61 @@
+---
+layout: default
+title: "Puppet Server Configuration Files: metrics.conf"
+canonical: "/puppetserver/latest/config_file_metrics.html"
+---
+
+The `metrics.conf` file configures Puppet Server's [metrics services](./puppet_server_metrics.markdown) and [v2 metrics API](./metrics-api/v2/metrics_api.markdown).
+
+## Settings
+
+All settings in the file are contained in a HOCON `metrics` section.
+
+-   `server-id`: A unique identifier to be used as part of the namespace for metrics that this server produces.
+
+-   `registries`: A section that contains settings to control which metrics are reported, and how they're reported.
+    -   `puppetserver`: A section that wraps registries settings.
+        -   `metrics-allowed`: An array of metrics to report. See the [metrics documentation](./puppet_server_metrics.markdown) for details about individual metrics.
+        -   `reporters`: Can contain `jmx` and `graphite` sections with a single Boolean `enabled` setting to enable or disable each reporter type.
+-   `reporters`: Configures reporters that distribute metrics to external services or viewers.
+    -   `graphite`: Contains settings for the Graphite reporter.
+        -   `host`: A string containing the Graphite server's hostname or IP address.
+        -   `port`: Contains the Graphite service's port number.
+        -   `update-interval-seconds`: Sets the interval on which Puppet Server will send metrics to the Graphite server.
+
+## Example
+
+Puppet Server ships with a default `metrics.conf` file in Puppet Server's `conf.d` directory, similar to the below example with additional comments.
+
+```
+metrics: {
+    server-id: localhost
+    registries: {
+        puppetserver: {
+            # specify metrics to allow in addition to those in the default list
+            #metrics-allowed: ["compiler.compile.production"]
+
+            reporters: {
+                jmx: {
+                    enabled: true
+                }
+                # enable or disable Graphite metrics reporter
+                #graphite: {
+                #    enabled: true
+                #}
+            }
+
+        }
+    }
+
+    reporters: {
+        #graphite: {
+        #    # graphite host
+        #    host: "127.0.0.1"
+        #    # graphite metrics port
+        #    port: 2003
+        #    # how often to send metrics to graphite
+        #    update-interval-seconds: 5
+        #}
+    }
+}
+```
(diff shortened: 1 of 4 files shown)


gguillotte:
FYI @camlow325 @justinstoller

documentation/puppet_server_metrics.markdown
```diff
@@ -0,0 +1,440 @@
...
+
+-   **Request durations:** This graph breaks down the average response times for different types of requests made by Puppet agents. This indicates how expensive catalog and report requests are compared to the other types of requests. It also provides a way to see changes in catalog compilation times when you modify your Puppet code. A sharp curve upward for all of the types of requests indicates an overloaded server, and they should trend downward after reducing the load on the server.
+
+-   **Request ratios:** This graph shows how many requests of each type that Puppet Server has handled. Under normal circumstances, you should see about the same number of catalog, node, or report requests, because these all happen once per agent run. The number of file and file metadata requests correlate to how many remote file resources are in the agents' catalogs.
+
+-   **Communications with PuppetDB:** This graph tracks the amount of time it takes Puppet Server to send data and requests for common operations to, and receive responses from, PuppetDB.
+
+-   **File Sync:** This graph tracks how long Puppet Server spends on File Sync operations, for both its storage and client services.
```
camlow325: Might be best to leave this out or tag it as "for PE only" since these ones aren't available in OSS.

documentation/puppet_server_metrics.markdown
```diff
@@ -0,0 +1,440 @@
...
+
+    -   `puppetlabs.<MASTER-HOSTNAME>.compiler.find_node`: The time spent retrieving node data. If the Node Classifier (or another ENC) is configured, this includes the time spent communicating with it.
+
+    -   `puppetlabs.<MASTER-HOSTNAME>.compiler.static_compile`: The time spent compiling [static catalogs][].
+
+    -   `puppetlabs.<MASTER-HOSTNAME>.compiler.static_compile_inlining`: The time spent inlining metadata for static catalogs.
+
+    -   `puppetlabs.<MASTER-HOSTNAME>.compiler.static_compile_postprocessing`: The time spent post-processing static catalogs.
+
+#### File sync metrics
```
camlow325: Same comment here as above about these being for PE only.

gguillotte:
@camlow325 Removed those references and fixed links throughout to point to Server and Puppet content instead of PE.

camlow325:
@justinstoller - could you take a quick look at this?

documentation/config_file_metrics.markdown
```diff
@@ -0,0 +1,61 @@
...
+The `metrics.conf` file configures Puppet Server's [metrics services](./puppet_server_metrics.markdown) and [v2 metrics API](./metrics-api/v2/metrics_api.markdown).
+
+## Settings
+
+All settings in the file are contained in a HOCON `metrics` section.
+
+-   `server-id`: A unique identifier to be used as part of the namespace for metrics that this server produces.
+
+-   `registries`: A section that contains settings to control which metrics are reported, and how they're reported.
+    -   `puppetserver`: A section that wraps registries settings.
```
justinstoller: AFAICT, this actually maps to a registry name. So there's a default `puppetserver` metrics registry that we use and these are its settings. I guess minimum it's `registry's` rather than `registries`. Is my understanding correct, @camlow325 ?

documentation/config_file_metrics.markdown
```diff
@@ -0,0 +1,61 @@
...
+        -   `metrics-allowed`: An array of metrics to report. See the [metrics documentation](./puppet_server_metrics.markdown) for details about individual metrics.
+        -   `reporters`: Can contain `jmx` and `graphite` sections with a single Boolean `enabled` setting to enable or disable each reporter type.
```
justinstoller: It can contain other things (it can override settings defined in the `metrics.reporters` sections) but that's a more advanced thing for folks using multiple registries.

Don't think we want to lead people down that path (ie we only really want folks to use this section to enable/disable reporters for the `puppetserver` registry), but... I don't know why I'm writing this, you wrote exactly what we want folks to use this for.

documentation/puppet_server_metrics.markdown
```diff
@@ -0,0 +1,423 @@
...
+
+4.  Navigate to and select the edited JSON file.
+
+This loads a dashboard with nine graphs that display various metrics exported from the Puppet Server to the Graphite server. (For details, see [Using the Grafana dashboard](#using-the-sample-grafana-dashboard).) However, these graphs will remain empty until you enable Puppet Server's Graphite metrics.
+
+### Enabling Puppet Server's Graphite support
+
+Configure Puppet Server's [`metrics.conf`](./config_file_metrics.markdown) file to enable and use the Graphite server.
+
+1.  Set the `enabled` parameter to true in `metrics:registries:puppetserver:reporters:graphite`:
```
justinstoller: Crap.... I was using dot notation to talk about drilling into hocon tables. Did I do that wrong? (see: [`metrics.metrics-webservice.jolokia.servlet-init-params`](https://github.com/puppetlabs/puppetserver/blob/master/documentation/metrics-api/v2/metrics_api.markdown#configuring-jolokia))

justinstoller:
lgtm overall

One thing I was wondering about, does anyone think that the `metrics.conf` would be more understandable if we moved the `metrics.reporters` section above the `metrics.registries` section? ie, `metrics.reporters` generically configures our possible reporters. In `metrics.registries.<registry>.reporters` we enable reporters and do any special configuration for them per registry.

documentation/config_file_metrics.markdown
gguillotte: Yeah, I'd like to limit this to metrics that Puppet Server's designed to work with. We can always update this if things get more complex later.

gguillotte:
@justinstoller I don't mind things being in a cascading order as they are now when more specific settings override general settings. If that's the case, though, I should also clarify that in the docs.

gguillotte:
Updated to address @justinstoller's feedback.

documentation/config_file_metrics.markdown
camlow325: Yeah, we can override the more global default for `update-interval-seconds` here, for example.  I'm good with the idea of that being an "advanced" thing that we don't expect users to want to do generally (and, therefore, not bother documenting here) - even if it is something that the `trapperkeeper-metrics` service supports.

documentation/puppet_server_metrics.markdown
camlow325: Dot notation does seem more standard that colons.  So, that seems like a good change (and that @gguillotte already made it.  😄).

documentation/puppet_server_metrics.markdown
```diff
@@ -0,0 +1,423 @@
...
+
+    -   `puppetlabs.<MASTER-HOSTNAME>.memory.non-heap.used`
+
+    -   `puppetlabs.<MASTER-HOSTNAME>.memory.non-heap.max`
+
+### Modifying Puppet Server's exported metrics
+
+In addition to the above default metrics, you can also export metrics measuring specific environments and nodes.
+
+The `metrics.registries.puppetserve.metrics-allowed` parameter in [`metrics.conf`][] takes an array of strings representing the metrics you want to enable.
```
camlow325: s/puppetserve/puppetserver

camlow325:
Just the one minor typo nit and I think we're good to go on this one.

camlow325:
🚢

PR #1436, opened 2017-06-22
(SERVER-1850) Server 5.0 release notes

gguillotte:
-   Removes Server 2.x release notes and points to the docs site as an archive.
-   Adds Server 5.0 release notes.
-   Updates related docs with relevant information.

### DIFF
--- a/documentation/configuration.markdown
+++ b/documentation/configuration.markdown
@@ -102,15 +102,17 @@ puppetlabs.services.ca.certificate-authority-service/certificate-authority-servi
 #puppetlabs.services.ca.certificate-authority-disabled-service/certificate-authority-disabled-service
 ~~~
 
-## Configuring JRuby Version
+## Configuring the JRuby Version
 
-By default, Puppet Server utilizes a version of the JRuby 1.7.x series, running in a mode which conforms to MRI/Ruby language version 1.9. Starting with the Puppet Server 5.0 release, however, it is now possible to configure Puppet Server to instead use a version of the JRuby 9k series (9.x.y.z). The specific Ruby language version that each JRuby 9k release conforms to is expected to evolve over time, tracking newer Ruby 2.x language versions.  For example, the JRuby 9.1.8.0 release conforms to Ruby language version 2.3.1 whereas 9.1.9.0 conforms to Ruby language version 2.3.3.
+By default, Puppet Server uses a version of the JRuby 1.7.x series, running in a mode which conforms to MRI/Ruby language version 1.9. Starting with the Puppet Server 5.0 release, however, you can configure Puppet Server to instead use the JRuby 9k series (9.x.y.z). The Ruby language version supported by each JRuby 9k release is expected to evolve over time, tracking newer Ruby 2.x language versions. For example, the JRuby 9.1.8.0 release conforms to Ruby language version 2.3.1, whereas 9.1.9.0 conforms to Ruby language version 2.3.3.
 
-Note that the use of JRuby 9k with Puppet Server is still considered somewhat experimental.  We have not encountered any significant functional issues with it in internal testing so far and the JRuby community is solidly behind it as their base going forward. We have, however, observed some [targeted performance regressions](https://github.com/jruby/jruby/issues/4112#issuecomment-242504130) and increased memory usage as compared to the JRuby 1.7 series. We hope to see these issues improved upon over time, though, and at some point be able to switch to 9k as the default JRuby for Puppet Server. In the meantime, we are very interested in hearing about external users' experiences with JRuby 9k and hope to work with the JRuby community to make incremental improvements.
+Using JRuby 9k with Puppet Server is still considered somewhat experimental. We have not encountered any significant functional issues with it in internal testing so far, and the JRuby community is solidly behind it as their base going forward. We have, however, observed some [targeted performance regressions](https://github.com/jruby/jruby/issues/4112#issuecomment-242504130) and increased memory usage compared to the JRuby 1.7 series.
 
-Note also that with the ability to configure Puppet Server to use JRuby 9k for Ruby language 2.x support, we have also now removed support for the `jruby-puppet.compat-version` setting.  The JRuby community no longer supports the use of the `compat-version` setting for configuring Ruby language 2.0, which is [effectively broken](https://github.com/jruby/jruby/issues/4613) in the latest version in the JRuby 1.7 series, 1.7.27.  If the `compat-version` setting is present in your Puppet Server configuration, the puppetserver service will fail to startup with an error message which describes steps you can follow in order to configure the use of JRuby 9k instead.
+We hope to see these issues resolved over time, and at some point be able to switch to 9k as the default JRuby. In the meantime, we are very interested in hearing about external users' experiences with JRuby 9k and hope to work with the JRuby community to make incremental improvements.
 
-To provide configurable support for JRuby 1.7 vs. 9k, Puppet Server 5.0 and later packages include both versions of JRuby and their upstream dependencies.  Because of this, the size of the rpm and deb packages for Puppet Server 5.0 is significantly larger than for the Puppet Server 2.x series - around 30 MB.
+With the ability to configure Puppet Server to use JRuby 9k for Ruby language 2.x support, we have also removed support for the `jruby-puppet.compat-version` setting from `puppetserver.conf`. The JRuby community no longer supports the `compat-version` setting for configuring Ruby language 2.0, which is [effectively broken](https://github.com/jruby/jruby/issues/4613) as of JRuby 1.7.27. If the `compat-version` setting is present in your Puppet Server configuration, the `puppetserver` service exits with an error message that describes steps you can follow in order to configure the use of JRuby 9k instead.
+
+To provide configurable support for JRuby 1.7 vs. 9k, Puppet Server 5.0 and later packages include both versions of JRuby and their upstream dependencies. Because of this, the packages are about 30 MB larger than in the Puppet Server 2.x series.
 
 To change the JRuby version Puppet Server uses, you can edit the init config file.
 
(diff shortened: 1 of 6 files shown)


documentation/release_notes.markdown
```diff
@@ -10,329 +10,158 @@ canonical: "/puppetserver/latest/release_notes.html"
...
+-   [SERVER-1800](https://tickets.puppetlabs.com/browse/SERVER-1800)
 
--   [EZ-102](https://tickets.puppetlabs.com/browse/EZ-102)
+### Deprecation: Legacy `auth.conf` settings are no longer enabled by default
 
-## Puppet Server 2.7.1
+By default, Puppet Server uses the HOCON-based [`auth.conf`][auth.conf] file introduced in Puppet Server 2.4. This uses the `trapperkeeper-authorization` methods of evaluating access to HTTP endpoints, instead of the legacy Puppet `auth.conf` file.
 
-Released November 21, 2016.
+Puppet Server 5.0 completes this switch by changing the default value of the `use-legacy-auth-conf` setting in [`puppetserver.comf`](./config_file_puppetserver.markdown) from true to false.
```
camlow325: s/puppetserver.comf/puppetserver.conf

documentation/release_notes.markdown
```diff
@@ -10,329 +10,158 @@ canonical: "/puppetserver/latest/release_notes.html"
...
 
-Released November 8, 2016.
+-   [SERVER-1819](https://tickets.puppetlabs.com/browse/SERVER-1819)
+-   [PUP-6060](https://tickets.puppetlabs.com/browse/PUP-6060)
 
-This is a feature and bug-fix release of Puppet Server.
+### New Feature: Performance metrics
 
-> **Warning:** If you're upgrading from Puppet Server 2.4 or earlier and have modified `bootstrap.cfg`, `/etc/sysconfig/puppetserver`, or `/etc/default/puppetserver`, see the [Puppet Server 2.5 release notes first](#potential-breaking-issues-when-upgrading-with-a-modified-bootstrapcfg) **before upgrading** for instructions on avoiding potential failures.
+Puppet Server 5.0 includes metrics support previously released in Puppet Enterprise, including [Grafana and Graphite support](./puppet_server_metrics.markdown), both [PE-style](./metrics-api/v1/metrics_api.markdown) and [Jokola-powered](./metrics-api/v2/metrics_api.markdown) metrics API endpoints, and the developer dashboard.
```
camlow325: s/Jokola/Jolokia

documentation/release_notes.markdown
```diff
@@ -10,329 +10,158 @@ canonical: "/puppetserver/latest/release_notes.html"
...
--   [SERVER-1455](https://tickets.puppetlabs.com/browse/SERVER-1455)
--   [EZ-84](https://tickets.puppetlabs.com/browse/EZ-84)
+-   [SERVER-1723](https://tickets.puppetlabs.com/browse/SERVER-1723)
 
 ### Other changes
 
--   [SERVER-1310](https://tickets.puppetlabs.com/browse/SERVER-1310): Error messages in Puppet Server 2.5.0 use the same standard types as other Puppet projects.
--   [SERVER-1121](https://tickets.puppetlabs.com/browse/SERVER-1121): JRuby pool management code for the [Trapperkeeper Webserver Service][Trapperkeeper] is now its own open-source project, [puppetlabs/jrubyutils](https://github.com/puppetlabs/jruby-utils).
+-   [SERVER-1807](https://tickets.puppetlabs.com/browse/SERVER-1807): Puppet Server 5.0 embeds `hocon` gem version 1.2.5 in the default Puppet Server gem path, an upgrade from v1.1.3 used in Server v2.7.2.
+-   [SERVER-1772](https://tickets.puppetlabs.com/browse/SERVER-1772): Added ezbake option to specify additional uberjars to be built and installed with the main project, and allows `puppetserver` to include 2 jars for JRuby 1.7 and JRuby 9k dependencies.
```
camlow325: I think it'd be okay to leave this out of the public release notes.  End users won't care about the ezbake support that we added to enable packaging both JRuby versions and the notes you have above about adding JRuby 9k support seem to cover this fine.

documentation/release_notes.markdown
```diff
@@ -10,329 +10,158 @@ canonical: "/puppetserver/latest/release_notes.html"
...
+- [SERVER-1262](https://tickets.puppetlabs.com/browse/SERVER-1262)
+- [SERVER-1261](https://tickets.puppetlabs.com/browse/SERVER-1261)
+- [SERVER-1260](https://tickets.puppetlabs.com/browse/SERVER-1260)
+- [SERVER-1259](https://tickets.puppetlabs.com/browse/SERVER-1259)
 
-Puppet Server automatically communicates with Puppet's servers to check for updates. Puppet Server 2.7 adds the option to stop checking for updates by creating a new configuration file, [`product.conf`][product.conf], setting `check-for-updates` to false, then [restarting Puppet Server](./restarting.markdown).
+### New Feature: Optional support for JRuby 9k
 
-For more information, see the [`product.conf`][product.conf] documentation.
+Puppet Server 5.0 packages include the dependencies for both JRuby 1.7 (running Ruby language version 1.9.3) and JRuby 9k (running Ruby language version 2.3 or later). Puppet Server 5.0 uses JRuby 1.7 by default.
```
camlow325: Do you think it would be worth mentioning somewhere in the release notes that the packages are about 30 MB larger now because of the JRuby 9k dependencies that we're bundling in?  We mention this in the detailed notes for enabling JRuby 9k but it may be good to mention this here as well since this is something that users might notice / be curious about right away as they are upgrading.

camlow325:
I think it might be good to say something in the release notes about the switch from PSON to JSON as the default serialization format for agent-master communications for Puppet 5.  I imagine we'll cover that to some degree in the puppet-agent 5 release notes.  While most of the work to enable this in the server was done in the Ruby Puppet code base - not Puppet Server - most users will probably perceive this change as something which was done to some degree in the server.  Maybe a brief note with a reference to the PSON->JSON epic, [PUP-3852](https://tickets.puppetlabs.com/browse/PUP-3852), would be good?

camlow325:
I think we should point out in the release notes that we had some changes in the supported OSes (dropped Precise, Trusty, and Wily) and installation procedures (install jessie-backports for jre-8 on Jessie, SP1 or later needed for SLES 12) because of the switch from JDK 7 to JDK 8.  This is something a number of users will probably notice right away.

documentation/configuration.markdown
```diff
@@ -121,55 +123,67 @@ To change the JRuby version Puppet Server uses, you can edit the init config fil
 
 ### Enabling JRuby 9k
 
-1. Open the init config file and add the following line:
+1.  Open the init config file and add the following line:
```
adrienthebo: What is the "init config file"? If we don't explain that term elsewhere then users will probably find that confusing.

gguillotte:
@camlow325 @adrienthebo New commit addresses feedback.

documentation/release_notes.markdown
```diff
@@ -16,7 +16,46 @@ For release notes on versions of Puppet Server prior to Puppet Server 5, see [do
...
-This is a major release of Puppet Server.
+This is a major release of Puppet Server, and corresponds with the major release of Puppet 5.0, which also includes many changes and new features relevant to Puppet Server users. 
+
+### Platform changes
+
+Puppet Server 5.0 packages are available for RHEL 6 and 7, Debian 8 (Jessie), Ubuntu 16.04 (Xenial), and SLES 12 SP1.
+
+Puppet Server 5.0 is built with JDK 8, and therefore cannot run on a Java 7 runtime. Server 5.0 packages now depend exclusively upon the `openjdk-8-jre-headless` package. Since Ubuntu Precise and Trusty, and versions of Debian prior to 8 (Jessie), do not distribute that package, we do not provide Puppet Server packages for those operating systems.
+
+For SLES 12 SP1, install the `jessie-backports` repository to add access to Java 8. For details, see [Installing From Packages](./install_from_packages.markdown).
```
camlow325: s/SLES 12 SP1/Debian 8 (Jessie)

gguillotte:
@camlow325 New commit addresses feedback.

documentation/release_notes.markdown
```diff
@@ -16,7 +16,46 @@ For release notes on versions of Puppet Server prior to Puppet Server 5, see [do
...
+To retain the Puppet 4.x behavior, add the [`puppet.conf`](./configuration.markdown) setting `node_cache_terminus = write_only_yaml`. Note that `write_only_yaml` is deprecated, and users are encouraged to migrate to PuppetDB in order to retrieve node information.
+
+-   [SERVER-1819](https://tickets.puppetlabs.com/browse/SERVER-1819)
+-   [PUP-6060](https://tickets.puppetlabs.com/browse/PUP-6060)
+
+#### Deprecation/New Feature: Puppet 5 uses JSON serialization by default
+
+Previous versions of Puppet exclusively used PSON, our vendored version of `pure_json`, for serializing communication between agents and masters. Testing showed that PSON serialization's performance was significantly worse than using JSON, and JSON adds opportunities for easier and better interoperability with other tools and programming languages.
+
+As part of this, Puppet Server 5.0 requests JSON over PSON from Puppet 5.0 agents by default, and consumes JSON facts and reports provided by Puppet 5.0 agents.
```
camlow325: Technically speaking, the Puppet Server never requests JSON/PSON from the agent.  The communication is all driven from the agent instead.

Puppet 5 agents will, by default, request that the server provide responses to it in application/json or text/pson - whereas Puppet 4 agents would only request text/pson.  Puppet 5 agents will also, by default, encode parts of their request bodies (the full payload for a reports request and the facts parameter in a catalog request) in application/json - whereas Puppet 4 agents would try to use text/pson for these request bodies instead.

Puppet Server 5 has been enhanced to:

* Produce application/json responses when a Puppet 5 agent requests them.
* Consume application/json request bodies sent from the Puppet 5 agent.

Note that the Puppet 5 agent will attempt to downgrade to requesting / sending only text/pson (and not application/json) when it detects that the server it is talking to is not capable of supporting application/json - for example, a 2.x version of Puppet Server.

Does this make sense?

camlow325:
I'm 👍 on this now.  @adrienthebo - WDYT?

PR #1437, opened 2017-06-22
(MAINT) Add http-client metrics-enabled setting to puppetserver.conf doc

camlow325:
This commit adds info for the new metrics-enabled setting in the
http-client section of the puppetserver.conf file to the
config_file_puppetserver.markdown doc.


camlow325:
@gguillotte - I'd be curious what you think about this one.  Looks like we didn't add any docs for the pre-existing http-client settings other than the inline conf file examples, so I followed that pattern here for the new `metrics-enabled` setting that we'll be introducing for Puppet 5.  Does this seem reasonable?

gguillotte:
Looks fine to me.

PR #1439, opened 2017-06-22
(SERVER-1847) Update Puppetserver version to 5.0.0

adrienthebo:


adrienthebo:
@puppetlabs/puppetserver-maintainers shall we do this?

PR #1442, opened 2017-06-24
(SERVER-1055) Add automated upgrade acceptance tests

adrienthebo:


acceptance/suites/pre_suite/upgrade/10_install_puppet.rb
```diff
@@ -0,0 +1,31 @@
...
+    logger.info "PUPPET_LEGACY_VERSION is not set!"
+    logger.info "  using default value of #{default_puppet_version}"
+    puppet_version = default_puppet_version
+  end
+
+  hosts.each { |host| host.install_package('puppet-agent', '', puppet_version) }
+end
+
+step "Install legacy Puppetserver" do
+  master.install_package('puppetserver', '', '2.7.2')
```
adrienthebo: This should probably be tunable.

camlow325: This is probably a reasonable thing to do.  One other option might be to just configure the PC1 repo and do a puppetserver install, letting whatever "latest" release build is available there be laid down.  That might be advantageous in terms of not wanting/needing to bump this later.  I suppose that would give us a latest released 2.x by default but maybe we'd want to go to a latest released 5.x later on, when that repo configuration has been solidified?  Upgrading from a fixed older-than-latest released version isn't terrible either, though.  Happy to go with whatever seems best to you here.

camlow325: For the PC1 repo config as the base, that might also be beneficial for avoiding having to pin either a puppet-agent or puppetserver version to install, which could be nice.

acceptance/suites/pre_suite/upgrade/10_install_puppet.rb
```diff
@@ -0,0 +1,31 @@
...
+  hostname = on(master, 'facter hostname').stdout.strip
+  fqdn = on(master, 'facter fqdn').stdout.strip
+  dir = master.tmpdir(File.basename('/tmp'))
+
+  lay_down_new_puppet_conf( master,
+                           {"main" => { "dns_alt_names" => "puppet,#{hostname},#{fqdn}",
+                                       "verbose" => true }}, dir)
+end
+
+puppetserver_initialize_ssl
```
camlow325: As a sanity check that the legacy installation is sound, I wonder if it would be good to at least do one agent run - even if just for the agent on the master itself - before upgrading puppetserver.  WDYT?

camlow325: ah, n/m.  Looks like that already happens as part of the `puppetserver_initialize_ssl` helper [here](https://github.com/puppetlabs/puppetserver/blob/d47353d2ffde378f080844716af5cc7833a6eb2b/acceptance/lib/helper.rb#L133-L136).

camlow325:
The tests appeared to do what I expected them to and passed fine for me on a multinode beaker config - both CentOS 7, one master/agent, one agent only.  Here's the beaker command I ran:

```
PACKAGE_BUILD_VERSION=5.0.1.master.SNAPSHOT.2017.06.26T1018 bundle exec beaker --xml --debug --root-keys --repo-proxy --preserve-hosts always --config ~/multihost-centos7.cfg --type aio --keyfile ~/.ssh/id_rsa-acceptance --tests acceptance/suites/upgrade_tests --helper acceptance/lib/helper.rb --options-file acceptance/config/beaker/options.rb --load-path acceptance/lib --pre-suite acceptance/suites/pre_suite/upgrade
```

The only comment /  question I have at this point is about whether it might be okay to just install/configure the PC1 repo and install the latest puppet-agent and puppetserver from those as a base, rather than pinning versions, and then just upgrading to the latest puppetserver / puppet-agent mergely/nightly builds from there (as this is already doing).

jpinsonault:
I'm 👍 aside from Jeremy's questions

adrienthebo:
We now install the latest version of puppet-agent and puppetserver instead of hardcoded versions, and in addition I've added some platform specific fixes for various issues in the pre suite.

camlow325:
Should we target this at 5.0.x and merge up to master - in case we want to use this stuff for a 5.0.x release that we might do?

adrienthebo:
Branch retargeted.

I think we should merge this as-is; it's going to be a while before I have the ci-job-configs changes needed to get this in the pipelines but there's no reason to stall the PR on it.

camlow325:
@jpinsonault or @justinstoller - you good with this?

PR #1444, opened 2017-06-28
(docs) Fix Server, Puppet version refs.

gguillotte:
-   Use Puppet Platform name instead of Puppet Collections.
-   Use Puppet 5.0, not 2.4.
-   Fix list formatting issues.


camlow325:
@gguillotte - I think it would be good to target this PR at 5.0.x instead of master.  I tried retargeting it via the github but it looked like it was going to try to bring in other changes which looked wrong, so I reset it to master for now.

gguillotte:
@camlow325

you got it, dude

PR #1454, opened 2017-07-07
(maint) Upgrade puppet agents for upgrade acceptance testing

adrienthebo:
We need to have the latest version of the puppet agent installed on all
agent systems to ensure that the systems under test are the same version
as the tests being run.

### DIFF
--- a/acceptance/suites/upgrade_tests/10_upgrade.rb
+++ b/acceptance/suites/upgrade_tests/10_upgrade.rb
@@ -1,3 +1,4 @@
+require 'puppetserver/acceptance/compat_utils'
 test_name "Upgrade Puppetserver"
 
 install_opts = options.merge({ :dev_builds_repos => ["PC1"]})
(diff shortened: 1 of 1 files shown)


acceptance/suites/upgrade_tests/10_upgrade.rb
```diff
@@ -23,6 +26,12 @@
   on(master, puppet("resource service puppetserver ensure=running"))
 end
 
+step "Upgrade Puppet agents" do
+  nonmaster_agents.each do |agent|
```
justinstoller: Why don't we want to also run this on the master?

camlow325: It's probably redundant since the puppetserver upgrade should transitively upgrade puppet-agent as well, since the newer puppetserver will have a newer puppet-agent dependency.  Might be nice to not upgrade the agent on the master so that we can prove the agent does get upgraded as expected just by upgrading puppetserver.

PR #1455, opened 2017-07-07
(MAINT) Bump puppet-agent to 33cc892e8f

adrienthebo:
This commit bumps the puppet submodule and puppet-agent versions for
testing to f24e03659 and 4.99.0.403.g33cc892, respectively.


justinstoller:
This looks good, content-wise. I have a question though, will this cause conflicts when we do an eventual merge down from master into the branch, and if so, frankly, do we even care (assuming it will be pretty easy to resolve)?

adrienthebo:
This is a cherry pick, and hopefully it won't cause conflicts while merging down, but if it would cause conflicts then basically everything else will cause conflicts at this point.

justinstoller:
👍

PR #1457, opened 2017-07-07
(DO NOT MERGE) (RE-8458) Update tests to use the puppet repos

melissa:
In order to test against the right set of packages, we need to update
these tests to hit the new repos. Rather than pin to puppet 5, we're
just going to pin to the rolling puppet repos. This will change
automatically to puppet6 once we change these repos to point to puppet6.

### DIFF
--- a/acceptance/suites/pre_suite/compat_common/30_install_dev_repos.rb
+++ b/acceptance/suites/pre_suite/compat_common/30_install_dev_repos.rb
@@ -1,4 +1,4 @@
-install_opts = options.merge({ :dev_builds_repos => ["PC1"]})
+install_opts = options.merge({ :dev_builds_repos => ["puppet"]})
 
 if test_config[:puppetserver_install_type] == :package
   package_build_version = ENV['PACKAGE_BUILD_VERSION']
--- a/acceptance/suites/pre_suite/upgrade/10_install_puppet.rb
+++ b/acceptance/suites/pre_suite/upgrade/10_install_puppet.rb
@@ -1,5 +1,5 @@
-step "Install PC1 repository" do
-  install_puppetlabs_release_repo(hosts, 'pc1')
+step "Install puppet repository" do
+  install_puppetlabs_release_repo(hosts, 'puppet')
 end
 
 hosts.each { |agent|
--- a/acceptance/suites/pre_suite/foss/30_install_dev_repos.rb
+++ b/acceptance/suites/pre_suite/foss/30_install_dev_repos.rb
@@ -1,4 +1,4 @@
-install_opts = options.merge({ :dev_builds_repos => ["PC1"]})
+install_opts = options.merge({ :dev_builds_repos => ["puppet"]})
 
 if test_config[:puppetserver_install_type] == :package
   package_build_version = ENV['PACKAGE_BUILD_VERSION']
--- a/acceptance/suites/pre_suite/foss/35_install_pdb.rb
+++ b/acceptance/suites/pre_suite/foss/35_install_pdb.rb
@@ -1,7 +1,7 @@
 matching_puppetdb_platform = puppetdb_supported_platforms.select { |r| r =~ master.platform }
 skip_test unless matching_puppetdb_platform.length > 0
 
-install_opts = options.merge( { :dev_builds_repos => ["PC1"] })
+install_opts = options.merge( { :dev_builds_repos => ["puppet"] })
 repo_config_dir = 'tmp/repo_configs'
 
 step "Install PuppetDB repository" do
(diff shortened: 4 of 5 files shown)


acceptance/suites/pre_suite/compat_common/30_install_dev_repos.rb
```diff
@@ -1,4 +1,4 @@
-install_opts = options.merge({ :dev_builds_repos => ["PC1"]})
+install_opts = options.merge({ :dev_builds_repos => ["puppet"]})
```
melissa: The only thing I'm unsure of here is how beaker handles this. I haven't looked closely at this method to make sure this will work

acceptance/suites/pre_suite/upgrade/10_install_puppet.rb
```diff
@@ -1,5 +1,5 @@
-step "Install PC1 repository" do
-  install_puppetlabs_release_repo(hosts, 'pc1')
+step "Install puppet repository" do
+  install_puppetlabs_release_repo(hosts, 'puppet')
```
camlow325: @adrienthebo Correct me if I'm wrong, but I think the way these tests are currently built, we're expecting to do an upgrade from Puppet 4 / Puppet Server 2.x to Puppet / Puppet Server 5 (latest).  If we were to change this line to use the new "puppet" repo for Puppet 5 rather than "pc1", I wonder if these tests would break.  WDYT?

camlow325: I tried running this locally with an EL-7 based job configuration.  The job failed at this step with the following error:

````
  dzxqhy95cr51jg5.delivery.puppetlabs.net 13:14:24$ rpm --replacepkgs -Uvh http://yum.puppetlabs.com/puppetlabs-release-puppet-el-7.noarch.rpm
    curl: (22) The requested URL returned error: 404 Not Found
    error: skipping http://yum.puppetlabs.com/puppetlabs-release-puppet-el-7.noarch.rpm - transfer failed
    Retrieving http://yum.puppetlabs.com/puppetlabs-release-puppet-el-7.noarch.rpm
````

It doesn't appear that files named like "puppetlabs-release-puppet-*" are present under http://yum.puppetlabs.com yet.  Have these repo configs not been created yet?

acceptance/suites/pre_suite/compat_common/30_install_dev_repos.rb
camlow325: FWIW, when I ran the Puppet Server smoke tests against this PR, the tests all passed for me.  I don't know with this line and the related ones below it do exactly with beaker but the changes don't appear to adversely affect the job, in any event.  Maybe this doesn't typically matter in the Puppet Server pipeline anyway since we have it download internal repo configs directly from the builds.puppetlabs server?

melissa: I'm digging into beaker re: this variable, and it looks like this isn't even used.

acceptance/suites/pre_suite/upgrade/10_install_puppet.rb
melissa: Yes, this is broken. These repos do exist, but they have a different format. My comment above, it turns out that value isn't used anywhere, which is why it doesn't break anything. This, on the other hand, is used, which is why it does break things. 

I'm hesitating because this method is removed in the latest release of beaker. What would it take for you to update to use the latest beaker version for testing?

camlow325: @melissa I don't know what specific blockers we might have in moving to the latest beaker version.  We'd probably just have to try it out and make changes where necessary for anything that might break.  I could submit a separate ticket for us to investigate that and we can block this PR the resolution of the new ticket, if that makes sense to you.  WDYT?

melissa: Yes, that seems like a reasonable path forward. I'm talking to @kevpl at the moment, and it looks like any supporting methods around installing the release packages have been removed. So that alone will take some thought if you are going to upgrade beaker

melissa: https://github.com/puppetlabs/beaker-puppet/pull/9 will address this change and allow beaker to set up the new puppet5 repos

melissa:
This PR is blocked on https://github.com/puppetlabs/beaker-puppet/pull/9 and on puppetserver updating the version of beaker it uses

melissa:
The PR I opened against beaker-puppet allows users to set up both the old pc1 repos and the new style puppet5/puppet repos

justinstoller:
@melissa I think this branch, 5.0.x, is going to go away. Can you retarget this towards master?

melissa:
Changed. I still need to get my PR against beaker-puppet merged, and we need to get puppetserver using a more recent version of beaker

PR #1458, opened 2017-07-10
(SERVER-1870) create Rake tasks to update puppet submodule and acceptance pins

er0ck:
* find latest passing puppet-agent
* update puppet submodule to the puppet SHA contained in the
puppet-agent build
* update beaker options file with puppet and puppet-agent versions

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,13 +1,20 @@
 require 'open3'
+require 'open-uri'
+require 'json'
+require 'pp'
 
 PROJECT_ROOT = File.dirname(__FILE__)
 ACCEPTANCE_ROOT = ENV['ACCEPTANCE_ROOT'] ||
   File.join(PROJECT_ROOT, 'acceptance')
+BEAKER_OPTIONS_FILE = File.join(ACCEPTANCE_ROOT, 'config', 'beaker', 'options.rb')
 PUPPET_SRC = File.join(PROJECT_ROOT, 'ruby', 'puppet')
 PUPPET_LIB = File.join(PROJECT_ROOT, 'ruby', 'puppet', 'lib')
 PUPPET_SPEC = File.join(PROJECT_ROOT, 'ruby', 'puppet', 'spec')
 FACTER_LIB = File.join(PROJECT_ROOT, 'ruby', 'facter', 'lib')
 PUPPET_SERVER_RUBY_SRC = File.join(PROJECT_ROOT, 'src', 'ruby', 'puppetserver-lib')
+PUPPET_SUBMODULE_PATH = File.join('ruby','puppet')
+CURRENT_BRANCH = '5.0.x'
+JENKINS_BRANCH = CURRENT_BRANCH
 
 TEST_GEMS_DIR = File.join(PROJECT_ROOT, 'vendor', 'test_gems')
 TEST_BUNDLE_DIR = File.join(PROJECT_ROOT, 'vendor', 'test_bundle')
@@ -44,7 +51,7 @@ def basic_smoke_test(package_version)
   beaker += " --preserve-hosts always"
   beaker += " --type aio"
   beaker += " --helper acceptance/lib/helper.rb"
-  beaker += " --options-file acceptance/config/beaker/options.rb"
+  beaker += " --options-file #{BEAKER_OPTIONS_FILE}"
   beaker += " --load-path acceptance/lib"
   beaker += " --config acceptance/scripts/hosts.cfg"
   beaker += " --keyfile ~/.ssh/id_rsa-acceptance"
@@ -62,13 +69,101 @@ def re_run_basic_smoke_test
   beaker += " --preserve-hosts always"
   beaker += " --type aio"
   beaker += " --helper acceptance/lib/helper.rb"
-  beaker += " --options-file acceptance/config/beaker/options.rb"
+  beaker += " --options-file #{BEAKER_OPTIONS_FILE}"
   beaker += " --load-path acceptance/lib"
   beaker += " --config acceptance/scripts/hosts.cfg"
   beaker += " --keyfile ~/.ssh/id_rsa-acceptance"
   beaker += " --tests acceptance/suites/tests/00_smoke"
 
   sh beaker
+end
+
+def jenkins_passing_json_parsed
+  jenkins_url = "https://jenkins-master-prod-1.delivery.puppetlabs.net/view/" \
+    "puppet-agent/view/#{JENKINS_BRANCH}/view/Suite/job/" \
+    "platform_puppet-agent_intn-van-promote_suite-daily-promotion-#{JENKINS_BRANCH}" \
+    "/lastSuccessfulBuild/api/json"
+  uri = URI.parse(jenkins_url)
+  begin
+    # DO NOT use uri-open if accepting user input for the uri
+    #   we've done some simple correction here,
+    #   but not enough to cleanse malicious user input
+    jenkins_result = uri.open(redirect: false)
+  rescue OpenURI::HTTPError => e
+    abort "ERROR: Could not get lastSuccessfulBuild for #{JENKINS_BRANCH} of puppet-agent: '#{e.message}'"
+  end
+
+  begin
+    jenkins_result_parsed = JSON.parse(jenkins_result.read)
+  rescue JSON::ParserError => e
+    abort "ERROR: Could not get lastSuccessfulBuild's valid json for #{JENKINS_BRANCH}: '#{e.message}'"
+  end
+
+  begin
+    jenkins_result_parameters = jenkins_result_parsed['actions'][0]['parameters']
+  rescue
+    abort "ERROR: Could not get lastSuccessfulBuild's actions or parameters for #{JENKINS_BRANCH}"
+  end
+
+  jenkins_result_parsed['actions'][0]['parameters']
+end
+
+def lookup_passing_puppetagent_sha(my_jenkins_passing_json)
+  begin
+    my_jenkins_passing_json.find{|x| x['name'] == 'SUITE_COMMIT'}['value']
+  rescue
+    abort "ERROR: Could not get lastSuccessfulBuild's SUITE_COMMIT value for #{JENKINS_BRANCH}"
+  end
+end
+def lookup_passing_puppet_sha(my_jenkins_passing_json)
+  begin
+    my_jenkins_passing_json.find{|x| x['name'] == 'puppet_COMPONENT_COMMIT'}['value']
+  rescue
+    abort "ERROR: Could not get lastSuccessfulBuild's puppet_COMPONENT_COMMIT value for #{JENKINS_BRANCH}"
+  end
+end
+
+def git_passing_puppet_version
+   #FIXME: this should be updated when the package yaml file contains the metadata we need
+   #  we have to replace the hyphens with dots because, vanagon
+  `cd #{PUPPET_SUBMODULE_PATH}; git describe`.strip.gsub(/-/,'.')
+end
+
+def replace_puppet_pins(passing_puppetagent_sha, passing_puppet_version)
+  # read beaker options hash from its file
+  puts("replacing puppet sha and version in #{BEAKER_OPTIONS_FILE} " \
+       "with agent sha: #{passing_puppetagent_sha} and puppet version: #{passing_puppet_version}")
+  beaker_options_from_file = eval(File.read(BEAKER_OPTIONS_FILE))
+  # add puppet version values
+  beaker_options_from_file[:puppet_version]       = passing_puppet_version
+  beaker_options_from_file[:puppet_build_version] = passing_puppetagent_sha
+  File.write(BEAKER_OPTIONS_FILE, beaker_options_from_file.pretty_inspect)
+end
+
+namespace :puppet_submodule do
+  desc 'update puppet submodule commit'
+  task :update_puppet_version do
+    #  ensure we fetch here, or the describe done later could be wrong
+    my_jenkins_passing_json = jenkins_passing_json_parsed
+    git_checkout_command = "cd #{PUPPET_SUBMODULE_PATH} && git fetch origin && " \
+      "git checkout #{lookup_passing_puppet_sha(my_jenkins_passing_json)}"
+    puts("checking out known passing puppet version in submodule: `#{git_checkout_command}`")
+    system(git_checkout_command)
+    # replace puppet version and sha pins in beaker options file
+    replace_puppet_pins(lookup_passing_puppetagent_sha(my_jenkins_passing_json), git_passing_puppet_version)
+  end
+  desc 'commit and push; CAUTION: WILL commit and push, upstream, local changes to the puppet submodule and acceptance options'
+  task :commit_push do
+    git_commit_command = "git checkout #{CURRENT_BRANCH} && git add #{PUPPET_SUBMODULE_PATH} " \
+      "&& git add #{BEAKER_OPTIONS_FILE} && git commit -m '(maint) update puppet submodule version and pins'"
+    git_push_command = "git checkout #{CURRENT_BRANCH} && git push origin HEAD:#{CURRENT_BRANCH}"
+    puts "committing submodule and pins via: `#{git_commit_command}`"
+    system(git_commit_command)
+    puts "pushing submodule and pins via: `#{git_push_command}`"
+    system(git_push_command)
+  end
+  desc 'update puppet versions and commit and push; CAUTION: WILL commit and push, upstream, local changes to the puppet submodule and acceptance options'
+  task :update_puppet_version_w_push => [:update_puppet_version, :commit_push]
 end
 
 namespace :spec do
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -23,6 +23,8 @@ module PuppetServerExtensions
                          nil, "Puppet Server Version",
                          "PUPPETSERVER_VERSION", nil, :string)
 
+    # This value now comes from the beaker options file, or PUPPET_VERSION env var
+    #   the beaker options file can have this value updated via the rake task
     puppet_version = get_option_value(options[:puppet_version],
                          nil, "Puppet Version", "PUPPET_VERSION",
                          "5.0.0.72.gf7c6ceb",
@@ -31,6 +33,8 @@ module PuppetServerExtensions
 
     # puppet-agent version corresponds to packaged development version located at:
     # http://builds.delivery.puppetlabs.net/puppet-agent/
+    # This value now come from the beaker options file, or PUPPET_BUILD_VERSION env var
+    #   the beaker options file can have this value updated via the rake task
     puppet_build_version = get_option_value(options[:puppet_build_version],
                          nil, "Puppet Agent Development Build Version",
                          "PUPPET_BUILD_VERSION",


Rakefile
```diff
@@ -71,6 +75,64 @@ def re_run_basic_smoke_test
...
+end
+
+namespace :puppet_submodule do
+  puppet_pins_file_path = File.join('acceptance','lib','helper.rb')
+  desc 'update puppet submodule commit'
+  task :update_puppet_version do
+    #hard code current branch
+    #  it's too difficult to reliably find trunk of a feature branch
+    #  ugh... now jenkins job naming is out of date with new component branches
+    current_branch_trunk = '5.0.x'
```
er0ck: when this gets merged up to master, this should be changed to `'master'`

er0ck:
this could use more error handling on the http calls.
it could use some tests.  but it either works or it doesn't and won't be used to fail down stream jobs

er0ck:
@justinstoller thoughts on this?

Rakefile
```diff
@@ -71,6 +75,63 @@ def re_run_basic_smoke_test
...
+def get_passing_puppet_version(current_branch_trunk)
+   #FIXME: this should be updated when the package yaml file contains the metadata we need
+   #  we have to replace the hyphens with dots because, vanagon
+  `cd #{PUPPET_SUBMODULE_PATH}; git describe`.strip.gsub(/-/,'.')
+end
+
+def replace_puppet_pins(passing_puppetagent_sha, passing_puppet_version)
+  # read beaker options hash from its file
+  beaker_options_file_path = 'acceptance/config/beaker/options.rb'
+  puts("replacing puppet sha and version in #{beaker_options_file_path}, with agent sha: #{get_passing_puppetagent_sha(current_branch_trunk)} and puppet version: #{get_passing_puppet_version(current_branch_trunk).to_s}")
```
justinstoller: I think we could use the passed in values here instead of calling to the accessors methods and returning their memoized values.

Rakefile
```diff
@@ -71,6 +75,63 @@ def re_run_basic_smoke_test
   sh beaker
 end
 
+def get_passing_json(current_branch_trunk)
```
justinstoller: I have seen the word "trunk" used in SVN to denote what would be the equivalent of "master" for us. I might have been using trunk wrong in the past but I think that "current_branch" would provide the similar level of understanding in the code here.

Rakefile
```diff
@@ -71,6 +75,63 @@ def re_run_basic_smoke_test
...
+  jenkins_url = "https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/#{current_branch_trunk}/view/Suite/job/platform_puppet-agent_intn-van-promote_suite-daily-promotion-#{current_branch_trunk}/lastSuccessfulBuild/api/json"
+  @memoized_passing_json ||= Net::HTTP.get(URI.parse(jenkins_url))
```
justinstoller: I would avoid memoizing a value that is dependent on the parameter passed in, unless you check that the value passed in is the same as the one used to compute the memoized value.

It might be nice to hit Jenkins in the rake task and then pass the JSON, or deserialized hash, into the methods that would otherwise call this also at within the top of the rake task, then passing their values into the replace_pins method, removing the need to memoize in the first place.

justinstoller: I don't see a place were we get the JSON and then act on it directly. It seems like everywhere we call this we then parse it with `JSON.parse`. In that case it might be valuable to parse it here and when robust-ifying we can then add the potential parsing error handling in one place.

Rakefile
```diff
@@ -71,6 +75,63 @@ def re_run_basic_smoke_test
...
+end
+def get_passing_puppetagent_sha(current_branch_trunk)
+  get_passing_json(current_branch_trunk)
+  JSON.parse(@memoized_passing_json)['actions'][0]['parameters'].find{|x| x['name'] == 'SUITE_COMMIT'}['value']
```
justinstoller: Calling a method to mutate a variable and then directly operating on that variable is something that I would probably avoid even in an OO language. `get_passing_json` returns the output of `@memoized_passing_json` so I would either save the return of `get_passing_json` as a local value, or call it directly in the `JSON.parse` parameter list. That way `get_passing_*_sha` methods aren't coupled to the internal structure of how `get_passing_sha` is memoizing the data.

Rakefile
```diff
@@ -71,6 +75,63 @@ def re_run_basic_smoke_test
...
+    git_checkout_command = "cd #{PUPPET_SUBMODULE_PATH} && git fetch origin && git checkout #{get_passing_puppet_sha(current_branch_trunk)}"
+    puts("checking out known passing puppet version in submodule: `#{git_checkout_command}`")
+    system(git_checkout_command)
+    # replace puppet version and sha pins in:
+    replace_puppet_pins(get_passing_puppetagent_sha(current_branch_trunk), get_passing_puppet_version(current_branch_trunk))
+  end
+  desc 'commit and push; CAUTION: WILL commit and push, upstream, local changes to the puppet submodule and acceptance helper'
+  task :commit_push do
+    #  commit and push
+    puts("commiting and pushing submodule and pins via: `git add #{PUPPET_SUBMODULE_PATH}; git add #{puppet_pins_file_path}; git commit -m '(maint) update puppet submoudule version and pins'`")
```
justinstoller: I would probably put the git commands in the same kind of local variable here as you do above for the `git_checkout_command`, so the thing reported and the thing done are always in sync.

Rakefile
```diff
@@ -71,6 +75,63 @@ def re_run_basic_smoke_test
...
+    system("git checkout #{current_branch_trunk} && git add #{PUPPET_SUBMODULE_PATH} && git add #{puppet_pins_file_path} && git commit -m '(maint) update puppet submoudule version and pins' && git push origin")
```
justinstoller: We probably shouldn't assume we're tracking the remote `current_branch_trunk` and specify it when pushing, ie `git push origin HEAD:current_branch_trunk`.

justinstoller: If we're using `current_branch_trunk` in multiple places I would probably then turn it into a constant and move it to the top of the file.

acceptance/lib/helper.rb
```diff
@@ -23,18 +23,22 @@ def self.initialize_config(options)
...
                          nil, "Puppet Version", "PUPPET_VERSION",
-                         "4.99.0.403.g33cc892",
+                         nil,
                          :string) ||
                          get_puppet_version
 
     # puppet-agent version corresponds to packaged development version located at:
     # http://builds.delivery.puppetlabs.net/puppet-agent/
+    # This value now come from the beaker options file
+    #   the beaker options file can have this value updated via the rake task
```
justinstoller: This is something that might be out of the scope of this PR (ie you don't have to clean up every little thing in our acceptance helpers that you could after this change), but do we need these any more? Should we simply remove them from here or does that break something? Do you think this is valuable as documentation?

Rakefile
er0ck: yeah, you are right.  damn this is sloppy.

acceptance/lib/helper.rb
er0ck: this comment is probably not quite clear enough.  

i believe `#get_option_value` finds a final value from various places.  i changed the default value to nil, as the default should probably come from the options file `options[:puppet_build_version]` below.  but yes, it also provides documentation as well as an env_var override, `PUPPET_BUILD_VERSION` below.

er0ck:
@justinstoller i believe all the review comments have been addressed.  thanks!

Rakefile
```diff
@@ -71,6 +76,78 @@ def re_run_basic_smoke_test
...
+    "/lastSuccessfulBuild/api/json"
+  uri = URI.parse(jenkins_url)
+  begin
+    # DO NOT use uri-open if accepting user input for the uri
+    #   we've done some simple correction here,
+    #   but not enough to cleanse malicious user input
+    jenkins_result = uri.open(redirect: false)
+  rescue OpenURI::HTTPError => e
+    status = e.io.status[0] # => 3xx, 4xx, or 5xx
+    raise RuntimeError "ERROR: Could not get lastSuccessfulBuild for #{CURRENT_BRANCH} of puppet-agent: '#{e.message}'"
```
justinstoller: missing a comma, this causes:
```
sellout:puppetserver justin$ bundle exec rake puppet_submodule:update_puppet_version
rake aborted!
NoMethodError: undefined method `RuntimeError' for main:Object
/Users/justin/Backup/code/work/server/puppetserver/Rakefile:92:in `rescue in jenkins_passing_json_parsed'
/Users/justin/Backup/code/work/server/puppetserver/Rakefile:85:in `jenkins_passing_json_parsed'
/Users/justin/Backup/code/work/server/puppetserver/Rakefile:132:in `block (2 levels) in <top (required)>'
/Users/justin/Backup/code/work/server/puppetserver/.bundle/gems/gems/rake-12.0.0/exe/rake:27:in `<top (required)>'
```

Rakefile
```diff
@@ -71,6 +76,78 @@ def re_run_basic_smoke_test
...
+    "platform_puppet-agent_intn-van-promote_suite-daily-promotion-#{CURRENT_BRANCH}" \
+    "/lastSuccessfulBuild/api/json"
+  uri = URI.parse(jenkins_url)
+  begin
+    # DO NOT use uri-open if accepting user input for the uri
+    #   we've done some simple correction here,
+    #   but not enough to cleanse malicious user input
+    jenkins_result = uri.open(redirect: false)
+  rescue OpenURI::HTTPError => e
+    status = e.io.status[0] # => 3xx, 4xx, or 5xx
```
justinstoller: this variable is unused

Rakefile
er0ck: weird

Rakefile
```diff
@@ -71,6 +76,78 @@ def re_run_basic_smoke_test
...
+       "with agent sha: #{passing_puppetagent_sha} and puppet version: #{passing_puppet_version}")
+  beaker_options_from_file = eval(File.read(beaker_options_file_path))
+  # add puppet version values
+  beaker_options_from_file[:puppet_version]       = passing_puppet_version
+  beaker_options_from_file[:puppet_build_version] = passing_puppetagent_sha
+  File.write(beaker_options_file_path, beaker_options_from_file.pretty_inspect)
+end
+
+namespace :puppet_submodule do
+  puppet_pins_file_path = File.join('acceptance','lib','helper.rb')
```
justinstoller: Do we want this to be the beaker_options_file_path?

er0ck: yep.  sheesh.  i'll take another pass through this thing

er0ck:
so this all works.  but it assumes origin is the fork to which to push as this will be the name of the remote for the  puppetlabs fork in jenkins.  but this doesn't help those working locally if puppetlabs is named upstream or puppetlabs or 🐩 

thoughts?

Rakefile
```diff
@@ -1,13 +1,19 @@
...
 ACCEPTANCE_ROOT = ENV['ACCEPTANCE_ROOT'] ||
   File.join(PROJECT_ROOT, 'acceptance')
+BEAKER_OPTIONS_FILE = File.join(ACCEPTANCE_ROOT, 'config', 'beaker', 'options.rb')
 PUPPET_SRC = File.join(PROJECT_ROOT, 'ruby', 'puppet')
 PUPPET_LIB = File.join(PROJECT_ROOT, 'ruby', 'puppet', 'lib')
 PUPPET_SPEC = File.join(PROJECT_ROOT, 'ruby', 'puppet', 'spec')
 FACTER_LIB = File.join(PROJECT_ROOT, 'ruby', 'facter', 'lib')
 PUPPET_SERVER_RUBY_SRC = File.join(PROJECT_ROOT, 'src', 'ruby', 'puppetserver-lib')
+PUPPET_SUBMODULE_PATH = File.join('ruby','puppet')
+CURRENT_BRANCH = '5.0.x'
```
er0ck: this will need changing to `'master'` upon merge-up

justinstoller:
> but it assumes origin is the fork to which to push as this will be the name of the remote for the puppetlabs fork in jenkins.... ...thoughts?

I'm alright with hardcoding the remote as origin and the branch name at this point. Ideally it would be parameterized with autodetected, correct defaults. But I'm hesitant to say that'd give us a good r.o.i. at this point.

---

I tested this by
  * temporarily changing my fork to be `origin`
  * checking out a branch erictest off of master
  * merging this into it
  * checkout out the last commit so we were in a detatched head state (which IIRC is what Jenkins is in)
  * changing the value of `CURRENT_BRANCH` to be `erictest`
  * the jenkins url to hit the lastSuccessful for `master`

I was then able to run the `update_puppet_version` and `commit_push` tasks successfully (see [the push here](https://github.com/justinstoller/puppetserver/commits/erictest).

That's a similar enough reproduction of what would happen in Jenkins to satisfy me as far as functional testing this goes.

---

Thanks for all your effort on this, @er0ck, I really appreciate it!

@adrienthebo Head ups, I only ran this in isolation. I don't know how it will interact with the pipeline changes you've been working on, or if they necessitate some additional functional testing.

er0ck:
i also tested it by repeatedly running the `commit_and_push` task with no changes to ensure git still would do what i expected and not push a new commit.  also re-running the `update` task does nothing so a subsequent commit and push also does nothing.
i also tested the various exception handling by introducing bad values upstream.

there is no dedicated testing for this.  at somepoint we should probably move this to some ruby module outside the Rakefile.  it will declutter the rakefile as well as make it easier to keep some spec tests around if we find this valuable.

Rakefile
```diff
@@ -71,6 +77,91 @@ def re_run_basic_smoke_test
...
+    puts("checking out known passing puppet version in submodule: `#{git_checkout_command}`")
+    system(git_checkout_command)
+    # replace puppet version and sha pins in beaker options file
+    replace_puppet_pins(lookup_passing_puppetagent_sha(my_jenkins_passing_json), git_passing_puppet_version)
+  end
+  desc 'commit and push; CAUTION: WILL commit and push, upstream, local changes to the puppet submodule and acceptance options'
+  task :commit_push do
+    git_commit_push_command = "git checkout #{CURRENT_BRANCH} && git add #{PUPPET_SUBMODULE_PATH} " \
+      "&& git add #{BEAKER_OPTIONS_FILE} && git commit -m '(maint) update puppet submodule version and pins' && git push origin HEAD:#{CURRENT_BRANCH}"
+    puts "commiting and pushing submodule and pins via: `#{git_commit_push_command}`"
```
adrienthebo: Small nit - "commiting" -> "committing"

Rakefile
```diff
@@ -71,6 +77,91 @@ def re_run_basic_smoke_test
   sh beaker
 end
 
+def jenkins_passing_json_parsed
```
adrienthebo: These look to be duplicating https://github.com/puppetlabs/puppetserver/blob/master/acceptance/lib/helper.rb#L290-L312 , maybe we should coalesce these helper methods.

Rakefile
```diff
@@ -71,6 +77,91 @@ def re_run_basic_smoke_test
...
+    my_jenkins_passing_json = jenkins_passing_json_parsed
+    git_checkout_command = "cd #{PUPPET_SUBMODULE_PATH} && git fetch origin && " \
+      "git checkout #{lookup_passing_puppet_sha(my_jenkins_passing_json)}"
+    puts("checking out known passing puppet version in submodule: `#{git_checkout_command}`")
+    system(git_checkout_command)
+    # replace puppet version and sha pins in beaker options file
+    replace_puppet_pins(lookup_passing_puppetagent_sha(my_jenkins_passing_json), git_passing_puppet_version)
+  end
+  desc 'commit and push; CAUTION: WILL commit and push, upstream, local changes to the puppet submodule and acceptance options'
+  task :commit_push do
```
adrienthebo: Can we decouple the commit and push actions? That way I can test the entire bump/commit path without affecting the live github repo.

Rakefile
er0ck: oooh.  good catch!  i didn't know about those

er0ck: on second thought..  it's gonna take some refactoring to combine these helpers.  
the helpers in the acceptance lib don't have any tests, so i don't really want to mess with them in this timeframe.  

can we see if this works as is, and ticket the refactor of this stuff for later?

Rakefile
er0ck: do you need separate rake tasks for that?  or are you cool with just two `system` calls and you can test by commenting them out?

er0ck:
another thing this assumes is `CURRENT_BRANCH`

i couldn't test this as-is when IN `5.0.x` because then i don't have the changes.  if i use update_puppet_version it makes changes and leaves them in whatever my current branch is.  
then the commit and push don't work as we're checking out the 5.0.x branch and git complains about stashing them.

also, do people ever want to run this on a feature branch, to run acceptance against latest puppet?

justinstoller:
We talked irl, but to update this thread.

> do people ever want to run this on a feature branch, to run acceptance against latest puppet?

I think once this is going the hope will be that we're updating each branch with latest passing puppet daily, so short lived branches should never be too out of date.

For long lived branches rebasing onto on to the head of the upstream merge target (assuming it has had its submodule bumped) has been recommended at Puppet in the past.

Long lived branches, tightly coupled to ongoing puppet changes, when the merge target has been broken (and so there's no recent "lastPassing" to bumped to) has happened in the past, but is emblematic of a larger problem and I don't feel bad about folks having to manually commit a submodule bump in that case to continue their wip.

Rakefile
```diff
@@ -1,13 +1,20 @@
...
   File.join(PROJECT_ROOT, 'acceptance')
+BEAKER_OPTIONS_FILE = File.join(ACCEPTANCE_ROOT, 'config', 'beaker', 'options.rb')
 PUPPET_SRC = File.join(PROJECT_ROOT, 'ruby', 'puppet')
 PUPPET_LIB = File.join(PROJECT_ROOT, 'ruby', 'puppet', 'lib')
 PUPPET_SPEC = File.join(PROJECT_ROOT, 'ruby', 'puppet', 'spec')
 FACTER_LIB = File.join(PROJECT_ROOT, 'ruby', 'facter', 'lib')
 PUPPET_SERVER_RUBY_SRC = File.join(PROJECT_ROOT, 'src', 'ruby', 'puppetserver-lib')
+PUPPET_SUBMODULE_PATH = File.join('ruby','puppet')
+CURRENT_BRANCH = '5.0.x'
+JENKINS_BRANCH = CURRENT_BRANCH
```
justinstoller: :bow:

justinstoller:
needs rebase

er0ck:
@adrienthebo i think this is ready to go

justinstoller:
Tested the latest changes while pushing to https://github.com/justinstoller/puppetserver/tree/testerict and all seemed to work okay.

er0ck:
did this get edited and merged up to master?

justinstoller:
Not that I know of?

er0ck:
on some teams, they who do the merge do the merge-up.  but since this needs an edit.  it should probably be PR'd

justinstoller:
@er0ck gotcha, traditionally in Server we've done PRs for merge ups or merge downs periodically (end of feature, end of sprint, release prep) but don't usually try to do merge ups or downs for per-PR work.

Thank you for opening https://github.com/puppetlabs/puppetserver/pull/1465 !

er0ck:
probably a good idea.  pushing upstream is a typo-vector waiting to happen.

PR #1459, opened 2017-07-10
(SERVER-1842) HTTP client metrics docs.

gguillotte:
Adds HTTP client metrics docs, and links to them from the nav and existing metrics docs.

PR #1460, opened 2017-07-11
(SERVER-1771) Scaling & perf docs.

gguillotte:
Adds scaling and performance docs, and integrates them with (or links them to) the new metrics docs.


documentation/puppet_server_metrics_performance.markdown
```diff
@@ -0,0 +1,78 @@
...
+
+[metrics]: ./puppet_server_metrics.markdown
+[tuning guide]: ./tuning_guide.markdown
+[metrics API]: ./metrics-api/v1/metrics_api.markdown
+[status API]: ./status-api/v1/services.markdown
+[developer dashboard]: #using-the-developer-dashboard
+[Graphite]: https://graphiteapp.org
+[Grafana]: http://grafana.org
+[sample Grafana dashboard]: ./sample-puppetserver-metrics-dashboard.json
+[`grafanadash`](https://forge.puppet.com/cprice404/grafanadash)
```
camlow325: Should this one and the following one have back-ticks around the words inside of the brackets?  That may be fine; I'm not really sure how the formatting works.

documentation/puppet_server_metrics_performance.markdown
```diff
@@ -0,0 +1,78 @@
...
+
+If wait time increases but borrow time stays the same, your Server infrastructure might be serving too many agents. This indicates that Server can easily handle requests but is receiving too many at once to keep up.
+
+If both wait and borrow times are increasing, something else on your server is causing requests to take longer to process. The longer borrow times suggest that Puppet Server is struggling more than before to process requests, which has a cascading effect on wait times. Correlate borrow time increases with other events whenever possible to isolate what activities might cause them, such as a Puppet code change.
+
+If you are setting up Puppet Server for the first time, start by increasing your Server infrastructure's capacity through additional JRubies (if your server has spare CPU and memory resources) or compile masters until your free JRubies are at least 0 and average number of free JRubies are at least 1. Once your system can handle its request volume, you can start looking into more specific performance improvements.
```
camlow325: Would it be better to say "until your free JRubies are above 0" rather than "until your free JRubies are at least 0"?

documentation/puppet_server_metrics_performance.markdown
```diff
@@ -0,0 +1,78 @@
...
+
+#### Adding more JRubies
+
+If you need to add JRubies, remember that Puppet Server is tuned by default to use one fewer than your total number of CPUs, with a maximum of 4 CPUs, for the number of available JRubies. You can change this by setting `max-active-instances` in [`puppetserver.conf`][], under the `jruby-puppet` section. Don' set this value to something greater than the number of CPUs, because if there aren't enough processes available, it won't matter how many JRubies in the pool there are — the server won't be able to use them.
```
camlow325: s/Don'/Don't

camlow325: > Don' set this value to something greater than the number of CPUs, because if there aren't enough processes available, it won't matter how many JRubies in the pool there are — the server won't be able to use them.

This can sound like good advice initially but there have been situations where users have set the `max-active-instances` to a value much greater than the number of CPUs and have seen performance improvements as a result.  The most common case for this, I believe, is when a catalog includes lots of file resources.  In those cases, the delivery of the file content payload, unlike catalog compilations, is more often I/O bound than CPU bound so having the ability for a number of those to run concurrently can result in greater throughput - albeit with the tradeoff of potentially having significantly greater memory usage to support the additional JRubies.  Maybe rather than going into this level of detail, it would be better to just strike the last sentence here, though?

documentation/puppet_server_metrics_performance.markdown
```diff
@@ -0,0 +1,78 @@
...
+
+Each JRuby also has a certain amount of memory overhead that it needs in order to be able to load both Puppet's Ruby code and your Puppet code. In other words, your available memory limits how much Puppet code you can process. As a general rule, adding a JRuby requires a bare minimum of 40MB of memory --- the amount of memory for the scripting container, plus Puppet's Ruby code, plus an additional 5MB) --- just to compile an almost-empty catalog. For real-world catalogs, add at least an additional 15MB per additional JRuby.
```
camlow325: A couple of notes about these memory numbers:

1. The numbers refer to persistent memory usage per JRuby.  The peak memory usage during operations like catalog compilations could grow to a much higher number.  I don't think we have any good measurements around peak memory usage to relay at this point, though.

2. For reference, the 40 MB number here was measured using JRuby 1.7 with Puppet Server and with the `jruby-puppet.compile-mode` setting set to `off`.  From early testing, we have seen memory usage be significantly higher with JRuby 9k and `jruby-puppet.compile-mode` set to `off` - roughly 61 MB or more - although that may improve in future JRuby 9k versions.  More memory can be consumed as well when the `jruby-puppet.compile-mode` is set to `jit` as well, with the tradeoff being potentially better performance.

3. I suppose the "real-world" measurement is pretty subjective.  In this case, we used the application of the `role::by_size::small` class from the [puppetlabs_perf_control](https://github.com/puppetlabs/puppetlabs-puppetserver_perf_control/blob/production/site/role/manifests/by_size/small.pp) repo to compute this value.  The catalog it produces installs a Tomcat and Postgres server along with 50 or so file resources related to managing user home directories.  Do you think it would be helpful to include any extra context about the catalog compilations we were measuring?

documentation/puppet_server_metrics_performance.markdown
```diff
@@ -0,0 +1,78 @@
...
+
+-   A Puppet code change, such as a faulty or complex new function. The Grafana dashboard should show if functions start taking significantly longer, and the experimental dashboard and [status API][] endpoint also list the lengthiest function calls (showing the top 10 and top 40, respectively) based on aggregate execution times.
+-   Adding many file resources at once.
+
+In cases like these, there might be more efficient ways to author your Puppet code, you might be extending Puppet to the point where you need to add JRubies or compile masters even if you aren't adding more agents.
+
+Slowdowns in PuppetDB can also cause catalog compilations to take more time: if you use exported resources or the `puppetdb_query` function and PuppetDB has a problem, catalog compilation times will increase.
+
+Puppet Server also sends agents' facts and the compiled catalog to PuppetDB during catalog compilation. The [status API][] for the master service reports metrics for these operations under [`http-client-metrics`][HTTP client metrics], and in the Grafana dashboard in the "External HTTP Communications" graph.
```
camlow325: The backticks around "http-client-metrics" might not need to be here - not sure.

documentation/puppet_server_metrics_performance.markdown
```diff
@@ -0,0 +1,78 @@
...
+
+Puppet Server also requests facts as HTTP requests while handling a node request, and submits reports via HTTP requests while handling of a report request. If you have an HTTP report processor set up, the Grafana dashboard shows metrics for `Http report processor,` as does the [status API][] endpoint under `http-client-metrics` in the master service, for metric ID `['puppet', 'report', 'http']`. Delays in the report processor are passed on to Puppet Server.
+
+### Memory leaks and usage
+
+A memory leak or increased memory pressure can stress Puppet Server's available resources. In this case, the Java VM will spend more time doing garbage collection, causing the GC time and GC CPU % metrics to increase. These metrics are available in the [developer dashboard][] and [status API][] endpoint, as well as in the `/metrics/v1/mbeans` endpoint.
+
+If you can't identify the source of a memory leak, you might need to add more RAM to your Server or an additional compile master.
```
camlow325: I hesitate making this recommendation - although I understand some people would probably resort to this anyway.  Adding RAM just prolongs the time to failure with an OutOfMemory error and adding another compile master may not help with preventing the other compile masters from OOMing either.

If anything here, it might be worth pointing out the [max-requests-per-instance](https://github.com/puppetlabs/puppetserver/blob/56ee2195b67db4fd865985081357736d779c3183/documentation/config_file_puppetserver.markdown#settings) setting here as this is frequently used as a way to mitigate the effects of a memory leak, avoiding OutOfMemory crashes but with some performance impact.

documentation/puppet_server_metrics_performance.markdown
gguillotte: Needs to be a reference link anyway, so I'll revisit.

documentation/puppet_server_metrics_performance.markdown
gguillotte: How about "until you have more than 0 free JRubies"?

documentation/puppet_server_metrics_performance.markdown
gguillotte: I can strike that, but it'd probably be better if we could at least summarize the relationship between the setting, the number of CPUs, and Server performance. What can users measure to determine the best `max-active-instances` value?

documentation/puppet_server_metrics_performance.markdown
gguillotte: For points 1 and 2, I can fold some of that context into this section.

For point 3, that's an excellent point  Explaining our methodology would provide context and show users a way to measure usage on their own.

documentation/puppet_server_metrics_performance.markdown
gguillotte: Probably, yeah. I'll address.

documentation/puppet_server_metrics_performance.markdown
gguillotte: Thanks, that's helpful advice. I'll revisit this.

documentation/puppet_server_metrics_performance.markdown
camlow325: Maybe it would be worth suggesting that since catalog requests, the most expensive ones that agents make, are CPU-bound that it's typically best to start with a `max-active-instances` value which is no greater than the number of CPU cores on the host machine.  For cases where the catalogs that the master is compiling lead to a large number of follow-up "file" requests (file_content, file_metadata, etc.) which are more I/O than CPU bound, it may be possible to improve performance further by setting `max-active-instances` to a value greater than the number of available CPU cores.

@npwalker, since I think you have more relevant real-world experience with this than we do, any specific recommendations around the wording that you'd suggest here?

documentation/puppet_server_metrics_performance.markdown
```diff
@@ -0,0 +1,78 @@
...
+
+Puppet Server also sends agents' facts and the compiled catalog to PuppetDB during catalog compilation. The [status API][] for the master service reports metrics for these operations under [`http-client-metrics`][HTTP client metrics], and in the Grafana dashboard in the "External HTTP Communications" graph.
+
+Puppet Server also requests facts as HTTP requests while handling a node request, and submits reports via HTTP requests while handling of a report request. If you have an HTTP report processor set up, the Grafana dashboard shows metrics for `Http report processor,` as does the [status API][] endpoint under `http-client-metrics` in the master service, for metric ID `['puppet', 'report', 'http']`. Delays in the report processor are passed on to Puppet Server.
+
+### Memory leaks and usage
+
+A memory leak or increased memory pressure can stress Puppet Server's available resources. In this case, the Java VM will spend more time doing garbage collection, causing the GC time and GC CPU % metrics to increase. These metrics are available in the [developer dashboard][] and [status API][] endpoint, as well as in the `/metrics/v1/mbeans` endpoint.
```
justinstoller: I believe you can get this data from both v1 and v2 of the metrics endpoint. We might want to just link to the metrics page?

documentation/puppet_server_metrics_performance.markdown
justinstoller: There's also Chris' [blog post](https://puppet.com/blog/puppet-server-advanced-memory-debugging) about debugging memory leaks that might be helpful to link to? That might be linked elsewhere (since this is more the why rather than how-to)?

documentation/puppet_server_metrics_performance.markdown
justinstoller: Apologies if this is mentioned elsewhere, how does this relate to our base recommendations? eg the [PE](https://docs.puppet.com/pe/latest/sys_req_hw.html) & [FOSS](https://docs.puppet.com/puppet/4.9/system_requirements.html#hardware) hardware reqs.

documentation/scaling_puppet_server.md
```diff
@@ -0,0 +1,131 @@
...
+If you use the inventory service or exported resources, use [PuppetDB](https://docs.puppet.com/puppetdb/) and point all of your Puppet masters at a shared PuppetDB instance. A reasonably robust PuppetDB server can handle many Puppet masters and many thousands of agents.
+
+See the [PuppetDB documentation](https://docs.puppet.com/puppetdb/latest/) for instructions on deploying a PuppetDB server, then configure every Puppet master to use it. Note that every Puppet master will need to have its own [whitelist entry](https://docs.puppet.com/puppetdb/latest/configure.html#certificate-whitelist) if you're using HTTPS certificates for authorization.
+
+## Keeping manifests and modules synchronized across masters
+
+You must ensure that all Puppet masters have identical copies of your manifests, modules, and [external node classifier](https://docs.puppet.com/puppet/5.0/nodes_external.html) data. Examples include:
+
+-   Using a version control system such as Git, Mercurial, or Subversion to manage and sync your manifests, modules, and other data.
+-   Running an out-of-band `rsync` task via `cron`.
```
justinstoller: Do we want to mention r10k here?

justinstoller:
Looks great, Garrett!

documentation/puppet_server_metrics_performance.markdown
gguillotte: @justinstoller No relationship that I know of offhand to the PE requirements, but I can carry over the "number of agents being served, how frequently those agents check in, how many resources are being managed on each agent, and the complexity of the manifests and modules in use " caveat on the OSP reqs to here since that seems relevant.

documentation/puppet_server_metrics_performance.markdown
gguillotte: @justinstoller Chris links back to the tuning guide, which might be a useful doc to expand using content from his blog post.

documentation/scaling_puppet_server.md
gguillotte: @justinstoller We can!

documentation/puppet_server_metrics_performance.markdown
justinstoller: 👍 but understandably out of scope of this work I assume (:

documentation/puppet_server_metrics_performance.markdown
gguillotte: I see the problem now—the HTTP client metrics are in another ~castle~ PR. I've fixed the link, but it won't completely work until/unless PR #1459 is merged.

gguillotte:
Addressed feedback from @camlow325 and @justinstoller.

documentation/puppet_server_metrics_performance.markdown
```diff
@@ -38,13 +38,17 @@ If wait time increases but borrow time stays the same, your Server infrastructur
...
 
 #### Adding more JRubies
 
-If you need to add JRubies, remember that Puppet Server is tuned by default to use one fewer than your total number of CPUs, with a maximum of 4 CPUs, for the number of available JRubies. You can change this by setting `max-active-instances` in [`puppetserver.conf`][], under the `jruby-puppet` section. Don' set this value to something greater than the number of CPUs, because if there aren't enough processes available, it won't matter how many JRubies in the pool there are — the server won't be able to use them.
+If you must add JRubies, remember that Puppet Server is tuned by default to use one fewer than your total number of CPUs, with a maximum of 4 CPUs, for the number of available JRubies. You can change this by setting `max-active-instances` in [`puppetserver.conf`][puppetserver.conf], under the `jruby-puppet` section.
```
justinstoller: What does `with a maximum of 4 CPUs` mean? Does that mean the JRubies are capped at 3? This seemed to make sense to me yesterday - so I must not have had enough coffee today - but I just can't parse that clause.

documentation/puppet_server_metrics_performance.markdown
```diff
@@ -38,13 +38,17 @@ If wait time increases but borrow time stays the same, your Server infrastructur
...
+
+As a general rule, adding a JRuby requires a bare minimum of 40MB of memory under JRuby 1.7, and at least 60MB under JRuby9k if the `jruby-puppet.compile-mode` setting in [`puppetserver.conf`][puppetserver.conf] is set to `off` --- the amount of memory for the scripting container, plus Puppet's Ruby code, plus additional memory overhead --- just to compile an nearly empty catalog.
+
+For real-world catalogs, you can generally add at least an additional 15MB for each JRuby you add. We calculated this amount by comparing a minimal catalog compilation to a catalog with a [basic role](https://github.com/puppetlabs/puppetlabs-puppetserver_perf_control/blob/production/site/role/manifests/by_size/small.pp) that installs Tomcat and Postgres servers. You can calculate a similar value unique to your infrastructure by measuring `puppetserver` memory usage during your infrastructure's catalog compilations and comparing it to compiling a minimal catalog for a similar number of nodes.
```
justinstoller: these paragraphs

documentation/puppet_server_metrics_performance.markdown
gguillotte: @justinstoller That's my interpretation of the pre-docs, and I didn't modify this part much from the pre-docs. I'm not sure where or how this is enforced in the code, though.

documentation/puppet_server_metrics_performance.markdown
```diff
@@ -73,6 +77,6 @@ Puppet Server also requests facts as HTTP requests while handling a node request
 
 ### Memory leaks and usage
 
-A memory leak or increased memory pressure can stress Puppet Server's available resources. In this case, the Java VM will spend more time doing garbage collection, causing the GC time and GC CPU % metrics to increase. These metrics are available in the [developer dashboard][] and [status API][] endpoint, as well as in the `/metrics/v1/mbeans` endpoint.
+A memory leak or increased memory pressure can stress Puppet Server's available resources. In this case, the Java VM will spend more time doing garbage collection, causing the GC time and GC CPU % metrics to increase. These metrics are available in the [developer dashboard][] and [status API][] endpoint, as well as in the mbeans metrics available from both the [`/metrics/v1/mbeans`](./metrics-api/v1/metrics_api.markdown) or [`/metrics/v2/mbeans`](./metrics-api/v2/metrics_api.markdown) endpoints.
```
justinstoller: The path portion of the endpoints change from `v1` to `v2` so that its `metrics/v1/mbeans` but `metrics/v2/read` (most likely).

gguillotte:
@camlow325 

> I'm especially hazy on some of the details on the DNS SRV / load balancing configuration in the scaling_puppet_server.md doc. They would probably do a much better job vetting that than I could.

Would it help for me to split these two docs into separate PRs?

camlow325:
@gguillotte It's okay with me if you keep this all as one PR.  If we don't get any feedback from outside of the team and we're otherwise good with it, merging the whole thing together seems reasonable to me.

documentation/puppet_server_metrics_performance.markdown
camlow325: Yeah, the default number of JRubies is equal to the number of CPU cores minus 1.  If 4 or more CPU cores are available, the default number of JRubies would be 3.

documentation/puppet_server_metrics_performance.markdown
Sharpie: I think our PE docs still recommend 512 MB per JRuby --- we'll probably want to reconcile these two numbers to avoid confusion.

documentation/puppet_server_metrics_performance.markdown
justinstoller: Gotcha, for some reason that wording is more understandable to me.

documentation/scaling_puppet_server.md
```diff
@@ -0,0 +1,131 @@
...
+title: "Scaling Puppet Server with compile masters"
+canonical: "/puppetserver/latest/scaling_puppet_server.html"
+---
+
+To scale Puppet Server for many thousands of nodes, you'll need to add Puppet masters dedicated to catalog compilation. These Servers are known as **compile masters**, and are simply additional load-balanced Puppet Servers that receive catalog requests from agents and synchronize the results with each other.
+
+## Planning your load-balancing strategy
+
+The rest of your configuration depends on how you plan on distributing the agent load. Determine what your deployment will look like before you add any compile masters, but **implement load balancing as the last step** only after you have the infrastructure in place to support it.
+
```
Sharpie: Might be worth calling out the PE docs on load balancing:

https://docs.puppet.com/pe/latest/install_lei_load.html

The stuff on DNS and SRV records looks great, but PE users will need to take the above documentation into consideration as well.

documentation/scaling_puppet_server.md
```diff
@@ -0,0 +1,131 @@
...
+### Directing individual agents to a central CA
+
+On every agent, set the [`ca_server`](https://docs.puppet.com/puppet/latest/reference/configuration.html#caserver) setting in [`puppet.conf`](https://docs.puppet.com/puppet/latest/reference/config_file_main.html) (in the `[main]` configuration block) to the hostname of the server acting as the certificate authority. If you have a large number of existing nodes, it is easiest to do this by managing `puppet.conf` with a Puppet module and a template.
+
+> **Note:** Set this setting *before* provisioning new nodes, or they won't be able to complete their initial agent run.
+
+### Pointing DNS `SRV` records at a central CA
+
+If you [use `SRV` records for agents](#using-dns-srv-records), you can use the `_x-puppet-ca._tcp.$srv_domain` DNS name to point clients to one specific CA server, while the `_x-puppet._tcp.$srv_domain` DNS name handles most of their requests to masters and can point to a set of compile masters.
+
```
Sharpie: Similarly, PE users will want to avoid tinkering with the CA settings as the `puppet_enterprise` module handles this.

documentation/scaling_puppet_server.md
gguillotte: This is aimed at open-source Server users. If we're worried about PE users landing here, we should point back to relevant PE docs, but if there's relevant content in the PE docs we should just fold the relevant parts into here. (And vice-versa.)

Sharpie: Yeah, PE users will definitely come across these so we should probably include at least a notice box that points over to the PE LEI and HA docs.

documentation/puppet_server_metrics_performance.markdown
camlow325: I believe the "512 MB" number was derived more from empirical experience of the memory requirements of a system under load whereas the numbers here pertain to minimum persistent memory usage (and tend to discount active / peak request load).  I don't think any of this new information would displace the historical "512 MB" recommendation.

@Sharpie - do you think mentioning this level of "minimum persistent usage" info is useful to external users or would you suggest that we jettison it completely?  I definitely agree with you that it would be good to avoid giving users information that they may see as conflicting.

gguillotte:
Rebased to resolve conflicts with #1459.

documentation/puppet_server_metrics_performance.markdown
Sharpie: I think it's fine to leave this in as it is useful, empirically determined info. However, maybe we should add something to the second paragraph indicating that memory usage could go much beyond 15 MB depending on population size, amount of puppet code and complexity or volume of data that goes into compilation.

gguillotte: It's omitted from that excerpt, but the paragraph before it notes that:

> Catalog compilation can consume more memory, and Puppet Server's total memory usage depends on the number of agents being served, how frequently those agents check in, how many resources are being managed on each agent, and the complexity of the manifests and modules in use.

Sharpie: Ah yes. That breaks it down pretty well. Might be worth re-enforcing it just a touch in the second paragraph so that folks read "15 MB based on <tomcat example>. Actual usage could be much more based on the above factors." vs. "15 MB based on <tomcat example>" then get surprised when it turns out to be 100s of MB for their usage.

gguillotte: @Sharpie I've re-emphasized this, and also cited the PE guide to note that 512MB recommendation.

PR #1461, opened 2017-07-13
(SERVER-1866) Copy the cacrl to the hostcrl file on change at runtime

camlow325:
This commit uses the tk-filesystem-watcher service to listen for changes
to the cacrl file.  When a change is detected, the cacrl file is copied
over to the hostcrl file.  This allows a watcher that the newer
tk-jetty9 service has established on changes to the hostcrl file to
trigger a reload of the CRL being used by the webserver, with no restart
to the webserver service being required.

The tk-jetty9 bump in this PR for the crl watcher changes also includes
a fix for previously failing gzip encoding for server responses.
To account for this fix, a couple of updates to the
environment-classes-int-test tests were added to account for the --gzip
text appended to some response etags.

### DIFF
--- a/acceptance/suites/tests/service_framework/reload_updated_crl.rb
+++ b/acceptance/suites/tests/service_framework/reload_updated_crl.rb
@@ -1,4 +1,4 @@
-test_name '(SERVER-1380) Server loads updated CRL after being reloaded'
+test_name 'Server reloads updated CRL without a restart'
 
 puppetservice=options['puppetservice']
 cert_to_revoke="reload_updated_crl"
@@ -78,24 +78,9 @@ step 'Revoke cert' do
   on(master, puppet('cert', 'revoke', cert_to_revoke))
 end
 
-if options[:type] == 'pe'
-  step 'Copy cacrl to hostcrl in order to work around SERVER-911' do
-    on(master, 'cp "$(puppet config print cacrl)" "$(puppet config print hostcrl)"')
-  end
-end
-
-step 'Validate that noop agent run successful after cert revoked but before reload' do
-  on(master, puppet('agent', '--test',
-                    '--server', server,
-                    '--certname', cert_to_revoke,
-                    '--noop'))
-end
-
-step 'Reload the server' do
-  reload_server
-end
-
-step 'Validate that noop run for revoked agent fails with SSL error after server reload' do
+step 'Validate that noop run for revoked agent fails with SSL error after CRL reload' do
+  # Sleep for a few seconds to give the server a chance to reload the updated CRL
+  sleep 5
   on(master, puppet('agent', '--test',
                     '--server', server,
                     '--certname', cert_to_revoke,
--- a/project.clj
+++ b/project.clj
@@ -1,5 +1,5 @@
 (def ps-version "2.7.3-SNAPSHOT")
-(def tk-jetty9-version "1.7.1-SNAPSHOT")
+(def tk-jetty9-version "1.8.0")
 
 (defn deploy-info
   [url]
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,5 @@
 (def ps-version "2.7.3-SNAPSHOT")
+(def tk-jetty9-version "1.7.1-SNAPSHOT")
 
 (defn deploy-info
   [url]
@@ -63,6 +64,7 @@
                  [puppetlabs/trapperkeeper-authorization]
                  [puppetlabs/trapperkeeper-scheduler]
                  [puppetlabs/trapperkeeper-status]
+                 [puppetlabs/trapperkeeper-filesystem-watcher "1.0.1"]
                  [puppetlabs/kitchensink]
                  [puppetlabs/ssl-utils]
                  [puppetlabs/ring-middleware]
@@ -109,8 +111,8 @@
 
   :profiles {:dev {:source-paths  ["dev"]
                    :dependencies  [[org.clojure/tools.namespace]
-                                   [puppetlabs/trapperkeeper-webserver-jetty9]
-                                   [puppetlabs/trapperkeeper-webserver-jetty9 nil :classifier "test"]
+                                   [puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty9-version]
+                                   [puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty9-version :classifier "test"]
                                    [puppetlabs/trapperkeeper nil :classifier "test" :scope "test"]
                                    [puppetlabs/kitchensink nil :classifier "test" :scope "test"]
                                    [ring-basic-authentication]
@@ -141,12 +143,12 @@
                                                ;; lein depend on clojure 1.6.
                                                [org.clojure/clojure nil]
                                                [puppetlabs/puppetserver ~ps-version]
-                                               [puppetlabs/trapperkeeper-webserver-jetty9 nil]
+                                               [puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty9-version]
                                                [org.clojure/tools.nrepl nil]]
                       :plugins [[puppetlabs/lein-ezbake "1.1.7"]]
                       :name "puppetserver"}
              :uberjar {:aot [puppetlabs.trapperkeeper.main]
-                       :dependencies [[puppetlabs/trapperkeeper-webserver-jetty9]]}
+                       :dependencies [[puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty9-version]]}
              :ci {:plugins [[lein-pprint "1.1.1"]]}
              :voom {:plugins [[lein-voom "0.1.0-20150115_230705-gd96d771" :exclusions [org.clojure/clojure]]]}}
 
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -783,8 +783,14 @@
                                          localcacrl
                                          cacrl-as-string)]
                (if (zero? attempts-left)
-                 (log/error (i18n/trs "Unable to synchronize crl file {0} to {1}: {2}"
-                                      cacrl localcacrl (.getMessage write-exception)))
+                 (log/error (format "%s\n%s\n%s"
+                                    (i18n/trs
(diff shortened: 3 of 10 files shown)


camlow325:
This will fail Travis because it is dependent upon changes in a new tk-webserver-jetty9 release, which would be done as part of https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/pull/177.  I'm just putting this up for review for now.

acceptance/suites/tests/service_framework/reload_updated_crl.rb
```diff
@@ -78,24 +78,9 @@
...
-end
-
-step 'Reload the server' do
-  reload_server
-end
-
-step 'Validate that noop run for revoked agent fails with SSL error after server reload' do
+step 'Validate that noop run for revoked agent fails with SSL error after CRL reload' do
+  # Sleep for a few seconds to give the server a chance to reload the updated CRL
+  sleep 5
```
camlow325: This may not strictly be necessary.  The test passed fine for me without it.  Just seemed like this would make the test a bit less likely to fail.

acceptance/suites/tests/service_framework/reload_updated_crl.rb
```diff
@@ -1,4 +1,4 @@
-test_name '(SERVER-1380) Server loads updated CRL after being reloaded'
+test_name 'Server reloads updated CRL without a restart'
```
camlow325: The test was originally exercising a service reload to see that the new CRL was being used.  With the addition of the file system watcher and associated Jetty reload, a service HUP should no longer be necessary.  I thought it might be best to just change the test to validate that the runtime reload occurs rather than doing the HUP, although maybe an argument could be made for testing both the HUP and filesystem watcher.

project.clj
```diff
@@ -1,4 +1,5 @@
 (def ps-version "2.7.3-SNAPSHOT")
+(def tk-jetty9-version "1.7.1-SNAPSHOT")
```
camlow325: This will need to be replaced with a new tk-jetty9 release when one is available.

project.clj
```diff
@@ -63,6 +64,7 @@
                  [puppetlabs/trapperkeeper-authorization]
                  [puppetlabs/trapperkeeper-scheduler]
                  [puppetlabs/trapperkeeper-status]
+                 [puppetlabs/trapperkeeper-filesystem-watcher "1.0.1"]
```
camlow325: tk-filesystem-watcher is now being managed with clj-parent, as of version 0.1.9, but it was not as of the version that puppetserver 2.x is currently using, 0.1.7.  It may be best to do another pre-1.0.0 release of clj-parent which bumps to the new tk-jetty9 version (when available) and uses the tk-filesystem-watcher version as well.

camlow325:
I ran the full Puppet Server acceptance smoke test suite with an EL-7 build of these changes and the tests all passed.

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
```diff
@@ -35,7 +42,19 @@
          auth-handler
          puppet-version)
        {:normalize-request-uri true})
-     (assoc context :auth-handler auth-handler)))
+     (watch-protocol/add-watch-dir! watcher
+                                    (fs/parent ca-crl-file)
+                                    {:recursive true})
```
camlow325: I left a similar comment in the related tk-jetty9 PR, https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/pull/177#discussion_r127269511, about how it would probably be better here to look into enhancing the tk-filesystem-watcher service to be able to watch just the immediate parent directory for the CRL file rather than recursively watching the entire directory tree.

camlow325: For reference, I submitted a ticket for the work to add support for non-recursive directory watching to the tk-filesystem-watcher service, [TK-448](https://tickets.puppetlabs.com/browse/TK-448).  It might be reasonable to consider merging this ahead of that work getting done.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -747,9 +766,33 @@
   should be copied to."
   ([cacrl :- schema/Str
     localcacrl :- schema/Str]
-    (when (fs/exists? cacrl)
-      (ks/mkdirs! (fs/parent localcacrl))
-      (fs/copy cacrl localcacrl))))
+   (when (not= cacrl localcacrl)
+     (let [max-attempts 5]
+       (when (and (fs/exists? cacrl))
```
haus: I think this `and` isn't needed anymore as there is only one condition.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -747,9 +766,33 @@
...
+             (when-let [write-exception (write-local-cacrl!
+                                         localcacrl
+                                         cacrl-as-string)]
+               (if (zero? attempts-left)
+                 (log/error (i18n/trs "Unable to synchronize crl file {0} to {1}: {2}"
+                                      cacrl localcacrl (.getMessage write-exception)))
+                 (do
+                   (Thread/sleep 1000)
+                   (let [new-ca-crl-last-modified (fs/mod-time cacrl)]
+                     (recur (if (= ca-crl-last-modified new-ca-crl-last-modified)
```
haus: So this could loop indefinitely if there were a competing process also modifying the crl and causing a failure to write it out.

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
```diff
@@ -265,3 +266,73 @@
...
+         (is (= 200 (:status node-response-before-revoke)))))
+     (testing "node certificate can be revoked successfully"
+       (let [revoke-response (cert-status-request "revoked")]
+         (is (= 204 (:status revoke-response)))))
+     (testing "node request after revocation fails"
+       (let [ssl-exception-for-request? #(try
+                                           (client-request)
+                                           false
+                                           (catch ConnectionClosedException e
+                                             (is true)))]
```
haus: It feels a little weird to have an assertion in the catch block?

haus:
Looks good overall to me. I think some of the cacrl copying code could maybe use a few comments and I'm a little worried about that loop that isn't guaranteed to terminate.

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
justinstoller: Yeah, is this something that could be done with `thrown?`?

haus: Or just `true`, as the `cond` below would get it still.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
camlow325: Yeah, that's probably true.  What would you suggest we do instead here?

haus: Arbitrary number of loop max? Not sure what should happen at the max count though

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -747,9 +766,33 @@
...
+                ca-crl-last-modified (fs/mod-time cacrl)]
+           (let [cacrl-as-string (slurp cacrl)]
+             (when-let [write-exception (write-local-cacrl!
+                                         localcacrl
+                                         cacrl-as-string)]
+               (if (zero? attempts-left)
+                 (log/error (i18n/trs "Unable to synchronize crl file {0} to {1}: {2}"
+                                      cacrl localcacrl (.getMessage write-exception)))
+                 (do
+                   (Thread/sleep 1000)
```
MosesMendoza: This timeout also seems like it could be reduced - maybe halved?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
haus: @camlow325 was there a specific concern you were guarding against here with the last modified check? It seems like even if we log an error here and move along, that when the modification finishes it will just trigger a new fs-watcher event and try again.
I'm inclined to remove the modification time check and simplify the logic here.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -747,9 +766,33 @@
...
+         ;; cacrl file being copied.
+         (loop [attempts-left max-attempts
+                ca-crl-last-modified (fs/mod-time cacrl)]
+           (let [cacrl-as-string (slurp cacrl)]
+             (when-let [write-exception (write-local-cacrl!
+                                         localcacrl
+                                         cacrl-as-string)]
+               (if (zero? attempts-left)
+                 (log/error (i18n/trs "Unable to synchronize crl file {0} to {1}: {2}"
+                                      cacrl localcacrl (.getMessage write-exception)))
```
MosesMendoza: Can we add some more context to this error message so folks know what the implications are and what to try next? I'm thinking something like:
...
```
Unable to synchronize crl file {0} to {1}: {2}. Recent changes to the 
CRL may not have taken effect. To load the updated CRL file reload 
or restart the Puppet Server service."
```
 Or something like that.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
camlow325: It's probably fine to just assume that a later modification event would be triggered in the event that the first set of retries all fail.  I don't know enough about when an OS may decide to trigger the notification so I can't say that would be a bad idea - and it definitely seems simpler.  Seems fine to just go with that for now and make it more sophisticated later only if needed.  Assuming the last modified logic were removed here, seems like it would probably be good to remove it from the [associated tk-webserver-jetty9 changes](https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/c409d7916df418f546e5b352c278236733073b40/java/com/puppetlabs/trapperkeeper/services/webserver/jetty9/utils/InternalSslContextFactory.java#L126-L145) as well, since it basically does the same sort of thing.

haus:
@MosesMendoza updated

MosesMendoza:
lincoln-boombox

PR #1462, opened 2017-07-13
Merge 5.0.x at ed415838 into master

adrienthebo:


adrienthebo:
I'm going to run some acceptance smoke tests on this for sanity.

adrienthebo:
This change might trigger failures in CI because the puppet-agent builds that we've been pinned to look to have been cleaned up. However this could be a really good chance to verify https://github.com/puppetlabs/puppetserver/pull/1458.

justinstoller:
Since this is a merge up does that mean that 5.0.x will also start failing?

camlow325:
> Since this is a merge up does that mean that 5.0.x will also start failing?

I don't think so since this PR preserves the pins for puppet-agent and the Ruby Puppet submodule which are already on the master branch - not bringing the newer ones in from the 5.0.x branch.

justinstoller:
> I don't think so since this PR preserves the pins

I guess I should ask, if these pins are broken, and doing this merge up will cause a CI run that changes no code but we know is going to fail... Why don't we change those pins in 5.0.x and then do the merge up?

justinstoller:
Sounds like I read this wrong

PR #1465, opened 2017-07-17
(server-1870) add ci job for automatically bumping puppet submodule commit

er0ck:
* merge-up from 5.0.x
* update pin from `5.0.x` to `master`

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,13 +1,20 @@
 require 'open3'
+require 'open-uri'
+require 'json'
+require 'pp'
 
 PROJECT_ROOT = File.dirname(__FILE__)
 ACCEPTANCE_ROOT = ENV['ACCEPTANCE_ROOT'] ||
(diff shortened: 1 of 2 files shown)


MosesMendoza:
I'm not familiar yet with the workflow on puppet-server so forgive me for the noob question - but why cherry-pick from 5.0.x to master, instead of merging 5.0.x to master? Is master not ahead of 5.0.x?

er0ck:
yeah, i probably should have.  but i had to update the constant to master after that commit.  i didn't think it mattered really.  if people prefer, i can certainly merge, instead

MosesMendoza:
@er0ck would you mind? I admit i'm a bit pedantic - it would be great to avoid the duplicate commit that we'll have when we do the next merge-up after this cherry-pick, just to keep the history clean. I personally prefer `--no-ff --log` flags to the merge so its clear what the merge-up contained too.

er0ck:
no problemo.

PR #1466, opened 2017-07-19
(SERVER-1894) install `apt-transport-https` on Debian for PDB

justinstoller:
We install PDB and smoke test it in
`acceptance/suites/tests/00_smoke/puppetdb_integration.rb`. As of
2017-07-17 the puppetlabs-puppetdb module will attempt to install
postgresql using a repo with in the https url. apt-get cannot use
https by default, the `apt-transport-https` package installs the addon
needed to use apt-get with https.

This patch installs `apt-transport-https` on debian variants.

### DIFF
--- a/acceptance/suites/tests/00_smoke/puppetdb_integration.rb
+++ b/acceptance/suites/tests/00_smoke/puppetdb_integration.rb
@@ -25,6 +25,10 @@ if !master.is_pe?
     on(master, puppet('module install puppetlabs-puppetdb'))
   end
 
+  if master.platform.variant == 'debian'
+    master.install_package('apt-transport-https')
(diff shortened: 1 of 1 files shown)


MosesMendoza:
@justinstoller just confirming target branch - no need for this in LTS ?

justinstoller:
Good question.

We don't test 2.x against Debian 8 and it doesn't sound like we should be testing it against PDB 5 (the corresponding version of the module brings in the version of the postgresql module that causes us issues on Debian 8).

I think we will need to separately pin 2.x to use ~4.x version of the PDB module & PDB, but I don't think we would use this in the 2.x branch.

PR #1467, opened 2017-07-19
(SERVER-1893) Cherry-pick JRE switching in upgrade tests from master

justinstoller:
This cherry-pick's b4b9938 from master into 5.0.x.


justinstoller:
I've confirmed locally that this one cherry-pick resolves the primary issue of puppetserver not starting in the upgrade test.

In a run of the full suite my hosts were lost around an hour and half into it, there were no failures in the test suite at that point...

PR #1470, opened 2017-07-20
(SERVER-1315) allow autosign w/bundled ca cert

MosesMendoza:
* Prior to this commit, if the CA certificate was a bundled certificate chain,
  autosigning would fail with an error "The PEM stream must contain exactly 1
  certificate". This commit updates the auto-signing to use `pem->certs`
  instead of `pem->cert`, and operate on the first certificate in the chain.
  The first certificate is the leaf certificate in a given chain bundle.

* This commit adds a test that asserts we can autosign given a CA certificate
  file containing more than one certificate (a bundle). The CA fixtures will be
  useful in validating other behavior on a certificate bundle if needed - they
  are a CA cert and CRL for an intermediate and root CA, and the private key
  for the intermediate.


adrienthebo:
This change makes me a bit anxious because it's exposing the coupling of two different concepts - the CA certificate that we're using to sign certificates, and the CA certificate bundle that we trust when validating SSL connections. We need the CA certificate for information like the authority key identifier when signing certificates and revoking CRLs, as @justinstoller noted. We can require that if the CA certificate file is a bundle then Puppet's CA certificate must be first, and if the wrong certificate is first then I think that operations will just fail so this isn't super dangerous. However we've had long standing issues with coupling unrelated concepts in SSL code and it tends to bite us later, so if we can detangle these two things I'd be happier.

That all said I'm not coming up with a lot of alternatives that make me particularly excited. Perhaps we should have yet another setting for the trusted CA certificate bundle? That was the route that I was pushing about a year ago; decoupling trusted CA bundles and the signing certificate.

justinstoller:
> Perhaps we should have yet another setting for the trusted CA certificate bundle?

What would that look like, @adrienthebo ? Like do you have any ideas how would we decouple those concepts in the code between server and agent?

adrienthebo:
One option is to keep the semantics of `cacert` the same, since it's literally the CA certificate. AFAIK we only use that as a single PEM encoded entity for signing certs so we don't have to tweak that. For the agent and compile master side, we can adapt the `localcacert` setting to use for bundles, and either keep the name or rename it to `cabundle` or somesuch.

One of the implications of all of this though is how we copy the `cacert` to the `localcacert` location; we'll have to be a bit smarter than relying on semi-blind copies here.

Alternately we can stick with this implementation as-is; I'm not married to this approach and doing this will mean more work. $0.50.

justinstoller:
So it kinda sounds like disambiguating the CA cert and CA cert bundle is out of scope for this PR.

If so, I think there's [still more work to do](https://github.com/puppetlabs/puppetserver/pull/1470#discussion_r128871582) to handle the move from cert to cert bundle and update the equivalent tests before merging this.

Does that seem right to you, @adrienthebo ?

adrienthebo:
@justinstoller sounds correct to me

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -938,7 +938,8 @@
    csr :- CertificateRequest
    {:keys [cacert cakey signeddir ca-ttl serial cert-inventory]} :- CaSettings]
   (let [validity    (cert-validity-dates ca-ttl)
-        cacert      (utils/pem->cert cacert)
+        ;; if part of a CA bundle, the intermediate CA will be first in the chain
+        cacert      (first (utils/pem->certs cacert))
         signed-cert (utils/sign-certificate (utils/get-subject-from-x509-certificate
                                              cacert)
                                             (utils/pem->private-key cakey)
```
justinstoller: We may want to update our terminology to ensure we're not talking about root "upstream" cert/key.

We may also want to check the leaf cert that we pull from the cert chain matches the private key that we read from disk and error if they do not match.

justinstoller:
Superseded by https://github.com/puppetlabs/puppetserver/pull/1480

PR #1471, opened 2017-07-24
(PUP-7794) Update gettext-setup gem dependency

Magisus:
Code to load translations in modules requires gettext-setup > 0.17. This
commit updates the gettext-setup dependency to the most recent version,
0.26.


Magisus:
We're going to want to merge this up to avoid warnings in the other branches as well.

PR #1473, opened 2017-07-28
(SERVER-1755) Do not deserialize msgpack payloads to UTF-8 string

Magisus:
Msgpack data, like other binary data, cannot be deserialized to a UTF-8
string. In general, we try to decode all payloads this way when we
receive requests. This commit adds an exception for payloads from
msgpack, sending them as is to Ruby Puppet.

This also updates the unit test concerning binary content types.

### DIFF
--- a/test/unit/puppetlabs/services/request_handler/request_handler_core_test.clj
+++ b/test/unit/puppetlabs/services/request_handler/request_handler_core_test.clj
@@ -121,7 +121,8 @@
   (testing "request with binary content type does not consume body"
     (let [body-string "some random text"]
       (doseq [content-type [nil "" "application/octet-stream"
-                            "APPLICATION/octet-Stream"]]
+                            "APPLICATION/octet-Stream"
+                            "application/x-msgpack"]]
         (let [body-reader (StringReader. body-string)
               wrapped-request (core/wrap-params-for-jruby
                                 {:body         body-reader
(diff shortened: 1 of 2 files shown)


test/unit/puppetlabs/services/request_handler/request_handler_core_test.clj
```diff
@@ -121,7 +121,8 @@
   (testing "request with binary content type does not consume body"
     (let [body-string "some random text"]
       (doseq [content-type [nil "" "application/octet-stream"
-                            "APPLICATION/octet-Stream"]]
+                            "APPLICATION/octet-Stream"
```
adrienthebo: If this has a more concrete integration test it might be worth extending test coverage to actually making a HTTP request, but that might be more work than it's worth.

jpinsonault: I couldn't find a higher level test after a quick search. I think I'm personally ok if this just stays a unit test. It runs it through the `core/wrap-params-for-jruby` fn, and all requests that use a jruby instance are wrapped by that fn  [here](https://github.com/puppetlabs/puppetserver/blob/master/src/clj/puppetlabs/services/request_handler/request_handler_core.clj#L268)

jpinsonault:
I'm 👍

PR #1474, opened 2017-08-02
(SERVER-1866) Add filesystem-watch-service to user_repl

haus:
Without this change running puppetserver from the repl fails trying to
find the filesystem-watch-service. This commit adds it to allow folks to
run from the repl.


justinstoller:
Tested this locally. Seems like a trivial oversight from the first PR, so I'm going to merge without two reviews.

PR #1475, opened 2017-08-07
(SERVER-1866)(TK-149) Make crl reloading optional in puppetserver 2.x

MosesMendoza:
* We encountered issues with the file system watcher service running on Java 7.
  The puppet server 2.x line was previousy updated to use the file system watch
  service for automatic crl reload functionality. However, the 2.x supports
  Java 7. To avoid potentially destabilizing installs, especially so late in
  the 2 series, make the file system watcher optional, defaulting to off. Users
  can enable for Java 8 installs if desired by adding the file system watch
  service to the bootstrap.cfg.
* Add a test that validates that the puppet server will not automatically
  reload the crl if the file system watcher service has not been included in
  the bootstrap.cfg.
* In order to facilitate this, we add a new test helper macro that allows you
  to specify an additional set of services to be included along the default set
  loaded from dev/bootstrap.cfg. We update the previous test to use this new
  helper to explicitly include the file system watcher service in the bootstrap
  and validate that crl reload behavior, and then we do *not* include it in a
  separate test to validate that the previously expected behavior remains
  intact.
* We comment out the service in dev/bootstrap.cfg to ensure it is not loaded by
  other tests. This will only be for the 2.x series - in later series we will
  load the service by default.

### DIFF
--- a/src/clj/puppetlabs/services/ca/certificate_authority_service.clj
+++ b/src/clj/puppetlabs/services/ca/certificate_authority_service.clj
@@ -12,10 +12,11 @@
 
 (tk/defservice certificate-authority-service
   CaService
-  [[:PuppetServerConfigService get-config get-in-config]
-   [:WebroutingService add-ring-handler get-route]
-   [:AuthorizationService wrap-with-authorization-check]
-   [:FilesystemWatchService create-watcher]]
+  {:required [[:PuppetServerConfigService get-config get-in-config]
+              [:WebroutingService add-ring-handler get-route]
+              [:AuthorizationService wrap-with-authorization-check]]
+   :optional [FilesystemWatchService]}
+
   (init
    [this context]
    (let [path (get-route this)
@@ -23,14 +24,7 @@
          puppet-version (get-in-config [:puppetserver :puppet-version])
          custom-oid-file (get-in-config [:puppetserver :trusted-oid-mapping-file])
          oid-mappings (ca/get-oid-mappings custom-oid-file)
-         auth-handler (fn [request] (wrap-with-authorization-check request {:oid-map oid-mappings}))
-         ca-crl-file (.getCanonicalPath (fs/file
-                                         (get-in-config
-                                          [:puppetserver :cacrl])))
-         host-crl-file (.getCanonicalPath (fs/file
-                                           (get-in-config
-                                            [:puppetserver :hostcrl])))
-         watcher (create-watcher)]
+         auth-handler (fn [request] (wrap-with-authorization-check request {:oid-map oid-mappings}))]
      (ca/validate-settings! settings)
      (ca/initialize! settings)
      (log/info (i18n/trs "CA Service adding a ring handler"))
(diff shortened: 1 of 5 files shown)


MosesMendoza:
ping @haus @justinstoller for review

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
```diff
@@ -45,20 +39,32 @@
...
-          (when (some #(and (:changed-path %)
-                            (= (.getCanonicalPath (:changed-path %))
-                               ca-crl-file))
-                      events)
-            (ca/retrieve-ca-crl! ca-crl-file host-crl-file)))))
-     (assoc context :auth-handler auth-handler
-                    :watcher watcher)))
+     (-> context
+         (assoc :auth-handler auth-handler)
+         ((fn [context]
```
justinstoller: I don't think this necessarily adds to the readability of what's happening, and while using fp, I don't think it's more idiomatic than say adding
```clj
context' (assoc :auth-handler auth-handler)
```
in the [above `let`](https://github.com/puppetlabs/puppetserver/pull/1475/files#diff-3e475421e7cb19ba1f757a133f75e650R27), and then having just the `if-let` (without the anonymous fn) operate on `context'` rather than `context`.

haus: agree

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
```diff
@@ -45,20 +40,29 @@
...
-        watcher
-        (fn [events]
-          (when (some #(and (:changed-path %)
-                            (= (.getCanonicalPath (:changed-path %))
-                               ca-crl-file))
-                      events)
-            (ca/retrieve-ca-crl! ca-crl-file host-crl-file)))))
-     (assoc context :auth-handler auth-handler
-                    :watcher watcher)))
+       (if-let [filesystem-watch-service
```
justinstoller: whitespace is off on this

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
```diff
@@ -45,20 +40,29 @@
...
+                                             {:recursive true})
+             (watch-protocol/add-callback!
+               watcher
+               (fn [events]
+                 (when (some #(and (:changed-path %)
+                                   (= (.getCanonicalPath (:changed-path %))
+                                     ca-crl-file))
+                             events)
+                   (ca/retrieve-ca-crl! ca-crl-file host-crl-file)))))
+           (assoc context' :watcher watcher))
```
justinstoller: This watcher is unused if `ca-crl-file` and `host-crl-file` are are the same, but still associated into the context, correct?

That seems unnecessary but I'm maybe missing something. And probably out of scope of what's happening in this PR.

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
MosesMendoza: ooh yes

MosesMendoza: actually the whitespace is off on this whole function because of the single space indent after the args

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
MosesMendoza: yeah I wonder why that was done that way. I guess it seems like they're always going to be different files (different paths) in a standard puppetserver setup so maybe it wasn't considered?

MosesMendoza: maybe because we have to return the context no matter what and since we end up with 3 different possible conditional forks otherwise we end up repeating ourselves a lot if we don't

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
MosesMendoza: fixed whitespace

justinstoller:
I'm good on this if someone else wants to merge!

PR #1476, opened 2017-08-07
(SERVER-1886)(TK-448) Use non-recursive form of watcher

MosesMendoza:
* This includes commits from https://github.com/puppetlabs/puppetserver/pull/1475, which should be merged first. To review the only intended change for this PR, see the commit `(SERVER-1886)(TK-448) Use non-recursive form of watcher`

* This PR switches to using the non-recursive from of the file system
  watcher for crl reload functionality. Without this, a directory created in
  the SSL dir owned by root would cause the service to fail.

* This is dependent on https://github.com/puppetlabs/trapperkeeper-filesystem-watcher/pull/17 being merged first

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,5 +1,5 @@
 (def ps-version "2.7.3-SNAPSHOT")
-(def tk-jetty9-version "1.8.0")
+(def tk-jetty9-version "1.8.1")
 
 (defn deploy-info
   [url]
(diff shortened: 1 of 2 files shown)


justinstoller:
Just making flagging the already mentioned fact that this is waiting on  #1475

MosesMendoza:
this has been rebased now that #1475 is merged and only includes the commits for review

PR #1478, opened 2017-08-09
(maint) Improve submodule bump

justinstoller:
In the master branch we are failing because the assumption in the rake task is that from the parsed JSON returned by jenkins we can find the parameters we need at
```rb
jenkins_result_parsed['actions'][0]['parameters']
```

However our master branch is currently failing because the parameter action is not the first element in the actions. See screen shot below or the output at [the latest failure](https://jenkins-master-prod-1.delivery.puppetlabs.net/job/platform_puppetserver_periodic-singleton_master/7/)

This searches for the correct element in the array instead of assuming it will be in the first position. It also improves the logging of raised exceptions for better debugging in the future. It will need to be merged up to master to fix CI.

screen shot 2017-08-08 at 11 05 20 pm

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -100,12 +100,13 @@ def jenkins_passing_json_parsed
   end
 
   begin
-    jenkins_result_parameters = jenkins_result_parsed['actions'][0]['parameters']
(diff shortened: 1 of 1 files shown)


er0ck:
looks good at first glance.  testing...

PR #1480, opened 2017-08-09
(SERVER-1315) Extract and verify ca cert from bundle

adrienthebo:
In order to support certificate bundles the `cacert` file has changed from
just holding a single certificate to potentially holding a certificate
chain. When we need a copy of the CA certificate we need to extract the
first cert from the bundle and verify that it matches the signing key. This
commit relies on the `pem->ca-cert` function to extract and verify the cert
whenever we need to touch the CA certificate.

We need to release a new version of jvm-ssl-utils before we can merge this change.


justinstoller:
Is it possible that we also need to watch out for this when listing cert status(es)? [link](https://github.com/puppetlabs/puppetserver/blob/master/src/clj/puppetlabs/puppetserver/certificate_authority.clj#L1155)

justinstoller:
Should we extend the testing for the other functions (revoke, setup) as well?

MosesMendoza:
blocked on https://github.com/puppetlabs/jvm-ssl-utils/pull/79

adrienthebo:
This should be ready for merge and travis is passing.

adrienthebo:
@MosesMendoza @justinstoller ping?

justinstoller:
I'm still unsure if this should be extended to cert status and I think we should add some tests for the other methods/endpoints.

adrienthebo:
With respect to certificate status, that code operates on a signed certificate in the `$cadir/signed` directory. Files in that directory should only have a single certificate in them; if there's more than one certificate in a given file and they have different statuses, how do we report that? Any interaction with the `cacert` or `localcacert` settings will need to handle multiple certificates but everything else should generally expect a single certificate.

I'll look at increasing test coverage for all points that are using the `pem->ca-cert` function.

adrienthebo:
I've added test coverage for revoking a certificate while `cacert` holds a certificate bundle. Instead of using `pem->ca-cert` when revoking, we could just extract the public key from `cakey` and use that, and ignore `cacert` entirely - but that'd necessitate another release of jvm-ssl-utils.

@justinstoller am I missing any other test coverage?

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -448,6 +466,20 @@
       (revoke-existing-cert! settings "localhost")
       (is (true? (revoked? cert))))))
 
+(deftest revoke-with-bundled-ca-certs
+  (testing "The CA certificate file can be a bundle when revoking a certificate"
+    (let [settings (testutils/ca-sandbox! cadir)
```
MosesMendoza: did you mean to use the `bundle-cadir` here?

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -448,6 +466,20 @@
...
+          cert     (-> (:signeddir settings)
```
adrienthebo: The whitespace here is the same as existing tests; would you rather have one space between `cert` and the opening paren?

justinstoller: and by "here" I mean this PR, I think we want to not align values in let blocks in general

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
adrienthebo: Good catch, I copied too much from an existing test. Switching to use the bundle-cadir currently fails because we don't actually support CRL bundles; I'll fix that too.

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
justinstoller: I honestly don't care much about formatting but it was my understanding that we only put one space between keys and values in maps and names and values in assignment (and that the formatting in the CA was especially old).

PR #1481, opened 2017-08-10
(SERVER-249) Enable loading user-supplied JARs when puppetserver starts

Magisus:
This configures EZBake to create a directory for users to place JARs in, and adds that directory to the classpath setup that runs when the user starts puppetserver.

### DIFF
--- a/resources/ext/cli_defaults/cli-defaults.sh.erb
+++ b/resources/ext/cli_defaults/cli-defaults.sh.erb
@@ -7,4 +7,4 @@ if [ ! -e "$JRUBY_JAR" ]; then
   return 1
 fi
 
-CLASSPATH="${CLASSPATH}:${JRUBY_JAR}"
+CLASSPATH="${CLASSPATH}:${JRUBY_JAR}:/opt/puppetlabs/server/data/<%= EZBake::Config[:real_name] %>/jars/*"
--- a/project.clj
+++ b/project.clj
@@ -112,6 +112,7 @@
                        :build-type "foss"
                        :java-args ~(str "-Xms2g -Xmx2g "
                                      "-Djruby.logger.class=com.puppetlabs.jruby_utils.jruby.Slf4jLogger")
+                       :create-dirs ["/opt/puppetlabs/server/data/puppetserver/jars"]
                        :repo-target "PC1"
                        :bootstrap-source :services-d
                        :logrotate-enabled false}


Magisus:
Weird, the tests pass locally for me... Thoughts? I'm not familiar with this test suite at all.

justinstoller:
book keeping: the test failures are being worked in https://github.com/puppetlabs/puppetserver/pull/1482

resources/ext/cli_defaults/cli-defaults.sh.erb
```diff
@@ -7,4 +7,4 @@ if [ ! -e "$JRUBY_JAR" ]; then
   return 1
 fi
 
-CLASSPATH="${CLASSPATH}:${JRUBY_JAR}"
+CLASSPATH="${CLASSPATH}:${JRUBY_JAR}:/opt/puppetlabs/server/data/puppetserver/jars/*"
```
justinstoller: Does this need to be parameterized similarly to INSTALL_DIR?

Magisus: Probably couldn't hurt. Does this sometimes get used in the context of something in PE which causes the name to change, maybe?

Magisus: Updated.

jpinsonault:
rekicked travis after merging https://github.com/puppetlabs/puppetserver/pull/1482

Magisus:
Confirmed that this produces a package that looks right, see http://builds.delivery.puppetlabs.net/puppetserver/5.0.1.master.SNAPSHOT.2017.08.11T1543/artifacts/

adrienthebo:
Does this need acceptance or integration level testing?

jpinsonault:
Maybe, not sure if the risk warrants acceptance testing, or if there's a simple way we could test it. A custom jar and a custom puppet function that uses it might be more effort than it's worth?

MosesMendoza:
Could you also add a PR to https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md#puppetserver adding the new jar dir?

MosesMendoza:
Ping @shrug mind taking a look and validating our thinking on pathing here if you haven't?

resources/ext/cli_defaults/cli-defaults.sh.erb
```diff
@@ -7,4 +7,4 @@ if [ ! -e "$JRUBY_JAR" ]; then
   return 1
 fi
 
-CLASSPATH="${CLASSPATH}:${JRUBY_JAR}"
+CLASSPATH="${CLASSPATH}:${JRUBY_JAR}:/opt/puppetlabs/server/data/<%= EZBake::Config[:real_name] %>/jars/*"
```
shrug: Generally I'd expect anything under data to be, well, data and not something to be executed - but I'm not sure there's a better choice for this. We should validate the permissions that this gets created with, they should be fairly restrictive for write access, at least.

Magisus: The dir is created with a `700` permission, does that seem right? https://github.com/puppetlabs/ezbake/blob/fe21baaac64457935c540f3907815e1f57c51c46/resources/puppetlabs/lein-ezbake/template/global/install.sh.erb#L190-L192

Magisus: I chose to create the jar dir there because it is adjacent to where gems get installed.

Magisus:
@shrug did you have any further comments on this?

resources/ext/cli_defaults/cli-defaults.sh.erb
jpinsonault: Yeah, we might want to change the location. In general we (in the past, in puppetserver) put things under /opt/ that we don't intend the user to have to modify. Things we expect the user to modify have gone under /etc/puppetlabs/, and that might be better just because the user will have all their configuration there already. On the other hand we put all the other binaries under /opt/, so ¯\\\_(ツ)\_/¯

Magisus: I think that sounds like the split we generally follow on the agent side too.

PR #1482, opened 2017-08-10
(maint) Remove bindaddress config tests

adrienthebo:
Some clojure tests were inspecting setting information coming from
Puppet and Ruby to verify that the information could be fetched, but
this failed when Puppet itself changed a default value. This commit
changes the tested values to another string type value to check the same
code paths so we don't have to have tests depending on what version
of Puppet is in use.

### DIFF
--- a/test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
+++ b/test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
@@ -508,7 +508,7 @@
            jruby-puppet (:jruby-puppet jruby-instance)]
        (try
          (testing "Various data types"
-           (is (= "0.0.0.0" (.getSetting jruby-puppet "bindaddress")))
+           (is (= "ldap" (.getSetting jruby-puppet "ldapserver")))
            (is (= 8140 (.getSetting jruby-puppet "masterport")))
            (is (= false (.getSetting jruby-puppet "onetime"))))
          (finally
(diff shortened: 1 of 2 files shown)


test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
```diff
@@ -508,7 +508,7 @@
            jruby-puppet (:jruby-puppet jruby-instance)]
        (try
          (testing "Various data types"
-           (is (= "0.0.0.0" (.getSetting jruby-puppet "bindaddress")))
+           (is (= "puppet" (.getSetting jruby-puppet "user")))
```
MosesMendoza: 👍  as long as these tests aren't somehow exercised as a part of pe-puppetserver, where the user is pe-puppet ?

Magisus:
Ah, I see, we were just using these values incidentally to test something else. 👍

test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
adrienthebo: Fair point. How about the ldapserver value? That shouldn't change in the near future.

adrienthebo: I've amended the commit to use ldapserver.

PR #1484, opened 2017-08-11
(TK-149, SERVER-1866) Update acceptance test to enable CRL reloading

justinstoller:
Previously, we were testing whether or not CRL reloading worked in our
default configuration.

With recent work this only happens if explicitly enabled, recommended
only on Java 8.

This updates the test to enable the CRL reloading before attempting to
test for it, and only if the server is running Java 8.

### DIFF
--- a/acceptance/suites/tests/service_framework/reload_updated_crl.rb
+++ b/acceptance/suites/tests/service_framework/reload_updated_crl.rb
@@ -6,6 +6,11 @@ server = master.puppet['certname']
 host_crl_file = master.puppet['hostcrl']
 ca_crl_file = master.puppet['cacrl']
 inventory_file = master.puppet['cert_inventory']
+bootstrap = "/opt/puppetlabs/server/apps/#{options['puppetservice']}/config/services.d/bootstrap.cfg"
+watcher_service = 'puppetlabs.trapperkeeper.services.watcher.filesystem-watch-service/filesystem-watch-service'
+java_version = on(master, 'java -version').stdout
+
+skip_test 'Feature only supported on Java 8' unless java_version =~ /.* version "?1.8.*/
 
 teardown do
   step 'Remove revoked CRL cert' do
@@ -25,8 +30,12 @@ teardown do
         "then mv -f #{inventory_file}.bak #{inventory_file}; fi")
   end
 
+  step 'Disable crl reloading in bootstrap.cfg' do
+    on(master, "sed -i 's,#{watcher_service},##{watcher_service},' #{bootstrap}")
+  end
+
   step 'Restart puppetserver service after testing revoked CRL agent' do
-    bounce_service(master, puppetservice)
+    on(master, "service #{puppetservice} restart")
   end
 end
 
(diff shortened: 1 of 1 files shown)


acceptance/suites/tests/service_framework/reload_updated_crl.rb
```diff
@@ -25,8 +30,12 @@
         "then mv -f #{inventory_file}.bak #{inventory_file}; fi")
   end
 
+  step 'Disable crl reloading in bootstrap.cfg' do
+    on(master, "sed -i 's,#{watcher_service},##{watcher_service},' #{bootstrap}")
+  end
+
   step 'Restart puppetserver service after testing revoked CRL agent' do
-    bounce_service(master, puppetservice)
+    on(master, "service #{puppetservice} restart")
```
justinstoller: We need to restart the service here so we pick up the bootstrap changes.

acceptance/suites/tests/service_framework/reload_updated_crl.rb
```diff
@@ -6,6 +6,11 @@
 host_crl_file = master.puppet['hostcrl']
 ca_crl_file = master.puppet['cacrl']
 inventory_file = master.puppet['cert_inventory']
+bootstrap = "/opt/puppetlabs/server/apps/#{options['puppetservice']}/config/services.d/bootstrap.cfg"
+watcher_service = 'puppetlabs.trapperkeeper.services.watcher.filesystem-watch-service/filesystem-watch-service'
+java_version = on(master, 'java -version').stdout
+
+skip_test 'Feature only supported on Java 8' unless java_version =~ /.* version "?1.8.*/
```
justinstoller: Is this a horrible idea?

acceptance/suites/tests/service_framework/reload_updated_crl.rb
```diff
@@ -25,8 +30,12 @@
         "then mv -f #{inventory_file}.bak #{inventory_file}; fi")
   end
 
+  step 'Disable crl reloading in bootstrap.cfg' do
+    on(master, "sed -i 's,#{watcher_service},##{watcher_service},' #{bootstrap}")
```
justinstoller: Should we validate that the defaults are what we expect? Like, check that, in this branch, we have commented out the service before continuing? Potentially making that check prior to skipping if not on Java 8?

justinstoller:
@MosesMendoza I grabbed this just to figure out what was up (Just making sure we didn't totally break something yesterday). I don't know if this was the best approach though, feel free to run with it from here (;

acceptance/suites/tests/service_framework/reload_updated_crl.rb
jpinsonault: Might be worth putting that in a comment here

acceptance/suites/tests/service_framework/reload_updated_crl.rb
jpinsonault: 🙈 🙉

acceptance/suites/tests/service_framework/reload_updated_crl.rb
jpinsonault: From what I can see it doesn't sound like a concern, maybe I'm missing something

MosesMendoza:
superseded by https://github.com/puppetlabs/puppetserver/pull/1485

PR #1488, opened 2017-08-18
(maint) Manually bump puppet SHA and version

adrienthebo:


justinstoller:
I think we need to bump the submodule version as well

Magisus:
PR to fix the test failures: https://github.com/puppetlabs/puppet/pull/6134 Once this is merged, we will need to update this PR to point to that puppet ref instead.

PR #1489, opened 2017-08-19
(maint) optimizations for gem installations

underscorgan:
Invoking the jvm for every gem installation is very slow. In testing
chaining all the gem installs together removed ~6.5 minutes from build
times, and removing docs from the installs removed an additional 1.5
minutes.

### DIFF
--- a/resources/ext/build-scripts/install-vendored-gems.sh
+++ b/resources/ext/build-scripts/install-vendored-gems.sh
@@ -15,9 +15,11 @@ cat "${DIR}/gem-list.txt"
 
 echo "jruby-puppet: { gem-home: ${DESTDIR}/opt/puppetlabs/server/data/puppetserver/vendored-jruby-gems }" > jruby.conf
 
+gem_list=()
 while read LINE
 do
   gem_name=$(echo $LINE |awk '{print $1}')
   gem_version=$(echo $LINE |awk '{print $2}')
-  java -cp puppet-server-release.jar:jruby-1_7.jar clojure.main -m puppetlabs.puppetserver.cli.gem --config jruby.conf -- install ${gem_name} --version ${gem_version}
+  gem_list+=("$gem_name:$gem_version")
 done < "${DIR}/gem-list.txt"
+java -cp puppet-server-release.jar:jruby-1_7.jar clojure.main -m puppetlabs.puppetserver.cli.gem --config jruby.conf --no-ri --no-rdoc -- install "${gem_list[@]}"


MosesMendoza:
👍  thanks for the optimization.

The only thing that gives me pause is I'm not sure if we can use this `gem:version` syntax and also specify a platform-specific gem to install. Do you know if that's possible? I can see a world some day where we want to add in a `-java` platform-specific version of a gem for example. But we can always cross that bridge when we get there.

underscorgan:
@MosesMendoza do you have an example of a platform-specific gem so I can test?

MosesMendoza:
https://rubygems.org/gems/pry/versions/0.10.4-java

underscorgan:
@MosesMendoza having the versions like this doesn't affect the functionality of the `--platform` tag, but I think this could get weird if some of the gems have java platform gems and we want those and other have java platform gems but we don't want to install those. Passing `--platform java` and installing multiple gems where only one of them has a java platform seems to work. So, I think it would be reasonable to not worry about unless we end up in this situation because it feels like a weird edge case?

MosesMendoza:
agreed

underscorgan:
@MosesMendoza do you want review for this from anyone else from your team or is this good for merging? Is master the right place to target this?

MosesMendoza:
Hi @melissa  - yes good to merge but we're waiting on CI to go green before merging any more on master

underscorgan:
cool ,thanks @MosesMendoza

Magisus:
@underscorgan this appears to have failed in CI with
```
17:14:33 + java -cp puppet-server-release.jar:jruby-1_7.jar clojure.main -m puppetlabs.puppetserver.cli.gem --config jruby.conf --no-ri --no-rdoc -- install semantic_puppet:0.1.3 hocon:1.2.5 text:1.3.1 locale:2.1.2 gettext:3.2.2 gettext-setup:0.26 fast_gettext:1.1.0
17:14:37 
17:14:37 
17:14:37 Error(s) occurred while parsing command-line arguments: Unknown option: "--no-ri"Unknown option: "--no-rdoc"
17:14:37 
17:14:37   -c, --config CONFIG-PATH  Path to a configuration file or directory of configuration files. See the documentation for a list of supported file types.
17:14:37   -h, --help                Show help
17:14:37 debian/rules:21: recipe for target 'install/puppetserver' failed
17:14:37 make: *** [install/puppetserver] Error 1
```

Are those args maybe in the wrong spot?

underscorgan:
hm, weird. Will investigate, feel free to revert in the meantime.

PR #1491, opened 2017-08-22
(SERVER-1912) update pup submodule to fix missing method

MosesMendoza:
* picks up new puppet version that fixes issue regarding inaccessible method in
  i18n setup for modules

### DIFF
--- a/acceptance/config/beaker/options.rb
+++ b/acceptance/config/beaker/options.rb
@@ -11,5 +11,5 @@
  "puppetserver-confdir"=>"/etc/puppetlabs/puppetserver/conf.d",
  "puppetserver-config"=>
   "/etc/puppetlabs/puppetserver/conf.d/puppetserver.conf",
- :puppet_version=>"5.0.1.189.g60ee269",
- :puppet_build_version=>"4d5d1c6a7dcbdc4ec420a7878cc1b3a1e9c631aa"}
+ :puppet_version=>"5.0.1.281.g94da54a",
+ :puppet_build_version=>"700ad26637c8fdaf9e59923636943e533d490a47"}


MosesMendoza:
change submodule commit 60ee26915e0381be5246ae9ddebc518b63554bd9 to 94da54a4ccde03436d63cb590765fa42ff2e3d14, puppet/master latest as of 10:03 08-22

Magisus:
I'm currently building an agent to go along with this SHA, so we can update https://github.com/puppetlabs/puppetserver/blob/master/acceptance/config/beaker/options.rb#L14-L15

Magisus:
Builds are done, puppet-agent SHA is 700ad26637c8fdaf9e59923636943e533d490a47, puppet git-describe is 5.0.1-281-g94da54a (it appears we have not merged up the 5.1 branch to master yet).

MosesMendoza:
Updated

acceptance/config/beaker/options.rb
```diff
@@ -11,5 +11,5 @@
  "puppetserver-confdir"=>"/etc/puppetlabs/puppetserver/conf.d",
  "puppetserver-config"=>
   "/etc/puppetlabs/puppetserver/conf.d/puppetserver.conf",
- :puppet_version=>"5.0.1.189.g60ee269",
- :puppet_build_version=>"4d5d1c6a7dcbdc4ec420a7878cc1b3a1e9c631aa"}
+ :puppet_version=>"5.0.1-281-g94da54a",
```
Magisus: Think it matters that this contains dashes instead of dots? I'd honestly really like to know, because we keep having this problem and it's really annoying >.>

MosesMendoza: huh. yes it does matter. the dashes are converted to dots for packaging purposes, and this version string is used to grep for the right packages basically. thanks for the catch

MosesMendoza: updated

Magisus: Thanks for explaining that!

PR #1492, opened 2017-08-22
(SERVER-1925) List all tasks

lindboe:
This PR adds an endpoint that lists all tasks found in modules in a given environment. It queries Puppet's InfoService class to find this information. This will be consumed directly by orchestrator in PE, which will relay information to clients like the PE console or a command line tool so users know the tasks they have available to choose from. This endpoint is very similar to the environment_classes endpoint (although less complex), so much of the approach was copied from that code.

Task information is accessible for anyone with a valid certificate signed by the Puppet CA.

I wanted to get this up for review ASAP, so I don't have pre-docs yet. Currently working on them.

### DIFF
--- a/ezbake/config/conf.d/auth.conf
+++ b/ezbake/config/conf.d/auth.conf
@@ -137,6 +137,15 @@ authorization: {
             name: "puppetlabs static file content"
         },
         {
+            match-request: {
+                path: "/puppet/v3/tasks"
+                type: path
+            }
+            allow: "*"
+            sort-order: 500
+            name: "puppet tasks information"
+        },
+        {
             # Allow all users access to the experimental endpoint
             # which currently only provides a dashboard web ui.
             match-request: {
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -319,6 +319,28 @@
       (middleware-utils/json-response 200 info-for-json))))
 
 (schema/defn ^:always-validate
+  all-tasks-response! :- ringutils/RingResponse
+  "Process the info, returning a Ring response to be propagated back up to the
+  caller of the endpoint.
+  Returns environment as a list of objects for an eventual future in which
+  tasks for all environments can be requested, and a given task will list all
+  the environments it is found in."
+  [info-from-jruby :- List
+   environment :- schema/Str
+   jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)]
+  (let [format-task (fn [task-object]
+                      {:name (:name task-object)
+                       :environment [{:name environment
+                                      :code_id nil}]})]
+    (->> (map format-task info-from-jruby)
+         (middleware-utils/json-response 200))))
+
+(defn environment-not-found
+  "Ring handler to provide a standard error when an environment is not found."
+  [environment]
+  (rr/not-found (i18n/tru "Could not find environment ''{0}''" environment)))
+
+(schema/defn ^:always-validate
   environment-class-info-fn :- IFn
   "Middleware function for constructing a Ring response from an incoming
   request for environment_classes information."
@@ -341,7 +363,7 @@
                                      (if-none-match-from-request request)
                                      cache-generation-id
                                      environment-class-cache-enabled)
-        (rr/not-found (i18n/tru "Could not find environment ''{0}''" environment))))))
+        (environment-not-found environment)))))
 
 (schema/defn ^:always-validate
   module-info-from-jruby->module-info-for-json  :- EnvironmentModulesInfo
@@ -391,6 +413,23 @@
         (environment-module-response! module-info)))))
 
 (schema/defn ^:always-validate
+  all-tasks-fn :- IFn
+  "Middleware function for constructing a Ring response from an incoming
+  request for tasks information."
+  [jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)]
+  (fn [request]
+    (let [environment (jruby-request/get-environment-from-request request)]
+      (if-let [task-info-for-env
+               (jruby-protocol/get-tasks jruby-service
+                                         (:jruby-instance
+                                           request)
+                                         environment)]
+        (all-tasks-response! task-info-for-env
+                             environment
+                             jruby-service)
+        (environment-not-found environment)))))
+
+(schema/defn ^:always-validate
   wrap-with-etag-check :- IFn
   "Middleware function which validates whether or not the If-None-Match
   header on an incoming environment_classes request matches the last Etag
@@ -434,6 +473,15 @@
    (jruby-request/wrap-with-jruby-instance jruby-service)
    (jruby-request/wrap-with-environment-validation true)
    jruby-request/wrap-with-error-handling))
+
+(schema/defn ^:always-validate
+  all-tasks-handler :- IFn
+  "Handler for processing an incoming all_tasks Ring request"
+  [jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)]
+  (-> (all-tasks-fn jruby-service)
+      (jruby-request/wrap-with-jruby-instance jruby-service)
+      jruby-request/wrap-with-environment-validation
+      jruby-request/wrap-with-error-handling))
 
 (schema/defn ^:always-validate valid-static-file-path?
   "Helper function to decide if a static_file_content path is valid.
@@ -539,6 +587,8 @@
                                    environment-class-cache-enabled)
         environment-module-handler
         (environment-module-handler jruby-service)
+        all-tasks-handler
+        (all-tasks-handler jruby-service)
         static-file-content-handler
         (static-file-content-request-handler get-code-content-fn)]
     (comidi/routes
@@ -546,6 +596,8 @@
                   (environment-class-handler request))
       (comidi/GET ["/environment_modules" [#".*" :rest]] request
                   (environment-module-handler request))
+      (comidi/GET ["/tasks" [#".*" :rest]] request
+                  (all-tasks-handler request))
       (comidi/GET ["/static_file_content/" [#".*" :rest]] request
                   (static-file-content-handler request)))))
 
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -420,10 +420,11 @@
   (fn [request]
     (let [environment (jruby-request/get-environment-from-request request)]
       (if-let [task-info-for-env
-               (jruby-protocol/get-tasks jruby-service
-                                         (:jruby-instance
-                                           request)
-                                         environment)]
+               (sort-nested-info-maps
+                 (jruby-protocol/get-tasks jruby-service
+                                           (:jruby-instance
+                                             request)
+                                           environment))]
         (all-tasks-response! task-info-for-env
                              environment
                              jruby-service)
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -325,7 +325,7 @@
   Returns environment as a list of objects for an eventual future in which
   tasks for all environments can be requested, and a given task will list all
   the environments it is found in."
-  [info-from-jruby :- List
+  [info-from-jruby :- [{schema/Any schema/Any}]
    environment :- schema/Str
    jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)]
   (let [format-task (fn [task-object]
@@ -335,9 +335,9 @@
     (->> (map format-task info-from-jruby)
          (middleware-utils/json-response 200))))
 
-(defn environment-not-found
+(schema/defn environment-not-found :- ringutils/RingResponse
   "Ring handler to provide a standard error when an environment is not found."
-  [environment]
+  [environment :- schema/Str]
   (rr/not-found (i18n/tru "Could not find environment ''{0}''" environment)))
 
 (schema/defn ^:always-validate
--- a/src/ruby/puppetserver-lib/puppet/server/master.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/master.rb
@@ -102,6 +102,14 @@ class Puppet::Server::Master
     Puppet::Server::Config.terminate_puppet_server
   end
 
+   # @return [Array, nil] an array of hashes describing tasks
+  def getTasks(env)
+    environment = @env_loader.get(env)
+    unless environment.nil?
+      Puppet::InfoService.tasks_per_environment(environment.name)
+    end
+  end
+
   private
 
   def self.getModules(env)
--- a/test/integration/puppetlabs/puppetserver/testutils.clj
+++ b/test/integration/puppetlabs/puppetserver/testutils.clj
@@ -125,6 +125,30 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Interacting with puppet code and catalogs
 
+(schema/defn ^:always-validate module-dir
+  [env-name :- schema/Str
+   module-name :- schema/Str]
+  (fs/file conf-dir "environments" env-name "modules" module-name))
+
+(schema/defn ^:always-validate create-module :- schema/Str
+  [module-name :- schema/Str
+   options :- {(schema/optional-key :env-name) schema/Str
+               (schema/optional-key :module-version) schema/Str}]
+  (let [default-options {:env-name "production"
+                         :module-version "1.0.0"}
+        options (merge default-options options)
+        module-dir (module-dir (:env-name options) module-name)
+        metadata-json {"name" module-name
+                       "version" (:module-version options)
+                       "author" "Puppet"
+                       "license" "apache"
+                       "dependencies" []
+                       "source" "https://github.com/puppetlabs"}]
+    (fs/mkdirs module-dir)
+    (spit (fs/file module-dir "metadata.json")
+          (json/generate-string metadata-json))
+    (.getCanonicalPath module-dir)))
+
 (schema/defn ^:always-validate write-pp-file :- schema/Str
   ([pp-contents :- schema/Str
     module-name :- schema/Str]
@@ -143,33 +167,31 @@
     pp-name :- schema/Str
     env-name :- schema/Str
     module-version :- schema/Str]
-   (let [pp-file (fs/file conf-dir
-                          "environments"
-                          env-name
-                          "modules"
-                          module-name
+   (let [pp-file (fs/file (module-dir env-name module-name)
                           "manifests"
-                          (str pp-name ".pp"))
-         module-dir (fs/file conf-dir
-                             "environments"
-                             env-name
-                             "modules"
-                             module-name)
-         metadata-json {"name" module-name
-                        "version" module-version
-                        "author" "Puppet"
-                        "license" "apache"
-                        "dependencies" []
-                        "source" "https://github.com/puppetlabs"}]
+                          (str pp-name ".pp"))]
+     (create-module module-name {:env-name env-name :module-version module-version})
      (fs/mkdirs (fs/parent pp-file))
      (spit pp-file pp-contents)
-     (spit (fs/file module-dir "metadata.json")
-           (json/generate-string metadata-json))
      (.getCanonicalPath pp-file))))
 
 (schema/defn ^:always-validate write-foo-pp-file :- schema/Str
   [foo-pp-contents]
   (write-pp-file foo-pp-contents "foo"))
+
+(schema/defn ^:always-validate write-tasks-files :- schema/Str
+  [module-name :- schema/Str
+   task-name :- schema/Str
+   task-file-contents :- schema/Str]
+  (let [metadata-contents {"description" "This is a description. It describes a thing."}
+        module-dir (create-module module-name {})
+        tasks-dir (fs/file module-dir "tasks")
+        metadata-file-path (fs/file tasks-dir task-name ".json")]
+    (create-file metadata-file-path
+                 (json/encode metadata-contents))
+    (create-file (fs/file tasks-dir task-name ".sh")
+                 task-file-contents)
+    (.getCanonicalPath metadata-file-path)))
 
 (defn create-env-conf
   [env-dir content]
--- a/test/integration/puppetlabs/puppetserver/testutils.clj
+++ b/test/integration/puppetlabs/puppetserver/testutils.clj
@@ -125,7 +125,7 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Interacting with puppet code and catalogs
 
-(schema/defn ^:always-validate module-dir
+(schema/defn ^:always-validate module-dir :- File
   [env-name :- schema/Str
    module-name :- schema/Str]
   (fs/file conf-dir "environments" env-name "modules" module-name))
--- a/test/integration/puppetlabs/services/jruby/tasks_test.clj
+++ b/test/integration/puppetlabs/services/jruby/tasks_test.clj
@@ -10,7 +10,8 @@
             [schema.test :as schema-test]
             [puppetlabs.puppetserver.testutils :as testutils]
             [puppetlabs.trapperkeeper.app :as tk-app]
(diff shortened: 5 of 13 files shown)


src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -435,6 +474,16 @@
    (jruby-request/wrap-with-environment-validation true)
    jruby-request/wrap-with-error-handling))
 
+(schema/defn ^:always-validate
+  all-tasks-handler :- IFn
+  "Handler for processing an incoming all_tasks Ring request"
+  [jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)]
+  (->
```
holguinj: I think it's kinda weird to have a `->` on its own line like this

test/integration/puppetlabs/services/jruby/tasks_test.clj
```diff
@@ -0,0 +1,116 @@
...
+(defn make-task-dir
+  [env-dir mod-name]
+  (fs/file env-dir "modules" mod-name "tasks"))
+
+(schema/defn gen-task
+  "Assumes tasks dir already exists -- generates a set of task files for a
+  single task."
+  [env-dir task :- TaskOptions]
+  (let [task-name (:name task)
+        extensions '(".rb" "" ".sh" ".exe" ".py")
```
lindboe: ah yes, I had been looking for something like that. Thanks.

src/clj/puppetlabs/services/master/master_core.clj
lindboe: true! will fix

src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -102,6 +102,14 @@ def terminate
     Puppet::Server::Config.terminate_puppet_server
   end
 
+  def getTasks(env)
+    environment = @env_loader.get(env)
+    unless environment.nil?
+      tasks_per_env =
+        Puppet::InfoService.tasks_per_environment(environment.name)
+    end
```
lindboe: you're right the assignment isn't needed. that came from some previous implementations, will remove.

And yeah, I find it weird that returns nil if environment isn't found, but that's consistent with getClassInfoForEnvironment and other methods. The code overall seems to handle this by checking the environment exists before passing requests to these methods. I'm not sure I'm a fan, but consistency seemed better unless I was going to change this everywhere? I'm very open to other thoughts on this though; going to leave as-is until I get maintainer feedback.

lindboe:
Pushed fixes for the minor changes Justin suggested.

ezbake/config/conf.d/auth.conf
```diff
@@ -137,6 +137,15 @@ authorization: {
             name: "puppetlabs static file content"
         },
         {
+            match-request: {
+                path: "/puppet/v3/tasks"
+                type: path
```
haus: seems like locking this down to just GET might be good? I guess puppetserver will just 404 in the case of a PUT POST or DELETE so maybe it doesn't matter?

lindboe: Yeah, I don't expect we'll have anything that needs further restriction with regards to this endpoint, so even though it doesn't work, a 404 seems like a more clear result than saying you're not authorized to access something that turns out to be nothing

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -319,6 +319,28 @@
       (middleware-utils/json-response 200 info-for-json))))
 
 (schema/defn ^:always-validate
+  all-tasks-response! :- ringutils/RingResponse
+  "Process the info, returning a Ring response to be propagated back up to the
+  caller of the endpoint.
+  Returns environment as a list of objects for an eventual future in which
+  tasks for all environments can be requested, and a given task will list all
+  the environments it is found in."
+  [info-from-jruby :- List
```
haus: Are you going with `List` for future compatibility? It seems like a slightly more specific schema, even if it was likely to change later, might be better.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -319,6 +319,28 @@
...
+   environment :- schema/Str
+   jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)]
+  (let [format-task (fn [task-object]
+                      {:name (:name task-object)
+                       :environment [{:name environment
+                                      :code_id nil}]})]
+    (->> (map format-task info-from-jruby)
+         (middleware-utils/json-response 200))))
+
+(defn environment-not-found
```
haus: Putting some schema around this `defn` would be good.

src/clj/puppetlabs/services/master/master_core.clj
lindboe: Yeah, at the time I wasn't super sure how to write it because JRuby, but then that turned out to not be a thing for this function's input. I can make this better.

test/integration/puppetlabs/puppetserver/testutils.clj
```diff
@@ -5,7 +5,8 @@
             [cheshire.core :as json]
             [puppetlabs.http.client.sync :as http-client]
             [puppetlabs.kitchensink.core :as ks]
-            [puppetlabs.puppetserver.ringutils :as ringutils])
+            [puppetlabs.puppetserver.ringutils :as ringutils]
+            [clojure.string :as str])
```
haus: Looks like this is unused?

test/integration/puppetlabs/puppetserver/testutils.clj
```diff
@@ -125,6 +126,30 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Interacting with puppet code and catalogs
 
+(schema/defn ^:always-validate module-dir
```
haus: This schema could have a return type of `File` it looks like?

test/integration/puppetlabs/services/jruby/tasks_test.clj
```diff
@@ -0,0 +1,116 @@
...
+
+(use-fixtures :once schema-test/validate-schemas)
+
+(def TaskOptions
+  {:name schema/Str
+   :module-name schema/Str
+   :metadata? schema/Bool
+   :number-of-files schema/Num})
+
+(defn make-task-dir
```
haus: Nit: the name of this function is misleading as it doesn't actually make the directory.

test/integration/puppetlabs/services/jruby/tasks_test.clj
```diff
@@ -0,0 +1,116 @@
...
+  [env-dir mod-name]
+  (fs/file env-dir "modules" mod-name "tasks"))
+
+(schema/defn gen-task
+  "Assumes tasks dir already exists -- generates a set of task files for a
+  single task."
+  [env-dir task :- TaskOptions]
```
haus: it's a little weird when using schema to have a param that has a type share a line with an unschemaed param.

test/integration/puppetlabs/services/jruby/tasks_test.clj
```diff
@@ -0,0 +1,116 @@
...
+
+      (tk-bootstrap/with-app-with-config
+        app
+        jruby-testutils/jruby-service-and-dependencies
+        config
+        (let [jruby-service (tk-app/get-service app :JRubyPuppetService)
+              instance (jruby-testutils/borrow-instance jruby-service :test)
+              jruby-puppet (:jruby-puppet instance)
+              env-registry (:environment-registry instance)
+              _ (testutils/create-file (fs/file conf-dir "puppet.conf")
```
haus: Any reason not to make this file right above the `with-app-with-config` call so that it isn't in the `let` call?

test/integration/puppetlabs/services/master/tasks_int_test.clj
```diff
@@ -0,0 +1,101 @@
...
+            [puppetlabs.services.protocols.jruby-puppet :as jruby-protocol]
+            [cheshire.core :as cheshire]
+            [me.raynes.fs :as fs]
+            [clojure.tools.logging :as log]
+            [puppetlabs.services.jruby.jruby-puppet-testutils :as jruby-puppet-testutils])
+  (:import (com.puppetlabs.puppetserver JRubyPuppetResponse JRubyPuppet)
+           (java.util HashMap)))
+
+(def test-resources-dir
+  "./dev-resources/puppetlabs/services/master/environment_classes_int_test")
```
haus: It's been the pattern to copy the needed resources to a directory of the name of the current test.

test/integration/puppetlabs/services/master/tasks_int_test.clj
```diff
@@ -0,0 +1,101 @@
...
+              url)]
+    (try
+      (http-client/get url
+        (merge
+          testutils/ssl-request-options
+          {:as :text}))
+      (catch Exception e
+        (throw (Exception. "tasks http get failed" e))))))
+
+(defn purge-all-env-caches
```
haus: This and the next 2 functions look unused. Copypasta?

test/integration/puppetlabs/services/master/tasks_int_test.clj
```diff
@@ -0,0 +1,101 @@
...
+            [puppetlabs.trapperkeeper.testutils.bootstrap :as tk-bootstrap-testutils]
+            [puppetlabs.trapperkeeper.testutils.webserver :as jetty9]
+            [puppetlabs.services.master.master-core :as master-core]
+            [puppetlabs.services.protocols.jruby-puppet :as jruby-protocol]
+            [cheshire.core :as cheshire]
+            [me.raynes.fs :as fs]
+            [clojure.tools.logging :as log]
+            [puppetlabs.services.jruby.jruby-puppet-testutils :as jruby-puppet-testutils])
+  (:import (com.puppetlabs.puppetserver JRubyPuppetResponse JRubyPuppet)
+           (java.util HashMap)))
```
haus: HashMap import looks unused. The JRuby* imports also look unused.

src/ruby/puppetserver-lib/puppet/server/master.rb
holguinj: yeah, fair enough on the nil thing.

haus:
This looks overall fine to me functionally.

src/clj/puppetlabs/services/master/master_core.clj
lindboe: will do

test/integration/puppetlabs/puppetserver/testutils.clj
lindboe: thanks for looking for that stuff, I forgot to audit for it

test/integration/puppetlabs/services/master/tasks_int_test.clj
```diff
@@ -0,0 +1,101 @@
...
+   (str "https://localhost:8140/puppet-admin-api/v1/environment-cache?"
+        "environment="
+        env)
+   testutils/ssl-request-options))
+
+(defn response-etag
+  [request]
+  (get-in request [:headers "etag"]))
+
+(defn response->clojure
```
holguinj: it might be clearer if this were called `parse-response` or something

test/integration/puppetlabs/puppetserver/testutils.clj
lindboe: Yup, it could. Is it typical in this code base to always include a return type schema? I usually find the input types most useful for debugging purposes, return types on simple functions less so

test/integration/puppetlabs/services/jruby/tasks_test.clj
lindboe: will change

test/integration/puppetlabs/services/jruby/tasks_test.clj
lindboe: Yeah, I'll change that

test/integration/puppetlabs/services/jruby/tasks_test.clj
lindboe: No, that looks fine to me, I think this is an artifact of me breaking up this into two let blocks from the copied test

test/integration/puppetlabs/puppetserver/testutils.clj
haus: It's pretty typical to include return types when feasible.

test/integration/puppetlabs/services/master/tasks_int_test.clj
lindboe: will remove

test/integration/puppetlabs/services/master/tasks_int_test.clj
lindboe: Okay, I'll fix that

test/integration/puppetlabs/services/master/tasks_int_test.clj
lindboe: Most likely, will remove

test/integration/puppetlabs/services/master/tasks_int_test.clj
```diff
@@ -0,0 +1,101 @@
+(ns puppetlabs.services.master.tasks-int-test
+  (:require [clojure.test :refer :all]
```
haus: Of these requires, I don't think `clojure.string`, `puppetlabs.trapperkeeper.testutils.bootstrap`, `puppetlabs.trapperkeeper.testutils.webserver`, `puppetlabs.services.master.master-core`, `puppetlabs.services.protocols.jruby-puppet`, `clojure.tools.logging`, or `puppetlabs.services.jruby.jruby-puppet-testutils` are used.

test/integration/puppetlabs/services/jruby/tasks_test.clj
```diff
@@ -0,0 +1,116 @@
...
+                            set)
+                        (-> (get-tasks "env1")
+                            mc/sort-nested-info-maps
+                            set))
+                     "Unexpected info retrieved for 'env1'"))
+
+               (testing "for environment that does not exist"
+                 (is (nil? (get-tasks "env2"))))
+
+               (finally
```
haus: I'm not sure it's worth a comment here, but it took me a bit to figure out why we're explicitly returning the jruby-instance.

lindboe: Is that unusual? I just assumed that's how you have to do it

test/integration/puppetlabs/services/master/tasks_int_test.clj
lindboe: Can do

test/integration/puppetlabs/services/master/tasks_int_test.clj
lindboe: ugh, thanks

test/integration/puppetlabs/services/jruby/tasks_test.clj
haus: No, i don't think it's unusual. It might have just been far enough away from the borrow that i didn't connect them. But the other tests that do this don't explain why either, so this is fine.

test/integration/puppetlabs/services/master/tasks_int_test.clj
lindboe: oh lord, there's both puppetlabs.puppetserver.bootstrap-testutils and puppetlabs.trapperkeeper.testutils.bootstrap, that confused me momentarily

lindboe:
Last round of feedback addressed

src/ruby/puppetserver-lib/puppet/server/master.rb
MosesMendoza: looks like `tasks_per_env` got left behind

lindboe: Yeah, I haven't updated this part yet, wanted to leave the thread open for discussion on the nil part

test/integration/puppetlabs/puppetserver/testutils.clj
```diff
@@ -125,6 +125,31 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Interacting with puppet code and catalogs
 
+(schema/defn ^:always-validate
+  module-dir :- File
+  [env-name :- schema/Str
+   module-name :- schema/Str]
+  (fs/file conf-dir "environments" env-name "modules" module-name))
+
+(schema/defn ^:always-validate create-module :- schema/Str
```
MosesMendoza: awesome, thanks for pulling this out

test/integration/puppetlabs/puppetserver/testutils.clj
```diff
@@ -125,6 +125,31 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Interacting with puppet code and catalogs
 
+(schema/defn ^:always-validate
+  module-dir :- File
```
MosesMendoza: (nit) all the other tests have this on one line, any reason for the inconsistency here?

lindboe: Ah that was something I noticed was happening in `master-core`, so I copied that since I looked there last. Guess they're stylistically different across namespaces :)

haus: I think it may have been the case there to keep the lines under 80 chars.

test/integration/puppetlabs/services/jruby/tasks_test.clj
```diff
@@ -0,0 +1,115 @@
...
+  [env-dir :- (schema/cond-pre File schema/Str)
+   task :- TaskOptions]
+  (let [task-name (:name task)
+        extensions (cycle '(".rb" "" ".sh" ".exe" ".py"))
+        task-dir (fs/file env-dir "modules" (:module-name task) "tasks")]
+    (fs/mkdirs task-dir)
+    (when (:metadata? task)
+      (fs/create (fs/file task-dir (str task-name ".json"))))
+    (dotimes [n (:number-of-files task)]
+      (fs/create (fs/file task-dir (str task-name (nth extensions n ".rb")))))))
```
MosesMendoza: I interpreted `:number-of-files` as an upper-unbounded number of files to create but this will create the same five files repeatedly if passed a number > 5. Intentional?

lindboe: oh yeah duh of course it will. I can change that and limit the number, I can't imagine wanting to test more than that outside of perf tests.

test/integration/puppetlabs/services/jruby/tasks_test.clj
```diff
@@ -0,0 +1,115 @@
...
+  (dorun (map (partial gen-task env-dir) tasks)))
+
+(defn create-env
+  [env-dir tasks]
+  (testutils/create-env-conf env-dir "")
+  (gen-tasks env-dir tasks))
+
+(defn expected-tasks-info
+  [tasks]
+  (->> tasks
```
MosesMendoza: it feels odd to use the thread-last when there's only one form? i.e., it feels to me like it would be cleaner without it, as:
```clojure
  (map (fn [{:keys [name module-name]}]
        {:module {:name module-name}
          :name (if (= "init" name)
                  module-name
                  (str module-name "::" name))})
    tasks))
```
but i'm also new here so I could be off

test/integration/puppetlabs/services/jruby/tasks_test.clj
```diff
@@ -0,0 +1,115 @@
...
+  (testing "requesting all tasks"
+    (let [code-dir (ks/temp-dir)
+          conf-dir (ks/temp-dir)
+          config (jruby-testutils/jruby-puppet-tk-config
+                   (jruby-testutils/jruby-puppet-config
+                     {:master-code-dir (.getAbsolutePath code-dir)
+                      :master-conf-dir (.getAbsolutePath conf-dir)}))]
+
+      (testutils/create-file (fs/file conf-dir "puppet.conf")
+                             "[main]\nenvironment_timeout=unlimited\nbasemodulepath=$codedir/modules\n")
```
MosesMendoza: doesn't need to happen in this PR but I'm guessing this setup pattern is probably repeated a lot in our tests and may be worth pulling out to its own util/helper. not sure.

test/integration/puppetlabs/services/jruby/tasks_test.clj
```diff
@@ -0,0 +1,115 @@
...
+                          (.getTasks jruby-puppet env))]
+
+          (try (create-env env-1-dir env-1-tasks)
+               (testing "for environment that does exist"
+                 (is (= (-> env-1-tasks
+                            expected-tasks-info
+                            set)
+                        (-> (get-tasks "env1")
+                            mc/sort-nested-info-maps
+                            set))
```
MosesMendoza: Why do we cast the results from both of these as sets? Is there a way to so that wouldn't hide if there were unexpected duplicate entries in the results?

test/integration/puppetlabs/services/jruby/tasks_test.clj
lindboe: Yup, that's from an old implementation, will update

test/integration/puppetlabs/services/jruby/tasks_test.clj
lindboe: Yeah, I'm not sure I fully understand all the stuff that's happening with puppet.conf and environments in these tests, so I was reluctant to touch that

test/integration/puppetlabs/services/jruby/tasks_test.clj
lindboe: You can sort nested in both, I suppose, that does seem to be more common here.

test/integration/puppetlabs/services/master/tasks_int_test.clj
MosesMendoza: wow

lindboe:
I just pushed a docs commit. Review from the docs team will wait until I also write up the docs for the task detail endpoint, but I wanted to push these docs in case they were helpful to someone. So with that in mind, I'd like to keep review light on the docs for now (if you have larger suggestions that don't need to be immediately addressed, I can make a note of them to be addressed during the later edits).

src/ruby/puppetserver-lib/puppet/server/master.rb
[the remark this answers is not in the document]
lindboe: If no one wants to discuss the nil-if-env-doesn't-exist pattern, i'll just remove the assignment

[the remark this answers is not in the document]
MosesMendoza: IMO its fine as-is. We're not going to let a request through the indirector for an environment that doesn't exist anyway. Might be worth adding a yard tag on the method though noting that it will return nil in this case.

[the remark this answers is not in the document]
lindboe: A yard tag? Is yard generated for Puppet server?

Also, I don't think the Ruby method returns nil, I think it might throw? Which is why the unless guard exists in the other code.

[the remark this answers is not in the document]
MosesMendoza: Sorry - yes I'm confident we don't actually do anything with the tags in puppet server. A generic comment is fine. I was just thinking I wouldn't do any sort of cognitive double-take if I saw something like `# @return [Array, nil] an array of hashes describing tasks` here. 

And yes exactly - at the base level the implementation of `tasks_per_environment` will raise an exception if the environment doesn't exist because of `get!` here: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/info_service/task_information_service.rb#L6

lindboe:
Updated in response to feedback.

test/integration/puppetlabs/services/master/tasks_int_test.clj
```diff
@@ -0,0 +1,74 @@
...
+                                                "code_id" nil}]})
+            response (get-all-tasks "production")]
+        (testing "a successful status code is returned"
+          (is (= 200 (:status response))
+              (str
+                "unexpected status code for response, response: "
+                (ks/pprint-to-string response))))
+        (testing "the expected response body is returned"
+          (is (= expected-response
+                 (parse-response response))))))))
```
aperiodic: Is this supposed to be sorted already? It fails for me unless I sort the value from parse-response

lindboe: No, there's no determined sort order. I'll add a guard to sort it though, good call.

lindboe: Sort function added

MosesMendoza:
@lindboe want to squash?

PR #1497, opened 2017-08-25
(SERVER-1866) add docs, release notes

MosesMendoza:
This PR addresses two issues:

First, add documentation for automatic CRL refresh to the documentation directory and prepare for the 2.8.0 release with an updated release notes entry.

Second, Documentation suggests that users only configure services in /etc/puppetlabs/puppetserver/conf.d, not the config directory in /opt/puppetlabs. This commit adds a commented out entry for the file system watcher to the existing ca.cfg user configuration. While not in the `ca` namespace, since this setting affects the behavior of the CA it makes sense to place here rather than create a new additional file.

### DIFF
--- a/documentation/crl_refresh.markdown
+++ b/documentation/crl_refresh.markdown
@@ -1,0 +1,27 @@
+---
+layout: default
+title: "Puppet Server: Automatic CRL refresh"
+canonical: "/puppetserver/latest/crl_refresh.html"
+---
+
+Starting in version 2.8.0, Puppet Server can automatically reload an updated CRL into the running SSL context, so that the revocation of an agent's certificate no longer requires a restart of the service to take effect. Prior versions required an explicit restart or reload of this service to reload the CRL, resulting in some small amount of downtime to effect the revocation of a certificate. With automatic CRL refresh enabled, revocation is now transparent and requires no service downtime.
+
+### Enabling
+
+To enable automatic CRL refresh, modify your Puppet Server services bootstrap configuration file to include the following line:
+
+`puppetlabs.trapperkeeper.services.watcher.filesystem-watch-service/filesystem-watch-service`
+
+This line should already exist, commented out, in the default ca.cfg shipped in Puppet Server 2.8.0, but if you are upgrading and have modified your ca.cfg, adding this line manually may be required. See [Service Bootstraping](./configuration.markdown#service-bootstrapping) for information on how to update your Puppet Server's services bootstrap configuration.
+
+### Java Compatiblity
+
+_This feature is only recommended for systems running Java 8._ It exhibited some instability when Puppet Server was running on systems running Java 7. We believe this is possibly due to bugs and platform-specific behavior in the underlying Java file system watcher implementation that were addressed in Java 8 but not in Java 7.
+
+### Implementation
+
+Automatic CRL refresh leverages the the [trapperkeeper file system watcher](https://github.com/puppetlabs/trapperkeeper-filesystem-watcher) to watch for changes to the CRL file, and loads the updated CRL on change.
+
+### Contributors
+
+Thanks to Jeremy Barlow, who laid the groundwork for this feature in Puppet Server.
--- a/documentation/crl_refresh.markdown
+++ b/documentation/crl_refresh.markdown
@@ -4,11 +4,11 @@ title: "Puppet Server: Automatic CRL refresh"
 canonical: "/puppetserver/latest/crl_refresh.html"
 ---
 
-Starting in version 2.8.0, Puppet Server can automatically reload an updated CRL into the running SSL context, so that the revocation of an agent's certificate no longer requires a restart of the service to take effect. Prior versions required an explicit restart or reload of this service to reload the CRL, resulting in some small amount of downtime to effect the revocation of a certificate. With automatic CRL refresh enabled, revocation is now transparent and requires no service downtime.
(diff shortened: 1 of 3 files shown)


documentation/release_notes.markdown
```diff
@@ -12,6 +12,20 @@ canonical: "/puppetserver/latest/release_notes.html"
 
 For release notes on versions of Puppet Server prior to Puppet Server 2.5, see [docs.puppet.com](https://docs.puppet.com/puppetserver/2.4/release_notes.html).
 
+## Puppet Server 2.8.0
+
+Released August 31, 2017
+
+This is a feature release of Puppet Server.
+
+> **Warning:** If you're upgrading from Puppet Server 2.4 or earlier and have modified `bootstrap.cfg`, `/etc/sysconfig/puppetserver`, or `/etc/default/puppetserver`, see the [Puppet Server 2.5 release notes first](#potential-breaking-issues-when-upgrading-with-a-modified-bootstrapcfg) **before upgrading** for instructions on avoiding potential failures.
```
justinstoller: What are the settings in sysconfig/default that this effects?

documentation/crl_refresh.markdown
```diff
@@ -0,0 +1,27 @@
+---
+layout: default
+title: "Puppet Server: Automatic CRL refresh"
+canonical: "/puppetserver/latest/crl_refresh.html"
+---
+
+Starting in version 2.8.0, Puppet Server can automatically reload an updated CRL into the running SSL context, so that the revocation of an agent's certificate no longer requires a restart of the service to take effect. Prior versions required an explicit restart or reload of this service to reload the CRL, resulting in some small amount of downtime to effect the revocation of a certificate. With automatic CRL refresh enabled, revocation is now transparent and requires no service downtime.
```
jtappa: s/effect/affect

Magisus: I'm reading this as "a small amount of downtime to make the revocation happen", in which case "effect" is correct.

jtappa: Sorry, didn't realize there were two on this line, I'm referring to the second–

"...some small amount of downtime to effect the revocation of..." should be "...some small amount of downtime to **affect** the revocation of"

jtappa: Affect is a verb, effect is a noun. In this case, it's affect.

Magisus: Effect is also a verb, see the second definition here: http://www.dictionary.com/browse/effect?s=t That is the usage in the sentence about downtime, if I'm understanding Moses' description here.

gguillotte: I agree with @Magisus that effect is technically being used correctly here, but it might be simpler and less confusing to simply say "downtime to revoke a certificate".

documentation/crl_refresh.markdown
```diff
@@ -0,0 +1,27 @@
...
+
+### Enabling
+
+To enable automatic CRL refresh, modify your Puppet Server services bootstrap configuration file to include the following line:
```
gguillotte: Can we specify up front that this should be `/etc/puppetlabs/puppetserver/services.d/ca.cfg`, at least by default?

documentation/crl_refresh.markdown
```diff
@@ -0,0 +1,27 @@
...
+
+This line should already exist, commented out, in the default ca.cfg shipped in Puppet Server 2.8.0, but if you are upgrading and have modified your ca.cfg, adding this line manually may be required. See [Service Bootstraping](./configuration.markdown#service-bootstrapping) for information on how to update your Puppet Server's services bootstrap configuration.
+
+### Java Compatiblity
+
+_This feature is only recommended for systems running Java 8._ It exhibited some instability when Puppet Server was running on systems running Java 7. We believe this is possibly due to bugs and platform-specific behavior in the underlying Java file system watcher implementation that were addressed in Java 8 but not in Java 7.
+
+### Implementation
+
+Automatic CRL refresh leverages the the [trapperkeeper file system watcher](https://github.com/puppetlabs/trapperkeeper-filesystem-watcher) to watch for changes to the CRL file, and loads the updated CRL on change.
```
gguillotte: s/the the/the

documentation/release_notes.markdown
```diff
@@ -12,6 +12,20 @@ canonical: "/puppetserver/latest/release_notes.html"
 
 For release notes on versions of Puppet Server prior to Puppet Server 2.5, see [docs.puppet.com](https://docs.puppet.com/puppetserver/2.4/release_notes.html).
 
+## Puppet Server 2.8.0
+
+Released August 31, 2017
```
gguillotte: Is this still the expected release date?

jpinsonault: Yeah, that's my understanding

documentation/release_notes.markdown
gguillotte: [JVM memory allocation](https://docs.puppet.com/puppetserver/latest/install_from_packages.html#memory-allocation) and [max heap size](https://docs.puppet.com/puppetserver/latest/tuning_guide.html#jvm-heap-size).

documentation/crl_refresh.markdown
MosesMendoza: Apologies - I explicitly intended `effect` the verb form, but I agree it could be clearer. Will update.

MosesMendoza:
Updated per feedback

PR #1498, opened 2017-08-28
(SERVER-1898) Update version for 2.8.0 release

jpinsonault:
This commit will automatically trigger CI to build packages in prep for the 2.8.0 release

Should probably wait for https://github.com/puppetlabs/puppetserver/pull/1497 to get merged first

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,4 @@
-(def ps-version "2.7.3-SNAPSHOT")
+(def ps-version "2.8.0")
 (def tk-jetty9-version "1.8.1")
 
 (defn deploy-info


Magisus:
For server, do we usually wait to bump the version till all the code has landed? Or is it alright to increment this number ahead of time? (we weren't very strict about it in Agent world, and for a while even had a policy of bumping it immediately following a release, in preparation for the next one)

jpinsonault:
We should wait. Just figured I would put this PR to be ready. We should definitely wait for any code changes, since CI will build packages for 2.8.0 once this is merged, and those will be the artifacts we release later. But I think all the code changes are in. All that's left is a docs ticket, https://github.com/puppetlabs/puppetserver/pull/1497 

It'd be better to wait for that though since the sha that we tag should ideally contain those docs.

Magisus:
Cool, got it. Good to know that that number actually does something. In Ruby and C++ land, it's purely cosmetic >.>

jpinsonault:
Yeah I'm not sure how I feel about it, seems too magical and "helpful"

PR #1499, opened 2017-08-30
(SERVER-1933) Copy cacrl to hostcrl when cacrl file changes

Magisus:
The latest version of trapperkeeper-webserver-jetty9 enabled hot
reloading of the CRL. Whenever this reload happens, we want to copy
those changes to our hostcrl. This commit uses the filesystem-watcher to
watch for changes to the cacrl file and trigger an update to our hostcrl
when they occur.

This functionality can be disabled by not loading the filesystem-watcher
service.

This feature has been ported from the 2.x branch.


test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
```diff
@@ -276,3 +278,133 @@
...
+               (let [node-response-before-revoke (client-request)]
+                 (is (= 200 (:status node-response-before-revoke)))))
+             (testing "node certificate can be successfully revoked"
+               (let [revoke-response (cert-status-request "revoked")]
+                 (is (= 204 (:status revoke-response)))))
+             (testing "node request after revocation fails"
+               (let [ssl-exception-for-request? #(try
+                                                   (client-request)
+                                                   false
+                                                   (catch SSLException e
```
Magisus: In the newer version of Jetty, the exception thrown here has apparently changed from ConnectionClosed to SSLException.

MosesMendoza: reminds me of https://github.com/puppetlabs/puppetserver/commit/fd17c79e85a3b8584feedc75edee554a0040a724

Magisus: I think I didn't bring that commit in with this because it seemed Java 7 specific.

MosesMendoza: Yep that was my hypothesis at the time.. maybe I was wrong

Magisus: Well I haven't seen any IOExceptions. All that SSLException, which is definitely tied to newer Jetty.

Magisus:
Moses and I validated this manually by standing up a local puppetserver and checking in with Puppet, then revoking the cert and making sure the agent could not check in again, without restarting the server. Everything worked as expected, and the error emitted was consistently `certificate unknown`. Since this seems to have the desired behavior, we concluded that the changed error message must result from a change in the underlying libraries. `certificate unknown` seems like a better error message than `connection closed` in this case, so I'd say it's a positive change.

dev/bootstrap.cfg
```diff
@@ -16,6 +16,8 @@ puppetlabs.trapperkeeper.services.metrics.metrics-service/metrics-service
 puppetlabs.trapperkeeper.services.metrics.metrics-service/metrics-webservice
 puppetlabs.services.jruby.jruby-metrics-service/jruby-metrics-service
 puppetlabs.services.analytics.analytics-service/analytics-service
+# To disable automatic CRL reloading, comment the line below.
+puppetlabs.trapperkeeper.services.watcher.filesystem-watch-service/filesystem-watch-service
```
MosesMendoza: This bootstrap.cfg is just used for dev environment purposes. There's a couple other .cfg's that actually go into production. I think we want to add this to the ca.cfg, which ultimately lands in the conf dir we expect users to edit.

documentation/release_notes.markdown
```diff
@@ -12,7 +12,19 @@ canonical: "/puppetserver/latest/release_notes.html"
...
-## Puppet Server 5.0
+## Puppet Server 5.1.0
+
+Released September 12, 2017
+
+This is a feature release of Puppet Server.
+
+### New Feature: Automatic CRL refresh on certificate revocation
+
+Puppet Server 5.1.0 includes the ability to automatically refresh the CRL in the running SSL context when any changes to that file have occurred, namely the addition of a revoked certificate. Prior to this release, revoking an agent's certificate required restarting the Puppet Server process before that revocation would be honored and the agent denied authentication. Now revocation will be effective psuedo-immediately (some threshold of milliseconds) without requiring a restart of the server. This feature is enabled by default See [automatic CRL refresh](./crl_refresh.markdown) for details on disabling.
```
MosesMendoza: Add a `.` between default and See?

dev/bootstrap.cfg
MosesMendoza: Perfect. But we should probably also remove the line from the ezbake bootstrap.cfg because both it and the ca.cfg go live in packages, and the resulting service load set is the union of both files - which means as-is they could comment it out in ca.cfg and it would still get loaded.

Magisus: Okay good to know. Do we want to update 2.x to only have it in one spot then too? It's currently commented out in both places.

MosesMendoza: I figured it was fine to leave in 2.x because a user would only actually uncomment it one place (the ca.cfg) but it might be useful to us to know its there / available via the internal bootstrap.cfg ? not sure

Magisus: Works for me. So long as nothing bad happens when it's included twice, that seems fine. If you've uncommented at least one, you're expecting the feature to be enabled.

MosesMendoza: right

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -371,6 +372,24 @@
   (fs/create (fs/file path))
   (spit path (format-serial-number 1)))
 
+(schema/defn write-local-cacrl! :- (schema/maybe Exception)
+  "Spits the contents of 'cacrl' string to the localcacrl' file location if
```
justinstoller: missing a beginning `'` before localcacrl.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -371,6 +372,24 @@
...
+  the 'cacrl' string contains valid CRL pem data. On success, return nil.
+  On failure, return the Exception captured from the failed attempt to parse
+  the CRL pem data."
+  [localcacrl :- schema/Str
+    cacrl :- schema/Str]
```
justinstoller: this line is indented one space too far.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -747,9 +766,39 @@
   should be copied to."
   ([cacrl :- schema/Str
     localcacrl :- schema/Str]
-    (when (fs/exists? cacrl)
-      (ks/mkdirs! (fs/parent localcacrl))
-      (fs/copy cacrl localcacrl))))
+    (when (not= cacrl localcacrl)
```
justinstoller: This, while the same indentation level as the previous code, is also off by one.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -747,9 +766,39 @@
...
+      (let [max-attempts 25]
+        (when (and (fs/exists? cacrl))
```
justinstoller: This `and` is unnecessary.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -747,9 +766,39 @@
...
+                                       "Unable to synchronize crl file {0} to {1}: {2}"
+                                       cacrl localcacrl (.getMessage write-exception))
+                                     (i18n/trs
+                                       "Recent changes to the CRL may not have taken effect.")
+                                     (i18n/trs
+                                       "To load the updated CRL, reload or restart the Puppet Server service.")))
+                  (do
+                    (Thread/sleep 100)
+                    (let [new-ca-crl-last-modified (fs/mod-time cacrl)]
+                      (recur (if (= ca-crl-last-modified new-ca-crl-last-modified)
```
justinstoller: Two things, 
This seems dangerous. If, in attempting to write the CRL file out, it updates the file's modified time but still throws an IOException then this will loop forever. I think I'd prefer to keep this the same as the method in 2.x.

Secondly, if we did do it this way, I think we should assign a local variable for the output of the conditional have the recur be the last thing.

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
```diff
@@ -10,32 +12,53 @@
 
 (tk/defservice certificate-authority-service
   CaService
-  [[:PuppetServerConfigService get-config get-in-config]
-   [:WebroutingService add-ring-handler get-route]
-   [:AuthorizationService wrap-with-authorization-check]]
+  {:required [[:PuppetServerConfigService get-config get-in-config]
+              [:WebroutingService add-ring-handler get-route]
+              [:AuthorizationService wrap-with-authorization-check]]
+   :optional [FilesystemWatchService]}
```
justinstoller: Originally we didn't make this optional, but then made it optional because we didn't expect it to work on all supported versions of Java. We no longer support versions of Java that we were concerned about so I think it makes sense to go back to how we were originally going to do it have this be mandatory.

iow, why would our users want to turn automatic CRL reloading off? and why would we let them?

justinstoller:
Most of my comments are cosmetic, though I'm concerned about the change to max-attempts, and would rather us ship this as a mandatory feature in the 5.x series and have as few toggles as needed.

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
MosesMendoza: I think that's fair. Talking with @haus he also suggested we shouldn't make it optional in 5.x. My thinking behind leaving it as optional (and shipping it on by defaults) is that if for some reason this doesn't perform well or falls over under load, we can just tell people how to turn it off without needing to immediately ship them a new release

MosesMendoza: (not that we do that with any other features, so i'm probably just being too nervous there)

src/clj/puppetlabs/puppetserver/certificate_authority.clj
Magisus: I don't see this one. Don't we want it to be two spaces inside the previous parenthesis level? And if only one, why?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
Magisus: Thanks, looks like I missed a spot where we updated that in 2.x as well. Fixing.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
justinstoller: See the function above this one for the way it's supposed to look, this comes from
https://github.com/bbatsov/clojure-style-guide#multiple-arity-indentation

Basically, we want to arguments and body to line up.

However, neither this nor the one before it should have the additional parens around them. It looks like maybe both of these used to be multi-arity? Or maybe the style has changed? I think this is some pretty old code.

Sorry, I didn't originally expand the code above this to see that it *wasn't* a multi-arity function. I would recommend removing the set of parens around the arguments with that knowledge.

Magisus: Okay that makes sense.

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
justinstoller: I wrote a big long thing about this, but it boiled down to: I think this is the kind of thing that is a good candidate for toggling, but we've already committed to shipping this on by default commercially for our customers that are most likely to be affected by it, so if we are concerned about its performance then I think we should be doing more testing there rather than making it toggle-able here.

MosesMendoza: 👍  that makes sense to me

documentation/crl_refresh.markdown
```diff
@@ -0,0 +1,19 @@
+---
+layout: default
+title: "Puppet Server: Automatic CRL refresh"
+canonical: "/puppetserver/latest/crl_refresh.html"
+---
+
+Starting in version 5.1.0, Puppet Server can automatically reload an updated CRL into the running SSL context, so that the revocation of an agent's certificate no longer requires a restart of the service to take effect. Prior versions required an explicit restart or reload of this service to reload the CRL, resulting in some small amount of downtime to effect the revocation of a certificate. With automatic CRL refresh enabled, revocation is now transparent and requires no service downtime.
```
justinstoller: > With automatic CRL refresh enabled

Is redundant now

Magisus:
Updated.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -730,26 +749,51 @@
...
+        (ks/mkdirs! (fs/parent localcacrl))
+        ;; Make up to 'max-attempts' tries to copy the cacrl file to the
+        ;; localcacrl file. The content of the cacrl file is read into memory
+        ;; and parsed for valid CRL pem content during each attempt. The content
+        ;; in memory is written to the localcacrl file only if it is valid. This
+        ;; validation is done to protect against a partially written (which could
+        ;; happen if an asynchronous revocation is in progress) or currupt cacrl
+        ;; file being copied.
+        (loop [attempts-left max-attempts]
+           (let [cacrl-as-string (slurp cacrl)]
```
MosesMendoza: nit > i think this is indented one space to far

PR #1502, opened 2017-08-30
(SERVER-1927) Add /tasks/:module/:task endpoint for detailed task information

holguinj:
The purpose of this endpoint is to take some low-level data from puppet's `TaskInformationService`, flesh it out using some additional logic, and serve it to clients who will (probably) end up fetching an actual task executable from the `tasks` mount point.

The PR is divided into two commits:

1. Add the `getTaskData` method to the `JRubyPuppet` interface, which mostly just passes through to the `TaskInformationService`'s `task_data` method.
2. Add the `describe-task` function, which reads the files from the TaskData and returns the detailed task description, and the `/tasks/:module/:task` endpoint that calls that function.

### DIFF
--- a/documentation/puppet-api/v3/task_detail.json
+++ b/documentation/puppet-api/v3/task_detail.json
@@ -1,0 +1,63 @@
+{
+  "$schema":     "http://json-schema.org/draft-04/schema#",
+  "title":       "Task detail",
+  "description": "Detailed information about a specific task",
+  "type": "object",
+  "properties": {
+    "metadata": {
+      "description": "The contents of the <task name>.json metadata file. This will be empty if the file doesn't exist, or is empty. Otherwise it is allowed to return arbitrary keys here.",
+      "type": "object",
+      "properties": {},
+      "required": [],
+      "additionalProperties": true
+    },
+    "files": {
+      "description": "A list of executable files associated with the requested task, and data about them. The data is intended to assist in fetching their contents from the task file-content endpoint.",
+      "type": "array",
+      "items": {
+        "description": "A single file associated with the requested task, and data about it.",
+        "type": "object",
+        "properties": {
+          "filename": {
+            "description": "The name of the file (which should match the second part of the task name) plus its extension.",
+            "type": "string"
+          },
+          "sha256": {
+            "description": "The SHA256 of the contents of the file.",
+            "type": "string"
+          },
+          "size_bytes": {
+            "description": "The size of the file, in bytes.",
+            "type": "number"
+          },
+          "uri": {
+            "description": "Information on how to request the file contents from a Puppet master node. This will only provide a relative path, since clients may want to request from a compile master instead of the primary master.",
+            "type": "object",
+            "properties": {
+              "path": {
+                "description": "A relative URI for accessing the task contents. A client can form the full URI by requesting 'master:port/path?params'.",
+                "type": "string"
+              },
+              "params": {
+                "description": "A map of query params to use when requesting task contents. Required to include in the URI when making a request for task contents.",
+                "type": "object",
+                "properties": {
+                  "environment": {
+                    "description": "The environment the task is in, a required query param on request to task contents.",
+                    "type": "string"
+                  }
+                },
+                "required": ["environment"],
+                "additionalProperties": true
+              }
+            },
+            "required": ["path", "params"],
+            "additionalProperties": false
+          }
+        },
+        "required": ["metadata", "files", "code_id"],
+        "additionalProperties": false
+      }
+    }
+  }
+}
--- a/documentation/puppet-api/v3/task_detail.markdown
+++ b/documentation/puppet-api/v3/task_detail.markdown
@@ -1,7 +1,7 @@
 ---
 layout: default
-title: "Puppet Server: Puppet API: Tasks"
-canonical: "/puppetserver/latest/puppet-api/v3/tasks.html"
+title: "Puppet Server: Puppet API: Task detail"
+canonical: "/puppetserver/latest/puppet-api/v3/task-detail.html"
 ---
 
 [deprecated WEBrick Puppet master]: https://docs.puppet.com/puppet/latest/reference/services_master_webrick.html
@@ -17,6 +17,11 @@ added '.json' extension. For example, the "install" task in a module "apache" co
 consist of the executable file `install.rb` and the metadata file
 `install.json`. This task would have the display name "apache::install".
 
+This endpoint, `/puppet/v3/tasks/:module/:taskname`, allows you to fetch the
+details about a task: its metadata, if present, and its associated executable
+files. The file entries have additional data on how to fetch their contents so
+they can be downloaded and run.
+
 This endpoint is available only when the Puppet master is running Puppet Server, not
 on Ruby Puppet masters, such as the [deprecated WEBrick Puppet master][]. It also ignores
 the Ruby-based Puppet master's authorization methods and configuration. See the
@@ -30,15 +35,16 @@ the regular expression `\A[a-z][a-z0-9_]*\z` (excluding extensions).
 A task will be listed if only metadata for it exists. How many files are
 associated with a task can be found by querying that task's details.
 
-### Does not read files
-This endpoint will not parse metadata or read any other files, only file names.
+### Does read files
+This endpoint will read in contents of metadata and other task files, so it may
+be more expensive than the `/tasks` endpoint.
 
 ### Uses `application/json` Content-Type
 
 The Content-Type in the response to an task API query is
 `application/json`.
 
-## `GET /puppet/v3/tasks?environment=:environment`
+## `GET /puppet/v3/tasks/:module/:task?environment=:environment`
 
 (Introduced in Puppet Server 5.1.0.)
 
@@ -65,32 +71,35 @@ environment will be returned for the call.
 #### GET request with results
 
 ```
-GET /puppet/v3/tasks?environment=env
+GET /puppet/v3/tasks/module/taskname?environment=env
 
 HTTP/1.1 200 OK
 Etag: b02ede6ecc432b134217a1cc681c406288ef9224
 Content-Type: application/json
 
-[
-  {
-    "name":"apache::install",
-    "environment":[
-      {
-        "name":"env",
-        "code_id":null
+{
+  "metadata": {
+    "description": "Install a package",
+    "parameters": {
+      "name": {
+        "description": "The package to install",
+        "type": "String[1]"
       }
-    ]
+    }
   },
-  {
-    "name":"dashboard::configure",
-    "environment":[
-      {
-        "name":"env",
-        "code_id":null
+  "files": [
+    {"filename": "taskname.rb",
+      "SHA256": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
+      "size_bytes": 1024,
+      "uri:" {
+        "path": "/puppet/v3/file_content/tasks/module/taskname.rb",
+        "params": {
+          "environment": "production"
+        }
       }
-    ]
-  }
-]
+    }
+  ]
+}
 ```
 
 #### Environment does not exist
@@ -99,7 +108,7 @@ If you send a request with an environment parameter that doesn't correspond to t
 directory environment on the server, the server returns an HTTP 404 (Not Found) error:
 
 ```
-GET /puppet/v3/tasks?environment=doesnotexist
+GET /puppet/v3/tasks/module/taskname?environment=doesnotexist
 
 HTTP/1.1 404 Not Found
 
@@ -109,7 +118,7 @@ Could not find environment 'doesnotexist'
 #### No environment given
 
 ```
-GET /puppet/v3/tasks
+GET /puppet/v3/tasks/module/taskname
 
 HTTP/1.1 400 Bad Request
 
@@ -119,7 +128,7 @@ You must specify an environment parameter.
 #### Environment parameter specified with no value
 
 ```
-GET /puppet/v3/tasks?environment=
+GET /puppet/v3/tasks/module/taskname?environment=
 
 HTTP/1.1 400 Bad Request
 
@@ -132,17 +141,45 @@ If the environment parameter in your request includes any characters that are
 not `A-Z`, `a-z`, `0-9`, or `_` (underscore), the server returns an HTTP 400 (Bad Request) error:
 
 ```
-GET /puppet/v3/tasks?environment=bog|us
+GET /puppet/v3/tasks/module/taskname?environment=bog|us
 
 HTTP/1.1 400 Bad Request
 
 The environment must be purely alphanumeric, not 'bog|us'
 ```
 
+#### Module does not exist
+
+If you send a request for a task in a module that doesn't correspond to the
+name of a module on the server, the server returns an HTTP 404 (Not Found)
+error:
+
+```
+GET /puppet/v3/tasks/doesnotexist/taskname?environment=env
+
+HTTP/1.1 404 Not Found
+
+Could not find module 'doesnotexist'
+```
+
+#### Task does not exist or does not have a valid name
+
+If you send a request for a task in that doesn't correspond to the name of a
+task on the server, but the module does exist, the server returns an HTTP 404
+(Not Found) error:
+
+```
+GET /puppet/v3/tasks/module/doesnotexist?environment=env
+
+HTTP/1.1 404 Not Found
+
+Could not find task 'doesnotexist'
+```
+
 ### Schema
 
-A tasks response body conforms to the
-[tasks schema](./tasks.json).
+A tasks detail response body conforms to the
+[task detail schema](./task_detail.json).
 
 ### Authorization
 
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -3,6 +3,7 @@
            (clojure.lang IFn)
            (java.util List Map Map$Entry)
            (org.jruby RubySymbol)
+           (org.jruby.exceptions RaiseException)
            (com.codahale.metrics MetricRegistry Gauge)
            (java.lang.management ManagementFactory))
   (:require [me.raynes.fs :as fs]
@@ -25,7 +26,9 @@
             [puppetlabs.metrics.http :as http-metrics]
             [puppetlabs.http.client.metrics :as http-client-metrics]
             [puppetlabs.http.client.common :as http-client-common]
-            [puppetlabs.trapperkeeper.services.status.status-core :as status-core]))
+            [puppetlabs.trapperkeeper.services.status.status-core :as status-core]
+            [clojure.java.io :as io]
+            [clojure.tools.logging :as log]))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Constants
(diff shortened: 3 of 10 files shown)


test/integration/puppetlabs/services/jruby/tasks_test.clj
```diff
@@ -58,8 +74,8 @@
                   (str module-name "::" name))})
        tasks))
 
-(deftest ^:integration all-tasks-test
-  (testing "requesting all tasks"
+(deftest ^:integration task-data-test
```
holguinj: The diff is confusing here. The all-tasks-test is below this one, I didn't replace it.

test/integration/puppetlabs/services/jruby/tasks_test.clj
```diff
@@ -89,6 +103,85 @@
...
+
+          (try (create-env env-1-dir env-1-tasks)
+            (testing "when the environment, module, and task do exist"
+              (testing "with the init task"
+                (let [res (get-task-data "env1" "apache" "init")]
+                  (is (nil? (schema/check mc/TaskData res)))))
+
+              (testing "with another named task"
+                (let [res (get-task-data "env1" "apache" "install")]
+                  (is (nil? (schema/check mc/TaskData res))))))
```
lindboe: These assertions should probably also make sure one piece of data is correct to make sure we got the right thing, like a file in the files list with "install" in the name

test/integration/puppetlabs/services/jruby/tasks_test.clj
```diff
@@ -89,6 +103,85 @@
...
+                              (-> (.getTaskData jruby-puppet env module task)
+                                  mc/sort-nested-info-maps))]
+
+          (try (create-env env-1-dir env-1-tasks)
+            (testing "when the environment, module, and task do exist"
+              (testing "with the init task"
+                (let [res (get-task-data "env1" "apache" "init")]
+                  (is (nil? (schema/check mc/TaskData res)))))
+
+              (testing "with another named task"
```
lindboe: Can there be a test case for tasks with metadata but no other files? Maybe you tested that at a higher level

src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -110,6 +110,11 @@ def getTasks(env)
     end
   end
 
+  def getTaskData(environment_name, module_name, task_name)
+    qualified_task_name = task_name == 'init' ? module_name : "#{module_name}::#{task_name}"
```
lindboe: This line's a bit hard to read, I was thinking maybe some parens around the equality comparison might help? Or a question mark in the assignment variable (`qualified_task_name?`)
And possibly a comment explaining this interface, since it's a little hard to figure out why it is the way it is

test/integration/puppetlabs/services/jruby/tasks_test.clj
```diff
@@ -21,33 +22,48 @@
...
-  (dorun (map (partial gen-task env-dir) tasks)))
+  (dorun (map (partial gen-empty-task env-dir) tasks)))
 
 (defn create-env
   [env-dir tasks]
   (testutils/create-env-conf env-dir "")
-  (gen-tasks env-dir tasks))
+  (gen-empty-tasks env-dir tasks))
+
+(defn env-dir
```
lindboe: Does this actually get used? Looks like it's defined in a let block in the test. Did you mean to factor this out of the two tests?

test/integration/puppetlabs/services/jruby/tasks_test.clj
```diff
@@ -58,8 +74,8 @@
                   (str module-name "::" name))})
        tasks))
 
-(deftest ^:integration all-tasks-test
-  (testing "requesting all tasks"
+(deftest ^:integration task-data-test
+  (testing "requesting data about a specific task"
     (let [code-dir (ks/temp-dir)
           conf-dir (ks/temp-dir)
           config (jruby-testutils/jruby-puppet-tk-config
```
lindboe: This is repeating code that was factored out in a function below this deftest form (puppet-tk-config)

test/integration/puppetlabs/services/jruby/tasks_test.clj
```diff
@@ -89,6 +103,85 @@
...
+              (jruby-testutils/return-instance jruby-service instance :test))))))))
+
+(schema/defn puppet-tk-config
+  [code-dir :- File, conf-dir :- File]
+  (jruby-testutils/jruby-puppet-tk-config
+    (jruby-testutils/jruby-puppet-config
+      {:master-code-dir (.getAbsolutePath code-dir)
+       :master-conf-dir (.getAbsolutePath conf-dir)})))
+
+(def puppet-conf-file-contents
```
lindboe: looks like this was also intended to be reused in a refactor

test/integration/puppetlabs/services/jruby/tasks_test.clj
holguinj: That's covered in some higher level tests in the next commit

test/integration/puppetlabs/services/jruby/tasks_test.clj
holguinj: I updated the tests below to actually use this

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -341,11 +353,41 @@
     (->> (map format-task info-from-jruby)
          (middleware-utils/json-response 200))))
 
+(schema/defn ^:always-validate
+  describe-task :- TaskDetails
+  [task-data :- TaskData
+   env-name :- schema/Str
+   module-name :- schema/Str]
+  (let [?metadata (try (some-> task-data :metadata_file slurp json/decode)
```
lindboe: I don't have a strong preference either way, but since you only refer to metadata_file or files keys here, you could destructure that in the args list, especially since there's a schema explaining what you expect

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -341,11 +353,41 @@
...
+                   (catch Exception e
+                     nil))
+        describe-file (fn [file]
```
lindboe: I know in the classifier it was a pattern to make anonymous functions for fns that only get used in that one function, but it's not my own perference, as it makes functions longer. What would you think about breaking this out?

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -341,11 +353,41 @@
...
          (middleware-utils/json-response 200))))
 
+(schema/defn ^:always-validate
+  describe-task :- TaskDetails
+  [task-data :- TaskData
+   env-name :- schema/Str
+   module-name :- schema/Str]
+  (let [?metadata (try (some-> task-data :metadata_file slurp json/decode)
+                   (catch Exception e
+                     nil))
```
lindboe: Hrm, not returning anything about the exception seems less than ideal. Can we at least log it?

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -341,11 +353,41 @@
...
+                        (let [size (fs/size file)
+                              sha256 (ks/file->sha256 (io/file file))
+                              base-name (fs/base-name file)]
+                          {:filename base-name
+                           :sha256 sha256
+                           :size_bytes size
+                           :uri {:path (str "/puppet/v3/file_content/tasks/" module-name "/" base-name)
+                                 :params {:environment env-name}}}))]
+    {:metadata (or ?metadata {})
+     :files (mapv describe-file (:files task-data))}))
```
lindboe: why mapv?

holguinj: I tend to favor `mapv` when I know that I don't want laziness.

src/clj/puppetlabs/services/master/master_core.clj
holguinj: yeah, logging it is a good idea

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -436,6 +478,62 @@
...
+          module (get-in request [:route-params :module-name])
+          task (get-in request [:route-params :task-name])]
+      (let [?task-details (try (task-details jruby-service
+                                             (:jruby-instance request)
+                                             environment
+                                             module
+                                             task)
+                            (catch RaiseException e
+                              (cond
+                                (exception-matches? e #"(EnvironmentNotFound)")
```
lindboe: Why the parens? You shouldn't need a capture group for these, right?

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -436,6 +478,62 @@
...
+(schema/defn ^:always-validate
+  task-details-fn :- IFn
+  "Middleware function for constructing a Ring response from an incoming
+  request for detailed task information."
+  [jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)]
+  (fn [request]
+    (let [environment (jruby-request/get-environment-from-request request)
+          module (get-in request [:route-params :module-name])
+          task (get-in request [:route-params :task-name])]
+      (let [?task-details (try (task-details jruby-service
```
lindboe: Shouldn't need this nested let, but I can see an argument for breaking it out into its own function

src/clj/puppetlabs/services/master/master_core.clj
holguinj: those are the only two keys in the map anyway. destructure or no destructure, I don't see much of a difference either way I guess.

documentation/puppet-api/v3/task_detail.json
```diff
@@ -0,0 +1,62 @@
...
+                "description": "A relative URI for accessing the task contents. A client can form the full URI by requesting 'master:port/path?params'.",
+                "type": "string"
+              },
+              "params": {
+                "description": "A map of query params to use when requesting task contents. Required to include in the URI when making a request for task contents.",
+                "type": "object",
+                "properties": {
+                  "environment": {
+                    "description": "The environment the task is in, a required query param on request to task contents.",
+                    "type": "string"}
```
haus: The trailing `}` here seems off stylistically from the rest of the doc

documentation/puppet-api/v3/task_detail.markdown
```diff
@@ -0,0 +1,196 @@
...
+
+[`auth.conf`]: ../../config_file_auth.markdown
+[`puppetserver.conf`]: ../../config_file_puppetserver.markdown
+
+The tasks API provides access to task information stored in modules. Tasks are
+files stored in `tasks` subdirectory of a module. A task consists of an
+executable file, with an optional metadata file with the same name with an
+added '.json' extension. For example, the "install" task in a module "apache" could
+consist of the executable file `install.rb` and the metadata file
+`install.json`. This task would have the display name "apache::install".
```
haus: Was there discussion on how this could be confusing if there was also a manifest called `install` (install.pp) in the apache module?

src/clj/puppetlabs/services/master/master_core.clj
haus: I've never used `?` to precede a varname before. Is that being used to indicate that it might be nil?

src/clj/puppetlabs/services/master/master_core.clj
haus: something like `generate-task-metadata` or something? That would also be nicely unit testable.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -335,11 +353,41 @@
...
+        describe-file (fn [file]
+                        (let [size (fs/size file)
+                              sha256 (ks/file->sha256 (io/file file))
+                              base-name (fs/base-name file)]
+                          {:filename base-name
+                           :sha256 sha256
+                           :size_bytes size
+                           :uri {:path (str "/puppet/v3/file_content/tasks/" module-name "/" base-name)
+                                 :params {:environment env-name}}}))]
+    {:metadata (or ?metadata {})
```
haus: Seems like you could do the `{}` defaulting in your catch above instead of here and not need the `or` here.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -430,6 +478,62 @@
                              jruby-service)
         (environment-not-found environment)))))
 
+(defn exception-matches?
```
haus: Probably would be good to wrap this in a schema.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -430,6 +478,62 @@
...
+  [^Exception e pattern]
+  (->> e
+       .getMessage
+       (re-find pattern)
+       boolean))
+
+(schema/defn ^:always-validate
+  task-details :- TaskDetails
```
haus: maybe a short docstring would be appropriate here?

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -430,6 +478,62 @@
...
+                                    task-name)
+      sort-nested-info-maps
+      (describe-task environment-name module-name)))
+
+(schema/defn ^:always-validate
+  task-details-fn :- IFn
+  "Middleware function for constructing a Ring response from an incoming
+  request for detailed task information."
+  [jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)]
+  (fn [request]
```
haus: Not sure if destructuring would help at all here. `{:keys [route-params jruby-instance] :as request}` or something. Might not be worth it.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -430,6 +478,62 @@
...
+
+                                (exception-matches? e #"(MissingModule)")
+                                ::missing-module
+
+                                (exception-matches? e #"(TaskNotFound)")
+                                ::missing-task
+
+                                :else
+                                (throw e))))]
+        (case ?task-details
```
haus: Is there a reason not to inline these into the `cond` above and just have the `json-response` after the `task-details` call in your try?

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -591,18 +698,22 @@
...
-                                   environment-class-cache-enabled)
+                                   environment-class-cache-enabled),
         environment-module-handler
-        (environment-module-handler jruby-service)
+        (environment-module-handler jruby-service),
         all-tasks-handler
-        (all-tasks-handler jruby-service)
+        (all-tasks-handler jruby-service),
+        task-details-handler
+        (task-details-handler jruby-service),
```
lindboe: this is very much an opinion thing but I hate commas and prefer newlines if necessary, but I still find this readable without commas (and maybe the two-line situation isn't necessary?)

src/clj/puppetlabs/services/master/master_core.clj
holguinj: @haus yup, that's the way I've seen it used. it's not *that* common, but it is a thing.

src/clj/puppetlabs/services/master/master_core.clj
holguinj: I'll split it out

src/clj/puppetlabs/services/master/master_core.clj
holguinj: Yeah, that would be possible. I kind of like doing it here because it's a little easier to see that there's a default `{}` without all the clutter from the try/catch stuff.

src/clj/puppetlabs/services/master/master_core.clj
haus: i also find the commas off-putting

haus:
I made it through the code changes and commented on anything that looked off. I didn't see anything scary, so I'm good there. I didn't get to the tests at all.

src/clj/puppetlabs/services/master/master_core.clj
haus: Yea I'm not sure which is clearer, so I'm fine either way.

src/clj/puppetlabs/services/master/master_core.clj
holguinj: yeah, I'm not sure it's worth it

src/clj/puppetlabs/services/master/master_core.clj
holguinj: murdering the commas, replacing with newlines

holguinj:
updated

documentation/puppet-api/v3/task_detail.markdown
lindboe: It's an example, I just used "apache" because that seems to be a commonly-used example module. I don't know the name of any planned tasks/modules.

documentation/puppet-api/v3/task_detail.json
```diff
@@ -0,0 +1,63 @@
+{
+  "$schema":     "http://json-schema.org/draft-04/schema#",
+  "title":       "Task detail",
```
pcarlisle: is it normal to align these

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -25,7 +26,10 @@
             [puppetlabs.metrics.http :as http-metrics]
             [puppetlabs.http.client.metrics :as http-client-metrics]
             [puppetlabs.http.client.common :as http-client-common]
-            [puppetlabs.trapperkeeper.services.status.status-core :as status-core]))
+            [puppetlabs.trapperkeeper.services.status.status-core :as status-core]
+            [cheshire.core :as json]
```
pcarlisle: already imported above as cheshire

documentation/puppet-api/v3/task_detail.json
lindboe: I have no idea, I mainly just copied what already existed

test/integration/puppetlabs/services/jruby/tasks_test.clj
```diff
@@ -203,3 +203,80 @@
...
+                     {:name "about"
+                      :module-name "apache"
+                      :metadata? true
+                      :number-of-files 0}]
+              get-task-details (fn [env module task]
+                                 (mc/task-details jruby-service jruby-puppet env module task))]
+
+          (try (create-env (env-dir code-dir "production") tasks)
+
+              (testing "when the environment exists"
```
pcarlisle: minor indentation issue

src/clj/puppetlabs/services/master/master_core.clj
pcarlisle: yes, the naming in this file is confusing, having both describe-task which returns TaskDetails and task-details which returns TaskDetails, I'd have no idea what does what from just looking at that

pcarlisle:
It seems like the ruby code (in Puppet::Module::Task and the associated information service) considers the name of a task to be the fully qualified name, so `apache::install` or for the init task simply `apache`, but the API here treats the name as simply `install` or `init` and you look it up inside a module. You're even translating between the two explicitly.

holguinj:
Updated. To your last point, the ruby interface is definitely weird (the module gets named twice, once on its own and once as part of the task name), so I tried to just limit the damage from that to one place so that the rest of the code would look more reasonable.

lindboe:
Yeah, the overall plan (outside of our team) has gone back and forth on whether "init" was going to be special-cased. Also, the orchestrator mock task implementation expects this interface; I'd rather bring the Puppet implementation in-line with this one, if anything, so that we're explicit about what is being requested at lower levels instead of special-casing one task (a UI convenience).

Magisus:
Resolved the gettext problem that was causing Travis to fail, this should be good to go 👍

PR #1504, opened 2017-09-01
(PUP-7881) Update gettext-setup to version 0.28

Magisus:
This update contains code that makes the initial state of the locale
English rather than a random locale selected based on the order in which
translation files were found.


Magisus:
This version of the gem is not compatible with Ruby 1.9.3

Magisus:
Updated to bump to version 0.28, which restores Ruby 1.9.3 compatibility.

PR #1508, opened 2017-09-12
(maint) Add name to task details response.

aperiodic:
Add a 'name' field to the response from the task details endpoint, whose
value is the task's fully-qualified name (which will be just the module
name for the init task).

### DIFF
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -154,6 +154,7 @@
 (def TaskDetails
   "A filled-in map of information about a task."
   {:metadata {schema/Str schema/Any}
+   :name schema/Str
    :files [{:filename schema/Str
             :sha256 schema/Str
             :size_bytes schema/Int
@@ -377,6 +378,15 @@
      :size_bytes size
      :uri uri}))
 
+(defn full-task-name
+  "Construct a full task name from the two components. If the task's short name
+  is 'init', then the second component is omitted so the task name is just the
+  module's name."
+  [module-name task-shortname]
+  (if (= task-shortname "init")
+    module-name
+    (str module-name "::" task-shortname)))
+
 (schema/defn ^:always-validate
   task-data->task-details :- TaskDetails
   "Fills in a bare TaskData map by examining the files it refers to,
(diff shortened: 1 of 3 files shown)


src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -376,8 +386,9 @@
                     (catch Exception e
                       (log/warn e
                                 (i18n/tru "An exception occurred while reading the metadata file for the ''{0}'' task"
-                                          (str module-name "::" task-name)))))]
+                                          (str module-name "::" task-name )))))]
```
adrienthebo: Why wouldn't this error message also use the `full-task-name` function?

aperiodic: The `full-task-name` function will omit the task part of the name for the init task, so `exec::init` will show up as `exec`. To aid in debugging, we felt it would be best to always have both components.

MosesMendoza:
@aperiodic looks like a conflict - can you rebase?

test/integration/puppetlabs/services/master/tasks_int_test.clj
```diff
@@ -105,13 +106,22 @@
...
                                              "size_bytes" 20
                                              "uri" {"path" "/puppet/v3/file_content/tasks/shell/poc.sh"
                                                     "params" {"environment" "production"}}}]}]
             (is (= expected-response (parse-response response)))))
 
+        (testing "the name of the init task is just the module's name"
+          (testutils/write-tasks-files "exec" "init" "eval \"$@\"" {"description" "it is"})
+          (is (= "exec"
+                 (-> (get-task-details "production" "exec::init")
+                     (doto println)
```
MosesMendoza: Out of curiosity, what is the purpose of this line?

adrienthebo:
Does this still need to get merged?

test/integration/puppetlabs/services/master/tasks_int_test.clj
aperiodic: Oops - left in from debugging

aperiodic:
@adrienthebo yes, I'll rebase it today.

MikaelSmith:
This seems like we might target it at 5.1.x (but not merge until 5.1.2 goes out).

justinstoller:
Is this still desired? Are we just waiting on a rebase?

MikaelSmith:
We can probably just target it at master.

PR #1511, opened 2017-09-14
Revert "(RE-8994) Update to ezbake 1.6.0"

underscorgan:
Reverts puppetlabs/puppetserver#1509


underscorgan:
Need to trace down a sles12 bug

PR #1512, opened 2017-09-15
(SERVER-1972) Support Debian 9 in test setup

MikaelSmith:
Expect PuppetDB packages on Debian 9 (stretch).

Not quite sure where this needs to be targeted. Also relies on unreleased PuppetDB packages.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -71,6 +71,7 @@ module PuppetServerExtensions
     [
       /debian-7/,
       /debian-8/,
+      /debian-9/,
       /el/, # includes cent6,7 and redhat6,7
       /ubuntu-12/,
       /ubuntu-14/,


MikaelSmith:
Depends on a PuppetDB nightly supporting Debian 9 I'd imagine.

justinstoller:
So I see we've added Debian 9 support to Server 5.1.x, I assume that would be the appropriate branch to target this? This PR says that it depends on PuppetDB nightlies supporting Debian 9, is that true, can we merge this regardless, but not update our matrices to support testing with a db on Debian 9 until it has been released? Is there a place to find out when that's happened? Has it already happened?

MikaelSmith:
This is about testing against Debian 9 PuppetDB. I don't believe that's ready yet, and we should wait for https://github.com/puppetlabs/ci-job-configs/pull/3464.

justinstoller:
When should this be merged?

MikaelSmith:
I think @underscorgan was still finishing that up.

underscorgan:
PuppetDB has deb9 packages now, so I think this should work?

PR #1516, opened 2017-09-22
(SERVER-1993) Add support for serving tasks via static-file-content

MikaelSmith:
Supports serving tasks via the static-file-content endpoint, and uses it in the tasks API when files are available via the endpoint.

Restructures tests to use a `with-running-server` fixture to simplify setup.

Blocked on a release with https://github.com/puppetlabs/clj-parent/pull/84.

### DIFF
--- a/dev-resources/puppetlabs/services/master/tasks_int_test/echo
+++ b/dev-resources/puppetlabs/services/master/tasks_int_test/echo
@@ -1,0 +1,3 @@
+#!/usr/bin/env sh
+
+echo $@
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -158,7 +158,8 @@
             :sha256 schema/Str
             :size_bytes schema/Int
             :uri {:path schema/Str
-                  :params {:environment schema/Str}}}]})
+                  :params {:environment schema/Str
+                           (schema/optional-key :code_id) schema/Str}}}]})
 
 (defn obj-or-ruby-symbol-as-string
   "If the supplied object is of type RubySymbol, returns a string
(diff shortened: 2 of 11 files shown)


MikaelSmith:
Something about this doesn't work right yet. `task-data->task-details` is erroring in an unexpected way.

MikaelSmith:
Looks like file sync uses a different type of exception that I'll need to catch.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -355,22 +356,33 @@
...
-        base-name (fs/base-name file)]
+        base-name (fs/base-name file)
+        ;; we trust the file path from Puppet, so extract the subpath from file
+        static-path (re-find (re-pattern (str #"[^/]+/" module-name "/tasks/" base-name)) file)
+        uri (try
+              ;; if code content can be retrieved, then use static_file_content
+              (get-code-content env-name code-id static-path)
+              {:path (str "/puppet/v3/static_file_content/" static-path)
+               :params {:environment env-name :code_id code-id}}
+              (catch IllegalStateException e
```
MikaelSmith: Probably need to catch any exception here. And log the failure.

MikaelSmith:
Fixed-up. Use in practice results in
```
2017-09-22 17:16:03.412448 DEBUG leatherman.curl:344 - requesting https://10-32-167-127.rfc1918.puppetlabs.net:8140/puppet/v3/static_file_content/site/service/tasks/init?code_id=urn%3Apuppet%3Acode-id%3A1%3Ad0e790d1a5b45ee7ab947b96b4df6d4026e19428%3Bproduction&environment=production&.
2017-09-22 17:16:03.523225 DEBUG leatherman.curl:291 - request completed (status 200).
```

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -355,22 +356,34 @@
...
+  [get-code-content env-name code-id module-name file]
   (let [size (fs/size file)
         sha256 (ks/file->sha256 (io/file file))
-        base-name (fs/base-name file)]
+        base-name (fs/base-name file)
+        ;; we trust the file path from Puppet, so extract the subpath from file
+        static-path (re-find (re-pattern (str #"[^/]+/" module-name "/tasks/" base-name)) file)
+        uri (try
+              ;; if code content can be retrieved, then use static_file_content
+              (get-code-content env-name code-id static-path)
```
MikaelSmith: If basemodulepath includes a directory that's not managed that contains the same module and task with different content, we'll get the wrong file (and pxp-agent will error downloading it because the SHA doesn't match).

MikaelSmith:
Will need a bump of clj-kitchensink.

dev-resources/puppetlabs/services/master/tasks_int_test/echo
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env sh
+
+echo $@
```
MikaelSmith: Also note to myself, need to fix these up with a version that returns the correct file content (now that we're checking the hash). Integration test fails currently.

haus:
@MikaelSmith do you know how this will work in PE with file sync?

MikaelSmith:
Let's chat about it.

MikaelSmith:
This expects all masters to be somewhat consistent about where managed code is located. This is somewhat consistent with Puppet's static catalog, where the environmentpath is expected to be consistent across masters.

Presumably if the module is managed on one master, file sync will result in it being managed on other masters. I would hope that guarantees the relative path within the environment will be the same. If basemodulepath pre-empts the module on the MoM, then this endpoint will fall back to serving that one via `file_content`; for this to work correctly, the same basemodulepath would be needed on other masters to ensure the correct file is retrieved. If basemodulepath pre-empts the module on other masters but not the MoM, then this endpoint will use `static_file_content` and should successfully download the selected file.

MikaelSmith:
I think this is now safe to merge. pe-puppetserver will also need a clj-parent bump at https://github.com/puppetlabs/pe-puppetserver/pull/230.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -355,22 +356,35 @@
...
+        uri (try
+              ;; if code content can be retrieved, then use static_file_content
+              (when (not= sha256 (ks/stream->sha256 (get-code-content env-name code-id static-path)))
+                (throw (Exception. (i18n/trs "file did not match versioned file contents"))))
+              {:path (str "/puppet/v3/static_file_content/" static-path)
+               :params {:environment env-name :code_id code-id}}
+              (catch Exception e
+                (log/debug (i18n/trs "Static file unavailable for {0}: {1}" file e))
+                {:path (format "/puppet/v3/file_content/tasks/%s/%s" module-name base-name)
+                 :params {:environment env-name}}))]
```
MosesMendoza: just stylistically, do you feel like the uri assignment should be its own function? feels like we're doing a lot in here.

MikaelSmith: Sure, I could go for that.

MikaelSmith: Well, I end up having to pass everything I define in describe-task-file as an argument. So it doesn't feel simpler.

MikaelSmith: Going to leave it unchanged.

MosesMendoza:
@puppetlabs/puppetserver-maintainers want to take a look before i merge?

MikaelSmith:
I think this is clear to merge now.

MosesMendoza:
Will merge once we get server/master CI green

PR #1517, opened 2017-09-22
(DOCUMENT-729) Note static_catalogs setting behavior.

gguillotte:
Add `static_catalogs` to the puppet.conf setting differences doc to clarify when and how it's respected.


justinstoller:
I see that related PRs have been posted for older versions of Puppet, should this be targeted at older branches of server as well?

Magisus:
How long has this setting been a thing (and behaved this way)? If "forever", I'm with Justin, let's target at 2.x (or 5.1.x).

gguillotte:
I targeted at master before I knew that we were back to doing z-version branches, so I'm happy to retarget to other branches. For docs purposes, this change needs to be made to any version branches where it's relevant.

justinstoller:
I'd say retarget to the oldest branch that we need these changes in, and then we can merge those changes up to the latest release (I don't know about every intervening release). At least that's the usual protocol. This is a pretty trivial diff, so I'm not too concerned.

gguillotte:
ha ha NOPE, that ain't gonna work

PR #1518, opened 2017-09-25
(SERVER-1979) Update version to 5.1.1

Magisus:


adrienthebo:
Is this PR targeted at the right branch? It looks to be targeted at master but is based off of the 5.1.x branch on your fork.

Magisus:
Nope, it wasn't, good catch

PR #1521, opened 2017-09-25
(SERVER-1922) Set puppet log level from logback

adrienthebo:

must be set.

### DIFF
--- a/src/ruby/puppetserver-lib/puppet/server/logger.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/logger.rb
@@ -27,6 +27,11 @@ end
 class Puppet::Server::Logger
   def self.init_logging
     Puppet::Util::Log.newdestination(:logback)
+  end
+
+  # @note This must be called after Puppet settings are configured; otherwise
+  #   the default log level will overwrite this setting.
+  def self.set_log_level_from_logback
     Puppet[:log_level] = level_from_logback(get_logger)
   end
 
--- a/src/ruby/puppetserver-lib/puppet/server/logger.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/logger.rb
@@ -27,5 +27,25 @@ end
 class Puppet::Server::Logger
   def self.init_logging
     Puppet::Util::Log.newdestination(:logback)
+    Puppet[:log_level] = level_from_logback(get_logger)
+  end
+
+  def self.level_from_logback(logger)
+    case
+    when logger.isDebugEnabled()
+      'debug'
+    when logger.isInfoEnabled()
+      'info'
+    when logger.isWarnEnabled()
+      'warning'
+    when logger.isErrorEnabled()
+      'err'
+    else
+      'notice'
+    end
+  end
+
+  def self.get_logger
+    LoggerFactory.getLogger("puppetserver")
   end
 end
--- a/src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
@@ -1,4 +1,5 @@
 require 'puppet/server'
+require 'puppet/server/logger'
 
 class Puppet::Server::PuppetConfig
   def self.initialize_puppet(puppet_config)
@@ -20,6 +21,8 @@ class Puppet::Server::PuppetConfig
         end
     )
     Puppet[:trace] = true
+
+    Puppet::Server::Logger.set_log_level_from_logback
 
     # (SERVER-410) Cache features in puppetserver for performance.  Avoiding
     # the cache is intended for agents to reload features mid-catalog-run.
--- a/src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
@@ -29,9 +29,6 @@ class Puppet::Server::PuppetConfig
       Puppet[:always_retry_plugins] = false
     end
 
-    # Crank Puppet's log level all the way up and just control it via logback.
-    Puppet[:log_level] = "debug"
-
     master_run_mode = Puppet::Util::RunMode[:master]
     app_defaults = Puppet::Settings.app_defaults_for_run_mode(master_run_mode).
         merge({:name => "master",
--- a/spec/puppet-server-lib/puppet/jvm/config_spec.rb
+++ b/spec/puppet-server-lib/puppet/jvm/config_spec.rb
@@ -1,6 +1,27 @@
 require 'puppet/server/config'
 
 describe Puppet::Server::Config do
+  context "initializing Puppet Server" do
+    context "setting the Puppet log level from logback" do
+      let(:logger) do
+        stub("Logger", isDebugEnabled: false, isInfoEnabled: true, isWarnEnabled: false, isErrorEnabled: false)
+      end
+
+      before :each do
+        Puppet::Server::Logger.expects(:get_logger).returns(logger)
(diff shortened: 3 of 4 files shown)


src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
```diff
@@ -29,9 +29,6 @@ def self.initialize_puppet(puppet_config)
       Puppet[:always_retry_plugins] = false
     end
 
-    # Crank Puppet's log level all the way up and just control it via logback.
```
Magisus: Is there code to actually do this controlling that should be ripped out? Or was it sort of automatic?

src/ruby/puppetserver-lib/puppet/server/logger.rb
```diff
@@ -27,5 +27,25 @@ def handle(msg)
 class Puppet::Server::Logger
   def self.init_logging
     Puppet::Util::Log.newdestination(:logback)
+    Puppet[:log_level] = level_from_logback(get_logger)
+  end
+
+  def self.level_from_logback(logger)
+    case
+    when logger.isDebugEnabled()
+      'debug'
```
MosesMendoza: should these be symbols?

Magisus: It appears to have previously been a string (see the last file in this diff).

MosesMendoza: indeed

adrienthebo: Puppetserver expects that this option will be set on the commandline and thus will be a string; if you pass a symbol Puppet explodes.

src/ruby/puppetserver-lib/puppet/server/logger.rb
```diff
@@ -27,5 +27,25 @@ def handle(msg)
 class Puppet::Server::Logger
   def self.init_logging
     Puppet::Util::Log.newdestination(:logback)
+    Puppet[:log_level] = level_from_logback(get_logger)
```
MosesMendoza: If I'm not mistaken, I think the trouble with this approach is that you can only set the log level once, when we start up. Future changes to the logback loglevel won't be reflected by the puppet loglevel.

I think we want something that queries the logback level when a log message is generated, so you can change it any time and have it be reflected across server & puppet.

src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
adrienthebo: This function is called to initialize Puppet within Puppetserver, but there was an existing function elsewhere that [enabled Puppetserver specific logging](https://github.com/puppetlabs/puppetserver/blob/master/src/ruby/puppetserver-lib/puppet/server/config.rb#L21), I just used that existing hook for setting the log level.

Magisus: I meant code to "control it via logback". Like, was there special filtering logic or something that we no longer need? (I don't really know anything about logback)

src/ruby/puppetserver-lib/puppet/server/logger.rb
adrienthebo: Under what conditions will the logback configuration be re-read? That file can't be read every time that a log message is generated, can it? As it stands if you reload the JRuby pool then the [Puppetserver master class](https://github.com/puppetlabs/puppetserver/blob/master/src/ruby/puppetserver-lib/puppet/server/master.rb#L28) should be regenerated and reloaded.

src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
adrienthebo: The :logback log destination works by grabbing the appropriately named logger and [relaying log messages](https://github.com/puppetlabs/puppetserver/blob/master/src/ruby/puppetserver-lib/puppet/server/logger.rb#L13-L23) to that instance. That logger instance itself should be determining if a message logged via `#info` should be dropped or actually logged.

Magisus: Cool thanks

src/ruby/puppetserver-lib/puppet/server/logger.rb
bacongravy: Does it need to automatically pick up the setting change, or could we document a manual step for users to perform after changing the setting?  Catching a signal (either SIGHUP, or SIGUSR1) is a common pattern I've seen for providing a way to tell a running process to re-read its configuration.  That said, signal handling is super tricky and if Puppet Server doesn't already catch those signals properly I wouldn't suggest adding code to do so at this point in the release cycle.

MosesMendoza: Ah sure. I was under the impression you could dynamically update the log level in puppet server without having to do a server restart. In puppet you can just update the log level at any time and voila, new log level.

adrienthebo: That's a fair point that Puppet can internally reconfigure logging but it's a dastardly thing to do; I don't think that we need to handle cases where some extension causes Puppet to change its log level at runtime.

I did some checking and it looks like logging is only configured when a [tk app is built](https://github.com/puppetlabs/trapperkeeper/blob/master/src/puppetlabs/trapperkeeper/core.clj#L44), so I'm not entirely sure that the trapperkeeper reload logic will re-read the logback.xml. However the `initialize_puppet_server` method will be invoked each time a [JRuby instance is created](https://github.com/puppetlabs/puppetserver/blob/master/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj#L139), so I think that the Ruby logic will be called more frequently than the logback.xml file will be read.

justinstoller: TK will re-initialize logback during service reload. Logback has a "scan" feature that should check the file every (configurable) 60s and update its configuration if it's changed. We've had issues with logback's scan feature working reliably (see [TK-426](https://tickets.puppetlabs.com/browse/TK-426)).

adrienthebo: welp

An alternative approach is to move the log level checking into the logger itself in `#handle` if it's essential to pick up these changes, but if TK will reinitialize logback during service reload then this implementation will pick up the log level changes at that point.

adrienthebo: (I'm going to do some verification that reloading picks up the log level as expected.)

justinstoller: You're right, finch. I was wrote that reply without seeing yours. I'm not seeing where tk re-initializes logging after a reload. (I guess I thought that [TK-424](https://tickets.puppetlabs.com/browse/TK-424) had been done).

src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
Sharpie: As long as adding:

```
<logger name="puppetserver" level="DEBUG"/>
```

To `/etc/puppetlabs/puppetserver/logback.xml` and restarting the puppetserver service still works, then we should be good from a user workflow perspective.

MosesMendoza:
Seems reasonable. First we do `Puppet::Server::Config.initialize_puppet_server` which sets up the logback destination as its first task, then we call `Puppet::Server::PuppetConfig.initialize_puppet` in which we do all of the puppet setttings setup, and thus query logback for the log level and set puppet's log level correspondingly.

PR #1523, opened 2017-09-25
(docs) Re-add re-splay re-lease note.

gguillotte:
Restores the 5.1.0 release note about the re-splay/Retry-After header feature, now that it'll land in Puppet agent. This should merge before a Server 5.2.x release so we can pin the 5.1 Server docs to include this note.

### DIFF
--- a/documentation/release_notes.markdown
+++ b/documentation/release_notes.markdown
@@ -18,6 +18,17 @@ Released September 13, 2017
 
 This is a feature and bug-fix release of Puppet Server.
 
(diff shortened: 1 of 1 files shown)


Magisus:
Does it matter which branch this gets merged to for what you want to do? Master is currently slated to be the 5.2 release, so if this is something that needs to _ship_ before 5.2, this PR should target the 5.1.x branch.

gguillotte:
Will retarget, thanks @Magisus.

PR #1525, opened 2017-09-26
(maint) generalize Puppet version check test

adrienthebo:


spec/puppet-server-lib/puppet/jvm/master_spec.rb
```diff
@@ -24,7 +24,7 @@
     end
 
     it "returns the correct puppet version number" do
-      expect(subject).to eq('4.8.1')
```
MosesMendoza: seems fine as long as we dont take the same changes to master where its 5.x.x

PR #1526, opened 2017-09-26
(SERVER-1922) Set puppet log level from logback

adrienthebo:

must be set.


MosesMendoza:
cherry-pick PR of #1521

PR #1528, opened 2017-09-27
(maint) Update to ezbake 1.6.2

underscorgan:


underscorgan:
This pulls in some minor bug fixes

underscorgan:
@justinstoller I'll run a test but you may end up hitting the missing tarball bug, which would mean you'll want to pull this in before the release.

underscorgan:
Yeah, looks like you're hitting the lack of tarball bug (http://builds.delivery.puppetlabs.net/puppetserver/5.1.2.SNAPSHOT.2017.09.27T0916/artifacts/). This will cause issues while shipping and your source tarball wouldn't end up on downloads.puppetlabs.com so I'd recommend pulling this in before release. 

As a plus, this is a very minimal [changeset](https://github.com/puppetlabs/ezbake/compare/1.6.1...1.6.2) and the only change to the packaging is that the tarball will be preserved.

I ran a test against this branch and it looks a-okay. http://builds.delivery.puppetlabs.net/puppetserver/5.1.2.SNAPSHOT.2017.09.27T0922/artifacts/

PR #1530, opened 2017-09-27
(SERVER-1995) Fix puppet submodule updating

Magisus:
This PR updates the puppet submodule to point to the most recent passing puppet-agent#5.3.x SHA. 

It also updates the submodule bump rake task to accept environment variables for branch configuration. It also updates the hard-coded fallbacks so that the Jenkins job doesn't immediately undo the bump in the previous commit. I will open a follow-up PR updating the Jenkins job to supply these env vars.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -13,8 +13,10 @@ PUPPET_SPEC = File.join(PROJECT_ROOT, 'ruby', 'puppet', 'spec')
 FACTER_LIB = File.join(PROJECT_ROOT, 'ruby', 'facter', 'lib')
 PUPPET_SERVER_RUBY_SRC = File.join(PROJECT_ROOT, 'src', 'ruby', 'puppetserver-lib')
 PUPPET_SUBMODULE_PATH = File.join('ruby','puppet')
-CURRENT_BRANCH = 'master'
-JENKINS_BRANCH = CURRENT_BRANCH
+# Branch of puppetserver for which to update submodule pins
+PUPPETSERVER_BRANCH = ENV['PUPPETSERVER_BRANCH'] || '5.1.x'
+# Branch of puppet-agent to track for passing puppet SHA
+PUPPET_AGENT_BRANCH = ENV['PUPPET_AGENT_BRANCH'] || '5.3.x'
 
 TEST_GEMS_DIR = File.join(PROJECT_ROOT, 'vendor', 'test_gems')
 TEST_BUNDLE_DIR = File.join(PROJECT_ROOT, 'vendor', 'test_bundle')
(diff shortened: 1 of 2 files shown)


acceptance/config/beaker/options.rb
```diff
@@ -11,5 +11,5 @@
  "puppetserver-confdir"=>"/etc/puppetlabs/puppetserver/conf.d",
  "puppetserver-config"=>
   "/etc/puppetlabs/puppetserver/conf.d/puppetserver.conf",
- :puppet_version=>"5.2.0.33.g571b94f",
- :puppet_build_version=>"b9b9d3f893abf6d60aa20558816341d6794351cf"}
+ :puppet_version=>"5.2.0.68.g384311c"
```
adrienthebo: This is missing a trailing comma. which is causing beaker to fail:

```
bundler: failed to load command: beaker
SyntaxError: (eval):15: syntax error, unexpected tSYMBEG, expecting '}'
 :puppet_build_version=>"b1172c7f8c93541c39f060e114062ad16227201a"}
  ^
```

Magisus: Whoops, updated.

adrienthebo:
The acceptance test failure that I just fixed on 4.10.x might be broken on 5.1.x; I just got this failure when smoke testing this change:

```
Failed Tests Cases:
      Test Case ruby/puppet/acceptance/tests/loader/autoload_from_resource_type_decl.rb reported: #<Minitest::Assertion: Expected execution log file to be empty on agent node gnh5nw2rpelt3po.delivery.puppetlabs.net.
    Expected "\nfound_type_tst: 2017-09-27 22:19:26 +0000\nfound_type_tst: 2017-09-27 22:19:26 +0000" to be empty.>
        Test line: ruby/puppet/acceptance/tests/loader/autoload_from_resource_type_decl.rb:118:in `block (5 levels) in run_test'
```

I'm going to look into whether we need to cherry pick that fix on 5.1.x next.

adrienthebo:
Cherry picking 1442d9cc546c35f283e60a66e3719d40a2552add onto the puppet submodule resolves the CI failure.

Magisus:
👍

adrienthebo:
See https://github.com/puppetlabs/puppet/pull/6242

Rakefile
```diff
@@ -13,8 +13,10 @@ PUPPET_SPEC = File.join(PROJECT_ROOT, 'ruby', 'puppet', 'spec')
 FACTER_LIB = File.join(PROJECT_ROOT, 'ruby', 'facter', 'lib')
 PUPPET_SERVER_RUBY_SRC = File.join(PROJECT_ROOT, 'src', 'ruby', 'puppetserver-lib')
 PUPPET_SUBMODULE_PATH = File.join('ruby','puppet')
-CURRENT_BRANCH = 'master'
-JENKINS_BRANCH = CURRENT_BRANCH
+# Branch of puppetserver for which to update submodule pins
+CURRENT_BRANCH = ENV['PUPPETSERVER_BRANCH'] || '5.1.x'
+# Branch of puppet-agent to track for passing puppet SHA
+JENKINS_BRANCH = ENV['PUPPET_AGENT_BRANCH'] || '5.3.x'
```
justinstoller: I think your environment names are much better names for the constants as well, can we go with those everywhere?

justinstoller:
Looks great. In fact I think the env names are better than the existing constant names and it would be good to replace the latter with the former, if it isn't a big deal, when we update this PR for the agent CI fix (If I followed the convo, we're going to wait and have this PR pull in https://github.com/puppetlabs/puppet/commit/d21c5a749a87cf4e4308880e5daf331ccf8fd49e, right?)

Magisus:
Yes that sounds right. Since that got merged, I'll update this now. Please note that this will probably fail the first time through, because the agent pipeline is having a problem with the Jenkins data we're querying. Should be fixed the next time their 5.3.x pipeline runs through.

PR #1532, opened 2017-09-29
(maint) Update puppet-agent pin to 1.10.8.31.g445a971

adrienthebo:
A defect in the Puppet agent acceptance tests was causing spurious
failures in CI; this commit updates the Puppet submodule commit and
Puppet agent pins to pick up the acceptance test fix and keep all of the
packaging in sync.


adrienthebo:
I've verified this by locally running beaker against this changeset, executing the failing test.

PR #1537, opened 2017-10-02
(docs)(SERVER-1982) Add 5.1.3 release notes.

gguillotte:


gguillotte:
With two approvals, I'm going to merge this since it sounds like packages are going live in the next few minutes. @adrienthebo I can address any review feedback in a follow-up PR.

PR #1538, opened 2017-10-04
(doc) Add guide for configuring Puppet Server as an intermediate CA

adrienthebo:


documentation/intermediate-ca.md
```diff
@@ -0,0 +1,191 @@
...
+
+When Puppet Server or PE is installed it will autogenerate a number of certificates. Regenerating all of these certificates can be a time consuming process, but this problem can be side stepped by generating an intermediate CA certificate that reuses the existing key pair that Puppet Server/PE has already created.
+
+### Configure the Puppet agent `certificate_revocation` checking
+
+Intermediate CA support is incompatible with full chain revocation checking, as Puppet is unable to load multiple CRLs. When an intermediate CA certificate is in use Puppet must have the `certificate_revocation` setting either set to `false`, or `leaf`. By default Puppet will attempt to perform full chain checking and will fail when encountering a non-root CA certificate, so this setting **must** be reconfigured on all Puppet agents before continuing.
+
+    puppet config set --section main certificate_revocation leaf
+
+See [PUP-3788](https://tickets.puppetlabs.com/browse/PUP-3788) for more information on chain revocation checking.
```
MosesMendoza: I wonder if we should link to a docs site page or something rather than a bug..

documentation/intermediate-ca.md
```diff
@@ -0,0 +1,191 @@
...
+
+### Generate an intermediate CA certificate request
+
```
MosesMendoza: I wonder if we should add a quick blurb of what to do with this ruby.

I.e.,
```
Save the following ruby script to a file, csrgen.rb. Then run `ruby csrgen.rb`.
```

documentation/intermediate-ca.md
```diff
@@ -0,0 +1,191 @@
...
+    end
+    Puppet::Application::GenCACSR.new.run
+
+Use this CSR to generate the intermediate CA certificate that Puppet will be using.
+
+### Construct the CA certificate chain
+
+Puppet Server needs to present the full certificate chain to clients in order for the client to authenticate the server. The certificate chain is constructed by concatenating the CA certificates together, starting with the new intermediate CA certificate and descending to the root CA certificate.
+
+    cat puppet-intermediate-certificate.pem \
```
MosesMendoza: I can get it from context, but we should probably include where each of these files came from. Ie,
```
In this example, puppet-intermediate-certificate.pem is the output of the previous ruby script. You can find it at <default location on disk>.
org-intermediate-certificate.pem and org-root-certificate.pem are certificates in use by your organization.
```

documentation/intermediate-ca.md
```diff
@@ -0,0 +1,191 @@
...
+        org-intermediate-certificate.pem \
+         org-root-certificate.pem > ca-bundle.pem
+
+### Verify the Puppet master host certificate against the CA bundle
+
+Before installing the new certificate bundle, verify that the bundle can be used to verify the existing host certificate on the CA server. If this step fails then the CA certificate bundle is invalid and cannot be installed.
```
MosesMendoza: What would the user do now if the bundle was invalid?

documentation/intermediate-ca.md
adrienthebo: I'm amenable to that, but I'd prefer a documentarian to make a decision on that. Unfortunately as you'll see later in the document there are a lot of bugs/sharp edges here so I'm not sure how to approach documentation for so many issues.

documentation/intermediate-ca.md
```diff
@@ -0,0 +1,191 @@
...
+
+    cp ca-bundle.pem /etc/puppetlabs/puppet/ssl/ca/ca_crt.pem
+
+### Update the Puppet agent CA certificate bundle
+
+The Puppet agent caches the CA certificate and will not automatically fetch the new CA certificate bundle from the master. This caveat applies to the Puppet CA server itself as well. On the CA server, copy the CA certificate bundle to the local CA certificate path:
+
+    cp /etc/puppetlabs/puppet/ssl/ca/ca_crt.pem /etc/puppetlabs/puppet/ssl/certs/ca.pem
+
+If any Puppet agents have already been created, copy the CA certificate bundle to `/etc/puppetlabs/puppet/ssl/certs/ca.pem`.
```
MosesMendoza: Can you clarify that this file needs to be copied to this directory on every agent? I think they might think we mean locally on the master.

documentation/intermediate-ca.md
```diff
@@ -0,0 +1,191 @@
...
+
+Note that due to [PUP-6697](https://tickets.puppetlabs.com/browse/PUP-6697) Puppet agents are unable to download CA certificate bundles and will only save the first certificate in the bundle. Since the first certificate in the bundle will be the Puppet intermediate CA certificate, if the Puppet agent downloads and saves this single CA certificate it will not be able to complete verification of the Puppet Server certificate and will fail to connect.
+
+There may be other settings in the `webserver.conf` file; the other settings should be left alone.
+
+### Restart Puppet Server
+
+Once the new CA certificate is installed, restart pe-puppetserver.
+
+## Troubleshooting
```
MosesMendoza: This section is awesome! Thank you! Our users thank you.

documentation/intermediate-ca.md
adrienthebo: I like this system where you write good documentation and I just copy/paste it into the docs. Let's keep doing this!

Otherwise phrased, that's an excellent point and I'll add that.

documentation/intermediate-ca.md
```diff
@@ -0,0 +1,199 @@
...
+
+        csr.add_attribute(ext_req)
+        csr.sign(ca_key, OpenSSL::Digest::SHA256.new)
+
+        puts csr.to_s
+      end
+    end
+    Puppet::Application::GenCACSR.new.run
+
+Use this CSR to generate the intermediate CA certificate that Puppet will be using.
```
gguillotte: What is the expected result of running `csrgen.rb`? When I run it, there is no output and no files are created.

documentation/intermediate-ca.md
```diff
@@ -0,0 +1,199 @@
...
+
+The Puppet agent caches the CA certificate and will not automatically fetch the new CA certificate bundle from the master. This caveat applies to the Puppet CA server itself as well. On the CA server, copy the CA certificate bundle to the local CA certificate path:
+
+    cp /etc/puppetlabs/puppet/ssl/ca/ca_crt.pem /etc/puppetlabs/puppet/ssl/certs/ca.pem
+
+If any Puppet agents have already been created, copy the CA certificate bundle to `/etc/puppetlabs/puppet/ssl/certs/ca.pem` for every Puppet agent that was created before the new CA certificate was generated.
+
+Note that due to [PUP-6697](https://tickets.puppetlabs.com/browse/PUP-6697) Puppet agents are unable to download CA certificate bundles and will only save the first certificate in the bundle. Since the first certificate in the bundle will be the Puppet intermediate CA certificate, if the Puppet agent downloads and saves this single CA certificate it will not be able to complete verification of the Puppet Server certificate and will fail to connect.
+
+There may be other settings in the `webserver.conf` file; the other settings should be left alone.
```
gguillotte: This is the first mention of `webserver.conf` in this document. Is there some context that could help me understand why a user would need to be told to not do anything with that configuration file?

documentation/intermediate-ca.md
```diff
@@ -0,0 +1,199 @@
...
+
+### Verify the Puppet master host certificate against the CA bundle
+
+Before installing the new certificate bundle, verify that the bundle can be used to verify the existing host certificate on the CA server. If this step fails then the CA certificate bundle is invalid and cannot be installed.
+
+    openssl verify -CAfile ca-bundle.pem $(puppet master --configprint hostcert)
+
+### Update the Puppet CA certificate bundle
+
+    cp ca-bundle.pem /etc/puppetlabs/puppet/ssl/ca/ca_crt.pem
```
gguillotte: Would it be prudent to tell users to back up the existing CA certificate before doing this?

documentation/intermediate-ca.md
```diff
@@ -0,0 +1,199 @@
...
+
+If any Puppet agents have already been created, copy the CA certificate bundle to `/etc/puppetlabs/puppet/ssl/certs/ca.pem` for every Puppet agent that was created before the new CA certificate was generated.
+
+Note that due to [PUP-6697](https://tickets.puppetlabs.com/browse/PUP-6697) Puppet agents are unable to download CA certificate bundles and will only save the first certificate in the bundle. Since the first certificate in the bundle will be the Puppet intermediate CA certificate, if the Puppet agent downloads and saves this single CA certificate it will not be able to complete verification of the Puppet Server certificate and will fail to connect.
+
+There may be other settings in the `webserver.conf` file; the other settings should be left alone.
+
+### Restart Puppet Server
+
+Once the new CA certificate is installed, restart pe-puppetserver.
```
gguillotte: Should this note puppetserver in addition to pe-puppetserver?

Is a full puppetserver service restart required?

documentation/intermediate-ca.md
```diff
@@ -0,0 +1,199 @@
...
+    On the master:
+      puppet cert clean pe-20173nightly-master.puppetdebug.vlan
+    On the agent:
+      1a. On most platforms: find /etc/puppetlabs/puppet/ssl -name pe-20173nightly-master.puppetdebug.vlan.pem -delete
+      1b. On Windows: del "\etc\puppetlabs\puppet\ssl\certs\pe-20173nightly-master.puppetdebug.vlan.pem" /f
+      2. puppet agent -t
+
+##### Cause
+
+The `puppet cert generate` command should use the first certificate in the `cacert` file, but unintentionally uses the `localcacert` file. If the first certificate in this file is not the Puppet intermediate CA certificate Puppet will load the CA key and a mismatched CA certificate, and will fail with a misleading error message. The logic for verifying that the key and certificate match is shared with the agent key/certificate verification and will report that the agent key and certificate are mismatched when the CA certificate and key are mismatched.
```
gguillotte: > The logic for verifying that the key and certificate match is shared with the agent key/certificate verification and will report that the agent key and certificate are mismatched when the CA certificate and key are mismatched.

Can you clarify this? Specifically:

> The logic for verifying that the key and certificate match is shared with the agent key/certificate verification

What is sharing the logic, and to what is that logic being shared?

> and will report that the agent key and certificate are mismatched when the CA certificate and key are mismatched.

I feel like I'm missing some nuance here. It reports the key and certificate are mismatched, when the key and certificate are mismatched?

In other words, what is the difference between:

> the agent key and certificate are mismatched

and

> the CA certificate and key are mismatched

documentation/intermediate-ca.md
```diff
@@ -0,0 +1,199 @@
...
+
+### Update the Puppet agent CA certificate bundle
+
+The Puppet agent caches the CA certificate and will not automatically fetch the new CA certificate bundle from the master. This caveat applies to the Puppet CA server itself as well. On the CA server, copy the CA certificate bundle to the local CA certificate path:
+
+    cp /etc/puppetlabs/puppet/ssl/ca/ca_crt.pem /etc/puppetlabs/puppet/ssl/certs/ca.pem
+
+If any Puppet agents have already been created, copy the CA certificate bundle to `/etc/puppetlabs/puppet/ssl/certs/ca.pem` for every Puppet agent that was created before the new CA certificate was generated.
+
+Note that due to [PUP-6697](https://tickets.puppetlabs.com/browse/PUP-6697) Puppet agents are unable to download CA certificate bundles and will only save the first certificate in the bundle. Since the first certificate in the bundle will be the Puppet intermediate CA certificate, if the Puppet agent downloads and saves this single CA certificate it will not be able to complete verification of the Puppet Server certificate and will fail to connect.
```
ericldelaney: If I understand this correctly we need to copy the ca.pam file over because of this defect. Correct?

documentation/intermediate-ca.md
ericldelaney: Yes this is awesome!

Could we add prompts that tell the user where things are being run. [root@master]# , [root@agent]# 

The new administrator or someone who doesn't know PKI, but is required by their business to support it will thank you.

documentation/intermediate_ca_configuration.markdown
```diff
@@ -0,0 +1,286 @@
...
+Error: Could not request certificate: SSL_connect returned=1 errno=0 state=error: certificate verify failed: [unable to get issuer certificate for /CN=Puppet Enterprise CA generated on <CA server fqdn> at +2017-10-03 00:13:16 +0000]
+```
+
+#### Cause
+
+If the Puppet agent does not have a copy of the CA certificate bundle, when it downloads the CA cert from the Puppet Server, it receives only the first certificate in the bundle. In an intermediate CA certificate configuration, that certificate will be the Puppet intermediate CA certificate, and without the rest of the certificate chain Puppet will not be able to validate the certificate.
+
+#### Solution
+
+Copy the full CA certificate bundle from the Puppet Server to `/etc/puppetlabs/puppet/ssl/certs/ca.pem` on the agent.
```
ericldelaney: Need the linux versus windows paths

documentation/intermediate_ca_configuration.markdown
```diff
@@ -0,0 +1,286 @@
...
+This error message looks similar to the error produced by an incorrect `certificate_revocation` setting. The difference between the two cases is that if `certificate_revocation` is incorrectly set, the subject/CN will be the subject of the CA certificate, while a CRL with the wrong issuer will generate an error mentioning the hostname of the Puppet Master.
+
+#### Cause
+
+Puppet does not verify that a CRL was issued by a specific CA, so it's possible to distribute and update a CRL whose issuer is not the Puppet CA. In addition, the `puppet cert revoke` and `puppet cert clean` commands will modify a CRL issued by a different CA and save the new CRL with the wrong issuer.
+
+Because the issuer of the CRL does not match the issuer of the Puppet Server host, Puppet is unable to determine that the loaded CRL is associated with the certificate. This mismatch can be verified by comparing the issuer of the host certificate and the downloaded CRL.
+
+```
+diff -q <(openssl x509 -in $(puppet agent --configprint hostcert) -noout -issuer) <(openssl crl -in $(puppet agent --configprint hostcrl) -noout -issuer)
```
ericldelaney: windows version

documentation/intermediate-ca.md
```diff
@@ -0,0 +1,199 @@
+# Intermediate CA support with Puppet Server
+
+## Configuration guide
+
+When Puppet Server or PE is installed it will autogenerate a number of certificates. Regenerating all of these certificates can be a time consuming process, but this problem can be side stepped by generating an intermediate CA certificate that reuses the existing key pair that Puppet Server/PE has already created.
```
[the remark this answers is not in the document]
adrienthebo: There are a few reasons for regenerating a PKI. First, as you noted, a security breach is a big one. Another one is rotating the CA key pair, either due to the age of the key pair or to create a more secure key pair.

documentation/intermediate-ca.md
adrienthebo: @gguillotte fixed this one up.

documentation/intermediate-ca.md
adrienthebo: Wait, there's not output? @gguillotte if you're in the office you should come show me; a CSR should be printed to stdout. I can change this script to both print the CSR to stdout and save it to a file.

documentation/intermediate-ca.md
adrienthebo: This is tricky because something went wrong but there are many things that could go wrong, and you need to have a good understanding of TLS to figure this out. I don't know how to document this. Perhaps this?

> If this failed, just burn everything down and move into a cave in the woods because that's the only way that you can avoid the evil taint of bad crypto.

documentation/intermediate-ca.md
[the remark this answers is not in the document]
adrienthebo: In addition this guide is heavily focused on regenerating a CA cert for an existing PKI but some users might want to do that rekeying. Do you think that's worth documenting in a bit more detail?

documentation/intermediate-ca.md
adrienthebo: Oops, I wound up deleting the section where this was in because it wasn't needed but I left this line. I'll fix that. Good catch!

documentation/intermediate-ca.md
adrienthebo: Yes, see line 82.

documentation/intermediate_ca_configuration.markdown
adrienthebo: Replicating this functionality on Windows doesn't seem trivial - @Iristyle do you know how to do something like this on Windows?

adrienthebo:
@MosesMendoza @ericldelaney @gguillotte I'd like to get this merged in the next three days to tie up this loose end. I can do another pull request to add more information like Windows specific paths in a follow up pull request but unless there are no show stoppers I think we should merge this now.

gguillotte:
@adrienthebo At the top of my list today, sorry for the delays.

PR #1548, opened 2017-10-26
(maint) Downgrade psql version for integration (cherry pick from master)

justinstoller:
The 5.2.0 release of the puppetlabs-postgresql module introduced an
ordering bug that can cause installation to fail; this commit pins the
module to an older version to sidestep the issue.

### DIFF
--- a/acceptance/suites/tests/00_smoke/puppetdb_integration.rb
+++ b/acceptance/suites/tests/00_smoke/puppetdb_integration.rb
@@ -22,6 +22,10 @@ test_name 'PuppetDB integration'
 ## report was sent from Puppet Server to PuppetDB
 if !master.is_pe?
   step 'Install PuppetDB module' do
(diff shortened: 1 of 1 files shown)


ericldelaney:
LGTM

Magisus:
Is the bug supposed to get fixed at some point, so we can unpin this? If so, is there a ticket tracking that we can keep an eye on?

justinstoller:
This is tracked on our side at SERVER-2003, which I've linked to the original modules ticket that introduced this error, MODULES-5635. I also chimed in on the PR and it looks like the original author (it was a community contribution) is looking into fixing. I've gotten no traction from the modules team however. Do you think we should have a ticket on our side to track the revert??

Magisus:
Nah, I think that's okay. The modules ticket appears to still be "Open", I'll keep an eye on it and see what happens.

PR #1529, opened 2017-09-27
(RE-8457) Update to ship to puppet5 repos

melissa:

PR #1556, opened 2017-11-03
(maint) Merge 5.1.x into master

justinstoller:
This merge up discards updates to the project.clj (version number,
master is pointing to 5.2.0 now) and our puppet source code pins (our
master branch automatically uptakes code from puppet/master).

This does pull in documentation and git history for our 5.1.x releases.

 Conflicts:
	acceptance/config/beaker/options.rb
	project.clj
	ruby/puppet

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -71,6 +71,7 @@ module PuppetServerExtensions
     [
       /debian-7/,
       /debian-8/,
+      /debian-9/,
       /el/, # includes cent6,7 and redhat6,7
       /ubuntu-12/,
       /ubuntu-14/,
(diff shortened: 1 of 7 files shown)


Magisus:
So back to our conversation yesterday, honestly I still see value in keeping the 5.1.x branch separate, if mainly for consistency with other component of the Platform, which have a master branch for "future work" and a "current released version" branch. It makes it easier to communicate with other teams both inside and outside our group if we all follow the same branching strategy, and to me that outweighs the minor annoyance of managing two streams.

Magisus:
That has nothing to do with this PR, just that we shouldn't follow up by deleting the 5.1.x branch.

justinstoller:
> we shouldn't follow up by deleting the 5.1.x branch.

Though I want to, I wouldn't unilaterally do that (: I do intend to hopefully bring it up in a retro after the release. I definitely see value in:

> [making] it easier to communicate with other teams both inside and outside our group if we all follow the same branching strategy

Despite not loving 5.1.x and 5.1.4-release branches. I'd enjoy having a conversation about the general pros and cons offline, but I don't think I'll do them justice here.

gguillotte:
Without making a judgment on the value of the 5.1.x branch, please be aware that deleting the 5.1.x branch would cause open source docs builds to fail, because they're currently pinned to that branch. If anyone decides to delete that branch, it is critical that tech pubs be aware ***before*** 5.1.x is deleted so we can repin to a tag, commit, or different branch.

Magisus:
Thanks for the heads-up, @gguillotte !

justinstoller:
You know, we should hold off on this and merge https://github.com/puppetlabs/puppetserver/pull/1529 first.

Magisus:
Since #1508 is now targeted at master, is this ready?

justinstoller:
I think so

PR #1559, opened 2017-11-10
(SERVER-2022) Improve PuppetDB installation in acceptance tests

justinstoller:
This moves our pre-suite to use the puppetdb module installation method that was being used to smoke test puppetdb in `acceptance/suites/tests/00_smoke/puppetdb_integration.rb`. Now that test only does the actual querying of puppetdb.

It also reverts 32e8328, while RE is currently building packages for puppetdb we have not announced support for it, nor does the puppetdb module support installing from it.

It also makes a couple fixups to the acceptance test script to aid first time testers.


justinstoller:
@kris-bosland I can't find Jen's Github handle, will you let her know that I think she should review this as well?

craig-gomes:
I'm pretty sure this is Jen's github handle
https://github.com/8675309

Magisus:
Can confirm, that's Jen.

kris-bosland:
Just a few questions:
1. Not entirely familiar with the syntax GitHub is showing me - is 35_install_pdb.rb removed entirely?
2. Adding beaker-hostgenerator to testrun.sh seems like a big change - what happened?
3. Is there a process for determining when we should support debian-9?
Thanks!

jtappa:
@kris-bosland 
1. yep, that file has been completely removed, and 95_install_pdb.rb has been added as a new file.
2. can't help there, sorry
3. Usually I think that's under the platform OS team's umbrella.

Magisus:
beaker-hostgenerator was already being used elsewhere in that script, see the bottom of the diff for that file.

justinstoller:
Oh, @8675309, I didn't realize you went by Jenny!

Sharpie:
In the test I just ran it's pulling PuppetDB 4.1 down for use with Puppet Server 5.2-SNAPSHOT and Puppet 5.3.3, which triggers PUP-7605 and fails during catalog storage.

This may not have shown up prior to the re-factor as we were doing one puppet run to configure PuppetDB and then validating that run submitted a report. Since one run was both configuring and excercising PuppetDB, we weren't getting complete coverage as PuppetDB wasn't present during fact storage and catalog storage.

Sharpie:
Looks like the `PUPPETDB_BUILD_VERSION` defaults to 4.1.0 if not set otherwise:

https://github.com/puppetlabs/puppetserver/blob/5.1.4/acceptance/lib/helper.rb#L47

Should bump that up to 5.x for Puppet Server 5.x.

justinstoller:
I think this is ready for re-review @kris-bosland @8675309

kris-bosland:
I ran the beaker tests and they passed.  I think it is ready to merge.

Magisus:
Does this need to be merged at the same time as #6436 to avoid breakages? Or can this go in on its own?

justinstoller:
Opened https://github.com/puppetlabs/puppet/pull/6462 instead of 6436 to address feedback (and target the correct Puppet branch). The Puppet PR should be merged prior to this one.

Magisus:
The Puppet PR has been merged, once the submodule has been updated to point to that new version, this can be merged.

PR #1561, opened 2017-11-15
(RE-9730) Use artifactory instead of nexus

sarameisburger:
Previously would send releases and snapshots to nexus, we want to ship these to artifactory instead

PR #1562, opened 2017-11-15
(RE-9730) Use artifactory instead of nexus

sarameisburger:
Previously would send releases and snapshots to nexus, we want to ship these to artifactory instead


sarameisburger:
Tested this out and it should work but I think it uses an older version of ezbake so I wasn't able to stage the builds. Works for 5.1.x branch https://github.com/puppetlabs/puppetserver/pull/1561

PR #1572, opened 2017-11-21
(SERVER-2025) Record metrics for max-queued-requests

Sharpie:
This patch adds a new JRuby metric, `queue-limit-hit-meter`, that counts the
number of times the `max-queued-requests` limit was hit and provides
one minute, five minute and fifteen minute measurements on the rate at which
the hits are occurring. This metric can be used to monitor the amount of
traffic that Puppet Server is rejecting due to the `max-queued-request` limit.

The count and five minute rate are also added to the status output for the
JRuby service.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/jruby_request.clj
+++ b/src/clj/puppetlabs/puppetserver/jruby_request.clj
@@ -60,11 +60,13 @@
    max-queued-requests :- schema/Int
    max-retry-delay :- schema/Int]
   (let [metrics-map (jruby-metrics/get-metrics metrics-service)
-        {:keys [requested-instances]} metrics-map
+        {:keys [requested-instances queue-limit-hit-meter]} metrics-map
         err-msg (i18n/trs "The number of requests waiting for a JRuby instance has exceeded the limit allowed by the max-queued-requests setting.")]
     (fn [request]
       (if (>= (count @requested-instances) max-queued-requests)
         (let [response (output-error request {:msg err-msg} 503)]
+          ;; Record an occurance of the rate limit being hit to metrics.
+          (.mark queue-limit-hit-meter)
           (if (pos? max-retry-delay)
             (assoc-in response [:headers "Retry-After"]
                       (-> (rand) (* max-retry-delay) int str))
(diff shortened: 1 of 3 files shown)


Sharpie:
I implemented this using a Meter, but could go with a plain Counter. The advantage of the Meter is that it also tracks 1, 5, and 15 minute rates on how fast the count is increasing. This could be computed externally by recording the count to a metrics database, but having it available out-of-the box is nice.

Sharpie:
We already use several Timers, which combine a Histogram and a Meter, to measure stuff on the HTTP request path. Would be interesting to know exact performance numbers, but I expect the library keeps it as cheap as possible to enable "measure anything, measure everything".

src/clj/puppetlabs/puppetserver/jruby_request.clj
```diff
@@ -59,12 +59,14 @@
    metrics-service :- (schema/protocol jruby-metrics/JRubyMetricsService)
    max-queued-requests :- schema/Int
    max-retry-delay :- schema/Int]
-  (fn [request]
-    (let [metrics-map (jruby-metrics/get-metrics metrics-service)
-          requested-count (-> metrics-map :requested-instances deref count)]
-      (if (>= requested-count max-queued-requests)
-        (let [err-msg (i18n/trs "The number of requests waiting for a JRuby instance has exceeded the limit allowed by the max-queued-requests setting.")
-              response (output-error request {:msg err-msg} 503)]
+  (let [metrics-map (jruby-metrics/get-metrics metrics-service)
```
haus: Are the elements in the metrics-map mutable? If not, lifting them out of the function seems like they will be set only on startup and then never updated again.

haus: Looks like queue-limit-hit-meter is a metrics object and requested-instances is an atom.

haus: @Sharpie This looks like a safe change, but was there a reason for it?

haus: ah never mind. i can read commit messages i swear.

Sharpie: The metrics map comes from the service context of the JRubyMetricsService. Should be immutable with the possible exception of a new state generated during reloads. The original commit that added rate limiting, PR #1365, also added a dependency on JRubyMetricsService to the RequestHandlerService which calls this function during init to construct a ring handler. So, both the atom and meter should be immutable within the closure of the ring handler.

PR #1573, opened 2017-11-22
(SERVER-1974) Add additional compiler metrics to status output

Sharpie:
This patch adds some additional patterns to the MetricsPuppetProfiler class and extends the `catalog-metrics` arrary reported by `status/v1/services/puppet-profiler` to include:

  - `find_node`: Measures the time it took for the External Node Classifier to return node data during compilation.

A new `puppetdb-metrics` array is also present and will be populated if PuppetDB is used. Metrics included here are:

  - `query`: Time spent waiting for PuppetDB to return query results.
  - `resource_search`: Time spent waiting for PuppetDB to collect exported resources.
  - `facts_encode`: Time spent in the Ruby terminus transforming fact data for PuppetDB.
  - `command_submit_replace_facts`: Time spent waiting for PuppetDB to accept fact data.
  - `catalog_munge`: Time spent in the Ruby terminus transforming catalog data for PuppetDB.
  - `command_submit_replace_catalog`: Time spent waiting for PuppetDB to accept catalog data.
  - `report_convert_to_wire_format_hash `: Time spent in the Ruby terminus transforming report data for PuppetDB.
  - `command_submit_store_report`: Time spent waiting for PuppetDB to accept report data.

Having these measurements available allows metrics from the `status/` API to show whether the ENC or PuppetDB are bottlenecking catalog compilation.

### DIFF
--- a/src/clj/puppetlabs/services/puppet_profiler/puppet_profiler_core.clj
+++ b/src/clj/puppetlabs/services/puppet_profiler/puppet_profiler_core.clj
@@ -34,6 +34,10 @@
                        :count schema/Int
                        :mean schema/Num
                        :aggregate schema/Num}]
+    :puppetdb-metrics [{:metric schema/Str
+                        :count schema/Int
+                        :mean schema/Num
+                        :aggregate schema/Num}]
     :inline-metrics [{:metric schema/Str
                       :count schema/Int}]}})
 
@@ -72,6 +76,21 @@
      :mean mean
      :aggregate (* count mean)}))
 
+(defn puppetdb-metric
+  [[metric-name timer]]
+  (let [count (.getCount timer)
+        mean (metrics/mean-millis timer)
+        ;; Metric databases like Graphite and InfluxDB treat
+        ;; spaces and periods as delimiters.
+        ;;
+        ;; TODO: Should probably be abstracted into
+        ;; trapperkeeper-metrics as a general utility function.
+        munged-name (str/replace metric-name #"\.| " "_")]
+    {:metric munged-name
+     :count count
+     :mean mean
+     :aggregate (* count mean)}))
+
 (defn inline-metric
   [[metric-name timer]]
   (let [count (.getCount timer)]
@@ -91,6 +110,9 @@
           catalog-metrics (->> (map catalog-metric (.getCatalogTimers metrics-profiler))
                                (sort-by :aggregate)
                                reverse)
+          puppetdb-metrics (->> (map puppetdb-metric (.getPuppetDBTimers metrics-profiler))
+                                (sort-by :aggregate)
+                                reverse)
           inline-metrics (->> (map inline-metric (.getInliningTimers metrics-profiler))
                                (sort-by :count)
                                reverse)]
@@ -98,6 +120,7 @@
           (assoc-in [:experimental :function-metrics] (take 50 function-metrics))
           (assoc-in [:experimental :resource-metrics] (take 50 resource-metrics))
           (assoc-in [:experimental :catalog-metrics] catalog-metrics)
+          (assoc-in [:experimental :puppetdb-metrics] puppetdb-metrics)
           (assoc-in [:experimental :inline-metrics] inline-metrics)))
     status))
 
--- a/acceptance/suites/tests/00_smoke/puppetdb_integration.rb
+++ b/acceptance/suites/tests/00_smoke/puppetdb_integration.rb
@@ -10,6 +10,9 @@
 ## run made it into PuppetDB. Additionally, the STDOUT of the agent run is
 ## tested for the presence of a Notify resource that was exported by another
 ## node.
+##
+## Finally, the output of the Puppet Server HTTP /status API is tested
+## to ensure that metrics related to PuppetDB communication were recorded.
 #
 
 # We only run this test if we'll have puppetdb installed, which is gated in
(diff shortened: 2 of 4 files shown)


Sharpie:
Flagging as "DO NOT MERGE" until #1559 lands and fixes the PuppetDB testsuite up a bit.

Sharpie:
@npwalker Let me know if this metrics set looks good or if there are any additional Puppet metrics that would be useful to get into the status API.

npwalker:
@Sharpie combing through it seems like a good set of metrics to add.  Can you provide the output from the status endpoint with these changes in place?  I'd love to see what they look like coming out of there.  

Have you confirmed that adding these metrics to the status endpoint "just works" with the metrics collection and graphing?

Sharpie:
Here's the result of applying the patch to PE 2017.3.2, I've trimmed the resources and functions arrays for brevity:

```
[root@pe-201732-master splode]# curl -k https://localhost:8140/status/v1/services/puppet-profiler?level=debug|python -m json.tool


```

npwalker:
@Sharpie I'm not a huge fan of the array of hashes it seems more difficult to find what you're looking for.  I'd be more into something like the following 

```
 "puppetdb-metrics": {
    "command_submit_replace_catalog" : 
                {
                    "aggregate": 1060,
                    "count": 2,
                    "mean": 530,
                },
 " catalog_munge":
                {
                    "aggregate": 592,
                    "count": 2,
                    "mean": 296,
                 }, ...
```

So a hash of hashes where the key is the name of the metric.  This is more similar to how the jruby metrics are granted the way you did it is more like the http metrics and some others.  

I wonder if we shouldn't ever be presenting anything other than the mean like the jruby metrics?

I know that parsing the http metrics presented a small challenge to @reidmv.

Sharpie:
Switching from an array to a hash would probably be better, but I'd call it out-of-scope for this change since it fundamentally alters the shape of the data returned by the API whereas the current patch set just extends the data but keeps the same shape.

npwalker:
@Sharpie hmmmmm I can see that.  Just as long as you know how to parse it for the metrics I think that'll be fine.

justinstoller:
@Sharpie do you want to check this against the current work in #1559 ? The current state of that branch works for me but I want to make sure I didn't miss anything.

Sharpie:
@justinstoller Sure. I'll re-base this patchset onto that PuppetDB branch and see what happens :)

Sharpie:
@justinstoller This should be good to go once 5.1.x gets merged up to master.

Turned out that data submission to PuppetDB was de-activated due to puppet.conf being rolled back. I fixed this up in `puppetdb_integration.rb` and changed the assertions to ensure that fresh data is submitted during the test. Also expanded the checks from just reports to include facts, catalogs and exported resources. These changes could be extracted to a separate pull request if it makes sense.

justinstoller:
Ack, looks good on the whole. I thought that the `Enable PuppetDB` step shouldn't be needed but I should run it locally first. Will try to get this merged in this week in prep for the January/February foss releases.

Sharpie:
I rebased this back onto 5.1.x so that the diffs look sane. Was planning to kick it back over to master and pop the "DO NOT MERGE" label off once 5.1.x gets mereged up.

Alternately, could leave it on 5.1.x and pop the label now if it seems like a good fit for the 5.1.x FOSS releases.

Sharpie:
Moved this back to master and re-ran `00_smoke/puppetdb_integration.rb` locally. Things are still looking green.

src/clj/puppetlabs/services/puppet_profiler/puppet_profiler_core.clj
```diff
@@ -72,6 +76,21 @@
...
+(defn puppetdb-metric
+  [[metric-name timer]]
+  (let [count (.getCount timer)
+        mean (metrics/mean-millis timer)
+        ;; Metric databases like Graphite and InfluxDB treat
+        ;; spaces and periods as delimiters.
+        ;;
+        ;; TODO: Should probably be abstracted into
+        ;; trapperkeeper-metrics as a general utility function.
+        munged-name (str/replace metric-name #"\.| " "_")]
```
justinstoller: Catalog metrics are already outputting in this dot delineated way, right (ie `compiler.find_node`) (or are we munging those names in a different location that I'm missing)? Would this mean that similarly formatted metrics from the profiler will have different representations in the status endpoint? And if so, would it be better to have them all similarly formatted in the status endpoint (even if that means the PDB ones are less convenient) and shoot for adding a v2 that has a more convenient format for all the metrics?

Sharpie: Yeah, but we're already doing transformation there. I.E `catalog.find_node` is shortened to `find_node` by the regex captures in `MetricsPuppetProfiler.java`. The additional work done here by munging the PuppetDB metrics is to reconcile the fact that the way we've instrumented the compiler Ruby code with`Puppet::Util::Profiler.profile` resulted in a very flat namespace:

  - `compiler.find_node`
  - `compiler.compile`
  - `compiler. static_compile`
  - etc.

But PuppetDB has ended up with a deeply nested namespace combined with odd spacing:

  - `puppetdb.facts.encode`
  - `puppetdb.command.submit.replace facts`
  - `puppetdb.catalog.munge`
  - `puppetdb.command.submit.replace catalog`

There's already a lot of drift between how things are named in the `/status` API and the `/metrics` API. The munging above makes the formatting of names in the `puppetdb-metrics` section of the `/status` API consistent with other sections like `catalog-metrics` which I think is more important than trying to keep it consistent with `/metrics`.

PR #1575, opened 2017-11-29
(SERVER-2034) Add -master back into version string

haus:
Some of the automated performance tests search for the latest build with
the word -master in the version string. This broke when -master was
removed from the version string between 5.0.1 and 5.1.0. This commit
adds it back in to get those tests working again.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,4 @@
-(def ps-version "5.2.0-SNAPSHOT")
+(def ps-version "5.2.0-master-SNAPSHOT")
 (def jruby-1_7-version "1.7.27-1")
 (def jruby-9k-version "9.1.11.0-1")
 


haus:
An alternative approach would be to have the perf job parse project.clj from puppetserver's master branch, or to maintain some configuration setting of what the latest version is. I'm open to either of those options if they seem better.

project.clj
```diff
@@ -1,4 +1,4 @@
-(def ps-version "5.2.0-SNAPSHOT")
+(def ps-version "5.2.0-master-SNAPSHOT")
```
melissa: This pulls in the jar from a maven repo, right?

melissa: a la https://artifactory.delivery.puppetlabs.net/artifactory/maven/puppetlabs/puppetserver/ or http://nexus.delivery.puppetlabs.net/content/repositories/releases/puppetlabs/puppetserver/

haus: No, this defines the version strings used when deploying to maven.

melissa: oooh, my bad. My brain was in pe-puppet-server-extensions world.

PR #1578, opened 2017-12-08
(SERVER-2054) [#puppethack] Add support for PUT /puppet/v3/facts

Sharpie:
This patch adds a comidi handler to the routing tree of the MasterService that
responds to PUT requests for `/puppet/v3/facts`.

### DIFF
--- a/dev-resources/puppetlabs/services/master/master_service_test/confdir/puppet.conf
+++ b/dev-resources/puppetlabs/services/master/master_service_test/confdir/puppet.conf
@@ -1,6 +1,7 @@
 [master]
 certname = localhost
 node_terminus = exec
+facts_terminus = yaml
 external_nodes = "/bin/sh ./dev-resources/puppetlabs/services/master/master_service_test/enc.sh"
 reports = http
 
--- a/ezbake/config/conf.d/auth.conf
+++ b/ezbake/config/conf.d/auth.conf
@@ -117,6 +117,17 @@ authorization: {
             name: "puppetlabs report"
         },
         {
+            # Allow nodes to update their own facts
+            match-request: {
+                path: "^/puppet/v3/facts/([^/]+)$"
+                type: regex
+                method: put
+            }
+            allow: "$1"
+            sort-order: 500
+            name: "puppetlabs facts"
+        },
+        {
             match-request: {
                 path: "/puppet/v3/status"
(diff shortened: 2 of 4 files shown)


Sharpie:
Adding "DO NOT MERGE" as the test attached to this PR will fail until puppetlabs/puppet#6434 is merged and the `ruby/puppet` submodule is updated to bring that changeset in.

Sharpie:
Another thing that might be worth considering is whether we want a default auth rule that allows access to this endpoint. The original docs from the `puppet facts upload` face state that auth.conf has to be modified to allow the upload.

However, it would be pretty easy to add something to the default ruleset that allows nodes to upload their own facts:

```
puppet_authorization::rule {'node fact upload':
  ensure               => $ensure,
  match_request_path   => '^/puppet/v3/facts/([^/]+)$',
  match_request_type   => 'regex',
  match_request_method => 'put',
  allow                => '$1',
  sort_order           => 601,
  path                 => '/etc/puppetlabs/puppetserver/conf.d/auth.conf',
  notify               => $_puppetserver_service,
}
```

justinstoller:
Good point, Charlie. Will you update https://github.com/puppetlabs/puppetserver/blob/master/ezbake/config/conf.d/auth.conf with the equivalent of that Puppet code? We will also need to record somewhere that there will be a puppet_enterprise module change as well in the process of promoting this into PE.

Sharpie:
Rebased onto master, fixed a merge conflict, and updated ezbake auth.conf.

ericldelaney:
@justinstoller could you please review this.

This is the other half of the PUP-8232 https://github.com/puppetlabs/puppet/pull/6434 which probably wan't pass travis until PUP-8232 is merged and we rerun travis.

justinstoller:
fyi, we're blocking this PR on puppetlabs/puppet#6434 which is currently going back and forth between @Sharpie and @joshcooper . Once that is merged we'll be open to merging this. IIRC, we're hoping to have this in for the Platform 5.5 release which needs its code landed ~ end of this week.

justinstoller:
Now that https://github.com/puppetlabs/puppet/pull/6434 has been merged it should be brought into puppetserver as a new submodule after it has passed agent's CI. Once that happens we can re-kick travis and should see these tests go green!

PR #1581, opened 2017-12-14
(SERVER-1780) Test and document jruby profiler settings

haus:
JRuby profiling support was added to jruby-utils in 0.11.0, so this
commit adds some docs around the relevant config settings and also
augments one of the jruby tests to validate that the profiler settings
work as expected.
Also includes an update to the latest jruby 9k.

### DIFF
--- a/documentation/config_file_puppetserver.markdown
+++ b/documentation/config_file_puppetserver.markdown
@@ -78,6 +78,10 @@ The `puppetserver.conf` file contains settings for Puppet Server software. For a
 
     * `compile-mode`: Optional, experimental. Used to control JRuby's "CompileMode", which may improve performance. The default value is `off`, which is the most conservative value. A value of `jit` enables JRuby's "just-in-time" compilation of Ruby code. A value of `force` causes JRuby to attempt to pre-compile all Ruby code.
 
+    * `profiling-mode`: Optional. Used to enable JRuby's profiler and set it to one of the supported modes. The default value is `off`, but it can be set to one of `api`, `flat`, `graph`, `html`, `json`, `off`, and `service`. See [ruby-prof](https://github.com/ruby-prof/ruby-prof/blob/master/README.rdoc#reports) for details on what the various modes do.
+
+    * `profiling-output-file`: Optional. Used to set the output file to direct JRuby profiler output. Should be a fully qualified path writable by the service user. If not set will default to a random name inside the service working directory.
+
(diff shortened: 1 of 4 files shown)


haus:
The jruby-utils release is pending once jenkins responds with something other than a 504.

project.clj
```diff
@@ -1,6 +1,6 @@
 (def ps-version "5.2.0-master-SNAPSHOT")
 (def jruby-1_7-version "1.7.27-1")
-(def jruby-9k-version "9.1.11.0-1")
+(def jruby-9k-version "9.1.15.0-2")
```
lindboe: Was this also intended as a combo update to 9.1.15?

haus: pretty much. i can break it into 2 PRs if that's easier. the jruby-utils releases were a little intermixed.

documentation/config_file_puppetserver.markdown
```diff
@@ -78,6 +78,10 @@ The `puppetserver.conf` file contains settings for Puppet Server software. For a
 
     * `compile-mode`: Optional, experimental. Used to control JRuby's "CompileMode", which may improve performance. The default value is `off`, which is the most conservative value. A value of `jit` enables JRuby's "just-in-time" compilation of Ruby code. A value of `force` causes JRuby to attempt to pre-compile all Ruby code.
 
+    * `profiling-mode`: Optional. Used to enable JRuby's profiler and set it to one of the supported modes. The default value is `off`, but it can be set to one of `api`, `flat`, `graph`, `html`, `json`, `off`, and `service`. See [ruby-prof](https://github.com/ruby-prof/ruby-prof/blob/master/README.rdoc#reports) for details on what the various modes do.
+
+    * `profiling-output-file`: Optional. Used to set the output file to direct JRuby profiler output. Should be a fully qualified path writable by the service user. If not set will default to a random name inside the service working directory.
```
lindboe: My lazy question asking: do we just do nothing if you provide an output file but no mode

haus: we set profileroutput still but it is a noop

haus: noop as in nothing is written to the file

project.clj
lindboe: In principle it's probably better in 2 PRs but right now whatever

lindboe:
LGTM

PR #1586, opened 2018-01-02
(SERVER-2022) Don't run pdb tests if pdb isn't there

haus:
PDB installation is gated on a list of supported platforms, which
differs(ed?) from the platforms that puppetserver builds and tests for.
Previously the pdb smoke tests were unconditionally run, even if pdb
wouldn't be there. This commit limits them both to only the platforms
listed in the puppetdb_supported_platforms method.

### DIFF
--- a/acceptance/suites/pre_suite/foss/95_install_pdb.rb
+++ b/acceptance/suites/pre_suite/foss/95_install_pdb.rb
@@ -2,7 +2,6 @@
 matching_puppetdb_platform = puppetdb_supported_platforms.select { |r| r =~ master.platform }
 skip_test unless matching_puppetdb_platform.length > 0
 
-require 'json'
 
 test_name 'PuppetDB setup'
 sitepp = '/etc/puppetlabs/code/environments/production/manifests/site.pp'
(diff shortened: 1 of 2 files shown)


haus:
It looks like https://github.com/puppetlabs/puppetserver/blob/2bb41296497f1d962bebb8f83f220000f70b485f/acceptance/lib/helper.rb#L65-L79 is probably out of date and maybe not needed anymore? I didn't try adjusting that for this PR, I just wanted to get tests green.

lindboe:
Do we know why the supported platforms are different between puppet server and puppetdb? that seems odd.

acceptance/suites/pre_suite/foss/95_install_pdb.rb
```diff
@@ -2,7 +2,6 @@
 matching_puppetdb_platform = puppetdb_supported_platforms.select { |r| r =~ master.platform }
 skip_test unless matching_puppetdb_platform.length > 0
 
-require 'json'
```
lindboe: Why this change?

haus: oh i should have pulled it into a different commit, but it isn't used in that test.

haus:
@lindboe it may be from a time when platforms were more in flux? IIRC there is often a little bit of lag for new platforms, as first agent must be ready for the platform, then puppetserver, and then puppetdb. depending on priorities it can lag quite a bit.

lindboe:
Okay so let me know if I'm breaking this down right: The issue is that the pre-suite to install PE is only installing it or some platforms, but the test was being run for all. The former install step is independent of what platforms PDB may or may not actually support. Yeah?

lindboe:
LGTM

PR #1587, opened 2018-01-03
(SERVER-2064) Use /bin/sh for commands that use shell characters

haus:
Commands that use pipes and other shell characters are currently handled
incorrectly by puppetserver's Puppet::Server::Execution.execute method.
They are parsed by apache commons exec and treated as arguments to the
binary being invoked, which is almost certainly the wrong thing. In
order to allow commands with pipes and other shell characters to be
correctly invoked, this commit brings the execute method into line with
what MRI's Kernel.exec does by default. It checks the command string
against a list of known shell characters and if any are found, it
invokes the command with /bin/sh -c. The list of characters can be found
at https://github.com/ruby/ruby/blob/v2_4_2/process.c#L2166.

### DIFF
--- a/spec/puppet-server-lib/puppet/jvm/execution_spec.rb
+++ b/spec/puppet-server-lib/puppet/jvm/execution_spec.rb
@@ -56,6 +56,30 @@ describe Puppet::Server::Execution do
       expect(result.exitstatus).not_to eq 0
     end
 
+    context "it should support other shellisms" do
+      it "should support pipes" do
+        result = test_execute("echo foo | wc -c")
+        expect(result).to be_a Puppet::Util::Execution::ProcessOutput
+        expect(result.lstrip).to eq "4\n"
+        expect(result.exitstatus).to eq 0
+      end
+
+      it "should support environment variables" do
+        result = test_execute(%(FOO=bar python -c "import os; print os.environ['FOO']"))
+        expect(result).to be_a Puppet::Util::Execution::ProcessOutput
+        expect(result).to eq "bar\n"
+        expect(result.exitstatus).to eq 0
+      end
+
+      it "should support environment newlines" do
+        result = test_execute(%(echo "foo\nbar"))
+        expect(result).to be_a Puppet::Util::Execution::ProcessOutput
+        expect(result.lines.count).to eq 2
+        expect(result.exitstatus).to eq 0
+      end
+    end
+
+
     it "should raise an error if `failonfail` is true and the process returns non-zero" do
       expect {
         test_execute("false", {:failonfail => true})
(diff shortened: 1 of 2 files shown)


src/ruby/puppetserver-lib/puppet/server/execution.rb
```diff
@@ -16,6 +16,8 @@ def self.initialize_execution_stub
   end
 
   private
+  SHELL_CHARACTERS = "*?{}[]<>()~&|\\$;'`\"\n#=".chars.to_a
```
haus: The `=` here differs slightly from MRI ruby. MRI ruby does special handling to check for `=` *before* the binary (for environment variables), and only invokes the shell in that case. This means we will run using `sh` for any command with `=` before or after, while MRI will only invoke `sh` with `=` before the command.

justinstoller: Why `to_a`? What does `chars` return?

spec/puppet-server-lib/puppet/jvm/execution_spec.rb
```diff
@@ -56,6 +56,30 @@ def test_execute(command, options = {})
       expect(result.exitstatus).not_to eq 0
     end
 
+    context "it should support other shellisms" do
+      it "should support pipes" do
+        result = test_execute("echo foo | wc -c")
+        expect(result).to be_a Puppet::Util::Execution::ProcessOutput
+        expect(result.lstrip).to eq "4\n"
```
justinstoller: Left strip and then expecting a result with a trailing newline seems weird to me. Why not just strip both? Or not strip either side? Is the preceding whitespace variable but the newline both mandatory and important?

src/ruby/puppetserver-lib/puppet/server/execution.rb
haus: in 1.9 it returns an enumerator. in 2.x, an array :(

haus: want a comment about that?

spec/puppet-server-lib/puppet/jvm/execution_spec.rb
haus: stripping it all seems fine. i didn't want to figure out how wc deals with leading whitespace.

src/ruby/puppetserver-lib/puppet/server/execution.rb
haus: too late. you get a comment

spec/puppet-server-lib/puppet/jvm/execution_spec.rb
haus: updated.

PR #1589, opened 2018-01-08
(PUP-8127) Update matcher for parse errors in environment cache tests

justinstoller:
Update assertion for new error format introduced in
https://github.com/puppetlabs/puppet/commit/f31193ee68666c18d336a56532500b9f86b1627c


haus:
sweet

haus:
is it worth tagging this change with the same jira as the puppet change?

justinstoller:
yeah that's a good idea, amend incoming

PR #1590, opened 2018-01-08
(maint) Test correct branches for 5.1.x and 2.x

haus:


.travis.yml
```diff
@@ -20,6 +20,7 @@ notifications:
 branches:
   only:
     - master
+    - 5.1.x
     - stable
```
Magisus: Can we get rid of stable here?

Magisus:
I'm actually kind of curious why we're even restricting the branches Travis runs on at all. I don't think most of our other projects do this, and I'm not convinced it's buying us much, especially given the new branching scheme that uses version numbers.

haus:
Yea i'm not sure. If i remove the branches will it just run everywhere?

Magisus:
That's what it looks like. The Puppet repo does not have a `branches` section.

haus:
This is failing travis for the thing @justinstoller 's PR fixes

Magisus:
Should this be merged into 2.x?

haus:
@Magisus no idea. i can retarget there and we can just merge up all the way

haus:
or whatever

haus:
so good

haus:
replaced by https://github.com/puppetlabs/puppetserver/pull/1591

PR #1593, opened 2018-01-08
(maint) Merge 5.1.x into master

haus:


haus:
Waiting on https://github.com/puppetlabs/puppet/commit/f31193ee68666c18d336a56532500b9f86b1627c to pass CI so the submodule can be updated. tests will fail until that happens.

PR #1595, opened 2018-01-09
 (SERVER-1922) Defer settting Puppet log level, use logback

haus:

must be set.

When Puppetserver is booting, it first configures the Ruby/JVM bindings
which includes tying logback in as a log destination, and then
configures Puppet. Because Puppet itself is configured after
Puppetserver it's possible for the default Puppet log level to overwrite
the logback derived log level. This commit defers the log level
configuration until Puppet has been initialized to ensure that the
logback determined level takes precedence.


lindboe:
LGTM

PR #1596, opened 2018-01-10
(PUP-8249) Remove gettext-setup and update fast_gettext

Magisus:
This commit updates the fast_gettext version to 1.1.1, which contains a
file encoding bug fix. It also removes gettext-setup, which has ceased
to be a puppet runtime dependency.

### DIFF
--- a/resources/ext/build-scripts/gem-list.txt
+++ b/resources/ext/build-scripts/gem-list.txt
@@ -3,5 +3,4 @@ hocon 1.2.5
 text 1.3.1
 locale 2.1.2
 gettext 3.2.2
-gettext-setup 0.28
-fast_gettext 1.1.0
+fast_gettext 1.1.1


Magisus:
Is there any compatibility reasons why we need to keep installing gettext-setup? I know in the agent we considered removing it something of a breaking change (and therefore haven't done it yet), does that apply here?

justinstoller:
We only require it if Puppet does. I'm not sure why it's a breaking change on Puppet's end, though if it was part of an API for server side extensions (like report termini) then it might be?

haus:
I think termini and custom functions would be the only places this would be exposed to users.

Magisus:
I think the argument was more, it's something we shipped to people's systems as part of the agent package, and they might be using it for things as part of puppet development (say, generating PO files for modules with translations in unsupported languages), less than that it breaks anything with Puppet itself.

PR #1600, opened 2018-01-12
(TK-462) Bring in updated clj-parent dependencies

justinstoller:


haus:
The only dependencies that we pull in that changed are tk-jetty9 and tk-status (https://github.com/puppetlabs/clj-parent/compare/1.4.3...1.6.0).

PR #1603, opened 2018-01-17
(SERVER-1954) Roll bidi back to 1.23.1

haus:
Starting in bidi 1.25.0, paths that contain invalid URI characters throw
an exception. Unfortunately, puppetserver uses tk-jetty9 and explicitly
tells tk-jetty9 to normalize URIs before handing them to the routing
layer. This means paths with encoded spaces are passed to bidi decoded,
and then have an exception raised, which in turn causes puppetserver to
return a 500. As this is undesireable behavior, this commit rolls the
bidi version in use to 1.23.1, which still allows decoded strings.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -47,6 +47,7 @@
                  [org.yaml/snakeyaml]
                  [commons-lang]
                  [commons-io]
+                 [bidi "1.23.1"]
 
                  [clj-time]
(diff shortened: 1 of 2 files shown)


haus:
@Magisus no, but there are enough tk level tests for puppet/server interactions that i'm not super worried about those

PR #1604, opened 2018-01-18
(SERVER-1954) Update comidi to 0.3.2

Magisus:
This commit updates comidi to a version that is compatible with bidi
2.1.3, in which the bug with parsing URIs with spaces has been fixed.
This allows us to roll our bidi version forward to a working version,
instead of pinning back.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -47,7 +47,7 @@
                  [org.yaml/snakeyaml]
                  [commons-lang]
                  [commons-io]
-                 [bidi "1.23.1"]
+                 [bidi "2.1.3"]
 
                  [clj-time]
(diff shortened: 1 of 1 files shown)


project.clj
```diff
@@ -47,7 +47,6 @@
                  [org.yaml/snakeyaml]
                  [commons-lang]
                  [commons-io]
-                 [bidi "1.23.1"]
```
haus: We talked in person about this, but we'll need to keep the bidi override in place to get to 2.1.3 otherwise clj-parent will win.

Magisus: Updated

lindboe:
+1

PR #1609, opened 2018-01-25
(SERVER-1954) Bump jruby-utils to 0.12.0

Magisus:
This commit updates jruby-utils to 0.12.0, which changes the encoding
settings for the JRuby scripting containers to be UTF-8 by default,
instead of US-ASCII. This avoids a corner-case bug when interpolating
translated strings containing Unicode characters.


haus:
@Magisus i think you need to pull in the fixes from https://github.com/puppetlabs/puppetserver/commit/0ee60fa7305d7c7280338ba911916be4046fb6be#diff-098f6bcd4621d373cade4e832627b4f6

PR #1611, opened 2018-01-26
(SERVER-2066) Add hiera-eyaml gem to list of installed gems

Magisus:
This commits adds the hiera-eyaml gem to the list of gems that get
installed with puppetserver. This makes it easier for users to take
advantage of the eyaml encrypted backend for hiera.

### DIFF
--- a/resources/ext/build-scripts/gem-list.txt
+++ b/resources/ext/build-scripts/gem-list.txt
@@ -5,3 +5,4 @@ locale 2.1.2
 gettext 3.2.2
 gettext-setup 0.28
 fast_gettext 1.1.0
+hiera-eyaml 2.1.0


Magisus:
Package built with this change: builds.delivery.puppetlabs.net/puppetserver/5.2.0.master.SNAPSHOT.2018.01.26T0922

resources/ext/build-scripts/gem-list.txt
```diff
@@ -5,3 +5,4 @@ locale 2.1.2
 gettext 3.2.2
 gettext-setup 0.28
 fast_gettext 1.1.0
```
haus: should this version be updated or will that come from a merge up?

Magisus: It should probably come from the merge-up, but we should also update to 1.1.2 in the 5.1.x branch.

Magisus:
The gem appears to be usable via the `puppetserver ruby` command, the same as other executable gems vendored with the server.

PR #1619, opened 2018-01-31
(maint) Add build_tar: FALSE to build_defaults

mwaggett:
Our automation is attempting to sign tarballs that aren't there. This will make sure we don't try to do that.


mwaggett:
Shipping barfed without this, but then making this change made everything good again.

Magisus:
OMG that gif is great.

PR #1622, opened 2018-02-02
(PE-23484) Conditionally disable PDB

justinstoller:


justinstoller:
Saw this issue with my recent PR. This is I *think* the approach to take. Still need to test it locally though.

acceptance/suites/tests/00_smoke/master_starts_if_agent_run_before_ssl_files_inited.rb
```diff
@@ -7,6 +7,10 @@
 puppetservice=options['puppetservice']
 ssldir = master.puppet['ssldir']
 backup_ssldir = master.tmpdir("agent_run_before_master_init_ssldir_backup")
+reports_termini = on(master, puppet("config get --section master reports")).stdout.chomp
```
justinstoller: `config has no 'get' subcommand` (this should be `print` I think)

justinstoller:
I've tested this in FOSS with redhat (pdb integration) and ubuntu (no pdb), testing in pe-pse next.

justinstoller:
I've tested this from the irving branch of pe-pse as well using both redhat7 and ubuntu1604 masters without issues.

acceptance/suites/tests/00_smoke/master_starts_if_agent_run_before_ssl_files_inited.rb
```diff
@@ -7,6 +7,10 @@
 puppetservice=options['puppetservice']
 ssldir = master.puppet['ssldir']
 backup_ssldir = master.tmpdir("agent_run_before_master_init_ssldir_backup")
+reports_termini = on(master, puppet("config print --section master reports")).stdout.chomp
+puppetdb_enabled = reports_termini =~ /puppetdb/
+route_file_exists = on(master, "ls /etc/puppetlabs/puppet/").stdout =~ /routes\.yaml/
```
haus: nit: I think `test -f /etc/puppetlabs/puppet/routes.yaml` would be better here, but :woman_shrugging:

haus:
I think this looks reasonable, but I have no idea what problem it is trying to solve (I assume some configuration difference between PE tests and FOSS tests?) Maybe a few comments in the disable/re-enable code to explain what is happening and/or a more detailed commit message? Otherwise I'm good.

Magisus:
Thanks for those comments.

PR #1623, opened 2018-02-07
(PE-23484) Disable FOSS PuppetDB integration test when running against PE

justinstoller:
Previously we were installing PuppetDB in the FOSS test presuite and
then configuring it in the PuppetDB integration test. This configuring
caused the PuppetDB integration test to fail when ran in PE. Originally,
the effort to put the PuppetDB installation in the presuite was so that
it could be both installed and configured there and the test suite would
run in a more real world like scenario.

Consequently, we attempted to make the presuite both install and
configure PuppetDB and make the PuppetDB test work regardless of the
install type. After several attempts, it's become apparent that the work
to do would require re-writing a considerable number of tests within the
Puppet project as well.

This patch reverts configuring PuppetDB in the presuite, moving the
configuration of it back to the integration test itself, which is now
configured to skip when run against PE.

### DIFF
--- a/acceptance/suites/pre_suite/foss/95_install_pdb.rb
+++ b/acceptance/suites/pre_suite/foss/95_install_pdb.rb
@@ -4,7 +4,7 @@ skip_test unless matching_puppetdb_platform.length > 0
 
 
 test_name 'PuppetDB setup'
-sitepp = '/tmp/configure_puppetdb.pp'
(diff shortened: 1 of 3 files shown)


justinstoller:
I haven't ran this locally yet, but I'm beginning to think the other approach doesn't warrant the effort it's taking.

Magisus:
I think we should just validate this in Jenkins rather than running all the combinations locally, given how close we are to the stop ship for 5.2.0.

PR #1624, opened 2018-02-07
(CPR-515) Only run config tasks on install, not upgrades

shrug:
This depends on https://github.com/puppetlabs/ezbake/pull/472 to enable
setting tasks that run on package install, but not upgrade.

This should address the issue of package upgrades overwriting custom
config values.

### DIFF
--- a/resources/ext/ezbake.conf
+++ b/resources/ext/ezbake.conf
@@ -20,7 +20,7 @@ ezbake: {
                # has probably the way to do it for now.  There might be a better
                # way to get rid of the hard-coded paths here, but I don't
                # know it.
-               postinst: [
+               postinst-install: [
                  "install --owner={{user}} --group={{user}} -d /opt/puppetlabs/server/data/puppetserver/jruby-gems",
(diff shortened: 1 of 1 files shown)


mwaggett:
This looks like we're moving all the things to be run after install only. Is that what we want? Or are there some things that we would also want to run on upgrade?

shrug:
@mwaggett These particular tasks only need to be run once (on install)

shrug:
The relevant ezbake change has been merged, we'll update this as soon as the ezbake release has been cut

justinstoller:
What's up with this/do you need a review from anyone on the server team?

Is this a thing in scope for the next FOSS release? IIRC, we're releasing off of master next, so if we want this in Platform 5.5 we should get it in and merged up this week.

Speaking of release vehicles this is targetting 5.1.7 right now (to go into the next Hoyt z), is changing the upgrade behavior in a z cool?

underscorgan:
This depends on #1648 

I think changing upgrade behavior in a Z should be fine, since the existing upgrade behavior is definitely a bug.

PR #1625, opened 2018-02-08
(maint) Revert "(maint) Add build_tar: FALSE to build_defaults"

mwaggett:
This reverts commit 755a06aa627bbf2c9f82bc1698ebdbe4f971891c.
The FOSS_ONLY setting (used for shipping) was not properly retrieving tarballs,
but our ship process was still attempting to sign tarballs, causing the ship to
fail. This setting was added to prevent the attempted signing, but now that we
have fixed the underlying problem, this setting is no longer needed.

### DIFF
--- a/ext/build_defaults.yaml
+++ b/ext/build_defaults.yaml
@@ -2,4 +2,3 @@
 packaging_url: 'git://github.com/puppetlabs/packaging.git --branch=master'
 packaging_repo: 'packaging'
 project: 'puppetserver'
-build_tar: FALSE


mwaggett:
https://github.com/puppetlabs/packaging/pull/788 needs to go in first.

haus:
Can we get a little more detail about what's going on with this PR? We're going to start building the tar again?

mwaggett:
@haus Did you read https://tickets.puppetlabs.com/browse/RE-10238? Let me know if you need more info than that.

haus:
Do you know how this will work with the ezbake version in use still setting it to false?

mwaggett:
@haus The change has been reverted in ezbake, but even pinned to an older version, we have environment variables to override. This should only be relevant at ship time.

haus:
@mwaggett Okay. That makes sense. Do you mind updating the commit message with more specifics of what the problem that is being solved here is? Public users won't have access to the RE project in jira to see the details.

mwaggett:
@haus updated

haus:
@mwaggett thanks!

haus:
@mwaggett just so i'm clear, with this change we'll be shipping and signing tarballs for puppetserver releases?

mwaggett:
@haus correct, this enables shipping the puppetserver tarballs that end up here: http://downloads.puppetlabs.com/puppet/

PR #1627, opened 2018-02-08
Merge 5.1.x -> master

Magisus:


Magisus:
This should be good to go now, this changeset looks right.

PR #1635, opened 2018-02-16
(SERVER-1663) Document usage of ~/.gemrc and package provider with `puppetserver gem` util

justinstoller:


justinstoller:
Halp, @gguillotte I can't use words good.

justinstoller:
danke

PR #1636, opened 2018-02-23
(SERVER-2024) Clean up puppet install in beaker pre-suite

Magisus:
The logic around installing puppet-agent on the agent nodes during test
setup had suffered a series of refactors that left some nonsense logic
in place. This commit updates that pre-suite to no longer check the
`puppet_version` variable, which a) referred to puppet itself, not the
agent, and b) was unnecessary for install because the version getting
installed is determined by which dev repo gets installed in a previous
step.

The install logic for Windows correctly uses `puppet_build_version` and
was left unchanged.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -21,12 +21,6 @@ module PuppetServerExtensions
                          nil, "Puppet Server Version",
                          "PUPPETSERVER_VERSION", nil, :string)
 
-    puppet_version = get_option_value(options[:puppet_version],
-                         nil, "Puppet Version", "PUPPET_VERSION",
-                         "1.10.9",
-                         :string) ||
-                         get_puppet_version
-
     # puppet-agent version corresponds to packaged development version located at:
     # http://builds.delivery.puppetlabs.net/puppet-agent/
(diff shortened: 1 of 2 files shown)


Magisus:
After this change, we no longer explicitly use `puppet_version` anywhere.How do people feel about removing it from the beaker options file? I sometimes find it useful as a tracker of what version of the submodule is in use, since git and Github make that hard to discover, but it feels a little weird to track an option that is never used.

justinstoller:
I think we should remove `puppet_version` from `options.rb` and the `Rakefile`. I think its confusing whenever a manual bump needs to be done.

Magisus:
Okay, updated. This will need a bit of additional work when we merge up to the 5.x branches because we have an actual options file there.

justinstoller:
I'll give this a run after lunch. There's a typo in the commit message, can you fix that real quick?

Magisus:
Updated I think, if a missing "on" is what you saw.

haus:
@Magisus "Remover" to "Remove"

Magisus:
Ah, derp. Fixed both places.

PR #1638, opened 2018-02-27
(maint) Run tests against jruby 9k as well as 1.7

haus:
Previously we would only run travis and specs against the default jruby.
This is undesirable as we would like to catch test failures against 9k
before merging, so this commit adds a matrix to the travis configuration
to run against both jruby versions. The Rakefile is also updated to
support running specs against jruby9k.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,14 @@
 language: clojure
 lein: 2.7.1
 dist: trusty
-jdk:
-- openjdk8
-- oraclejdk8
+matrix:
+  include:
+  - env: JRUBY_VERSION=9k
+    jdk: openjdk8
+  - env: JRUBY_VERSION=1.7
+    jdk: oraclejdk8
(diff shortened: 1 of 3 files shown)


haus:
this may take a bit of tweaking as it involves travis

haus:
https://travis-ci.org/puppetlabs/puppetserver/builds/347048472 looks like what i think we want. i assigned oracle to one jruby version and openjdk to the other to maintain coverage there.

haus:
loooooking good

haus:
`Facter Could not retrieve fact='hostname', resolution='<anonymous>': Failed while executing '/bin/hostname': Cannot allocate memory - /bin/hostname`

haus:
oh-thats-terrifying

haus:
Well I have no idea if the heap change made a different or not, but it's green.

PR #1639, opened 2018-02-28
(maint) Use correct variable name for nss_package_name

haus:


justinstoller:
Were we seeing this fail somewhere? SLES?

haus:
I dunno. I've never seen it fail. I'm not sure if this code has ever worked.

acceptance/suites/pre_suite/foss/70_install_puppet.rb
```diff
@@ -63,7 +63,7 @@ def install_puppet_from_msi( host, opts )
   end
 
 step "Upgrade nss to version that is hopefully compatible with jdk version puppetserver will use." do
-  nss_package=nil
+  nss_package_name=nil
```
haus: In case anyone cares, `case` in ruby does not create a new scope. Additionally, as ruby sees/parses variables it will initialize them, so this line isn't needed at all.

justinstoller: This line isn't needed but I (at least) find Ruby's behavior surprising and think this change is a good one.

lindboe: I'd be fine without the line but if y'all want it I'm cool with that too

PR #1641, opened 2018-03-01
(maint) Remove service_framework tests from puppetserver

haus:
The service_framework tests are fairly costly tests that don't provide a
lot of value to us. I don't believe they've ever failed from a service
regression. This commit removes them entirely.


acceptance/suites/tests/service_framework/error_during_restart.rb
```diff
@@ -1,33 +0,0 @@
...
-step 'Ensure puppetserver fails to start' do
-  on(master, "service #{puppetservice} stop")
-  on(master, "service #{puppetservice} start", {:acceptable_exit_codes => [1]})
-end
-
-step 'Ensure puppetserver fails to start after a restart' do
-  # Sleep a few seconds to try to provoke contention between the automatic
-  # restart that systemd would do - for restart=on-failure - and the
-  # 'manually requested' one below.  The 'manually requested' restart should
-  # complete but with an error (1) status code.
```
lindboe: From hipchat room: This is really testing trapperkeeper's restart behavior, not so much server behavior. We should rely on trapperkeeper's tests for this.

PR #1643, opened 2018-03-06
(maint) Set environment_timeout to 0 in tasks tests

haus:
Previously environment_timeout was set to unlimited in the tasks tests.
Due to a puppet bug, we still saw new environments as they were created.
Once that bug was fixed, new environments would not be seen with it set
to unlimited because they would not be created until after puppet had
started up. Now we set it to 0 to ensure new environments are seen.

### DIFF
--- a/dev-resources/puppetlabs/services/master/environment_classes_int_test/puppet.conf
+++ b/dev-resources/puppetlabs/services/master/environment_classes_int_test/puppet.conf
@@ -2,4 +2,4 @@
 
 certname = localhost
 environmentpath = $confdir/environments
-environment_timeout = unlimited
+environment_timeout = 0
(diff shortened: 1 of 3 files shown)


lindboe:
should we get rid of the environment.conf if it's not necessary?

lindboe:
Oh god so now I realize the test had this config all along: https://github.com/puppetlabs/puppetserver/blob/5.1.x/dev-resources/puppetlabs/services/master/tasks_int_test/puppet.conf which is possibly what gave us the 0 timeout behavior before. but now that's not being respected, so maybe the local environment.conf stuff was overriding it?

lindboe:
Also, why did we change the other environment.confs? were they causing failures?

haus:
@lindboe yes, they were also causing failures in those tests

haus:
@lindboe that would explain why the tasks-int-test tests weren't failing at least.

lindboe:
...oh yup I mixed up jruby/tasks_test and master/tasks_int_test

haus:
green

PR #1647, opened 2018-03-09
(SERVER-2098) (SERVER-2133) Multijson and agent shared gem dir

Magisus:
This adds the JrJackson gem and the new gem directory from puppet-agent.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -37,7 +37,7 @@
 
   :min-lein-version "2.7.1"
 
-  :parent-project {:coords [puppetlabs/clj-parent "1.7.1"]
+  :parent-project {:coords [puppetlabs/clj-parent "1.7.3"]
                    :inherit [:managed-dependencies]}
 
   :dependencies [[org.clojure/clojure]
@@ -54,6 +54,7 @@
                  [liberator]
                  [org.apache.commons/commons-exec]
                  [io.dropwizard.metrics/metrics-core]
+                 [com.fasterxml.jackson.module/jackson-module-afterburner]
 
                  ;; We do not currently use this dependency directly, but
                  ;; we have documentation that shows how users can use it to
--- a/resources/ext/build-scripts/gem-list.txt
+++ b/resources/ext/build-scripts/gem-list.txt
@@ -5,3 +5,4 @@ locale 2.1.2
 gettext 3.2.2
 fast_gettext 1.1.2
 hiera-eyaml 2.1.0
+jrjackson 0.4.5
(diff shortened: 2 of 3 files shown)


Magisus:
Is there any reason to also have puppetserver install MultiJson? Since all of the code depending on it is in Puppet, I feel like no...

resources/ext/build-scripts/gem-list.txt
```diff
@@ -5,3 +5,4 @@ locale 2.1.2
 gettext 3.2.2
 fast_gettext 1.1.2
 hiera-eyaml 2.1.0
+jrjackson 0.3.8
```
justinstoller: Why not 0.4.5?

justinstoller:
>Is there any reason to also have puppetserver install MultiJson? Since all of the code depending on it is in Puppet, I feel like no...

Agree

resources/ext/build-scripts/gem-list.txt
Magisus: I thought there was a reason, but looking through the changelog now I'm not seeing one. It might have had something to do with me making sure it worked with various JRuby versions. I'll update.

Magisus:
Updated to 0.4.5, seemed to work alright in a basic test.

Magisus:
Turns out both versions of JrJackson require Jackson's [Afterburner module](https://github.com/FasterXML/jackson-modules-base/tree/master/afterburner). I've added a commit to this pulling it in, but I'm curious if this is something that should go in clj-parent instead?

justinstoller:
Doing a quick scan of our Github org I didn't see any other projects pulling in afterburner, and doing a quick scan of cheshire (the JSON serialization library most clojure projects use) it doesn't look like cheshire could easily take advantage of afterburner if we started adding it to our projects.

This seems like something that is only going to be used by jrjackson for the foreseeable future/without a bit of work from other teams, so I'd only put it in Puppet Server.

@pcarlisle and @lindboe might have a better idea of the larger clojure ecosystem's jackson usage here though?

Magisus:
It seems like a very specific set of optimizations for JSON (de)serialization in particular, so I'm not terribly surprised nothing else is using it. If it doesn't need to go in clj-parent, then I think this is ready to go. The errors disappear with the project.clj line added in the last commit.

lindboe:
I have no information to add, this sounds fine to me.

project.clj
```diff
@@ -85,6 +85,8 @@
                  [puppetlabs/comidi]
                  [puppetlabs/i18n]
 
+                 [com.fasterxml.jackson.module/jackson-module-afterburner "2.7.1"]
```
pcarlisle: Does this version have to be synchronized now with the version of jackson transitively pulled in by cheshire from clj-parent?

Magisus: Not sure. @justinstoller ?

justinstoller: ppfft, this is only an issue because of https://tickets.puppetlabs.com/browse/SERVER-1731 jrjackson vendors the jars it needs (and is set to 2.9.4 currently).

It synchronizes its versions of jackson projects and we probably should too. That's probably a good enough reason put it into clj-parent?

lindboe: I talked with @Magisus and she said she'll move it into clj-parent

PR #1650, opened 2018-03-12
(maint) Add Puppet 6 to compatibility tests

lindboe:
This commits add Puppet 6 as an acceptable version to the "modern"
agents in compatibility tests.

### DIFF
--- a/acceptance/suites/puppet3_tests/puppet3_version_test.rb
+++ b/acceptance/suites/puppet3_tests/puppet3_version_test.rb
@@ -10,7 +10,7 @@ end
 
 step "Check that Puppet Server has Puppet 4.x or 5.x installed"
 on(master, puppet("--version")) do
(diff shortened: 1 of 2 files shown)


Magisus:
It feels a little annoying that we will have to keep adding version numbers, but I also kind of like that it forces us to revisit the utility of these tests.

Do we want to add a similar test for Puppet 5 -> 6?

lindboe:
If we do I can add that when I cut the branch to separate concerns

Magisus:
That sounds good.

acceptance/suites/puppet3_tests/puppet3_version_test.rb
```diff
@@ -10,7 +10,7 @@
 
 step "Check that Puppet Server has Puppet 4.x or 5.x installed"
```
pcarlisle: the descriptions are dated now

lindboe: well

PR #1652, opened 2018-03-16
(SERVER-2129) Improve failure message for PuppetDB test

Sharpie:
This commit updates the PuppetDB integration test to make it more robust
and easier to diagnose:

  - The test now handles dataset timestamps that return as `null`.

  - The test retries the query for updated data up to three times. This
    mitigates failures due to race conditions by giving PuppetDB extra
    time to process commands into the database.

  - An informative message is printed upon failure that lists which datasets
    were missing or not updated along with which logs to check for failures.

### DIFF
--- a/acceptance/suites/tests/00_smoke/puppetdb_integration.rb
+++ b/acceptance/suites/tests/00_smoke/puppetdb_integration.rb
@@ -107,13 +107,38 @@ EOM
   end
 end
 
-step 'Validate server sent agent data to PuppetDB' do
+step 'Validate PuppetDB successfully stored agent data' do
   query = "curl http://localhost:8080/pdb/query/v4/nodes/#{master_fqdn}"
-  response = JSON.parse(on(master, query).stdout.chomp)
-  %w[facts_timestamp catalog_timestamp report_timestamp].each do |dataset|
-    assert_operator(Time.iso8601(response[dataset]),
-                    :>,
-                    run_timestamp,
-                   "#{dataset} updated in PuppetDB")
+  agent_datasets = %w[facts_timestamp catalog_timestamp report_timestamp]
+  missing_datasets = [ ]
+  retries = 3
+
+  retries.times do |i|
+    logger.debug("PuppetDB query attempt #{i} for updated agent data...")
+
+    missing_datasets = [ ]
+    response = JSON.parse(on(master, query).stdout.chomp)
+
+    dataset_states = response.select {|k, v| agent_datasets.include?(k)}.map do |k, v|
+      t = Time.iso8601(v) rescue nil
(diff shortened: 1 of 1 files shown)


Sharpie:
@Magisus @lindboe This change should make failures in the PuppetDB integration test easier to diagnose. In the case of the recent issue with JrJackson, the error message would change from:

```
ArgumentError: invalid date: nil
```

To:

```
Minitest::Assertion: PuppetDB did not return updated data for kvypwyu7b7netvb.delivery.puppetlabs.net after
3 consecutive queries. The following timestamps were
missing or not updated:

  report_timestamp

Check puppetserver.log for errors that may have ocurred during
data submission.

Check puppetdb.log for errors that may have ocurred during
data processing.
```

Let me know there's any improvements that could be made to the message.

Magisus:
Ooh, that's much better! We did track this down a real error, the log of which was in `puppetdb.log`, but I personally wouldn't have thought to look there, so this is great.

justinstoller:
This is awesome, @Sharpie ! I think this will also fix: https://tickets.puppetlabs.com/browse/SERVER-2129, right?

Sharpie:
Yeah, the retries should definitely help with that transient. I've retitled the commit -- let me know if there's a better branch than 5.3.x to land it in.

Magisus:
I would land this in 5.1.x.

Sharpie:
5.1.x won't work -- that was prior to the refactor and build-out I did for the integration tests in #1573 which landed in Puppet Server 5.2.0.

Magisus:
Ah okay, this is probably fine then. But we might want to hold till after tagging the release.

acceptance/suites/tests/00_smoke/puppetdb_integration.rb
```diff
@@ -107,13 +107,38 @@
...
+
+    missing_datasets = [ ]
+    response = JSON.parse(on(master, query).stdout.chomp)
+
+    dataset_states = response.select {|k, v| agent_datasets.include?(k)}.map do |k, v|
+      t = Time.iso8601(v) rescue nil
+      [k, t]
+    end.to_h
+
+    missing_datasets = dataset_states.select {|k, v| v.nil? || (v < run_timestamp)}.keys
```
justinstoller: Should we be asserting that the value is greaterthan the run_timestamp here?

Sharpie: Nope. This is checking for datasets that didn't make it in: either they're `null` because no data is there, or the timestamp is older than the time taken before the puppet run.

Later on, we assert that the result of this selection is empty.

PR #1654, opened 2018-03-19
(maint) Update to SNAPSHOT version

Magisus:


project.clj
```diff
@@ -1,4 +1,4 @@
-(def ps-version "5.3.0")
+(def ps-version "5.3.0-SNAPSHOT")
```
justinstoller: I think this should point to the next release, ie `5.3.1-SNAPSHOT`

Magisus: Oh yeah. Updated.

PR #1655, opened 2018-03-19
Merge 5.3.x -> master

Magisus:


project.clj
```diff
@@ -1,4 +1,4 @@
-(def ps-version "5.2.1-master-SNAPSHOT")
+(def ps-version "5.3.1-SNAPSHOT")
```
Magisus: Does this need `master` in it still? Should this be 6.0.0 at this point?

justinstoller: I think the answer is 'yes' to both `6.0.0-master-SNAPSHOT`

Magisus: Updated.

PR #1657, opened 2018-03-20
(SERVER-2095) Remove JRuby 1.7 support

pcarlisle:


haus:
Have you considered pulling the jruby jar back into the uberjar? It would make using AOT a lot easier for puppetserver.

pcarlisle:
Yeah, but I think I'm going to put it off.

project.clj
```diff
@@ -203,7 +201,7 @@
                                                ;; https://github.com/technomancy/leiningen/issues/2216
                                                ;; Hopefully we can remove those `nil`s (if we care)
                                                ;; and this comment when lein 2.7.2 is available.
-                                               ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
```
justinstoller: looks like there's a few whitespace oddities in this file (this and the two changes above)

pcarlisle:
still a little broken

pcarlisle:
should be good to go

acceptance/README.md
```diff
@@ -51,7 +51,13 @@ changes.
 
 ### PACKAGE_BUILD_VERSION when testing a new build with local changes
 
-1. Run `lein with-profile ezbake ezbake build`. This will take a while, and
+1. Run
+   ```
+   lein clean
```
lindboe: this is applicable to older branches too, right?

pcarlisle: Probably yes

lindboe: maybe should have separate PR for that

pcarlisle: sure we can backport this later

lindboe:
Do we want to deprecate the JRUBY_JAR setting and document it as such (and then possibly also log a warning if it's used?)

pcarlisle:
Right now it's still usable if you want to try a different version. If we do SERVER-2157 then we should do that.

lindboe:
I was just thinking about not supporting an option for longer than we need to. SERVER-2157 meaning we should still support it or no?

lindboe:
talked in person: Patrick's adding a note to SERVER-2157 that we should remove the JRUBY_JAR support then and log a warning because it won't do anything meaningful anymore

test/unit/puppetlabs/services/jruby/jruby_puppet_core_test.clj
```diff
@@ -167,10 +167,7 @@
                                     nil)]
 
       (testing "jruby-config default values are used if not provided"
-        ; jruby-utils defaults to jit when running 9k
-        (if jruby-schemas/using-jruby-9k?
-          (is (= :jit (:compile-mode initialized-jruby-config)))
-          (is (= :off (:compile-mode initialized-jruby-config))))
+        (is (= :jit (:compile-mode initialized-jruby-config)))
```
Magisus: Do we plan to update jruby-utils to remove the logic around selecting the default compile-mode? https://github.com/puppetlabs/jruby-utils/blob/6a77927e73c9fcade6265f8044242224565f1ace/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj#L24-L29

lindboe: Oh yeah, good call, that will need an update.

Magisus: I'll file a ticket about that.

PR #1660, opened 2018-03-21
(maint) Remove deprecated setting in tests

pcarlisle:
In puppet 5 app_management doesn't do anything and this is just
producing useless warnings.


lindboe:
If it's Puppet 5, that also applies to the 5.1.x branch or no?

PR #1665, opened 2018-03-27
(SERVER-2147) Update docs with new 9k instructions

Magisus:
This commit adds some additional configuration instructions for those
enabling JRuby 9k. These settings were determined via extensive Gatling
testing to chase down performance degradations when running with 9k.


Magisus:
This should actually go against 5.1.x I think.

PR #1666, opened 2018-03-27
(SERVER-2147) Update docs with new 9k instructions

Magisus:
This commit adds some additional configuration instructions for those
enabling JRuby 9k. These settings were determined via extensive Gatling
testing to chase down performance degradations when running with 9k.

### DIFF
--- a/documentation/configuration.markdown
+++ b/documentation/configuration.markdown
@@ -144,8 +144,13 @@ The init configuration file location depends on your operating system.
     ```
     Jun 01 19:58:16 myhost puppetserver[24001]: Unable to find specified JRUBY_JAR:/opt/puppetlabs/server/apps/puppetserver/jruby-wrong.jar
     ```
+2.  Set other configuration options for better performance
 
-2.  Restart the `puppetserver` service.
+    To ensure good performance with JRuby 9k, the following settings should also be updated:
+    * In `puppetserver.conf`, `jruby-puppet.compile-mode` should be set to `jit`. This setting must be switched back when reverting to 1.7.
+    * Add `-XX:ReservedCodeCacheSize=512m` to `JAVA_ARGS`, typically defined in `/etc/sysconfig/puppetserver`. This scales up the JVM's code cache to the size needed by JRuby 9k. This setting also behaves well when using JRuby 1.7, so it is safe to leave this setting on if you are switching back and forth.
+
+3.  Restart the `puppetserver` service.
 
     You must fully restart the service to use the new JRuby version. A service reload or `kill -HUP` reload is not sufficient.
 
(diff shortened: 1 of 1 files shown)


Magisus:
Is there anything else people think we should mention?

documentation/configuration.markdown
```diff
@@ -144,8 +144,13 @@ The init configuration file location depends on your operating system.
     ```
     Jun 01 19:58:16 myhost puppetserver[24001]: Unable to find specified JRUBY_JAR:/opt/puppetlabs/server/apps/puppetserver/jruby-wrong.jar
     ```
+2.  Set other configuration options for better performance
 
-2.  Restart the `puppetserver` service.
+    To ensure good performance with JRuby 9k, the following settings should also be updated:
+    * In `puppetserver.conf`, `jruby-puppet.compile-mode` should be set to `jit`. This setting must be switched back when reverting to 1.7.
```
justinstoller: In Server 5.3.0 and later, if using JRuby 9k, `jit` is the default compile mode and doesn't need to be toggled.

Magisus: We can update this when we merge up then to leave that part out.

Magisus: This section doesn't even exist at all anymore in the master branch, so that is already taken care of.

lindboe:
This does make we wonder where if anywhere we help figure out how big of a server they need to run puppet server. I found https://puppet.com/docs/puppetserver/5.1/tuning_guide.html#jvm-heap-size but that doesn't say totals, just how much heap to set. I am thinking somewhere we should tell people they should have at least that heap + codecache + enough for the rest of your system to run since it's not clear if you don't know the JVM how these interact

documentation/configuration.markdown
justinstoller: Master's 6.0.0 now and 9k only, right?

Magisus: Yep.

documentation/configuration.markdown
```diff
@@ -144,8 +144,13 @@ The init configuration file location depends on your operating system.
...
+    * Add `-XX:ReservedCodeCacheSize=512m` to `JAVA_ARGS`, typically defined in `/etc/sysconfig/puppetserver`. This scales up the JVM's code cache to the size needed by JRuby 9k. This setting also behaves well when using JRuby 1.7, so it is safe to leave this setting on if you are switching back and forth.
```
gguillotte: What does this setting do in 1.7 if enabled?

If this setting is safe in 1.7, can Server set it by default so the user doesn't need to?

Magisus: We haven't updating our packaging to do so yet, but we have encouraged people to start turning 9k on. I think we are planning to set this by default going forward. When we do, we can update the docs for that release.

PR #1667, opened 2018-03-28
Merge 5.1.x -> 5.3.x

Magisus:


documentation/configuration.markdown
```diff
@@ -144,8 +144,13 @@ The init configuration file location depends on your operating system.
     ```
     Jun 01 19:58:16 myhost puppetserver[24001]: Unable to find specified JRUBY_JAR:/opt/puppetlabs/server/apps/puppetserver/jruby-wrong.jar
     ```
+2.  Set other configuration options for better performance
 
-2.  Restart the `puppetserver` service.
+    To ensure good performance with JRuby 9k, the following settings should also be updated:
+    * In `puppetserver.conf`, `jruby-puppet.compile-mode` should be set to `jit`. This setting must be switched back when reverting to 1.7.
```
justinstoller: This doesn't need to be here for 5.3.x.

Magisus: Removed.

justinstoller: Sorry, I should have called it out. There's a corresponding comment below in the reverting to 1.7 that can also be removed.

Magisus: Oh derp. Removed that too.

PR #1668, opened 2018-03-28
(SERVER-2154) Add Puppet 5 compat tests

Magisus:
This commit adds tests for compatability between Puppet 5 agents and a
Puppet 6 master, similar to the existing compat tests for Puppet 3 and
4.

### DIFF
--- a/acceptance/suites/puppet5_tests/puppet5_version_test.rb
+++ b/acceptance/suites/puppet5_tests/puppet5_version_test.rb
@@ -3,9 +3,9 @@ test_name "Testing Master/Agent backwards compatibility"
 # Agent running on the master is current, not legacy.
 legacy_agents = agents.reject { |agent| agent == master }
 
-step "Check that legacy agents have Puppet 3.x installed"
+step "Check that legacy agents have Puppet 5.x installed"
 on(legacy_agents, puppet("--version")) do
-  assert(stdout.start_with? "3.", "puppet --version does not start with major version 3.")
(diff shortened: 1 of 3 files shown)


Magisus:
So, this feels like something that could maybe be refactored to reduced some repetition, but I was unsure how much effort we wanted to expend on redoing the way these tests are run to parameterize them. I've opened https://github.com/puppetlabs/ci-job-configs/pull/4239 with the ci-job-configs changes.

acceptance/suites/puppet5_tests/puppet5_version_test.rb
```diff
@@ -0,0 +1,20 @@
...
+legacy_agents = agents.reject { |agent| agent == master }
+
+step "Check that legacy agents have Puppet 5.x installed"
+on(legacy_agents, puppet("--version")) do
+  assert_match(/\A5\./, stdout, "puppet --version does not start with major version 5.")
+end
+
+step "Check that Puppet Server has Puppet 6.x installed"
+on(master, puppet("--version")) do
+  assert_match(/\A6/, stdout, "puppet --version does not start with major version 6.x or 6.x")
```
pcarlisle: "6.x or 6.x"

pcarlisle: also should we tighten the check this is copied from to only be 6?

Magisus: That seems right, yeah.

PR #1669, opened 2018-03-29
(SERVER-2166) Note extension authors should use Puppet execution helper

justinstoller:


justinstoller:
@gguillotte halp

Is there any other place you think this should be called out. I'm pretty sure it's just a best practice when writing Puppet extensions.

justinstoller:
Let me know if that helps

documentation/puppetserver_vs_passenger.markdown
```diff
@@ -70,3 +70,22 @@ logged using `Puppet.info` will be logged in Puppet Server and would not have
 been under Passenger. See
 [Puppet Server Configuration Files: logback.xml](./config_file_logbackxml.md)
 for more information, as well as instructions on how to set log levels in Puppet Server.
+
+
+## External Command Execution
+
+When writing extensions that target the server, extension authors should be
+careful to use Puppet's built in execution helper: `Puppet::Util::Execution.execute`
+rather than native Ruby methods (eg "backticks", `system`, or `spawn`).
```
lindboe: `e.g.,`

might also help to show what a backtick looks like but I don't know if markdown is very friendly about that (``` doesn't work :D)

documentation/puppetserver_vs_passenger.markdown
```diff
@@ -70,3 +70,22 @@ logged using `Puppet.info` will be logged in Puppet Server and would not have
...
+most Linux servers, which in a large production environment can easily cause
+Out of Memory errors at the OS level. Puppet Server provides a lighter weight
+way of creating sub-processes which it exposes through
+`Puppet::Util::Execution.execute`.
+
+You should use this during development of all extensions unless there are
+better methods available. For example you should use
+`Puppet::Util::Execution.execute` when writing Ruby based functions, custom
+report processors, Hiera backends, and faces. When writing custom providers
+the `command` helper should be used if possible to help determine suitability.
```
lindboe: This sentence is hard for me to understand. 1) Is there a link we can provide explaining what the `command` helper is? 2) What does "to help determine suitability" mean? I'm probably just missing some knowledge here, but other people might be too.

gguillotte: > 1) Is there a link we can provide explaining what the command helper is?

I'm not sure what the `command` helper refers to here, but if it's `Puppet::Provider::Command` or something in the `Puppet::Util::Execution` YARD docs, then it'd have to be to rubydoc.info. http://www.rubydoc.info/gems/puppet/Puppet/Provider/Command

> 2) What does "to help determine suitability" mean? I'm probably just missing some knowledge here, but other people might be too.

+1

justinstoller:
Think that addresses this round of feedback.

justinstoller:
> Does this apply to 5.1.x (Hoyt)?

Yeah, this applies to all the things, forever ago. I'll update the PR accordingly.

justinstoller:
Retargeted at 2.x

justinstoller:
@Sharpie My understanding was that they standardized everything to call `posix_spawn` in 9k, but 9k has a larger resident set and more active churn in the active set.

I was originally confused because I thought `posix_spawn` didn't do a full `fork` and `exec` (and the delay in my response was from me doing some research on it). What I *think* the state of things is, is that `posix_spawn` doesn't fork in OSX and Solaris. It's not implemented at the kernel level in Linux but in GlibC prior to 2.24 uses `fork` under the hood while GlibC 2.24 and later uses `clone` in a similar manner to `vfork`.

Fwiw, I reproduced an out of memory error on a 2cpu, 6G mem vagrant box with:
```
$ PUPPETSERVER_HEAP_SIZE=5G lein repl
user=> (go)
user=> (def jruby (:scripting-container (first (jruby-pool))))
user=> (def script "20.times.map { Thread.new { sleep 1; 20.times { puts `/bin/hostname` } } }.map {|t| t.join }")
user=> (.runScriptlet jruby script)
```
This immediately fails with ENOMEM on Cent 7, but successfully runs to completion on Ubuntu Bionic (the closest thing to a supported OS with GlibC >= 2.24). It also works fine on my Mac with a heap of 14G (my macbook has 16G total).

justinstoller:
Also this has three approvals so I'm going to go ahead and merge.

PR #1670, opened 2018-03-30
(SERVER-2100) Install lein-exec plugin when in CI

Magisus:
The lein-exec plugin is needed for the updated lein release job. This
commit ensures the plugin gets installed for use in Jenkins.


Magisus:
Relates to https://github.com/puppetlabs/ci-job-configs/pull/4248

PR #1671, opened 2018-04-02
(maint) Use default heap size in travis tests

justinstoller:
We are currently causing ENOMEM errors in our travis testing from
running Facter 2.x in our CI with JRuby 9k. We expect switching to using
Facter 3 when they implement java support for the gem to fix this issue.

Lowering the heap size should help with the memory consumption of the
forked processes that facter 2 creates. In local testing lowering the
heap caused negligible test run time differences.

### DIFF
--- a/ext/travisci/test.sh
+++ b/ext/travisci/test.sh
@@ -2,9 +2,6 @@
 
 set -e
 
-export PUPPETSERVER_HEAP_SIZE=4G
-
-echo "Using heap size: $PUPPETSERVER_HEAP_SIZE"
 echo "Total memory available: $(grep MemTotal /proc/meminfo | awk '{print $2}')"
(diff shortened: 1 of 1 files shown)


justinstoller:
Locally I saw essentially the same timings at 5G vs the defaults:
```
real    27m38.628s
user    59m56.761s                                
sys     5m11.639s
```
vs
```
real    26m41.002s                                                                            
user    73m45.211s                                                                            
sys     5m9.484s
```
See what actually happens in travis

justinstoller:
No out of memory errors, runtime of 37 minutes 11 seconds, compared with https://github.com/puppetlabs/puppetserver/pull/1668 that ran for 34 minutes 29 seconds.

PR #1677, opened 2018-04-04
(SERVER-2157) Re-bundle JRuby jar

Magisus:
This commit updates puppetserver's build process to start bundling its
JRuby jar again, now that we are only supporting JRuby 9k. If the
JRUBY_JAR setting is used, either at the command line or in the init
config, a warning will be issued indicating that it is no longer needed.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,5 +1,4 @@
 (def ps-version "6.0.0-master-SNAPSHOT")
-(def jruby-version "9.1.16.0-1")
 
 (defn deploy-info
   [url]
@@ -65,7 +64,7 @@
                  [net.logstash.logback/logstash-logback-encoder]
 
                  [puppetlabs/jruby-utils "2.0.0"]
-                 [puppetlabs/jruby-deps ~jruby-version]
+                 [puppetlabs/jruby-deps "9.1.16.0-1"]
 
                  ;; JRuby 1.7.x and trapperkeeper (via core.async) both bring in
                  ;; asm dependencies.  Deferring to clj-parent to resolve the version.
@@ -119,8 +118,7 @@
                        :logrotate-enabled false}
                 :resources {:dir "tmp/ezbake-resources"}
                 :config-dir "ezbake/config"
-                :system-config-dir "ezbake/system-config"
-                :additional-uberjars [[puppetlabs/jruby-deps ~jruby-version]]}
+                :system-config-dir "ezbake/system-config"}
 
   :deploy-repositories [["releases" ~(deploy-info "https://artifactory.delivery.puppetlabs.net/artifactory/clojure-releases__local/")]
                         ["snapshots" ~(deploy-info "https://artifactory.delivery.puppetlabs.net/artifactory/clojure-snapshots__local/")]]
@@ -208,7 +206,7 @@
                                                ;; brings in its own version, and older versions of
                                                ;; lein depend on clojure 1.6.
                                                [org.clojure/clojure nil]
-                                               [puppetlabs/puppetserver ~ps-version :exclusions [puppetlabs/jruby-deps]]
+                                               [puppetlabs/puppetserver ~ps-version]
                                                [puppetlabs/trapperkeeper-webserver-jetty9 nil]
                                                [org.clojure/tools.nrepl nil]]
                       :plugins [[puppetlabs/lein-ezbake "1.8.1"]]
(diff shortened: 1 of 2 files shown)


Magisus:
I built a package with these changes and played around with it; it seems to behave correctly, including appropriate warnings either issued to stdout or logged.

project.clj
```diff
@@ -208,7 +207,7 @@
                                                ;; brings in its own version, and older versions of
                                                ;; lein depend on clojure 1.6.
                                                [org.clojure/clojure nil]
-                                               [puppetlabs/puppetserver ~ps-version :exclusions [puppetlabs/jruby-deps]]
+                                               [puppetlabs/puppetserver ~ps-version]
```
justinstoller: Can we get rid of this line complete?

lindboe: +1

project.clj
```diff
@@ -119,8 +119,7 @@
                        :logrotate-enabled false}
                 :resources {:dir "tmp/ezbake-resources"}
                 :config-dir "ezbake/config"
-                :system-config-dir "ezbake/system-config"
-                :additional-uberjars [[puppetlabs/jruby-deps ~jruby-version]]}
```
justinstoller: We might want to still call it out at the top, but I think w/o this line we, can if we want, inline the jruby-version string now.

resources/ext/cli_defaults/cli-defaults.sh.erb
```diff
@@ -1,10 +1,7 @@
 INSTALL_DIR="/opt/puppetlabs/server/apps/<%= EZBake::Config[:real_name] %>"
 
-JRUBY_JAR="${JRUBY_JAR:-${INSTALL_DIR}/jruby-9k.jar}"
-
-if [ ! -e "$JRUBY_JAR" ]; then
-  echo "Unable to find specified JRUBY_JAR: ${JRUBY_JAR}" 1>&2
-  return 1
+if [ -n "$JRUBY_JAR" ]; then
+  echo "Warning: the JRUBY_JAR setting is no longer needed and will be ignored." 1>&2
```
lindboe: I wonder if it's also a good idea to log a warning on startup for puppet server if JRUBY_JAR has been set?

Magisus: I'm not 100% sure how it works, but this warning DOES get issued when puppetserver starts. It goes to the log.

project.clj
Magisus: I'll try it, however, it was still present before we added the exclusion, see https://github.com/puppetlabs/puppetserver/commit/59ebc8cf6e5834a14822746578d8c1194fa65132 which is why I left it in.

project.clj
Magisus: Might as well inline it. Keep it cleaner.

resources/ext/cli_defaults/cli-defaults.sh.erb
lindboe: Oh... that's weird

project.clj
justinstoller: I believe https://github.com/puppetlabs/puppetserver/pull/319 is where it was added

justinstoller: Not that there's any information in that ticket why this is necessary. However looking at PDB they have the same line in their [project.clj](https://github.com/puppetlabs/puppetdb/blob/master/project.clj#L180-L183) with a ticket number that actually explains this pretty well.

tl;dr
Everyone ends the ticket believing this should no longer be necessary and Jeremy includes it in [SERVER-1053](https://tickets.puppetlabs.com/browse/SERVER-1053) in a long list of things to look at when automating Puppet Server's release (ie removing this instead of updating the cljupdatedeps script). We closed Jeremy's ticket after a re-orged because no one knew what issues he was referencing in his list any more.

resources/ext/cli_defaults/cli-defaults.sh.erb
justinstoller: How'd you start Puppet Server? I'd expect this to go to stderr when running `puppetserver start` and when using a service that calls `puppetserver start` for it to go somewhere like systemd's journal.

Magisus: Oh I started it with `service puppetserver start`, and it did go to the journal. I bet with `puppetserver start` it would go to stderr.

project.clj
justinstoller: Re-opened [EZ-35](https://tickets.puppetlabs.com/browse/EZ-35) for allowing us to remove this line.

project.clj
Magisus: Updated.

Magisus:
Rebased to take into account jruby-utils bump.

pcarlisle:
Should we be eliminating jruby-deps next? I initially thought this would include that.

Magisus:
We probably could (I'll need to dig in a little deeper to make sure I understand the implications of that). I can do that as a follow-up PR.

PR #1678, opened 2018-04-04
(maint) Use puppet5 repo when downloading agents for compat test

Magisus:
Previously we were not specifying the repo to use for downloading puppet
5 agents, so the test failed. This commit adds the puppet_collection
option with the 'puppet5' repo.


Magisus:
I have run these tests now and verified that things get installed correctly.

PR #1683, opened 2018-04-12
(PUP-8659) Remove JrJackson

justinstoller:
JrJackson seems to coerce non-UTF8 strings into UTF8 when generating
JSON, rather than raising and execption like the built in JSON module
does.

Puppet relies on this behavior to correctly downgrade to PSON when
there is binary data in the catalog/facts/report.

This remove JrJackson until we can sort out how to properly serialize
binary data with Jackson.

### DIFF
--- a/resources/ext/build-scripts/gem-list.txt
+++ b/resources/ext/build-scripts/gem-list.txt
@@ -5,4 +5,3 @@ locale 2.1.2
 gettext 3.2.2
 fast_gettext 1.1.2
 hiera-eyaml 2.1.0
-jrjackson 0.4.5


jhelwig:
"rather than raising and exception" => "rather than raising an exception"

justinstoller:
Same code, potentially better commit message.

I functionally tested this by removing the jrjackson gem from my local dev setup (`lein gem uninstall jrjackson`), starting the server from the repl, and running puppet from source against the server with a file resource as described in the failing acceptance test (`catalog_with_binary_data.rb`).

tvpartytonight:
👍

PR #1685, opened 2018-04-19
fixing up slash and and/or usage in docs

kayayarai:


gguillotte:
To get this into the docs earlier, and to fix some commit message issues that I think would trip up Jenkins, I'll close this PR but make the changes in a followup targeting the 5.3.x branch.

PR #1686, opened 2018-04-19
(docs)(DOC-3552)(REVIEW ONLY) Create a Server API index.

gguillotte:
The Puppet Server docs index (/documentation/index.md) contains links to
documentation for all Puppet Server-specific API endpoints. However, it
does not provide the same context as the Puppet HTTP API documentation,
and points to too much unrelated content to be a specific point of
entry about HTTP API endpoints.

Copy and revise the [Puppet HTTP API index](https://puppet.com/docs/puppet/latest/http_api/http_api_index.html)
and add details about Puppet Server-specific API endpoints.

This document must be maintained with changes, additions, removals, or
deprecations to both Puppet and Puppet Server HTTP API endpoints, and
should also be maintained for consistency whenever the Puppet HTTP API
index is updated.

### DIFF
--- a/documentation/http_api_index.markdown
+++ b/documentation/http_api_index.markdown
@@ -11,7 +11,7 @@ Many of these endpoints are the same as the [Puppet HTTP API](https://puppet.com
 
 The V1 and V2 APIs were removed in Puppet 4.0.0. The routes that were previously under `/` or `/v2.0` can now be found under the [`/puppet/v3`](#puppet-v3-http-api) API or [`/puppet-ca/v1`](#ca-v1-http-api) API.
 
-Starting with Puppet Server 2.1, Server provides both the current and previous API endpoints, and can serve nodes running Puppet agent 3.x and 4.x. However, Rack masters, WEBrick masters, and Puppet Server 2.0 and earlier cannot serve nodes running Puppet 3.x.
+Starting with Puppet Server 2.1, Server provides both the current and previous API endpoints, and can serve nodes running Puppet agent 3.x and 4.x. However, Rack masters, WEBrick masters, and Puppet Server 2.0 cannot serve nodes running Puppet 3.x.
 
 ## Puppet and Puppet CA APIs
 
--- a/documentation/http_api_index.markdown
+++ b/documentation/http_api_index.markdown
@@ -1,0 +1,141 @@
+---
+layout: default
+title: 'Puppet Server HTTP API: Index'
+---
+
+Puppet Server provides several services via HTTP API, and the Puppet agent application uses those services to resolve a node's credentials, retrieve a configuration catalog, retrieve file data, and submit reports.
+
+Many of these endpoints are the same as the [Puppet HTTP API](https://puppet.com/docs/puppet/latest/http_api/http_api_index.html).
+
+## V1/V2 HTTP APIs (removed)
+
+The V1 and V2 APIs were removed in Puppet 4.0.0. The routes that were previously under `/` or `/v2.0` can now be found under the [`/puppet/v3`](#puppet-v3-http-api) API or [`/puppet-ca/v1`](#ca-v1-http-api) API.
+
+Starting with Puppet Server 2.1, Server provides both the current and previous API endpoints, and can serve nodes running Puppet agent 3.x and 4.x. However, Rack masters, WEBrick masters, and Puppet Server 2.0 and earlier cannot serve nodes running Puppet 3.x.
+
+## Puppet and Puppet CA APIs
+
+Beginning with Puppet 4, Puppet's HTTP API was split into two separately versioned APIs:
+
+-   An API for configuration-related services
+-   An API for the certificate authority (CA).
+
+All configuration endpoints are prefixed with `/puppet`, while all CA endpoints are prefixed with `/puppet-ca`. All endpoints are explicitly versioned: the prefix is always immediately followed by a string like `/v3` (a directory separator, the letter `v`, and the version number of the API).
+
+### Authorization
+
+Authorization for `/puppet` endpoints is still controlled with Puppet's `auth.conf` authorization system.
+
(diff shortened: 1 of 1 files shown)


gguillotte:
@justinstoller FYI, this is to add a new piece of requested documentation.

documentation/http_api_index.markdown
```diff
@@ -0,0 +1,141 @@
...
+-   An API for configuration-related services
+-   An API for the certificate authority (CA).
+
+All configuration endpoints are prefixed with `/puppet`, while all CA endpoints are prefixed with `/puppet-ca`. All endpoints are explicitly versioned: the prefix is always immediately followed by a string like `/v3` (a directory separator, the letter `v`, and the version number of the API).
+
+### Authorization
+
+Authorization for `/puppet` endpoints is still controlled with Puppet's `auth.conf` authorization system.
+
+Puppet Server ignores `auth.conf` for `/puppet-ca` endpoints. Access to the `certificate_status` endpoint is configured in Puppet Server's `ca.conf` file, and the remaining CA endpoints are always accessible. Rack Puppet master servers still use `auth.conf` for `/puppet-ca`.
```
justinstoller: ca.conf is deprecated in favor of new auth.conf

documentation/http_api_index.markdown
```diff
@@ -0,0 +1,141 @@
...
+Beginning with Puppet 4, Puppet's HTTP API was split into two separately versioned APIs:
+
+-   An API for configuration-related services
+-   An API for the certificate authority (CA).
+
+All configuration endpoints are prefixed with `/puppet`, while all CA endpoints are prefixed with `/puppet-ca`. All endpoints are explicitly versioned: the prefix is always immediately followed by a string like `/v3` (a directory separator, the letter `v`, and the version number of the API).
+
+### Authorization
+
+Authorization for `/puppet` endpoints is still controlled with Puppet's `auth.conf` authorization system.
```
justinstoller: Its unclear if this is talking about old auth.conf or new auth.conf. The /puppet endpoints are managed by *both*, with (I think?) the general recommendation that old (ie Puppet's) auth.conf be set to `allow *` and new (ie Puppet Server's) auth.conf contain's the actual authorization rules. The `puppet-ca` routes are controlled solely by the new auth.conf.

documentation/http_api_index.markdown
```diff
@@ -0,0 +1,141 @@
...
+-   `:indirection` is the indirection to which the request is dispatched.
+-   `:key` is the "key" portion of the indirection call.
+
+Using this API requires significant understanding of how Puppet's internal services are structured, but the following documents specify what is available and how to interact with it.
+
+### Configuration management services
+
+The Puppet agent application directly uses these servcies to manage the configuration of a node.
+
+These endpoints accept payload formats formatted as JSON or PSON (MIME types of `application/json` and `text/pson`, respectively) except for `File Content` and `File Bucket File`, which always use `application/octet-stream`.
```
justinstoller: JSON is the default, and while I don't think we've formally deprecated it we're looking to replace the remaining usages of PSON in Puppet 6 (unless users have binary content they should not be using PSON, and generally they should only be receiving binary content via the `File Content` and `File Bucket File` endpoints. I'd appreciate it if we didn't put JSON and PSON on the same footing going forward and caveated it's usage as fallback for binary content that we hope to provide a better solution for soon (this is the Rich Data Types feature).

documentation/http_api_index.markdown
```diff
@@ -0,0 +1,141 @@
...
+-   [File bucket file](https://puppet.com/docs/puppet/latest/http_api/http_file_bucket_file.html)
+-   [File content](https://puppet.com/docs/puppet/latest/http_api/http_file_content.html)
+-   [File metadata](https://puppet.com/docs/puppet/latest/http_api/http_file_metadata.html)
+-   [Report](https://puppet.com/docs/puppet/latest/http_api/http_report.html)
+
+### Informational services
+
+These services are not directly used by Puppet agent, but can be used by other tools.
+
+-   [Status](https://puppet.com/docs/puppet/latest/http_status.html)
```
justinstoller: This exists and should be documented but folks should prefer the Puppet Server status api instead: https://github.com/puppetlabs/puppetserver/tree/master/documentation/status-api/v1

Also this link is broken (it needs an `http_api` in it)

documentation/http_api_index.markdown
```diff
@@ -0,0 +1,141 @@
...
+It also includes these unique APIs, with endpoints containing other URL prefixes:
+
+-   [Status API](./status-api/v1/services.markdown), at `/status/v1/services`
+-   [Metrics v1 (mbeans) API](./metrics-api/v1/metrics_api.markdown), at `/metrics/v1/mbeans`
+-   [Metrics v2 (Jolokia) API](./metrics-api/v2/metrics_api.markdown), at `/metrics/v2/`
+-   Admin API, at `/puppet-admin-api/v1/`:
+    -   [Environment cache](./admin-api/v1/environment-cache.markdown), at `/puppet-admin-api/v1/environment-cache`
+    -   [JRuby pool](./admin-api/v1/jruby-pool.markdown), at `/puppet-admin-api/v1/jruby-pool`
+
+### Error responses
```
justinstoller: Puppet Server specific, or implemented (like the CA), api endpoints have a slightly different error schema of
```
{
  "msg": "",
  "kind": ""
}
```

documentation/http_api_index.markdown
```diff
@@ -0,0 +1,141 @@
...
+| Server can't find an endpoint handler for an HTTP request | 404 Not Found |
+
+Except for HEAD requests, error responses contain a body of a uniform JSON object with the following properties:
+
+-   `message`: (`String`) A human-readable message explaining the error.
+-   `issue_kind`: (`String`) A unique label to identify the error class.
+
+There is also a [JSON schema for error objects](../schemas/error.json).
+
+## CA V1 HTTP API
```
justinstoller: When running Puppet Server we use clojure based CA that does not go through the Ruby indirector based CA. While we worked hard to maintain compatibility with the Ruby indirector based CA and all this documentation should be correct as far as how to use the API is concerned, it does not, actually, go through the indirector any more.

PR #1689, opened 2018-04-26
(SERVER-1975) Time JRuby use per-borrow reason

Sharpie:
The changeset in this PR enhances the JRuby Metrics Service with the ability to time JRuby use per-borrow reason. Prior to these changes, the service only kept a single global borrow timer. This timer could show when JRuby performance was slowing down, but could not reveal whether the slowdown was coming from a specific use such as catalog requests, report uploads or file serving.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
@@ -2,6 +2,7 @@
   (:require [schema.core :as schema]
             [puppetlabs.metrics :as metrics]
             [puppetlabs.services.jruby-pool-manager.jruby-schemas :as jruby-schemas]
+            [clojure.string :as str]
             [clojure.tools.logging :as log]
             [puppetlabs.trapperkeeper.services.status.status-core :as status-core]
             [puppetlabs.comidi :as comidi]
@@ -59,6 +60,8 @@
    :lock-held-timer Timer
    :lock-requests Atom
    :lock-status Atom
+   :hostname schema/Str
+   :metric-registry MetricRegistry
    :sampler-job-id schema/Any
    :queue-limit-hit-meter Meter})
 
@@ -154,6 +157,48 @@
    (comp add-duration-to-instance instance-request-info)
    instances))
 
+(schema/defn timer-for-borrow-reason :- Timer
+  "Create or return a named timer for a JRuby borrow reason.
+
+  If the JRuby instance was borrowed to service a HTTP request, then
+  the timer name is generated from the sanitized Comidi :route-id.
+
+  Returns a timer named 'other' if the borrow reason does not match
+  one of the above cases."
+  [{:keys [hostname metric-registry]} :- JRubyMetrics
+   request :- TimestampedReason]
+  (let [request-info (instance-request-info request)
+        reason (:reason request-info)
+        timer-name (cond
+                     (nil? (schema/check HttpRequestReasonInfo reason))
+                       (-> (get-in reason [:request :route-id])
+                           ;; Remove ":" characters created by stringifying
+                           ;; Clojure keywords.
+                           (str/replace ":" "")
+                           ;; Remove segments created from regexes by Comidi.
+                           ;; Regexes contain characters that create escaping
+                           ;; problems in downstream services that use regexes
+                           ;; to search for or select metrics by name.
+                           (str/replace #"-\/.*\/" ""))
+                     :else "other")]
+    (->> timer-name
+         (str "jruby.borrow-timer.")
+         (metrics/host-metric-name hostname)
+         (.timer metric-registry))))
+
+(schema/defn borrow-timers :- java.util.Map
+  "Returns a map of borrow timers from JRuby Metrics"
+  [{:keys [metric-registry hostname]} :- JRubyMetrics]
+  (let [metric-namespace (metrics/host-metric-name hostname "jruby.borrow-timer")
+        ;; TODO: v4.0 of Dropwizard Metrics has a MetricFilter/startsWith
+        ;;       static method that returns a filter which can be passed
+        ;;       directly to .getTimers.
+        metric-filter (partial filter (fn [[k v]]
+                                        (str/starts-with? k metric-namespace)))]
+    (->> (.getTimers metric-registry)
+         metric-filter
+         (into {}))))
+
 (schema/defn track-successful-borrow-instance!
   [{:keys [borrow-count borrowed-instances]} :- JRubyMetrics
    jruby-instance :- JRubyInstance
@@ -186,16 +231,16 @@
     (log/warn (trs "Unable to find request event for borrowed JRuby instance: {0}" event))))
 
 (schema/defn track-return-instance!
-  [{:keys [return-count borrowed-instances borrow-timer]} :- JRubyMetrics
-   {jruby-instance :instance} :- jruby-schemas/JRubyReturnedEvent]
+  [{:keys [return-count borrowed-instances borrow-timer] :as metrics} :- JRubyMetrics
+   {:keys [instance reason]} :- jruby-schemas/JRubyReturnedEvent]
   (.inc return-count)
-  (when (jruby-schemas/jruby-instance? jruby-instance)
-    (let [id (:id jruby-instance)]
+  (when (jruby-schemas/jruby-instance? instance)
+    (let [id (:id instance)]
       (if-let [ta (get @borrowed-instances id)]
-        (do
-          (.update borrow-timer
-                   (- (System/currentTimeMillis) (:time ta))
-                   (TimeUnit/MILLISECONDS))
+        (let [elapsed-time (- (System/currentTimeMillis) (:time ta))
+              per-reason-timer (timer-for-borrow-reason metrics ta)]
+          (.update borrow-timer elapsed-time (TimeUnit/MILLISECONDS))
+          (.update per-reason-timer elapsed-time (TimeUnit/MILLISECONDS))
           (swap! borrowed-instances dissoc id))
         (log/warn (trs "JRuby instance ''{0}'' returned, but no record of when it was borrowed!" id))))))
 
@@ -300,6 +345,8 @@
    :lock-requests (atom {})
    :lock-status (atom {:current-state jruby-pool-lock-not-in-use
                        :last-change-time (timestamp)})
+   :metric-registry registry
+   :hostname hostname
    :sampler-job-id nil
    :queue-limit-hit-meter (.meter registry (metrics/host-metric-name hostname "jruby.queue-limit-hit-meter"))})
 
--- a/src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
@@ -163,6 +163,9 @@
   If the JRuby instance was borrowed to service a HTTP request, then
   the timer name is generated from the sanitized Comidi :route-id.
 
+  If a Clojure Keyword was used as the JRuby borrow reason, then the
+  timer is generated from the Keyword's name and namespace.
+
   Returns a timer named 'other' if the borrow reason does not match
   one of the above cases."
   [{:keys [hostname metric-registry]} :- JRubyMetrics
@@ -180,6 +183,10 @@
                            ;; problems in downstream services that use regexes
                            ;; to search for or select metrics by name.
                            (str/replace #"-\/.*\/" ""))
+                     (keyword? reason)
+                       (if-let [name-space (namespace reason)]
+                         (str name-space "." (name reason))
+                         (name reason))
                      :else "other")]
     (->> timer-name
          (str "jruby.borrow-timer.")
--- a/src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
@@ -88,6 +88,12 @@
    :reason RequestReasonInfo
    :time schema/Num})
 
+(def TimerSummary
+  {:count schema/Int
+   :mean schema/Num
+   :max schema/Num
+   :rate schema/Num})
+
 (def JRubyMetricsStatusV1
   {(schema/optional-key :experimental)
    {:jruby-pool-lock-status JRubyPoolLockStatus
@@ -102,6 +108,7 @@
               :average-free-jrubies schema/Num
               :average-borrow-time schema/Num
               :average-wait-time schema/Num
+              :borrow-timers {schema/Keyword TimerSummary}
               :requested-instances [InstanceRequestInfo]
               :borrowed-instances [InstanceRequestInfo]
               :num-pool-locks schema/Int
@@ -204,6 +211,36 @@
                                         (str/starts-with? k metric-namespace)))]
     (->> (.getTimers metric-registry)
          metric-filter
+         (into {}))))
+
+(schema/defn summarize-borrow-timers
+  "Generates a summary for each JRuby borrow timer
+
+  This function iterates through the list of registered JRuby borrow timers
+  and summarizes each one."
+  [{:keys [hostname] :as metrics} :- JRubyMetrics]
+  (let [timer-namespace (metrics/host-metric-name hostname "jruby.borrow-timer")
+        get-name (fn [timer-name]
+                   (if (= timer-namespace timer-name)
+                     "total"
+                     (str/replace-first timer-name
+                                        (str timer-namespace ".") "")))
+        timer-summary (fn [timer]
+                        (let [snapshot (.getSnapshot timer)]
+                        {:count (.getCount timer)
+                         :mean (->> snapshot
+                                   .getMean
+                                   (.toMillis TimeUnit/NANOSECONDS))
+                         :max  (->> snapshot
+                                    ;; Using 99.9th percentile instead of the
+                                    ;; max as it is properly weighted for the
+                                    ;; past 5 minutes.
+                                    (.get999thPercentile)
+                                    (.toMillis TimeUnit/NANOSECONDS))
+                         :rate (.getFiveMinuteRate timer)}))]
+    (->> (borrow-timers metrics)
+         (map (fn [[k v]] (vector (keyword (get-name k))
+                                  (timer-summary v))))
          (into {}))))
 
 (schema/defn track-successful-borrow-instance!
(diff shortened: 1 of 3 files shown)


Sharpie:
Put this up as "DO NOT MERGE". Some details aren't finalized and the `/status` output doesn't have tests yet, but the implementation is at a point where it'd be useful to get some feedback.

Prior to the patchset, this is what we get from `/status` around JRuby performance:

```
curl -sk https://localhost:8140/status/v1/services/jruby-metrics\?level\=debug |jq '.status.experimental.metrics|with_entries(select(.key|test("borrow")))'

{
  "borrow-count": 10,
  "borrow-timeout-count": 0,
  "borrow-retry-count": 0,
  "borrowed-instances": [],
  "average-borrow-time": 180
}
```

You can see how many times a JRuby was used to process a request and the average time taken to process all requests (within the last 5 minutes).  With the changes, we get:

```
$ curl -sk https://localhost:8140/status/v1/services/jruby-metrics\?level\=debug |jq '.status.experimental.metrics|with_entries(select(.key|test("borrow")))'

{
  "borrow-count": 10,
  "borrow-timeout-count": 0,
  "borrow-metrics": [
    {
      "reason": "total",
      "count": 10,
      "mean": 143,
      "aggregate": 1430
    },
    {
      "reason": "puppet-v3-file_metadatas",
      "count": 4,
      "mean": 267,
      "aggregate": 1068
    },
    {
      "reason": "puppet-v3-catalog",
      "count": 2,
      "mean": 90,
      "aggregate": 180
    },
    {
      "reason": "puppet-v3-node",
      "count": 2,
      "mean": 68,
      "aggregate": 136
    },
    {
      "reason": "puppet-v3-report",
      "count": 2,
      "mean": 42,
      "aggregate": 84
    }
  ],
  "borrow-retry-count": 0,
  "borrowed-instances": [],
  "average-borrow-time": 143
}
```

Which gives us performance metrics for each type of operation that JRubies are being borrowed for. For consistency, I patterned the output after what the Puppet Profiler Service produces for timing of function and resource metrics. However, there are a couple of design questions that we might want to consider:

  - The output is structured as a list of items, but a hash may be an easier datastructure to consume?

  - The output just includes `mean`, `count`, and `aggregate`. Should we return other metrics such as percentiles?

  - The `aggregate` metric is not useful since it is produced by multiplying the count of operations since startup by the mean time per operation **over the last 5 minutes** (see [TK-468](https://tickets.puppetlabs.com/browse/TK-468) for more details). The resulting number is not meaningful due to the difference in timescale. Is there a different sort of total time that would be useful to track?

src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
```diff
@@ -2,6 +2,7 @@
   (:require [schema.core :as schema]
             [puppetlabs.metrics :as metrics]
             [puppetlabs.services.jruby-pool-manager.jruby-schemas :as jruby-schemas]
+            [clojure.string :as string]
```
justinstoller: nit, almost everyone else we use `:as str` even though that shadows `clojure.core/str`. I *think* our require is idiomatic but going to check in the clojurians channel.

Sharpie: Looks like Puppet Server is split with 7 instances of `:as str` and 4 instances of `:as string`. Happy to switch to `:as str` as that seems to be the preference.

Sharpie: Replaced with `:as str`.

src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
```diff
@@ -85,6 +88,14 @@
    :reason RequestReasonInfo
    :time schema/Num})
 
+(def BorrowMetric
+  {:reason schema/Str
+   :count schema/Int
+   :mean schema/Num
+   ;; FIXME: This metric is currently busted for reasons outlined in
+   ;;        TK-468. Find a suitable replacement or remove it.
+   :aggregate schema/Num})
```
justinstoller: I would think max or maybe 99.9th percentile would be better.

src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
```diff
@@ -152,6 +166,73 @@
...
+                           (str/replace #"-\/.*\/" ""))
+                     (keyword? reason)
+                       (if-let [name-space (namespace reason)]
+                         (str name-space "." (name reason))
+                         (name reason))
+                     :else "other")]
+    (->> timer-name
+         (str "jruby.borrow-timer.")
+         (metrics/host-metric-name hostname)
+         (.timer metric-registry))))
```
justinstoller: Do we want to remove older metrics when creating newer ones? Is there a hard stop at how many borrow metrics we may have?

src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
```diff
@@ -344,6 +440,7 @@
                                    :average-free-jrubies (metrics/mean free-jrubies-histo)
                                    :average-borrow-time (metrics/mean-millis borrow-timer)
                                    :average-wait-time (metrics/mean-millis wait-timer)
+                                   :borrow-metrics (top-borrow-summaries metrics 50)
```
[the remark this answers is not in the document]
Sharpie: 50 resulted from me copying most of the logic from the Puppet ProfilerService which limits the resource and function timers to 50 entries in status output:

https://github.com/puppetlabs/puppetserver/blob/5.3.2/src/clj/puppetlabs/services/puppet_profiler/puppet_profiler_core.clj#L120-L121

It's definitely on the high side as there are only 14 Ruby routes defined over in the MasterService:

https://github.com/puppetlabs/puppetserver/blob/5.3.2/src/clj/puppetlabs/services/master/master_core.clj#L715

Maybe 24? Or just remove the limit?

src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
```diff
@@ -107,6 +119,8 @@
               :queue-limit-hit-count schema/Int
               :queue-limit-hit-rate schema/Num}}})
 
+(declare borrow-timers)
```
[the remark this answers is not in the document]
Sharpie: I could, but it felt better placed down in the "Public" section as it is intended to be called from other namespaces that need to get the list of borrow timers. The forward declaration is here because the status summary methods in the "Private" section use it.

Could definitely move it up here, but that seems like it might run counter to the existing organization of the file:

  - Definitions
  - Private functions
  - Public functions

src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
Sharpie: My current thought is to drop aggregate altogether and use `:count` as the sorting key. The upstream Dropwizard Metrics library has added a sum to Timers which looks like it  will ship in version 5.x:

https://github.com/dropwizard/metrics/pull/1254

Once that lands, could add `:sum` here and to other places that report data from Timers. Then eventually deprecate and drop the `:aggregate` entries (somehow? Not sure what the best deprecation strategy would be for this sort of data.)

Percentiles would be great too and by bundling them in we could get something close to the Prometheus Summary metric:

https://prometheus.io/docs/concepts/metric_types/#summary

@npwalker @suckatrash @abottchen @jarretlavallee Any thoughts on which percentile values would be the most helpful?

src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
Sharpie: Right now there is no hard stop. We just dynamically make a new counter for each borrow reason. The MetricRegistry class takes care of ensuring this is done in a thread-safe manner.

We could implement some sort of cleanup, but we already have MetricsPuppetProfiler.java dynamically creating a separate timer for each unique combination of certname, environment, and resource or function name that occurs during catalog compilation. That commonly creates 1000s of timer objects, so we don't have much to gain by implementing cleanup logic for the couple dozen that would be created here.

We should also totally look into the RAM usage from spawning all those timers in MetricsPuppetProfiler.java. I don't think it has ever presented as a "memory leak", but it is an unbounded allocator of a lot of persistent objects.

src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
```diff
@@ -152,6 +166,73 @@
...
+(schema/defn summarize-borrow-timers :- [BorrowMetric]
+  "Generates a summary for each JRuby borrow timer
+
+  This function iterates through the list of registered JRuby borrow timers
+  and summarizes each one."
+  [{:keys [hostname] :as metrics} :- JRubyMetrics]
+  (let [timer-namespace (metrics/host-metric-name hostname "jruby.borrow-timer")
+        get-name (fn [timer-name]
+                   (if (= timer-namespace timer-name)
+                     "total"
```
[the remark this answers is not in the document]
Sharpie: Yup! The existing `borrow-timer` has a name that ends with `jruby.borrow-timer`. The timers added by this PR all have longer names: `jruby.borrow-timer.<something>`.

So, this conditional associates the existing `borrow-timer` with the name "total".

src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
```diff
@@ -298,9 +379,24 @@
    :lock-requests (atom {})
    :lock-status (atom {:current-state jruby-pool-lock-not-in-use
                        :last-change-time (timestamp)})
+   :metric-registry registry
+   :hostname hostname
```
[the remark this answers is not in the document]
Sharpie: We could move all of these timers into their own registry, but there are a couple of complicating factors:

  - We'd have to also put the existing `borrow-timer` into the new registry in order to keep the full set together.

  - We'd have to copy in logic for wiring the registry into JMX and the Graphite reporter (if configured).

By using the existing registry we get the second point for free from the TK init routines, which saves a lot of work.

samwoods1:
Some general comments:
I agree that a hash would be easier to consume than a list
Also agree on using count as the sorting key and dropping aggregate (or deprecating it somehow)
Does any of this code have unit tests? Should it? It seems like if our customers and/or CS are relying on this code for information about their environments, it is worthy of having some unit tests.

src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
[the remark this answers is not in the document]
justinstoller: And `borrow-timers` would be used in SERVER-2198 after this and https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/pull/192 are merged? Just making sure I follow the reasoning behind that visibility level.

src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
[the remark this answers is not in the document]
justinstoller: Let's remove the limit.

This kinda ties into the hard limit of timers conversation. I would imagine that right now 50 is effectively all of them. And if we end up doing something that accidentally automatically creates a bunch of timers we could end up hiding how many we're actually creating in that case.

I think there will be thresholds that determine interesting vs uninteresting data about borrow-timers, but I have a feeling that those thresholds will either be based on change in a timer vs itself over time, or a timer vs the whole group, both sets of queries I think are best done post metric collection and to enable them we'd need to send all the timers.

src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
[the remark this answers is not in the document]
Sharpie: Not quite. The forward declaration of `borrow-timers` is there purely to preserve the existing organization of the source file.

If puppetlabs/trapperkeeper-webserver-jetty9#192 is merged, SERVER-2198 becomes a one-line addition to the `track-return-instance!` function:

    (MDC/put "jruby.borrow-time" elapsed-time)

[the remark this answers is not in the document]
justinstoller: Okay. I wouldn't mark borrow-timers as a public function unless we have an external consumer now or plan to in the immediate future.

src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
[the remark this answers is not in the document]
justinstoller: Yeah, that's a good point about not creating an additional registry that folks would have to configure.

I think that the surrounding code assumes objects that can be initialized and then acted upon, queried, for better or worse, though.

What if, instead of saving the hostname and registry into the `metrics` map, we created an atom that contained them and the existing `borrow-timer` metric. Then we can create an "update-borrow-timers" kinda function that takes the `borrow-timers` atom and the `TimestampedReason` and updates the total and per-reason timers (like lines 274-277 in `track-return-instance!`). We can then also just pass that atom into `summarize-borrow-timers` instead of the whole metrics map. And per https://github.com/puppetlabs/puppetserver/pull/1689/files#r189341818 basically get rid of `top-borrow-summaries`.

I don't know if that'd be worth the effort, but I'm not a huge fan of passing all of metric through `top-borrow-summaries` into `summarize-borrow-metrics` or adding `hostname` and `registry` to the metrics map.

justinstoller:
iirc, this has been reviewed and is waiting on the author for updates.

src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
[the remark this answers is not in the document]
Sharpie: I've removed the "top 50" limit.

src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
Sharpie: Dropped `aggregate` and added 99.9th percentile as `max` as it has the same "past five minutes" weighting as the `mean`. Also added the FiveMinuteRate as `rate` to give a measurement of request rate.

src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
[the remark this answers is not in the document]
Sharpie: After hemming and hawing about this quite a bit, I don't think it's worth the extra effort. `hostname` and `registry` have to be in the metrics map in order to dynamically generate new timers. Adding an atom could remove the need for the `borrow-timers` method and memoize the overhead of filtering the metrics list. But, it would push complexity into other functions and we already do several similar filter operations per-status call over in the `puppet-profiler` service.

This definitely could be done better, but I think:

  - It's good enough at the moment.
  - The best approach to eliminate state like `hostname` may be to improve the encapsulation that the `trapperkeeper-metrics` service offers.

Sharpie:
Alright, this one is finally ready for another look. I removed the "top 50" filter in the status output, swapped the useless `aggregate` measurement out for `max` and `rate`, and added some integration tests to ensure the expected values are returned from the `status/` endpoint after JRuby operations occur.

Sharpie:
This is what the output currently looks like:

```
# curl -ks https://localhost:8140/status/v1/services/jruby-metrics?level=debug| \
  /opt/puppetlabs/puppet/bin/ruby -rjson -e 'puts JSON.pretty_generate(JSON.parse($stdin.read).dig("status", "experimental", "metrics", "borrow-timers"))'
{
  "total": {
    "count": 14,
    "mean": 515,
    "max": 2949,
    "rate": 0.029877917992745882
  },
  "puppet-v3-catalog": {
    "count": 2,
    "mean": 2374,
    "max": 2949,
    "rate": 0.12559796690547395
  },
  "puppet-v3-file_metadata": {
    "count": 2,
    "mean": 97,
    "max": 147,
    "rate": 0.12559796690547395
  },
  "puppet-v3-file_metadatas": {
    "count": 6,
    "mean": 114,
    "max": 412,
    "rate": 0.37064052705790257
  },
  "puppet-v3-node": {
    "count": 2,
    "mean": 535,
    "max": 1317,
    "rate": 0.12352201510956584
  },
  "puppet-v3-report": {
    "count": 2,
    "mean": 140,
    "max": 209,
    "rate": 0.12770880783313002
  }
}
```

src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
```diff
@@ -154,6 +164,85 @@
...
+        timer-summary (fn [timer]
+                        (let [snapshot (.getSnapshot timer)]
+                        {:count (.getCount timer)
+                         :mean (->> snapshot
+                                   .getMean
+                                   (.toMillis TimeUnit/NANOSECONDS))
+                         :max  (->> snapshot
+                                    ;; Using 99.9th percentile instead of the
+                                    ;; max as it is properly weighted for the
+                                    ;; past 5 minutes.
```
justinstoller: I don't understand

Sharpie: The metrics library defaults to using an ExponentiallyDecayingReservoir which stores the last 1028 measurements. Statistics generated from this reservoir, such as the `mean` and `rate` are weighted to be representative of the past 5 minutes.

Except for the `max`. Max is just the largest number in the list of 1028 measurements.

So, if we used `max` we'd have two measurements that are weighted to represent the past 5 minutes and one that isn't. 99.9th percentile does use the same weighting, so it can be meaningfully compared with the `mean` and `rate`.

src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
```diff
@@ -163,7 +163,10 @@
   If the JRuby instance was borrowed to service a HTTP request, then
   the timer name is generated from the sanitized Comidi :route-id.
 
-  Returns a timer named 'other' if the borrow reason does not match
+  If a Clojure Keyword was used as the JRuby borrow reason, then the
+  timer is generated from the Keyword's name and namespace.
+
+  Returns a named labeled 'other' if the borrow reason does not match
```
Magisus: Looks like we lost the word "timer" here.

Sharpie: Fixed!

justinstoller:
I threw the dnm label on this precisely because I think this is finally ready to merge, but we have already crossed the stop ship for the 5.3.x "z" (and not without controversy). I will be cutting the release pipelines for 5.3.9 and then we can merge this in and up.

justinstoller:
release branches are pushed and pipelines are deployed, we should be go to go merging to any of our usual branches now

PR #1691, opened 2018-05-07
(maint) Fix mocked Ring requests and JRuby metrics error handling

Sharpie:
The Travis tests for PR #1689 were timing out. Turned out that the cause was validation failing on mocked Ring requests that had previously never been exposed to the schemas in `puppetlabs.services.jruby.jruby-metrics-core`. This PR corrects three issues:

  - The callback function registered by `puppetlabs.services.jruby.jruby-metrics-core` has no error trap. This caused the exceptions thrown by schema to be unlogged which meant that JRuby instances were silently never returned to the pool --- hence the test suite timeout.

  - Several mocked ring requests were using `:GET` or `"get"` as their HTTP request method. Both Comidi and the Ring specs require lower-cased keywords like `:get`.

  - The schema for `jruby-metrics-core/instance-request-info` expects the HTTP request map to include information from a Comidi route handler. This caused mocks that had not been routed through the MasterService to fail.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
@@ -143,7 +143,9 @@
     (assoc-in instance [:reason :request]
               {:uri (:uri request)
                :request-method (:request-method request)
-               :route-id (get-in request [:route-info :route-id])})
+               :route-id (get-in request
+                                 [:route-info :route-id]
+                                 "unknown-http-request")})
     instance))
 
 (schema/defn ^:always-validate requested-instances-info :- [InstanceRequestInfo]
--- a/src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
@@ -304,14 +304,19 @@
 (schema/defn jruby-event-callback
   [metrics :- JRubyMetrics
    event :- jruby-schemas/JRubyEvent]
-  (case (:type event)
-    :instance-requested (track-request-instance! metrics event)
-    :instance-borrowed (track-borrow-instance! metrics event)
-    :instance-returned (track-return-instance! metrics event)
-    :lock-requested (track-lock-requested! metrics (:reason event))
-    :lock-acquired (track-lock-acquired! metrics (:reason event))
-    :lock-released (track-lock-released! metrics (:reason event))
-
+  (if-let [[func & args] (case (:type event)
+                           :instance-requested [track-request-instance! metrics event]
+                           :instance-borrowed [track-borrow-instance! metrics event]
+                           :instance-returned [track-return-instance! metrics event]
+                           :lock-requested [track-lock-requested! metrics (:reason event)]
+                           :lock-acquired [track-lock-acquired! metrics (:reason event)]
+                           :lock-released [track-lock-released! metrics (:reason event)]
+                           nil)]
+    (try
+      (apply func args)
+      (catch Exception e
+        (log/error e (trs "Error ocurred while recording metrics for jruby event type: {0}" (:type event)))
+        (throw e)))
     (throw (IllegalStateException. (trs "Unrecognized jruby event type: {0}" (:type event))))))
 
 (schema/defn ^:always-validate v1-status :- status-core/StatusCallbackResponse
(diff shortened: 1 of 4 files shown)


src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
```diff
@@ -140,10 +147,13 @@
...
-              {:uri (:uri request)
-               :request-method (:request-method request)
-               :route-id (get-in request [:route-info :route-id])})
+    (if (nil? (schema/check ComidiRequest request))
+      ;; Schema check passed. We can assume this is a full Comidi request.
+      (assoc-in instance [:reason :request]
+                {:uri (:uri request)
+                 :request-method (:request-method request)
+                 :route-id (get-in request [:route-info :route-id])})
+      (assoc instance :reason ::unknown-http-request))
```
Sharpie: The other option would be to make `:route-id` an optional key in the schema.

src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
```diff
@@ -304,14 +314,19 @@
...
+                           :instance-returned [track-return-instance! metrics event]
+                           :lock-requested [track-lock-requested! metrics (:reason event)]
+                           :lock-acquired [track-lock-acquired! metrics (:reason event)]
+                           :lock-released [track-lock-released! metrics (:reason event)]
+                           nil)]
+    (try
+      (apply func args)
+      (catch Exception e
+        (log/error e (trs "Error ocurred while recording metrics for jruby event type: {0}" (:type event)))
+        (throw e)))
```
Sharpie: Re-throwing the exception keeps the same behavior that existed prior to this PR. I suppose we could also swallow the exception here in order to allow the JRuby instance to be returned to the pool. Potential downside is that metrics could start drifting, and a pile-up of uncleared `instance-requested` events could cause `max-queued-requests` to start denying requests event though instances had been returned.

src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
justinstoller: So we say the instance that we pass in should be a [`TimestampedReason`](https://github.com/puppetlabs/puppetserver/pull/1691/files#diff-06e8dcaea6b81022d68725a975548198R65) which is `{:time Long :reason schema/Any}`. Does it make sense to tighten up the input to this method instead?

Sharpie: Ah, the problem isn't the input schema. It's the output schema of `TimestampedReasonWithRequestInfo` which requires that any `:reason` that is a map [that contains a `:request` key](https://github.com/puppetlabs/puppetserver/pull/1691/files#diff-06e8dcaea6b81022d68725a975548198R77) shall have a [non-null :route-id](https://github.com/puppetlabs/puppetserver/pull/1691/files#diff-06e8dcaea6b81022d68725a975548198R71)

A regular Ring mock causes this output assertion to fail as there is no Comidi `:route-info` for the `:route-id` to be extracted from.

justinstoller: I'm sorry, I was having a weird day when I reviewed these two PRs the first time. I meant to ask that comment wrt the schema check on line 150:
```clj
    (if (nil? (schema/check ComidiRequest request))
```
Should we, instead of adding more checking inside the method, tighten up the validation that schema is doing on the initial input? ie, if `TimestampedReason` contains the `:request` key that key should be a `ComidiRequest` (I assume something with the `:request` key should have some kind of valid request even if its not comidi)?

If not, ie if we need to support non-comidi requests, then for this section can we just give the `:route-id` a default value? Like:
```clj
                 :route-id (get-in request [:route-info :route-id] ::unknown-http-request)})
```

justinstoller: I guess my concern with this function's changes are: is this doing work in our function body that schema should be doing for us in the input validation? and is it actually valid behavior to support non-comidi request objects outside of tests (old compojure or liberator maybe??)?

Sharpie: > Should we, instead of adding more checking inside the method, tighten up the validation that schema is doing on the initial input? ie, if TimestampedReason contains the :request key that key should be a ComidiRequest (I assume something with the :request key should have some kind of valid request even if its not comidi)?

The `:request` key should always hold a Ring request map which can optionally include some Comidi metadata added by the routes in the `MasterService`. The output schema of the `instance-request-info` function currently requires that Comidi metadata be present.

If we tighten up the input validation, then a bunch of unit tests for the `RequestHandlerService` that currently mock Ring request maps will have to start mocking Ring+Comidi request maps or pull in a dependency on the `MasterServer` --- both of those options seem like bad directions to go as they increase the coupling of the code.

I like the idea of just adding a default to the `get-in` function that returns `"unknown-http-request"` for `:route-id` if the Comidi metadata isn't present. Seems like the most elegant solution that will satisfy the existing schemas.

Sharpie:
Dropped the additional schema check and added "unknown-http-request" as a default to the `get-in` call that retrieves the Comidi route ID.

PR #1692, opened 2018-05-09
(SERVER-2132) Remove semantic_puppet

jtappa:
I'm not certain, but I think this may have to wait for merge until https://github.com/puppetlabs/puppet/pull/6827 is merged.

### DIFF
--- a/resources/ext/build-scripts/gem-list.txt
+++ b/resources/ext/build-scripts/gem-list.txt
@@ -1,4 +1,4 @@
-semantic_puppet 0.1.3
+semantic_puppet 1.0.2
 hocon 1.2.5
 text 1.3.1
 locale 2.1.2


jtappa:
@Magisus as I understand it, puppet-agent will get it from here: https://github.com/puppetlabs/puppet-runtime/pull/51

Magisus:
I'm specifically asking what directory that gem component installs to. I haven't kept track if we've made the shared dir the default or not...

jtappa:
Hmmm, I'm not certain, but @joshcooper would know.

joshcooper:
Ah right, @Magisus. It looks like puppet-agent (via the puppet-runtime) is installing the gem into the normal gem home:

```
# find /opt/puppetlabs/ -type f -name semantic_puppet-1.0.2.gemspec
/opt/puppetlabs/puppet/lib/ruby/gems/2.4.0/specifications/semantic_puppet-1.0.2.gemspec
```

@jtappa could you submit a PR to the puppet-runtime component so that `semantic_puppet` gets installed into `/opt/puppetlabs/puppet/lib/ruby/vendor_gems` (and whatever the windows equivalent is)? I think we're already installing `multi_json` in there too. We'll need to block this PR until that gets sorted out.

jtappa:
https://github.com/puppetlabs/puppet/pull/6827 and https://github.com/puppetlabs/puppet-runtime/pull/51 have been merged, so I think this is ready to go now?

justinstoller:
Hey, Jorie! I was functionally reviewing this and I realized that this is going to cause an issue with our dev environment, since Puppet will need semantic_puppet, but we won't have a reference to it anywhere besides the agent packaging.

We currently setup the dev environment using the `dev-setup` script which reads this file and installs the gems mentioned in it into our vendor directory (see [here](https://github.com/puppetlabs/puppetserver/blob/master/dev-setup#L23)).

Could we add this to another list, that can just be used by `dev-setup` and not by our packaging tool?

joshcooper:
@jtappa @justinstoller Since puppet has a hard runtime dependency on `semantic_puppet` (and `fast_gettext`, etc), I'm thinking we should just bump the `semantic_puppet` version to match what puppet currently depends on (`~> 1.0`). We should probably bump the `puppet` submodule at the same time, so that puppetserver is using a consistent version of puppet and semantic_puppet?

For the longer term, we're trying to get away from having multiple lists of puppet dependencies (Gemfile, gemspec, puppet-agent, puppetserver) and supporting all of the ways puppet can be installed (gem, bundler, puppet-agent packages, git acceptance, puppetserver dev). See https://tickets.puppetlabs.com/browse/PA-1077. Can puppetserver introspect the puppet gem, or load its gemspec, and install whatever gems it depends on? That way we can change puppet dependencies in the puppet repo, and nothing in puppetserver would need to change.

justinstoller:
> So I'm wondering, can we move all of the gems that puppet-agent has a hard dependency on into puppet-agent's shared directory, e.g. locale, gettext, fast_gettext, text, semantic_puppet? We already have hocon there. Then puppetserver doesn't need to vendor the gems during its packaging, and we can only use gem-list when running in dev mode?

I think we can from server team's perspective. I'm a :hugefan: of that idea.

> Can puppetserver introspect the puppet gem, or load its gemspec, and install whatever gems it depends on?

Our dev-setup script is just written in bash and can call JRuby, so if all those hard, server-side runtime dependencies are in one place then we can go with that approach too. (I assume we eventually want to do both of these things, but that there's still maybe work to consolidate that gem dependency info in Puppet?)

joshcooper:
Something like this might work:

```
irb(main):001:0> require 'puppet'
=> true
irb(main):010:0> spec.dependencies.each do |dep|
irb(main):011:1* puts "gem install #{dep.name} --version '#{dep.requirements_list.join(',')}'"
irb(main):012:1> end
gem install facter --version '< 4,>= 2.0.1'
gem install hiera --version '< 4,>= 3.2.1'
gem install fast_gettext --version '~> 1.1.2'
gem install locale --version '~> 2.1'
gem install multi_json --version '~> 1.13'
gem install hocon --version '~> 1.0'
gem install net-ssh --version '< 5,>= 3.0'
gem install CFPropertyList --version '~> 2.2'
```

I'm not sure the comma separated version list will work... but maybe just ignore `>=`?

jtappa:
What action should I be taking to follow up on this? Open another PR and edit `dev-setup`? Or let someone on @puppetlabs/puppetserver-maintainers take care of that and have this be blocked until that's in place?

Magisus:
How are we going to handle the case where we need something JRuby-specific, like e.g. JrJackson, if we remove the gem list? I like the idea, just want to make sure we keep that case in mind.

justinstoller:
I think we can put the gem-list.txt back in that case. But that situation, where the gem-list in the puppetserver repo only specifies the puppetserver specific gems, if any, is the ideal. Moving out the puppet specific gems, I think, is a good step towards that.

jtappa:
Updated PR per our meeting today

Magisus:
We will look into removing the gem list as a build resource once all of these gems are provided in puppet-agent's shared directory.

PR #1693, opened 2018-05-09
(SERVER-2205) Allow multiple CRLS in cacrl

justinstoller:
Previously, when loading the cacrl multiple PEMs would be ignored on
first load and when revoking, only the CA's CRL would be written back out
to the cacrl (so any CRL chain would only exist on disk until the first
revocation).

This work assumes when checking certificate statuses we only need to
check the status against the CRL that we manage (ie we're not going to
have to check the status of certs issued by upstream CAs in an
intermediate setup). This was the implicit behavior before, but it is
now codified in the `ssl-utils/pem->ca-crl` helper.

For certificate revocation it loads all the CRLs that could be in a
CRL chain into memory, creates a new CRL for itself with the revoked
cert added and then writes out all the CRLs back to disk.


src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1024,7 +1025,7 @@
     (when (or existing-cert? existing-csr?)
       (let [status (if existing-cert?
                      (if (utils/revoked?
-                           (utils/pem->crl cacrl) (utils/pem->cert cert))
+                           (first (utils/pem->crls cacrl)) (utils/pem->cert cert))
```
Magisus: Will we have to do something similar for `pem->cert`? I seem to remember we thought it could already load multiple, but if that's true, it might be nice to also update the names there to make it clearer.

justinstoller: Yeah, I think we added a `pem->ca-cert` that does the `(first (pem->certs cacert))`. I can add a `pem->ca-crl` or `pem->leaf-crl` that encapsulates this.

justinstoller:
This depends on https://github.com/puppetlabs/jvm-ssl-utils/pull/83 which to get into here will require a new version clj-parent....

justinstoller:
Running this locally (with the pre-suite updated not to use autosigning to work around PUP-8905), the smoke acceptance tests pass:
```
      Test Suite: tests @ 2018-06-05 08:12:37 -0700
       Total Suite Time: 1429.72 seconds
      Average Test Time: 109.98 seconds
              Attempted: 13
                 Passed: 13
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 13
```
The build is:
```
puppetserver justin (mergejustin *$%) ‖ echo $PACKAGE_BUILD_VERSION
6.0.0.master.SNAPSHOT.2018.06.05T0801
```

dev-resources/puppetlabs/puppetserver/certificate_authority_test/pems/bundle/ca/ca_crl.leaf.pem
```diff
@@ -0,0 +1,19 @@
...
+o8Wrca84Q3scgGbKtYMVYhfzpZxf79Zj4VEiVn5dG7ZxtLUyuWGztPE/sW08YS9R
+cEgza16upwQOofnZPL7bS2Yv0yDFzNwIyJKfZZlWaMfIvivGyLDl/2zu7PBsYiTx
+QgY7r37UqZrXhXPUnvnHzoqknd0NwPYO+gH7/z4Vo0DLzBP4vebuulufC47GH9Eq
+2D48nU560sUes1G/HSnDfZaQgH6yypnNEiJ1K6E9hSbFuKMwVo1v7LaqgO+xPhnL
+NzZIz5K+JkVCmEKsQkufnu6ZsHz3qAC3/egZDn6P+z8XtlWfVSdawrfiEzXGQh3m
+vMsrAmylp8ZgwTDaIwXweiEQF/fNYSBf28vjUDfBCWlrUYD3at8x/JBbGBvTwQ0G
+1FS33eVRb4MjKuG4tfS1Xt4MTGfu6ewAqtbemp9fP83h+cGYSqUbdkCqOy9/cYYZ
+pLmC41MAXOvvsE2Tz6PtShAXLBak6BkFNFxGCTWzZ/wjwsCzFpcrw78tZdpfT76E
+ttYGX24dUfGqQIlR9YG8r+LJzpC3/lLa/kLS849Ik3GyaQO+T04P7JWefZzOjA==
+-----END X509 CRL-----
```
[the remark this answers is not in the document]
justinstoller: Updated

PR #1694, opened 2018-05-11
(SERVER-2206) Improve logging configuration

Sharpie:
This commit updates the logback configuration bundled with puppetserver
packages to implement the following improvements:

  - Set scanPeriod="60 seconds" so that settings like log levels can be
toggled without requiring a service restart.

  - Use RFC 3339 compliant timestamps that include the offset from UTC.

This is a port of the enhancements made to pe-puppetserver in 2018.1.0.


gguillotte:
👀

Should I file a followup after this is merged to update https://github.com/puppetlabs/puppetserver/blob/master/documentation/config_file_logbackxml.markdown? Specifically: 

> Scanning can potentially degrade logging performance and is not enabled by default.

This appears to enable scanning by default. Does it still degrade logging performance?

Sharpie:
> This appears to enable scanning by default.

Scanning has been enabled by default since PR #165 added `scan="true"` to v0.1.0 of Puppet Server. An update to the logback library caused this setting to be ignored unless `scanPeriod` was also set.

Sharpie:
Blarg. That would be puppetlabs/ezbake#165, which was then re-applied in puppetlabs/puppetserver#575.

gguillotte:
@Sharpie More context, emphasis mine:

> ## scan and scanPeriod

> Logback supports noticing and reloading configuration changes without requiring a restart, a feature Logback calls scanning. To enable this, set the scan and scanPeriod attributes in the <configuration> element of logback.xml. Scanning can potentially degrade logging performance and is not enabled by default.

```
<configuration scan="true" scanPeriod="60 seconds">
```

> **Due to a bug in Logback, the scanPeriod must be set to a value; setting only scan="true" will not enable configuration scanning.**

Setting `scan="true"` did not enable scanning. Adding the `scanPeriod` enables scanning. This PR adds a default `scanPeriod`. Ergo, this PR enables scanning by default.

It seems like I'm missing something? Were the original docs wrong and scanning has been enabled the entire time?

Sharpie:
Scanning has been enabled since 2015 --- it just broke recently due to the logback bug. Looks like the notes about performance are tagged [TK-426](https://github.com/puppetlabs/puppetserver/commit/a819f892a526d57684c51406916a910fca784e3b), perhaps based on [Joe's comment in TK-426](https://tickets.puppetlabs.com/browse/TK-426?focusedCommentId=413812&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-413812). However, the logback docs referenced by that comment make no mention of performance impacts:

https://logback.qos.ch/manual/configuration.html#autoScan

I'll do some digging on the logback side to see if there was something in their history that changed.

Sharpie:
Looks like the Logback docs changed recently:

https://github.com/qos-ch/logback/commit/5a81da139538#diff-8938c7525b8bc6981d7df30f6d45fe3e

Looks like the gist of the change was that scanning used to be implemented by a Filter where:

> scanning is done "in-thread", that is any time a printing method of logger is invoked.
> ...
> Given that <code>ReconfigureOnChangeFilter</code> is invoked every time <em>any</em> logger is invoked, regardless of logger level, <code>ReconfigureOnChangeFilter</code> is absolutely performance critical.

The new implementation watches for config changes in a background worker thread so the operation is no longer "in-thread" with the actual logging operations, which would be much, much better in terms of performance. Looks like this change went into logback 1.1.7:

https://github.com/qos-ch/logback/commit/1518573

Which happens to be the version that broke `scan="true"` unless `scanPeriod` was set and thus caused TK-426.

Magisus:
For reference, the pe-puppetserver change: https://github.com/puppetlabs/pe-puppetserver/pull/265

Sharpie:
Added "DON'T MERGE" for the moment, going to add two changes:

  - Going to remove `scan` and `scanPeriod` from `request-logging.xml` as scanning actually isn't a thing for the `logback-access` library like it is for `logback-classic`: [LOGBACK-1136](https://jira.qos.ch/browse/LOGBACK-1136). No sense in giving the illusion that something works when it doesn't.

  - Will update the docs to clarify scan performance.

Sharpie:
Alright. This should be good to go now.

documentation/config_file_logbackxml.markdown
```diff
@@ -57,7 +57,7 @@ The user account that owns the Puppet Server process must have write permissions
 
 #### `scan` and `scanPeriod`
 
-Logback supports noticing and reloading configuration changes without requiring a restart, a feature Logback calls **scanning**. To enable this, set the `scan` and `scanPeriod` attributes in the `<configuration>` element of `logback.xml`.  Scanning can potentially degrade logging performance and is not enabled by default.
+Logback supports noticing and reloading configuration changes without requiring a restart, a feature Logback calls **scanning**. To enable this, set the `scan` and `scanPeriod` attributes in the `<configuration>` element of `logback.xml`:
```
justinstoller: Can we mention that this is the default setting for logback.xml but does not apply to request logging?

Sharpie:
Added two notes to the docs:

  - Scanning is enabled by default for the main application log.

  - The scan feature is not implemented by the request log.

gguillotte:
@justinstoller LGTM :shipit:

PR #1695, opened 2018-05-22
(SERVER-2214) Configure gempath for clojure tests

Magisus:


justinstoller:
I think we should put `./install-test-gems` into `dev` and call it from the `dev-setup` script as well. I don't think there's a development workflow where we won't want to run tests.

justinstoller:
I think we should merge this so we can get master's tests passing and we can address any moving around of scripts or refactoring afterwards

PR #1697, opened 2018-05-25
(CPR-553) Migrate puppetserver docker build scripts to the puppetserver repo

underscorgan:
These came from https://github.com/puppetlabs/puppet-in-docker

### DIFF
--- a/docker/puppetserver-standalone/Dockerfile
+++ b/docker/puppetserver-standalone/Dockerfile
@@ -1,0 +1,53 @@
+FROM ubuntu:16.04
+
+ENV PUPPET_SERVER_VERSION="5.3.1" DUMB_INIT_VERSION="1.2.1" UBUNTU_CODENAME="xenial" PUPPETSERVER_JAVA_ARGS="-Xms256m -Xmx256m" PATH=/opt/puppetlabs/server/bin:/opt/puppetlabs/puppet/bin:/opt/puppetlabs/bin:$PATH PUPPET_HEALTHCHECK_ENVIRONMENT="production"
+
+LABEL org.label-schema.maintainer="Puppet Release Team <release@puppet.com>" \
+      org.label-schema.vendor="Puppet" \
+      org.label-schema.url="https://github.com/puppetlabs/puppet-in-docker" \
+      org.label-schema.name="Puppet Server (No PuppetDB)" \
+      org.label-schema.license="Apache-2.0" \
+      org.label-schema.version=$PUPPET_SERVER_VERSION \
+      org.label-schema.vcs-url="https://github.com/puppetlabs/puppet-in-docker" \
+      org.label-schema.vcs-ref="b75674e1fbf52f7821f7900ab22a19f1a10cafdb" \
+      org.label-schema.build-date="2018-05-09T20:11:01Z" \
+      org.label-schema.schema-version="1.0" \
+      org.label-schema.dockerfile="/Dockerfile"
+
+RUN apt-get update && \
+    apt-get install -y wget=1.17.1-1ubuntu1 && \
+    wget https://apt.puppetlabs.com/puppet5-release-"$UBUNTU_CODENAME".deb && \
+    wget https://github.com/Yelp/dumb-init/releases/download/v"$DUMB_INIT_VERSION"/dumb-init_"$DUMB_INIT_VERSION"_amd64.deb && \
+    dpkg -i puppet5-release-"$UBUNTU_CODENAME".deb && \
+    dpkg -i dumb-init_"$DUMB_INIT_VERSION"_amd64.deb && \
+    rm puppet5-release-"$UBUNTU_CODENAME".deb dumb-init_"$DUMB_INIT_VERSION"_amd64.deb && \
+    apt-get update && \
+    apt-get install --no-install-recommends git -y puppetserver="$PUPPET_SERVER_VERSION"-1"$UBUNTU_CODENAME" && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* && \
+    gem install --no-rdoc --no-ri r10k
+
+COPY puppetserver /etc/default/puppetserver
+COPY logback.xml /etc/puppetlabs/puppetserver/
+COPY request-logging.xml /etc/puppetlabs/puppetserver/
+
+RUN puppet config set autosign true --section master
+
+COPY docker-entrypoint.sh /
+
+EXPOSE 8140
+
+ENTRYPOINT ["dumb-init", "/docker-entrypoint.sh"]
+CMD ["foreground" ]
+
+HEALTHCHECK --interval=10s --timeout=10s --retries=90 CMD \
+  curl --fail -H 'Accept: pson' \
+  --resolve 'puppet:8140:127.0.0.1' \
+  --cert   $(puppet config print hostcert) \
+  --key    $(puppet config print hostprivkey) \
+  --cacert $(puppet config print localcacert) \
+  https://puppet:8140/${PUPPET_HEALTHCHECK_ENVIRONMENT}/status/test \
+  |  grep -q '"is_alive":true' \
+  || exit 1
+
+COPY Dockerfile /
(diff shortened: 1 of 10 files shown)


docker/puppetserver-standalone/Dockerfile
```diff
@@ -0,0 +1,53 @@
+FROM ubuntu:16.04
+
+ENV PUPPET_SERVER_VERSION="5.3.1" DUMB_INIT_VERSION="1.2.1" UBUNTU_CODENAME="xenial" PUPPETSERVER_JAVA_ARGS="-Xms256m -Xmx256m" PATH=/opt/puppetlabs/server/bin:/opt/puppetlabs/puppet/bin:/opt/puppetlabs/bin:$PATH PUPPET_HEALTHCHECK_ENVIRONMENT="production"
+
+LABEL org.label-schema.maintainer="Puppet Release Team <release@puppet.com>" \
+      org.label-schema.vendor="Puppet" \
+      org.label-schema.url="https://github.com/puppetlabs/puppet-in-docker" \
```
tvpartytonight: Does this URL need to change to this repo's name, `puppetserver`? Also wondering the same for `vcs-url` and `vcs-ref` below.

docker/puppetserver-standalone/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+      org.label-schema.name="Puppet Server (No PuppetDB)" \
+      org.label-schema.license="Apache-2.0" \
+      org.label-schema.version=$PUPPET_SERVER_VERSION \
+      org.label-schema.vcs-url="https://github.com/puppetlabs/puppet-in-docker" \
+      org.label-schema.vcs-ref="b75674e1fbf52f7821f7900ab22a19f1a10cafdb" \
+      org.label-schema.build-date="2018-05-09T20:11:01Z" \
```
tvpartytonight: Should this be hardcoded here?

docker/puppetserver-standalone/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+      org.label-schema.schema-version="1.0" \
+      org.label-schema.dockerfile="/Dockerfile"
```
tvpartytonight: Why do we copy the `Dockerfile` into the image? I didn't see any reference to this practice [here](https://github.com/projectatomic/ContainerApplicationGenericLabels).

docker/puppetserver-standalone/Dockerfile
```diff
@@ -0,0 +1,53 @@
+FROM ubuntu:16.04
+
+ENV PUPPET_SERVER_VERSION="5.3.1" DUMB_INIT_VERSION="1.2.1" UBUNTU_CODENAME="xenial" PUPPETSERVER_JAVA_ARGS="-Xms256m -Xmx256m" PATH=/opt/puppetlabs/server/bin:/opt/puppetlabs/puppet/bin:/opt/puppetlabs/bin:$PATH PUPPET_HEALTHCHECK_ENVIRONMENT="production"
```
tvpartytonight: Can these be listed out on multiple lines, like how `LABEL` is done?

docker/puppetserver-standalone/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+RUN puppet config set autosign true --section master
+
+COPY docker-entrypoint.sh /
+
+EXPOSE 8140
+
+ENTRYPOINT ["dumb-init", "/docker-entrypoint.sh"]
+CMD ["foreground" ]
+
+HEALTHCHECK --interval=10s --timeout=10s --retries=90 CMD \
```
tvpartytonight: 10 seconds seems a little often to me; what is there a reason for this frequency?

docker/puppetserver-standalone/puppetserver
```diff
@@ -0,0 +1,22 @@
...
+
+# These normally shouldn't need to be edited if using OS packages
+USER="puppet"
+GROUP="puppet"
+INSTALL_DIR="/opt/puppetlabs/server/apps/puppetserver"
+CONFIG="/etc/puppetlabs/puppetserver/conf.d"
+BOOTSTRAP_CONFIG="/etc/puppetlabs/puppetserver/services.d/,/opt/puppetlabs/server/apps/puppetserver/config/services.d/"
+SERVICE_STOP_RETRIES=60
+
+# START_TIMEOUT can be set here to alter the default startup timeout in
```
tvpartytonight: I'm no release engineer, but systemd was a part of xenial, right? Maybe we don't need this comment here...maybe it is just copied and pasted from somewhere else...

docker/puppetserver-standalone/docker-entrypoint.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+chown -R puppet:puppet /etc/puppetlabs/puppet/ssl
+chown -R puppet:puppet /opt/puppetlabs/server/data/puppetserver/
+
+if test -n "${PUPPETDB_SERVER_URLS}" ; then
```
tvpartytonight: Should this be part of the standalone?

docker/puppetserver-standalone/Dockerfile
underscorgan: actually, this looks not too bad to implement, working on an update

underscorgan:
@tvpartytonight updated!

docker/puppetserver-standalone/docker-entrypoint.sh
tvpartytonight: Yeah, I was wondering if the entrypoint should be overwritten by the puppetserver with puppetdb version? Not a blocker at all for this PR, just noticing that it seems to be a part of those concerns.

docker/puppetserver-standalone/puppetserver
tvpartytonight: I'm guessing it's not relevant, and we can take it out? Also not a blocker, but, the less meaningless settings, the better...

underscorgan: updated!

docker/puppetserver-standalone/docker-entrypoint.sh
underscorgan: I think I'd like to leave this as-is for now, this feels cleaner than duplicating the entrypoint script or having a new entrypoint script for puppetserver that adds this check then calls the existing puppetserver-standalone entrypoint.

docker/puppetserver-standalone/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+      org.label-schema.name="Puppet Server (No PuppetDB)" \
+      org.label-schema.license="Apache-2.0" \
+      org.label-schema.version=$PUPPET_SERVER_VERSION \
+      org.label-schema.vcs-url="https://github.com/puppetlabs/puppetserver" \
+      org.label-schema.vcs-ref=$vcs_ref \
+      org.label-schema.build-date=$build_date \
+      org.label-schema.schema-version="1.0" \
+      org.label-schema.dockerfile="/Dockerfile"
+
+RUN apt-get update && \
```
e-gris: This is the common, accepted format for RUN scripts inside Dockerfiles, so my approach is non-standard. With that said, I still think copying in a script from source code and running that has lower maintenance costs and a reduced chance of intermittent bugs. 

COPY srcbin/install_puppet /install_puppet
RUN /install_puppet

docker/puppetserver-standalone/docker-entrypoint.sh
e-gris: Executable scripts should not have a '.sh' extension.

See: https://stackoverflow.com/questions/27813563/what-is-the-bash-file-extension

docker/puppetserver-standalone/docker-entrypoint.sh
```diff
@@ -0,0 +1,16 @@
...
+  sed -i "s@^server_urls.*@server_urls = ${PUPPETDB_SERVER_URLS}@" /etc/puppetlabs/puppet/puppetdb.conf
+fi
+
+# Configure puppet to use a certificate autosign script (if it exists)
+# AUTOSIGN=true|false|path_to_autosign.conf
+if [[ -n "${AUTOSIGN}" ]]; then
```
e-gris: Quotes especially not needed inside '[['

Also "test -n" is used above and "[[ -n " is used here. Probably better to choose one.

docker/puppetserver-standalone/Dockerfile
underscorgan: I don't want to do any major refactoring as part of this initial work, I'm just trying to get the existing automation to a state where it works in the project repos and can have pipelines setup for auto publishing.

docker/puppetserver-standalone/docker-entrypoint.sh
underscorgan: ok, I'll take a look at the test / [[. This is a copy of a script from elsewhere, so there may be some wonkiness.

underscorgan:
@e-gris `.sh` extensions aren't necessary, but we use them fairly consistently throughout this and other projects. I'm going to leave that extension as is.

PR #1698, opened 2018-05-25
(CPR-557) Add script to build puppetserver docker images

underscorgan:
The scripts allow us to set up a pipeline to build these containers on
pipelines.puppet.com.

### DIFF
--- a/docker/ci/build
+++ b/docker/ci/build
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+
+set -eux
+
+BUNDLER_PATH=.bundle/gems
+
+: ===
+: === bundle install to get ready
+: ===
+bundle install --path $BUNDLER_PATH
+
+: ===
+: === pull updated base images
+: ===
+bundle exec puppet-docker update-base-images ubuntu:16.04
+
+# Haven't yet figured out how to run a container with a mounted file in pipelines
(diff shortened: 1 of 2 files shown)


underscorgan:
Depends on #1697

mwaggett:
Out of curiosity, what's the difference between puppetserver and puppetserver-standalone?

underscorgan:
puppetserver-standalone doesn't include puppetdb, but puppetserver does.

docker/ci/build
```diff
@@ -4,6 +4,29 @@ set -eux
...
+
+  : ===
+  : === run linter on the docker files
+  : ===
+   bundle exec puppet-docker local-lint $container_name
+
+  : ===
+  : === build and test $container_name
+  : ===
+  bundle exec puppet-docker build $container_name --repository pcr-internal.puppet.net/release-engineering
```
mwaggett: Should we concerned about community members being able to build?

underscorgan: At least in it's current state, this script is intended entirely for running on CI. It may be valuable at some point to have this all environment variable'd out, but I think for right now I think community members running the `puppet-docker` commands individually is reasonable

tvpartytonight:
LGTM; thanks for the PRs @underscorgan

PR #1703, opened 2018-06-01
(maint) Upgrade cljs tools to re-enable building

justinstoller:
Prior to this our cljs building broke because of an incompatibility
between clojurescript and newer versions of clojure.spec.

This upgrades cljsbuild and clojurescript, excludes cljsbuild from
bringing in its version of clojurescript and sets an explicit dependency
in our ezbake profile.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -28,7 +28,8 @@
       default-heap-size)))
 
 (def figwheel-version "0.3.7")
-(def cljsbuild-version "1.1.5")
+(def cljsbuild-version "1.1.7")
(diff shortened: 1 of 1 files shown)


justinstoller:
So I'm not totally sure why this works vs other configurations. I think I've convinced myself that lein's dependency resolution in at least 2.8 is fairly buggy and have stopped trusting what it's doing. But this is what it takes for me to get this working locally, and I'd love some extra eyes on it.

PR #1704, opened 2018-06-01
(maint) Refactor dev/test gem installation

justinstoller:
This commit moves us to using the much faster gem installation that RE
was using during actual packaging. It also moves the
`install-test-gets` script into the dev directory since it's a
development utility and updates `./dev-setup` to call it so there's only
one script call needed to setup a working dev environment.


justinstoller:
This and https://github.com/puppetlabs/ci-job-configs/pull/4594 should be merged at the same time.

justinstoller:
Still running tests locally, but initially this is working.

To test I think it should be:
`rm -rf ~/.puppetlabs/*`
`lein clean`
`./dev-setup`
`lein repl` and `(go)`
and
`lein test :all`

PR #1706, opened 2018-06-08
(SERVER-2172) Integrate basic puppetserver ca CLI

justinstoller:
:sigh: I think we were just joking about jerks that open 1k+ line PRs...

This really only does a few things:
1. Moves over the "external_ca_support" test from FOSS puppet so our CA testing will be in one place - where we test Server-Agent interaction (which unfortunately includes a bunch of fixture files and a lot of commits).
2. Removes the CA tests that existed in Puppet Server, they were disabled and tested a subset of the functionality that the ones we're importing do.
3. Splits the idea of a "gem-list.txt" into a "jruby-gem-list.txt" and a "mri-gem-list.txt" and updates install-vendored-gems.sh to account for the two types of gems we vendor now.
4. Adds a ca.erb subcommand that calls into the new gem we're vendoring
5. Adds an acceptance test (and helper file) to test the subcommand.

The net new work is in the last three bullets above, which is contained in the last two commits.

I'm happy to squash the commit history for the Puppet acceptance tests that I'm bringing over, and/or move it into its own PR. With the deprecation of the Ruby CA I do feel like these tests should live in the server repo now though. Opening this, hoping to get folks' feedback on the approach.


acceptance/lib/puppet_x/acceptance/pki.rb
```diff
@@ -0,0 +1,222 @@
...
+
+        crl
+      end
+
+      # Creates a self-signed root ca, then signs two node certs, revoking one of them.
+      # Creates an intermediate CA and one node cert off of it.
+      # Creates a second intermediate CA and one node cert off if it.
+      # Creates a leaf CA off of the intermediate CA, then signs two node certs revoking one of them.
+      # Revokes an intermediate CA.
+      # Returns the ca bundle, crl chain, and all the node certs
```
justinstoller: Whoops, this should go away since I changed the certs that are returned from this method.

Magisus:
I'm fine with the commit list, it's good to have that history, and I think we can still find what we need to review this PR.

justinstoller:
I can probably get rid of a few files (like the apache or openssl confs) and by refactoring the tests over to use the pki generation get rid of most of the static fixtures (though I'm hesitant to refactor the old tests until we've seen them ran through CI, or spend more time on it without breaking it up into another ticket  (; ).

resources/ext/cli/ca.erb
```diff
@@ -0,0 +1,30 @@
...
+  STDERR.puts "The import subcommand requires three arguments: private-key, cert-bundle, crl, in that order."
+  exit 1
+end
+
+ca_dir = '/etc/puppetlabs/puppet/ssl/ca'
+FileUtils.mkdir_p ca_dir
+Puppetserver::Ca::Stub.import(private_key, cert_bundle, crl_chain)
+FileUtils.touch(ca_dir + '/inventory.txt')
+File.open(ca_dir + '/serial', 'w+') {|f| f.puts '003' }
+FileUtils.chown_R 'puppet', 'puppet', ca_dir
```
Magisus: Do we ultimately want to move more of the logic in this file into the gem? What was the responsibility split you were envisioning between this wrapper and the gem?

justinstoller: Yeah, I originally just put in the minimum into the gem, but needed to iterate on it for this, so I just put those workarounds in the script for the time being.

I was thinking the script, when we're done with this ticket and we bump the version of the gem we depend on, should have just the minimum to run the library. Something like:
```rb
require 'puppetserver/ca/config'
require 'puppetserver/ca/cli'

user_overrides = Puppetserver::CA::Config.parse(ARGV)
configuration  = Puppetserver::CA::Config.new(user_overrides)

Puppetserver::CA::CLI.run!(config)
```

Magisus: That looks like what I was expecting to see, 👍

justinstoller:
Updated.

I think this is correct, but I also think the correct order of operation is to get Puppet Server master CI green (merge https://github.com/puppetlabs/puppet/pull/6869 and get it promoted into Puppet Server master), then merge https://github.com/puppetlabs/puppetserver/pull/1693, and then this.

That would allow functional testing of this w/o a _very snowflake-y_ dev build.

Magisus:
Agreed on the merge order.

acceptance/lib/puppet_x/acceptance/pki.rb
Magisus: Looks like this still needs to be removed?

resources/ext/cli/ca.erb
```diff
@@ -0,0 +1,30 @@
...
+private_key = ARGV[1]
+cert_bundle = ARGV[2]
+crl_chain   = ARGV[3]
+
+unless private_key && cert_bundle && crl_chain
+  STDERR.puts "The import subcommand requires three arguments: private-key, cert-bundle, crl, in that order."
+  exit 1
+end
+
+ca_dir = '/etc/puppetlabs/puppet/ssl/ca'
```
amysahli: I guess since we are hard-coding this, its not configurable. ^^

resources/ext/cli/ca.erb
```diff
@@ -0,0 +1,30 @@
+#!/opt/puppetlabs/puppet/bin/ruby
+
+require 'fileutils'
+require 'puppetserver/ca/stub'
+
+command = ARGV[0]
+
+unless command == "import"
+  STDERR.puts "  'import' is the only valid subcommand, try:"
+  STDERR.puts "    `puppetserver ca import <key> <bundle> <crl>"
```
amysahli: ^ these are supposed to be provided as paths to the files? (for key, bundle, crl)

resources/ext/cli/ca.erb
justinstoller: It is, but the idea behind this PR was to create a "walking skeleton", "tracer bullet" esque, happy-path feature that proved we could integrated the puppetserver-cli gem and have an end to end demo of the idea.

Since this was originally written I've implemented most of the functionality for this ticket in https://github.com/puppetlabs/puppetserver-ca-cli/pull/1 and after that is merged this file will be re-written to look more like: https://github.com/puppetlabs/puppetserver-ca-cli/pull/1/files#diff-df26495b80ddb61db7362b0be03dd86e

amysahli: ok.  Sorry.  Missing context here

PR #1708, opened 2018-06-12
(PUP-8913) Optionally return the crl based on headers present

amysahli:


src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -56,11 +58,27 @@
       ;; Respond to all CSR validation failures with a 400
       (middleware-utils/plain-response 400 msg))))
 
+      ;;A recipient MUST ignore If-Modified-Since if the request contains an
+      ;; If-None-Match header field;
```
amysahli: ^^ I left these comments here until this can be addressed.  According to the documentation we should ignore this header if "If-None-Match" header is also sent.  I'm unsure if I should implement this logic.  "If-None-Match" header is sent, should I just always send the crl then?  Thoughts?

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -56,11 +58,27 @@
...
-      (rr/content-type "text/plain")))
+  "Always return the crl if no 'if-modified-since' header is provided or if that header is not in correct http-date format.
+  If the header is present and correctly formatted, only return the crl if the master crl is newer than the agent crl."
+  [request {:keys [cacrl]}]
+  (let [agent-crl-last-modified-val (rr/get-header request "If-Modified-Since")
+        agent-crl-last-modified-date-time (try
+                                            (time-format/parse
+                                              (time-format/formatters :rfc822)
+                                              (string/replace agent-crl-last-modified-val #"GMT" "+0000"))
+                                            (catch Exception e
```
amysahli: If there is no header sent, an exception will be raised with the string/replace method.  This seemed the driest implementation I could think of (considering I want to send the crl if there is no header, or if the header is malformed in any way from http-date format).  However, I know catching exceptions is kind of messy, especially since sending a nil value for this header is legit.  Let me know if you'd like me to check specifically for nil value instead of raising an error.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -56,11 +58,27 @@
...
+                                              "unknown"))
+        master-crl-last-modified-date-time (ca/get-crl-last-modified cacrl)]
+    (if (or (= "unknown" agent-crl-last-modified-date-time) (time/after? master-crl-last-modified-date-time agent-crl-last-modified-date-time))
```
amysahli: is "unknown" too hacky here?  I could just return nil and check for the presence of the variable (nil?), but I kind of like "unknown" since its clearer what's going on.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
Magisus: It's hard to say if the "IF" in that comment is the overriding branch, e.g. "If you intend to send this other header, here's how it interacts with 'If-Modified-Since'." Considering we control the only real client for this server, it might be okay as is. But if you do decide to special-case this, I think we should stick to the behavior we have elsewhere of sending the CRL by default.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
Magisus: I do find this a little hard to parse; its unclear without your note where the exception is coming from. I'd be in favor of making this more explicit somehow.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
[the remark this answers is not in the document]
amysahli: done!

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
amysahli: I've reformatted it...not as dry, but easier to understand now I think.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
amysahli: Yeah, if I do add some logic, I think it would just be: if we also get the "If-None-Match" header, then always send the crl (ignore anything in the "If-Modified-Since" header.  Not sure if I need to add anything though.  Any other thoughts?

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -56,11 +58,30 @@
...
+    [request {:keys [cacrl]}]
+    (let [agent-crl-last-modified-val (rr/get-header request "If-Modified-Since")
+          agent-crl-last-modified-date-time (if agent-crl-last-modified-val
+                                              (try
+                                                (time-format/parse
+                                                  (time-format/formatters :rfc822)
+                                                  (string/replace agent-crl-last-modified-val #"GMT" "+0000"))
+                                                (catch Exception e
+                                                  :unknown))
+                                              :unknown)
```
Magisus: I think we should put this processing in its own function, there's a lot going on here with the `if` and the `try`.

Also, is it possible to be more specific about what exception we expect, i.e. is there a more specific `Exception` type that is being thrown by the `string/replace`?

amysahli: Will move to own function.  It will be an exception raised by (time/parse my-custom-formatter).  Lemme see if it makes sense to be more specific there.

Magisus: Yeah, I think so. That would make it clearer what went wrong that we are squashing.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -56,11 +59,34 @@
...
+    (let [agent-crl-last-modified-val (rr/get-header request "If-Modified-Since")
+          agent-crl-last-modified-date-time (if agent-crl-last-modified-val
+                                              (try
+                                                (format-http-date agent-crl-last-modified-val)
+                                                (catch IllegalArgumentException e
+                                                  :invalid-http-date))
+                                              :unknown)
+          master-crl-last-modified-date-time (ca/get-crl-last-modified cacrl)]
+      (if (or (= :unknown agent-crl-last-modified-date-time)
+              (= :invalid-http-date agent-crl-last-modified-date-time)
```
amysahli: so obviously i could just return nil (and check only once for nil) instead of :unknown or :invalid-http-date, but I thought this would be easier to follow/understand.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
amysahli: Ok.  separated it out into its own function and catching that specific error now.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
Magisus: I would be fine with just checking for `nil`, I think, since we don't have any differing behavior between the two cases. Unless you think we might need to add something for one of the cases?

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -56,11 +59,34 @@
...
+    "Always return the crl if no 'if-modified-since' header is provided or if that header is not in correct http-date format.
+    If the header is present and correctly formatted, only return the crl if the master cacrl is newer than the agent crl."
+    [request {:keys [cacrl]}]
+    (let [agent-crl-last-modified-val (rr/get-header request "If-Modified-Since")
+          agent-crl-last-modified-date-time (if agent-crl-last-modified-val
+                                              (try
+                                                (format-http-date agent-crl-last-modified-val)
+                                                (catch IllegalArgumentException e
+                                                  :invalid-http-date))
+                                              :unknown)
```
Magisus: Personally I would move everything starting with the `if` into that function you made, so that the `let` binding here is just naming the result of calling that function, but I'd appreciate the thoughts of someone more experienced with Clojure. Still trying to get a feel for what is considered readable.

If you do that, I would definitely just return nil for the unhappy paths. Then this becomes "Ask for agent modified time, either get it or get nothing, handle those two cases."

amysahli: ^ I agree with second part.  I considered both options.  I was thinking...that if this method were to be called in the future that the error handling desired may be different (so maybe it should exist w/out error handling packaged in).  Also, it would have to accept either a nil or string argument, so I couldn't use schema/str.  And not sure what clojure thinks about this ambiguity.

pcarlisle: i think moving it makes sense

it could go either way, but handle-get-certificate-revocation-list is long enough that it makes sense to extract that

you could always have two functions, one that does the parsing and one that does error handling, but someone who needs to distinguish that could always do that later

amysahli: ^ Updated.  Thanks for the great advice everyone.

melissa:
jenkins please test this

Magisus:
@melissa we don't have any PR testing for Puppetserver, unless you are working on adding it?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1136,6 +1137,14 @@
   [cacrl :- schema/Str]
   (slurp cacrl))
 
+(schema/defn ^:always-validate
+  get-crl-last-modified :- org.joda.time.DateTime
```
justinstoller: nit: once we've imported this above we can, and idiomatically do, refer to it as just `DateTime`.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -56,11 +59,33 @@
       ;; Respond to all CSR validation failures with a 400
       (middleware-utils/plain-response 400 msg))))
 
-(defn handle-get-certificate-revocation-list
-  [{:keys [cacrl]}]
-  (-> (ca/get-certificate-revocation-list cacrl)
-      (rr/response)
-      (rr/content-type "text/plain")))
+  (schema/defn format-http-date :- (schema/maybe org.joda.time.DateTime)
```
justinstoller: ditto

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -56,11 +59,33 @@
...
+    "Formats an http-date into joda time.  Returns nil for malformed or nil http-dates"
+    [http-date :- (schema/maybe schema/Str)]
+    (if http-date
```
justinstoller: If you don't want to return a value for the else block, or the else is `nil` you can just use the `when` macro instead of `if`.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -56,11 +59,33 @@
...
+      (try
+        (time-format/parse
+          (time-format/formatters :rfc822)
+          (string/replace http-date #"GMT" "+0000"))
+        (catch IllegalArgumentException e
+          nil))
+      nil))
+
+  (defn handle-get-certificate-revocation-list
+    "Always return the crl if no 'if-modified-since' header is provided or if that header is not in correct http-date format.
```
justinstoller: Can you wrap these lines a little earlier. I think most docs in this file/repo try to keep the wrap around at 80 ish characters.

justinstoller: Might want to capitalize 'if-modified-since' as we do below

justinstoller: Instead of `in correct http-date format` we also might just want to call out `rfc822`

justinstoller: Wow, those were super nit-picky of me, sorry.

amysahli: The only thing I didnt change was http-date to rfc822, since our http date has to end in GMT, which is not the case for rfc822 i believe. (rfc822 <pubDate>Wed, 02 Oct 2002 15:00:00 +0200</pubDate> is OK, for example, but wouldnt be a valid http-date since it needs to end in GMT :)

justinstoller: That's fair

PR #1709, opened 2018-06-12
(SERVER-2231) Publish infra-CRL - preliminary changes for early review

jsane:
NOTE - This is not meant to be a pull request. DO NOT MERGE. 

I am just hoping to get an early feedback here around a couple of things and thought it would be easier to get it with some code in place than via/over just documentation or email/hipchat discussions. 

- Design: Does it look reasonable for what it is trying to achieve? The idea is simple - if any of the master certificate is getting revoked we add it to an infra-CRL in addition to the regular/existing crl. For that we maintain a list of all master certificate serial numbers in a file generated during init time. I have avoided use of globals with 'state' being maintained in files. Hopefully there should not be any performance concerns with reading the file with compile master serial numbers during every revocation as they should not be very often and the file would be relatively small. 
- Conventions/styling: for some of the names used for functions, config settings etc. if they look odd and not. Placement of the newly introduced methods - if they seem out of place. I have added some comments at places just for review purposes at this stage. Can remove them if they seem rather excessive.
- Opens: I guessed which methods/functions (endpoints used for servicing crls over http/network) that would require changes to return the infra-CRL once AFTER this is actually working and tested adequately. 
- Misc: Any Clojure novice-ness in the implementation that may have better alternatives.  And/or any unaccounted for conditions/situations. 

I am yet to add defaults for the newly introduced config settings in puppet (lib/puppet/defaults.rb) which I will be doing soon. I am thinking of maintaining the two new files for maintaining list of compile masters and their certificate serial numbers in the cadir ($ssldir/ca). 
The other thing remaining is the monitoring the list of compile masters and restarting CA anytime that list changes (by sending a sighup to ??? trapperkeeper service).  

Assuming it is ok to target master branch.

thanks


jsane:
Am aware of travis failures and think adding the new defaults to puppet should address some of them.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -762,6 +823,11 @@
                (i18n/trs ":localcacert ({0}) could not be found and no file at :cacert ({1}) to copy it from"
                     localcacert cacert))))))
 
+;; Open/question - Is this the method that would be invoked when an agent requests
+;; for a CRL over HTTP/REST? If so once we have the infra-crl functioning then perhaps
+;; this would need to be updated
```
justinstoller: No, what this does is copy the cacrl from its location to the location that the agent/master will use it (localcacrl). Its used on start up, but also we register a file watcher for the cacrl, so when it is updated the master will copy the file over and refresh its ssl-context.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1129,6 +1195,9 @@
         {:outcome :not-found
          :message msg }))))
 
+;; Open/question - Is this the method that would be invoked when an agent requests
+;; for a CRL over HTTP/REST? If so once we have the infra-crl functioning then perhaps
```
justinstoller: yes

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -505,14 +553,27 @@
...
+    (utils/crl->pem! cacrl (:cacrl ca-settings))
+    (utils/crl->pem! infra-crl (:infra-crl ca-settings)))
+
+    ;; Generate a file containing serial numbers of the configured compiler master certificates, if any.
+    ;; Open - Would this be a right place to generate this as we want the compile master serials list to
+    ;; get generated everytime the compile masters are updated. And looking at this function this appears
+    ;; more like a one-time only method (generates ca keys, cert.
+    ;; If so, what would be an appropriate place for doing the below which can get called as many times as
+    ;; needed.
+    (generate-compile-master-serials (:compile-masters ca-settings) (:compile-master-serials ca-settings))
```
justinstoller: you may want to look into registering a filesystem watcher on the compile-masters file and running this when it changes.

justinstoller: See the certificate-authority-service for an example of doing so with the crl.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -475,6 +486,43 @@
...
+    (line-seq (io/reader cm-file)))
+
+(schema/defn generate-compile-master-serials
+  "Given a file containing names of all compile master nodes
+   it will create a file containing serial numbers of their certificates.
+   Both files will have names/serial numbers on separate lines.
+   In case the compile masters file does not exist it will create an 
+   empty file of compile master serial numbers."
+   [compile-masters-file :- schema/Str
+    compile-masters-serials-file :- schema/Str
```
justinstoller: What do we want to do when the serials-file already exists? Do we want to overwrite it? Do we only want to load the certs that aren't already in it (may cause issues with the `allow-duplicate-certnames` setting)? Do we want to load all the certs but only add serials that do not already exist in it? What happens if we can't find a cert for an infra node?

justinstoller:
I'm a little worried that loading the certnames to treat as infra nodes from a file won't be robust enough. If the node is removed from the node group first there might be situations when the infra-crl isn't updated when the cert is finally revoked.

However, everything that I try to think of (specifying during revocation that this is a node you want to put in the infra-crl, or attaching metadata to infra certs that if present when revoking we add the node to the infra-crl) don't seem like obvious wins.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -68,7 +68,16 @@
    :ruby-load-path           [schema/Str]
    :gem-path                 schema/Str
    :signeddir                schema/Str
-   :serial                   schema/Str})
+   :serial                   schema/Str
+   ;; This would point to a file containing a list of compile master node names (FQDNs)including MoM
+   ;; This would be either provisioned by PE or created by user (input configuration)
+   ;; Would like it to be optional (though the below type qualifier may not be capturing the intent)
+   :compile-masters          (schema/maybe schema/Str)
```
justinstoller: newer settings usually call out that they are a path in their name

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1263,7 +1332,17 @@
                               (.getPublicKey (utils/pem->ca-cert cacert cakey))
                               serial)]
     (utils/crl->pem! new-crl cacrl)
-    (log/debug (i18n/trs "Revoked {0} certificate with serial {1}" subject serial))))
+    (log/debug (i18n/trs "Revoked {0} certificate with serial {1}" subject serial))
+   
+    ;; Publish infra-crl if a compile-master is getting revoked.
+    (if (seq-contains? (read-compile-masters compile-master-serials) serial)
```
justinstoller: `if`s without an else clause are idiomatically written with the `when` macro.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
jsane: Yes the serials-file would be regenerated (overwritten) every time compile masters change as it is supposed to maintain the 'current set of compile masters'. Allow-duplicate-certnames refers to whether CA will issue multiple certificates with same subject names so hopefully should not matter. And I hope/assume current interfaces (UI or cli) do not allow for revoking same certificate more than once. Else this could be an issue for existing CRLs as well depending on how CA deals with same serial getting revoked more than once. 
The case of not being able to find a certificate for an infra node would be an issue. I think there is a backup option of figuring that assuming inventory.txt exists and is reliable.

jsane:
Yeah there don't seem to be a good way out of the scenario you pointed out - if it is removed from the node group first and revoked later it would not get into the infra-crl. Though I would imagine anyone doing that do not want that node to act as a compile master and hopefully will have also adjusted agent side configs or load balancers to take it out. Else we can put in a clear security warning in the documentation about doing such things.  These should not be routine or common occurrences. In any case agents were not doing any revocation checking till. Am not trying to use that as an excuse to not anything about it. Just that I do not see any good option to address it programmatically (now). 

I do think however if the certificates had some role information stamped in them as an extension it would definitely help. Then we won't even need to maintain the compile masters list etc. as just by looking at the certificate itself one can do the needful.

jsane:
@justinstoller Whenever you get a chance could you pl take a look at the latest version I pushed and my responses to your comments. Agree there are few corner cases but if you add notes to the documentation about such possibilities (plus how this feature works so once they know the mechanics whoever can intervene manually to rectify/avoid the situation). 
I think one main open is where to maintain defaults for the new config settings that are to live in certificate-authority. Sorry for the bother but I cannot seem to figure travis the failures for this PR.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
justinstoller: > And I hope/assume current interfaces (UI or cli) do not allow for revoking same certificate more than once.

The current implementation doesn't prevent that behavior. Every call to revoke a cert will add an entry to the CRL.
See lack of checking whether or not the cert has already been revoked in:
  * [puppetserver](https://github.com/puppetlabs/puppetserver/blob/master/src/clj/puppetlabs/puppetserver/certificate_authority.clj#L1275)
  * [jvm-ssl-utils](https://github.com/puppetlabs/jvm-ssl-utils/blob/master/src/java/com/puppetlabs/ssl_utils/SSLUtils.java#L346)
  * [puppet](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/ssl/certificate_revocation_list.rb#L49)

justinstoller:
> These should not be routine or common occurrences. 

I do think there are customers that use Puppet in cloud environments and routinely spin up and down compile masters. I don't think we need to block on making sure this works great for that use case, but we should make sure that we aren't doing harm to them. You may want to talk to folks internally that work with AWS deployments about the scope of ephemeral compile masters.

justinstoller:
This travis error:
```
lein test puppetlabs.general-puppet.general-puppet-int-test
2018-06-15 20:53:59,043 ERROR [async-dispatch-2] [p.t.internal] Error during service init!!!
java.lang.IllegalArgumentException: No implementation of method: :as-file of protocol: #'clojure.java.io/Coercions found for class: clojure.lang.PersistentVector
	at clojure.core$_cache_protocol_fn.invokeStatic(core_deftype.clj:583)
	at clojure.java.io$fn__10838$G__10820__10843.invoke(io.clj:35)
	at clojure.java.io$file.invokeStatic(io.clj:424)
	at clojure.java.io$file.invoke(io.clj:418)
	at clojure.lang.AFn.applyToHelper(AFn.java:154)
	at clojure.lang.RestFn.applyTo(RestFn.java:132)
	at clojure.core$apply.invokeStatic(core.clj:659)
	at clojure.core$apply.invoke(core.clj:652)
	at me.raynes.fs$file.invokeStatic(fs.clj:56)
	at me.raynes.fs$file.doInvoke(fs.clj:50)
	at clojure.lang.RestFn.invoke(RestFn.java:410)
	at me.raynes.fs$parent.invokeStatic(fs.clj:239)
	at me.raynes.fs$parent.invoke(fs.clj:236)
	at puppetlabs.puppetserver.certificate_authority$create_parent_directories_BANG_.invokeStatic(certificate_authority.clj:303)
```
is because we're trying to make the parent directories we need when initializing our ssl files. The list of ca related directories is created with `settings->cadir-paths` which blacklists settings in `CaSettings` that shouldn't be considered ca related directories. Adding items to CaSettings will require commensurate changes in `settings->cadir-paths` or changing `settings->cadir-paths` to be a whitelist instead of a blacklist.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -68,7 +68,21 @@
    :ruby-load-path           [schema/Str]
    :gem-path                 schema/Str
    :signeddir                schema/Str
-   :serial                   schema/Str})
+   :serial                   schema/Str
+   ;; This would be list of compile master node names (FQDNs) including MoM
+   ;; It would either be provisioned by PE or created & maintained by user (input configuration)
+   ;; OPEN: where do we get to specify defaults for the new configuration settings
```
justinstoller: Probably in `config->ca-settings`

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -68,7 +68,21 @@
...
+   ;; that will reside in ca.conf within certificate-authority?
+   ;; Would the installer be able to populate specifically the compile-masters with 
+   ;; current host name? 
+   :compile-masters          [schema/Str]    
+   ;; This would point to a file containing a list of serial numbers of compile master certificates
+   ;; corresponding to the nodes from the above file. This would be re-generated everytime the
+   ;; compile-masters list is updated.
+   ;; Q - is there a way to maintain this as an in-memory list?
```
justinstoller: What do you mean?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1129,6 +1207,7 @@
         {:outcome :not-found
          :message msg }))))
 
+;; TODO - Once we have the infra-crl functioning then this would need to be updated
```
justinstoller: What will this return? Will it always return the infra-crl?

justinstoller: How will this work with "intermediate CAs"

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -475,6 +494,51 @@
...
+  "Given a list of compile master nodes it will create a file containing 
+   serial numbers of their certificates (listed on separate lines).
+   Compile masters is expected have at least one entry (MoM)"
+   [cm-nodes :- [schema/Str]
+    compile-masters-serials-file :- schema/Str
+    signeddir :- schema/Str]
+
+   ;; TODO: We may want to account for the possibility of compile masters to
+   ;; be empty (by mistake, accident). In which case this can use the name
+   ;; of the current node as the MoM as a fallback
```
justinstoller: Is this feature something that will be opted into (I assume not)?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -475,6 +494,51 @@
...
+  (comment
+   (with-open [wtr (clojure.java.io/writer compile-masters-serials-file)]
+     (doseq [cm-node cm-nodes]
+       (let [cm-serial  (-> (path-to-cert signeddir cm-node)
+                         (utils/pem->cert)
+                         (utils/get-serial))]
+          (log/debug (format "%s: %3d\n" "Serial " cm-serial))
+          ;; These writes are failing with an error of "No matching method found: write for class java.io.BufferedWriter"
+          ;; The filename is valid and writable. Same code works in a test app
+          ;; (.write wtr cm-serial)
```
justinstoller: It might not be able to  infer what cm-serial will be, and so can't chose the right method. Try adding a [type hint](https://clojure.org/reference/java_interop#typehints). Something like `(.write wtr ^String cm-serial)`.

justinstoller:
Thanks, Jayant. I think I still have some questions about this would work in a cloud deployment but it looks like its coming along. Hopefully, there's enough feedback there to unblock you.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
jsane: Currently during every revocation this serials file needs to be read. And while it should not have any perf implications I was thinking if there was a way to maintain the serials in memory at all times then that could be avoided.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
jsane: Yes it would always return infra-crl - on the assumption that other compile masters do NOT use this to get the crl and only pure agent nodes would use it. They should be using the full CRL. I now wonder we will have a problem in case of a split install with console services, puppetdb - maybe not since agents do not talk to console. In some cases they actually might like when obtaining rbac token using a cli running from any agent node. But it takes in actual creds and does not rely on client cert so that should be ok. I will think more though. 

Correct me but here's my understanding of our "intermediate CA" support: PuppetCA is not a root but is issued by some other CA with an arbitrary level of hierarchy above. Obviously all entities within puppet infrastructure would need access to all intermediate issuers and their revocation status. PuppetCA, however, would only be responsible for all puppet certificates. So the infra-crl should be able to correctly report on status of all compile masters and MoM. 
That brings up a question (am suspecting an issue/gap presently with our intermediate CA support). All our certificate revocations need to be able to figure+get the revocation status for all issuers above PuppetCA. That may need to be provisioned in some manner either we adding an endpoint that serves these other CRLs (which puppetserver maintains uptodate copies of) or they be able to fetch CRLs themselves based on info in the intermediate CA certs (crlDistributionPoint) or do OCSP checking (AuthorityInfoAccess). The later two are possibilities depending on whatever PKI setup we become part of.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
jsane: Am thinking not. Though we could make it configurable if there might be issues with cloud deployments using ephemeral compile masters and there are concerns around infra-crl not accurately reflecting revocation status of all compile masters. But then they will have to deal with potentially large CRLs (since even agents would be ephemeral and thus large number of revocations). 

As I think about it, I feel it should not be too difficult to stamp an extension indicating the role of the node. Though need to think if it can also be done reliably in a pure foss deployment (with no classifier etc.)

src/clj/puppetlabs/puppetserver/certificate_authority.clj
justinstoller: Right now we're adding the ability for the cacrl to contain multiple crls in it. The `cacrl` setting would contain the crl that the CA adds and then also the crls for all of the authorities above it. The agent would then download a single file with multiple crls in it and respect each of them for full chain checking.
See https://github.com/puppetlabs/puppetserver/pull/1693 for that work in serverland.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
justinstoller: So, I think the custom extension denoting the puppet infra role would be valuable and itself not a lot of work, but it would require coordination across our products and making a policy about how to create certs for puppet infra, which we don't have currently and have always wavered on. That level of decisions and coordinating probably make it a significantly harder thing to implement then this approach.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
jsane: ok cool though that assumes the PKI can integrate with must support CRLs. Thats fair expectation I guess. I have not looked thru the above PR (will do) but who would be ensuring the combined CRL is kept up to date with all crls for CA's above us?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
jsane: Agree. It occurs to me the scheme I was thinking (of automating the extension stamping transparently) actually might not work. It breaks down in case a node get repurposed for some other role but continues to use the original certificate. But if you were alluding to some more explicit manner of how infra certificates are created  (that can avoid above situations) then that should work sans the coordination challenges.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
jsane: I tried setting some defaults there but then getting tripped up in the 
"(schema/defn ^:always-validate 
  config->ca-settings :- CaSettings ....) " schema validation function. I have tried many different ways to convince it but am sure I am not doing it right. The case I was trying handle was if the new config settings are missing in the certificate-authority group in the config file then it should use the defaults. I do not want to make those config settings as optional either. Or may be if I mark them as optional in CaSettings defn that might do it - but I need to check if it would use the defaults in that case.  Am going to try that now. If that does not work I will push an update with comments around my various attempts and may be you can take a look sometime.

jsane: ok I think I managed to make it use some defaults for the new config params for infraCRL in case they are not specified in any config files (ca.conf or puppetserver.conf). 
Had a question on the default values for some of the paths. I tried using values like "/tmp/crl/...." etc but it errors out in travis. Think I can get away with "/tmp/..." - I would expect that to exist everywhere but obviously that won't be a proper default. 
How do I specify a default that will work in either cases where it can/should use "~/.puppetlabs/etc/..." or "/etc/puppetlabs/puppet/ssl/...". 
Also question on - how you do test Clojure CA interfaces when running server from sources?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
justinstoller: Our assumption is that the crls are only for providing a chain of trust to the our CA and that the rest of the puppet infra is signed by that, so you'd only need to update it in the case of migrating/re-keying your CA, which should be very infrequent.

We're also working on a CLI tool to help with importing a chain when bootstrapping intermediate CAs which might be able to mitigate some of that pain.

jsane:
@justinstoller 
I think I need some help. 
1. With travis failures. I think I know what is happening but can't seem to figure where some of the definitions like expected schemas (that now need to be updated with new fields in CaSettings map) or what it passes to "initialize-master-ssl" method during the initialize-master-ssl!-test-with-keylength-settings test.  On the later (initialize-master-ssl! method) I verified that with the code in place all invocations of this method from within CA and master service seem to work fine (.i.e the test is passing it a map of config settings w/o the new keys). 
2. I want to do the following for defaults of infra-crl-path and compile-masters-serials-path (set-ca-config-defaults method). Get the cacrl, extract out the folder portion and append the appropriate file names to make defaults for the above. 
thx

justinstoller:
We talked in the office about the current blockers.

Let me know if you run into anything else or want to pair on writing an integration test. I think you need to rebase and resolve the conflicts to get travis testing your PR again.

jsane:
I rebased. 
Huh..I don't understand whats causing the latest travis failure? The updated CA settings in ca_testutils.clj are there; wonder if there is some other place which needs to know about the updated schema for CA settings (with the new test for if-modified-since). 
I also noticed that there was a typo in one of the config params in ca_testutils.clj (compile-master-serials-path was misspelled as compile-masters-serials-path) but strangely it was passing travis yesterday. I tried correcting it but the above/current travis is not related to it. Temporarily I have left misspelled config name.

test/unit/puppetlabs/services/ca/ca_testutils.clj
```diff
@@ -27,7 +27,10 @@
         :privatekeydir (str ssldir "/private_keys")
         :requestdir     (str ssldir "/certificate_requests")
         :csr-attributes (str confdir "/csr_attributes.yaml")
-        :keylength      512})))
+        :keylength      512
+        :compile-masters ([])
```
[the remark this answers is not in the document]
jsane: ok. I can/will remove them but they are not causing the travis failure (they are just redundant) as it passed travis tests yesterday with them.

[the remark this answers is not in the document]
justinstoller: These are where your travis tests are failing. `()` is being treated like a function invocation, it's try to call `[]` (the `PersistantVector` literal) with zero args. If you want nested array-like structures try `[[]]`, if you specifically want a list containing a vector you need to quote the list `'([])`.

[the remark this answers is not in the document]
justinstoller: whoops, late send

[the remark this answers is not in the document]
jsane: Hmm... you are probably right but that is strange. It did pass yesterday with those parens. The other weird part was the misspelled config params were not getting flagged.

jsane:
Curious why this is deemed as blocker?

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
```diff
@@ -349,3 +350,54 @@
...
+  (testing "Revoking compile master certificate"
+    (logutils/with-test-logging
+      (bootstrap/with-puppetserver-running-with-mock-jrubies
+       "JRuby mocking is safe here because all of the requests are to the CA
+       endpoints, which are implemented in Clojure."
+        app
+        ;; Q: Does one need to make up other config params like for eg. infra-crl-path
+        ;; or can we assume anything unspecified will exist with appropriate defaults
+        ;; Would the below override whatever defaults?
+        {:certificate-authority {:compile-masters ["dummy-compile-master"]}}
```
justinstoller: `bootstrap/with-puppetserver-running-with-mock-jrubies` will [use](https://jenkins-master-prod-1.delivery.puppetlabs.net/login?from=%2Fme%2Fconfigure) the configuration at `/dev/puppetserver.conf` with this used as an override for what's in there. Any additional overriding/defaults would be handled in the code you're writing.

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
```diff
@@ -349,3 +350,54 @@
...
+        (testing "should update infrastructure CRL"
+          (let [ca-cert (bootstrap/get-ca-cert-for-running-server)
+                subject "dummy-compile-master"
+                ;; In one of the other unit tests I noticed the localhost cert is
+                ;; being revoked. If that does not affect things we could do that instead of 
+                ;; creating a new one. But wondering it actually may have impact on other 
+                ;; tests since this is integration
```
justinstoller: The `bootstrap/with-puppetserver-running-with-mock-jrubies` will spin up a new instance of puppetserver, you can give it a custom confdir in your configuration which will move the default cadir as well. You can create fixtures in that cadir that will not affect other tests. See [this CA test](https://github.com/puppetlabs/puppetserver/blob/master/test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj#L185-L203) for an example of setting the confdir to [this test fixture directory](https://github.com/puppetlabs/puppetserver/tree/master/dev-resources/puppetlabs/services/certificate_authority/certificate_authority_int_test/ca_true_test/master/conf) that the test then manipulates the contents of the csr directory.

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
```diff
@@ -349,3 +350,54 @@
...
+                cm-cert (bootstrap/get-cert-signed-by-ca-for-running-server
+                             ca-cert
+                             subject)
+                 ;; TODO - Extract serial number of abvoe cert. Is there an easy way
+                 ;; to do it here or can we make assumption on the expected serial number
```
justinstoller: https://github.com/puppetlabs/puppetserver/blob/master/src/clj/puppetlabs/puppetserver/certificate_authority.clj#L1263

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
```diff
@@ -349,3 +350,54 @@
...
+                 ;;
+
+                 ;; Q: Do we need to do something to explicitly moved the certificate
+                 ;; to CA's signed folder
```
justinstoller: The CA writes a copy of the cert out to the signed dir when it signs it.

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
```diff
@@ -349,3 +350,54 @@
...
+                                                {:body (str "{\"desired_state\": \""
+                                                            action
+                                                            "\"}")
+                                                 :headers {"content-type"
+                                                           "application/json"}}))) ]
+            (testing "Infra CRL should contain the revoked cert"
+               (let [revoke-response (cert-status-request "revoked")]
+                 ;; If the revocation was successful we want to check the infra CRL contains above certificate
+                 ;; Can I just change the below to have an AND check as I have attempted below
+                 ;; Q: How do I access CaSettings or just certificate-authority settings here to look up infra-crl-path?
```
justinstoller: Couple different levels of testing here. Since you're going through the HTTP API I would expect that you'd go through it as well to validate that the correct CRL was returned when trying to download it. Otherwise you could also call the CA function `revoke-existing-cert!` that's called by the HTTP API and test at a lower level. You can also pass in a custom path to the infra-crl, and serial-file right? If you're explicitly setting paths, I don't think it's inappropriate to then read those paths at the end of the test.

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
```diff
@@ -349,3 +350,54 @@
...
+                                                 :headers {"content-type"
+                                                           "application/json"}}))) ]
+            (testing "Infra CRL should contain the revoked cert"
+               (let [revoke-response (cert-status-request "revoked")]
+                 ;; If the revocation was successful we want to check the infra CRL contains above certificate
+                 ;; Can I just change the below to have an AND check as I have attempted below
+                 ;; Q: How do I access CaSettings or just certificate-authority settings here to look up infra-crl-path?  
+                 (is (and (= 204 (:status revoke-response)) (utils/revoked? (utils/pem->crl (:infra-crl-path settings)) cert)))))
+          )))))
+    ;; TODO/Open - Should we also have an other way round test like if revoke a non compile master
```
justinstoller: Yeah, I don't think it would be bad in an integration test to revoke a compiler and non-compiler and then check both crls to ensure the correct content. More granular than that and I'd move to a lower level of testing.

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
jsane: I think I will just issue a new cert and so should not have to move the configuration. A new cert and infra CRL will not affect anything.

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
jsane: Actually the q was a bit leftover as I found the util/revoked? method which seems to fit the bill (nice thing it takes in the crl file and cert).

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
jsane: Yes I know (when a cert is issued over proper interfaces) but was not sure the method being used in test would do it.

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
jsane: yeah perhaps we should test the downloading part. And checking that the revoked cert shows up in the CRL will ensure 'correctness' (which am trying to do in the current version of the test). Btw, I have not yet changed the place where the http get crl API to return the infra (that will then allow testing the downloads correct crl). We need a new config setting to control whether to return full crl or infra. Should that be part of this same PR?   

Coming back to my question of - how do I access the CaSettings configuration? Basically I want to access :infra-crl-path from "settings" (what would the settings be in this case)?

justinstoller: That value isn't readily available that I know of

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
jsane: Hmm.. looks like this particular method - bootstrap/get-cert-signed-by-ca-for-running-server - used to sign the cert does not place the cert in signed folder. 
Got this travis error: 
  actual: java.io.FileNotFoundException: ./target/master-conf/ssl/ca/signed/dummy-compile-master.pem (No such file or directory)

jsane:
@justinstoller 
Looks like am not understanding how this test setup actually works. Meaning whether one can assume whether it would have created "localhost" certificate when puppetserver starts. That does not seem to be the case when running the integration test because attempting to revoke it errors claiming it does not know about it as seen in the last two/three travis failures. 

Also per my comment in the recent few pushes, certificates listed in compile-masters config must exist during CA service initialization (thats when it attempts to generate a serials list from them). So the approach I was trying (and what we were thinking) earlier of generating a certificate as part of the test and designating that as a compile master would not work. Unless we (can) update the compile masters config after creating the new certificate and then forcing a reload of the service (as would be the case in production). 

I have manually verified that if a compile master gets revoked, the infraCRL gets updated correctly. If /when you have time can you pl take a look and/or actually update the test code itself to an approach you think should work? thx

jsane:
In the latest rev of the integration test, I can get it to generate a cert in first run, have it revoked in the second run. However the infraCRL wont record the revoked cert when it should have. And I know from manual testing that it should have. 
The compile-masters config setting, that i specified in the test, seems to be the suspect (though it looks correct) because the compile master serials file is empty. 
Is there a way to capture the log messages, generated from within CA, when running the test? They won't show in console or anywhere under the specified master config directory?

jsane:
This should now be close for a review. If the integration test needs to be doing any additional checks  pl let me know. Apart any other inputs - think one of the util type methods I introduced in the certificate_authority_int_test.clj may need to be moved somewhere appropriate. 
Otherwise if it looks ok I will squash the commits.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -69,7 +69,19 @@
    :ruby-load-path           [schema/Str]
    :gem-path                 schema/Str
    :signeddir                schema/Str
-   :serial                   schema/Str})
+   :serial                   schema/Str
+   ;; This would be list of compile master node names (FQDNs) including MoM
+   ;; It would either be provisioned by PE or created & maintained by user (input configuration)
+   :compile-masters          [schema/Str]
```
justinstoller: I'd maybe change this to be "infra-certs" or "infra-certnames" to keep consistent with our usage of "infra" elsewhere (ie it's a "infra-crl" not a "compiler-masters-crl".

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -69,7 +69,19 @@
    :ruby-load-path           [schema/Str]
    :gem-path                 schema/Str
    :signeddir                schema/Str
-   :serial                   schema/Str})
+   :serial                   schema/Str
+   ;; This would be list of compile master node names (FQDNs) including MoM
```
justinstoller: I believe these need to be certnames, not node names, or FQDNs (though we often combine those concepts).

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -69,7 +69,19 @@
...
+   ;; It would either be provisioned by PE or created & maintained by user (input configuration)
+   :compile-masters          [schema/Str]    
+   ;; This would point to a file containing a list of serial numbers of compile master certificates
```
justinstoller: If you're going to add comments, and a little bit is good, I would make them more declarative, like "path to file containing serials found for certs listed in `compile-masters`".

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -69,7 +69,19 @@
...
+   ;; corresponding to the nodes from the above file. This would be re-generated everytime the
+   ;; compile-masters list is updated.
+   :compile-master-serials-path   schema/Str
+   ;; Path to Infrastructure CRL file containing master certificates
+   :infra-crl-path                schema/Str
+   ;; Option to continue using full CRL instead of infrastructure CRL if desired
+   ;; Infrastructure CRL would be enabled by default.
+   :disable-infra-crl         schema/Bool })
```
justinstoller: I think when we do settings we should keep them positive, eg `enable-infra-crl`. Sometimes negative settings confuse folks.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -262,7 +274,16 @@
           :keylength
           :manage-internal-file-permissions
           :ruby-load-path
-          :gem-path))
+          :gem-path
+          :compile-masters
+          ;; We plan to use one of the existing directories for maintaining the
+          ;; compile master serials and infra-crl so removing them from here
```
justinstoller: I don't follow.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -393,6 +416,21 @@
...
 
+(schema/defn set-ca-config-defaults
+  "Sets defaults for any of the missing config parameters for infra CRL"
+  [infraCrlConfig
+   cadir]
+  (log/debug (str (i18n/trs "set-ca-config-defaults received input:")
+                  "\n"
+                  (ks/pprint-to-string infraCrlConfig)))
+  ;; TODO - We want to use the current node/certificate as the default in here instead of starting empty
+  (-> (assoc infraCrlConfig :compile-masters (if (nil? (:compile-masters infraCrlConfig)) [] (:compile-masters infraCrlConfig)))
```
justinstoller: I think this would be better with a `merge` rather than `assoc` and a ternary conditional if the value is set.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -393,6 +416,21 @@
       (catch CRLException e
         e))))
 
+(schema/defn set-ca-config-defaults
+  "Sets defaults for any of the missing config parameters for infra CRL"
+  [infraCrlConfig
```
justinstoller: variables should be kebab-case rather than camelCase

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -470,6 +508,32 @@
...
+(schema/defn generate-compile-master-serials
+  "Given a list of compile master nodes it will create a file containing 
+   serial numbers of their certificates (listed on separate lines).
+   Compile masters is expected have at least one entry (MoM)"
+   [cm-nodes :- [schema/Str]
+    compile-masters-serials-file :- schema/Str
+    signeddir :- schema/Str]
+
+   (log/debug (format "%s\n%s" (i18n/trs "generate-compile-master invoked ") compile-masters-serials-file))
+   (with-open [wtr (clojure.java.io/writer compile-masters-serials-file)]
```
justinstoller: this should just be `io` rather than `clojure.java.io`

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -470,6 +508,32 @@
...
+
+(schema/defn generate-compile-master-serials
+  "Given a list of compile master nodes it will create a file containing 
+   serial numbers of their certificates (listed on separate lines).
+   Compile masters is expected have at least one entry (MoM)"
+   [cm-nodes :- [schema/Str]
+    compile-masters-serials-file :- schema/Str
+    signeddir :- schema/Str]
+
+   (log/debug (format "%s\n%s" (i18n/trs "generate-compile-master invoked ") compile-masters-serials-file))
```
justinstoller: This is probably too high of detail for logging, or if we really want to keep it in it would be a `trace`, I think.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -470,6 +508,32 @@
...
+   (with-open [wtr (clojure.java.io/writer compile-masters-serials-file)]
+     (doseq [cm-node cm-nodes]
+       (try
+         (let [cm-serial  (-> (path-to-cert signeddir cm-node)
+                           (utils/pem->cert)
+                           (utils/get-serial))]
+            (log/debug (format "%s: %3d\n" "Serial " cm-serial))
+            (.write wtr (str cm-serial))
+            (.newLine wtr))
+         (catch Exception ex
```
justinstoller: We should catch something more specific than Exception

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -740,7 +809,11 @@
            hostprivkey)))
 
     :else
-    (generate-master-ssl-files! settings certname ca-settings)))
+    (generate-master-ssl-files! settings certname ca-settings))
+  (log/debug (format "%s\n%s" (i18n/trs "initialize-master-ssl invoked. CA Settings: ") (ks/pprint-to-string ca-settings)))
+  ;; Generate a file containing serial numbers of the configured compiler master certificates, if any.
+  (generate-compile-master-serials (:compile-masters ca-settings) (:compile-master-serials-path ca-settings) (:signeddir ca-settings))
```
justinstoller: You may want to just pass in `ca-settings`, or the subset of keys that you want, and destructure the map in the argument list for `generate-compile-master-serials` rather than use positional arguments.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -898,12 +971,19 @@
...
    service return a map with of all the CA settings."
   [{:keys [puppetserver jruby-puppet certificate-authority]}]
-  (-> (select-keys puppetserver (keys CaSettings))
-      (assoc :ruby-load-path (:ruby-load-path jruby-puppet))
-      (assoc :gem-path (str/join (System/getProperty "path.separator")
-                                 (:gem-path jruby-puppet)))
-      (assoc :access-control (select-keys certificate-authority
-                                          [:certificate-status]))))
+    (let [certificate-authority-defaults (set-ca-config-defaults certificate-authority (fs/parent (:cacrl puppetserver)))]
+      (log/debug (str (i18n/trs "config->ca-settings value of certificate-authority-defaults :")
```
justinstoller: In general most of these log messages seem better suited for understanding the flow of the code when you're writing a feature, rather than debugging a production issue.

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
```diff
@@ -349,3 +352,148 @@
...
+
+;; This may need to be moved to one of the utils file/ns
+(defn get-ca-signed-cert
+  [ca-cert
+   certname
+   serial
+   master-conf-dir]
+  (let [ca-public-key  (.getPublicKey ca-cert)
+        ca-private-key (ssl-utils/pem->private-key
+                        (str master-conf-dir "/ssl/ca/ca_key.pem"))
```
justinstoller: We load the private-key based on the confdir but have the cert passed in, why is that? Should we either have the caller pass in the CA objects or load them from the path relative to confdir?

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
```diff
@@ -349,3 +352,148 @@
...
+          )))
+
+        (bootstrap/with-puppetserver-running-with-mock-jrubies
+         "JRuby mocking is safe here because all of the requests are to the CA
+         endpots, which are implemented in Clojure."
+          app
+          {:jruby-puppet {:master-conf-dir master-conf-dir}
+           :certificate-authority {:compile-masters ["compile-master"]
+                                   :disable-infra-crl true}}
+          (testing "Verify full CRL is returned "
```
justinstoller: How are these last two tests different from the `"Verify correct CRL is returned depending on disable-infra-crl"` test above?

test/unit/puppetlabs/services/ca/ca_testutils.clj
```diff
@@ -50,7 +50,11 @@
    :signeddir             (str cadir "/signed")
    :serial                (str cadir "/serial")
    :ruby-load-path        jruby-testutils/ruby-load-path
-   :gem-path              jruby-testutils/gem-path})
+   :gem-path              jruby-testutils/gem-path
+   :compile-masters       []
+   :compile-master-serials-path (str cadir "/cm-serials.txt")
+   :infra-crl-path (str cadir "/infra-crl.pem")
```
justinstoller: spacing's off

dev-resources/puppetlabs/services/ca/certificate_authority_core_test/master/conf/ssl/ca/infra-crl.pem
```diff
@@ -0,0 +1,16 @@
+-----BEGIN X509 CRL-----
```
justinstoller: What does this do and how does it interact with the tests?

dev/puppetserver.conf
```diff
@@ -168,3 +168,12 @@ authorization: {
         }
     ]
 }
+
+certificate-authority: {
+   # This will house following new configuration settings 
+   # for infrastructure CRL
+   # Enable them to specify any non-default values
+   # - infra-crl-path:
```
justinstoller: would probably put `""` here to match up with the empty array you place after `compile-masters`

justinstoller:
I'm still wary about putting the infra nodes' certnames in a config file that requires the server to be HUPed on update. I'd rather see us take a path to the certname file and put a filesystem watcher on it, reloading the values as it changes.

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
```diff
@@ -349,3 +352,148 @@
...
+            (let [ca-cert (ssl-utils/pem->cert (str master-conf-dir "/ssl/ca/ca_crt.pem"))
+                  cm-cert (get-ca-signed-cert ca-cert subject 23 master-conf-dir)
+                  node-cert (get-ca-signed-cert ca-cert node-subject 24 master-conf-dir)])))
+        (bootstrap/with-puppetserver-running-with-mock-jrubies
+         "JRuby mocking is safe here because all of the requests are to the CA
+         endpots, which are implemented in Clojure."
+          app
+          {:jruby-puppet {:master-conf-dir master-conf-dir}
+           :certificate-authority {:compile-masters ["compile-master"]
+                                   :disable-infra-crl false}}   ;; Got to be able to use subject here
```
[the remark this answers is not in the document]
jsane: I wanted to be able to use the "subject" local var defined above (that has same value as "compile-master")

src/clj/puppetlabs/puppetserver/certificate_authority.clj
jsane: agree current impl is inelegant and verbose. I don't quite understand what you are suggesting and/or how but will give it a try.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
jsane: It depends on what infrastructure constitutes like if it also includes things like puppetdb and console nodes in case of an split install then it would be misleading. I just wanted to be specific to only mean compile masters. But no heartburn if you prefer it like that. 
By that token then infra-crl is misleading as well...will change it I guess (in a bit).

test/unit/puppetlabs/services/ca/ca_testutils.clj
jsane: Did you mean that for infra-crl-path? Whats the convention when keyname is long like in case of compile-master-serials-path?

dev-resources/puppetlabs/services/ca/certificate_authority_core_test/master/conf/ssl/ca/infra-crl.pem
jsane: It is dummy (copy of existing crl.pem in same folder). I stumbled upon the need for the file to exist when testing the certificate_authority_core_test.clj (crl-endpoint-test) when passing infra-crl-path to handle-get-certificate-revocation-list. crl-endpoint-test by itself does not use/need it. Since the unit test do not run the CA service that file is not created as it happens when running integration tests.

dev/puppetserver.conf
jsane: Think the uncommented compile-masters accidentally got checked in. I can just comment them all out; the configuration parameter names are self explanatory on the expected values unless we want to provide some sample example values (while still leaving them commented).

jsane:
That could be a concern if someone is spinning up/down compile masters very frequently. Originally I had started with that approach but Joshua P had expressed objections to it. 
If it is serious enough I will get back to it after I wrap some of the other things on my plate. Else we can put this PR on hold and I will complete it whenever I can.

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
jsane: The last two verify correct crl is returned over http endpoints (more of an end to end tests). The ones above them directly invoke handle-get-certificate-revocation-list. This is per the multiple levels of tests you had suggested.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
jsane: So I ran into an issue. I removed the compile-master-serials-path and infra-crl-path from settings->cadir-paths. Then launched puppetserver while specifying these two configs to point to some non-existing paths. Would have expected the directories, from the paths specified in values of these two to get created during init (by create-parent-directories).
But it doesn't seem to. 
If I leave the above two as defaults then they end up re-using the same folder as that of cacrl (which does get created by create-parent-directories) and things work out then.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
jsane: Not sure if this is what you meant in my latest push.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
jsane: Figured this

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -740,7 +790,10 @@
            hostprivkey)))
 
     :else
-    (generate-master-ssl-files! settings certname ca-settings)))
+    (generate-master-ssl-files! settings certname ca-settings))
+  ;; Generate a file containing serial numbers of the configured compiler master certificates, if any.
```
justinstoller: I don't think you need this comment line

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -740,7 +790,10 @@
...
+  (generate-compile-master-serials ca-settings)
```
justinstoller: This line should aligned with the one above it

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -740,7 +790,10 @@
...
+)
```
justinstoller: This closing parens should be on the line above

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -393,6 +408,19 @@
       (catch CRLException e
         e))))
 
+(schema/defn set-ca-config-defaults
+  "Sets defaults for any of the missing config parameters for infra CRL"
+  [infra-crl-config
+   cadir]
+  ;; TODO - We want to use the current node/certificate as the default in here instead of starting empty
+  ;; Open - Is there a way to eliminate these temps below? Cannot quite use an ->
+  (let [t1 (if (nil? (:compile-masters infra-crl-config)) (merge infra-crl-config {:compile-masters []}) infra-crl-config)
```
justinstoller: When I mentioned using merge before I was thinking something more along the lines of (I think you were using the threading macro before, but creating an intermediate variable would work too):
```clj
(let [defaults {:compile-masters []
                :compile-masters-serials-path (str cadir "/cm-serials.txt")
                :infra-crl-path (str cadir "/infra-crl.pem")
                :enable-infra-crl true}]
  (-> infra-crl-config
    (merge defaults)
    (dissoc :certificate-status)))
```

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -69,7 +69,18 @@
    :ruby-load-path           [schema/Str]
    :gem-path                 schema/Str
    :signeddir                schema/Str
-   :serial                   schema/Str})
+   :serial                   schema/Str
+   ;; List of for all compile master certificates including MoM
+   ;; provisioned by PE or user in case of FOSS
+   :compile-masters          [schema/Str]
```
justinstoller: still think changing our terminology from "compile-master" to "infra" or "infra-node" would be beneficial in the long run (eg. we include the MoM already).

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1267,7 +1326,18 @@
...
+    (log/debug (i18n/trs "Revoked {0} certificate with serial {1}" subject serial))
+   
+    ;; Publish infra-crl if a compile-master is getting revoked.
+    (when (and (fs/exists? compile-master-serials-path) (seq-contains? (read-compile-masters compile-master-serials-path) (str serial)))
+        (let [new-infra-crl (utils/revoke (utils/pem->crl infra-crl-path)
+                            (utils/pem->private-key cakey)
+                            (.getPublicKey (utils/pem->ca-cert cacert cakey))
+                            serial)]
+          (utils/crl->pem! new-infra-crl infra-crl-path)
+        (log/info (i18n/trs "Compile master certificate being revoked; publishing updated infrastructure CRL"))))
```
justinstoller: This should be on the same column as the line above.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
jsane: I wanted it to be outside of the above conditional and execute every time since we don't know if the infra-nodes changed since the last time.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
jsane: Thanks! I knew there had to be a better way. Though the order of args to merge matters. I changed it to have the intended behavior of explicit values over riding defaults.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
jsane: Yeah. regret all the nits.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -470,13 +497,33 @@
...
+            (log/warn (i18n/trs (str "Failed to find/load certificate for compile master: " cm-node)))))
+    )))
+
 (schema/defn generate-ssl-files!
   "Given the CA settings, generate and write to disk all of the necessary
   SSL files for the CA. Any existing files will be replaced."
   [ca-settings :- CaSettings]
-  (log/debug (str (i18n/trs "Initializing SSL for the CA; settings:")
-                  "\n"
-                  (ks/pprint-to-string ca-settings)))
```
Magisus: Is there a reason this log message was removed? (Sorry if I missed a comment about it, lots of review on this)

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -689,9 +741,6 @@
   [settings :- MasterSettings
    certname :- schema/Str
    ca-settings :- CaSettings]
-  (log/debug (format "%s\n%s"
-                     (i18n/trs "Initializing SSL for the Master; settings:")
-                     (ks/pprint-to-string settings)))
```
Magisus: Same question here. I've definitely used these for debugging before.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
jsane: @justinstoller 
Am looking into adding a file system watcher for infra nodes file and was not clear on some things.  Questions on code https://github.com/puppetlabs/puppetserver/blob/master/src/clj/puppetlabs/services/ca/certificate_authority_service.clj#L46-56
- about the (when (not= ca-crl-file host-crl-file...) check. This seems to copy content of cacrl, anytime it changes, to local/host crl file. Am guessing I don't need to maintain two files like this. 
- the code seems to register a watcher for "changed-path" (?) type of event on ca-crl-file or may be it is saying "ca-crl-file" was the changed file within the watched directory (its parent). I am interested in "file got updated" type of event for infra-nodes. Do I also need to watch the directory housing infra-nodes file and check if it was the infra-nodes file that changed then invoke the callback?

jsane:
@justinstoller 
About maintaining the list of compile masters in a separate file that would reside in cadir folder. I talked to Joshua Partlow. So PE configuration cannot use contents of any of the config files it configures/manages as the input => PE module wont be able to see the updated value of cadir, if set, in ca.conf which could then be used to provision the compile masters file with the expectation that it would reside in cadir. One can specify parameters to classes in PE to be able to override any defaults but thats about it. A user can manually update cadir value in ca.conf at anytime which puppetserver will see the updated value but PE wont know about it. 
Implications and workaround options: 
Compile-masters list file would not move to new folder if a user updates cadir in ca.conf etc. It would stay in same default location or whatever the overridden value specified by the PE level parameter. 
Workaround option - Hardcode the location of the compile masters file (like is the case with ca.conf) and not insist it stay in same place as cadir (should it move). It can start off in the default folder of '/etc/puppetlabs/puppetserver/ca'.

justinstoller:
I think it's fine to assume that the file will live in '/etc/puppetlabs/puppetserver/ca' with the rest of the CA files.

jsane:
I have the file monitoring working but its ugly. Due to the dependency on PE the file path to infra nodes list needs to be hard coded (needs to be same across both). So we don't want to expose it as a user configurable setting. But without a default value for it during ca service init, we cannot set a file watcher on it. I have a way of setting defaults for the new config settings for this feature but it happens late. There needs to be a way to set CA defaults somewhat similar to they being set in puppet (defaults.rb) which are preset/available before CA service initializes. Given that we are weaning off puppet dependence is there any other alternate mechanism? Of course the requirement is to be able to set the default in code and not in user visible config file - at least for this particular config.  Lets chat tomorrow.

dev/puppetserver.conf
```diff
@@ -168,3 +168,13 @@ authorization: {
...
     ]
 }
+
+certificate-authority: {
+   # This will house following new configuration settings 
+   # for infrastructure CRL
+   # Enable them to specify any non-default values
+
+   # infra-crl-path: ""
+   # infra-node-serials-path: ""
```
justinstoller: remove these two paths from being configurable.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -179,7 +199,7 @@
 
 (def required-ca-files
   "The set of SSL related files that are required on the CA."
-  #{:cacert :cacrl :cakey :cert-inventory :serial})
+  #{:cacert :cacrl :cakey :cert-inventory :serial :infra-node-serials-path :infra-crl-path})
```
justinstoller: We talked about splitting this between the infra-crl related files and not and then in the caller conditionally combining if `enable-infra-crl` is true

jsane:
@justinstoller pl let know if/when you have sometime this week to look into the travis failure. Would like to be ready for 6.0 (not sure when it is but guess close). thanks,

justinstoller:
Hey, Jayant! I'm actually currently working on this. I think I have the tests working correctly now and am looking at squashing and rebasing onto master. Does that sound alright?

jsane:
Great! Thats good news. Am itching to know what was the issue; will sync up sometime. ok to squash commits (and seems to need rebase). thanks

justinstoller:
This work was merged in https://github.com/puppetlabs/puppetserver/pull/1760 !!!

PR #1710, opened 2018-06-12
(SERVER-2215) Upgrade cljs workflows to use fighweel-sidecar

lazartea:
Currently, we have a dependency on leiningen as a library so that we can run ``lein figwheel`` to access the cljs-dashboard-widgets. This commit uses figwheel-sidecar instead to start the figwheel server without calling the lein task. This allows us to remove dependencies on both leiningen and logback-classic.


lazartea:
It's also worth noting that part of SERVER-2215 mentioned making the figwheel served dashboard interact with a running puppet server. While this was done within in the cljs-dashboard-widgets repo, I didn't do it here because sending an XMLHttpRequest to remote servers that aren't on the same origin causes an error. Adding the headers to fix that error was straightforward in cljs-dashboard-widgets, but it looked like it would be a lot more convoluted in puppetserver.

PR #1711, opened 2018-06-13
(maint) switch manifest namespace

underscorgan:


distelli-manifest.yml
```diff
@@ -1,4 +1,4 @@
-release-engineering/puppetserver:
+pe-and-platform/puppetserver:
```
tvpartytonight: For future reference, what does this correlate with in Puppet Pipelines? Is it group names?

er0ck: the first part (before the `/`) is a context, similar to AWS contexts

(here is a screenshot of mine, including the "SLV" context, which is kinda like a group)
screen shot 2018-06-18 at 3 43 15 pm

 it does not correspond to a repo, directly.  if that's what you mean...

following through on tony's question...
anyone know whom is in control of `pe-and-platform` and can we get added to it?

PR #1715, opened 2018-06-14
(SERVER-2212) Merge 5.3.x master

tvpartytonight:


Magisus:
Oh jeez. We need to do these more often...

project.clj
```diff
@@ -1,4 +1,6 @@
 (def ps-version "6.0.0-master-SNAPSHOT")
+(def jruby-1_7-version "1.7.27-1")
+(def jruby-9k-version "9.1.16.0-1")
```
justinstoller: We don't want to bring these up to master

Magisus:
I don't see the submodule or the the puppet-agent ref being updated by this, which is correct. @justinstoller you didn't mean "don't bring the commits", did you? Because that isn't really possible...

justinstoller:
No, I didn't mean "don't bring in the commits". Previously, at least I thought, the PR included changing the actual submodule pin, which we don't want.

Magisus:
Does this look okay to you now then @justinstoller ? If so, mind changing your review?

PR #1717, opened 2018-06-21
(SERVER-2169) Support running under Java 9 and 10

austb:


test/unit/puppetlabs/services/jruby/jruby_puppet_testutils.clj
```diff
@@ -367,6 +368,7 @@ create-mock-pool-instance :- JRubyInstance
   ([services config]
    (add-mock-jruby-pool-manager-service services config create-mock-jruby-puppet))
   ([services config mock-jruby-puppet-fn]
+   (#'ks-classpath/ensure-modifiable-classloader)
```
austb: How do we feel about this hacky method of calling that function?

.travis.yml
```diff
@@ -1,8 +1,9 @@
 language: clojure
-lein: 2.7.1
+lein: 2.8.1
 dist: trusty
 jdk:
   - oraclejdk8
+  - oraclejdk10
```
austb: Do we want to test on both version or just 8/10 (and if so, which one)?

test/unit/puppetlabs/services/jruby/jruby_puppet_testutils.clj
amysahli: Discussed this with Austin-  Would be better to call ```add-classpath``` method since its not private (and since ensure-modifiable-classloader is only called in the code from w/in add-classpath), but looks like that's not a possible solution here.  @austb is that right?

.travis.yml
amysahli: @austb why not  - oraclejdk9 ?  Would that be a different ticket?  Was this the thing you mentioned about tags?  Ticket commit message mentions Java 9.

austb: 9 and 10 don't seem to be very different, 10 was sort of a "trial run" for a faster release cadence for the JDK. Also the end goal is Java 11 support, so I didn't think it was worth testing.

amysahli: @pcarlisle I noticed you had 7, 8, and 9 here:  https://github.com/puppetlabs/clj-kitchensink/pull/99/files
Do you have any preference?

test/unit/puppetlabs/services/jruby/jruby_puppet_testutils.clj
justinstoller: Is making sure the classloader doesn't need to be modifiable part of the work to support Java 11? Or is something like `ensure-modifiable-classloader` the long term solution when we're on Java 11?

pcarlisle: As far as I know this is okay for 11.

.travis.yml
pcarlisle: i think 8 and 10 is fine

test/unit/puppetlabs/services/jruby/jruby_puppet_testutils.clj
justinstoller: If this is the longer term fix, should we release ks with it as a public fn?

austb: Yes, it seems like something people may want to do without calling add-classpath (where it is called by ks currently).

pcarlisle: nobody should ever call this, what we're doing is awful

justinstoller: So, this approach is fine for Java 11, but it's awful and nobody should ever do it?

Should we inline what ks is doing in this test so we don't have to commit to it as a public api there, or should we investigate a less awful way to implement it (do you know off the top of your head if that's even an option)?.

austb: I am unclear what the test actually tests, it seems to mock everything (and one of the things it mocks is probably what would call ks)

pcarlisle: > So, this approach is fine for Java 11, but it's awful and nobody should ever do it?

It's fine in that it's not going to break, and given the testing we have here it's an okay way to get it working, but anyone who really needs to do things with classloaders like that outside of our public API should be figuring out what they need and doing it explicitly. I'd really like to not maintain that function as part of our API. I think inlining the implementation is reasonable.

pcarlisle: @austb Which test do you mean? this is just a utility function called from a bunch of tests

austb: Ah yes, @pcarlisle probably should have clarified that this line is necessary to prevent an error in the test [facter-jar-loaded-during-init](https://github.com/puppetlabs/puppetserver/blob/master/test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj#L19-L36)

justinstoller: That test doesn't seem _super_ valuable. If I'm understanding it correctly the behavior of the function is tested in [jruby-puppet-core-test/add-facter-to-classpath-test](https://github.com/puppetlabs/puppetserver/blob/master/test/unit/puppetlabs/services/jruby/jruby_puppet_core_test.clj#L53-L89) and `facter-jar-loaded-during-init` is really only testing that we call the `add-facter-jar-to-system-classloader!` function during the [service initialization](https://github.com/puppetlabs/puppetserver/blob/master/src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj#L36). I could see either way, but I'd probably delete the test myself.

PR #1719, opened 2018-06-28
(PE-24670) Add default authorization rule for status service endpoint

amysahli:
The status service endpoint previously did not check the puppetserver
auth.conf file.  Now that it does, we will give it a default rule to
allow unauthenticated requests (default behavior).

### DIFF
--- a/acceptance/README.md
+++ b/acceptance/README.md
@@ -89,7 +89,7 @@ discussion in this document assumes the content of
 To use one of the pre-defined config files, specify the path to it
 using `beaker`'s `--config` option. For example:
 
-    bundle exec beaker --config acceptance/beaker/jenkins/redhat7-64m-64a.cfg ...
+    bundle exec beaker --config acceptance/config/beaker/jenkins/redhat7-64m-64a.cfg ...
 
 ### Using beaker-hostgenerator
 
--- a/acceptance/suites/tests/authorization/default_rules.rb
+++ b/acceptance/suites/tests/authorization/default_rules.rb
@@ -135,6 +135,11 @@ with_puppet_running_on(master, {}) do
     assert_allowed
   end
 
+  step 'status service endpoint' do
+    curl_unauthenticated('/status/v1/services')
+    assert_allowed
+  end
+
   step 'static file content endpoint' do
     # We'd actually need to perform a commit and use its code-id in order to
     # get back a 200, but we know that a 400 means we got past authorization
--- a/ezbake/config/conf.d/auth.conf
+++ b/ezbake/config/conf.d/auth.conf
@@ -46,6 +46,17 @@ authorization: {
             name: "puppetlabs csr"
         },
         {
+            # Allow unauthenticated access to the status service endpoint
+            match-request: {
+                path: "/status/v1/services"
+                type: path
+                method: get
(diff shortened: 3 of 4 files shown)


justinstoller:
Needs to be updated to depend on a yet-to-be-released version of tk-status. Are we updating the dependency directly, or are we using clj-parent? Should we be targeting 5.3.x? Don't know if anyone else in @puppetlabs/puppetserver-maintainers has opinions on that? (maybe @Sharpie, as well)

Sharpie:
Given `/status` output from Puppet Server is something we want to restrict access to, it seems likely we'll want to do it for PuppeDB, Console Services, and Orchestration Services. `clj-parent` seems like the right spot for the update as it is shared by all TK services.

justinstoller:
I agree we should put this into clj-parent 2.x and put that into master and promote it into Johnson.

I think the next question should be: Since this is targeting a release in Irving, do we release this as part of Puppet Server FOSS 5.3.x, or do we just update it in pe-puppet-server-extensions for Irving? (and if we do put it into FOSS, should we put it into clj-parent 1.7.x or set it as an explicit version override in Server's project.clj).

Considering where Puppet Server is in the 5.x lifecycle (maintenance bug/fix) I would normally say to either skip doing the FOSS release or use an explicit override, but considering where we are in the Irving lifecycle I would say that we should make a clj-parent 1.7.x release and do a FOSS Server release (seems early in a two year support cycle to start accumulating one-offs/diffs between FOSS and PE).

Sharpie:
Putting it into clj-parent 1.7.x and merging up makes the most sense to me. Given this behavior is opt-in, I don't see any particular need to split it off to pe-puppetserver.

justinstoller:
So, it sounds like the order of operations is:
  1. Merge https://github.com/puppetlabs/trapperkeeper-status/pull/89 and do a 1.1.0 release of tk-status with the changes
  2. Update and release new versions of clj-parent at 1.7.8 and 2.0.2 with the new version of tk-status.
  3. Update this PR to target 5.3.x and have it pull in the new clj-parent 1.7.8.
  4. Merge up to master from 5.3.x and over write the 1.7.8 version of clj-parent with the 2.0.2 version.
  5. The two versions of Puppet Server (5.3.x & master) will auto promote into Irving and Johnson respectively. pe-pse and pe-ps will need to have their versions of clj-parent bumped prior to the autopromotion.

nb: 3-5 should happen around the same time so the PRs for all of them should be prepped together.

justinstoller:
Let me know if you want any help targeting this at 5.3.x

justinstoller:
I built this into a package at [5.3.4.SNAPSHOT.2018.07.09T1405](http://builds.delivery.puppetlabs.net/puppetserver/5.3.4.SNAPSHOT.2018.07.09T1405/) and ran it against the smoke suite (that includes the test changed).
```
      Test Suite: tests @ 2018-07-09 14:15:07 -0700
       Total Suite Time: 1228.77 seconds
      Average Test Time: 81.92 seconds
              Attempted: 15
                 Passed: 15
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 15
```

PR #1725, opened 2018-07-12
(SERVER-2237) Use `python3` when `python` is not available

Magisus:
Ubuntu 18.04 does not ship with a `python` executable, instead
providing only `python3`. This commit updates our test that uses Python
to check for the `python` executable and use `python3` instead if it
cannot be found.

### DIFF
--- a/acceptance/suites/tests/code_commands/code_scripts.rb
+++ b/acceptance/suites/tests/code_commands/code_scripts.rb
@@ -202,6 +202,13 @@ step 'SETUP: Configure the code-id script' do
   reload_server
 end
 
+step 'SETUP: Find Python executable'
+  if on(master, 'which python', :acceptable_exit_codes => [0, 1]).exit_code == 1
+    python_bin = 'python3'
+  else
+    python_bin = 'python'
+  end
+
 step 'Get the current code-id'
   current_code_id=on(master, '/opt/puppetlabs/server/apps/puppetserver/code-id-command_script.sh production').stdout.chomp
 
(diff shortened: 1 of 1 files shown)


Magisus:
Please let me know if there's a better way to accomplish this.

Magisus:
I ran this locally and the test now passes.

acceptance/suites/tests/code_commands/code_scripts.rb
```diff
@@ -210,6 +217,6 @@ class { '::ntp': }
   key       ="/etc/puppetlabs/puppet/ssl/private_keys/#{fqdn}.pem"
   hostcert  ="/etc/puppetlabs/puppet/ssl/certs/#{fqdn}.pem"
   auth_str  ="--cacert #{cacert} --key #{key} --cert #{hostcert}"
-  result=on(master, "curl --silent #{auth_str} https://#{fqdn}:8140/puppet/v3/catalog/#{fqdn}?environment=production | python -m json.tool").stdout
+  result=on(master, "curl --silent #{auth_str} https://#{fqdn}:8140/puppet/v3/catalog/#{fqdn}?environment=production | #{python_bin} -m json.tool").stdout
```
justinstoller: Is this line needed? I thought `python -m json.tool` was used to pretty print a string of valid json. I would assume that `JSON.parse` could handle an un-pretty printed string of JSON though...

It doesn't look like we print the result anywhere humans would have to read it either.

Sharpie: As I recall, the result of `on( )` calls are logged by beaker when `--debug` is on.

So, this is likely here for those moments when a human is trawling through the output trying to figure out why the test failed.

justinstoller: fair

PR #1726, opened 2018-07-13
(SERVER-2237) Use `python3` when `python` is not available

Magisus:
Ubuntu 18.04 does not ship with a `python` executable, instead
providing only `python3`. This commit updates our test that uses Python
to check for the `python` executable and use `python3` instead if it
cannot be found.

This is a cherry pick of https://github.com/puppetlabs/puppetserver/commit/b8cf75a12cc678ed90bd387974424c9a29eb66f1.

### DIFF
--- a/acceptance/suites/tests/code_commands/code_scripts.rb
+++ b/acceptance/suites/tests/code_commands/code_scripts.rb
@@ -202,6 +202,13 @@ step 'SETUP: Configure the code-id script' do
   reload_server
 end
 
+step 'SETUP: Find Python executable'
+  if on(master, 'which python', :acceptable_exit_codes => [0, 1]).exit_code == 1
+    python_bin = 'python3'
+  else
+    python_bin = 'python'
+  end
+
 step 'Get the current code-id'
   current_code_id=on(master, '/opt/puppetlabs/server/apps/puppetserver/code-id-command_script.sh production').stdout.chomp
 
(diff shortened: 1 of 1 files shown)


acceptance/suites/tests/code_commands/code_scripts.rb
```diff
@@ -202,6 +202,13 @@ class { '::ntp': }
   reload_server
 end
 
+step 'SETUP: Find Python executable'
+  if on(master, 'which python', :acceptable_exit_codes => [0, 1]).exit_code == 1
```
amysahli: Do you need to have the "acceptable_exit_code" bit in here for it to work?

Magisus: I think beaker automatically crashes the run if anything exits non-zero when it hasn't been flagged. Hence all the wrappers around Puppet calls you see that say "0 or 2", because with detailed exit codes getting 2 is usually fine too.

Magisus: Which is interesting, because failing to find python returns 127, not 1. Maybe more testing is needed.

justinstoller: `which` should return 0 or 1 depending on if the executable is found, `bash` should return 127 if an executable isn't found, is that now what you're seeing?

Magisus: Ah, yes. Sorry for the confusion. I was second guessing myself after looking at the failure output again.

Magisus:
Re-verified that this works correctly (there's a chance I was running the wrong test yesterday, so I wanted to make sure).

PR #1728, opened 2018-07-13
(SERVER-2225) Move cadir outside of SSL dir

lazartea:
Currently, puppetserver's CA directory defaults to /etc/puppetlabs/puppetserver/ssl. This commit changest the default location to /etc/puppetlabs/puppetserver/ca, separating the CA and SSL dirs. The new setting is specified in the puppetserver.conf in the (no longer deprecated) certificate-authority setting.

There's still some work to be done here (I think the dev-resources directory should be updated to follow our new structure), and we need to start thinking about deprecation and CLI integration. For now, though, I was hoping to get some eyes on this at its current state.


src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -109,6 +110,55 @@
...
+
+(defn default-serial
+  [cadir]
+  (str cadir "/serial"))
+
+(schema/defn ^:always-validate
+initialize-ca-config
+  [ca-data-structure]
+  (-> ca-data-structure
+      (update-in [:cadir] #(or % default-cadir))
```
justinstoller: You can, and should if you keep this structure, just use `update` here and remove the `[]` from the key.

You _might_ also want to try setting the `cadir` to the desired value and then creating a map of defaults and merging them?

You can also use argument destructuring to get the value from a map or a default [see](https://clojure.org/guides/destructuring#_keyword_arguments).

This is something I would write:
```clj
(defn init-ca
  [{:keys [cadir] :or {cadir default-cadir} :as ca-data}]
  (let [defaults {:cadir cadir
                  :cacert (default-cacert cadir)
                  :cacrl (default-cacrl cadir)
                  ;; ...
                 }]
    (merge defaults ca-data)))
```

But I think the main thing here is readability, I think lifting the values to a map at the top of the function brings them into focus and removes as much ceremony around setting them as possible. I'd say experiment with the approaches until you find something you think someone w/o a lot of clojure experience will easily understand in a year or two's time (I could see, based on that, making an argument against destructuring).

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
```diff
@@ -64,13 +64,13 @@
 
   (retrieve-ca-cert!
     [this localcacert]
-    (ca/retrieve-ca-cert! (get-in-config [:puppetserver :cacert])
-                          localcacert))
+    (let [settings (ca/config->ca-settings (get-config))]
```
justinstoller: I think `ca/config->ca-settings` is rather heavyweight and none of its values should change during the service's lifetime. I think it would be good to compute it once during `init` and save it into the `context` then grab the service's context in these protocol implementations.

justinstoller: See start and end lines [here](https://github.com/puppetlabs/puppetserver/blob/master/src/clj/puppetlabs/services/config/puppet_server_config_service.clj#L32-L37) for how we do that in the config service.

src/clj/puppetlabs/services/config/puppet_server_config_core.clj
```diff
@@ -110,12 +101,13 @@
...
-  [override-webserver-settings! webserver-settings puppet-config]
-  (let [{:keys [hostcert localcacert cacrl hostprivkey]} puppet-config
+  [override-webserver-settings! webserver-settings ca-settings puppet-config]
+  (let [{:keys [hostcert localcacert hostprivkey]} puppet-config
         overrides {:ssl-cert     hostcert
                    :ssl-key      hostprivkey
                    :ssl-ca-cert  localcacert
-                   :ssl-crl-path cacrl}]
+                   :ssl-crl-path (:cacrl ca-settings)
+                   }]
```
justinstoller: This should be on the line above

src/clj/puppetlabs/services/config/puppet_server_config_service.clj
```diff
@@ -28,14 +29,17 @@
...
                               puppet-config)
         (assoc context :puppet-config
                        {:puppetserver puppet-config}))))
 
   (get-config
     [this]
     (let [context        (tk-services/service-context this)
-          puppet-config  (:puppet-config context)]
+          puppet-config  (:puppet-config context)
+          tk-config      (get-config)]
```
justinstoller: what's with this `tk-config`?

test/integration/puppetlabs/puppetserver/bootstrap_testutils.clj
```diff
@@ -39,6 +39,9 @@
 (def master-log-dir
   "./target/master-var/log")
 
+(def ca-dir
+  "./target/ca")
```
justinstoller: In the fixtures directory we put the cadir under the confdir (though not in the ssldir).
eg `dev-resources/p/s/c_a/certificate_authority_int_test/ca_true_test/master/conf/ca/ca_pub.pem`

I would expect the equivalent for the fixtures within the ./target dir to be `./target/master-conf/ca/ca_pub.pem`.

test/integration/puppetlabs/puppetserver/bootstrap_testutils.clj
```diff
@@ -50,6 +53,7 @@
         (assoc-in [:jruby-puppet :master-var-dir] master-var-dir)
         (assoc-in [:jruby-puppet :master-run-dir] master-run-dir)
         (assoc-in [:jruby-puppet :master-log-dir] master-log-dir)
+        (assoc-in [:certificate-authority :cadir] ca-dir)
```
justinstoller: Ha

src/clj/puppetlabs/services/config/puppet_server_config_service.clj
lazartea: good catch, I think that might be left over from some previous debugging I was doing.

test/integration/puppetlabs/puppetserver/bootstrap_testutils.clj
```diff
@@ -40,7 +40,7 @@
   "./target/master-var/log")
 
 (def ca-dir
-  "./target/ca")
+  "./target/master-conf/ca")
```
lazartea: At the moment, this is passing tests for me locally. I'm very skeptical, and think that @justinstoller might be right that there's another piece to this that I've overlooked.

resources/ext/config/conf.d/puppetserver.conf
```diff
@@ -74,3 +74,9 @@ profiler: {
     # enable or disable profiling for the Ruby code; defaults to 'true'.
     #enabled: true
 }
+
+# settings related to the certificate authority
+certificate-authority: {
```
justinstoller: Should we put this in its own file?

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
```diff
@@ -55,22 +55,23 @@
...
+    (ca/initialize-master-ssl! master-settings certname (:ca-settings context))))
 
-  (retrieve-ca-cert!
-    [this localcacert]
-    (ca/retrieve-ca-cert! (get-in-config [:puppetserver :cacert])
-                          localcacert))
+(retrieve-ca-cert!
+  [this localcacert]
+  (let [context (tk-services/service-context this)]
+    (ca/retrieve-ca-cert! (:cacert (:ca-settings context)) localcacert)))
```
justinstoller: This isn't big enough to really warrant going back and changing it, (but I think it's fun to talk about (: ). For future reference, using multiple, nested map accesses like this isn't terribly idiomatic. Alternatives are:
```clj
(get-in context [:ca-settings :cacert])
```
Which is especially good if you have a default you want to supply as the third argument
or
```clj
(-> context :ca-settings :cacert)
```

also, if you are assigning a variable and then only using it once to dig another value from it you can just set the variable to that value initially, eg:
```clj
(let [cacert (-> this tk-services/service-context :ca-settings :cacert)]
  (ca/retrieve-ca-cert! cacert localcacert))
```

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
```diff
@@ -55,22 +55,23 @@
                        events)
              (ca/retrieve-ca-crl! ca-crl-file host-crl-file)))))
       (assoc context :auth-handler auth-handler
-                     :watcher watcher)))
+                     :watcher watcher
+                     :ca-settings settings)))
 
-  (initialize-master-ssl!
-   [this master-settings certname]
```
justinstoller: I could see making this indented two spaces rather than one below its function name (we used to have a policy that this should be one space, but intellij's tooling made it unenforceable). However, the aligning of the protocols you updated against the margin of this file is incorrect. The function names' opening paren should all be aligned with the init function above (two spaces in).

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
lazartea: Oh, thank you for pointing this out! I'm still learning a lot of the nuances of clojure, so I really appreciate it :) I think it might be worth going back and changing, especially since I'm already going to fix those white space issues you mentioned above.

resources/ext/config/conf.d/puppetserver.conf
lazartea: hm. Do you mean like a ``certificate_authority.conf`` or am I misunderstanding?

Magisus: I think the docs tend to refer to a `ca.conf`.

lazartea: ah, that makes more sense. Should I move it then?

justinstoller: Jayant's also adding things to the `certificate-authority` section and I *think* it's going to be a thing that warrants its own, albeit small, file by the time we ship Puppet 6. I'm a little torn on whether its better to do that now or if I'm setting us up for YAGNI.

Genuinely unsure on that.

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
```diff
@@ -55,23 +55,24 @@
...
-     (ca/initialize-master-ssl! master-settings certname settings)))
+     [this master-settings certname]
+     (let [settings (tk-services/service-context this)]
+       (ca/initialize-master-ssl! master-settings certname (:ca-settings settings))))
 
   (retrieve-ca-cert!
-    [this localcacert]
-    (ca/retrieve-ca-cert! (get-in-config [:puppetserver :cacert])
-                          localcacert))
+     [this localcacert]
```
lazartea: @justinstoller and now I think I've added too much whitespace... Should that space before ``[this localcacert]`` and the next line be removed?

justinstoller: yes

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
```diff
@@ -55,23 +55,24 @@
...
-                         localcacrl))
+     [this localcacrl]
+     (let  [cacrl (-> this tk-services/service-context :ca-settings :cacrl)]
+       (ca/retrieve-ca-crl! cacrl localcacrl)))
 
-  (get-auth-handler
-    [this]
-    (:auth-handler (tk-services/service-context this))))
+    (get-auth-handler
+       [this]
```
justinstoller: there's too much spaaaaaaaaaaaaace

jsane:
Curious - what happens to the cadir default in puppet/lib/puppet/defaults.rb and/or when specified in puppet.conf? Am assuming the default as set in src/clj/puppetlabs/puppetserver/certificate_authority.clj instead will always override it.

Magisus:
That is the plan, yes. We were also considering updating the defaults in Puppet though to correspond to the new puppetserver directory, just to send a more consistent message. But we haven't fully decided yet. The settings in Puppet will definitely be deprecated and scheduled to be removed in Puppet 7.

jsane:
When are we planning to merge this? I would like to use this for testing what I am doing

Magisus:
We are planning to merge it in time for Puppet 6. However, there is quite a bit of work that needs to be done before we can, most notably around [SERVER-1747](https://tickets.puppetlabs.com/browse/SERVER-1747) and some associated test updates. That being said, I would definitely try to incorporate or plan for this in any major CA work you're currently doing. Sorry that we can't merge it quite yet.

jsane:
That is fine. Turns out PE cannot use contents of any of the config files it manages as input to other config settings. So I wont be able to use or need this.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -258,6 +308,7 @@
           :autosign
           :ca-name
           :ca-ttl
+          :cadir
```
amysahli: This seems like a file/directory.  Why do we want to exclude it?

PR #1729, opened 2018-07-15
Add release notes for Server 5.3.4

gguillotte:
This assumes SERVER-2237 lands in time for release.


justinstoller:
@magis	didn't @amysa	's status work go into 5.3.4?

Magisus:
Yes, it did. Those were all PE tickets, which is probably why the notes aren't showing up here. Ticket is https://tickets.puppetlabs.com/browse/PE-24670

gguillotte:
@Magisus There are no Server tickets for that work? It'd be best if I had a ticket to link to and PE tickets aren't accessible externally.

CC @branan Data point for automating release note compilation: Outliers like this complicate the process.

PR #1730, opened 2018-07-17
(SERVER-2172) Update intermediate ca test to work with PE

justinstoller:


Magisus:
If this passes the pe-pse tests in the referenced PR, I'm 👍

justinstoller:
I've tested that this fixes the PE issue in the linked PR. I'm currently running this against FOSS as well to check for regression.

PR #1732, opened 2018-07-18
(maint) Add the ability to set the puppetserver hostname to something…

underscorgan:
… other than 'puppet'


Magisus:
Note that this won't promote into PE until after the current release goes out.

underscorgan:
@Magisus thanks! no worries, this doesn't affect pe at all at this point

PR #1733, opened 2018-07-20
(maint) Improve intermediate ca acceptance test

justinstoller:
This skips the test if there's no non-CA agent (we need at least those to hosts to properly test downloading all of the ssl files) and should resolve issues we're seeing in CI where we are running this test with only a single master & agent host.

While investigating `host.puppet['setting']` I remembered that `puppet_config` was added and should be used going forward (works with config sections rather than applications). See https://github.com/puppetlabs/beaker-puppet/commit/f5dcaa031bd7c6a15dc3914328960260238473ed#diff-1ad39dbbbad154ea718af70e36bca745 for details.

Will test locally with a mono install and w/o and report back.

### DIFF
--- a/acceptance/suites/tests/certificate_authority/intermediate_ca.rb
+++ b/acceptance/suites/tests/certificate_authority/intermediate_ca.rb
@@ -15,8 +15,8 @@ test_name 'Intermediate CA setup' do
         __FILE__))
   end
 
-  ssl_directory    = master.puppet['ssldir']
-  ca_directory     = master.puppet['cadir']
+  ssl_directory    = puppet_config(master, 'ssldir', section: 'master')
+  ca_directory     = puppet_config(master, 'cadir', section: 'master')
   auth_conf        = "/etc/puppetlabs/puppetserver/conf.d/auth.conf"
   test_directory   = master.tmpdir('intermediate_ca_files')
   backup_directory = master.tmpdir('intermediate_ca_backup')
@@ -93,11 +93,13 @@ test_name 'Intermediate CA setup' do
 
 
   # Remove the old CA infrastructure from an agent
-  on test_agent, "rm -rf #{test_agent.puppet['ssldir']}"
+  ssldir = puppet_config(test_agent, 'ssldir', section: 'agent')
+  cert = puppet_config(master, 'hostcert', section: 'master')
+  cacert = puppet_config(master, 'localcacert', section: 'master')
+  key = puppet_config(master, 'hostprivkey', section: 'master')
 
-  ssl_options = "--cert #{master.puppet['hostcert']} " +
-                "--cacert #{master.puppet['localcacert']} " +
-                "--key #{master.puppet['hostprivkey']}"
+  on test_agent, "rm -rf #{ssldir}"
+  ssl_options = "--cert #{cert} --cacert #{cacert} --key #{key}"
 
   # Create a dummy agent
   on test_agent,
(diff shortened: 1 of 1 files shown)


acceptance/suites/tests/certificate_authority/intermediate_ca.rb
```diff
@@ -88,14 +91,15 @@ def fixture(filename)
...
+  cacert = puppet_config(master, 'localcacert', section: 'master')
+  key = puppet_config(master, 'hostprivkey', section: 'master')
 
-  ssl_options = "--cert #{master.puppet['hostcert']} " +
-                "--cacert #{master.puppet['localcacert']} " +
-                "--key #{master.puppet['hostprivkey']}"
+  on test_agent, "rm -rf #{ssldir}"
+  ssl_options = "--cert #{cert} --cacert #{cacert} --key #{key}"
 
   # Create a dummy agent
```
Magisus: Would you mind making this and the other comments in this file into calls to `step`? It makes the output much easier to parse.

justinstoller:
The initial draft of the PR (w/o added steps) being correctly skipped on a single master agent test run:
```
Begin acceptance/suites/tests/certificate_authority/intermediate_ca.rb
Intermediate CA setup
  requires an agent not running on the CA
Begin teardown
End teardown
Warning: acceptance/suites/tests/certificate_authority/intermediate_ca.rb skipped in 0.00 seconds
```
The second version of the PR (w/ added steps) being ran with the split external ca test to ensure no weird artifacts are left behind:
```
      Test Suite: tests @ 2018-07-20 14:20:13 -0700
       Total Suite Time: 598.40 seconds
      Average Test Time: 299.20 seconds
              Attempted: 2
                 Passed: 2
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 2
```

This was ran with FOSS (our current failing PE tests are, I believe solely related to host layout rather than FOSSvsPE-ness and FOSS cycle time is faster) using

```
PACKAGE_BUILD_VERSION=6.0.0.master.SNAPSHOT.2018.07.19T2149
```
the last passing FOSS build

acceptance/suites/tests/certificate_authority/intermediate_ca.rb
```diff
@@ -88,38 +98,46 @@ def fixture(filename)
...
-             %q<-H 'Content-Type: application/json' -d '{"desired_state":"revoked"}' > +
-             "--url https://#{master}:8140/puppet-ca/v1/certificate_status/fake_agent_name?environment=production"
+  step 'Verify revocation works as expected' do
+    on master, "curl -XPUT #{ssl_options} " +
+               %q<-H 'Content-Type: application/json' -d '{"desired_state":"revoked"}' > +
+               "--url https://#{master}:8140/puppet-ca/v1/certificate_status/fake_agent_name?environment=production"
 
-  on master, "curl -XDELETE -H 'Content-Type: application/json' #{ssl_options} " +
-             "--url https://#{master}:8140/puppet-ca/v1/certificate_status/fake_agent_name?environment=production"
+    on master, "curl -XDELETE -H 'Content-Type: application/json' #{ssl_options} " +
```
amysahli: what's the purpose of the delete here?

justinstoller: Revoking a cert doesn't remove it from the CA's signed certs directory, this will delete the revoked cert as well.

PR #1734, opened 2018-07-23
(docs)(DOCUMENT-889)(SERVER-2253) profiling-mode

gguillotte:
Update `profiling-mode` setting docs to specify that it profiles service
startup.


samwoods1:
👍

PR #1736, opened 2018-07-31
(SERVER-2266) Note universe is required on 18.04.

gguillotte:
Ubuntu 18.04 needs the universe repository enabled to install `openjdk-8-jre-headless`, which is required by Server. Document this.

### DIFF
--- a/documentation/install_from_packages.markdown
+++ b/documentation/install_from_packages.markdown
@@ -43,7 +43,10 @@ update-alternatives --set java /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java
 
 ### Ubuntu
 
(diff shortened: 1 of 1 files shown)


gguillotte:
I'm putting this up quickly due to the importance of noting this, but I'd like some confirmation that this is required only on 18.04, and if it also affects 16.04 then whether this needs to also be noted in older Server docs.

Magisus:
I don't think it affects any other OS, but maybe @geoffnichols or his team can verify?

Magisus:
Platform OS says Ubuntu 18.04 only, but possibly need something similar for PuppetDB.

PR #1746, opened 2018-08-14
(maint) Exclude packaging when running JRuby specs

justinstoller:


Magisus:
Yeah agreed, this probably needs to go everywhere, even though it's only in our Jenkins CI where it's causing trouble (because we don't have CI for the older branches).

PR #1747, opened 2018-08-15
(maint) Exclude packaging when running JRuby specs

justinstoller:
The fun continues


mwaggett:
@justinstoller Are you still having issues? What errors are you seeing?

justinstoller:
I hate our conditional specs setup.

I think this is working now.

I've gotten it to pass running the previously included reproducer.

Running Beaker I had test failures (when beaker tried to require 'test/unit/assertions') because I apparently don't have a working Ruby 1.9 environment on my computer any more. However the packages successfully installed which is what would have failed, I think, if this PR broke something.

justinstoller:
Now watch travis fail

PR #1749, opened 2018-08-15
Merge 2.6.x -> 2.x

Magisus:
This also updates ezbake to the latest version.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -91,7 +91,7 @@ namespace :spec do
       BUNDLE_GEMFILE='#{PUPPET_SRC}/Gemfile' \
       GEM_HOME='#{TEST_GEMS_DIR}' GEM_PATH='#{TEST_GEMS_DIR}' \
       lein run -m org.jruby.Main \
-        -S bundle install --without extra development --path='#{TEST_BUNDLE_DIR}'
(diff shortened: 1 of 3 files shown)


project.clj
```diff
@@ -149,7 +149,7 @@
                                                [puppetlabs/puppetserver ~ps-version]
                                                [puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty9-version]
                                                [org.clojure/tools.nrepl nil]]
-                      :plugins [[puppetlabs/lein-ezbake "1.1.13"]]
```
justinstoller: I think this would implicitly remove Java 7 support from Puppet Server 2.x (see https://github.com/puppetlabs/ezbake/blob/master/CHANGELOG.md#120---2017-4-12) /cc @underscorgan

mwaggett: ugh I think you're right. I knew that happened somewhere, but didn't find it when skimming the ezbake changelog. I guess I can backport this stuff.

Magisus: In that case I think I'll do the mergeup without the ezbake update in this branch, and add it for 5.3.x, so we can see that go green.

justinstoller:
I'll merge once travis is done

PR #1751, opened 2018-08-16
(WIP) Run AppVeyor Docker Tests

Iristyle:


Iristyle:
Looks like we might be running into https://github.com/inspec/inspec/issues/2379

Iristyle:
Working off of #1789 for now... leaving this here as it has the `enable-lcow.ps1` script that's no longer being used.

PR #1756, opened 2018-08-20
Revert submodule and ezbake updates

Magisus:
The Jenkins submodule job ran and updated our puppet pins to invalid values. Ezbake 1.8.6 has a bug. This undoes both of those changes.


justinstoller:
This submodule has already passed integration tests once (before our automation accidentally downgraded us), I don't think there's a reason to wait for travis here.

PR #1757, opened 2018-08-20
(maint) Docker: change how we deal with puppetserver configuration

lutter:
We now expect all of /etc/puppetlabs/puppet to live on a volume, rather
than just expecting ssl/ to live there.

That way, users can modify puppetserver configuration by editing files on
the volume from outside the container, and have those changes persist
across restarts/recreation of the container.

Since users should be able to start puppetserver with an empty volume, and
have the container populate it with sensible defaults, we also stash the
relevant configuration files in /var/tmp/puppet from where we retrieve them
if the container is started with those files on the volume. The files in
question are auth.conf, hiera.yaml, puppet.conf, and puppetdb.conf.

The only settings that we need to know when starting on an empty volume are
the certname and dns_alt_names for the server's certificate as they are a
pain to change later. The certname gets taken from the container's
hostname, and the dns_alt_names can be passed in with the DNS_ALT_NAMES
environment variable.

### DIFF
--- a/docker/puppetserver-standalone/docker-entrypoint.sh
+++ b/docker/puppetserver-standalone/docker-entrypoint.sh
@@ -1,7 +1,21 @@
 #!/bin/bash
 
-chown -R puppet:puppet /etc/puppetlabs/puppet/ssl
+chown -R puppet:puppet /etc/puppetlabs/puppet/
 chown -R puppet:puppet /opt/puppetlabs/server/data/puppetserver/
+
+# During build, pristine config files get copied to this directory. If
+# they are not in the current container, use these templates as the
+# default
+TEMPLATES=/var/tmp/puppet
+
+cd /etc/puppetlabs/puppet
+for f in auth.conf hiera.yaml puppet.conf puppetdb.conf
+do
+    if ! test -f $f ; then
+        cp -p $TEMPLATES/$f .
+    fi
+done
+cd /
 
 if test -n "${PUPPETDB_SERVER_URLS}" ; then
   sed -i "s@^server_urls.*@server_urls = ${PUPPETDB_SERVER_URLS}@" /etc/puppetlabs/puppet/puppetdb.conf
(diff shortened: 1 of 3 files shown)


lutter:
@stahnma @underscorgan this is still WIP .. I am still trying to validate these changes, but wanted to share what I am thinking to make sure this makes sense to you. I've tried it out with containers that I built based FROM puppet/puppetserver but still need to make sure these changes don't have tyops or similar.

docker/puppetserver-standalone/docker-entrypoint.sh
```diff
@@ -13,4 +26,11 @@ if test -n "${AUTOSIGN}" ; then
   puppet config set autosign "$AUTOSIGN" --section master
 fi
 
+# Allow setting the dns_alt_names for the server's certificate. This
+# setting will only have an effect when the container is started without
+# an existing certificate on the /etc/puppetlabs/puppet volume
+if test -n "${DNS_ALT_NAMES}"; then
+  puppet config set dns_alt_names "${DNS_ALT_NAMES}" --section master
```
underscorgan: Should we only run the `puppet config set` if the cert doesn't exist yet? I know practically it has no effect, but could lead to a confusing config file?

docker/puppetserver-standalone/docker-entrypoint.sh
```diff
@@ -1,8 +1,21 @@
...
-chown -R puppet:puppet /etc/puppetlabs/puppet/ssl
+chown -R puppet:puppet /etc/puppetlabs/puppet/
 chown -R puppet:puppet /opt/puppetlabs/server/data/puppetserver/
 
+# During build, pristine config files get copied to this directory. If
+# they are not in the current container, use these templates as the
+# default
+TEMPLATES=/var/tmp/puppet
+
+cd /etc/puppetlabs/puppet
```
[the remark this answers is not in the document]
lutter: It might be clearer to just `cd /` back when we're done in that directory.

docker/puppetserver-standalone/docker-entrypoint.sh
lutter: Good point. The deluxe version would be to also log a warning if DNS_ALT_NAMES is set to something other than what's in `puppet.conf`

docker/puppetserver-standalone/docker-entrypoint.sh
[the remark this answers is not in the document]
stahnma: push/pop is a bash-ism and if we ever get these on alpine, bash is not the default shell, so +1 on  `cd`

docker/puppetserver-standalone/docker-entrypoint.sh
```diff
@@ -1,8 +1,22 @@
 #!/bin/bash
 
-chown -R puppet:puppet /etc/puppetlabs/puppet/ssl
+chown -R puppet:puppet /etc/puppetlabs/puppet/
```
stahnma: This should fix most of my NFS mount issues 🙏

PR #1758, opened 2018-08-20
(SERVER-2156) add support for multi-platform tasks

adreyer:
This adds support for multi-platform tasks by including all
"implementations in the "files" response for TaskDetails. In order to
prevent duplicating logic in both puppet-server and ruby this moves
metadata reading and validation out of clj in puppet-server and into
puppet itself.

fix tests

### DIFF
--- a/documentation/puppet-api/v3/task_detail.json
+++ b/documentation/puppet-api/v3/task_detail.json
@@ -52,11 +52,11 @@
               }
             },
             "required": ["path", "params"],
-            "additionalProperties": false
+            "additionalProperties": true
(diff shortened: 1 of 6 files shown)


nicklewis:
Since `files` is returned separately from `implementations` in the metadata, does this mean orchestrator will need to correlate the two values against one another when deciding which implementation to use?

adreyer:
pxp-agent will in most cases. the filename key should map the name in the implementation.

MikaelSmith:
Puppet PR merged.

PR #1759, opened 2018-08-21
5.3.x

MFredette:
Release notes for Puppet Server 5.3.5.


documentation/release_notes.markdown
```diff
@@ -12,6 +12,20 @@ canonical: "/puppetserver/latest/release_notes.html"
 
 For release notes on versions of Puppet Server prior to Puppet Server 5, see [docs.puppet.com](https://docs.puppet.com/puppetserver/2.8/release_notes.html).
 
+## Puppet Server 5.3.5
+
+Released 21 August, 2018
+
+### New features
+
+- We have added a new command line tool for interacting with the Puppet CA, under the `puppetserver ca` command. This tool can be used to generate an intermediate CA for Puppet Server, and to generate, sign, revoke, clean, and list certs. The Puppet 5 series still contains caveats about manually copying these certs to agents. All of these actions are executed by making requests to Puppet Server's CA API, in particular the `certificate_status` and `certificate_statuses` endpoints.
```
Magisus: Can we change the line about intermediate CA caveats to read something like 
```
The Puppet 5 series still contains all of the old caveats about using an intermediate CA, including the need to manually copy the certs to the agent, and the need to configure CRL checking (see https://puppet.com/docs/puppetserver/5.3/intermediate_ca_configuration.html) for details. Therefore the `generate` and `import` commands that create intermediate CAs should be used with caution."
```
 I ran into the CRL checking bug while manually validating this the other day; restrictions in puppetserver prevent revoking and cleaning certs when using an intermediate CA on this branch, unless CRL checking has been turned down or off.

I am also planning to send out an email about this tool and its caveats for 5.x, but it seems maybe good to have here too.

MFredette: I made this change in my new PR...

PR #1762, opened 2018-08-22
Merge 5.3.x -> master

Magisus:


Rakefile
```diff
@@ -174,8 +174,8 @@ namespace :spec do
       PATH='#{TEST_GEMS_DIR}/bin:#{path}' \
       BUNDLE_GEMFILE='#{PUPPET_SRC}/Gemfile' \
       GEM_HOME='#{TEST_GEMS_DIR}' GEM_PATH='#{TEST_GEMS_DIR}' \
-      lein run -m org.jruby.Main \
-        -S bundle install --without extra development --path='#{TEST_BUNDLE_DIR}' --retry=3
+      lein #{profile} run -m org.jruby.Main \
```
justinstoller: I don't think this exists in master? I'll double check in a second.

Magisus:
Looks like this is missing some Rakefile updates as well.

Rakefile
justinstoller: Didn't get this far, looks like you and travis are seeing the same thing I am. However, the reason I thought this line was off is because I *think* we used to use this for switching between 9k and 1.7 in our specs.

Magisus: Ah okay, makes sense. I'll try to fix the packaging issue and remove this.

Magisus:
I think the revert of the messed up mergeup is messing with things. Going to have to go through this more carefully.

Magisus:
Updated.

justinstoller:
This is failing for me locally, but I'm happy to trust travis and say this is my environment.

```
Tried to load org.apache.maven/maven-model version 3.0.4 but 3.5.0 was already loaded.
Bundler::GemNotFound: Could not find gem 'semantic_puppet (~> 1.0) java' in any of the gem sources listed in your Gemfile.
  block in verify_gemfile_dependencies_are_found! at /Users/justin/Backup/code/work/puppetserver/vendor/test_gems/gems/bundler-1.16.3/lib/bundler/resolver.rb:289
                                             each at org/jruby/RubyArray.java:1735
```

Magisus:
That looks like you may need to re-run `dev-setup`, to install semantic-puppet, which is longer vendored in puppet#master.

Magisus:
Gonna redo this so it can pick up https://github.com/puppetlabs/puppetserver/pull/1764

PR #1766, opened 2018-08-22
(SERVER-2290) Add a setting to enable signing CSRs with auth extensions

amysahli:
Before this commit we completely disallowed signing certificate requests
with authorization extensions from Puppetserver's certificate_status
endpoint.  Now users can choose to allow signing of certs with auth
extensions by setting allow-authorization-extensions to TRUE in the
puppetserver .conf file.  Setting is False by default.


amysahli:
Tested this locally by adding to my csr_attributes.yaml fil:
```
extension_requests:
  pp_authorization: "MrBean"
```
Creating a cetificate:
```
bundle exec puppet agent -t --server localhost --certname=smiffan
```
and then signing (once will dev/puppetserver.conf allow-authorization-extensions set to FALSE and once with it set to TRUE (after restarting server)
```
~/Development/puppetserver-ca-cli (SERVER-2276) $🐰  be ./exe/puppetserver-ca sign --all
Error:
    When attempting to sign certificate request 'smiffan', received
      code: 409
      body: CSR 'smiffanyjr2' contains an authorization extension, which is disallowed. To allow authorization extensions, set allow-authorization-extensions to true in your puppetserver.conf file, restart the puppetserver, and try signing this certificate again.

```
AND
```

~/Development/puppetserver-ca-cli (SERVER-2276) $🐰  be ./exe/puppetserver-ca sign --all
--
Successfully signed certificate request for smiffany


```

resources/ext/config/conf.d/puppetserver.conf
```diff
@@ -79,4 +79,5 @@ profiler: {
 certificate-authority: {
     # allow CA to sign certificate requests that have subject alternative names.
     # allow-subject-alt-names: false
+    # allow-authorization-extensions: false
```
Magisus: Maybe there should be a comment on this one too.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1069,21 +1076,21 @@
...
-                (i18n/trs "CSR ''{0}'' contains an authorization extension, which is disallowed." csr)
-                (i18n/trs "Use {0} to sign this request."
-                     (format "`puppet cert --allow-authorization-extensions sign %s`"
-                             (get-csr-subject csr))))})))))
+        (if (false? allow-authorization-extensions)
+          (sling/throw+
+            {:kind :disallowed-extension
+             :msg (format "%s %s %s"
+                          (i18n/trs "CSR ''{0}'' contains an authorization extension, which is disallowed." (get-csr-subject csr))
+                          (i18n/tru "To allow authorization extensions, set allow-authorization-extensions to true in your puppetserver.conf file,")
```
justinstoller: should these also be `trs` rather than `tru`?

justinstoller: Doesn't look like we're consistent through the file. I think `trs` respects the server side local and `tru` respects the requesters, ie for things that would be seen in the broswer. Does that sound, right?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -54,6 +54,7 @@
    and request handling during normal Puppet operation.
    Most of these are Puppet configuration settings."
   {:access-control           (schema/maybe AccessControl)
+   :allow-authorization-extensions  schema/Bool
```
justinstoller: We maintain this list here and in `ca_testutils.clj`, in ca_testutils we updated the column spacing. We might as well do that here. Or not align the spacing at all, fine either way.

justinstoller:
There an error message that could part of it translated one way and the other part translated another way. Other than that the code looks great.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
amysahli: I can align them (I saw the manage-file-internal-positions below was not aligned, so I didnt bother here).  BUt I'll fix 'em up.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
justinstoller: That sounds fine to me, @Magisus I think you're our i18n guru, what do you think?

Magisus: I actually have no idea how our clojure tooling works, sorry.

justinstoller:
Functionally reviewed this as well. Looks great!

PR #1767, opened 2018-08-22
(CPR-600) Auto-publish to dockerhub

underscorgan:
This updates the build script to build containers in the correct
namespace for publishing to dockerhub.


underscorgan:
welp, definitely targeted the wrong branch. Luckily was based on that branch originally so this didn't get messy.

PR #1768, opened 2018-08-23
(maint) Docker: change how we deal with puppetserver configuration

lutter:
We now expect all of /etc/puppetlabs/puppet to live on a volume, rather
than just expecting ssl/ to live there.

That way, users can modify puppetserver configuration by editing files on
the volume from outside the container, and have those changes persist
across restarts/recreation of the container.

Since users should be able to start puppetserver with an empty volume, and
have the container populate it with sensible defaults, we also stash the
relevant configuration files in /var/tmp/puppet from where we retrieve them
if the container is started with those files on the volume. The files in
question are auth.conf, hiera.yaml, puppet.conf, and puppetdb.conf.

The only settings that we need to know when starting on an empty volume are
the certname and dns_alt_names for the server's certificate as they are a
pain to change later. The certname gets taken from the container's
hostname, and the dns_alt_names can be passed in with the DNS_ALT_NAMES
environment variable.

### DIFF
--- a/docker/puppetserver-standalone/Dockerfile
+++ b/docker/puppetserver-standalone/Dockerfile
@@ -39,7 +39,9 @@ COPY puppetserver /etc/default/puppetserver
 COPY logback.xml /etc/puppetlabs/puppetserver/
 COPY request-logging.xml /etc/puppetlabs/puppetserver/
 
-RUN puppet config set autosign true --section master
+RUN puppet config set autosign true --section master && \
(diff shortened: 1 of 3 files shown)


lutter:
@underscorgan I cherry-picked 01fdba9e onto the 5.3.x branch. Let me know if there's a better way to get this onto the branch since I already merged it into master

underscorgan:
@lutter cherry-pick seems good, thanks!

PR #1775, opened 2018-08-29
(maint) Add back package:bootstrap and package:implode for compatibility

underscorgan:


Magisus:
This is the only one of these we want to merge like this, we'll merge the change up from here.

underscorgan:
okay. figured cherry-picking might be easier to get the change in quickly, i'll close the others.

Magisus:
I'll get this rolling as soon as Travis finishes. Most of these don't have pipelines, so we won't be waiting on CI at each step.

underscorgan:
@Magisus can this be merged up? It ended up missing for the 6.0 release

Magisus:
Oh womp. Sorry about that... if we get some time after this release maybe we'll get auto mergeups working.

underscorgan:
No worries. As long as it makes it in before the next group of releases, it causes issues with shipping

PR #1776, opened 2018-08-29
(maint) Add back package:bootstrap and package:implode for compatibility

underscorgan:


Magisus:
We really want to be doing these as mergeups. I can merge the 2.6.x one when travis finishes and then take care of the mergeups if you like.

underscorgan:
@Magisus that would be great, thanks.

PR #1780, opened 2018-08-30
(SERVER-2225) (SERVER-2222) Move CA settings to puppetserver

Magisus:
This PR combines @lazartea 's work with an additional commit to move the remaining CA-related settings to puppetserver. Needs cleanup, but appears to work for basic use cases.


src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -118,13 +119,78 @@
...
+  [cadir]
+  (str cadir "/ca_crl.pem"))
+
+(defn default-cakey
+  [cadir]
+  (str cadir "/ca_key.pem"))
+
+;; TODO how do we get certname here?
+(def default-caname
+  "Puppet CA: $certname")
```
Magisus: Maybe we can pull it out of puppet's settings?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -118,13 +119,78 @@
...
+
+;; TODO where should this actually live?
+(defn default-autosign
+  [cadir]
+  (str cadir "/autosign.conf"))
```
Magisus: This currently lives in puppet's confdir. Is it something we want to move, or not?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -118,13 +119,78 @@
...
+
+(def default-allow-duplicate-certs
+  false)
+
+;; 15 years in seconds
+(def default-ca-ttl
+  473000000)
```
Magisus: We currently use five years in Puppet, but there was a request from Nick Walker approved by Eric Sorenson to make it longer.

test/unit/puppetlabs/services/jruby/jruby_puppet_testutils.clj
```diff
@@ -228,18 +228,12 @@ create-mock-pool-instance :- JRubyInstance
     {"allow_duplicate_certs" false
      "autosign" true
      "keylength" 2048
-     "cacert" (str cadir "/ca_crt.pem")
      "ca_name" (str "Puppet CA: " certname)
```
Magisus: Need to update this file to remove the remaining removed settings.

src/clj/puppetlabs/services/config/puppet_server_config_core.clj
```diff
@@ -107,6 +134,12 @@
    (let [config (get-puppet-config* jruby-puppet)]
      (assoc config :puppet-version (.puppetVersion jruby-puppet)))))
 
+(defn resolve-ca-settings
+  [puppet-config server-ca-settings]
+  (when-let [cadir (:cadir server-ca-settings)]
+    (-> (interpolate-cadir cadir puppet-config)
+        (merge server-ca-settings))))
```
Magisus: Maybe we don't do the merge here and just return puppet's config hash with `cadir` interpolated? And then call this somewhere before we return the hash? It does seem a bit weird (like maybe hard to debug) to do that to something we're going to go around thinking of as settings out of puppet...

justinstoller:
Talked with Maggie yesterday and my understanding was that this was going to be split into two PRs:
 - One against 5.3.x to allow users to migrate their CA related settings to Puppet Server's conf.d (resolving deprecation warnings in the 5 series) - I believe that's most of the last four commits.
 - Then another PR after merging the previous work up which changes the default location of CA related files and ignores values set in the puppet.conf (to match the removal of those settings in Puppet 6) - the first three commits in this PR.

Magisus:
Closing this, I think this work is going to be undertaken elsewhere.

PR #1781, opened 2018-08-30
(SERVER-2287) Add OID for custom CLI auth extension

Magisus:
This commit adds a custom extension in Puppet's arc that is checked when
allowing access to the certificate_status(es) endpoint. This extension
should only be added to the master's cert by the CLI tool's generate
command, and it is disallowed when signing other certs with this CA.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -160,6 +160,11 @@
 (def subject-alt-names-oid
   "2.5.29.17")
 
+;; Extension that is checked when allowing access to the certificate_status(es)
+;; endpoint. Should only be present on the puppet master cert.
+(def cli-auth-oid
+  "1.3.6.1.4.1.34380.1.3.39")
+
 (def puppet-short-names
   "A mapping of Puppet extension short names to their OIDs. These appear in
   csr_attributes.yaml."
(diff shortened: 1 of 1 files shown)


ezbake/config/conf.d/auth.conf
```diff
@@ -45,6 +45,38 @@ authorization: {
...
+            # Allow the CLI tool to query and update a certificate status,
+            # using the master's hostcert.
+            match-request: {
+                path: "/puppet-ca/v1/certificate_status"
+                type: path
+                method: [get, put, delete]
+            }
+            allow: {
+                extensions: {
+                    pp_cli_auth: true
```
Magisus: Is there a different value than "true" we should be checking against here? I remember we had thought of including the hostname, but apparently the values for extensions in tk-auth have to match the value of the extension exactly, so I wasn't sure how to accomplish that programmatically, considering this is a static file that we ship.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -160,6 +160,11 @@
 (def subject-alt-names-oid
   "2.5.29.17")
 
+;; Extension that is checked when allowing access to the certificate_status(es)
+;; endpoint. Should only be present on the puppet master cert.
+(def cli-auth-oid
+  "1.3.6.1.4.1.34380.1.3.39")
```
Magisus: Should we be using a "crazier" number? What are valid values, and what is more likely to conflict with things people might already be using?

Magisus:
The PR adding this extension to the master cert in the gem is here: https://github.com/puppetlabs/puppetserver-ca-cli/pull/28

Magisus:
@jsane would you mind taking a look at this and letting us know whether this is a legitimate approach for allowing restricted access to these currently locked-down endpoints?

Magisus:
@ahpook do we have a sense of how people use custom extensions and what value we could choose to reserve for this one such that it's unlikely to conflict with what people are already using?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
Magisus: I think we want to disallow adding this particular custom extension to any old cert, restricting it to the generated master's cert. But I assume we should still document it in that list, with a note about how it's reserved.

Of course if this whole approach seems weird, please let us know. We wanted to be more restrictive than just allowing "localhost" to access these endpoints.

jsane:
That should be fine; I was curious if we could have used the existing pp_auth_role (xxx.13) extension unless it is not being stamped with appropriate roles.

PR #1782, opened 2018-08-31
(maint) Update CA CLI gem to 0.5.0

Magisus:


Magisus:
The test failure is related to a puppet change that is being fixed separately.

PR #1784, opened 2018-09-05
(SERVER-2302) Allow access to all status endpoints

justinstoller:
Previously, we added the ability for the status service to honor tk-auth
configs. To provide backwards compatability we opened up the default
tk-auth rules for the status service to be un-authenticated. However, in
that work we only opened up /status/v1/services, to allow calls to all
endpoints mounted under the status service this relaxes the path to only
be /status/v1.

During the previous work we also updated the tk-status service
documentation, however we missed documenting this change in our public
Puppet Server documentation. This includes that update as well.

### DIFF
--- a/documentation/status-api/v1/services.markdown
+++ b/documentation/status-api/v1/services.markdown
@@ -123,15 +123,13 @@ Content-Type: application/json
 
 ### Authorization
 
-Requests to the `services` endpoint are not authorized by either the
-[Trapperkeeper-based authorization process][`auth.conf`] introduced in Puppet
-Server 2.3.0, or the deprecated Ruby-based authorization process. For more
-information about the supported and deprecated Puppet Server authorization
-processes and configuration settings, see the
+Requests to the `services` endpoint are authorized by the
+[Trapperkeeper-based authorization process][`auth.conf`] as of Puppet
+Server 5.3.0. For more information about the supported Puppet Server
+authorization processes and configuration settings, see the
 [`auth.conf` documentation][`auth.conf`].
 
-Unless the `client-auth` setting is set to `required` for the webserver, a
-client should be permitted to make a request to this endpoint via SSL with or
-without the use of a client certificate. See
+One may also restrict access to the status service by changing the
+`client-auth` setting to `required` for the webserver. See
 [Configuring the Webserver Service](https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/master/doc/jetty-config.md#client-auth)
 for more information on the `client-auth` setting.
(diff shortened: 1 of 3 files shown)


justinstoller:
needs to be against 5.3.x....

Magisus:
So this seems fine, but earlier you had mentioned wanting to be more explicit in our auth rules about which different endpoints are supported. Does that seem like overkill here because we really do want blanket coverage with this rule? (i.e. is there anything else under status that we're also trying to open up with this?)

justinstoller:
I did want blanket coverage of the status service (though that currently includes just services and simple endpoints).

The difference between the two changes that I thought were salient were:
 - partial segment matching (eg using status to match status and statuses vs using status/v1 to match status/v1/simple and /status/v1/services)
 - different valid http verbs (and relatedly, very different security implications from accidental exposure).

Having said that, I if there's any question I do think we should prefer to err on the side of more explicit. I'll update accordingly.

justinstoller:
I've built this into a package with a version of 5.3.6.SNAPSHOT.2018.09.06T1446

I was successfully able to use that build in the pooler:
```
[root@htwpvrh14gywukw ~]# curl -vk https://htwpvrh14gywukw.delivery.puppetlabs.net:8140/status/v1/simple
* About to connect() to htwpvrh14gywukw.delivery.puppetlabs.net port 8140 (#0)
*   Trying 10.32.121.206...
* Connected to htwpvrh14gywukw.delivery.puppetlabs.net (10.32.121.206) port 8140 (#0)
* Initializing NSS with certpath: sql:/etc/pki/nssdb
* skipping SSL peer certificate verification
* NSS: client certificate not found (nickname not specified)
* SSL connection using TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA
* Server certificate:
*       subject: CN=htwpvrh14gywukw.delivery.puppetlabs.net
*       start date: Sep 05 21:56:04 2018 GMT
*       expire date: Sep 05 21:56:04 2023 GMT
*       common name: htwpvrh14gywukw.delivery.puppetlabs.net
*       issuer: CN=Puppet CA: htwpvrh14gywukw.delivery.puppetlabs.net
> GET /status/v1/simple HTTP/1.1
> User-Agent: curl/7.29.0
> Host: htwpvrh14gywukw.delivery.puppetlabs.net:8140
> Accept: */*
>
< HTTP/1.1 200 OK
< Date: Thu, 06 Sep 2018 22:00:12 GMT
< Content-Type: text/plain;charset=utf-8
< Vary: Accept-Encoding, User-Agent
< Content-Length: 7
< Server: Jetty(9.4.z-SNAPSHOT)
<
* Connection #0 to host htwpvrh14gywukw.delivery.puppetlabs.net left intact
running
```

PR #1788, opened 2018-09-07
(SERVER-2231) Disable using the infrastructure CRL by default

Magisus:
This commit updates puppetserver's configuration to not use a separate
infrastructure CRL by default.


src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -138,7 +138,7 @@
         defaults {:infra-nodes-path (str cadir "/infra_inventory.txt")
                   :infra-node-serials-path (str cadir "/infra_serials")
                   :infra-crl-path (str cadir "/infra_crl.pem")
-                  :enable-infra-crl true
+                  :enable-infra-crl false
```
justinstoller: can we update the documented default in dev/puppetserver.conf as well

PR #1790, opened 2018-09-10
(SERVER-2308) Update CA bootstrap code to add CLI auth extension

Magisus:
This commit updates the code that generates the master's host cert on
server start to add the auth extension needed by the CA CLI tool to talk
to the certificate_status endpoint.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -722,7 +722,10 @@
                          #{:key-encipherment
                            :digital-signature} true)
                        (utils/subject-key-identifier
-                         master-public-key false)]]
+                         master-public-key false)
+                       {:oid cli-auth-oid
+                        :critical false
+                        :value "true"}]]
     (validate-dns-alt-names! alt-names-ext)
     (when csr-attr-exts
       (validate-extensions! csr-attr-exts))
--- a/test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
+++ b/test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
@@ -685,6 +685,11 @@
                 "The Subject Alternative Names extension should contain the
                  master's actual hostname and the hostnames in $dns-alt-names")))
 
+        (testing "has CLI auth extension"
+          (let [cli-auth-ext (utils/get-extension-value hostcert cli-auth-oid)]
+            (is (= "true" cli-auth-ext)
+                "The master cert should have the auth extension for the CA CLI.")))
+
         (testing "is also saved in the CA's $signeddir"
           (let [signedpath (path-to-cert (:signeddir ca-settings) "master")]
             (is (fs/exists? signedpath))
@@ -1156,6 +1161,9 @@
                            {:oid      "2.5.29.14"
                             :critical false
                             :value    subject-pub}
+                           {:oid      "1.3.6.1.4.1.34380.1.3.39"
+                            :critical false
+                            :value    "true"}
                            {:oid      utils/subject-alt-name-oid
                             :critical false
                             :value    {:dns-name ["puppet" "subject"]}}]]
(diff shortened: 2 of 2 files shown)


src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -722,7 +722,10 @@
                          #{:key-encipherment
                            :digital-signature} true)
                        (utils/subject-key-identifier
-                         master-public-key false)]]
+                         master-public-key false)
+                       {:oid cli-auth-oid
+                        :critical false
+                        :value "true"}]]
```
Magisus: Would we prefer this as a utility method in `jvm-ssl-utils` instead? It seemed like maybe overkill given how simple it is, but I'm happy to make one.

Magisus:
Tested that this works when using the bootstrapping code to generate the CA.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
justinstoller: I don't think that putting it in ssl-utils is necessary though we may want to put an assertion in `initialize-master-ssl!-test` to call out how important it is and make sure it round trips like we expect?

Magisus: Will do. Didn't want to dig into testing until I knew that this was an okay approach.

Magisus:
Updated.

justinstoller:
Tested locally by deleting the ssl dir and starting puppetserver, updated auth to only allow certs with extension in the repl, then successfully created a new cert:
```
puppetserver justin (merge-maggie *$) ‖ git diff dev/
diff --git a/dev/puppetserver.conf b/dev/puppetserver.conf
index 5fdbf5f..54d1e7b 100644
--- a/dev/puppetserver.conf
+++ b/dev/puppetserver.conf
@@ -159,12 +159,26 @@ authorization: {
     rules: [
         {
             match-request: {
-                path: "/"
+                path: "/puppet-ca/v1/certificate"
                 type: "path"
+                method: [get, put, delete]
+            }
+            allow: {
+               extensions: {
+                   pp_cli_auth: "true"
+               }
             }
-            allow-unauthenticated: true
             sort-order: 1
-            name: "allow all"
+            name: "don't allow all"
+        }
+        {
+            match-request: {
+                path: "/"
+                type: "path"
+            }
+            deny: "*"
+            sort-order: 2
+            name: "deny all"
         }
     ]
 }
```
```
puppetserver-ca-cli justin (maint-derp-der $%) ‖ be exe/puppetserver-ca create --certname barberry
Successfully submitted certificate request for barberry
Successfully signed certificate request for barberry
Successfully saved certificate for barberry to /Users/justin/.puppetlabs/etc/puppet/ssl/certs/barberry.pem
Successfully saved private key for barberry to /Users/justin/.puppetlabs/etc/puppet/ssl/private_keys/barberry.pem
Successfully saved public key for barberry to /Users/justin/.puppetlabs/etc/puppet/ssl/public_keys/barberry.pem
```

smortex:
This change requires people upgrading from a previous Puppet version to tweak their master certificate to make authentication work for the new API endpoints.

Finding relevant instructions was not straightforward, I ended up writing a ruby script to fix this issue ([smortex/puppet-add-cli-auth-to-certificate](https://github.com/smortex/puppet-add-cli-auth-to-certificate)), but maybe sore general guidance would be welcomed from the Puppet community?

Magisus:
The instructions for the new CLI tool address this issue. We recommend that people update their auth.conf to allow the master's certname to access the endpoints: https://puppet.com/docs/puppetserver/6.0/subcommands.html#available-actions. But this approach also works (and is a little more secure), though more difficult.

smortex:
Oh my…  I completely missed this section.  Thanks for the pointer @Magisus!

Magisus:
Understandable, some docs links we have really need improvement. Thanks for the feedback, we'll try to make this more discoverable!

PR #1792, opened 2018-09-10
Set up CI with Azure Pipelines

Iristyle:


Iristyle:
This branch should have been against my fork, but the tooling created it here instead. Re-opening from fork.

PR #1793, opened 2018-09-10
(maint) Setup CI with Azure Pipelines

Iristyle:
This attempts to build puppetserver container using Azure and then consume / run it on a Windows hosts.

A few snags thus far:

* Can only build on Windows Server 2016 vmImage, which doesn't support LCOW, which prevents Linux container builds on Windows
* Server 1803 cannot build containers (because its hosted by 2016), but seems like it can run LCOW (with some work)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,72 @@
+pool:
+  # vmImage: 'VS2017-Win2016'
+  vmImage: Vs2017-Server2016
+  # vmImage: 'win1803'
+
+# strategy:
+#   matrix:
+#     - ruby_version: 24-x64
+
+steps:
+- powershell: |
+    # basic diagnostic host info
+    Get-ComputerInfo | select WindowsProductName, WindowsVersion, OsHardwareAbstractionLayer
+    Get-ChildItem Env: | % { Write-Host "$($_.Key): $($_.Value)"  }
+    docker version
+    docker images
+    docker info
+    sc.exe qc docker
+    # install Chocolatey - not necessary since image has it
+    # Set-ExecutionPolicy Bypass -Scope Process -Force; iex ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1'))
+    # install Ruby 2.5 and Go (for building LCOW)
+    choco install -y --no-progress ruby golang
+    #
+    # add to path
+    #
+    refreshenv
+    $Env:PATH += ';C:\tools\ruby25\bin'
+    #
+    # config the Go install
+    #
+    # Env:GOROOT already defaults to c:\tools\go
+    $Env:GOPATH = 'c:\go'
+    New-Item -Type Directory $Env:GOPATH
+    $Env:GOROOT = 'c:\tools\go'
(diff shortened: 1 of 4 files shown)


chrisrpatterson:
@Iristyle the Server 1803 image should be able to build containers for Windows but the image base needs to be Server 1803.  We also have the Ubuntu 1604 image which will enable you to build Linux containers.

Iristyle:
Hey @chrisrpatterson - thanks for the response.

It might not be clear, but we're interested in testing two primary things:

* Building Linux containers on Windows
* Running / testing that built Linux container on Windows

Please correct me if I'm mistaken, but I don't think there's a hosted agent (yet?) that can meet those requirements. I was unable to get a Linux container to build on either 2016 or Server 1803. Further, 1803 would be the only environment capable of running the container.

For now, we've setup a self-hosted agent on a _v3 VM (to supported nested virt), and I've setup recent builds of LCOW - that fully supports the workflow.


The best section on caveats I've found is in the docs at https://docs.microsoft.com/en-us/azure/devops/pipelines/languages/docker?view=vsts&tabs=yaml#microsoft-hosted-vs2017-windows-agents which state

> * Using Windows agents, you can only build Docker images with Windows Server 2016 as the container OS. You cannot build Docker images with Windows Server 1803 as the container OS since the host operating system on the virtual machines is Windows Server 2016.
> * We don't yet have a pool of Microsoft-hosted agents running Windows Server 1803. Until this is available, you can build Windows Server 1803 images using self-hosted agents.


I have tried the `vmImage: 'win1803'` which as I understand, is itself a container running on top of Server 2016? I think what we'd need for a hosted agent is something similar to the `Vs2017-Server2016` image, but sitting on top of a newer 1709+ base OS. Do you know if there are any such plans in the works for that?

Thanks again for the response!

chrisrpatterson:
@Iristyle Interesting, I am curious why specifically you want to build and test Linux containers on Windows.  Can you give me some additional details on that?  

We do not run any infrastructure in containers for builds, all jobs run on a fresh VM.  The win1803 image is a VM running Windows Server Core 1803 and the latest docker ee builds for Windows.  At the moment they are not running on the v3 skus but that is a change we are planning to deploy customers will have access to nested virtualization which enables a bunch of scenarios.

Iristyle:
@chrisrpatterson we're building this application stack (puppetserver), which is a Java app that only runs on Linux currently. We're enabling workflows for environments that only manage Windows to be able to run this server in a Windows environment. We want open-source Windows only contributors to be able to build / test in their environments as well. I can provide more details privately in email if that helps.

Thanks for the clarification on the 1803 VMs. Given no _v3 SKU, now I understand why builds failed! Any timeline on the _v3 infra? Or any way to opt in to a beta?

Thanks again for the info!

chrisrpatterson:
@Iristyle ok that makes sense.  

I will check with the team on the timeframe and let you know.

Iristyle:
Closing in favor of #1796 which uses a custom agent. Some of the unused helper scripts live on in https://gist.github.com/Iristyle/a4d58cd77afeda8830523e4be15d23de

PR #1796, opened 2018-09-11
(maint) Build Container in Azure Pipelines

Iristyle:
This enables building Docker images / containers in the Azure Devops Pipelines. Uses a custom agent in the `Default` agent pool that has Docker with LCOW supported installed / enabled, Ruby 2.5 and hadolint installed - all required for this build. It also has Go and Node.js installed, which are currently not required.

Currently requires https://github.com/puppetlabs/puppet_docker_tools/pull/26 (currently installing as a gem from that git ref) and has cherry-picked over test fixes in #1794 

Once those are merged, this can be rebased, edited to remove the cherry-picked commit and the git ref to the puppet_docker_tools PR, and merged.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+docker/puppetserver/*               text  eol=lf
+docker/puppetserver-standalone/*    text  eol=lf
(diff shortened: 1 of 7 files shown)


Iristyle:
Some of the helper script code that appeared here is still available in a gist at https://gist.github.com/Iristyle/a4d58cd77afeda8830523e4be15d23de

Iristyle:
Even though the Azure Pipelines check isn't showing here, it is green at https://puppetinc.visualstudio.com/puppetserver/_build/results?buildId=97 @underscorgan

PR #1798, opened 2018-09-12
Set up CI with Azure Pipelines

Iristyle:


Iristyle:
This was the resulting of resetting / recreating the Azure DevOps project... ignoring as its the same code as #1796

PR #1801, opened 2018-09-14
(maint) Manually bump puppet submodule for testing

justinstoller:


justinstoller:
This should get us to the head of 6.0.x so we can have a run through our pipelines without waiting for agent to pass (and have a better idea if there are any problems tomorrow morning).

justinstoller:
ugh, this isn't right

justinstoller:
Okay, the agent commit has gone far enough though the pipeline that this commit is now installable. I think this is ready to go now.


/cc @branan @Magisus

PR #1807, opened 2018-09-18
Don't build the latest tag of the container since 6.0.0 is out, and a few other fixups

underscorgan:


justinstoller:
Is this something we're going to need to toggle which each release?

underscorgan:
It needs to be toggled every X or Y release (basically when the current branch is no longer the latest).

PR #1810, opened 2018-09-19
(SERVER-2267) Get IP and DNS SANs to return in get-certificate-status

amysahli:


src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -377,6 +377,13 @@
   (mapv (partial str "DNS:")
         (utils/get-subject-dns-alt-names cert-or-csr)))
 
+(schema/defn subject-alt-names :- [schema/Str]
+  "Get the list of both DNS and IP alt names on the provided certificate or CSR.
+   Each name will be prepended with 'DNS:' or 'IP:'."
+  [cert-or-csr :- (schema/either Certificate CertificateRequest)]
+  (into (mapv (partial str "IP:") (utils/get-subject-ip-alt-names cert-or-csr))
+        (mapv (partial str "DNS:") (utils/get-subject-dns-alt-names cert-or-csr))))
```
amysahli: This is a WIP.  This would use my new "get-subject-ip-alt-names" method from jvm-ssl-utils to get the IP subject alt names and prepend them with "IP".

amysahli:
THIS is a WIP.  I'll add tests once my jvm-ssl-utils method is merged.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
amysahli: Note the similarity of this method to the method "dns-alt-names" above, which only got the dns-alt names.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1314,14 +1321,15 @@
...
-     :dns_alt_names (dns-alt-names cert-or-csr)
-     :fingerprint   default-fingerprint
-     :fingerprints  {:SHA1    (fingerprint cert-or-csr "SHA-1")
-                     :SHA256  default-fingerprint
-                     :SHA512  (fingerprint cert-or-csr "SHA-512")
-                     :default default-fingerprint}}))
+    {:name              subject
+     :state             (certificate-state cert-or-csr crl)
+     :dns_alt_names     (dns-alt-names cert-or-csr)
+     :subject_alt_names (subject-alt-names cert-or-csr)
```
amysahli: I didnt want to remove dns_alt_names from this payload, so I just added one that includes both IP AND DNS, which we can update our puppetserver-ca-cli to use.

justinstoller:
This looks like the right approach and code for part 1 of the ticket.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -612,17 +612,25 @@
...
-      (utils/subject-dns-alt-names (conj default-alt-names host-name) false))))
+  (let [split-alt-names (split-hostnames alt-names)
+        default-alt-names ["puppet"]
+        alt-names-list (reduce (fn [acc alt-name]
+                                 (if (str/starts-with? alt-name "IP:")
+                                   (update acc :ip conj (str/replace alt-name #"^IP:" ""))
+                                   (update acc :dns-name conj (str/replace alt-name #"^DNS:" ""))))
+                               {:ip [] :dns-name []} split-alt-names)
+        min-alt-names-list (if (empty? (get alt-names-list :ip))
+                             (dissoc alt-names-list :ip)
```
justinstoller: What happens if we don't `dissoc` here?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -674,26 +682,25 @@
...
+    (when-not (every? #{:dns-name :ip} name-types)
       (sling/throw+
         {:kind :invalid-alt-name
-         :msg (i18n/tru "Only DNS names are allowed in the Subject Alternative Names extension")}))
-
-    (doseq [name names]
-      (when (.contains name "*")
+         :msg (i18n/tru "Only DNS and IP names are allowed in the Subject Alternative Names extension")}))
+    (doseq [dns-name dns-names]
+      (when (.contains dns-name "*")
```
justinstoller: `*` is disallowed for only dns-names?

amysahli: As far as I know. IP addresses can't contain "*" and be valid anyway...i think.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
amysahli: Just tested and I don't need to do that (if it comes in as an empty vector, its just ignores it :)

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -1204,11 +1204,26 @@
...
+    (testing "correct subject alt name extensions are created for a master"
+      (let [dns-alt-names "onefish,twofish,DNS:threefish,IP:fourfish,fivefish,IP:sixfish"
+            exts-expected {:oid      "2.5.29.17"
+                            :critical false
+                            :value    {:dns-name ["onefish"
+                                                  "twofish"
+                                                  "threefish"
+                                                  "fivefish"
+                                                  "subject"]
+                                       :ip       ["fourfish"
```
justinstoller: I would expect this to fail with an invalid ip.

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -1340,22 +1355,28 @@
...
     (is (thrown+?
           [:kind :invalid-alt-name
-           :msg "Only DNS names are allowed in the Subject Alternative Names extension"]
-          (validate-dns-alt-names! {:oid "2.5.29.17"
-                                    :critical false
-                                    :value {:ip-address ["12.34.5.6"]}}))))
+           :msg "Only DNS and IP names are allowed in the Subject Alternative Names extension"]
+          (validate-subject-alt-names! {:oid "2.5.29.17"
+                                        :critical false
+                                        :value {:uri ["12.34.5.6"]}}))))
```
justinstoller: Did this fail before? Wondering since it was `:ip-address` instead of `:ip`?

Magisus:
Closing and reopening to see if it'll fix azure

justinstoller:
IIRC, we decided not to let azure pipelines block merging while we sort them out (new feature).

Magisus:
👍

justinstoller:
Normally it'd be nice to provide some validation of the IP addresses as well, though I don't know if we want to go down the route of parsing valid ipv4 and ipv6 addresses. Luckily the error from BC is actually fairly helpful:
```
2018-09-24 11:53:35,318 DEBUG [nREPL-worker-0] [p.p.certificate-authority] No CSR Attributes configuration file found at /Users/justin/.puppetlabs/etc/puppet/csr_attributes.yaml, CSR Attributes will not be loaded
2018-09-24 11:53:35,330 ERROR [nREPL-worker-0] [p.t.internal] Error during service init!!!
java.lang.IllegalArgumentException: IP Address is invalid
        at org.bouncycastle.asn1.x509.GeneralName.<init>(Unknown Source)
```

For:
```
puppetserver justin ((remotes/amysahli/SERVER-2267) *$) ‖ cat ~/.puppetlabs/etc/puppet/puppet.conf

[main]
certname = localhost
[agent]
server = localhost
[master]
dns_alt_names = foo,DNS:foo.example.com,IP:1.1.1.1,IP:baz
```

Magisus:
Yeah I was planning to rely on the SSL code to do the verification, since it does error helpfully.

Magisus:
I've put up https://github.com/puppetlabs/puppetserver/pull/1822 that retargets this at 6.0.x. I think we should merge that instead of this, and then backport to 5.3.x later along with the other changes around signing certs with alt names (new settings).

justinstoller:
Maggie rebased this onto 6.0.x for the upcoming 6.0.1 release in #1822. Closing as now these commits will get into master via a merge up.

PR #1811, opened 2018-09-19
(maint) Update CA conf file docs

Magisus:


documentation/infrastructure_crl.markdown
```diff
@@ -0,0 +1,15 @@
+---
+layout: default
+title: "Puppet Server: Intermediate CA"
+canonical: "/puppetserver/latest/infrastructure_crl.html"
+---
+
+# Infrastructure certificate revocation list
+
+The Puppet Server CA can create a smaller CRL that contains only revocations of those nodes that agents are expected to talk to during normal operations, for example, compile masters or hosts that agents connect to as part of agent-side functions. Puppet Server CA can distribute that CRL to agents, rather than the CRL it maintains with all node revocations.
```
amysahli: I don't love "smaller".  I guess "leaner" would be better, but then it sounds like some kind of meat :(

documentation/subcommands.markdown
```diff
@@ -32,6 +33,28 @@ lein <subcommand> -c /path/to/puppetserver.conf [--] [<args>]
...
+- `import`: import the CA's key, certs, and CRLs
+- `list`: list all certificate requests
+- `revoke`: revoke a given certificate
+- `sign`: sign a given certificate
+
+Because these commands utilize Puppet Server’s API, all except `setup` and `import` need the server to be running in order to work.
+
+Since these commands are shipped as a gem alongside Puppet Server, it can be updated out-of-band to pick up improvements and bug fixes. To upgrade it, run this command: `/opt/puppetlabs/puppet/bin/gem install -i /opt/puppetlabs/puppet/lib/ruby/vendor_gems puppetserver-ca`
+
+**Note:** These commands are available in Puppet 5, but in order to use them, you must update Puppet Server’s `auth.conf` to include a rule allowing the master’s certname to access the `certificate_status` and `certificate_statuses` endpoints. The same applies to upgrading in open source Puppet: if you're upgrading from Puppet 5 to Puppet 6 and are not regenerating your CA, you must whitelist the master’s certname. See [Puppet Server Configuration Files: auth.conf](/puppetserver/latest/config_file_auth.html) for details on how to use `auth.conf`.
```
amysahli: I wish we could link directly to an example of how to whitelist the master's certname instead of the general config file.   I'm imagining myself as a user sorting endlessly through auth conf file materials :(

Magisus: I can just put an example in here.

documentation/infrastructure_crl.markdown
Magisus: "trimmed-down"?

justinstoller: it also makes sense sans adjective

Magisus:
Updated.

documentation/infrastructure_crl.markdown
```diff
@@ -0,0 +1,15 @@
+---
+layout: default
+title: "Puppet Server: Intermediate CA"
+canonical: "/puppetserver/latest/infrastructure_crl.html"
+---
```
Magisus: Can we get this page added to the TOC?

justinstoller: Yes

justinstoller: I think the sidebar nav is controlled by _puppetserver_nav.html in this directory?

Magisus: Okay @MFredette wasn't sure. I can update that then.

justinstoller: Grr... Sorry, it looks like I'm wrong... I'm not sure what this is used for (maybe it was used before they switched doc sites? maybe we should delete it) From looking at the content on the live site it looks like the index file in this directory includes a TOC on Puppet Server's index page, while the left hand nav is controlled by [puppet-docs](https://github.com/puppetlabs/puppet-docs/blob/master/source/puppet/6.0/_puppet_toc.html#L105). This is both side-by-side https://puppet.com/docs/puppetserver/6.0/index.html

Magisus: Well I think updating the index is good, so I'm gonna leave this. I would also kind of like to update the left nav, I feel like it could be organized more clearly.

PR #1822, opened 2018-09-24
(SERVER-2267) Enabling IP subject alt names

Magisus:


Magisus:
I think we should supersede https://github.com/puppetlabs/puppetserver/pull/1810

PR #1829, opened 2018-10-02
Added release notes for 6.0.1.

MFredette:
Ready for merge when Puppet Platform release goes out.


documentation/known_issues.markdown
```diff
@@ -9,9 +9,13 @@ For a list of all known issues, visit our [Issue Tracker](https://tickets.puppet
 
 Here are a few specific issues that we're aware of that might affect certain users:
 
-## Server-side Ruby gems might need to be updated for upgrading from JRuby 1.7
+## puppetserver 6.0.0 dies with `java.lang.OutOfMemoryError: Metaspace`
```
Magisus: Can we maybe break this into two sections? They are both related to JRuby 9k, but the two pieces of advice (gem pathing vs. `ReservedCodeCache`) are sort of unrelated.

justinstoller: The codecache info is already below in "## Potential JAVA ARGS settings". Which is maybe a bad title for it. It might be nice to also call out that these are *broken* things that are ever going to be fixed, they're just gotchas in changing requirements that might trip folks up with upgrading.

MFredette: do you mean "never going to be fixed" Justin?

justinstoller: Actually, this looks like the revision of this section that was added in 84e77e1 and has been subsequently updated. I think this PR is pulling in an older version of the file?

documentation/known_issues.markdown
```diff
@@ -7,13 +7,21 @@ canonical: "/puppetserver/latest/known_issues.html"
 
 For a list of all known issues, visit our [Issue Tracker](https://tickets.puppet.com/browse/SERVER).
 
-Here are a few specific issues that we're aware of that might affect certain users:
+Here are some specific issues that we're aware of that might affect certain users, depending on their Puppet setup. Familiarize yourself with these issues, particularly before you upgrade to a new version of Puppet Server.
+
+## puppetserver 6.0.0 dies with `java.lang.OutOfMemoryError: Metaspace`
+
+When upgrading from Puppet Server 5 using JRuby 1.7 (9k was optional in those releases), Server-side gems (those installed manually with `puppetserver gem` command or using the puppetserver_gem package provider) may need to be updated to work with the newer JRuby. In most cases gems do not have APIs that break when upgrading from the Ruby versions implemented between JRuby 1.7 and JRuby 9k, so there may be no necessary updates, but users should be aware. Two notable exceptions are that the autosign gem should be 0.1.3 or later and yarddoc must be 0.9 or later.
```
Magisus: I'll put something up to remove this duplication after the release.

justinstoller:
I reset back to the 6.0.1 release notes commit, and then only commit back the changes to the release notes, leaving behind the known issues updates. I think force pushed to Michelle's branch. Sorry, that didn't retain the commit author, this was the easiest/fastest way I know how to do it.

PR #1831, opened 2018-10-02
(SERVER-2331) Merge 6.0.1 into master

justinstoller:


project.clj
```diff
@@ -1,4 +1,4 @@
-(def ps-version "6.0.1-SNAPSHOT")
+(def ps-version "6.0.2-SNAPSHOT")
```
Magisus: Do we want to update this to `6.1.0-SNAPSHOT`?

dev-resources/puppetlabs/services/certificate_authority/certificate_authority_int_test/infracrl_test/master/conf/ssl/ca/infra_serials
```diff
@@ -1 +1 @@
-23
+
```
Magisus: Why is this empty now?

justinstoller: infra-crl code should populate the infra_serials file using the certnames in the infra_inventory.txt. That this was set to the serial number of the old certname was an oversight in our original testing. Removing this provides better integration testing (and broke when I regenerated the certs).

project.clj
justinstoller: Yeah, I was just thinking about getting the merge up done, but makes sense to do it now.

justinstoller: updated

PR #1832, opened 2018-10-03
(SERVER-2322) Settings backport

Magisus:
This cherry-picks all the commits from the 6.0.x branch related to adding settings for allowing subject alt names and auth extensions when signing certs via the `certificate_status` endpoint.

It also adds the commit where we started shipping a `ca.conf` file, so that the same docs can apply to both 5.x and 6.x about where to specify these settings.


Magisus:
I tried this out, enabling the setting in my `dev/puppetserver.conf`, submitting a CSR with alt names, then signing it with `puppetserver ca sign`. Appears to work.

ezbake/config/conf.d/ca.conf
```diff
@@ -0,0 +1,10 @@
+certificate-authority: {
+    # allow CA to sign certificate requests that have subject alternative names.
+    # allow-subject-alt-names: false
+
+    # allow CA to sign certificate requests that have authorization extensions.
+    # allow-authorization-extensions: false
+
+    # enable the separate CRL for Puppet infrastructure nodes
+    # enable-infra-crl: false
```
justinstoller: I don't think we want to ship this file with this setting?

Magisus: Good catch.

PR #1839, opened 2018-10-08
(SERVER-2347) Fix package installation docs

Magisus:
This commit updates the instructions for setting up the Puppet CA in
Puppet 6.


Magisus:
In this file there are also extensive instructions for how to stop the puppet master service from various kinds of installs. Are all of these still relevant? I know we don't have a Webrick master anymore, but I suppose this is supposed to help people upgrading from old installs, which might still be of this style?

Magisus:
Cool, I'll clean that up.

Magisus:
Let me know if I missed anything.

PR #1844, opened 2018-10-10
(SERVER-2350) Register the CA service with the status service

Magisus:
This commit registers the CA service with the StatusService. It does not
register the DisabledCAService, so if that is enabled instead of the
regular CA service, no CA info will be available from the status
endpoint.


Magisus:
Figuring out how to add tests to this.

Magisus:
I think fixing the tests that failed when this was added is probably sufficient testing for now, given that we're not returning any special metadata based on level.

justinstoller:
Yeah, those are crappy tests to have failed because of this change, but considering there's almost no new logic being added I'm fine with letting those integration-y tests prove it's wired up correctly.

PR #1849, opened 2018-10-17
(maint) Redirect `git pull --unshallow` to /dev/null

underscorgan:


.travis.yml
```diff
@@ -3,7 +3,7 @@ lein: 2.8.1
 dist: trusty
 jdk:
   - oraclejdk8
-  - oraclejdk10
+  - openjdk10
```
underscorgan: FYI @justinstoller @Magisus  oraclejdk10 is no more, it eol'd in september and its downloads are no longer available.

PR #1850, opened 2018-10-17
Update health check script to use any hostname

nalyanyam:
Update healthcheck.sh
This allows HEALTHCHECK routine to work with any container hostname.


docker/puppetserver-standalone/healthcheck.sh
```diff
@@ -3,11 +3,12 @@
 set -x
 set -e
 
+export PUPPETHOST=$(hostname -f); \
```
Iristyle: Thanks for the PR @nalyanyam.

What is the particular problem that you've run into, and what does your environment look like?

nalyanyam: The problem is that this health check routine is bound to the Puppetserver hostname being named "puppet". In production we do use "puppet" hostname as an alias, but we have various environments where we use different hostnames for puppet server. I would therefore suggest that the code is modified to work with any hostname. otherwise the health check hangs for ever.

Iristyle: Ah, but in this case it should not be, based on how the curl call uses [--resolve](https://curl.haxx.se/docs/manpage.html#--resolve). This means that the name `puppet` is resolving to `127.0.0.1` based on `--resolve '${PUPPETHOST}:8140:127.0.0.1'`. The name `puppet` should actually not be relevant here.

We're also not sure, generally speaking, of how using `hostname -f` will behave when containers are hosted in various cloud providers with different networking setup. For instance, I know that on a host VM in Azure, behavior can vary based on OS, and whether the resource has publicly bound addresses / hostnames.

Did you actually ran into a problem running the healthcheck, or did you just see this code and think it needed to be changed? If you did have some failures running the healthcheck in your environment, do you have any logs or other supporting info?

Thanks!

nalyanyam: Yes, i ran into issues. the health check hangs, and the puppetserver does not therefore successfully start. The status shows (Health: Starting).
I thought puppet hostname was not necessary, but seems not to be the case because the query is using SSL certificates which should be specific to the hostname.

It should be easy to reproduce. Just start puppetserver docker with any other hostname  (not "puppet"). Was using docker compose file where i specified the hostname.

I was also not certain how this change will behave on different platforms and you are right, we might need a more generic health check to be platform-agnostic.

underscorgan:
```suggestion
hostname=$(puppet config print certname)
```

docker/puppetserver-standalone/healthcheck.sh
```diff
@@ -3,11 +3,12 @@
...
 curl --fail -H 'Accept: pson' \
---resolve 'puppet:8140:127.0.0.1' \
+--resolve '${PUPPETHOST}:8140:127.0.0.1' \
```
underscorgan:
```suggestion
--resolve '${hostname}:8140:127.0.0.1' \
```

docker/puppetserver-standalone/healthcheck.sh
```diff
@@ -3,11 +3,12 @@
...
 --cert   $(puppet config print hostcert) \
 --key    $(puppet config print hostprivkey) \
 --cacert $(puppet config print localcacert) \
-https://puppet:8140/${PUPPET_HEALTHCHECK_ENVIRONMENT}/status/test \
+https://${PUPPETHOST}:8140/${PUPPET_HEALTHCHECK_ENVIRONMENT}/status/test \
```
underscorgan:
```suggestion
https://${hostname}:8140/${PUPPET_HEALTHCHECK_ENVIRONMENT}/status/test \
```

docker/puppetserver-standalone/healthcheck.sh
underscorgan: Ok, it looks like the problem here is that you're setting dns alt names but those names don't include 'puppet' (which is the default behavior)

```
root@testserver:/# curl --fail -H 'Accept:pson' --resolve 'puppet:8140:127.0.0.1' --cert $(puppet config print hostcert) --key $(puppet config print hostprivkey) --cacert $(puppet config print localcacert) https://puppet:8140/production/status/test
curl: (51) SSL: no alternative certificate subject name matches target host name 'puppet'
root@testserver:/# puppetserver ca list --all
Signed Certificates:
    testserver   (SHA256)  7D:F8:1A:67:E8:ED:D8:9F:93:FE:82:D4:E9:E5:88:35:AD:6E:50:A4:E3:82:3F:24:F5:6C:32:AB:C7:A9:AB:E1      alt names: ["DNS:foo", "DNS:testserver"]
```

Since `hostname -f` is unreliable, I had a few suggested changes for this.

Also, if you could rebase your work on the 5.3.x so we can get it into all of the containers we're currently publishing that would be helpful.

Thank you for this contribution!

nalyanyam:
I closed this PR because of too many conflicts and opened a new one with the suggested changes.
https://github.com/puppetlabs/puppetserver/pull/1854

PR #1853, opened 2018-10-18
Merge 5.3.x -> 6.0.x

Magisus:


justinstoller:
Do we want to include these[1] changes into 6.0.x? Or should we switch to jdk11?


1. https://github.com/puppetlabs/puppetserver/blob/master/.travis.yml#L4-L11

Magisus:
Would including that get Travis passing?

justinstoller:
I think so

Magisus:
I'll pull them in. We can do the jdk11 thing separately.

PR #1854, opened 2018-10-19
Update health check script to use any hostname

nalyanyam:
This allows HEALTHCHECK routine to work with any container hostname.

### DIFF
--- a/docker/puppetserver-standalone/Dockerfile
+++ b/docker/puppetserver-standalone/Dockerfile
@@ -51,12 +51,13 @@ ENTRYPOINT ["dumb-init", "/docker-entrypoint.sh"]
 CMD ["foreground"]
 
 HEALTHCHECK --interval=10s --timeout=10s --retries=90 CMD \
+  hostname=$(puppet config print certname) && \
   curl --fail -H 'Accept: pson' \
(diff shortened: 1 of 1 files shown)


docker/puppetserver-standalone/Dockerfile
```diff
@@ -51,12 +51,13 @@ ENTRYPOINT ["dumb-init", "/docker-entrypoint.sh"]
 CMD ["foreground"]
 
 HEALTHCHECK --interval=10s --timeout=10s --retries=90 CMD \
+  hostname=$(puppet config print certname)
```
underscorgan:
```suggestion
  hostname=$(puppet config print certname) && \
```

nalyanyam:
Updated.

nalyanyam:
Thanks @underscorgan

nalyanyam:
@underscorgan , was just checking healthcheck.sh script again, and it looks like we should have put double quotes. the script still works but it does not interpolate the hostname in resolve line.
-   --resolve '${hostname}:8140:127.0.0.1' \
+  --resolve "${hostname}:8140:127.0.0.1" \

Should i send another PR to fix it?

underscorgan:
@nalyanyam ah, yes, good catch. PR would be great.

PR #1858, opened 2018-10-22
Provide ability to disable CA service

nalyanyam:
This maybe necessary when using external CA or in Multi-master
Puppet configuration setup

Not sure if this is the right approach. We are trying to use Puppetserver docker images to create a stack of Puppet servers and need a way to disable PuppetCA on Compile masters and only allow CA service on some docker containers designated as PuppetCAs.
This was just a quick hack. Please let me know in case there's a better way?

### DIFF
--- a/docker/puppetserver-standalone/Dockerfile
+++ b/docker/puppetserver-standalone/Dockerfile
@@ -23,7 +23,7 @@ LABEL org.label-schema.maintainer="Puppet Release Team <release@puppet.com>" \
       org.label-schema.dockerfile="/Dockerfile"
 
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends wget=1.17.1-1ubuntu1 ca-certificates && \
+    apt-get install -y --no-install-recommends wget=1.17.1-1ubuntu1 ca-certificates netcat && \
     wget https://apt.puppetlabs.com/puppet5-release-"$UBUNTU_CODENAME".deb && \
     wget https://github.com/Yelp/dumb-init/releases/download/v"$DUMB_INIT_VERSION"/dumb-init_"$DUMB_INIT_VERSION"_amd64.deb && \
     dpkg -i puppet5-release-"$UBUNTU_CODENAME".deb && \
--- a/docker/puppetserver-standalone/docker-entrypoint.sh
+++ b/docker/puppetserver-standalone/docker-entrypoint.sh
@@ -26,12 +26,6 @@ if test -n "${PUPPETSERVER_HOSTNAME}"; then
   /opt/puppetlabs/bin/puppet config set server "$PUPPETSERVER_HOSTNAME"
 fi
 
-# Configure puppet to use a certificate autosign script (if it exists)
-# AUTOSIGN=true|false|path_to_autosign.conf
-if test -n "${AUTOSIGN}" ; then
-  puppet config set autosign "$AUTOSIGN" --section master
-fi
-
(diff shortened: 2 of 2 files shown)


underscorgan:
@lutter I know you've been working on scaling/compile master stuff...

lutter:
Yes, I have similar stuff happening. I'll put up a PR today against pupperware so we can compare. A couple things I spotted is that it uses `nc -z` to check if the CA server is up which doesn't work if the CA is behind a load balancer (it succeeds when the lb is up, even if the backend is still down), that you have to set both `CA_DISABLE` and `CA_SERVER` even though there's only one correct combination, and the use of `sed` to disable the CA rather than replacing `ca.cfg` wholesale.

nalyanyam:
@underscorgan @lutter   Any progress on this PR? Anything we can update or change?

lutter:
@nalyanyam sorry for the long delays. I am trying to figure out a way to combine the functionality from this PR with the functionality that needs to go in for [this Pupperware PR](https://github.com/puppetlabs/pupperware/pull/24) It's a little messy since using the container with an external CA and using it behind a load balancer have some things in common and disagree in some respects.

lutter:
@nalyanyam rereading your initial comment, it seems what you are after is very similar to what I am trying to do in that [Pupperware PR](https://github.com/puppetlabs/pupperware/pull/24)

Would the following work for you:

- put all your Puppetmasters behind a load balancer
- terminate HTTPS in the load balancer (and have the lb talk HTTP to the puppetservers behind it)
- have one puppetserver act as the CA
- bring up other puppetservers as compile masters without a CA

Maybe if we can figure out where our needs differ, we can come up with a solution that works for both scenarios.

nalyanyam:
@lutter  
Yes, the above description captures more or less what we are trying to achieve.
Our test implementation has Puppetmasters (compile masters) behind a load balancer. We are using docker swarm with Traefik reverse proxy / lb ( https://traefik.io which seemlessly integrates with docker). 

Experienced issues with passing https traffic to the backend puppetservers through traefik, but worked around this by using docker swarm itself as a main lb. I think your strategy of terminating https on lb and have lb talk http to backend puppetservers is a good one and would work here as well.

The setup has a separate docker swarm for 2 Puppet CA servers, first because we need redundancy for CA server as well. The second reason is because it is not trivial to separate Puppet compile master and Puppet CA traffic (port 8140) in a single docker swarm.

We are still in test phase so quite open to test different lb  / or design strategies and i would be happy to test your https://github.com/puppetlabs/pupperware/pull/24 PR setup. 

The only issue i see is that this change binds users to use a specific HAProxy

mungo312:
Any news about this ? We need this function for a multi-master setup with every master having its own agents. Or is there a way to get a multi-master setup running with CA enabled in all masters?

underscorgan:
@mungo312 I've been working on this functionality in #1930 to better align with the [Pupperware scaling PR](https://github.com/puppetlabs/pupperware/pull/24), does that look like it would cover your use case?

underscorgan:
@nalyanyam thanks for this PR! Your work was helpful when I was able to revisit disabling the CA. https://github.com/puppetlabs/puppetserver/pull/1956 has been merged into master which adds the ability to disable the CA using the `CA_ENABLED` environment variable. This change is in the 6.2.0 tagged image.

nalyanyam:
@underscorgan Thanks for working on this. I will try out version 6.2.0 and see how it goes.

PR #1860, opened 2018-10-23
(maint) Remove repo_link_target and nonfinal_repo_link_target params

mwaggett:
This commit removes the `repo_link_target` and `nonfinal_repo_link_target`
params from build_defaults. These params are used by the packaging repo to link
our rolling puppet repo to the latest repo. Since puppet6 is now the latest
repo, we need to remove this or else every time we ship, the puppet repo will
get linked back to puppet5.


Magisus:
I believe you :P

PR #1863, opened 2018-10-25
[SPIKE] Use ruby to consolidate build and build.ps1

underscorgan:
Still need to update azure to use the new stuff, but here's an example of what the scripts could look like

### DIFF
--- a/docker/distelli-manifest.yml
+++ b/docker/distelli-manifest.yml
@@ -1,4 +1,9 @@
 pe-and-platform/puppetserver:
+  PreBuild:
+    - bundle install --path .bundle/gems
+    - bundle exec ci/pre_build
   Build:
-    - docker login -u "$DISTELLI_DOCKER_USERNAME" -p "$DISTELLI_DOCKER_PW" "$DISTELLI_DOCKER_ENDPOINT"
-    - /bin/bash -x ./ci/build
+    - bundle exec ci/build
+    - bundle exec ci/test
+  AfterBuildSuccess:
+    - bundle exec ci/publish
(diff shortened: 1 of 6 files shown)


underscorgan:
Windows stuff hasn't been backported to 5.3.x, but here's what the changes looked like against 6.0.x to get azure using these scriptlets (https://github.com/underscorgan/puppetserver/commit/f33139000cde75a87b578fc565d9a4e02d53c927), and a passing build https://dev.azure.com/PuppetInc/puppetserver/_build/results?buildId=348&view=logs

docker/distelli-manifest.yml
```diff
@@ -1,4 +1,9 @@
...
+  PreBuild:
+    - bundle install --path .bundle/gems
+    - bundle exec ci/pre_build
   Build:
-    - docker login -u "$DISTELLI_DOCKER_USERNAME" -p "$DISTELLI_DOCKER_PW" "$DISTELLI_DOCKER_ENDPOINT"
-    - /bin/bash -x ./ci/build
+    - bundle exec ci/build
+    - bundle exec ci/test
+  AfterBuildSuccess:
+    - bundle exec ci/publish
```
underscorgan: Tested these changes with only a changed namespace in the distelli-manifest at https://pipelines.puppet.com/release-engineering/builds/469030. The failure is expected, the job isn't set up with credentials to push to dockerhub.

underscorgan:
We're going to maintain separate build scripts, and break apart the linux-y one into separate steps, so closing this for now.

PR #1866, opened 2018-10-30
(maint) use JDK11 instead of 10

justinstoller:


justinstoller:
for some reason travis didn't run on that last push...

justinstoller:
Nice! I don't know what the issue is that Headius is talking about but it looks like we're now passing against JDK11 in Travis :shipit: At this point I'm happy to say we have experimental JDK11 support in Puppet Server.

PR #1867, opened 2018-10-31
(maint) Improve container testing

underscorgan:
Rather than testing the container has the expected files installed, make
sure the puppetserver container can start and we can run an agent
against it. Also test that we can execute r10k on the puppetserver
container.


underscorgan:
Tested on windows at https://github.com/puppetlabs/puppetserver/pull/1873

nwolfe:
Do we really need an entirely different spec directory (with almost identical contents) for both Dockerfiles?

Looks like the only difference in the images is which package version of puppetserver is installed?

I can look into this further, but I'm tempted to generalize these a bit further so we don't have the duplication with only some string value differences.

nwolfe:
Some thoughts...

I think what's bugging me is the number of permutations we have for Puppet Server and other projects.

There's:

* puppetserver
* puppetserver-standalone
* puppetserver.nightly
* puppetserver-standalone.nightly

The actual differences seem relatively small, yet we have all this duplication across our build and test scripts.

I'll take a closer look next week.

underscorgan:
@nwolfe The test scripts can definitely be combined and moved up a directory, the differences are unimportant. It'll be much easier to get that change in after https://github.com/puppetlabs/puppetserver/pull/1869 goes in since puppet_docker_tools has some assumptions about where the tests live. 

I feel like a lot of the build script duplications are dealt with with the makefile updates, I'm cleaning up the commits in #1869 so we can get that in / try to get that standardized.

Iristyle:
Testing is merged into the Azure 5.3.x branch now @underscorgan, so I think we're good for a rebase!

PR #1869, opened 2018-11-06
(maint) build using a makefile

underscorgan:


nwolfe:
To get away from all the `foo` and `foo-nightly` targets, I think we could...

- Make "is nightly" an ENV variable or something dynamic
- Combine the `foo` and `foo-nightly` targets into one and condition behavior on the ENV variable
- Make the `prep` task a NOOP if it's nightlies, so all the targets can still depend on it

docker/Makefile
```diff
@@ -0,0 +1,66 @@
+git_describe := $(shell git describe)
+version := $(shell echo $(git_describe) | sed 's/-.*//')
```
Iristyle: I'm not sure if this Makefile would also target Windows, but we won't have access to things like `sed`, or `/dev/null` (Windows has `NUL` which is similar there, but nothing as full-featured as `sed`)

underscorgan: @Iristyle I thought make didn't really work on windows?

underscorgan:
yeah, I think with the existing scripts we use an IS_NIGHTLY env variable, I can collapse those

docker/Makefile
Iristyle: It's more that it's not a default tool and fairly uncommon (except to maybe C++ folks). CMake would be the more "Microsoft approved" way of doing things, and even then, you at least have to install a choco package to get going.

I wasn't sure if the intent was to use `make` everywhere, but if this is only for non-Windows (where make is far more common), then the Windows concerns aren't that important.

underscorgan:
@nwolfe how about now?

docker/Makefile
```diff
@@ -0,0 +1,54 @@
...
+endif
+
+test: prep
+	@PUPPET_TEST_DOCKER_IMAGE=puppet/puppetserver-standalone:$(version) rspec puppetserver-standalone/spec
+	@PUPPET_TEST_DOCKER_IMAGE=puppet/puppetserver:$(version) rspec puppetserver/spec
+
+publish: prep
+	@docker push puppet/puppetserver-standalone:$(version)
+	@docker push puppet/puppetserver:$(version)
+ifeq ($(IS_LATEST),true)
```
nwolfe: Where does this `IS_LATEST` come from? 

Hadn't noticed this before, thought there was only `IS_NIGHTLY`.

underscorgan: IS_LATEST is an environment variable that we pass in during the pipelines job. Initially we were needing to edit jobs to get the latest tag built for the correct branch and that was getting really weird with mergeups

underscorgan:
Strawpoll: @stahnma @Iristyle @lutter @nwolfe Makefile for unix-y builds 👍 / 👎 

This would leave windows builds with the .ps1 file for a more windows-native experience and is a step in the direction of using straight docker commands instead of puppet_docker_tools since that just makes it harder to figure out what's going on.

stahnma:
I always ⬆️ Makefiles

docker/Makefile
```diff
@@ -0,0 +1,53 @@
+git_describe := $(shell git describe)
+vcs_ref := $(shell git rev-parse HEAD)
+build_date := $(shell date -u +%FT%T)
+hadolint_available := $(shell hadolint --help > /dev/null 2>&1; echo $$?)
+hadolint_command := hadolint --ignore DL3008 --ignore DL3018 --ignore DL4000 --ignore DL4001
+hadolint_container := hadolint/hadolint:latest
+
+ifeq ($(IS_NIGHTLY),true)
+	dockerfile := Dockerfile.nightly
+	version := puppet6-nightly
```
Iristyle: Since the later `push` only varies by the version, I would set a var here like this:

```make
container_version := latest
```

docker/Makefile
Iristyle: And set it to `$(version)` if not already set to `latest`

```make
container_version ?= $(version)

@docker push puppet/puppetserver-standalone:$(container_version)
@docker push puppet/puppetserver:$(container_version)
```

docker/Makefile
underscorgan: But if the container has `IS_LATEST` set we push both the numbered tag and the latest tag in separate push steps. I'm not following how this would help.

Iristyle: Yeah, sorry. I completely misread there and my comment is nonsensical.

PR #1873, opened 2018-11-09
(maint) Improve container testing

underscorgan:
Rather than testing the container has the expected files installed, make
sure the puppetserver container can start and we can run an agent
against it. Also test that we can execute r10k on the puppetserver
container.


underscorgan:
mostly want to make sure this doesn't 💥 windows

docker/puppetserver-standalone/spec/puppetserver_spec.rb
```diff
@@ -29,14 +31,11 @@ def puppetserver_health_check(container)
 
   it 'should start puppetserver successfully' do
     status = puppetserver_health_check(@container)
-    while status == 'starting'
+    while (status == 'starting' || status == "'starting'")
```
Iristyle: This looks familiar! ;0

docker/puppetserver/spec/puppetserver_spec.rb
```diff
@@ -29,14 +31,11 @@ def puppetserver_health_check(container)
...
-    while status == 'starting'
+    while (status == 'starting' || status == "'starting'")
       sleep(1)
       status = puppetserver_health_check(@container)
     end
-    unless status == 'healthy'
-      puts %x(docker logs #{@container})
-    end
-    expect(status).to eq('healthy')
+    expect(status).to match(/\'?healthy\'?/)
```
Iristyle: Do we get anything useful out of `docker logs` when things fail?

docker/puppetserver-standalone/spec/puppetserver_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      error_message = <<-MSG
+* * * * *
+  PUPPET_TEST_DOCKER_IMAGE environment variable must be set so we
+  know which image to test against!
+* * * * *
+      MSG
+      fail error_message
+    end
+
+    @container = %x(docker run --rm --detach --env DNS_ALT_NAMES=puppet --name puppet.test --hostname puppet.test --publish 8140:8140 #{@image}).chomp
```
Iristyle: Should we randomize the 8140 locally?

docker/puppetserver/spec/puppetserver_spec.rb
underscorgan: mm, yeah, that would probably be helpful for actually figuring out what went wrong if something failed, will add that back

docker/puppetserver-standalone/spec/puppetserver_spec.rb
underscorgan: Actually, since we're linking the agent to the master container I think we can get away without publishing the ports ... testing!

underscorgan: Yup, looks like that works \o/

docker/puppetserver-standalone/spec/puppetserver_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+describe 'puppetserver-standalone container' do
+  def puppetserver_health_check(container)
+    status = %x(docker inspect "#{container}" --format '{{.State.Health.Status}}').chomp
+    STDOUT.puts "queried health status of #{container}: #{status}"
+    return status
+  end
+
+  before(:all) do
+    %x(docker pull puppet/puppet-agent-alpine:latest)
+    @image = ENV['PUPPET_TEST_DOCKER_IMAGE']
```
Iristyle: Where is this value being set from? Calculated inside `puppet_docker_tools` or something?

underscorgan: yeah, that's calculated in `puppet_docker_tools` https://github.com/puppetlabs/puppet_docker_tools/search?q=PUPPET_TEST_DOCKER_IMAGE&unscoped_q=PUPPET_TEST_DOCKER_IMAGE

In moving to Makefile, it just gets set there before calling the tests.

PR #1874, opened 2018-11-09
(SERVER-2374) Update to beaker 4

Magisus:
Update to beaker 4 and beaker-puppet 1, along with some minor updates to the pre-suites to fix Windows installs.

### DIFF
--- a/acceptance/scripts/generic/testrun.sh
+++ b/acceptance/scripts/generic/testrun.sh
@@ -2,7 +2,7 @@
 set -x
 
 export GEM_SOURCE="https://artifactory.delivery.puppetlabs.net/artifactory/api/gems/rubygems/"
-export GENCONFIG_LAYOUT="${GENCONFIG_LAYOUT:-redhat6-64ma-ubuntu1404-64a-windows2008r2-64a}"
+export GENCONFIG_LAYOUT="${GENCONFIG_LAYOUT:-redhat6-64ma-ubuntu1604-64a-windows2012r2-64a}"
 export BEAKER_TESTSUITE="${BEAKER_TESTSUITE:-acceptance/suites/tests}"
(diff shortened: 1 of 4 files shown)


Magisus:
Not sure if we want this targeted further back (2.x or 5.1).

acceptance/scripts/generic/testrun.sh
```diff
@@ -2,7 +2,7 @@
 set -x
 
 export GEM_SOURCE="https://artifactory.delivery.puppetlabs.net/artifactory/api/gems/rubygems/"
-export GENCONFIG_LAYOUT="${GENCONFIG_LAYOUT:-redhat6-64ma-ubuntu1404-64a-windows2008r2-64a}"
+export GENCONFIG_LAYOUT="${GENCONFIG_LAYOUT:-redhat6-64ma-ubuntu1604-64a-windows2012r2-64a}"
```
Magisus: This isn't necessary, I just question whether or not we still support those old OSes every time I look at this file, so I updated them. Happy to put it back if we think the previous versions were better for some reason.

justinstoller: Ha, this is great. When I use this script I always update it to just be a single redhat7 vm (unless the tests I'm working on are platform specific or need a separate master and agent).

Magisus: Do we want to make it a smaller matrix? Though to be fair running Windows caught something for me here.

justinstoller: I'm fine leaving it larger and having folks have to pare it down if they know they don't need those vms.

justinstoller:
This looks good, running the script locally now (has gotten past the install and looks good so far)

PR #1876, opened 2018-11-09
Merge 5.3.x -> 6.0.x

Magisus:


justinstoller:
I don't know if there should be docker changes here since there's commits relating to them? Other than that everything looks good.

Magisus:
So I did some spelunking, and it looks like on this branch, they've factored out the healthcheck script into its own file, which had already incorporated the changes that they added in the 5.3.x branch (quoting fixes).

justinstoller:
Sounds good to me. And travis doesn't care about anything else in this PR.

PR #1878, opened 2018-11-14
(maint) remove cljs code and assets

justinstoller:
The developer dashboard receives little use and the continued
maintenance of it has become enough of a burden that we've re-evaluated
how much value we and the community are receiving from it. That
evaluation pointed out that our Support and Services departments own a
solution that leverages graphite within a docker container to query
a larger collection of data from the status endpoint. They also own a
polling job that can be installed to ensure historical data. They have
open sourced these tools and they are now the recommended way for users
to solve the problems that this dashboard solves.


justinstoller:
I wanted to put this up as it might be relevant to recent efforts to shrink our uberjar size. We probably want to have more discussion before actually landing this code (do we want to put it into an older branch?).

This reduces the uberjar size by 21M (32%)
```
puppetserver justin (no-cljs $) ‖ ls -lh puppet-server-uberjar-*
-rw-r--r--  1 justin  staff    65M Nov 14 10:22 puppet-server-uberjar-master.jar
-rw-r--r--  1 justin  staff    44M Nov 14 10:50 puppet-server-uberjar-no-cljs.jar
```

Sharpie:
Should we link this to [SERVER-2324](https://tickets.puppetlabs.com/browse/SERVER-2324) as the resolution for the developer dashboard not working?

justinstoller:
Probably, I did this in the context of Puppercon and reducing the size of our container image. I think folks would be alright with this as the resolution to SERVER-2324 (and worded the commit as such) though that would have this targeted at 5.3.x or earlier. Wanted to put this up if folks wanted to use it as part of their container work but I'll wait on the SERVER-2324 until Maggie gets back from RubyConf.

justinstoller:
Chatted with @Magisus and retargeted this for 5.3.x

justinstoller:
There we go, trying to get travis to do something sensible

PR #1879, opened 2018-11-14
(maint) Use AOT for uberjar profile

justinstoller:


justinstoller:
Bit of WIP, I need to make sure we can still ezbake and take up in PE correctly.

justinstoller:
So far I haven't seen any issue with taking this work up in PE or being able to create packages, I'm going to do a quick look at metadata elision and direct linking compiler optimizations and then do an acceptance run.

justinstoller:
I updated this to elide [:doc :file :line :added] and use direct linking per the clojure compilation page and then ran the beaker smoke tests locally. They passed. Normally, I'd think that is good enough testing pre-Jenkins (those tests doesn't exercise Puppet but they round trips things through clojure -> java -> ruby and back).

However, attempting those same optimizations in PE cause compilation failure. Not totally sure, but looks like maybe directly linking doesn't work with the jruby service (I know file-sync uses it to lock the pool that isn't used currently in FOSS which might be a cause).

So, for now we should probably wait on merging the direct linking optimization until we can figure out whether or not code that is incompatible with it is used anywhere in FOSS.

However I think the aot commit above is correct (if folks are happy with this I can update the commit message).

PR #1885, opened 2018-11-16
(maint) Fixups for version determination in CI

underscorgan:
We should use `git fetch --unshallow` instead of `git pull --unshallow`
so we aren't merging. We should also use recursively expanded variables
for variables relating to versions so that they are set following the
execution of the 'prep' command.


Iristyle:
This was accidentally targeted at master, so #1886 was opened up against the 5.3.x branch as well.

PR #1889, opened 2018-11-20
(maint) Fix PUP issue link in release notes

alexjfisher:


clairecadman:
@justinstoller I think it happens automatically, but doesn't it need to be cherry-picked up to 6.0 and master?

PR #1900, opened 2018-11-26
Add nightly alpine builds

pcarlisle:
This uses the alpine Dockerfile from puppet-agent. For now it's only
nightlies as this isn't present in older puppet-agent versions.

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -4,10 +4,16 @@ build_date := $(shell date -u +%FT%T)
 hadolint_available := $(shell hadolint --help > /dev/null 2>&1; echo $$?)
 hadolint_command := hadolint --ignore DL3008 --ignore DL3018 --ignore DL4000 --ignore DL4001
 hadolint_container := hadolint/hadolint:latest
+makefile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
+makefile_dir := $(dir $(makefile_path))
+agent_ref = $(shell grep puppet_build_version $(makefile_dir)/../acceptance/config/beaker/options.rb | cut -d\" -f2)
+
+BUILD_TARGETS = build-ubuntu
 
 ifeq ($(IS_NIGHTLY),true)
 	dockerfile := Dockerfile.nightly
 	version := puppet6-nightly
+	BUILD_TARGETS += build-alpine
 else
 	version = $(shell echo $(git_describe) | sed 's/-.*//')
 	dockerfile := Dockerfile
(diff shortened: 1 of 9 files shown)


docker/conf.d/auth.conf
```diff
@@ -0,0 +1,193 @@
+authorization: {
```
Magisus: I notice that this file doesn't have the `certificate_status` endpoints exposed. Is that intentional?

pcarlisle: no, i just copied these from wherever i thought the packages got them

docker/Makefile
```diff
@@ -21,22 +27,39 @@ endif
...
-	@docker run --rm -v $(PWD)/puppetserver/$(dockerfile):/Dockerfile -i $(hadolint_container) $(hadolint_command) Dockerfile
+	docker pull $(hadolint_container)
+	docker run --rm -v $(PWD)/puppetserver-standalone/$(dockerfile):/Dockerfile -i $(hadolint_container) $(hadolint_command) Dockerfile
+	docker run --rm -v $(PWD)/puppetserver/$(dockerfile):/Dockerfile -i $(hadolint_container) $(hadolint_command) Dockerfile
+	docker run --rm -v $(PWD)/puppetserver-standalone-alpine/Dockerfile:/Dockerfile -i $(hadolint_container) $(hadolint_command) Dockerfile
 endif
 
-build: prep
+puppet-agent/.git/config:
+	@git clone https://github.com/puppetlabs/puppet-agent.git puppet-agent
```
tvpartytonight: What is the reason why these are prefixed with the `@` while the above docker runs/hadolints are not?

pcarlisle: I removed the hadolint ones specifically beacuse without echoing the command you actually can't tell which file failed. These are here I guess because they're everywhere else. I'm inclined to remove all of them throughout the file though.

docker/puppetserver-standalone-alpine/Dockerfile
```diff
@@ -0,0 +1,87 @@
...
+COPY --from=agent /etc/puppetlabs /etc/puppetlabs
+COPY --from=agent /usr/local/bin /usr/local/bin
+COPY --from=agent /usr/local/lib/site_ruby /usr/local/lib/site_ruby
+COPY --from=agent /usr/local/lib/libfacter.so.* /usr/local/lib/
+COPY --from=agent /usr/local/share /usr/local/share
+RUN ln -s /usr/local/lib/libfacter.so.* /usr/local/lib/libfacter.so
+
+EXPOSE 8140
+
+CMD ["java", "-cp", "/puppet-server-release.jar", "clojure.main", "-m", "puppetlabs.trapperkeeper.main", "-b", "/etc/puppetlabs/puppetserver/bootstrap.cfg,/etc/puppetlabs/puppetserver/services.d", "-c", "/etc/puppetlabs/puppetserver/conf.d"]
```
nwolfe: If you wanted an init system the current tool of choice is `tini` for that, check out: https://github.com/krallin/tini#alpine-linux-package

Probably just  a 2-line addition to the file, but then the container should properly handle signals like stopping it via Ctrl-C.

pcarlisle:
this isn't 100% at the level of the others yet, i still need to do some things

docker/Makefile
pcarlisle: I think it's just the container ones, since all the files get copied to the same name it's hard to tell. I could also make them have different names, but generally I'm in favor of logging all the commands so it's easier to see what's happening and debug later.

pcarlisle:
We don't pass the tests because we don't install the puppet agent. Is there any reason to want that?

docker/puppetserver-standalone-alpine/Dockerfile
nwolfe: @pcarlisle Heads up...just learned that you probably always want to specify the `tini -g` flag so it properly handles killing process groups (see https://github.com/krallin/tini#process-group-killing)

For example `sh -c 'sleep 10'` wouldn't be killable via Ctrl-C with just `tini`.

So you'd want `ENTRYPOINT ["tini", "-g", "--"]`

underscorgan:
@pcarlisle it doesn't look like the tests require the puppet agent to be installed, they just pull a puppet-agent-alpine image to run something against the server

docker/Makefile
```diff
@@ -21,22 +27,39 @@ endif
...
 	@docker build --pull --build-arg vcs_ref=$(vcs_ref) --build-arg build_date=$(build_date) --build-arg version=$(version) --file puppetserver-standalone/$(dockerfile) --tag puppet/puppetserver-standalone:$(version) puppetserver-standalone
 	@docker build --build-arg vcs_ref=$(vcs_ref) --build-arg build_date=$(build_date) --build-arg version=$(version) --file puppetserver/$(dockerfile) --tag puppet/puppetserver:$(version) puppetserver
 ifeq ($(IS_LATEST),true)
 	@docker tag puppet/puppetserver-standalone:$(version) puppet/puppetserver-standalone:latest
 	@docker tag puppet/puppetserver:$(version) puppet/puppetserver:latest
 endif
 
+build: $(BUILD_TARGETS)
+
 test: prep
```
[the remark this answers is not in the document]
nwolfe: I think we need to add another line like this:

	@PUPPET_TEST_DOCKER_IMAGE=puppet/puppetserver-standalone-alpine:$(version) bundle exec rspec puppetserver-standalone-alpine/spec

But it gets even better! Morgan has a PR #1956 that consolidates the spec/ directories so they don't all need their own set of the same tests.

Iristyle:
I'm going to close this one for now, as we haven't had much traction on it in a while. We'd like to eventually get to it, but it's not on the short-term roadmap at least.

PR #1904, opened 2018-11-29
(maint) Change docker-entrypoint.sh to source from a directory

lutter:
The entrypoint script was getting big and hard to extend. This change
splits the entrypoint script into individual scripts in docker-entrypoint.d
and reduces docker-entrypoint.sh to running scripts from that directory.

### DIFF
--- a/docker/puppetserver-standalone/docker-entrypoint.sh
+++ b/docker/puppetserver-standalone/docker-entrypoint.sh
@@ -1,51 +1,11 @@
 #!/bin/bash
 
-chown -R puppet:puppet /etc/puppetlabs/puppet/
-chown -R puppet:puppet /opt/puppetlabs/server/data/puppetserver/
+set -e
 
-# During build, pristine config files get copied to this directory. If
-# they are not in the current container, use these templates as the
-# default
-TEMPLATES=/var/tmp/puppet
-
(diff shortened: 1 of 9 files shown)


docker/puppetserver-standalone/docker-entrypoint.sh
```diff
@@ -1,51 +1,9 @@
...
-cd /etc/puppetlabs/puppet
-for f in auth.conf hiera.yaml puppet.conf puppetdb.conf
-do
-    if ! test -f $f ; then
-        cp -p $TEMPLATES/$f .
-    fi
+for f in /docker-entrypoint.d/*.sh; do
+    echo "Running $f"
+    chmod +x "$f"
+    "$f"
```
Iristyle: Should we have a `set -e` at the top of the script to fail fast? We probably want any erroring script to kill the whole entrypoint and never allow healthy, right?

lutter: Good idea ! I updated the PR accordingly

Iristyle:
Per the extra vetting of `set -e` in #1909 I think this is good to go, aside from the base branch.

lutter:
@Iristyle I've retargeted the PR and all checks have now passed. Ok to merge ?

PR #1906, opened 2018-12-03
DOC-3845

kayayarai:
added advice about using max-queued-requests with older agents


Magisus:
Travis isn't going to tell us anything useful here, merging.

PR #1915, opened 2018-12-11
(CPR-650) Make puppetserver.conf more configurable

underscorgan:
This adds env variables `PUPPETSERVER_MAX_ACTIVE_INSTANCES` and
`PUPPETSERVER_MAX_REQUESTS_PER_INSTANCE` to allow `max-active-instances`
and `max-requests-per-instance` to be configured in
/etc/puppetlabs/puppetserver/conf.d/puppetserver.conf.

### DIFF
--- a/docker/puppetserver-standalone/Dockerfile
+++ b/docker/puppetserver-standalone/Dockerfile
@@ -9,7 +9,7 @@ ENV UBUNTU_CODENAME="xenial"
 ENV PUPPETSERVER_JAVA_ARGS="-Xms256m -Xmx256m"
 ENV PATH=/opt/puppetlabs/server/bin:/opt/puppetlabs/puppet/bin:/opt/puppetlabs/bin:$PATH
 ENV PUPPET_HEALTHCHECK_ENVIRONMENT="production"
-ENV PUPPETSERVER_MAX_ACTIVE_INSTANCES=''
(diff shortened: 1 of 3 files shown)


justinstoller:
Is there a default for max-active-instances that we want to be a default for docker images (iirc, it will choose between 1 ... 4 depending on how many CPUs the underlying OS has). It looks like we're setting `-Xms256m -Xmx256m`, for that we might want to limit ourselves to 1 JRuby instance?

underscorgan:
@justinstoller I'll trust you and @Magisus for the defaults here. We do go up to 512m in the 6.0.x branch since it was OOMing with 256. But can confirm it uses 1-4 based on CPUs by default right now.

Magisus:
I was just thinking about this being in line with our current defaults for regular installs. If we're restricting the memory, I'm with Justin, we should probably default lower.

underscorgan:
Does defaulting to 1 seem reasonable?

justinstoller:
Yes. You might be able to get away with 2 (say a jruby to server file metadata while one does a catalog compile). But I don't think you're going to get a good experience with more than 1 catalog compile at a time and 512m or less memory.

justinstoller:
I think we can then get rid of docker-entrypoint.d/35-update-puppetserver-conf.sh too.

underscorgan:
@justinstoller @Magisus updated!

PR #1932, opened 2019-01-03
(SERVER-2418) Update clj-parent to 1.7.19

Magisus:
This commit updates clj-parent to 1.7.19, which pulls in
jackson-databind 2.9.8, fixing several security issues.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -161,7 +161,7 @@ namespace :spec do
       gem_install_bundler = <<-CMD
       GEM_HOME='#{TEST_GEMS_DIR}' GEM_PATH='#{TEST_GEMS_DIR}' \
       lein #{profile} run -m org.jruby.Main \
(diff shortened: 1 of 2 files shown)


Magisus:
For now I'm just putting a version constraint on bundler, because I was having trouble getting the Rubygems update to work properly under JRuby. We should investigate this more, but since today is stop-ship for the 5.3.x release, we should prioritize getting this in.

Magisus:
Based on https://github.com/puppetlabs/ci-job-configs/blob/master/resources/scripts/unit-jruby.sh#L49, I think this version pin here might be sufficient to fix at least the unit test job in Jenkins too.

Side note, given that we already run these tests (both the Clojure and the Ruby) in Travis, is there benefit to also running them in Jenkins? Seems excessive, given that we still only run on one OS in Jenkins.

PR #1933, opened 2019-01-04
(SERVER-2418) Update clj-parent to 2.5.0

Magisus:
This commit updates clj-parent to 2.5.0, which brings in
jackson-databind 2.9.8, resolving some security issues.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -36,7 +36,7 @@
 
   :min-lein-version "2.7.1"
 
-  :parent-project {:coords [puppetlabs/clj-parent "2.4.1"]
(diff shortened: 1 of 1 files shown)


justinstoller:
I think travis will be resolved when we merge up https://github.com/puppetlabs/puppetserver/pull/1932. Either we should merge that up and then rebase this onto that, or do this bump as part of the merge up. Either way 👍

Magisus:
Superseded by https://github.com/puppetlabs/puppetserver/pull/1935

PR #1934, opened 2019-01-04
(MAINT) Update Azure CI to use docker CLI directly

nwolfe:
Update the Azure CI docker builds to match puppetdb/puppet-agent and
prefer the docker CLI over our puppet-docker-tools.

Also push the docker/.rspec file down into both the puppetserver and
puppetserver-standalone subdirectories so they can each be configured to
write output to their own results file and not overwrite one another.


docker/Makefile
```diff
@@ -31,16 +31,16 @@ endif
 
 build: prep
 	@docker build --pull --build-arg vcs_ref=$(vcs_ref) --build-arg build_date=$(build_date) --build-arg version=$(version) --file puppetserver-standalone/$(dockerfile) --tag puppet/puppetserver-standalone:$(version) puppetserver-standalone
-	@docker build --build-arg vcs_ref=$(vcs_ref) --build-arg build_date=$(build_date) --build-arg version=$(version) --file puppetserver/$(dockerfile) --tag puppet/puppetserver:$(version) puppetserver
+	@docker build --pull --build-arg vcs_ref=$(vcs_ref) --build-arg build_date=$(build_date) --build-arg version=$(version) --file puppetserver/$(dockerfile) --tag puppet/puppetserver:$(version) puppetserver
```
nwolfe: Seems like we always want `--pull` so I added it to this other build call.

docker/ci/build.ps1
```diff
@@ -19,27 +19,45 @@ function Get-ContainerVersion
...
-  bundle exec puppet-docker build $Name --no-cache --repository $Repository --version $Version --no-latest --build-arg namespace=$Repository
+
+  $build_date = (Get-Date).ToUniversalTime().ToString('o')
+  $docker_args = @(
+    '--pull',
+    '--build-arg', "vcs_ref=$Vcs_ref",
+    '--build-arg', "build_date=$build_date",
+    '--build-arg', "version=$Version",
+    '--file', "$Name/Dockerfile",
+    '--tag', "$Namespace/$Name:$Version",
```
nwolfe: I noticed we're not tagging as latest here like we are in puppetdb...

Turns out the Makefile actually does the tagging as latest via `docker tag $name:$version $name:latest` (in both PuppetDB and PuppetServer) so I don't think we need to worry about "latest" at this level.

docker/Makefile
underscorgan: nope, this one was intentional, we explicitly want to build from the puppetserver-standalone container we just built and pull will instead build from the latest published

docker/ci/build.ps1
```diff
@@ -19,27 +19,46 @@ function Get-ContainerVersion
...
+  $Version = (Get-ContainerVersion),
+  $Vcs_ref = $(git rev-parse HEAD))
 {
   Push-Location (Join-Path (Get-CurrentDirectory) '..')
-  bundle exec puppet-docker local-lint $Name
-  bundle exec puppet-docker build $Name --no-cache --repository $Repository --version $Version --no-latest --build-arg namespace=$Repository
+
+  $build_date = (Get-Date).ToUniversalTime().ToString('o')
+  $docker_args = @(
+    '--pull',
```
underscorgan: we don't always want `--pull`, we need to build the puppetserver container based on the puppetserver-standalone container we just built.

nwolfe: Did you mean to comment on the Makefile change where I added the second `--pull`?  Or do we really not want to do any pulling here in the powershell?

underscorgan: Well, we want pulling in the powershell, but it should be conditional so we aren't trying to pull when building the puppetserver container but are for the puppetserver-standalone.

nwolfe: How was that handled before this PR?  I don't remember seeing anything around conditional building.

underscorgan: We weren't using `--pull` prior to this, we were explicitly pulling the base image for the puppetserver-standalone container.

nwolfe: I think I fixed this in: https://github.com/puppetlabs/puppetserver/pull/1934/commits/f7e0911306bc447aa18be537c51ea788aca8fc01

nwolfe:
@underscorgan ok, got the conditional --pull'ing in here: https://github.com/puppetlabs/puppetserver/pull/1934/commits/50580c1d4c10380c51027f647008de93ff110569

Once this PR is 👍 I'll "Squash and Merge".

azure-pipelines.yml
```diff
@@ -44,47 +44,52 @@ steps:
...
-  name: build_prepare
+    Lint-Dockerfile -Path ./docker/puppetserver/Dockerfile
+  displayName: Lint PuppetServer Dockerfile
+  name: lint_puppetserver_dockerfile
 - powershell: |
     . ./docker/ci/build.ps1
-    Build-Container -Name puppetserver-standalone -Repository $ENV:BUILD_REPOSITORY
-  displayName: Build Puppetserver Standalone
-  name: build_puppetserver_standalone
+    Build-Container -Name puppetserver-standalone -Namespace $ENV:NAMESPACE
```
underscorgan: [nitpick] you may be able to get rid of all the `-Namespace $ENV:NAMESPACE` since it looks like that's the default for the methods in the ps1.

nwolfe: Yeah we're specifying namespace twice between build.ps1 and azure.yml in a few projects.

Iristyle: Yeah, being explicit isn't a bad idea either FWIW

PR #1945, opened 2019-01-16
(DOCS) Add Puppet Server 6.2.0 release notes

clairecadman:
This commit adds the release notes for Puppet Server 6.2.0.


justinstoller:
I think we need to update some tickets with the 6.2 label (where we weren't sure if we were doing a 6.1.1 or 6.2). This should also be targeted at master which is where we will release new "y" versions from for the time being.

PR #1947, opened 2019-01-17
(maint) String resources should not end with ","

ehom:
Also, remove extra comma in the sentence.


Magisus:
Looks like there are also at least two other tests in that file with a similar issue.

Magisus:
I'm not really sure why that test is still failing. I can help poke at it in a bit.

PR #1949, opened 2019-01-24
(Docs) profiler-output-file setting incorrectly documented

clairecadman:
The documentation for the jruby profiling output file setting is incorrect. The current documentation shows profiling-output-file while the actual setting name is profiler-output-file. (https://github.com/puppetlabs/jruby-utils/blob/d8a8f9be9499f78356f5874d41b16cbeb65cc758/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj#L167)

This is also true in previous versions of this doc.


Magisus:
If this is true in older versions of the doc, can we target this at 5.3.x instead?

justinstoller:
You need to rebase (or cherry-pick your commit into a new branch based off of 5.3.x).

PR #1952, opened 2019-01-24
(PDB-4232) Pin apt module to 6.2.1 when installing puppetDB

Magisus:
The 6.3.0 release of the puppetlabs-apt module broke the PDB module's
ability to install PDB on Debian. Until a solution can be found to
resolve this, this commit pins the version of the apt module in our
tests to allow PDB to install on Debian.


Magisus:
The PuppetDB team is planning to work with them modules team to find a real solution to the dependency issues they're having, as well as discuss future maintenance and ownership of the PDB module. But for now, this will get our tests passing again.

PR #1956, opened 2019-01-25
(maint) Add the ability to disable the CA in the puppetserver

underscorgan:
This will allow you to have compile masters or make use of external CAs.
Default behavior is not changed. To set up a compile master, start up
the puppetserver container with ENABLE_CA=false and
CA_HOSTNAME=\<hostname\>


underscorgan:
Still To do -- update testing

docker/puppetserver-standalone/Dockerfile
```diff
@@ -11,6 +11,7 @@ ENV PATH=/opt/puppetlabs/server/bin:/opt/puppetlabs/puppet/bin:/opt/puppetlabs/b
 ENV PUPPET_HEALTHCHECK_ENVIRONMENT="production"
 ENV PUPPETSERVER_MAX_ACTIVE_INSTANCES=1
 ENV PUPPETSERVER_MAX_REQUESTS_PER_INSTANCE=0
+ENV ENABLE_CA=true
```
nwolfe: Is the point of this PR to enable compile masters, or to simply add a flag for disabling the CA (for whatever purposes)?

I get a sense of both in different parts of the PR, I think.

If the point is to get additional compile masters, then maybe we should name the variables based on that, so we don't need to add additional ENV vars for other things that we need to toggle for compile masters.

For example, maybe this becomes `COMPILE_MASTER=false`, and then in the new ca.sh script we'd check `if [ "$COMPILE_MASTER" = "true" ]; then disable CA`.

This would make it clear in the pupperware compose file too what's going on, and we wouldn't have to add additional ENV vars for other things necessary to get a compile master.

Also thinking about how this looks in the compose file:
```yaml
services:
  puppet:
    image: puppet/puppetserver
  compiler1:
    image: puppet/puppetserver
    environment:
      - COMPILE_MASTER=true
  compiler2:
    image: puppet/puppetserver
    environment:
      - ENABLE_CA=false
      # - SOME_OTHER_ENV_WE_NEED_LATER_FOR_COMPILE_MASTER=true
```

And either way it's easy enough to change later, just wanted to think about the design/naming a bit.

nwolfe: Dangit, I missed your commit message 😄 

I see that this PR is not specifically for compile masters, so largely disregard my comment above.

In that case, since we're introducing 2 new variables that go together, I'd suggest naming it `CA_ENABLED` so it matches the related variable `CA_HOSTNAME`, and then any other CA-related settings can follow suit.

docker/puppetserver-standalone/docker-entrypoint.d/80-ca.sh
```diff
@@ -0,0 +1,44 @@
...
+  echo "turning off CA"
+  cat > /etc/puppetlabs/puppetserver/services.d/ca.cfg <<EOF
+puppetlabs.services.ca.certificate-authority-disabled-service/certificate-authority-disabled-service
+puppetlabs.trapperkeeper.services.watcher.filesystem-watch-service/filesystem-watch-service
+EOF
+
+  ssl_cert=$(puppet master --configprint hostcert)
+  ssl_key=$(puppet master --configprint hostprivkey)
+  ssl_ca_cert=$(puppet master --configprint localcacert)
+  certname=$(puppet config print certname)
```
nwolfe: I'm not sure we should even bother with calling into `puppet config` or `puppet master --configprint` since we're assuming the filepaths anyway:
- https://github.com/puppetlabs/puppetserver/pull/1956/files#diff-f9e912c5472d58a4ee06a97dc9592516R33
- https://github.com/puppetlabs/puppetserver/pull/1956/files#diff-f9e912c5472d58a4ee06a97dc9592516R37

And we're using `$HOSTNAME` in place of `puppet config print certname` in other container-only places too.

docker/puppetserver-standalone/Dockerfile
Iristyle: Yeah, +1 to syncing on `CA_` naming convention.

docker/puppetserver-standalone/docker-entrypoint.d/80-ca.sh
underscorgan: I think using `puppet config print certname` might be leftover from David's work where the compile masters and CA shared certs and certname != $HOSTNAME in that case, but I can probably get rid of that.

I think for the others I'd rather fix the places where the cert paths were assumed just to be more forward-proof / less fragile.

underscorgan: mm, so it looks like if the `$PUPPETSERVER_HOSTNAME` env variable is set certname may not be the same as `$HOSTNAME`, so leaving that check in would probably be better

docker/puppetserver-standalone/docker-entrypoint.d/80-ca.sh
```diff
@@ -0,0 +1,44 @@
...
+  hocon -f webserver.conf set webserver.ssl-crl-path /etc/puppetlabs/puppet/ssl/crl.pem
+  cd /
+
+  # bootstrap certs for the puppetserver
+  if [[ ! -f "/etc/puppetlabs/puppet/ssl/certs/${certname}.pem" ]]; then
+    while ! master_running; do
+      sleep 1
+    done
+
+    puppet agent --noop --server="${CA_HOSTNAME}"
```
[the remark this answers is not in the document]
underscorgan: yeah, this is just for getting ssl certs.

I took a look at the puppet ssl subcommand but it looks like that doesn't exist yet in 5.3.x

docker/puppetserver-standalone/docker-entrypoint.d/80-ca.sh
nwolfe: > I think for the others I'd rather fix the places where the cert paths were assumed just to be more forward-proof / less fragile.

We can't just do this across the board because many containers won't have a puppet CLI to call into; puppetserver is kinda special in this case.

> mm, so it looks like if the $PUPPETSERVER_HOSTNAME env variable is set certname may not be the same as $HOSTNAME, so leaving that check in would probably be better

I wouldn't expect those two ENV variables to be the same, but that's not what I was talking about.

I was just saying that we've simplified calls like `puppet config print certname` into just `$HOSTNAME` elsewhere, and was thinking we could do that here to.

Would that not work here?

Not a problem if we leave it like this, since this is the one container that actually does have the `puppet config` CLI available.

nwolfe: Oh I see how you updated this to not need the certname at all.

Nice!

Iristyle:
Oops, looks like we need a rebase now @underscorgan! Sorry 'bout that.

underscorgan:
@Iristyle rebased!

docker/puppetserver-standalone/docker-entrypoint.d/80-ca.sh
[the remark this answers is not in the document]
nwolfe: @underscorgan Do you have any sense if we should use the new `ssl.sh` script instead of the agent run here?

Puppetserver is the one of the containers that actually *will* have a puppet-agent available on it, but wondering if there's value in consistency across our containers and to prevent this agent run from organically growing other responsibilities in the future.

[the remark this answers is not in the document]
underscorgan: I don't have super strong feelings, though leaning slightly towards using ssl.sh for consistency.

underscorgan:
@Iristyle rebased! I think somewhere in the force pushing azure got confused, but the build seems to be running at https://dev.azure.com/PuppetInc/puppetserver/_build/results?buildId=1259

@nwolfe I looked a little at moving the agent run in 80-ca.sh to use ssl.sh, but `type` on the ubuntu images doesn't support `-p` (it works fine on alpine), but seems like maybe we should just hold off on that until we move to alpine rather than updating the script to support ubuntu.

Iristyle:
As soon as #2819 goes in, we should be able to merge this one... 🚋

PR #1962, opened 2019-02-01
Add new catalog endpoint

pcarlisle:
very wip do not merge

### DIFF
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -725,7 +725,7 @@
       :persistence {:facts schema/Str, :catalogs schema/Str, :reports schema/Str}
       (schema/optional-key :trusted_facts) {:values {schema/Any schema/Any}}
       (schema/optional-key :facts) {:values {schema/Any schema/Any}}
-      (schema/optional-key :job-id) schema/Int
+      (schema/optional-key :job_id) schema/Int
       (schema/optional-key :environment) schema/Str
       (schema/optional-key :classes) [schema/Any]
       (schema/optional-key :parameters) {schema/Any schema/Any}}
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -5,6 +5,7 @@
            (org.jruby RubySymbol)
            (org.jruby.exceptions RaiseException)
            (com.codahale.metrics MetricRegistry Gauge)
+           (com.fasterxml.jackson.core JsonParseException)
            (java.lang.management ManagementFactory))
   (:require [me.raynes.fs :as fs]
             [puppetlabs.puppetserver.ringutils :as ringutils]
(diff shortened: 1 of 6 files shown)


src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -710,6 +712,48 @@
...
+  [jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)]
+  (fn [request]
+    (let [{:keys [certname persistence trusted_facts facts environment classes parameters]}
+          (decode-catalog-post-body (slurp (:body request)))]
+      {:status 200
+       :headers {"Content-Type" "application/json"}
+       :body (json/encode
+              (jruby-protocol/compile-catalog jruby-service
+                                              (:jruby-instance request)
+                                              {:certname certname
```
tvpartytonight: This map should probably contain all the keys listed in the schema in the `decode-catalog-post-body` method?

Magisus: Agreed, I think they all end up being needed in the Ruby code.

src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -60,6 +60,10 @@ def handleRequest(request)
         response["X-Puppet-Version"])
   end
 
+  def compileCatalog(_request_data)
```
tvpartytonight: What is the reason for the `_` for `request_data`?

src/clj/puppetlabs/services/master/master_core.clj
pcarlisle: ok tony wrote that part

src/ruby/puppetserver-lib/puppet/server/master.rb
pcarlisle: signifies that it's unused (for now)

tvpartytonight:
I'll defer to @Magisus to actually merge this, since I wrote some of this too.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -710,6 +712,45 @@
...
+          (json/decode body true)
+          (catch JsonParseException e
+            (throw+ {:kind :bad-request
+                     :msg (format "Error parsing JSON: %s" e)})))]
+    (schema/validate
+     {:certname schema/Str
+      :persistence {:facts schema/Str, :catalogs schema/Str, :reports schema/Str}
+      (schema/optional-key :trusted_facts) {:values {schema/Any schema/Any}}
+      (schema/optional-key :facts) {:values {schema/Any schema/Any}}
+      (schema/optional-key :job-id) schema/Int
```
Magisus: Can we choose either hyphen or underscores for these keys (this vs. `trusted_facts` above)? I don't have a strong preference which we choose.

pcarlisle: @tvpartytonight where does this job-id come from?

pcarlisle: looks like just an error, i've fixed it to _

pcarlisle:
everything in here can probably be squashed

Magisus:
Can maybe do that as part of the merge with the fancy button... Or you can squash things and then we merge.

pcarlisle:
Feel free to use the button, either way is fine but that'll be faster if you're good with the changes. I just kept them separate for review purposes.

PR #1967, opened 2019-02-06
Add details to certificate status endpoint

runejuhl:
Currently fails a few tests.

See https://tickets.puppetlabs.com/browse/SERVER-2370

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -535,7 +535,7 @@
     (line-seq (io/reader infra-file)))
 
 (schema/defn generate-infra-serials
-  "Given a list of infra nodes it will create a file containing 
+  "Given a list of infra nodes it will create a file containing
    serial numbers of their certificates (listed on separate lines).
    It is expected have at least one entry (MoM)"
    [{:keys [infra-nodes-path infra-node-serials-path signeddir]} :- CaSettings]
@@ -549,13 +549,13 @@
                                  (utils/get-serial))]
                (.write wtr (str infra-serial))
                (.newLine wtr))
-            (catch java.io.FileNotFoundException ex 
+            (catch java.io.FileNotFoundException ex
                (log/warn
                  (i18n/trs
                    (str
                      "Failed to find/load certificate for Puppet Infrastructure Node:"
                      infra-node)))))))
-       (catch java.io.FileNotFoundException ex 
+       (catch java.io.FileNotFoundException ex
          (log/warn (i18n/trs (str infra-nodes-path " does not exist")))))))
 
 (schema/defn generate-ssl-files!
@@ -618,7 +618,7 @@
    provided then defaults will be used."
   [host-name :- schema/Str
    alt-names :- schema/Str]
-  (let [split-alt-names (split-hostnames alt-names)
+  (let [split-alt-names   (split-hostnames alt-names)
         default-alt-names ["puppet"]
         alt-names-list (reduce (fn [acc alt-name]
                                  (if (str/starts-with? alt-name "IP:")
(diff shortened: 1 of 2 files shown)


bastelfreak:
Ups, I mixed my github tabs and approved the wrong PR. Sorry!

runejuhl:
Hah, did that myself the other day. I'll see if I can't get the tests fixed tonight.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1317,24 +1320,41 @@
        (str/join ":")
        (str/upper-case)))
 
+(defn get-certificate-details
+  "Return details from a X509 certificate."
+  [^sun.security.x509.X509CertImpl cert]
```
justinstoller: Elsewhere in this file, instead of doing type hints we would do something like:
```clj
(schema/defn get-certificate-details
  "..."
  [cert :- X509CertImpl]
  (...))
```
Where `sun.security.x509.X509CertImpl` had already been imported in the `ns` and could we referenced just by the class name here.

justinstoller:
This looks great! One thing for later auditing: can you rebase and prepend `(SERVER-2370)` or `(maint)` to your commit messages?

runejuhl:
Sure @justinstoller, I'll try to get it out the door today.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
runejuhl: Fixed, thanks!

runejuhl:
There's a single failing test:

~~~
FAIL in (handle-put-certificate-request!-test) (certificate_authority_core_test.clj:242)
when autosign results in true it signs the CSR, writes the certificate to disk, and
                      returns a 200 response with empty plaintext body
expected: (true? (fs/exists? expected-path))
  actual: (not (true? false))
2019-02-12 21:27:12,912 WARN  [main] [p.p.certificate-authority] No certificate request for revoked-agent at expected path /tmp/null1550003232737-2157122427/ca/requests/revoked-agent.pem
~~~

I've tested against master and the tests fails there too, so I'm assuming I can safely ignore it.

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
```diff
@@ -99,59 +102,63 @@
...
+    (let [request  (mock/request :get
+                                 "/v1/certificate_revocation_list/mynode")
           response (handle-get-certificate-revocation-list
-                     request {:cacrl (test-pem-file "crl.pem") :infra-crl-path (test-pem-file "crl.pem") :enable-infra-crl false})]
+                    request {:cacrl (test-pem-file "crl.pem") :infra-crl-path (test-pem-file "crl.pem") :enable-infra-crl false})]
       (is (map? response))
       (is (= 200 (:status response)))
       (is (= "text/plain" (get-in response [:headers "Content-Type"])))
       (is (string? (:body response))))))
-  (testing "with a malformed http-date 'If-Modified-Since' header"
```
justinstoller: This alignment shouldn't be like this. It looks like your editor is doing it this way because the parens are off. I'm certain these `testing` forms should be within `deftest crl-endpoint-test`.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -94,16 +95,29 @@
...
-   :fingerprints      {schema/Keyword schema/Str}})
+  (let [base {:name              schema/Str
+              :state             CertificateState
+              :dns_alt_names     [schema/Str]
+              :subject_alt_names [schema/Str]
+              :fingerprint       schema/Str
+              :fingerprints      {schema/Keyword schema/Str}}]
+    ;; The map should either have all of the CertificateDetails keys or none of
+    ;; them
+    (schema/either
```
justinstoller: iirc, `schema/either` is deprecated in favor of `schema/conditional` or `schema/if` which would require a predicate to be written to test if a CertDetail key existed and if so use the merge keys, otherwise use the "base" keys. I'll double check that.

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
runejuhl: You're right. I'll push a rebased commit in a sec.

justinstoller:
> There's a single failing test

Yeah, that's an unfortunately flaky test for us.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
runejuhl: Yes, I think that's correct. I'll change it.

justinstoller:
Just a heads up, assuming there's no regressions in our internal acceptance testing, this will go out in a release in about a month (There will be a release next week, but it won't be there as we cut the build for final validation earlier).

ekohl:
Thanks a lot for fixing this @runejuhl. Also thanks to all others involved.

PR #1978, opened 2019-02-14
(maint) If consul is enabled, register the puppet service on startup

underscorgan:


docker/puppetserver-standalone/docker-entrypoint.d/90-consul.sh
```diff
@@ -0,0 +1,30 @@
...
+{
+  "name": "puppet",
+  "id": "$HOSTNAME",
+  "port": 8140,
+  "address": "$ipaddress",
+  "checks": [
+    {
+      "http": "https://$HOSTNAME:8140/production/status/test",
+      "tls_skip_verify": true,
+      "header": {"Accept": ["pson"]},
```
nwolfe: Ha oh wow, forgot about PSON!

Are we able to specify JSON here instead?

Or is this actually PSON?

docker/puppetserver-standalone/docker-entrypoint.d/90-consul.sh
```diff
@@ -0,0 +1,30 @@
...
+
+  cat <<SERVICEDEF > /puppet-service.json
+{
+  "name": "puppet",
+  "id": "$HOSTNAME",
+  "port": 8140,
+  "address": "$ipaddress",
+  "checks": [
+    {
+      "http": "https://$HOSTNAME:8140/production/status/test",
```
nwolfe: Should we use the ENV variable for the environment here?

This one: https://github.com/puppetlabs/puppetserver/blob/master/docker/puppetserver-standalone/Dockerfile#L18

underscorgan: woops I made that change locally and forgot to push it :P

docker/puppetserver-standalone/docker-entrypoint.d/90-consul.sh
underscorgan: I copied that from the healthcheck but it looks like the calls pass without specifying any extra headers, so will probably just get rid of that in both places

docker/puppetserver-standalone/docker-entrypoint.d/90-consul.sh
```diff
@@ -0,0 +1,29 @@
...
+
+  cat <<SERVICEDEF > /puppet-service.json
+{
+  "name": "puppet",
+  "id": "$HOSTNAME",
+  "port": 8140,
+  "address": "$ipaddress",
+  "checks": [
+    {
+      "http": "https://${HOSTNAME}:8140/${PUPPET_HEALTHCHECK_ENVIRONMENT}/status/test",
```
nwolfe: The healthcheck that hits this same endpoint includes the various SSL files to satisfy HTTPS.

Is that not necessary here with Consul? Is that what `tls_skip_verify` is for?

underscorgan: yeah, `tls_skip_verify=true` is basically like passing `--insecure` to curl like we do in the pdb entrypoint.

underscorgan:
This should be good to go!

docker/puppetserver-standalone/docker-entrypoint.d/90-consul.sh
```diff
@@ -0,0 +1,29 @@
...
+      "tls_skip_verify": true,
+      "interval": "30s",
+      "deregister_critical_service_after": "5m"
```
nwolfe: The corresponding puppetdb PR uses a value of `10m` for this.

The difference is fine, just wondering if that was intentional?

Was that to account for the longer startup time of puppetdb (since it may have to do all those database migrations)?

underscorgan: yeah, the difference was mostly since puppetdb was going to spin waiting for puppetserver to start before it started to start itself

PR #1979, opened 2019-02-14
(SERVER-2444) Process facts for flexible catalog request

Magisus:


Magisus:
I've tested this with regular and trusted facts passed to the request, and they are not correctly able to be referenced in manifests.

Next step, make sure fetching from PDB works properly.

Magisus:
I've validated that this can compile a manifest for the requested node using both regular and trusted facts, both with and without puppetdb. I'm now going to take some time to see if we can avoid the `Facts.indirection.find` call when querying PDB for facts, so that we are not relying at all on the global configuration of the indirector.

justinstoller:
This failed on the docker travis check. I don't think anything in here would change that, but I've rekicked the build. If we continue to see docker builds being flaky we should accost the docker team.

src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -60,8 +62,41 @@ def handleRequest(request)
...
   end
 
-  def compileCatalog(_request_data)
-    {}
+  def convert_java_args_to_ruby(hash)
+    Hash[hash.collect do |key, value|
+        # Stolen and modified from params_to_ruby in handler.rb
+        newkey = key.to_s
+        newkey.slice!(0)
+        if value.java_kind_of?(Java::ClojureLang::PersistentArrayMap)
```
tvpartytonight: Not sure if this`Java::ClojureLang::PersistentArrayMap` should be `Java::JavaUtil::List` like it is in the original method this is taken from.

src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -127,6 +162,73 @@ def getTaskData(environment_name, module_name, task_name)
...
+    else
+      # How should this be surfaced? Seems like we could maybe do better than a 500, unless
+      # that's accurate?
+      raise(Puppet::Error, "PuppetDB not configured, please provide facts with your catalog request.")
+    end
+  end
+
+  # Initialize our server fact hash; we add these to each client, and they
+  # won't change while we're running, so it's safe to cache the values.
+  def set_server_facts
```
tvpartytonight: Maybe put in a comment to reference where this method comes from in the [puppet compiler](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/indirector/catalog/compiler.rb#L390)?

src/ruby/puppetserver-lib/puppet/server/master.rb
pcarlisle: list is handled in the next branch

more importantly, this is not a reliable way to determine if something is a clojure map since the underlying class can change dynamically based on the size of the map

src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -127,6 +162,79 @@ def getTaskData(environment_name, module_name, task_name)
...
+                      fact_values['trusted'].nil? ? {} : fact_values['trusted']
+                    else
+                      request_data['trusted_facts']['values']
+                    end
+
+    return facts, trusted_facts
+  end
+
+  def get_facts_from_pdb(nodename, environment)
+    if Puppet[:storeconfigs_backend] == :puppetdb
```
justinstoller: I think this will always be `:puppetdb` regardless if PuppetDB is actually installed or configured - [storeconfig_backend](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/defaults.rb#L1891-L1898).

We might need to check both the backend and `Puppet[:storeconfigs] == true`? Or should we call `if Puppet::Node::Facts.indirection.class == :puppetdb`?

It looks like PuppetDB encourages consumers to configure the fact cache class to be `:yaml` and the generic fact indirection to be `:puppetdb` - [PuppetDB configuration](https://puppet.com/docs/puppetdb/6.2/connect_puppet_master.html#edit-routesyaml)??

I'm a little unclear how that corresponds to setting the facts cache class in the [`:storeconfigs` setting](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/defaults.rb#L1886)?

src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -127,6 +162,79 @@ def getTaskData(environment_name, module_name, task_name)
...
+    fact_values = if facts.nil?
+                    {}
+                  else
+                    facts.sanitize
+                    facts.to_data_hash['values']
+                  end
+
+    # Pull the trusted facts from the request, or attempt to extract them from
+    # the facts hash
+    trusted_facts = if request_data['trusted_facts'].nil?
```
justinstoller: This conditional feels backwards to me

src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -127,6 +162,79 @@ def getTaskData(environment_name, module_name, task_name)
 
   private
 
+  # @return Puppet::Node::Facts facts, Hash trusted_facts
+  def process_facts(request_data)
+    if request_data['facts'].nil?
+      facts = get_facts_from_pdb(request_data['certname'], request_data['environment'])
+    else
+      facts_from_request = request_data['facts']
```
justinstoller: If facts from request does not contain the key `values` (or that key is anything other than a Hash) then the Facts object will be invalid and blow up later.

src/ruby/puppetserver-lib/puppet/server/master.rb
Magisus: Oh good point about the default. I switched to using that after setting up PDB for myself (the instructions there say to set that setting to `puppetdb`), because the problem with using `Puppet::Node::Facts.indirection.class == :puppetdb` is that it still relies on the global configuration of the indirector, which I felt kind of defeated the point. I think maybe checking `Puppet[:storeconfigs] == true` makes more sense here.

The connection between the hook and the actual configured terminii seems to be https://github.com/puppetlabs/puppet/blob/master/lib/puppet/indirector/store_configs.rb#L6 for the terminus class and https://github.com/puppetlabs/puppet/blob/1a63db161de417c9dcbeffeb5b60790a6aee6503/lib/puppet/application_support.rb#L57 for the cache class. However, If we instantiate the terminus ourselves, we lose the automatic caching configured by `routes.yaml`.

src/ruby/puppetserver-lib/puppet/server/master.rb
Magisus: Do you know what would be more reliable here?

src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -127,6 +162,79 @@ def getTaskData(environment_name, module_name, task_name)
...
+      facts_from_request['name'] = request_data['certname']
+      facts = Puppet::Node::Facts.from_data_hash(facts_from_request)
+    end
+
+    fact_values = if facts.nil?
+                    {}
+                  else
+                    facts.sanitize
```
justinstoller: Feels weird to me to call sanitize here - like this conditional is turning whatever we have into a data structure, not doing additional, conditional processing.

justinstoller: I don't know, does putting the conditional sanitization up in the above conditional or something like this make sense?
```rb
facts.sanitize if facts
fact_values = facts ? facts.to_data_hash['values'] : {}
```

Magisus: Yeah I had just written something like the second one. I think it's clearer. Because the PDB terminus can explicitly return `nil` sometimes: https://github.com/puppetlabs/puppetdb/blob/master/puppet/lib/puppet/indirector/facts/puppetdb.rb#L84.

justinstoller: **Oh**! I see.... We only use this for finding `trusted_facts` _if_ `request_data['trusted_facts']` is `nil`.

src/ruby/puppetserver-lib/puppet/server/master.rb
Magisus: Is that something we'd want to validate here, or at the schema level in Clojure as we're parsing the body?

src/ruby/puppetserver-lib/puppet/server/master.rb
justinstoller: _And_ `Node#merge` calls `#sanitize` on the facts that are passed into it, which in our case is the `facts` variable.

Magisus: Yes. I still think splitting out sanitize makes sense though. The new thing is much shorter and clearer.

Magisus: What do you think about putting the creation of `fact_values` inside the `else` branch of the `trusted_facts` conditional? Does that make your above point more obvious?

src/ruby/puppetserver-lib/puppet/server/master.rb
justinstoller: It'd be nice to have any user input, if given, be syntactically correct. (though I kinda hate surfacing schema errors to users) I'm also not totally sure what other values from [`Node.from_data_hash`](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/node/facts.rb#L43-L59) are required, if any.

justinstoller: Non-answer of the day

src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -127,6 +164,79 @@ def getTaskData(environment_name, module_name, task_name)
...
+    end
+
+    facts.sanitize if facts
+
+    # Pull the trusted facts from the request, or attempt to extract them from
+    # the facts hash
+    trusted_facts = if request_data['trusted_facts']
+                      request_data['trusted_facts']['values']
+                    else
+                      fact_values = facts ? facts.to_data_hash : {}
```
justinstoller: previously, i think, this was `facts.to_data_hash['values']`

src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -127,6 +164,79 @@ def getTaskData(environment_name, module_name, task_name)
...
+      pdb_terminus = Puppet::Node::Facts::Puppetdb.new
+      request = Puppet::Indirector::Request.new(pdb_terminus.class.name,
+                                                :find,
+                                                nodename,
+                                                nil,
+                                                :environment => environment)
+      pdb_terminus.find(request)
+    else
+      # How should this be surfaced? Seems like we could maybe do better than a 500, unless
+      # that's accurate?
```
justinstoller: I think that depends on user intent, right?

If they sent a request expecting PDB to fill in facts but PDB isn't configured then its a server configuration error. If they wanted to specify their own facts (because they don't use PDB) but forgot, then it's a 400 - Bad Request.

400 feels... nicer, though from the way we've put default usage so far I think a 500 fits better (we just need to very explicitly document our assumptions when we release this).

src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -60,8 +62,45 @@ def handleRequest(request)
         response["X-Puppet-Version"])
   end
 
-  def compileCatalog(_request_data)
-    {}
+  def convert_java_args_to_ruby(hash)
+    Hash[hash.collect do |key, value|
+        # Stolen and modified from params_to_ruby in handler.rb
+        newkey = key.to_s
+        newkey.slice!(0)
```
justinstoller: wth

justinstoller: Is to remove a leading colon?

tvpartytonight: Yes. I wrote that method to unblock work for facts processing. It's not perfect!...or is it?

src/ruby/puppetserver-lib/puppet/server/master.rb
justinstoller: I would assume something like `Java::ClojureLang::IPersistentMap`? (See: [clojure's java interop docs](https://clojure.org/reference/data_structures#Maps) and [the java code](https://github.com/clojure/clojure/blob/4ef4b1ed7a2e8bb0aaaacfb0942729252c2c3091/src/jvm/clojure/lang/IPersistentMap.java), it looks like Java::Iterable would work, though I'd be surprised if that didn't also get Vectors as well).

tvpartytonight: It looks like [APersistentMap](https://github.com/clojure/clojure/blob/master/src/jvm/clojure/lang/APersistentMap.java) would also work?

src/ruby/puppetserver-lib/puppet/server/master.rb
justinstoller: Maybe it is...

src/ruby/puppetserver-lib/puppet/server/master.rb
Magisus: Yeah these were kind of my thoughts. I think given that we can't really know user intent, I think a 500 makes more general sense. I might make the error message more explicitly state what went wrong and the two options: configure PDB or supply your own facts.

src/ruby/puppetserver-lib/puppet/server/master.rb
pcarlisle: using the interface makes the most sense i think

justinstoller:
I think of `Puppet::Server::Master` as managing the bootstrapping of the Master's Ruby side code and providing an external interface, but I think'd be more maintainable to put any significant implementation details in other namespaces that the external interface delegates to.

Magisus:
Agreed! I look into doing this refactor today.

Magisus:
As part of this refactor I also attempted to make the complicated fact-processing code cleaner and more robust.

Magisus:
Please let me know if we don't like the "all class methods on a module" thing that I seem so prone to doing. I'm open to things more idiomatic to Ruby, it's just my go-to when there's no need to store any state.

src/ruby/puppetserver-lib/puppet/server/catalog.rb
```diff
@@ -0,0 +1,121 @@
...
+
+module Puppet::Server::Catalog
+
+  def self.convert_java_args_to_ruby(hash)
+    Hash[hash.collect do |key, value|
+        # Stolen and modified from params_to_ruby in handler.rb
+        newkey = key.to_s
+        # Java::ClojureLang::Keywords retain their leading colon when
+        # converted to a string
+        newkey.slice!(0)
```
justinstoller: I get what's happening here, though using Ruby's convenience methods might make it more readable:
`newkey.delete_prefix!(':')`

src/ruby/puppetserver-lib/puppet/server/catalog.rb
```diff
@@ -0,0 +1,121 @@
+require 'puppet/server'
+
+module Puppet::Server::Catalog
```
justinstoller: I usually do
```rb
module Puppet
  module Server
    module Catalog
    end
  end
end
```
Is that unnecessary here?

src/ruby/puppetserver-lib/puppet/server/catalog.rb
```diff
@@ -0,0 +1,121 @@
...
+        # Stolen and modified from params_to_ruby in handler.rb
+        newkey = key.to_s
+        # Java::ClojureLang::Keywords retain their leading colon when
+        # converted to a string
+        newkey.slice!(0)
+
+        if value.java_kind_of?(Java::ClojureLang::IPersistentMap)
+          [newkey, convert_java_args_to_ruby(value)]
+        else
+          [newkey, value.java_kind_of?(Java::JavaUtil::List) ? value.to_a : value]
```
justinstoller: This whole conditional feels to me more like a 3 branch switch on value's type. Would that be more readable or am totally bikeshedding? (I think Ruby's case wouldn't work well here either, so it'd end up being a 3 branch if/elsif/else...)

src/ruby/puppetserver-lib/puppet/server/catalog.rb
```diff
@@ -0,0 +1,121 @@
...
+    trusted_facts = extract_trusted_facts(request_data, facts)
+
+    return facts, trusted_facts
+  end
+
+  def self.extract_facts(request_data)
+    if request_data['facts'].nil?
+      facts = get_facts_from_pdb(request_data['certname'], request_data['environment'])
+
+      Puppet.warning "FACTS: #{facts}"
```
justinstoller: Debugging?

justinstoller:
I removed the "DO NOT MERGE" label, I've left some nit picks but I don't have any blocking issues that we couldn't address later. I would be happy if someone else approved this and decided to merge.

justinstoller:
I could go either way on the module vs class debate on this (either what you have here or turning this module into a class and instantiating it with the server facts).

Any preference @tvpartytonight 	and @pcarlisle 	?

src/ruby/puppetserver-lib/puppet/server/catalog.rb
Magisus: Oh.. yeah there are few weird edge cases, see https://techblog.thescore.com/2014/05/28/how-you-nest-modules-matters-in-ruby/. The biggest problem only arises if `Puppet::Server` were not defined in its own right (though both parts are in this case, so it works). This does work for me, but I'm happy to switch it.

src/ruby/puppetserver-lib/puppet/server/master.rb
Magisus: I am now checking more cases when processing facts, making sure that we won't blow up from things being nil at least.

src/ruby/puppetserver-lib/puppet/server/master.rb
Magisus: For now I have switched this to use `Puppet['storeconfigs'] == true`, later we can switch over to using Justin's cached indirector data if we like that idea, see https://gist.github.com/justinstoller/db1b9b761aa43e55615d787a49b7f9e7.

src/ruby/puppetserver-lib/puppet/server/catalog.rb
Magisus: Derp.

Magisus:
I thought about the second option too, storing `server_facts`. Happy to go that route if people like it more.

src/ruby/puppetserver-lib/puppet/server/catalog.rb
Magisus: Confirmed that both work in this case.

src/ruby/puppetserver-lib/puppet/server/catalog.rb
```diff
@@ -1,121 +1,124 @@
...
+          if facts.nil?
+            # Create an empty facts object
+            facts = Puppet::Node::Facts.new(request_data['certname'])
+          end
+
+          facts
+        else
+          # How should this be surfaced? Seems like we could maybe do better than a 500, unless
+          # that's accurate?
+          raise(Puppet::Error, "PuppetDB not configured, please provide facts with your catalog request.")
```
justinstoller: Playing with this more, I think this is part of determining whether to call PDB in the first place and might work better in the caller?

justinstoller:
Since I'm going with storing indirector info in my current work, I'm starting to lean more towards the class way, still love to hear opinions from others.

Magisus:
I think that might be more idiomatic anyway.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -0,0 +1,160 @@
...
+                                                  :find,
+                                                  nodename,
+                                                  nil,
+                                                  :environment => environment)
+        facts = pdb_terminus.find(request)
+
+        # If no facts have been stored for the node, PDB will return nil
+        if facts.nil?
+          # Create an empty facts object
+          facts = Puppet::Node::Facts.new(request_data['certname'])
```
justinstoller: there's no request_data in scope here

Magisus: Oh good catch. I'll test this branch (the case where you're asking PDB for facts for a node that doesn't exist) to make sure the rest of it works okay.

Magisus: Confirmed providing no facts and PDB also not having facts works. Though I would like to know whether that is good behavior. Should we allow catalogs to be compiled when no facts are available?

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -0,0 +1,160 @@
...
+        node
+      end
+
+      def convert_java_args_to_ruby(hash)
+        Hash[hash.collect do |key, value|
+            # Stolen and modified from params_to_ruby in handler.rb
+            newkey = key.to_s
+            # Java::ClojureLang::Keywords retain their leading colon when
+            # converted to a string
+            newkey.delete_prefix!(':')
```
tvpartytonight: I think this can be fixed later, but we should probably use [stringify-keys](https://clojuredocs.org/clojure.walk/stringify-keys) on the clojure side prior to calling compileCatalog.

Magisus: I can add that. That seems maybe better.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
justinstoller: I don't think that's a valid thing to do, though I'd defer to Sorenson on that kind of thing.

Magisus: Okay. In that case maybe this should raise. It seems... wrong to me to allow that.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
Magisus: Ended up instead changing the JSON parsing to not turn these keys into keywords in the first place.

PR #1980, opened 2019-02-15
(maint) Updates to the nightly containers

underscorgan:


underscorgan:
Idunno if we actually want to maintain these if we're moving to alpine / not using packaged versions of the software, just noticed with other changes I had been working on that this was broken

PR #1982, opened 2019-02-22
(MAINT) Replace deprecated --link usage

nwolfe:
Explicitly create and destroy our own docker network when running
container spec tests.

Hopefully this results in fewer network transients in CI as we were
seeing problems with resolving `puppet:8140`.

----

Also in this PR ...

(MAINT) Configurable image name

Allow the Docker image namespace "puppet" (i.e. puppet/puppetserver) to
be configurable, which makes it easier for local development of the
image.

We don't want to push these images to DockerHub ever, so we don't honor
the namespace override in the `make publish` target.

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -1,3 +1,4 @@
+NAMESPACE ?= puppet
 git_describe = $(shell git describe)
 vcs_ref := $(shell git rev-parse HEAD)
 build_date := $(shell date -u +%FT%T)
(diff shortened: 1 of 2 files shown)


Magisus:
So this is tangential, but I was looking at using https://hub.docker.com/r/puppet/puppetdb/ the other day, and noticed our instructions to use `--link` here. I noticed it was deprecated when I was trying to figure out what it did. If we think it's not advisable to use `link`, can we update that page to recommend something different and/or link to docs describing how we think this should actually be set up?

Of course, if we think no one in the real world is using any of these containers standalone (preferring pupperware), I guess it matters less. My use case (which I ended up abandoning), was to try to run PuppetDB in a container against a Puppetserver running locally from source (lein repl), which seems like maybe a valid dev workflow that could be worth documenting at least internally, given how complicated it is to set up PDB without packages.

justinstoller:
I appreciate the PR and, to the extent I understand what it's doing, approve. However it looks like Azure might not have a needed plugin to manage networks like this? All of the Azure tests failed.

nwolfe:
Yep, turns out we were using the deprecated feature for a reason!

Linux Containers On Windows (LCOW) doesn't yet work with `docker network create`.

The other commit in this PR is still valid so going to split that out as a new PR.

Also going to add a comment to our spec tests.

Closing this.

nwolfe:
Follow-up PR: https://github.com/puppetlabs/puppetserver/pull/1984

nwolfe:
> given how complicated it is to set up PDB without packages.

FYI the puppetdb image is now built from source, not packages, so this should be way easier now.

Just `git clone puppetdb && cd docker && make build`

PR #1983, opened 2019-02-22
(SERVER-2457) Save facts and catalogs when appropriate

justinstoller:


justinstoller:
Requires #1979 , only 5001665 is pertinent to this PR.

justinstoller:
This still needs testing/tests, but if anyone wants to give it a review as far as structure and style I'd appreciate it.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -722,9 +722,10 @@
                      :msg (format "Error parsing JSON: %s" e)})))]
     (schema/validate
      {(schema/required-key "certname") schema/Str
-      (schema/required-key "persistence") {(schema/required-key "facts") schema/Str, (schema/required-key"catalogs") schema/Str, (schema/required-key "reports") schema/Str}
-      (schema/optional-key "trusted_facts") {(schema/required-key "values") {schema/Any schema/Any}}
-      (schema/optional-key "facts") {(schema/required-key "values") {schema/Any schema/Any}}
+      (schema/required-key "persistence") {(schema/required-key "facts") schema/Bool
+                                           (schema/required-key"catalogs") schema/Bool}
```
Magisus: Missing a space here.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -6,6 +6,7 @@ class Compiler
 
       def initialize
         set_server_facts
+        @adapters_info = collect_adapters_info
```
Magisus: I like not using indirector terminology here.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -48,11 +81,11 @@ def convert_java_args_to_ruby(hash)
...
           # Stolen and modified from params_to_ruby in handler.rb
           if value.java_kind_of?(Java::ClojureLang::IPersistentMap)
-            [newkey, convert_java_args_to_ruby(value)]
+            [key, convert_java_args_to_ruby(value)]
           elsif value.java_kind_of?(Java::JavaUtil::List)
-            [newkey, value.to_a]
+            [key, value.to_a]
           else
-            [newkey, value]
+            [key, value]
```
Magisus: @tvpartytonight just put up a PR (#1987) to fix this. We can either close that or rebase this to pick that up if we merge it first.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -150,6 +183,158 @@ def set_server_facts
           end
         end
       end
+
+      # For indirection `:facts` returns Hash like:
+      #   {yaml:     Puppet::Node::Facts::Yaml,
+      #    puppetdb: Puppet::Node::Faccts::Puppdb,
```
Magisus: extra 'c' here

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -150,6 +183,158 @@ def set_server_facts
...
+        # Will be a symbol that can be passed into `Puppet.setting[<here>]`
+        terminus_setting = indirection_instance.terminus_setting
+
+        return indirected_class_reference, cache_terminus_name,
+          primary_terminus_name, terminus_setting
+      end
+
+      # Returns a Hash with symbol keys naming each indirection (eg :facts)
+      # Each indirection key refers to a Hash with the configured termini
+      # (name and class) for its primary and cache usages, if a termini is
```
Magisus: "a terminus"

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -150,6 +183,158 @@ def set_server_facts
...
+      #        :actual_terminus_class  => Puppet::Resource::Catalog::Compiler,
+      #        :terminus_setting       => :catalog_terminus,
+      #        :termini=>
+      #         {:compiler      => Puppet::Resource::Catalog::Compiler,
+      #          :json          => Puppet::Resource::Catalog::Json,
+      #          :msgpack       => Puppet::Resource::Catalog::Msgpack,
+      #          :rest          => Puppet::Resource::Catalog::Rest,
+      #          :store_configs => Puppet::Resource::Catalog::StoreConfigs,
+      #          :yaml          => Puppet::Resource::Catalog::Yaml,
+      #          :puppetdb      => Puppet::Resource::Catalog::Puppetdb}},
```
Magisus: Do we have a use for this list yet? I don't mind it being here, just curious.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -150,6 +183,158 @@ def set_server_facts
...
+
+          cache_terminus_class = find_terminus_class(indirection, cache_terminus_name)
+
+          actual_cache_name, actual_cache_class =
+            maybe_deref_storeconfigs(indirection,
+                                       cache_terminus_name,
+                                       cache_terminus_class)
+
+
+          # Every indirection needs a primary termini, cache however does not.
```
Magisus: "a primary terminus"

src/ruby/puppetserver-lib/puppet/server/compiler.rb
justinstoller: No, I started just collecting the info that the indirection knew about. Was useful when building this. But I don't think it'd be needed unless we let people choose specific termini to use. I can remove it.

justinstoller:
Up to 6a2ce60 has been built and lives here [6.2.2.SNAPSHOT.2019.02.25T1522](http://builds.delivery.puppetlabs.net/puppetserver/6.2.2.SNAPSHOT.2019.02.25T1522/). I was able to see that `replace_facts` and `replace_catalog` had been called and got a uuid from the respective termini.

I was not able to query facts for the node, though I was able to query and retrieve a catalog for the node. Don't know if that's my ability to use PDB or not.

Writing tests could be going better (:

spec/puppet-server-lib/puppet/jvm/compiler_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      }
+    }
+  end
+
+
+  it 'can be instantiated' do
+    Puppet::Server::Compiler.new
+  end
+
+  it 'calls termini to save as needed' do
```
justinstoller: This is a pretty sad test and just checks that our code in a few of paths doesn't blow up, but not that our inputs to the termini are correct. I don't know if we want to try to put expectations on the content of the requests given to them (and/or return a valid objects from `find`), or defer that kind of checking to acceptance tests.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -14,11 +15,59 @@ def compile(request_data)
...
       end
 
       private
 
+      def maybe_save(processed_hash, facts, catalog)
+        nodename = processed_hash['certname']
+        environment = processed_hash['environment']
+        persist = processed_hash['persistence']
+
+        Puppet.warning(facts.inspect)
```
justinstoller: whoops

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -14,11 +15,59 @@ def compile(request_data)
...
+        if persist['facts']
+          save_artifact(:facts, facts, nodename, environment)
+        end
+
+        if persist['catalog']
+          # TODO: We need to pass transaction_id in here some how.
```
justinstoller: I think this is my remaining hard todo that I know of (though I should do another round of manual testing). Did you find anything out about transaction_ids @Magisus , I vaguely recall you mentioning the three different ids we need to work with yesterday.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
justinstoller: Yeah, I'll rebase this out once everything else looks good.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
justinstoller: Okay, looked it up in the [docs](https://puppet.com/docs/puppet/6.2/http_api/http_catalog.html) (which I think you linked to yesterday). Looks like transaction_uuid can be provided by the http api caller. This may be unnecessary if not being persisted? If provided by the caller then it should be given to the compiler, saved with the puppetdb catalog terminus, and should be submitted back to the report processor. Does that sound correct?

justinstoller: Also, does that mean we should update the http body schema to allow optional keys listed in the linked documentation? Sorry if that's something we already talked about (I remembered my schema updates to just be around removing reports from persistence).

justinstoller: I plumbed transaction_uuid through the saving, though I haven't taken on any of the other api parameters in those docs.

justinstoller:
I think this is ready for feedback again, especially if @tvpartytonight or @pcarlisle want to chime in. After that I'll clean up the commits.

spec/puppet-server-lib/puppet/jvm/compiler_spec.rb
Magisus: I think I'd rather test this further at a higher level. Those kinds of details in tests like this make it even harder to change any of that stuff later (e.g. if for some reason we stop using this machinery).

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -14,11 +15,55 @@ def compile(request_data)
         node = create_node(processed_hash)
 
         catalog = Puppet::Parser::Compiler.compile(node, processed_hash['job_id'])
+
+        maybe_save(processed_hash, node.facts, catalog)
```
Magisus: Does this include trusted facts? Should it?

Magisus:
I'll try this out tomorrow.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
justinstoller: I don't believe so. PDB gets trusted facts by looking up the "local" trusted data for the node attached to the request. 

https://github.com/puppetlabs/puppetdb/blob/master/puppet/lib/puppet/indirector/facts/puppetdb.rb#L12-L28

Not saying that I have much faith in this implementation, but it looks right to me reading that code.

Magisus: Hmm... I wonder if normally `:trusted_information` is set to something in the context. We might need to look into that.

spec/puppet/indirector/this_is_stupid.rb
```diff
@@ -0,0 +1,9 @@
+require 'puppet/indirector/terminus'
+
+module Puppet
+  module Indirector
+    class ThisIsStupid < Puppet::Indirector::Terminus
```
pcarlisle: :suspect:

src/ruby/puppetserver-lib/puppet/server/compiler.rb
Magisus: So turns out PDB overwrites whatever is in the facts object being saved https://github.com/puppetlabs/puppetdb/blob/master/puppet/lib/puppet/indirector/facts/puppetdb.rb#L28, by either generating a default or pulling data out of the context https://github.com/puppetlabs/puppetdb/blob/master/puppet/lib/puppet/indirector/facts/puppetdb.rb#L13-L14 Neither of which is the trusted facts we’re sending along. Given that, we may have to use the context? :/

Magisus: This is relevant for the ACE compilations, where both facts and trusted facts are supplied (rather than fetched from PDB) and expected to be saved.

Magisus:
Other than the trusted facts issue, this seems to behave as I would expect.

pcarlisle:
How does thread safety in the indirector affect us in this case?

justinstoller:
safety, or the lack there of?

The understanding, I at least had, was the indirector is unsafe by being a global registry that can be updated at any time if we move to doing compilations in different threads.

The idea here was to save the indirector's mapping of indirection to termini before the jruby instance is handling requests, so compiles from this entry point don't have to use the indirector for termini lookups.

Access to this mapping could be frozen, or passed into Thread.new to make it a thread local variable if mutating in a single request was required.

That understanding of the indirector may be incorrect, this implementation may not actually improve the situation any, or the benefit may not be worth the effort? Do you have a different understanding of the indirector's thread safety or opinion on this implementation?

justinstoller:
I've rebased this down in https://github.com/puppetlabs/puppetserver/pull/1999
I've also also removed the indirector configuration caching in https://github.com/puppetlabs/puppetserver/pull/2000 for side by side comparison. Closing this one in favor of one of those two.

PR #1987, opened 2019-02-25
(maint) Use correct key name in hash conversion

tvpartytonight:


tvpartytonight:
I mean, it works on my laptop. I don't know about "definitively".

PR #1990, opened 2019-02-25
(SERVER-2474) Add docs for new catalog endpoint

Magisus:


documentation/puppet-api/v4/catalog_resource.json
```diff
@@ -0,0 +1,191 @@
+{
```
Magisus: This is copied directly from the Puppet repo. Not sure if this is a good idea to duplicate this here, but I wasn't sure what else to do, not being terribly familiar with JSON Schema. Thoughts?

Magisus:
The main docs page here has some questions attached to some of the params in ALL CAPS. Maybe we can answer some of them as part of this review?

documentation/puppet-api/v4/catalog_resource.json
justinstoller: I thought there was a way to reference Puppet Docs from within server markdown pages.

justinstoller: I think puppet_conf_setting_diffs.markdown contains links like what we could use. Maybe @jbondpdx has opinions on this?

documentation/puppet-api/v4/catalog.markdown
```diff
@@ -0,0 +1,74 @@
...
+Any classes that should be applied to the node. Can be omitted, in which case the classes
+will be looked up in the configured node classifier. (WHAT HAPPENS IF NONE IS CONFIGURED?
+IS THERE A CASE WHERE WE WOULD WANT NEITHER?)
+
+#### `parameters`
+Any parameters that should be added to the node. Can be omitted. (ARE THESE PARAMETERS ALA
+https://puppet.com/docs/puppet/5.3/lang_classes.html#class-parameters-and-variables
+OR PARAMETERS ALA https://github.com/puppetlabs/puppet/blob/master/lib/puppet/node.rb#L29?
+Are those two different things or somehow the same/connected? DO THESE ALSO COME FROM THE
+CLASSIFIER SOMETIMES?)
```
justinstoller: I thought so originally, though last week I thought we found out that the classifier returned parameters as a map of keys with class names and values of another map of parameters name to value. So, now I'm actually unsure if classes and parameters should even be two separate things.

justinstoller:
I should have responded yesterday when I first looked at this. I think content-wise what you have is good. I'd leave final formatting/wording to a docs review though because I'm terrible at that. Your in docs questions, I don't know the answer two and think it's going to require some experimentation to find the current values and discussion with product/stakeholders to figure out what the answers should be.

Magisus:
@pcarlisle so just to confirm, we want to drop the ability to specify classes and parameters here, right?

justinstoller:
👍

documentation/puppet-api/v4/catalog.markdown
```diff
@@ -0,0 +1,67 @@
...
+following form:
+```
+{
+  "certname": "<node name>",
+  "persistence": { "facts": <true/false>, "catalog": <true/false> },
+  # The rest are optional:
+  "facts": { "values": { "<fact name>": <fact value>, ... } },
+  "trusted_facts": { "values": { "<fact name>": <fact value>, ... } },
+  "environment": "<environment name>",
+  "prefer_requested_environment": <true/false>
```
Magisus: Thoughts on the name?

documentation/puppet-api/v4/catalog.markdown
```diff
@@ -0,0 +1,67 @@
...
+endpoint does not require a cert for the node whose catalog is being compiled. If not
+provided, Puppet will attempt to fetch the trusted facts for the node from PuppetDB or
+from the provided facts hash.
+
+#### `environment`
+The name of the environment for which to compile the catalog. If `prefer_requested_environemnt`
+is true, override the classified environment with this param. If it is false, only respect this
+if the classifier allows an agent-specified environment.
+
+#### `prefer_requested_environment`
```
justinstoller: I know, this isn't what we talked about, but can we just say, if the api caller has specified the environment use it, if they haven't specified it use the classifier, and if not classified fall back to production/whatever's default?

Magisus: I'm pretty sure we need to support the normal rules around agent-specified environments for ACE compilations though... I guess we can see if anyone responds to the email I sent that team overnight.

Magisus: @DavidS thoughts on that question? Do agentless nodes obey the same classifier environment rules as normal nodes?

Magisus: Got a "yes" on this from Alex.

documentation/puppet-api/v4/catalog.markdown
Magisus: We decided to remove these params from the request data entirely, since we don't currently have a use case that wants to supply them.

documentation/puppet-api/v4/catalog.markdown
Magisus: This is kind of nasty, but I think unfortunately necessary.

documentation/puppet-api/v4/catalog.markdown
```diff
@@ -0,0 +1,76 @@
...
+#### `persistence` (required)
+A hash containing two required keys, `facts` and `catalog`, which when set to true will
+cause the facts and reports to be stored in PuppetDB, or discarded if set to false.
+
+#### `facts`
+A hash with a required `values` key, containing a hash of all the facts for the node.
+If not provided, Puppet will attempt to fetch facts for the node from PuppetDB.
+
+#### `trusted_facts`
+A hash with a required `values` key containing a hash of the trusted facts for a node.
```
pcarlisle: `facts` has `values` because that's how a facts object is encoded and you could include other data at the same level as `values`, but trusted facts aren't currently shaped like this. We could still keep it that way, just wanted to check

documentation/puppet-api/v4/catalog.markdown
```diff
@@ -0,0 +1,76 @@
...
+
+#### `persistence` (required)
+A hash containing two required keys, `facts` and `catalog`, which when set to true will
+cause the facts and reports to be stored in PuppetDB, or discarded if set to false.
+
+#### `facts`
+A hash with a required `values` key, containing a hash of all the facts for the node.
+If not provided, Puppet will attempt to fetch facts for the node from PuppetDB.
+
+#### `trusted_facts`
```
pcarlisle: Do we want to call this `trusted_facts`? Puppetdb I think just uses `trusted` and internally to puppet it's always `trusted_data`. I don't know how we communicate this generally outside the company.

documentation/puppet-api/v4/catalog.markdown
Magisus: I added the `values` key to `trusted_facts` also because we had talked about keeping our options open for adding adjacent keys to all hashes in our schema. That being said, there was already one bug we found snuck in around that difference in structure. So I could go either way.

documentation/puppet-api/v4/catalog.markdown
Magisus: That's a good question. I think most of our external discussion of trusted facts calls them that, "trusted facts". And that is how they were referenced by the people who wrote the tickets for this endpoint. I personally consider PDB's and Puppet's internal names confusing implementation details.

Magisus:
I need to add the ids to this.

Magisus:
Let me know if anyone has more info about those id fields that could make the descriptions clearer. I was mostly going off of https://puppet.com/docs/puppet/6.3/format_report.html#puppet-transaction-report.

Magisus:
Updated to not expose `code_id`, since puppetserver generates this instead of the caller supplying it.

documentation/puppet-api/v4/catalog.markdown
```diff
@@ -0,0 +1,84 @@
...
+#### `options`
+
+A hash of options beyond direct input to catalogs.
+
+`prefer_requested_environment`
+Whether to always override a node's classified environment with the one supplied in the
+request. If this is true and no environment is supplied, fall back to the classified
+environment, or finally, 'production'.
+
+`capture_logs'
```
justinstoller: this backtick + single mark is wrong.

documentation/puppet-api/v4/catalog.markdown
```diff
@@ -0,0 +1,84 @@
...
+
+### Schema
+
+The catalog response body conforms to the [catalog schema](./catalog.json).
+
+### Authorization
+
+All requests made to the catalog API are authorized using the Trapperkeeper-based `auth.conf`.
+For more information about the Puppet Server authorization process and configuration settings,
+see the [`auth.conf` documentation][`auth.conf`].
```
justinstoller: I don't know if it's the broken backticks in `capture_logs`, but this link doesn't render correctly in the github UI.

Magisus: No it's just wrong, fixed.

documentation/puppet-api/v4/catalog.markdown
```diff
@@ -0,0 +1,84 @@
...
+
+### Schema
+
+The catalog response body conforms to the [catalog schema](./catalog.json).
+
+### Authorization
+
+All requests made to the catalog API are authorized using the Trapperkeeper-based `auth.conf`.
+For more information about the Puppet Server authorization process and configuration settings,
+see the [`auth.conf` documentation](https://github.com/puppetlabs/puppetserver/blob/master/documentation/config_file_auth.markdown).
```
justinstoller: I think we want to use a relative url like we do at https://github.com/puppetlabs/puppetserver/blob/master/documentation/admin-api/v1/jruby-pool.markdown#relevant-configuration
?

Magisus: I can do that. I picked a file at random looking for examples, and that one had all absolute URLs. Do you have a preference, given that we don't seem to have a convention in the repo?

justinstoller: I thought that links to puppet agent docs were absolute to puppet.com/docs pages, links to other puppetserver docs were relative, and links to further info in tk service dependencies were absolute to github.

If that's not the case then maybe there's been some changing best practices in Docs? (or us just being bad at following them?) @clairecadman do you know what the best practice is for docs links?

Magisus: Looking through more files, that does seem generally applicable. I'll update this.

clairecadman: Yes that works!

PR #1992, opened 2019-02-27
(maint) Update auth.conf for v4 catalog endpoints

justinstoller:
Previously, the v4 endpoints were undeclared, so by default deny all.

However for documentation, testing, and later management purposes it is
valuable to have the two endpoint calling schemes spelled out in the
default auth.conf.

### DIFF
--- a/ezbake/config/conf.d/auth.conf
+++ b/ezbake/config/conf.d/auth.conf
@@ -10,7 +10,18 @@ authorization: {
             }
             allow: "$1"
             sort-order: 500
-            name: "puppetlabs catalog"
+            name: "puppetlabs v3 catalog from agents"
+        },
+        {
+            # Allow services to retrieve catalogs on behalf of others
+            match-request: {
+                path: "^/puppet/v4/catalog/?$"
+                type: regex
+                method: post
+            }
+            deny: "*"
+            sort-order: 500
+            name: "puppetlabs v4 catalog for services"
         },
         {
             # Allow nodes to retrieve the certificate they requested earlier


ezbake/config/conf.d/auth.conf
```diff
@@ -10,7 +10,30 @@ authorization: {
...
+        },
+        {
+            # Allow nodes to retrieve their own catalog
+            match-request: {
+                path: "^/puppet/v4/catalog/([^/]+)$"
+                type: regex
+                method: [get, post]
+            }
+            # allow: "$1"
+            deny: "*"
```
Magisus: Are we including this version now to support the work @tvpartytonight is doing? I don't think we have plans to officially support these requests any time soon, right?

justinstoller: No, I just thought we'd break out the possible workflows of what's possible. Fine to remove what we aren't publicly supporting yet.

justinstoller: Updated. I think I've addressed this, lemme know if not.

ezbake/config/conf.d/auth.conf
```diff
@@ -10,7 +10,18 @@ authorization: {
...
             sort-order: 500
-            name: "puppetlabs catalog"
+            name: "puppetlabs v3 catalog from agents"
+        },
+        {
+            # Allow services to retrieve catalogs on behalf of others
+            match-request: {
+                path: "^/puppet/v4/catalog/?$"
+                type: regex
+                method: [get, post]
```
Magisus: So I think right now if you do a GET it will always return 404. Do we want to include GET here?

justinstoller: Good catch, no we probably shouldn't do that.

PR #1994, opened 2019-02-27
(maint) add distelli manifest file for nightly pipeline

underscorgan:


docker/distelli-manifest-nightly.yml
```diff
@@ -0,0 +1,9 @@
+pe-and-platform/puppetserver-nightly:
```
underscorgan: This file matches docker/distelli-manifest.yml except for this line, and unfortunately you can't have two pipelines share this file due to this line needing to match the pipeline name and the pipeline name needing to be unique

underscorgan:
we're discontinuing nightly containers

PR #1995, opened 2019-02-28
(SERVER-2475) Return logs from compilation with response

Magisus:
This commit adds code to wrap the catalog compilation that captures any
logging emitted during the compilation and returns it along with the
catalog. Normally, this logging would just go into the main puppetserver
log, and would be interleaved with any other compilations going on,
making it very hard to tell where the messages were coming from. This
way, the logs are tied to the catalog whose compilation produced them.

Note that this means that the logs will NOT also go to Puppet Server's
normal log destination.

### DIFF
--- a/src/ruby/puppetserver-lib/puppet/server/compiler.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/compiler.rb
@@ -9,15 +9,45 @@ module Puppet
       end
 
       def compile(request_data)
-        processed_hash = convert_java_args_to_ruby(request_data)
+        catalog, logs = capture_logs do
+          processed_hash = convert_java_args_to_ruby(request_data)
 
-        node = create_node(processed_hash)
+          node = create_node(processed_hash)
 
-        catalog = Puppet::Parser::Compiler.compile(node, processed_hash['job_id'])
-        catalog.to_data_hash
+          catalog = Puppet::Parser::Compiler.compile(node, processed_hash['job_id'])
+          catalog.to_data_hash
+        end
+
+        { catalog: catalog, logs: logs }
       end
 
       private
+
+      def capture_logs(&block)
+        logs = []
+        result = nil
+        Puppet::Util::Log.close_all
+        log_dest = Puppet::Test::LogCollector.new(logs)
+        begin
+          Puppet::Util::Log.newdestination(log_dest)
+          Puppet::Util::Log.with_destination(log_dest) do
+            result = yield
+          end
+        ensure
+          Puppet::Util::Log.newdestination(:console)
+          Puppet::Util::Log.close(log_dest)
+        end
+
+        log_entries = logs.map do |log|
+          begin
+            log.to_data_hash
+          rescue
+            log
+          end
+        end
+
+        return result, log_entries
+      end
 
       def create_node(request_data)
         # We need an environment to talk to PDB
--- a/src/ruby/puppetserver-lib/puppet/server/compiler.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/compiler.rb
@@ -1,4 +1,5 @@
 require 'puppet/server'
+require 'puppet/server/log_collector'
 
 module Puppet
   module Server
@@ -43,7 +44,7 @@ module Puppet
       def capture_logs(&block)
         logs = []
         result = nil
-        log_dest = Puppet::Test::LogCollector.new(logs)
+        log_dest = Puppet::Server::LogCollector.new(logs)
         Puppet::Util::Log.with_destination(log_dest) do
           result = yield
         end
--- a/src/ruby/puppetserver-lib/puppet/server/compiler.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/compiler.rb
@@ -19,11 +19,9 @@ module Puppet
         # Default to capturing errors and warnings from compiles
         options['capture_logs'] = true unless options['capture_logs']
 
-        processed_hash = convert_java_args_to_ruby(request_data)
-
         if options['capture_logs']
           catalog, logs = capture_logs do
-            compile_catalog(processed_hash)
+            compile_catalog(request_data)
           end
 
           { catalog: catalog, logs: logs }
@@ -84,20 +82,6 @@ module Puppet
         node.add_server_facts(@server_facts)
         node
       end
-
-      def convert_java_args_to_ruby(hash)
-        Hash[hash.collect do |key, value|
-          # Stolen and modified from params_to_ruby in handler.rb
-          if value.java_kind_of?(Java::ClojureLang::IPersistentMap)
-            [key, convert_java_args_to_ruby(value)]
-          elsif value.java_kind_of?(Java::JavaUtil::List)
-            [key, value.to_a]
-          else
-            [key, value]
-          end
-        end]
-      end
-
 
       # @return Puppet::Node::Facts facts, Hash trusted_facts
       def process_facts(request_data)
--- a/src/ruby/puppetserver-lib/puppet/server/compiler.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/compiler.rb
@@ -49,11 +49,11 @@ module Puppet
         end
 
         log_entries = logs.map do |log|
-          begin
-            log.to_data_hash
-          rescue
-            log
-          end
+          log.to_data_hash
+        end.select do |log|
+          # Filter out debug messages, which may be verbose and
+          # contain sensitive data
+          log['level'] == 'warning' || log['level'] == 'error'
         end
 
         return result, log_entries
--- a/src/ruby/puppetserver-lib/puppet/server/compiler.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/compiler.rb
@@ -26,16 +26,9 @@ module Puppet
       def capture_logs(&block)
         logs = []
         result = nil
-        Puppet::Util::Log.close_all
         log_dest = Puppet::Test::LogCollector.new(logs)
-        begin
-          Puppet::Util::Log.newdestination(log_dest)
-          Puppet::Util::Log.with_destination(log_dest) do
-            result = yield
-          end
-        ensure
-          Puppet::Util::Log.newdestination(:console)
-          Puppet::Util::Log.close(log_dest)
+        Puppet::Util::Log.with_destination(log_dest) do
+          result = yield
         end
 
         log_entries = logs.map do |log|
--- a/src/ruby/puppetserver-lib/puppet/server/compiler.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/compiler.rb
@@ -8,20 +8,37 @@ module Puppet
         set_server_facts
       end
 
+      # Compiles a catalog according to the spec provided from the
+      # request.
+      # @param [Hash] request_data details about the catalog to be compiled
+      # @return [Hash] containing either just the catalog or catalog and logs,
+      #                if capturing logs was enabled
       def compile(request_data)
-        catalog, logs = capture_logs do
-          processed_hash = convert_java_args_to_ruby(request_data)
+        options = request_data['options'] || {}
+        # Default to capturing errors and warnings from compiles
+        options['capture_logs'] = true unless options['capture_logs']
 
-          node = create_node(processed_hash)
+        processed_hash = convert_java_args_to_ruby(request_data)
 
-          catalog = Puppet::Parser::Compiler.compile(node, processed_hash['job_id'])
-          catalog.to_data_hash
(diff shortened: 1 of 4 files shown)


Magisus:
This is based on what CD4PE is currently doing. Do we think this total redirection of logs is acceptable? If not, is there any way to get it to go to both places? /cc @Sharpie

hlindberg:
A concern here is that logs can reveal secrets. In the past we discussed an option where we instead return something that makes it possible to retrieve the log for a compilation in a separate request (given you have permission to do that). That is also of value in support cases where an agent run ends with a cryptic error - the first thing we ask for is typically "can you give us the log" and users don't know how to get that from the server side logs. If they instead had a key and could request them, it would be possible to retrieve the relevant logged entries. This would also enable having separate permissions for getting a catalog vs. getting the log. (The log, especially in --debug mode can reveal all kinds things and unintentionally reveal sensitive information).

Magisus:
That's a good point. This endpoint is currently planned to be locked down to only two use cases, both of which we control (CD4PE Impact Analysis and ACE), so secrets seem like less of a concern in that case. Maybe we want to consider making that switch if/when we start using this endpoint for general purpose agent runs?

Magisus:
Related question, is this something that should be configurable? Should the caller be able to specify whether or not to capture logs? Or is that something we could add later if the consumers have an issue with this?

Is there any chance that there could be a TON of logs that make the response body huge? Is that something we should worry about, or would it be on the caller in that case to make sure to stream the body?

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -9,16 +9,46 @@ def initialize
...
+      def capture_logs(&block)
+        logs = []
+        result = nil
+        Puppet::Util::Log.close_all
+        log_dest = Puppet::Test::LogCollector.new(logs)
+        begin
+          Puppet::Util::Log.newdestination(log_dest)
+          Puppet::Util::Log.with_destination(log_dest) do
+            result = yield
+          end
```
Sharpie: Is it really necessary to call both `newdestination` and `with_destination`? 

Looks like `with_destination` its self just calls `newdestination`, yields, and then calls `close` in an ensure block. Seems like we could replace the whole begin/ensure here with that.

Sharpie:
> Note that this means that the logs will NOT also go to Puppet Server's
normal log destination.

We could probably drop the `Puppet::Util::Log.close_all` and have the compilation activity go to the main log in addition to the API response. This would be useful for remote debugging as it is often much easier for a user to send `puppetserver.log` over to support than to manually replicate a failing API call made by something like CD4PE.

Sharpie:
> Is there any chance that there could be a TON of logs that make the response body huge? Is that something we should worry about, or would it be on the caller in that case to make sure to stream the body?

Having a single compilation generate large amounts of logs is a rare event, but I suppose it could happen. Allowing users to toggle the behavior via a query parameter seems like it could be fairly easy. 

Maybe could even toggle the log level with `['none', 'error', 'warn', 'info', 'debug']` being the options?

hlindberg:
with debug on there is potentially a lot of data since every lookup then produces explain output

Sharpie:
Yeah, it definitely wouldn't be a default option. However, for those times you need DEBUG level output, turning it on for one compilation is much lower impact than turning it on for all compilations.

Also, toggling debug on and off for Ruby ops currently requires a service reload as changes to `logback.xml` don't propagate down to the Ruby loggers. So, perf issues aside, it'd be a massive win in reducing the number of manual steps and disruption a user has to undertake to get DEBUG logs for a single compile.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
Magisus: Oh that's nicer. I hadn't had a chance to go look at the internals of those yet, I'll give that a shot.

Sharpie: Yeah, I think you could just replace the entire begin/end block with:

```
Puppet::Util::Log.with_destination(log_dest) do
  result = yield
end
```

And drop the `Puppet::Util::Log.close_all`. I think it's a good thing for logs related to this endpoint to show up in `puppetserver.log` --- just as logs from the `/puppet/v3/catalog` endpoint do.

Magisus:
I think I'll leave the log level off for the time being, but I'll try adding a basic toggle for now.

What is the preference for the response schema for a toggleable bit of data like that, return a `logs` key with an empty array, or do not include a `logs` key in the response?

Sharpie:
I think leaving the `logs` key out of the response is better. Then you can disambiguate between "no logs returned" and "no logs generated", which would be important if we added `error` as a level option.

Magisus:
Should there be a default to the on/off param, or should it be required?

Sharpie:
I think it could be an optional parameter. I'm squarely on the fence as to whether the default should be on or off:

  - A default of `off` makes the average response size smaller and would be safer in certain pathological scenarios.

  - A default of `on` might encourage more people to finally address their errors and deprecation warnings before they evolve into serious problems.

Sharpie:
Ok, after waffling for a few minutes: I'd say default it to "on" so that people are aware of the new feature. If it becomes a problem, we can switch the default to "off". Moving the default from "on" to "off" feels like it would be less contentious than going the other direction.

Magisus:
And I just want to reiterate that for the time being, this affects ONLY compiles for ACE and CD4PE. So the vast majority of compile runs will still behave the old way.

justinstoller:
Does CD4PE surface this info currently? This seems like a nice to have that only CD4PE could use. It'd be nicer (I think) to only return warnings and error messages that compilation produced (seems like a much higher signal to noise ratio).

Magisus:
Talked with Justin in person, I think we're going to filter out anything that isn't `error` or `warning` for the time being, due to verbosity and security concerns. We know that CD4PE is actually surfacing this data, and that ACE is not, so I think for the time being we are also going to default this to `off`. We can change that later when we support more general use of this endpoint, but for the time being, it seems not so great to be shuttling around a bunch of data that's just getting dropped into the void.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -8,17 +8,57 @@ def initialize
...
+      def compile_catalog(request_data)
+        node = create_node(request_data)
+        catalog = Puppet::Parser::Compiler.compile(node, request_data['job_id'])
+        catalog.to_data_hash
+      end
+
+      def capture_logs(&block)
+        logs = []
+        result = nil
+        log_dest = Puppet::Test::LogCollector.new(logs)
```
justinstoller: If we were going to use this destination it would be nice to change its name and remove the bits in Puppet that make it seem to only be for testing (alternatively implement our own in this code base).

Magisus: Okay, can do.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -8,17 +8,57 @@ def initialize
...
+        Puppet::Util::Log.with_destination(log_dest) do
+          result = yield
+        end
+
+        log_entries = logs.map do |log|
+          log.to_data_hash
+        end.select do |log|
+          # Filter out debug messages, which may be verbose and
+          # contain sensitive data
+          log['level'] == 'warning' || log['level'] == 'error'
```
justinstoller: What happens when compilation errors? Are there different kinds of errors? I was just thinking that there's bound to be some errors that raise rather than log, or is that incorrect? Is it ever valid to return nothing from the endpoint (besides a 500)? Would we want to return the logs but not a catalog? Not necessarily blockers but related questions I realized I don't know the answer to.

Magisus: I'll play around with the error case.

Magisus: I think most parsing and compilation errors become 500s, we seem to use `fail` a lot in the compiler code around https://github.com/puppetlabs/puppet/blob/master/lib/puppet/pops/issues.rb.

Which means that nothing is returned, not even logs. I'm not sure if there are other kinds of errors that I don't know how to trigger, but it still seems good capture `error` level messages, just in case. For sure warnings are sent out with this.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
Magisus: Updated.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
hlindberg: @Magisus that `fail` is different from Ruby `fail` (yep, it is badly named). The `fail` in the evaluator/issue logic raises an exception.
Typically all exceptions are caught and investigated by the evaluator/compiler and they are possibly decorated with location information before re-raised (possibly as a different type of exception). Some errors/warnings (those that come from validation) can result in multiple logged issues. If there was just one, it is raised as the final cause of the failure, if there are many a "there were many issues" type of exception is raised at the end.

That leaves only catastrophic errors like OOM or out of stack that would just cause the runtime to crash.

Hope that helps.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -8,17 +9,57 @@ def initialize
         set_server_facts
       end
 
+      # Compiles a catalog according to the spec provided from the
+      # request.
+      # @param [Hash] request_data details about the catalog to be compiled
+      # @return [Hash] containing either just the catalog or catalog and logs,
+      #                if capturing logs was enabled
       def compile(request_data)
+        options = request_data['options'] || {}
```
tvpartytonight: With this pushing `convert_java_args_to_ruby` to after options processing, I think this might be setting options to a `Java::ClojureLang::IPersistentMap`, which is probably not what we want? I think this might mean that we should move that `convert_java_args_to_ruby` out to the `master.rb` file and call it there, so we don't run into this issue again...

Magisus: Oh good catch. Yeah maybe I'll add that refactor we mentioned on Justin's PR here.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -8,17 +9,57 @@ def initialize
...
+        if options['capture_logs']
+          catalog, logs = capture_logs do
+            compile_catalog(processed_hash)
+          end
 
-        catalog = Puppet::Parser::Compiler.compile(node, processed_hash['job_id'])
-        catalog.to_data_hash
+          { catalog: catalog, logs: logs }
+        else
+          catalog = compile_catalog(request_data)
```
tvpartytonight: Should this be `compile_catalog(processed_hash)`?

Magisus: Yes. I think I didn't test this branch manually actually. Thanks.

Magisus:
Updated with Tony's catches.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -8,17 +9,55 @@ def initialize
...
+      # Compiles a catalog according to the spec provided from the
+      # request.
+      # @param [Hash] request_data details about the catalog to be compiled
+      # @return [Hash] containing either just the catalog or catalog and logs,
+      #                if capturing logs was enabled
       def compile(request_data)
-        processed_hash = convert_java_args_to_ruby(request_data)
+        options = request_data['options'] || {}
+        # Default to capturing errors and warnings from compiles
+        options['capture_logs'] = true unless options['capture_logs']
```
pcarlisle: won't this make it always true

Magisus: Oh, shoot, yeah, I needed a `nil?` in there. I think Justin and I had discussed also making this false by default, so this line can be removed.

pcarlisle: In this kind of case also we should prefer to save the result to a local variable instead of updating the hash. Then you can get from the hash with a default like
```ruby
capture_logs = options.fetch('capture_logs', true)
```

Magisus: Oh I forgot about `fetch`, that's much nicer. Still think it makes sense to have this off by default though.

pcarlisle: if it's off by default then no

PR #1997, opened 2019-03-01
(DOC-3947) Update auth.conf docs

clairecadman:
This commit removes EOL documentation and links to more detailed examples.

### DIFF
--- a/documentation/config_file_auth.markdown
+++ b/documentation/config_file_auth.markdown
@@ -13,32 +13,14 @@ Puppet Server's `auth.conf` file contains rules for authorizing access to Puppet
 
 The new Puppet Server authentication configuration and functionality is similar to the legacy method in that you define rules in a file named `auth.conf`, and Puppet Server applies the settings when a request's endpoint matches a rule.
 
-However, Puppet Server now has its own `auth.conf` file that uses a new HOCON format with different parameters, syntax, and functionality.
-
-> ### Aside: Changes to Authorization in Puppet Server 2.2.0
->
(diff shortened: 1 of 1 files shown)


documentation/config_file_auth.markdown
```diff
@@ -142,7 +126,7 @@ A `match-request` can take the following parameters, some of which are required:
 
     > **Note:** While the new HOCON format does not provide a direct equivalent to the [deprecated][] `method` parameter's `search` indirector, you can create the equivalent rule by passing GET and POST to `method` and specifying endpoint paths using the `path` parameter.
 
--   **`query-params` and `environment`:** For endpoints on a Puppet 4 master, you can supply the `environment` as a query parameter suffix on the request's base URL. Use the optional `query-params` setting and provide the list of query parameters as an array to the setting's `environment` parameter.
+-   **`query-params` and `environment`:** Use the optional `query-params` setting and provide the list of query parameters as an array to the setting's `environment` parameter.
```
Magisus: Do we know if "environment" is part of tk-auth here (as in, that's the only value that can be in here), or if it's just being referenced as the most common example of a query param for our APIs?

I'll look into this.

Magisus: It looks like "environment" is not actually part of tk-auth, but just an example. I think we should reword this to more generally describe query-params. Something like:

-   **`query-params`:** Use the optional `query-params` setting and provide the list of query parameters. Each entry is a hash of the param name followed by a list of its values.

And then the example we already have below is good.

clairecadman: Thanks @Magisus! I just updated this section. Let me know if there is anything else or if you're happy for it to be merged.

PR #1999, opened 2019-03-02
(SERVER-2457) Save catalog artifacts

justinstoller:
This is built on builds.d.pl.n as 6.2.2.SNAPSHOT.2019.03.01T1602


justinstoller:
I need to write better commit messages, but I didn't want to until I knew which implementation we were going with.

PR #1998, opened 2019-03-01
(SERVER-2462) Test v4 Catalog Endpoint

justinstoller:

PR #2000, opened 2019-03-02
(SERVER-2457) save compile artifacts

justinstoller:
This is built on builds.d.pl.net as 6.2.2.SNAPSHOT.2019.03.01T1538

### DIFF
--- a/src/ruby/puppetserver-lib/puppet/server/compiler.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/compiler.rb
@@ -32,8 +32,24 @@ module Puppet
       private
 
       def compile_catalog(request_data)
+        persist = request_data['persistence']
+        save_options = request_data.slice('environment', 'transaction_id', 'certname')
+
         node = create_node(request_data)
+
+        if persist['facts']
+          save_facts(node.facts, node.trusted_data, save_options)
+        end
+
+        # Note: if we change this to use the indirection we may no longer
+        # need to call `save_catalog` below. See its documentation for
+        # further info.
         catalog = Puppet::Parser::Compiler.compile(node, request_data['job_id'])
+
+        if persist['catalog']
+          save_catalog(catalog, save_options)
+        end
+
         catalog.to_data_hash
       end
 
@@ -54,6 +70,50 @@ module Puppet
         end
 
         return result, log_entries
+      end
+
+      # Typically in our use case (~early 2019, with PuppetDB configured as
+      # the primary terminus and yaml as the cache) Indirection.save will
+      # save to both the primary terminus and cache.
+      #
+      # @param [Puppet::Node::Facts] facts
+      # @param [Hash] trusted_facts
+      # @param [Hash] options
+      # @option options [String] environment    Required
+      # @option options [String] certname       Required
+      def save_facts(facts, trusted_facts, options)
+        # trusted_facts are pulled from the context in at least the PDB terminus.
+        Puppet.override({trusted_information: trusted_facts}) do
+          Puppet::Node::Facts.indirection.save(facts, options.delete('certname'), options)
+        end
+      end
+
+      # The current Compiler terminus (which is the primary terminus for the
+      # Catalog indirection) does not implement save and so we
+      # cannot call Indirection#save directly. Typically, the catalog is
+      # "saved" to PDB because PDB becomes the cache terminus and
+      # Indirection#find will attempt to cache on successful lookup.
+      #
+      # Should we begin retrieving the catalog via Indirection#find this
+      # method may become unnecessary
+      #
+      # @param [Puppet::Resource::Catalog] catalog
+      # @param [Hash] options
+      # @option options [String] environment    Required
+      # @option options [String] certname       Required
+      # @option options [String] transaction_id Optional
+      def save_catalog(catalog, options)
+        if Puppet::Resource::Catalog.indirection.cache?
+          terminus = Puppet::Resource::Catalog.indirection.cache
+
+          request = Puppet::Indirector::Request.new(terminus.class.name,
+                                                    :save,
+                                                    options.delete('certname'),
+                                                    catalog,
+                                                    options)
+
+          terminus.save(request)
+        end
       end
 
       def create_node(request_data)
@@ -91,7 +151,7 @@ module Puppet
 
       def extract_facts(request_data)
         if request_data['facts'].nil?
-          if Puppet[:storeconfigs] == true
+          if Puppet::Node::Facts.indirection.terminus.name.to_s == "puppetdb"
             facts = get_facts_from_pdb(request_data['certname'], request_data['environment'])
           else
             raise(Puppet::Error, "PuppetDB not configured, please provide facts with your catalog request.")
--- a/spec/puppet-server-lib/puppet/jvm/compiler_spec.rb
+++ b/spec/puppet-server-lib/puppet/jvm/compiler_spec.rb
@@ -1,0 +1,8 @@
+require 'puppet/server/compiler'
+
+describe Puppet::Server::Compiler do
+
+  it 'can be instantiated' do
+    Puppet::Server::Compiler.new
+  end
+end
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -722,10 +722,12 @@
                      :msg (format "Error parsing JSON: %s" e)})))]
     (schema/validate
      {(schema/required-key "certname") schema/Str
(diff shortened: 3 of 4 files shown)


src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -14,11 +14,40 @@ def compile(request_data)
...
+      def maybe_save(processed_hash, facts, catalog)
+        nodename = processed_hash['certname']
+        persist = processed_hash['persistence']
+        options = processed_hash.
+                    slice("environment", "transaction_id").
+                    map {|key, val| [key.intern, val] }.to_h
+
+        if persist['facts']
+          if Puppet::Node::Facts.indirection.terminus_class.to_s == "puppetdb"
+            Puppet.override({trusted_information: processed_hash['trusted_facts']}) do
```
Magisus: This either needs to be `processed_hash['trusted_facts']['values']` or we need to pass in trusted facts separately, like we do with regular facts. I think the latter might be more correct, because it ensures that we save any trusted facts we pulled out, even if they weren't supplied in the request. (I don't think anyone wants to use the endpoint this way, but I want to cover our bases.)

Magisus: I played with this, and the way it is now, round-tripped trusted facts -- submitted in the request, saved to PDB, then queried on the next request -- come back wrapped in `values` when you pull them back out of PDB.

justinstoller: 👍

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -14,11 +14,40 @@ def compile(request_data)
...
+
         catalog.to_data_hash
       end
 
       private
 
+      def maybe_save(processed_hash, facts, catalog)
+        nodename = processed_hash['certname']
+        persist = processed_hash['persistence']
+        options = processed_hash.
```
tvpartytonight: I'm not seeing where these `options` are used after this assignment...

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -14,11 +14,40 @@ def compile(request_data)
         node = create_node(processed_hash)
 
         catalog = Puppet::Parser::Compiler.compile(node, processed_hash['job_id'])
+
+        maybe_save(processed_hash, node.facts, catalog)
```
tvpartytonight: If the `maybe_save` function fails, will the catalog not be returned in the response? I'm not sure if failure to save warrants returning an error in the response, but, I'm not too opinionated here; just seems like if we have a catalog that is ready to be `to_data_hash`-ed, it's shame to just throw that catalog away if `maybe_save` fails.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -14,11 +14,40 @@ def compile(request_data)
...
+
         catalog.to_data_hash
       end
 
       private
 
+      def maybe_save(processed_hash, facts, catalog)
```
pcarlisle: Could have a more meaningful name than processed_hash here, which doesn't really make sense in this context since it doesn't need to be distinguished in this function from the unprocessed one

(a better design there imo would be to have the java to ruby stuff contained entirely in the file that implements the interface, since that's where the other java stuff is, then everything in this file gets passed the already "processed" data and can just call it something semantically meaningful)

Magisus: Agreed about moving it back out to the interface file.

tvpartytonight:
It may be because I grew up listening to [alternative](https://947.radio.com/) music, but I think I am partial to this simpler implementation than what is in #1999. I think the direction of the the work in #1999 is something we eventually should do, but I'm not sure if it needs to be implemented right now to save artifacts for the new catalog endpoint.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
justinstoller: I know that the pdb terminus will raise if there's an issue saving, I _think_ that Puppet will raise if a terminus isn't installed, but will double check. My impression was that this was the closest to existing Puppet behavior, while following the same error messages that we give for `find`. Even if this is consistent with existing behavior, I don't know (or have an opinion on) what desired behavior should be.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
justinstoller: Should I add that on to this PR? I was thinking about rebasing this onto @Magisus logging PR? Can you look at @Magisus ' PR @pcarlisle 	and merge if it seems good to you?

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -15,23 +15,18 @@ def compile(request_data)
 
         catalog = Puppet::Parser::Compiler.compile(node, processed_hash['job_id'])
 
-        maybe_save(processed_hash, node.facts, catalog)
+        maybe_save(processed_hash['persistence'], node.facts,
+                   processed_hash['trusted_facts']['values'], catalog)
```
Magisus: I think we should use `node.trusted_data` here. As it is, if no trusted facts are provided in the request (let's say for some reason the person includes them as a `trusted` key in `facts` instead), we will save something different than we compiled with.

Magisus:
I just came across some discussion that pointed out that in Puppet currently, facts are saved before compile. Is that behavior we want to preserve?

src/ruby/puppetserver-lib/puppet/server/compiler.rb
Magisus: I did do this refactor in my PR, so maybe that should be merged first.

justinstoller:
For some reason I had convinced myself we were caching locally prior to compile, but I suppose it doesn't make sense for it to be caching without saving.

I think that behavior is weird, but I think maintaining existing functionality is probably best. I'll move that around, good catch.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -10,15 +10,43 @@ def initialize
...
 
+      def save_facts(facts, trusted_facts)
+          if Puppet::Node::Facts.indirection.terminus_class.to_s == "puppetdb"
+            Puppet.override({trusted_information: trusted_facts}) do
+              Puppet::Node::Facts.indirection.save(facts)
+            end
+          else
+            raise Puppet::Error, "PuppetDB not configured, not saving Facts"
+          end
+        end
```
Magisus: Soo... I think we might not want to be this restrictive. Since we are just calling into the indirector now to do the save, I think we should rely on it to do the right thing, and not raise if it can't find PDB.

@tvpartytonight ran into an issue with this when trying to use our new endpoint to use the PDB module to install PDB. The facts and catalog from that run should be saved (it's a normal agent run), but they can't be because PDB isn't present yet. This kind of issue probably isn't relevant for our current known use cases, but it will be if we ever try to use this endpoint more generally.

Please push back if you think there is value in having `persistence` mean `PDB only`.

justinstoller: I'm fine with this not requiring PDB, however, I don't think the change in approach to the Indirector has an effect on that decision.

Magisus: Oh sorry, I meant as opposed to what I was originally doing when I was instantiating the PDB terminus explicitly, and we were checking settings to make sure Puppet expected that to work.

Magisus: Which I guess we are still doing for retrieval. But I guess that part makes sense?

justinstoller:
Wait on this

justinstoller:
Commit 9eccbfe, ie post rebase, is built here: builds.delivery.puppetlabs.net/puppetserver/6.2.2.SNAPSHOT.2019.03.06T1112 is ran against 0ed53c5 in https://github.com/puppetlabs/puppetserver/pull/1998 with a RedHat7 master and Ubuntu 1604 agent running only the pre-suite and v4_catalog.rb test.
```
acceptance/suites/tests/catalog/v4_catalog.rb passed in 73.78 seconds
      Test Suite: tests @ 2019-03-06 11:25:28 -0800
       Total Suite Time: 73.78 seconds
      Average Test Time: 73.78 seconds
              Attempted: 1
                 Passed: 1
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 1
```
I've now ran the full suite as well and it appears the test doesn't teardown everything correctly (iiuc, it's leaving autosign = true in the puppet.conf) and causing the next two CA related tests to fail. I'm working on that now but I don't believe it's related to this code PR.

Magisus:
LGTM

PR #2003, opened 2019-03-06
(SERVER-2478) Do not require `infra_serials` file on startup

Magisus:
This commit updates the list of required CA files to no longer include
the infra_serials file. This file is always updated (or created) when
puppetserver starts and the infra CRL feature is enabled, so it is
unnecessary to require it to be present beforehand. This requirement was
unnecessarily causing issues with upgrade workflows.


Magisus:
Updated with gem bump.

justinstoller:
travis won't tell us anything about this newest commit.

PR #2005, opened 2019-03-06
(SERVER-2468) Use node indirection in v4/catalog

pcarlisle:
In order to use the classifier (or other terminus such as an ENC) we need
to use the indirector to retrieve the node. Whether to use the requested
environment is controlled by a new parameter in the request:

```json
{"options": {"prefer_requested_environment": <boolean>}}
```

### DIFF
--- a/spec/puppet-server-lib/puppet/jvm/compiler_spec.rb
+++ b/spec/puppet-server-lib/puppet/jvm/compiler_spec.rb
@@ -62,7 +62,6 @@ describe Puppet::Server::Compiler do
         let(:options) { { 'prefer_requested_environment' => true } }
 
         it 'uses the environment in the request' do
-          node = compiler.create_node(request_data)
           expect(node.environment.name).to eq(:fancy)
         end
       end
--- a/spec/puppet-server-lib/puppet/jvm/compiler_spec.rb
+++ b/spec/puppet-server-lib/puppet/jvm/compiler_spec.rb
@@ -48,20 +48,74 @@ describe Puppet::Server::Compiler do
 
       before(:each) do
         FileUtils.mkdir_p(File.join(Puppet[:environmentpath], environment))
-
-        Puppet::Node.indirection.expects(:find).returns(
-          Puppet::Node.new(certname, environment: 'production')
-        )
       end
 
       it 'by default uses the classified environment' do
+        Puppet::Node.indirection.expects(:find).returns(
+          Puppet::Node.new(certname, environment: 'production')
+        )
+
         expect(node.environment.name).to eq(:production)
+      end
+
+      context 'and facts are not submitted' do
+        let(:facts) { nil }
+
+        it 'requests facts from pdb after classifying and attempts to classify again' do
+          pdb_environments = sequence('pdb environments')
+          Puppet::Node::Facts.indirection.terminus.stubs(:name).returns(:puppetdb)
+          compiler.expects(:get_facts_from_pdb)
+                  .with(certname, 'fancy')
+                  .in_sequence(pdb_environments)
+                  .returns(Puppet::Node::Facts.new(certname))
+          compiler.expects(:get_facts_from_pdb)
+                  .with(certname, 'production')
+                  .in_sequence(pdb_environments)
+                  .returns(Puppet::Node::Facts.new(certname))
+          Puppet::Node.indirection.expects(:find).returns(
+            Puppet::Node.new(certname, environment: 'production')
+          ).twice
+
+          expect(node.environment.name).to eq(:production)
+        end
+
+        it 'makes a limited number of attempts to retrieve a node' do
+          %w[foo bar baz qux].each do |env|
+            FileUtils.mkdir_p(File.join(Puppet[:environmentpath], env))
+          end
+          environments = sequence('environments')
+          Puppet::Node::Facts.indirection.terminus.stubs(:name).returns(:puppetdb)
+          compiler.stubs(:get_facts_from_pdb).returns(Puppet::Node::Facts.new(certname))
+
+          Puppet::Node.indirection.expects(:find).returns(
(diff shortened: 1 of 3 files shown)


src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -29,6 +29,37 @@ def compile(request_data)
...
+        end
+
+        if prefer_requested_environment
+          node.environment = environment
+        end
+
+        # Merges facts into the node parameters.
+        # Ensures that facts will be surfaced as top-scope variables,
+        # along with other node parameters.
+        node.merge(facts.values)
```
pcarlisle: I think all termini do this already, by calling `node.fact_merge` which also sanitizes facts (which we're doing separately as well)

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -729,7 +729,8 @@
       (schema/optional-key "job_id") schema/Int
       (schema/optional-key "transaction_uuid") schema/Str
       (schema/optional-key "environment") schema/Str
-      (schema/optional-key "options") {(schema/optional-key "capture_logs") schema/Bool}
+      (schema/optional-key "options") {(schema/optional-key "capture_logs") schema/Bool
+                                       (schema/optional-key "prefer_requested_environment") schema/Bool}
       (schema/optional-key "classes") [schema/Any]
       (schema/optional-key "parameters") {schema/Any schema/Any}}
```
Magisus: I think we wanted to remove these last two.

Magisus:
I verified that this works in a PE install. It will get the catalog for the node as classified.

I ran into one weird thing while testing this, namely that whatever environment is supplied in the request will be used to fetch facts from PDB, even if a different environment would end up being used to compile the catalog, based on the opinion of the classifier. I don't know if there are any negative implications of that, but it does seem maybe odd, and errors strangely if the environment in the request doesn't exist.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
justinstoller: Should we get rid of it here, then?

src/clj/puppetlabs/services/master/master_core.clj
justinstoller: I'm fine leaving that for a fixup PR.

justinstoller:
> one weird thing while testing this, namely that whatever environment is supplied in the request will be used to fetch facts from PDB, even if a different environment would end up being used to compile the catalog

Do we need to make two calls to to PDB then?

spec/puppet-server-lib/puppet/jvm/compiler_spec.rb
```diff
@@ -48,20 +48,74 @@
...
 
-      it 'by default uses the classified environment' do
         expect(node.environment.name).to eq(:production)
       end
 
+      context 'and facts are not submitted' do
+        let(:facts) { nil }
+
+        it 'requests facts from pdb after classifying and attempts to classify again' do
+          pdb_environments = sequence('pdb environments')
```
Magisus: Oh no, are we still using mocha here? :(

(I assumed we weren't, so it took me ages to figure out what this was doing :P)

Magisus: It looks like we are relying on Puppet for this configuration. Is their mocha removal going to break all our spec tests?

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -35,20 +35,42 @@ def compile(request_data)
...
+          tries = 0
+          environment = requested_environment
+          while node.environment != environment
+            if tries > 3
+              raise Puppet::Error, _("Node environment didn't stabilize after %{tries} fetches, aborting run") % { tries: tries }
+            end
+
+            environment = node.environment
+            facts = get_facts_from_pdb(certname, environment.to_s)
+            node = Puppet.override(trusted_information: trusted_facts) do
```
Magisus: So this is assuming that `trusted_facts` will never change from what was either passed in with the request or pulled out of PDB intitially. I guess that makes sense, considering what they usually consist of?

tvpartytonight:
I've built a custom package of this PR + the custom changes to route `v3/catalog` to the `compile-catalog` clojure method. All expected tests pass in smoke, including the test that uses the exported resources, `acceptance/suites/tests/00_smoke/puppetdb_integration.rb`. The version used is PACKAGE_BUILD_VERSION=6.2.2.SNAPSHOT.2019.03.12T1027.

spec/puppet-server-lib/puppet/jvm/compiler_spec.rb
pcarlisle: Yeah we're getting everything from puppet. I've experimented with not doing so and it's not too hard. I can put up something for that after this.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
pcarlisle: I think the real reason I did this is that I didn't correctly follow the logic of how getting trusted facts can depend on the facts returned from puppetdb. But it does seem true that they shouldn't change.

spec/puppet-server-lib/puppet/jvm/compiler_spec.rb
Magisus: Oh that's good to hear. Thanks for taking a look!

src/ruby/puppetserver-lib/puppet/server/compiler.rb
Magisus: I'm fine leaving this as is then.

PR #2006, opened 2019-03-07
(SERVER-2479) Plumb code_id through to the v4 catalog endpoint

tvpartytonight:
Before this PR, the schema had a `job_id` that was propagated through to be used at compile time. However, all the code indicates that the proper name should be `code_id`,  and is furthermore determined by the `current-code-id-fn`.

This PR removes that `job_id` from the schema and plumbs the code_id function through to the v4 catalog endpoint.

### DIFF
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -736,7 +736,8 @@
 
 (schema/defn ^:always-validate
   v4-catalog-fn :- IFn
-  [jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)]
+  [jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)
+   current-code-id-fn :- IFn]
   (fn [request]
     (let [request-options (decode-catalog-post-body (slurp (:body request)))]
       {:status 200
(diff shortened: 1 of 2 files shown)


joshcooper:
FWIW, `code_id` was intended to be the SHA of the per-environment git checkout used for that compile, while `job_id` was the orchestrator job that kicked off the run (which could be the same across different nodes participating in the same job).

Magisus:
I think this does solve `code_id`, which gets us closer to supporting static catalogs.

But `job_id` is something different, as Josh says, that _should_ come in with the request (and is explicitly called out by the ACE folks in the email about the schema) and be handled correctly. We need to figure out what "correctly" means for that one.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -737,20 +736,24 @@
...
     (let [request-options (decode-catalog-post-body (slurp (:body request)))]
       {:status 200
        :headers {"Content-Type" "application/json"}
        :body (json/encode
               (jruby-protocol/compile-catalog jruby-service
                                               (:jruby-instance request)
-                                              request-options))})))
+                                              (assoc request-options
+                                                     "code_id"
+                                                     (current-code-id (jruby-request/get-environment-from-request request)))))})))
```
Magisus: I think you have one too many parentheses on the end of this line.

Magisus:
So this looks like `code_id` actually doesn't need to be a param to the request? Should I remove it from the schema docs until we know if we ever need to allow people to supply it?

src/clj/puppetlabs/services/master/master_core.clj
tvpartytonight: oh, it's because it is `current-code-id-fn` instead of `current-code-id`.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -726,7 +726,6 @@
                                            (schema/required-key "catalog") schema/Bool}
       (schema/optional-key "trusted_facts") {(schema/required-key "values") {schema/Str schema/Any}}
       (schema/optional-key "facts") {(schema/required-key "values") {schema/Str schema/Any}}
-      (schema/optional-key "job_id") schema/Int
```
Magisus: I think we should leave this in, because it will need to be there ultimately.

tvpartytonight: I removed it to reduce confusion with code_id, so when we actually add support for it, it will come in more cleanly in a single PR/commit instead of just dangling unused.

tvpartytonight:
I have a parallel branch of work [here](https://github.com/tvpartytonight/puppetserver/tree/SERVER-2473). That branch has v3 catalog rerouting, as well as a commit to twizzle the `current-code-id-fn` through the v3-reroute code.

Using PACKAGE_BUILD_VERSION=6.2.2.SNAPSHOT.2019.03.11T1653 which contained the work in this PR as well as the work in the branch for SERVER-2473, I was able to get the `acceptance/suites/tests/code_commands/code_scripts.rb` passing.

PR #2007, opened 2019-03-07
fixing text file busy error at entrypoint when using nfs

a-dawg:
When using docker/kubernetes cluster depending on the underlying storage layer you could stumble upon the following which prevents puppet from starting:
```
Running /docker-entrypoint.d/10-analytics.sh
/docker-entrypoint.sh: /docker-entrypoint.d/10-analytics.sh: /bin/sh: bad interpreter: Text file busy`
```
Adding sync in the entrypoint fixes the issue.

### DIFF
--- a/docker/puppetserver-standalone/docker-entrypoint.sh
+++ b/docker/puppetserver-standalone/docker-entrypoint.sh
@@ -5,6 +5,7 @@ set -e
 for f in /docker-entrypoint.d/*.sh; do
     echo "Running $f"
     chmod +x "$f"
+    sync
     "$f"
 done
 


nwolfe:
The same change was made recently in PuppetDB: https://github.com/puppetlabs/puppetdb/pull/2887

FYI @puppetlabs/pupperware

docker/puppetserver-standalone/docker-entrypoint.sh
```diff
@@ -5,6 +5,7 @@ set -e
 for f in /docker-entrypoint.d/*.sh; do
     echo "Running $f"
     chmod +x "$f"
+    sync
```
Iristyle: Instead of multiple `sync` calls can we do what we've done in PDB to perform this only once - https://github.com/puppetlabs/puppetdb/pull/2887/files#diff-86cc2c2467f4688ca87a833d8bf198b2R5

Thanks!

Iristyle:
Superseded by #2054

PR #2008, opened 2019-03-08
(SERVER-2462) Don't force storeconfigs

justinstoller:
To enable storeconfigs for catalogs we need to set it manually in the
puppet.conf, however we only want to do this if PDB is installed.


Magisus:
This looks right, I'll try to run it unless you're already on it.

justinstoller:
Yes, I ran this on Redhat7 (with PDB) in both isolation and as part of the ful suite, and on Ubuntu18 (w/o PDB) in isolation and in the full suite.

PR #2009, opened 2019-03-08
(SERVER-2477) Wrap v4 catalog endpoint with jruby queue limit

justinstoller:
Creates the queue-limit middleware in MasterService then passes it through core methods until we get to v4-catalog-handler where we wrap the v4 catalog handler in similar JRuby related middleware. Updates tests and legacy-routes since they reach into master-core.

### DIFF
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -754,9 +754,11 @@
 (schema/defn ^:always-validate
   v4-catalog-handler :- IFn
   [jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)
+   wrap-with-jruby-queue-limit :- IFn
    current-code-id-fn :- IFn]
   (-> (v4-catalog-fn jruby-service current-code-id-fn)
       (jruby-request/wrap-with-jruby-instance jruby-service)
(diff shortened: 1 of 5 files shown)


justinstoller:
Needs to be rebased on top of Tony's PR #2006 once that's merged.

Needs a change specific integration test.

justinstoller:
@pcarlisle (Patrick Carlisle) Recently edited these files

/cc @puppetlabs/puppetserver-maintainers

Magisus:
Just to make sure I'm understanding, the use of `identity` in this PR is to avoid having to do any multi-arity or conditional things? Makes sense, I just haven't seen it before.

justinstoller:
rebased onto master

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -845,8 +847,12 @@
   v4-routes :- bidi-schema/RoutePair
   [clojure-request-wrapper :- IFn
    jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)
+   wrap-with-jruby-queue-limit
```
Magisus: Should this have a schema annotation?

justinstoller: yeah

PR #2010, opened 2019-03-11
(maint) Install nss on RHEL 8

justinstoller:
Our image of RHEL 8 does not appear to have nss installed by default
this will conditionally install or upgrade depending on platform.

---

I haven't checked whether or not nss is necessary. The pre-suite still fails locally for me after the currently failing check because libfacter isn't available for JRuby.

Will check if nss is necessary and if the libfacter problem is specifically a RHEL8 problem

### DIFF
--- a/acceptance/suites/pre_suite/foss/70_install_puppet.rb
+++ b/acceptance/suites/pre_suite/foss/70_install_puppet.rb
@@ -21,7 +21,11 @@ step "Upgrade nss to version that is hopefully compatible with jdk version puppe
     nss_package_name="nss"
   end
   if nss_package_name
-    master.upgrade_package(nss_package_name)
+    if master['platform'] != 'el-8-x86_64'
+      master.upgrade_package(nss_package_name)
+    else
+      master.install_package(nss_package_name)
+    end
   else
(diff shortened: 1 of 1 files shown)


justinstoller:
It looks like nss is no longer installed by default because an alternative service (SSSD) is providing that feature in RHEL 8, so I don't *think* its necessary. However, the libfacter issue is preventing me from confirming.

justinstoller:
Removing the nss block entirely didn't seem to cause issues with a rhel6 master, though I'm a little hesitant to pull it out entirely for all versions yet/this close to a release.

justinstoller:
Not being able to confirm this with a full run of the acceptance tests, I think I'd like to merge this since it's the most similar to our pre-RHEL 8 configuration and moves the suite forward (and I'd like to point to a run failing in CI with the libfacter issue rather than just the output from my local machine).

Magisus:
That sounds good to me. This will let us move forward for now. But it needs to target 5.3.x, can you rebase?

justinstoller:
Whoops, thank you, rebased.

Magisus:
Travis etc. won't tell us anything about this change.

PR #2012, opened 2019-03-12
(SERVER-2497) Save job_id along with catalog

Magisus:
This commit passes the `job_id` from the v4 catalog request along to
PuppetDB to be saved alongside the catalog.


Magisus:
I tested manually that along with https://github.com/puppetlabs/puppetdb/pull/2851, `job_id` successfully gets saved to PuppetDB along with the catalog.

Magisus:
It doesn't do anything with the param unfortunately... there's not really much behavior there to test.

PR #2013, opened 2019-03-12
(maint) Remove classes and params from v4 catalog schema

justinstoller:


src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -728,9 +728,7 @@
       (schema/optional-key "facts") {(schema/required-key "values") {schema/Str schema/Any}}
       (schema/optional-key "transaction_uuid") schema/Str
       (schema/optional-key "environment") schema/Str
-      (schema/optional-key "options") {(schema/optional-key "capture_logs") schema/Bool}
```
tvpartytonight: I think we want to keep these options.

justinstoller: We do, they're below. GH is just showing it as removed because there's an additional curly brace on this line now.

tvpartytonight: Well now there's a conflict, prolly because the node change introduced a new key for the options map.

justinstoller: :fistshake:

PR #2014, opened 2019-03-13
(maint) Disable PDB on RHEL 8

Magisus:
This commit updates the list of supported PDB platforms to explicitly
exclude RHEL 8, because the puppetdb module is not currently supported
on that platform. This should be undone once RHEL 8 support is added to
the module.


Magisus:
I've run the pre-suite on both RHEL 8 and RHEL 6 (to make sure PDB is still getting installed there), with success.

PR #2015, opened 2019-03-13
(maint) Install python3 on RHEL 8

Magisus:
RHEL 8 does not come with a python installation by default, so unlike on
other platforms, neither `python` nor `python3` is available as an
executable. This commit updates the test that uses python to install
python3 prior to attempting to use it.

The long term fix here is to remove Python from our test suite, see
SERVER-2501.

### DIFF
--- a/acceptance/suites/tests/code_commands/code_scripts.rb
+++ b/acceptance/suites/tests/code_commands/code_scripts.rb
@@ -204,6 +204,9 @@ end
 
 step 'SETUP: Find Python executable'
   if on(master, 'which python', :acceptable_exit_codes => [0, 1]).exit_code == 1
+    if master['platform'] == 'el-8-x86_64'
+      master.install_package('python3')
+    end
     python_bin = 'python3'
   else
     python_bin = 'python'


Magisus:
I've verified this on RHEL 8.

Magisus:
If someone just knows of a quick alternative for what this test is doing (using json.tool to validate and pretty-print JSON), I'm happy to instead remove the usage. In the meantime, I've filed a ticket about that, https://tickets.puppetlabs.com/browse/SERVER-2501, due to the time constraint of the upcoming release.

justinstoller:
I don't think the use of json.tool is necessary at all. Ruby's JSON module should be able to parse non pretty printed output just fine. Though the error for invalid JSON would be in Ruby rather than in the system call and it wouldn't be pretty printed in the logs. (of course a line pretty printing the result could be added in Ruby if that was necessary)

justinstoller:
sorry for the close, wrong button

Magisus:
I can try it.

Magisus:
Closing in favor of https://github.com/puppetlabs/puppetserver/pull/2016

PR #2016, opened 2019-03-13
(SERVER-2501) Remove use of python from acceptance tests

Magisus:
Several times as we've added new supported platforms, the state of their
default Python installation (python3 only, none) has caused problems for
our code script test, which was using Python relatively superfluously to
validate and pretty-print JSON. This commit removes that use of Python,
leaving it to the Ruby JSON parse tool to handle errors. This means we
no longer have any straggling dependencies on Python in these tests.


Magisus:
I verified that this test still passes with this removal.

PR #2017, opened 2019-03-13
(maint) Require environment in v4 catalog endpoint

justinstoller:
Previously we were not requiring the caller of the v4 catalog endpoint
to provide an environment, assuming that the caller would know when it
was needed and supply the environment in those cases.

This led to the need to create defaults, either a specific environment
like "production" or Puppet's ":current_environment" within its context.
Neither of these are good defaults as default environments could be set
by users to some value besides production and
Puppet.lookup(:current_environment) has undefined, non-thread safe
semantics.

The simplest and most backwards compatible behavior is to always require
and environment from the endpoint caller.

This may be loosened in the future if we can define more concrete
workflows that may break backwards compatibility.

### DIFF
--- a/src/ruby/puppetserver-lib/puppet/server/compiler.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/compiler.rb
@@ -35,7 +35,7 @@ module Puppet
       def create_node(request_data)
         facts, trusted_facts = process_facts(request_data)
         certname = request_data['certname']
-        requested_environment = request_data['environment'] || 'production'
+        requested_environment = request_data['environment']
         transaction_uuid = request_data['transaction_uuid']
         prefer_requested_environment =
           request_data.dig('options', 'prefer_requested_environment')
--- a/acceptance/suites/tests/catalog/v4_catalog.rb
+++ b/acceptance/suites/tests/catalog/v4_catalog.rb
@@ -65,6 +65,7 @@ PUPPETCONF
 
   base_request = {
     "certname" => compile_subject,
+    "environment" => "production",
     "persistence" => {
       "facts" => false,
       "catalog" => false
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -724,11 +724,11 @@
      {(schema/required-key "certname") schema/Str
       (schema/required-key "persistence") {(schema/required-key "facts") schema/Bool
                                            (schema/required-key "catalog") schema/Bool}
+      (schema/required-key "environment") schema/Str
       (schema/optional-key "trusted_facts") {(schema/required-key "values") {schema/Str schema/Any}}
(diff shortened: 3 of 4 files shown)


src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -218,12 +218,13 @@ def extract_trusted_facts(request_data, facts)
       end
 
       def get_facts_from_pdb(nodename, environment)
+        env = environment || 'production'
```
Magisus: I kind of don't love providing this default in two places (also https://github.com/puppetlabs/puppetserver/blob/master/src/ruby/puppetserver-lib/puppet/server/compiler.rb#L38), but I'm not sure if it's worth passing `environment` separately all the way through...

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -218,12 +218,13 @@ def extract_trusted_facts(request_data, facts)
...
         pdb_terminus = Puppet::Node::Facts::Puppetdb.new
         request = Puppet::Indirector::Request.new(pdb_terminus.class.name,
                                                   :find,
                                                   nodename,
                                                   nil,
-                                                  :environment => environment)
+                                                  :environment => env)
```
justinstoller: Alternatively we can pass an options hash in here that just doesn't contain the `environment` key. In that case the `Indirector` will use whatever is the value of `Puppet.lookup(:current_environment)`. It might be possible that that value isn't "production"? I assume `:current_environment` is the environment of both the master and the agent who we're compiling the catalog for, but that other code hasn't had the chance of changing the  environment yet?

justinstoller: Also, we should probably update the environment option that we pass into save methods with whatever node resolves to (as is it will be `:current_environment`).

justinstoller:
@Magisus, I talked face to face with @pcarlisle . And we thought this was probably the safest route forward. What do you think?

Magisus:
Yeah that works for me. I know that our existing APIs generally require an environment, so this is at least consistent with that.

justinstoller:
Using [6.2.2.SNAPSHOT.2019.03.13T1623](http://builds.delivery.puppetlabs.net/puppetserver/6.2.2.SNAPSHOT.2019.03.13T1623/) this test passed on my local machine:

```acceptance/suites/tests/catalog/v4_catalog.rb passed in 76.08 seconds
      Test Suite: tests @ 2019-03-13 16:36:19 -0700
       Total Suite Time: 76.08 seconds
      Average Test Time: 76.08 seconds
              Attempted: 1
                 Passed: 1
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 1
```

Magisus:
I'll get something up to update the docs.

PR #2019, opened 2019-03-15
(maint) Do not test catalog endpoint in PE

justinstoller:


justinstoller:
I should test this locally before we merge, though this is the way we skip the PDB test in PE.

Magisus:
I found `confine :except, :type => 'pe'` in the `split_external_cas` test, which results in the following in our PE testing on Jenkins:
```
Standard Output

        Warning: No suitable hosts without: {:type=>"pe"}

No suitable hosts found without {:type=>"pe"}
Begin teardown
End teardown
```

They probably both do the same thing under the hood.

justinstoller:
Looks good,
`Warning: puppetserver/acceptance/suites/tests/catalog/v4_catalog.rb skipped in 0.00 seconds`

Magisus:
Travis is meaningless here.

PR #2023, opened 2019-03-19
(maint) Use rspec expectations instead of mocha

pcarlisle:
(maint) Use rspec expectations instead of mocha

This adds our own spec_helper file and moves us to using rspec expectations instead of mocha.

I also backported a test fix from master.


Rakefile
```diff
@@ -197,7 +198,7 @@ task :spec => ["spec:init"] do
     BUNDLE_GEMFILE='#{PUPPET_SRC}/Gemfile' \
     GEM_HOME='#{TEST_GEMS_DIR}' GEM_PATH='#{TEST_GEMS_DIR}' \
     lein #{profile} run -m org.jruby.Main \
-      -I'#{PUPPET_LIB}' -I'#{PUPPET_SPEC}' -I'#{FACTER_LIB}' -I'#{PUPPET_SERVER_RUBY_SRC}' \
```
Magisus: I think we can remove the `PUPPET_SPEC` definition above with this change.

pcarlisle: done

Magisus:
I can run these locally with no issue.

Magisus:
https://github.com/puppetlabs/puppetserver/pull/2024 should be merged first, to avoid merging up the mocha removal while there is still a test in master relying on it.

PR #2030, opened 2019-03-26
(FM-7881) Auth.conf v4 catalog parameterised allow

willmeek:
An an allow to the v4 catalog request, replacing the deny catchall

### DIFF
--- a/ezbake/config/conf.d/auth.conf
+++ b/ezbake/config/conf.d/auth.conf
@@ -17,9 +17,9 @@ authorization: {
             match-request: {
                 path: "^/puppet/v4/catalog/?$"
                 type: regex
-                method: post
+                method: [get, post]
             }
-            deny: "*"
+            allow: "$1"
             sort-order: 500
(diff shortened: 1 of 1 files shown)


justinstoller:
The v4 catalog endpoint doesn't currently accept a certname as a url segment like the v3 catalog endpoint (which is what is referenced by the `$1` in its `allow: $1`) so that won't work here.

However, the v4 catalog endpoint currently allows a requester to compile a catalog for _any_ node. This enables allowed API callers to mimic the capabilities of `puppet master --compile` from version 5.x and below. Being able to compile a catalog for _any_ node is considered a huge security risk and we want users who want that behavior to explicitly opt-in by either whitelisting specific nodes or cert extensions that are trusted.

The default of deny-all was chosen for that reason, afaik we don't plan to change it, and if we did it would have to be after some discussion in a ticket, I think.

Sorry to do this, efforts are very much appreciated, but I'm going to close this because as-of-now this is intended.

PR #2036, opened 2019-03-27
(SERVER-115) Avoid potential corruption by serializing crl writes

rbrw:
Note that this is just a trivial mitigation -- it does not prevent
corruption that might result from a server crash while writing, nor
does it prevent simultaneous crl read operations from seeing partial
updates.

To fix those, we might rework the code to make the updates to a temp
file while holding the lock, and then use Files/move to ATOMIC_MOVE
the update into place.


rbrw:
...a quick demonstration of the "general idea", but I'm happy to work on it further if y'all don't just want to pursue it yourselves.

PR #2038, opened 2019-03-28
(maint) Use publicly available version of rbac-client

pcarlisle:
We inadvertently made this version of puppetserver require a version of
rbac-client that was not publicly available.


Magisus:
We should also do this update in pe-pse and pe-puppetserver, just to keep everything consistent.

PR #2039, opened 2019-03-28
(maint) Updates for container analytics

underscorgan:


docker/puppetserver-standalone/docker-entrypoint.d/10-analytics.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-if [ "${PUPPERWARE_DISABLE_ANALYTICS}" = "true" ]; then
+if [ "${PUPPERWARE_ANALYTICS_ENABLED}" = "false" ]; then
     echo "($0) Pupperware analytics disabled; skipping metric submission"
```
nwolfe: FYI I had also updated the logic and message to match the default, like this: https://github.com/puppetlabs/r10k/commit/032d2b3ddd8a1fb2578291aca78062821f739fc7#diff-2f5e7558956c84e60baaf3a67c971dc9

But then we removed the message from r10k altogether.

Just wanted to mention all the parts related to the variable rename.

docker/spec/puppetserver_spec.rb
```diff
@@ -58,7 +58,7 @@ def puppetserver_health_check(container)
 
     result = run_command("docker run --rm --detach \
                    --env DNS_ALT_NAMES=puppet \
-                   --env PUPPERWARE_DISABLE_ANALYTICS=true \
+                   --env PUPPERWARE_ANALYTICS_ENABLED=false \
```
nwolfe: I just deleted these lines in r10k since it defaulted to false but not a biggie

underscorgan:
no idea why travis says this is still building, it definitely finished successfully 2 days ago.

PR #2040, opened 2019-03-29
(maint) Don't fail on an empty status in container tests

underscorgan:


underscorgan:
This fixes an issue that was causing puppetserver container tests to fail in our Pipelines setup.

nwolfe:
Can this cause an endless loop?

What if the status is "starting", and then the service dies, will it return a non-empty status here? Otherwise the loop would then just continue forever, right?

underscorgan:
ahh, good catch. Looks like it's possible, I'll add some logic to have that timeout after a few consecutive empty responses, the failures I've seen have all involved intermittent single empty responses.

underscorgan:
@nwolfe updated!

docker/spec/puppetserver_spec.rb
```diff
@@ -7,9 +7,22 @@
...
+    # assume that 5 empty responses in a row is a failure case
+    while status.empty? && empty_response_counter < 5
+      result = run_command("docker inspect \"#{container}\" --format '{{.State.Health.Status}}'")
+      status = result[:stdout].chomp
+      if status.empty?
+        empty_response_counter += 1
+        sleep 1
+      end
+      STDOUT.puts "queried health status of #{container}: #{status}"
+    end
```
nwolfe: FYI we've been using a time-based timeout in our spec tests, like this: https://github.com/puppetlabs/puppetdb/blob/master/docker/puppetdb/spec/puppetdb_spec.rb#L15-L24

Iristyle: Ah yeah, we'll probably refactor this a bit with the shared helpers.

More to come!

PR #2047, opened 2019-04-11
(DOC-3964) Add note re. `systemctl restart puppet`

clairecadman:
This commit adds a note to address a user feedback question re. the `systemctl restart puppet command`.

### DIFF
--- a/documentation/restarting.markdown
+++ b/documentation/restarting.markdown
@@ -66,3 +66,6 @@ There are three ways to trigger your Puppet Server environment to refresh and pi
 * Changes to [`ca.cfg`](./configuration.markdown#service-bootstrapping) to enable or disable Puppet Server's certificate authority (CA) service.
 
 For these types of changes, you must restart the process by using the operating system's service framework, for example, by using the `systemctl` or `service` commands.
+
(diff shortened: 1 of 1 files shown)


documentation/restarting.markdown
```diff
@@ -66,3 +66,6 @@ There are three ways to trigger your Puppet Server environment to refresh and pi
 * Changes to [`ca.cfg`](./configuration.markdown#service-bootstrapping) to enable or disable Puppet Server's certificate authority (CA) service.
 
 For these types of changes, you must restart the process by using the operating system's service framework, for example, by using the `systemctl` or `service` commands.
+
+> Note: To ensure that the Puppet master and CA service is running in a platform agnostic way, use the `puppet resource service puppetserver ensure=running` command. This command is equivalent to  `systemctl start puppetserver`. For more information on the resource command and managing a server’s desired state, see
```
justinstoller: I'm sorry, can we say

> This command is equivalent to  `systemctl start puppetserver` _on systems that support it_.

instead of

> This command is equivalent to  `systemctl start puppetserver`.

PR #2048, opened 2019-04-16
(maint) Update auth.conf test for new PE report rule

Magisus:
The deafult rule for the report endpoint in PE was recently updated.
This test now verifies both versions of the rule.


Magisus:
I tested this out locally, with success.

justinstoller:
Should this go into 6.3.x?

Magisus:
Oh, yes, probably, since I guess that's the stream that will stay in Kearney.

justinstoller:
You know nothing, Travis CI.

PR #2053, opened 2019-04-25
(maint) Clear Facter domain fact in Docker

Iristyle:
- Rather than relying on an FQDN everywhere, rely on short names
   for hosts inside the cluster.

- This could very well be a terrible idea!


Iristyle:
I think we have other ways of achieving this goal (like setting `PUPPETSERVER_HOSTNAME` value in `docker-compose.yml` to make sure these values are set appropriately - https://github.com/puppetlabs/puppetserver/blob/349692aa4519a56c83093efb858042012d0e2c6b/docker/puppetserver-standalone/docker-entrypoint.d/50-set-certname.sh#L4)

Removing the domain is pretty heavy-handed

PR #2055, opened 2019-04-29
(maint) Puppetserver shouldn't include facter 2

underscorgan:
Due to dependencies of the puppetserver-ca gem, the puppetserver package
includes the facter2 gem. This can lead to strange behavior in some
cases. This commit updates the mri gem installation to run with
`--no-dependencies` to not pull in facter2. This will break if that gem
ends up with additional dependencies, or if puppetserver ever removes
its dependency on the puppet-agent.


underscorgan:
This affects at least as far back as puppetserver 5.3.8 so I'll want to retarget it, but am looking for some guidance as to where this should be targeted.

underscorgan:
replaced by https://github.com/puppetlabs/puppetserver/pull/2056, targeting the correct branch

PR #2056, opened 2019-04-29
(maint) Puppetserver shouldn't include facter 2

underscorgan:
Due to dependencies of the puppetserver-ca gem, the puppetserver package
includes the facter2 gem. This can lead to strange behavior in some
cases. This commit updates the mri gem installation to run with
`--no-dependencies` to not pull in facter2. This will break if that gem
ends up with additional dependencies, or if puppetserver ever removes
its dependency on the puppet-agent.


justinstoller:
Is this because Puppet Server gems aren't resolved when Facter 3 is in the gem path, or will Facter 3 never resolve as the correct Facter dependency because it's not released on Rubygems? (I'll try this on a machine, thought I'd ask in case you knew off the top of your head)

If `gem install` will always resolve to Facter 2 then we might want to remove Facter as a runtime dependency? We've told folks that they can upgrade/install the gem out of band to fix an issues there are in their packages.

underscorgan:
@justinstoller it seems like facter3 is never resolving. If I try to install puppetserver-ca on a box using the agent ruby/gem i can see facter3 listed in the `gem list` but the `gem install` still pulls in 2.5.1.

I guess the main question for removing facter as a runtime environment is do we expect this gem to only be installed on a box that has puppet-agent installed? I'm not sure how heavily the gem depends on facter or what may go sideways if the dependency is removed.

PR #2059, opened 2019-05-01
(SERVER-2471) environment transports endpoint pt.1

justinstoller:


justinstoller:
This creates a new endpoint `/puppet/v3/environment_transports` that behaves like the existing `/puppet/v3/environment_classes` endpoint. It currently returns hardcoded dummy data, has a major bug that when called will invalidate the _classes_ endpoint's cache (ticketed as [SERVER-2532](https://tickets.puppetlabs.com/browse/SERVER-2532)), with both the implementation and tests copied from the classes endpoint w/ modifications as needed.

At this point I'd like to see both the implementation and tests refactored so that as much could be shared between the environment classes, modules, and transports endpoints as possible. However the size of the PR was becoming unwieldy and I wanted to run it by folks to see if this intermediate state was at a place we were comfortable reviewing/merging or if I should do the refactor first. The refactor might also go well with SERVER-2532.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -815,33 +885,36 @@
...
+        tasks-handler (all-tasks-handler jruby-service)
 
-        all-tasks-handler
-        (all-tasks-handler jruby-service)
+        transport-handler (environment-transport-handler jruby-service
+                                                         cache-enabled)
 
-        task-details-handler
-        (task-details-handler jruby-service get-code-content-fn current-code-id-fn)
+        task-handler (task-details-handler jruby-service
```
justinstoller: Ugh, I accidentally rebased away a maint commit where I changed the names of these handlers. Having them be the same name as fns w/in the namespace really bothered me. I can break this back out.

test/integration/puppetlabs/services/master/environment_transports_int_test.clj
```diff
@@ -0,0 +1,638 @@
...
+           ;; The test has the following basic steps:
+           ;;
+           ;; 1) Purge the current environment files on disk and hit the
+           ;;    environment-cache endpoint to flush the cache for all
+           ;;    environments, basically to ensure nothing is left around from
+           ;;    other tests.
+           ;; 2) Populate code for the 'test' and 'production' environments and
+           ;;    see that environment_classes queries return the right info for
+           ;;    them.
+           ;; 3) Do one more environment_classes query for each environment,
```
Magisus: I think this comment should be referring to `environment_transports` throughout?

test/integration/puppetlabs/services/master/environment_transports_int_test.clj
```diff
@@ -0,0 +1,638 @@
...
+         (is (= expected-response
+                (response->map response))))))))
+
+
+(deftest ^:integration environment-transports-integration-cache-enabled-test1
+  (let [debug-log "./target/environment-transports-integration-cache-enabled.log"]
+    (fs/delete debug-log)
+    (bootstrap/with-puppetserver-running app
+     {:global {:logging-config
+               (str "./dev-resources/puppetlabs/services/"
```
Magisus: Are we just doing this logging switchover to make the tests easier to debug if something goes wrong? I don't see us actually looking at the resulting logs anywhere...

test/integration/puppetlabs/services/master/environment_transports_int_test.clj
```diff
@@ -0,0 +1,638 @@
...
+           ;; set, done after the code on disk for both environments has changed.
+           ;;
+           ;; The test has the following basic steps:
+           ;;
+           ;; 1) Purge the current environment files on disk and hit the
+           ;;    environment-cache endpoint to flush the cache for all
+           ;;    environments, basically to ensure nothing is left around from
+           ;;    other tests.
+           ;; 2) Populate code for the 'test' and 'production' environments and
+           ;;    see that environment_classes queries return the right info for
```
Magisus: Same here, I think this should be `environment_transports`.

test/integration/puppetlabs/services/master/environment_transports_int_test.clj
```diff
@@ -0,0 +1,638 @@
...
+                   "etag did not change even though code did")
+               (is (= mock-response
+                      (response->map response))
+                   "unexpected body following code change")))))
+       (catch Exception e
+         (println "dumping puppetserver.log\n" (slurp debug-log))
+         (throw e))))))
+
+
+#_(deftest ^:integration environment-transports-integration-cache-enabled-test3
```
Magisus: Does this mean this test is skipped right now?

test/integration/puppetlabs/services/master/environment_transports_int_test.clj
```diff
@@ -0,0 +1,638 @@
...
+                   "unexpected body for test response")))))
+       (catch Exception e
+         (println "dumping puppetserver.log\n" (slurp debug-log))
+         (throw e))))))
+
+(deftest ^:integration
+         not-modified-returned-for-environment-transports-info-request-with-gzip-tag
+  (testing (str "SERVER-1153 - when the webserver gzips the response "
+                "containing environment_transports etag, the next request "
+                "roundtripping that etag returns an HTTP 304 (Not Modified)")
```
Magisus: If I'm reading https://tickets.puppetlabs.com/browse/SERVER-1153 right, it sounds like we may not need to be explicitly stripping `--gzip` anymore, since we're now on Jetty 9.4.x. Could removing [that stripping](https://github.com/puppetlabs/puppetserver/blob/master/src/clj/puppetlabs/services/master/master_core.clj#L244) be part of this, or would that better be done as its own thing in case it's problematic for some reason?

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -636,6 +636,76 @@
...
+   env :- schema/Str
+   jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)
+   request-tag :- (schema/maybe String)
+   cache-id :- (schema/maybe schema/Int)
+   cache-enabled? :- schema/Bool]
+  (let [resp-map (raw-transports->response-map raw-info env)
+        serialized-resp (cheshire/generate-string resp-map)]
+    (if cache-enabled?
+      (let [;; This isn't going to work with two different data sets per environmnt, is it?
+            tag (ks/utf8-string->sha1 serialized-resp)]
```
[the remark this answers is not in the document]
justinstoller: yes

test/integration/puppetlabs/services/master/environment_transports_int_test.clj
justinstoller: Yeah, there's a try and then catch block where we print out the whole log. I actually triggered these failures in my refactor and didn't find that they added much value. In fact the log was so long that I couldn't see all of my failed tests. I'll remove this.

test/integration/puppetlabs/services/master/environment_transports_int_test.clj
justinstoller: yeah, `#_` is a reader macro that comments out the following form. this whole deftest is handling updating caches after changing resources on disk, which isn't valid until we get rid of the stub ruby impl.

test/integration/puppetlabs/services/master/environment_transports_int_test.clj
justinstoller: I'm for doing it as its own thing and created [SERVER-2534](https://tickets.puppetlabs.com/browse/SERVER-2534)

PR #2060, opened 2019-05-02
Fix healthcheck when masterport is different than 8140

mcanevet:


Iristyle:
Thanks for the PR @mcanevet!

How are you initially injecting the port configuration? Are you using our container as a base for a custom container that you're maintaining? (Trying to determine if we should provide any additional configuration options through ENV vars or similar)

mcanevet:
We are using a configMap when on kubernetes or mount binding a config file when using a docker composition.

Iristyle:
I think we also want to make a similar change for https://github.com/puppetlabs/puppetserver/blob/master/docker/puppetserver-standalone/docker-entrypoint.d/90-consul.sh#L16

underscorgan:
Hi @mcanevet , thanks for this PR! Apologies for the delay in getting back to this, but I was looking at making custom master port more configurable in the containers, and I think I have something working in #2078 . 

Let me know if you need any other functionality for customizing the ports!

mcanevet:
Your solution is OK for me

PR #2062, opened 2019-05-02
 Refactor puppetserver container tests

underscorgan:


underscorgan:
Depends on https://github.com/puppetlabs/pupperware/pull/65

underscorgan:
This is ready for review now :)

docker/docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+
+services:
+  puppet:
+    hostname: puppet.test
+    image: ${PUPPET_TEST_DOCKER_IMAGE:-puppet/puppetserver-standalone}
+    environment:
+      - PUPPETSERVER_HOSTNAME=puppet.test
+      - PUPPERWARE_ANALYTICS_ENABLED=${PUPPERWARE_ANALYTICS_ENABLED:-false}
+      - DNS_ALT_NAMES=puppet,puppet.test,${DNS_ALT_NAMES:-}
+    dns_search: '.test'
```
Iristyle: It's interesting that this works, because I thought what I discovered in my experiments was that this is actually supposed to be:

```yaml
dns_search: test
```

docker/docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+      - CA_ENABLED=false
+      - CA_HOSTNAME=puppet.test
+    dns_search: '.test'
+    networks:
+      puppetserver_test:
+        aliases:
+          - compiler.test
+
+networks:
+  puppetserver_test:
```
Iristyle: 👍 

Even though this adds more to the file, we should probably make this explicit it in other compose files so we have better control of the name.

PR #2065, opened 2019-05-09
(SERVER-2471) environment transports endpoint pt.2

justinstoller:


justinstoller:
This is a refactor of what's in https://github.com/puppetlabs/puppetserver/pull/2059 so that the new transport endpoint and the old classes endpoint share the same caching logic. A more understandable view (ie the diff of this and the other PR) is here: https://github.com/justinstoller/puppetserver/compare/server2471...justinstoller:server2471-3

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -624,13 +556,68 @@
...
+  out that version). If the given and internal versions do not match the
+  request to set the tag is ignored, if they match the tag is stored and
+  the internal version is incremented."
+  [info :- {schema/Any schema/Any}
+   env :- schema/Str
+   jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)
+   request-tag :- (schema/maybe String)
+   cache-version :- (schema/maybe schema/Int)]
+  (let [body (cheshire/encode info)
+        tag (ks/utf8-string->sha1 body)]
```
Magisus: Do we have the same concerns about the SHA1 here @pcarlisle ?

pcarlisle: Since it's just for a cache I think it probably doesn't matter, though in cases like this if there's no cost to updating to sha256 I think we generally should.

PR #2068, opened 2019-05-21
(PE-26549) Document upcoming compile endpoint

justinstoller:


documentation/puppet-api/v4/compile.markdown
```diff
@@ -0,0 +1,93 @@
...
+  "transaction_uuid": "<uuid string>",
+  "job_id": "<id string>",
+  "options": { "capture_logs": <true/false>,
+               "log_level": <one of debug/info/warn>}
+}
+```
+
+#### `code_ast` (required)
+A parsed string of json encoded
+[PCore Generic Data](https://github.com/puppetlabs/puppet-specifications/blob/a87c9967ebd7071e0a548920f3262ac24cc3f4cd/language/data-types/pcore-generic-data.md)
```
justinstoller: I'm not a 100% on the correctness or value of the links in these sections, though I'm not sure what to put in here that would be helpful documentation without going on a code safari. Maybe, since the consumer of this is internal and already doing what they need to do we can get by with a cursory blurb that an AST string goes here?

documentation/puppet-api/v4/compile.markdown
```diff
@@ -0,0 +1,93 @@
...
+objects representing a Puppet Code AST, as provided in code giving a Puppet AST or Program to
```
justinstoller: I assume this should be a `Puppet::Pops::Model::Program` (since that is the top level in an AST that knows how to serialize itself to pcore) but I was losing the thread in bolt/pal/compiler so this line's a bit handwavy in an otherwise probably too detailed section.

documentation/puppet-api/v4/compile.markdown
```diff
@@ -0,0 +1,93 @@
...
+[Puppet::Pops::Serialization::ToDataConverter](https://github.com/puppetlabs/puppet/blob/68584eaf62361a47da8d825525a3423af1a388c4/lib/puppet/pops/serialization/to_data_converter.rb)
+which represents an intermediate step when compiling Puppet Code to a catalog
+when `rich_data = true` (the default in Puppet 6). The returned catalog is
```
justinstoller: I should probably just say "when rich_data is enabled" rather than use formated text.

documentation/puppet-api/v4/compile.markdown
```diff
@@ -0,0 +1,93 @@
...
+---
+
+The compile endpoint allows compilation of arbitrary pcore serialized Puppet
+Code ASTs (aka parsed, but unevaluated Puppet Code snippets). In addition to
+the code AST, the caller must provide the certname, environment, facts,
+trusted facts, and variables for compilation (the context in which this AST
+will be evaluated). Optionally, the caller may provide job_id, transaction_id,
+and configure returned logging. The request body must be JSON formatted and
+the caller must accept a JSON response. The server must not have rich_data
+disabled.
```
Magisus: Assuming it's phrased like this because it's enabled by default?

documentation/puppet-api/v4/compile.markdown
```diff
@@ -0,0 +1,93 @@
...
+
+
+## `POST /puppet/v4/compile`
+
+(Introduced in Puppet Server 6.5.0)
```
Magisus: I think we're on 6.4.0. We didn't ship a Y with the last platform release.

documentation/puppet-api/v4/compile.markdown
```diff
@@ -0,0 +1,93 @@
...
+
+The request body must look like:
+```
+{
+  "code_ast": "<pcore serialized ast",
```
Magisus: missing closing `>`

documentation/puppet-api/v4/compile.markdown
Magisus: +1

documentation/puppet-api/v4/compile.markdown
```diff
@@ -0,0 +1,93 @@
...
+
+#### `code_ast` (required)
+A parsed string of json encoded
+[PCore Generic Data](https://github.com/puppetlabs/puppet-specifications/blob/a87c9967ebd7071e0a548920f3262ac24cc3f4cd/language/data-types/pcore-generic-data.md)
+objects representing a Puppet Code AST, as provided in code giving a Puppet AST or Program to
+[Puppet::Pops::Serialization::ToDataConverter](https://github.com/puppetlabs/puppet/blob/68584eaf62361a47da8d825525a3423af1a388c4/lib/puppet/pops/serialization/to_data_converter.rb)
+which represents an intermediate step when compiling Puppet Code to a catalog
+when `rich_data = true` (the default in Puppet 6). The returned catalog is
+equivalent to a catalog returned with the MIME type
+"application/vnd.puppet.rich+json" normally.
```
Magisus: This is definitely a little hard to parse. Maybe it could be simplified a bit. Something like:
```
A parsed string of json encoded PCore Generic Data objects representing a Puppet Code AST. The When `rich_data` is enabled (the default in Puppet 6), the AST represents an intermediate step when compiling Puppet Code to a catalog. The returned catalog is equivalent to a catalog returned with the MIME type "application/vnd.puppet.rich+json" normally.
```

documentation/puppet-api/v4/compile.markdown
justinstoller: yep, should I change the way it's worded?

Magisus: No I think that makes sense given the default. And we call out that default elsewhere on this page.

documentation/puppet-api/v4/compile.markdown
justinstoller: I like that better, updated!

justinstoller:
I don't think the docker failure has anything to do with my docs update (:

documentation/puppet-api/v4/compile.markdown
```diff
@@ -0,0 +1,94 @@
...
+---
+
+[PCore Generic Data]: https://github.com/puppetlabs/puppet-specifications/blob/a87c9967ebd7071e0a548920f3262ac24cc3f4cd/language/data-types/pcore-generic-data.md
+[`auth.conf` documentation]: ../../config_file_auth.markdown
+
+The compile endpoint allows compilation of arbitrary pcore serialized Puppet
+Code ASTs (aka parsed, but unevaluated Puppet Code snippets). In addition to
+the code AST, the caller must provide the certname, environment, facts,
+trusted facts, and variables for compilation (the context in which this AST
+will be evaluated). Optionally, the caller may provide job_id, transaction_id,
```
Magisus: Can you put the fields in this block in `backticks`? I think they should stand out a bit more.

documentation/puppet-api/v4/compile.markdown
```diff
@@ -0,0 +1,94 @@
...
+A parsed string of json encoded [PCore Generic Data][]
+objects representing a Puppet Code AST. When `rich_data` is enabled (the
+default in Puppet 6), the AST represents an intermediate step when
+compiling Puppet Code to a catalog. The returned catalog is equivalent
+to a catalog returned with the MIME type `application/vnd.puppet.rich+json`
+normally.
+
+
+#### `certname` (required)
+The name of the node for whom the AST will be compiled.
```
Magisus: Can we change "whom" to "which"? It sounds a little funny reading it to myself.

Magisus:
I'll merge after those couple nits.

justinstoller:
addressed nits

Magisus:
Travis has nothing to tell us about this.

PR #2069, opened 2019-05-23
(SERVER-2541) compile endpoint - clojure work

justinstoller:
I think this, minus cleaning up commit messages is ready to do. Hopefully this works with what you're working on @Magisus.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
@@ -87,7 +87,7 @@
 
   (compile-ast
     [this jruby-instance compile-options]
-    {:foo "bar"})
+    (.compileAST jruby-instance compile-options))
 
   (get-environment-module-info
     [this jruby-instance env-name]
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
@@ -85,6 +85,10 @@
    [this jruby-instance request-options]
    (.compileCatalog jruby-instance request-options))
 
+  (compile
+    [this jruby-instance compile-options]
+    {:foo "bar"})
+
   (get-environment-module-info
     [this jruby-instance env-name]
     (.getModuleInfoForEnvironment jruby-instance env-name))
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
@@ -85,7 +85,7 @@
    [this jruby-instance request-options]
    (.compileCatalog jruby-instance request-options))
 
-  (compile
+  (compile-ast
     [this jruby-instance compile-options]
     {:foo "bar"})
 
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -788,9 +788,9 @@
       {:status 200
        :headers {"Content-Type" "application/json"}
        :body (json/encode
-               (jruby-protocol/compile jruby-service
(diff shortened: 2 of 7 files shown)


src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
```diff
@@ -81,6 +81,10 @@
    [this jruby-instance request-options]
    (.compileCatalog jruby-instance request-options))
 
+  (compile
```
justinstoller: this conflicts with a clojure var, I think I'll rename it to `compile-ast`, you're working on the Ruby side, right @Magisus ? Do you have thoughts on the protocol name? Should I implement the java interface as part of this work too?

justinstoller: renamed to `compile-ast` and implemented interface per convo in slack.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -852,13 +897,19 @@
   (let [v4-catalog-handler (v4-catalog-handler
                              jruby-service
                              wrap-with-jruby-queue-limit
-                             current-code-id-fn)]
+                             current-code-id-fn)
+        compile-handler' (compile-handler
```
Magisus: So I notice that we put the transports endpoint under v3 and this one under v4. Is there any policy we want to follow with choosing one of those going forward?

justinstoller: Sorry, missed this earlier.

I put them under versions that were similar to other work that we were doing, but that's not really a reason for _why_ they should be there. I could see this being a `v3` endpoint.

Magisus: We're already not doing regular API versioning (where the whole API is supposed to move to the new version when a change happens like the new `catalog` endpoint), so I guess it doesn't matter that much to anyone except us.

I think that given we just put the other net-new endpoint under v3, maybe this should go there too. It seems a little easier to remember that you only have to use `v4` for the one that has another meaning in `v3`.

justinstoller: moved this to the v3 api

justinstoller:
Did you still want review from @pcarlisle or @tvpartytonight, @Magisus ?

Magisus:
No that's fine. But I just realized as I was working on my other PR, can you move the docs under the v3 directory as part of this?

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -712,34 +712,50 @@
...
+   (schema/required-key "code_ast") schema/Str
+   (schema/required-key "trusted_facts") {(schema/required-key "values") {schema/Str schema/Any}}
+   (schema/required-key "facts") {(schema/required-key "values") {schema/Str schema/Any}}
+   (schema/required-key "variables") {(schema/required-key "values") {schema/Str schema/Any}}
+   (schema/optional-key "job_id") schema/Str
+   (schema/optional-key "transaction_uuid") schema/Str
+   (schema/optional-key "options") {(schema/optional-key "capture_logs") schema/Bool
+                                    (schema/optional-key "log_level") (schema/enum "debug" "info" "warn")}})
+
+(defn valid-body
```
pcarlisle: more clear if named validate-body, makes the name obviously parallel to schema/validate

pcarlisle: oh i see you want the difference of returning the body after validating.... this is still a little weird but i'm not as sure

justinstoller: `validated-body`?

justinstoller:
updated

Magisus:
Uh oh, conflict.

justinstoller:
rebased, I think this is ready for merge

PR #2070, opened 2019-05-28
(DOC-4135) Fix broken link

clairecadman:


clairecadman:
@Magisus Could we also merge this change to the 6.3.x branch?

Magisus:
Yep, in theory it will merge up automatically.

clairecadman:
Perfect, thanks!

documentation/compatibility_with_puppet_agent.markdown
```diff
@@ -7,7 +7,7 @@ canonical: "/puppetserver/latest/compatibility_with_puppet_agent.html"
 [ca.conf]: ./config_file_ca.markdown
 [auth.conf]: https://puppet.com/docs/puppet/latest/config_file_auth.html
 [future parser]: https://docs.puppet.com/puppet/3.8/experiments_future.html
-[upgrade puppet]: https://puppet.com/docs/puppet/latest/upgrade_major_pre.html
+[upgrade puppet]: https://puppet.com/docs/puppet/5.5/upgrade_major_pre.html
 [deprecated]: https://docs.puppet.com/puppetserver/2.2/deprecated_features.html
```
justinstoller: this is also broken

justinstoller:
This is weird. It's documentation about how to upgrade from Puppet 3.x to Puppet 4.x in the Puppet 6.x documentation and is updating a link to point to the Puppet 5.x docs. We might want to point to the Puppet 4.x docs (that version is no longer maintained but is the version we recommended folks go to from 3.x), or if we now want to recommend folks go directly from 3.x to 5.x we should update the wording throughout the page.

documentation/compatibility_with_puppet_agent.markdown
clairecadman: This link takes me here: https://github.com/puppetlabs/docs-archive/tree/master/puppetserver/2.2. Is that incorrect?

clairecadman:
Which one do you think we should recommend?

justinstoller:
I'd continue to recommend folks upgrade from one major to the next without skipping, even though those majors are no longer supported.

So, I wouldn't change any of the prose on the page, but I'd make https://puppet.com/docs/puppet/5.5/upgrade_major_pre.html -> https://puppet.com/docs/puppet/4.10/upgrade_major_pre.html, generally anywhere we have links to latest, or 5.x of puppet I'd point them at 4.10, and change the https://docs.puppet.com/puppetserver/2.2/deprecated_features.html link -> latest server in the 2 series: https://puppet.com/docs/puppetserver/2.7/deprecated_features.html.

clairecadman:
@justinstoller Thanks! I've updated these links.

PR #2072, opened 2019-05-31
(maint) Update puppetserver healthcheck to not use `puppet config print`

underscorgan:
Rather than using `puppet config print` to determine hostname and cert
locations, we hard-code directory paths and figure out the certname
based on what certificate we find on the filesystem.

This has some significant performance improvements over the previous
healthcheck.

Tested on a macbook, with old script:
$ time ./healthcheck.sh

real    0m2.405s
user    0m2.130s
sys     0m0.160s

tested on the same macbook, with new script:
$ time ./healthcheck.sh

real    0m0.092s
user    0m0.030s
sys     0m0.000s

### DIFF
--- a/docker/puppetserver-standalone/healthcheck.sh
+++ b/docker/puppetserver-standalone/healthcheck.sh
@@ -4,7 +4,7 @@ set -x
 set -e
 
 certname=$(ls /etc/puppetlabs/puppet/ssl/certs | grep --invert-match ca.pem) && \
-hostname=${certname%.*} && \
+hostname=$(basename $certname .pem) && \
 hostprivkey=/etc/puppetlabs/puppet/ssl/private_keys/$certname && \
 hostcert=/etc/puppetlabs/puppet/ssl/certs/$certname && \
 localcacert=/etc/puppetlabs/puppet/ssl/certs/ca.pem && \
--- a/docker/puppetserver-standalone/healthcheck.sh
+++ b/docker/puppetserver-standalone/healthcheck.sh
@@ -3,12 +3,16 @@
 set -x
 set -e
 
-hostname=$(puppet config print certname) && \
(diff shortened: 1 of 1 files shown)


docker/puppetserver-standalone/healthcheck.sh
```diff
@@ -3,12 +3,16 @@
 set -x
 set -e
 
-hostname=$(puppet config print certname) && \
+certname=$(ls /etc/puppetlabs/puppet/ssl/certs | grep -v ca.pem) && \
```
Iristyle: Nit - can we change `-v` to `--invert-match` for readability?

Thanks!

underscorgan:
@Iristyle updated!

docker/puppetserver-standalone/healthcheck.sh
```diff
@@ -3,12 +3,16 @@
 set -x
 set -e
 
-hostname=$(puppet config print certname) && \
+certname=$(ls /etc/puppetlabs/puppet/ssl/certs | grep --invert-match ca.pem) && \
+hostname=${certname%.*} && \
```
Iristyle: This construct is not one I was familiar with either and had to look it up as well... 

> ${var%Pattern} Remove from $var the shortest part of $Pattern that matches the back end of $var.

But since we're matching everything with `.*`, I don't think I understand why the `%.*` is needed?

underscorgan: @Iristyle we're stripping off the file extension here, that `.` isn't a meta-character it's the literal `.`, so `puppet.test.pem` would go to `puppet.test`

underscorgan: ah, cool. I hadn't. That seems more straightforward, will switch to that

underscorgan:
@Iristyle updated!

PR #2073, opened 2019-05-31
(SERVER-2520) Allow setting log_level capturing compile logs

Magisus:
This commit adds the ability to specify the logging level when capturing
logs for catalog compile performed with the v4/catalog endpoint. Note
that if `debug` is used, the response might be very verbose and could
leak senstive information (similar to regular debug logging).


Magisus:
Need to test this.

pcarlisle:
It seems odd to warn that a setting could leak sensitive data when it's open to be set by the client. If this is a real risk there should be a way to lock it down on the server.

Magisus:
Are you talking about a way to configure the server not to allow `debug` to be requested, that someone could toggle if they were concerned?

pcarlisle:
Yeah something like that is what I was thinking. As it is now, you might as well reword the warning to "you might gain access to sensitive information by using this"

Magisus:
Oh. Duh, good point. Updated with setting.

I've also manually verified this now. Looking into writing a test now.

Magisus:
Is this worth testing at the integration level (like with the v4 catalog endpoint using it)? We didn't have any test around the log capture previously.

Magisus:
Is this worth testing at the integration level? We didn't have any testing around the log capture previously.

justinstoller:
Is gaining access to Sensitive data different then being able to read a catalog for another node?

Magisus:
Not sure, is Sensitive data in plain text in the serialized catalog?

justinstoller:
So, do we need a setting to allow sensitive data in the logs, when it will be in the catalog (the other key in the response hash)?

Magisus:
I don't... think so? Not sure what that setting would do, I think most of the sensitive stuff that shows up in logs does so by accident.

Magisus:
@pcarlisle given that the sensitive data is already in the clear in the catalog, do you think it's still useful to have the setting preventing use of `debug` (e.g. to prevent people from requesting extremely verbose logging), or is that just YAGNI?

pcarlisle:
I'm not sure. I don't know what else might end up in there. It's probably ok.

Magisus:
Updated to remove the `disallow-debug` setting.

src/ruby/puppetserver-lib/puppet/server/logging.rb
```diff
@@ -0,0 +1,26 @@
+require 'puppet/server/log_collector'
+
+module Puppet
+  module Server
+    module Logging
+
+      def self.capture_logs(level, &block)
+        old_log_level = Puppet.settings[:log_level]
+        Puppet.settings[:log_level] = level if level
```
justinstoller: What would an invocation of `Logging.capture_logs(false) do .. end` mean?

I would guess that `level` is more of a logging level override and might make more sense as an optional param if we want callers to use Puppet Server's default logging level.

src/ruby/puppetserver-lib/puppet/server/logging.rb
```diff
@@ -0,0 +1,26 @@
...
+
+        logs = []
+        result = nil
+        log_dest = Puppet::Server::LogCollector.new(logs)
+        Puppet::Util::Log.with_destination(log_dest) do
+          result = yield
+        end
+
+        log_entries = logs.map(&:to_data_hash)
```
justinstoller: With the increase in logging at debug level are we concerned about this additional mapping over the logs, or is that concern yagni at this point?

src/ruby/puppetserver-lib/puppet/server/logging.rb
Magisus: That makes sense.

Magisus: Supplying the `nil` here as a default (not overriding puppetserver's log level) is functionally what this does already, since if the option wasn't present in the request, nil will be passed. We don't have any places in code where `level` is not passed in. Did you have something different in mind than just `capture_logs(level = nil, &block)`?

src/ruby/puppetserver-lib/puppet/server/logging.rb
Magisus: Not sure if we really have a choice? The clojure needs to get a hash back to properly parse to JSON for the response. What were you thinking instead?

src/ruby/puppetserver-lib/puppet/server/logging.rb
justinstoller: not really, though I might change the name of the variable to something conveying that the level is an override of the existing process logging level. or a comment/doc to that effect with valid values.

src/ruby/puppetserver-lib/puppet/server/logging.rb
justinstoller: Updating the LogCollector's `<<` method to call `to_data_hash` on its input so we don't have to map over all the log events after they were collected.

Magisus: Oh yeah that's much better. Thanks.

Magisus:
Updated.

justinstoller:
Test failures seem legit.

Magisus:
That test now passes for me locally...

Magisus:
The tests are passing now.

PR #2074, opened 2019-05-31
(maint) Update travis to use xenial image

pcarlisle:
Trusty is going EOL and travis is migrating.


pcarlisle:
i'm going to test how this works on each branch in prs

PR #2076, opened 2019-06-04
(maint) Don't run container tests unless there are changes to containers

underscorgan:


Iristyle:
I think I may have missed the conversation on this one -- this works because we're not actually building anything that goes into the container from source?

underscorgan:
@Iristyle yeah, since we're just installing packages if nothing changed in the 'docker' directory nothing will have changed that affects the container tests. This is mostly to cut down on transient failures on PRs

PR #2078, opened 2019-06-05
Make ssl-port configurable in the puppetserver container

underscorgan:


docker/puppetserver-standalone/webserver.conf
```diff
@@ -0,0 +1,6 @@
+webserver: {
+    access-log-config: /etc/puppetlabs/puppetserver/request-logging.xml
+    client-auth: want
+    ssl-host: 0.0.0.0
+    ssl-port: ${PUPPET_MASTERPORT}
```
nwolfe: Does this ENV variable get resolved? It might; they don't in pe-console-services at least.

Also, I noticed this webserver.conf file is modified via `hocon` in the entrypoint script: https://github.com/puppetlabs/puppetserver/pull/2078/files#diff-f9e912c5472d58a4ee06a97dc9592516R31

Just pointing that out since I hadn't seen that before.

underscorgan: Ah, good call about using `hocon` to manage webserver.conf, I've updated!

underscorgan:
pupperware commit has been merged in, so this should be good to go!

nwolfe:
Oh, guess we should add the new configuration variables to the README too

underscorgan:
Yeah, I think leaving `EXPOSE` with 8140 makes sense since that will be accurate in most cases / that makes the normal case work as expected with compose (it seems to auto-expose ports for whatever the containers have `EXPOSE`d)

Added README updates

PR #2079, opened 2019-06-05
(DOCUMENT-852) Clarify which puppet.conf sections puppetserver uses

clairecadman:
https://tickets.puppetlabs.com/browse/DOCUMENT-852

### DIFF
--- a/documentation/configuration.markdown
+++ b/documentation/configuration.markdown
@@ -9,9 +9,7 @@ canonical: "/puppetserver/latest/configuration.html"
 [`puppetserver.conf`]: ./config_file_puppetserver.markdown
 [deprecated]: ./deprecated_features.markdown
 
-Puppet Server honors most settings in `puppet.conf` and picks them up automatically. However, for some tasks, such as configuring the web server or an external Certificate Authority (CA), Puppet Server has separate configuration files and settings.
-
-These files and settings are described below. For more information about differences between Puppet Server and the Ruby Puppet master's use of `puppet.conf` settings, see  [Puppet Server: Differing Behavior in `puppet.conf`](./puppet_conf_setting_diffs.markdown).
+Puppet Server automatically loads settings in the `main` and `master` section, and honors most other settings in [`puppet.conf`](https://github.com/puppetlabs/puppetserver/blob/master/src/clj/puppetlabs/services/config/puppet_server_config_core.clj#L18-L49). However, for some tasks, such as configuring the web server or an external Certificate Authority (CA), Puppet Server has separate configuration files and settings. These files and settings are described below. For more information about differences between Puppet Server and the Ruby Puppet master's use of `puppet.conf` settings, see  [Puppet Server: Differing Behavior in `puppet.conf`](./puppet_conf_setting_diffs.markdown).
 
 ## Configuration Files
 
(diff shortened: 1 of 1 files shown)


clairecadman:
@Magisus could you please review for accuracy? In response to DOCUMENT-852. 
(https://tickets.puppetlabs.com/browse/DOCUMENT-852)

documentation/configuration.markdown
```diff
@@ -9,9 +9,7 @@ canonical: "/puppetserver/latest/configuration.html"
...
-
-These files and settings are described below. For more information about differences between Puppet Server and the Ruby Puppet master's use of `puppet.conf` settings, see  [Puppet Server: Differing Behavior in `puppet.conf`](./puppet_conf_setting_diffs.markdown).
+Puppet Server automatically loads settings in the `main` and `master` section, and honors most other settings in [`puppet.conf`](https://github.com/puppetlabs/puppetserver/blob/master/src/clj/puppetlabs/services/config/puppet_server_config_core.clj#L18-L49). However, for some tasks, such as configuring the web server or an external Certificate Authority (CA), Puppet Server has separate configuration files and settings. These files and settings are described below. For more information about differences between Puppet Server and the Ruby Puppet master's use of `puppet.conf` settings, see  [Puppet Server: Differing Behavior in `puppet.conf`](./puppet_conf_setting_diffs.markdown).
```
justinstoller: Puppet Server will prefer values set in the master section to the same settings set in the main section, if there are duplicates.

eg, the config file:
```
[master]
setting1 = foo
setting2 = baz

[main]
setting1 = bar
setting3 = buzz
```

Will resolve `setting1` to `foo`, `setting2` to `baz`, `setting3` to `buzz`, and `setting4` will be whatever its default value is.

Puppet Server will only do the above (load and honor) those values listed in the linked document.

fwiw, I think it'd be better to write out those settings in the docs than link to the code. In code they start with colons and  use hyphens, however when writing the actual settings the colons are dropped and the hyphens are turned into underscores, eg the linked `:allow-duplicate-certs` becomes the setting `allow_duplicate_certs`

clairecadman:
@justinstoller I've updated the PR. Do you think this is better? I'm not sure whether this is the right place for the long list.

justinstoller:
I don't think the list is bad, though I see your concern about whether this is the best place for it. Overall, I think the wording and listing of these settings is a good improvement.

justinstoller:
I kinda inserted myself into this one, would love to know what @Magisus and @jcbollinger (Sorry if you're not the John I'm trying to ping) think of it.

jcbollinger:
If I'm following correctly, the approved text says that settings in the `[main]` section take precedence over those in the `[master]` section.  Is that really true?  It conflicts with [the general docs for puppet.conf](https://puppet.com/docs/puppet/6.4/config_file_main.html), which say

> `main` is the global section used by all commands and services. **It can be overridden by the other sections**.

(Emphasis added)

jcbollinger:
The question of which section takes precedence aside, the form of the documentation update is great -- just the sort of thing I was hoping for.

justinstoller:
You're right, John. I read that wrong. Thanks!

justinstoller:
Remembered I'm capable of making those changes as well  (:

PR #2080, opened 2019-06-11
(MAINT) Separate container build targets

nwolfe:
Just so it's easier to work with each image during container
development.

I was playing around with using puppetserver-standalone as a base image
for another image and needed to make changes to the Dockerfile, but I
didn't want to spend the time building both images each time.

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -15,7 +15,7 @@ dockerfile := Dockerfile
 PUPPERWARE_ANALYTICS_STREAM ?= dev
 
 prep:
-	@git fetch --unshallow ||:
+	@git fetch --unshallow 2> /dev/null ||:
 	@git fetch origin 'refs/tags/*:refs/tags/*'
 
 lint:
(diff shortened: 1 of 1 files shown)


docker/Makefile
```diff
@@ -50,12 +56,12 @@ build: prep
 		--tag $(NAMESPACE)/puppetserver:$(version) \
 		puppetserver
 ifeq ($(IS_LATEST),true)
-	@docker tag $(NAMESPACE)/puppetserver-standalone:$(version) \
-		$(NAMESPACE)/puppetserver-standalone:latest
 	@docker tag $(NAMESPACE)/puppetserver:$(version) \
 		$(NAMESPACE)/puppetserver:latest
 endif
 
+build: build-standalone build-full
```
nwolfe: FYI I verified that `build-standalone` does *not* get called twice during `make build`.

We could also just define this as `build: build-full` and then rely on `build-full: build-standalone` 🤷‍♂

nwolfe:
Not sure why this is constantly failing in one Travis cell but none of the others.

This change isn't worth the investigation and further potential changes, so just going to close it.

PR #2082, opened 2019-06-17
(SERVER-2540) Compile AST snippets

Magisus:
This adds the Ruby logic to compile catalogs from snippets of Puppet AST. It is built on https://github.com/puppetlabs/puppetserver/pull/2069 (logging refactor) and https://github.com/puppetlabs/puppetserver/pull/2069 (Clojure endpoint exposure).

See https://github.com/puppetlabs/bolt/blob/master/lib/bolt/catalog.rb#L63-L90 for the bolt version of this functionality.

### DIFF
--- a/spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
+++ b/spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
@@ -1,0 +1,64 @@
+require 'spec_helper'
+
+require 'puppet_pal'
+
+require 'puppet/server/ast_compiler'
+
+def generate_ast(code)
+  Puppet::Pal.in_tmp_environment("ast") do |pal|
+    pal.with_catalog_compiler do |compiler|
+      ast = compiler.parse_string(code)
+      Puppet::Pops::Serialization::ToDataConverter.convert(ast,
+                                                           rich_data: true,
+                                                           symbol_to_string: true)
+    end
+  end
+end
+
+def request(code)
+  {"certname" => "localhost",
+   "facts" => {"values" => {"my_fact" => "fact_value"}},
+   "trusted_facts" => {"values" => {"my_trusted" => "trusted_value"}},
+   "options" => {"capture_logs" => false},
+   "environment" => "production",
+   "code_ast" => generate_ast(code).to_json,
+   "variables" => {"values" => {"foo" => "bar"}}}
+end
+
+def find_notify(catalog)
+  catalog['resources'].find do |item|
+    item['type'] == 'Notify'
+  end
+end
+
+describe Puppet::Server::ASTCompiler do
+  context 'when compiling AST' do
+    it 'handles basic resources' do
+      response = Puppet::Server::ASTCompiler.compile(request("notify { 'my_notify': }"))
+      notify = find_notify(response[:catalog])
+      expect(notify).not_to be_nil
+      expect(notify['title']).to eq("my_notify")
+    end
+
+    it 'correctly interpolates supplied variables' do
+      response = Puppet::Server::ASTCompiler.compile(request('notify { "$foo": }'))
+      notify = find_notify(response[:catalog])
+      expect(notify).not_to be_nil
+      expect(notify['title']).to eq("bar")
+    end
+
+    it 'correctly interpolates supplied facts' do
+      response = Puppet::Server::ASTCompiler.compile(request('notify { "$my_fact": }'))
+      notify = find_notify(response[:catalog])
+      expect(notify).not_to be_nil
+      expect(notify['title']).to eq("fact_value")
+    end
+
+    it 'correctly interpolates supplied trusted facts' do
+      response = Puppet::Server::ASTCompiler.compile(request('notify { "${trusted[\'my_trusted\']}": }'))
+      notify = find_notify(response[:catalog])
+      expect(notify).not_to be_nil
+      expect(notify['title']).to eq("trusted_value")
+    end
+  end
+end
--- a/src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
@@ -1,0 +1,53 @@
+require 'puppet_pal'
+require 'puppet/server/log_collector'
+require 'puppet/server/logging'
+
+module Puppet
+  module Server
+    class ASTCompiler
+      def self.compile(compile_options)
+        options = compile_options['options'] || {}
+
+        log_level = options['log_level']
+        code = JSON.parse(compile_options['code_ast'])
+
+        if options['capture_logs']
+          catalog, logs = Logging.capture_logs(log_level) do
+            compile_ast(code, compile_options)
+          end
+
+          { catalog: catalog, logs: logs }
+        else
+          catalog = compile_ast(code, compile_options)
+          { catalog: catalog }
+        end
+      end
+
+      def self.compile_ast(code, compile_options)
+        Puppet[:node_name_value] = compile_options['certname']
+
+        # Use the existing environment with the requested name
+        Puppet::Pal.in_environment(compile_options['environment'],
+                                   envpath: Puppet[:environmentpath],
+                                   facts: compile_options['facts']['values'],
+                                   variables: compile_options['variables']['values']) do |pal|
+          Puppet.lookup(:pal_current_node).trusted_data = compile_options['trusted_facts']['values']
+
+          # This compiler has been configured with a node containing
+          # the requested environment, facts, and variables, and is used
+          # to compile a catalog in that context from the supplied AST.
+          pal.with_catalog_compiler do |compiler|
+            # We have to parse the AST inside the compiler block, because it
+            # initializes the necessary type loaders for us.
+            ast = Puppet::Pops::Serialization::FromDataConverter.convert(code)
+
+            compiler.evaluate(ast)
+            compiler.compile_additions
+            compiler.catalog_data_hash
+          end
+        end
+      end
+      private_class_method :compile_ast
+    end
+  end
+end
--- a/documentation/puppet-api/v3/compile.markdown
+++ b/documentation/puppet-api/v3/compile.markdown
@@ -36,7 +36,7 @@ The request body must look like:
   "transaction_uuid": "<uuid string>",
   "job_id": "<id string>",
   "options": { "capture_logs": <true/false>,
-               "log_level": <one of debug/info/warn>}
+               "log_level": <one of debug/info/warning/err>}
 }
 ```
 
--- a/src/ruby/puppetserver-lib/puppet/server/compiler.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/compiler.rb
@@ -73,10 +73,6 @@ module Puppet
           node.environment = requested_environment
         end
 
-        # Merges facts into the node parameters.
-        # Ensures that facts will be surfaced as top-scope variables,
-        # along with other node parameters.
-        node.merge(facts.values)
         node.trusted_data = trusted_facts
         node.add_server_facts(@server_facts)
         node
--- a/src/ruby/puppetserver-lib/puppet/server/master.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/master.rb
@@ -10,6 +10,7 @@ require 'puppet/server/config'
 require 'puppet/server/puppet_config'
 require 'puppet/server/network/http/handler'
 require 'puppet/server/compiler'
+require 'puppet/server/ast_compiler'
 
 require 'java'
 
@@ -69,7 +70,7 @@ class Puppet::Server::Master
   end
 
   def compileAST(compile_options)
-    {mock: "data"}
+    Puppet::Server::ASTCompiler.compile(convert_java_args_to_ruby(compile_options))
   end
 
   def getClassInfoForEnvironment(env)


Magisus:
I haven't been able to get this to work manually yet, still tracking down if that's an issue with my test input or the code calls. Some review is still appreciated; maybe an issue with the code will jump out at someone.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -0,0 +1,44 @@
...
+          catalog = compile_ast(compile_options)
+          { catalog: catalog }
+        end
+      end
+
+      def self.compile_ast(compile_options)
+        code = compile_options['code_ast']
+        target = compile_options['certname']
+
+        Puppet::Pal.in_tmp_environment(compile_options['environment'],
```
justinstoller: Do we have to use Pal?

Magisus: I don't know that we _have_ to. I was going to try to get it working using PAL first, since we have an example of that flow, and when we looked over this code, I don't think anyone had a good reason _not_ to. Its API is definitely better defined for what we're trying to do than the other entry points.

Magisus:
Fixed the issue, turns out the AST needs to be parsed into a Hash before you attempt to convert it to rich data and compile it. It silently does something else not mentioned in the docs comment if passed a string.

I know Ruby doesn't really want to enforce types, but things still work out better if methods don't do completely different things based on what type they're passed :( There are multiple layers of that going on here.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
Magisus: That being said, now that this is working, I'll start looking at alternatives, if we think this is gross or unnecessary. Though I think we might wind up with more complicated code here.

Do you have some particular things in mind you'd like to avoid? I've cut out almost all of the settings configuration already, compared to Bolt.

justinstoller: I just don't know what this does or is meant to do.

Magisus:
So, I took inventory of everything that PAL is doing for us. A lot of it is really similar to what we are doing in our other `Compiler` class, though it is more tailored for a situation where the node isn’t a real node that is classified or anything (e.g. it doesn't do the complicated environment resolution). 

However, it does handle custom variables and their validation, which we don't support in the other context.

I didn’t come across any indirections; it just creates objects directly like we were originally trying to do, and can get away with that because it doesn't attempt to load or store anything from PDB or the classifier.

I looked at what we're doing in our other `Compiler` class, but the use case there is heavily optimized for logic around the side effects before or after compile, most of which is irrelevant here. I'm not sure it's worth the complexity of adding another axis of configurability -- in where the code to be compiled comes from (which takes us down completely different code paths) --  to reuse parts of that.

Personally, I’m kind of in favor of continuing to use PAL here. There is a little bit of extraneous logic for other use cases, but I don't see a lot of value in pulling out yet another version of this sequence of setting up a compile environment, given that this already does exactly what we need if used correctly. The resulting code is essentially 20 lines, because the API is designed for this use case. But if someone feels differently, I'd love to hear other ideas.

I've tested what's in this PR with code that relies on installed modules, and with custom variables. But I'm interested in ideas of other cases we should try.

Magisus:
Tested that this can correctly use `puppetdb_query` function with no additional work, when Puppet Server is configured with PuppetDB.

Magisus:
I have also confirmed that PAL respects both `basemodulepath` and `environment.conf` out of the box, when using `in_environment`.

Magisus:
This is ready for review. I have one other manual test case I'd like to do, making sure you can do hiera lookups.

Magisus:
Just kidding, I think I found some weird state leakage. Once you do a compile with this code, all other normal catalog compiles don't work until you restart the server. Digging into that.

Magisus:
For posterity: I've been testing this by generating JSON AST using `Bolt::Catalog.new.generate_ast(<code string>).to_json`, writing it out to a file, escaping the quotes (be careful with quoted strings inside strings in the resulting JSON, they need to be double escaped), and using the following request to curl the server:
```
ast=$(cat ast.json)

curl -k -H "Content-Type: application/json" --request POST --data "{\"certname\":\"localhost\",\"facts\":{\"values\":{\"my_fact\":\"fact_value\"}},\"trusted_facts\":{\"values\":{\"my_trusted\":\"trusted_value\"}}, \"options\":{\"capture_logs\":true,\"log_level\":\"debug\"},\"environment\":\"production\",\"code_ast\":\"${ast}\",\"variables\":{\"values\":{\"foo\":\"bar\"}}}" https://localhost:8140/puppet/v3/compile
```

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -0,0 +1,51 @@
...
+                                   envpath: Puppet[:environmentpath],
+                                   facts: compile_options['facts']['values'],
+                                   variables: compile_options['variables']['values']) do |pal|
+          Puppet.lookup(:pal_current_node).trusted_data = compile_options['trusted_facts']['values']
+
+          # This compiler has been configured with a node containing
+          # the requested environment, facts, and variables, and is used
+          # to compile a catalog in that context from the supplied AST.
+          pal.with_catalog_compiler do |compiler|
+            ast = Puppet::Pops::Serialization::FromDataConverter.convert(code)
```
justinstoller: Does the deserialization need to happen within this block?

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -0,0 +1,51 @@
...
+          { catalog: catalog, logs: logs }
+        else
+          catalog = compile_ast(compile_options)
+          { catalog: catalog }
+        end
+      end
+
+      def self.compile_ast(compile_options)
+        # FromDataConverter.convert below expects this to be a hash
+        code = JSON.parse(compile_options['code_ast'])
```
justinstoller: It seems weird that fact values have been recursively deserialized but the ast hasn't at this point. Should the code higher up be doing some more aggressive deserializing?

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -0,0 +1,51 @@
...
+          Puppet.lookup(:pal_current_node).trusted_data = compile_options['trusted_facts']['values']
+
+          # This compiler has been configured with a node containing
+          # the requested environment, facts, and variables, and is used
+          # to compile a catalog in that context from the supplied AST.
+          pal.with_catalog_compiler do |compiler|
+            ast = Puppet::Pops::Serialization::FromDataConverter.convert(code)
+            compiler.evaluate(ast)
+            compiler.compile_additions
+            compiler.with_json_encoding(&:encode)
```
justinstoller: this feels like a weird interface, but I guess it has nothing to do with this pr

spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    pal.with_catalog_compiler do |compiler|
+      ast = compiler.parse_string(code)
+      Puppet::Pops::Serialization::ToDataConverter.convert(ast,
+                                                           rich_data: true,
+                                                           symbol_to_string: true)
+    end
+  end
+end
+
+def request(ast)
```
justinstoller: Do we ever call `generate_ast` w/o immediately calling `request` on the result?

spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      req = request(ast_facts)
+      response = Puppet::Server::ASTCompiler.compile(req)
+      expect(response[:catalog]).to match(/Notify\[fact_value\]/)
+    end
+
+    it 'correctly interpolates supplied trusted facts' do
+      ast_trusted_facts = generate_ast('notify { "${trusted[\'my_trusted\']}": }')
+      req = request(ast_trusted_facts)
+      response = Puppet::Server::ASTCompiler.compile(req)
+      expect(response[:catalog]).to match(/Notify\[trusted_value\]/)
```
justinstoller: Does the `compile` method return a Hash with a `:catalog` key that is JSON string? Can we just return a Hash and do serialization all at once?

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
Magisus: Yeah that seems fine. This is a relic from when I thought the processing methods below were expecting a string.

spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
Magisus: No, and probably don't have to. I can combine this.

spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
Magisus: I don't think so, because of the interface of the PAL catalog compiler. The only way to get the catalog out is via the `with_json_encoding` method. The other methods don't actually return the catalog. So we'd have to deserialize that only to re-serialize it later.

Magisus: Do you think it makes a difference? We could probably add something to the API if we really think it's better to do it all at once.

justinstoller: it feels like a weird way to shape the response internally and the reason seems to be because the thing we're calling into has a weird interface. do you know why the interface to compiling in pal is that way?

aside, in our other endpoint we get a data structure that represents the catalog (from calling `to_data_hash` on a Catalog instance) and then serialize the whole thing at once. that feels more natural but if there's a good reason that pal needs to do the serializing then maybe there's a latent bug in the other endpoint?

Magisus: So, there is some post-processing that PAL does as part of the JSON encoding. It gives you the option to exclude virtual resources (this is on by default) and will clean up any resources that are missing names or titles. 

That being said, by simply making the [`catalog` method](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/pal/catalog_compiler.rb#L12-L16) public, we have access to the same catalog object that we turn to a hash in the other endpoint. I dug through the history, and that method has been private since the beginning, with no comment as to why. But I don't think it would be too bad to expose it... This seems like it was maybe just designed for another use case (a human use case seems probable, given that pretty printing is also on by default).

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
Magisus: So apparently something gets loaded as part of setting up these blocks that allows the `FromDataConverter` to return a `Puppet::Pops::Model::Program`, which lets us follow the right branch [here](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/pal/catalog_compiler.rb#L48). Doing it outside the block returns what looks like an anonymous class. I'll try to track down what needs to be loaded so we could do it outside, if that seems more appropriate.

justinstoller: I don't know that it needs to be deserialized outside of the block, though I find that behavior surprising. It would be nice to somehow make the dependency on where it's executed explicit if not remove it. What do you think?

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -0,0 +1,50 @@
...
+          Puppet.lookup(:pal_current_node).trusted_data = compile_options['trusted_facts']['values']
+
+          # This compiler has been configured with a node containing
+          # the requested environment, facts, and variables, and is used
+          # to compile a catalog in that context from the supplied AST.
+          pal.with_catalog_compiler do |compiler|
+            ast = Puppet::Pops::Serialization::FromDataConverter.convert(code)
+            compiler.evaluate(ast)
+            compiler.compile_additions
+            compiler.catalog.to_data_hash
```
Magisus: This won't work without an update to PAL making the `catalog` method public.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
Magisus: Yeah I do think that would be better, unless the required loading proves complicated. I get the impression some types just need to be brought into scope.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
Magisus: See https://github.com/puppetlabs/puppet/pull/7599 for that and other updates

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
Magisus: Turns out that it requires the Loaders to be initialized with the environment and then added to the context, where they are used to load various PCore types. Without this, a generic class is generated instead, in place of the official PCore type. I think it's best to leave that initialization up to PAL, and therefore do the AST conversion inside this block.

Magisus:
This is ready for review. I'll squash my feedback commit once we're happy with it.

spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
hlindberg: As mentioned in the other PR where an API was added - the reason is that we don't want all of the `Catalog` implementation to be considered API - and it should instead be available via a wrapper that provides the parts we do want as API. In the other PR, a method is added to serialize the catalog as a data_hash - that is fine as API.

Magisus: I'll update this now that the other PR is merged and the API finalized.

Magisus:
This is now using the `catalog_data_hash` API as added in https://github.com/puppetlabs/puppet/pull/7599.

Magisus:
This has now been rebased on top of the Puppet version providing the new API.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -0,0 +1,53 @@
...
+
+          { catalog: catalog, logs: logs }
+        else
+          catalog = compile_ast(code, compile_options)
+          { catalog: catalog }
+        end
+      end
+
+      def self.compile_ast(code, compile_options)
+        target = compile_options['certname']
```
tvpartytonight: It looks like `target` isn't used anywhere else in this method? 🎯

Magisus: True... looks like Bolt might set `Puppet[:node_name_value]` to the cert name, I didn't seem to need that, but maybe you do in some cases...

Magisus: Updated to do what Bolt does. PAL uses it notably here: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/pal/pal_impl.rb#L320 so setting that seems correct.

PR #2083, opened 2019-06-19
(SERVER-1914) Update clj-parent to version with FIPS bouncy-castle

Magisus:


justinstoller:
The code tests have all passed though the docker test has failed (from what the puppetware team tells me) spuriously twice. Regardless of its outcome I think this is good to merge.

PR #2085, opened 2019-06-19
(maint) Only run docker travis tests on master

justinstoller:


justinstoller:
Yeah, they're only doing master, and this test is causing trouble in your other PR.

nwolfe:
Heads up looks like this is also in `6.3.x`: https://github.com/puppetlabs/puppetserver/blob/6.3.x/.travis.yml#L15-L18

Not sure if it should be there or not, but it's out-of-date with what we're doing on `master` anyway, so it would probably fail when run.

PR #2086, opened 2019-06-24
Mergeup60to63

tvpartytonight:


tvpartytonight:
I was pretty sure that we wanted to keep openjdk8 for now...but I wasn't 100% sure...

justinstoller:
yeah, we want jdk8, though I'm not sure how we started pulling in a library that doesn't work on jdk11?

PR #2087, opened 2019-06-24
Build puppetserver snapshot packages for use in docker images

underscorgan:
The docker image build has been updated to start with an ezbake
local-build so that we're running puppetserver 'from source' in the
containers.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,8 @@
+/lib/
+/classes/
+/targets/
+target/
+/.bundle/
+/resources/puppetlabs/puppetserver/*.class
+/tmp/
+/dev-resources/i18n/bin
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,8 +1,9 @@
-/lib/
-/classes/
-/targets/
-target/
-/.bundle/
-/resources/puppetlabs/puppetserver/*.class
-/tmp/
-/dev-resources/i18n/bin
+*
+!.git
+!project.clj
+!docker
+!documentation
+!ezbake
+!resources
+!ruby
+!src
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,6 @@
 docker/puppetserver/*               text  eol=lf
 docker/puppetserver-standalone/*    text  eol=lf
 docker/puppetserver-standalone/docker-entrypoint.d/*    text  eol=lf
+resources/ext/build-scripts/*       text  eol=lf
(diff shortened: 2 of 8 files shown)


justinstoller:
what's the rationale for the :internal profile change?

underscorgan:
@justinstoller that's intended to make it easier for external folks to re-build our software, but I should break that out into a separate PR. I've been looking at a bunch of dev/contributor improvements and this ended up blending in with some wider work to make it easier to build ezbake packages externally.

justinstoller:
That'd be awesome, I ended up here from the cjc pr, and I'm not 100% on how its intended to be used (but clarifying that is my only qualm about that and this pr).

underscorgan:
@justinstoller split out the profile change to https://github.com/puppetlabs/puppetserver/pull/2089

.dockerignore
```diff
@@ -0,0 +1,8 @@
+/lib/
+/classes/
+/targets/
+target/
+/.bundle/
+/resources/puppetlabs/puppetserver/*.class
+/tmp/
+/dev-resources/i18n/bin
```
nwolfe: Not sure in this case, but it may be easier to invert these .dockerignore files so we ignore everything and then only add in the things we need, like we did for puppetdb: https://github.com/puppetlabs/puppetdb/blob/master/.dockerignore

That way you don't accidentally get a Gemfile.lock sneaking into your docker build context and changing your behavior, or some other random file that gets added to the repo but not added to .dockerignore.

Not a big deal, just mentioning the different approaches.

underscorgan: ahh, good call @nwolfe, will update

docker/minimal-spec/puppetserver-minimal_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    run_command('docker pull puppet/puppet-agent-ubuntu:latest')
+    if ENV['PUPPET_TEST_DOCKER_IMAGE'].nil?
+      fail <<-MSG
+      error_message = <<-MSG
+  * * * * *
+  PUPPET_TEST_DOCKER_IMAGE environment variable must be set so we
+  know which image to test against!
+  * * * * *
+      MSG
+		end
```
Iristyle: super minor nit... looks we have some tabs interspersed with spaces on a few lines here

Iristyle:
Even though this went red in Azure, merging since we know that's a transient we're working on elsewhere.

PR #2088, opened 2019-06-25
(SERVER-2576)

justinstoller:
Here's an example of the changes needed to support tk-jetty9 3.0.

They are essentially just configuring the default cipher-suites to include the older weaker ciphers until we can drop them in an X.

However, I think removal of the `so-linger-seconds` config option might be a bigger deal than I originally thought. If a user has it set the service will not start and there isn't a place where we first get a chance to munge the webserver's config. You can see this by checking out this branch and then running `lein repl` & `(go)`.

We may want to go and do a 3.0.1 release the allows the option even though it is ignored?


justinstoller:
Here's the change I'm proposing to tk-jetty9
https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/pull/206

Magisus:
Were we intending to update the jetty version in clj-parent or not yet? Current is 2.4.1.

Magisus:
I've released 3.0.1 which contains the schema updates for backwards compatibility.

PR #2090, opened 2019-06-27
(SERVER-2465) Allow ca certs and crl bundles to be in any order

pcarlisle:
Use the ca key to detect the correct certificate to use from a
certificate bundle on the ca, and the ca cert to detect the correct crl
from a crl bundle.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -36,7 +36,7 @@
 
   :min-lein-version "2.7.1"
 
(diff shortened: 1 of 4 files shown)


pcarlisle:
I did some manual testing with this and the ca cli and it seems ok but I'm 100% sure I know all the ins and outs of that.

PR #2092, opened 2019-07-02
(maint) Specify jvm opt allowing pry to work with Jruby

Magisus:
This commit adds a jvm option to the lein dev profile that enables the
use of pry and pry-nav.

This PR is testing if this setting negatively impacts our testing performance.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -142,7 +142,8 @@
                                    [com.cemerick/url "0.1.1"]]
 
                    ;; SERVER-332, enable SSLv3 for unit tests that exercise SSLv3
-                   :jvm-opts      ["-Djava.security.properties=./dev-resources/java.security"]}
+                   :jvm-opts      ["-Djava.security.properties=./dev-resources/java.security"
+                                   "-Djruby.debug.fullTrace=true"]}
 
              :testutils {:source-paths ["test/unit" "test/integration"]}
              :test {


justinstoller:
This took 34m and 19m on java 8 and 11, respectively. While compared to https://travis-ci.org/puppetlabs/puppetserver/builds/553042306 the old times seem to have been 30m and 16m each. We may want to wait and collect a couple more runs to make sure it's not just an infra blip.

justinstoller:
Looking at a few other runs it seems that this reliably increases the test time 15-20%.

If that's the case I don't really want to add this to the dev profile (or I'd like to override it in the test profile, or add it to a repl profile, or create a dedicated profile and have folks that want to do pry-nav stuff start the repl with an invocation like `lein with-profile +pry-nav irb`).

Magisus:
Agreed, we don't need any increases to our test time. I like the idea of having a dedicated profile. It's not always that you need to be debugging the Ruby code. Closing, and I'll look into doing that.

pcarlisle:
i think in the end anything with a dedicated profile will be no easier to invoke than just adding `JRUBY_OPTS=--debug` at the beginning of your command line

Magisus:
Oh, good to know you can do that. I'll just do that. Maybe adding something about it in the readme or even [here](https://github.com/puppetlabs/puppetserver/blob/master/documentation/dev_debugging.markdown#pry) would be good.

PR #2093, opened 2019-07-02
(MAINT) Use build.ps1 from pupperware gem

nwolfe:
Instead of copy/pasting it everywhere.


azure-pipelines.yml
```diff
@@ -11,73 +11,41 @@ steps:
...
-    ruby --version
-    gem --version
-    gem env
-    bundle --version
-    #
-    # Environment
-    #
-    Write-Host "`n`n$line`nEnvironment`n$line`n"
-    Get-ChildItem Env: | % { Write-Host "$($_.Key): $($_.Value)"  }
+    $gemfile = Join-Path -Path (Get-Location) -ChildPath 'docker/Gemfile'
```
Iristyle: Usually something like `Get-Location` is unreliable. which is why I have that `Get-CurrentDirectory` helper. I also think you can just use `.` there, like:

```powershell
$gemfile = Join-Path -Path '.' -ChildPath 'docker/Gemfile'
```

azure-pipelines.yml
```diff
@@ -11,73 +11,41 @@ steps:
...
+    $gempath = Join-Path -Path (Get-Location) -ChildPath 'docker/.bundle/gems'
+    bundle config --local gemfile $gemfile
+    bundle config --local path $gempath
+    bundle install
+  displayName: Fetch Dependencies
+  name: fetch_deps
+- powershell: |
+    . "$(bundle show pupperware)/ci/build.ps1"
```
Iristyle: Did you see if loading this once makes it available in all future `powershell` instances?

azure-pipelines.yml
nwolfe: I didn't want to copy/paste that custom `Get-CurrentDirectory` function into every repo, and it seemed like I should be able to use the PowerShell built-in function `Get-Location`.

I also tried `$PWD` and that worked; I feel like I had a problem trying to use `.`.

azure-pipelines.yml
nwolfe: I tried, didn't work unfortunately :(

azure-pipelines.yml
nwolfe: Happy to rewrite these, I just don't want to use a copy/pasted custom function if we can avoid it.

azure-pipelines.yml
Iristyle: Ok, wasn't sure if the instances were isolated or not... looks like they are!

nwolfe:
The plan with this is to wait for https://github.com/puppetlabs/puppetserver/pull/2087 to be merged in first, and then to incorporate changes made there into our shared `build.ps1` file, then update this PR accordingly.

PR #2099, opened 2019-07-04
(SERVER-2575) Use clj-shell-utils clojure lib

tvpartytonight:
Now that the shell-utils have been extracted and released, remove the
corresponding files from the puppetserver repo and utilize the new
clj-shell-utils lib.


justinstoller:
It looks like the version of shell-utils released was compiled on Java 11? We need to compile on Java 8 for all our FOSS releases.

tvpartytonight:
Now that [pepse has been upgraded to use clj-parent 3.1.1](https://github.com/puppetlabs/pe-puppet-server-extensions/commit/85cae614177a420b03e78790ea9fefc49dd5ee13), this should be good to merge.

PR #2100, opened 2019-07-05
(maint) Extend `server` config changes to upgrades

justinstoller:
Uses code from 0b26a8639 to finish work started (and described more
fully in) a116edeb8.


justinstoller:
Testing locally I've seen this pass the server `upgrade_tests` and Puppet's acceptance tests through `puppet/acceptance/tests/environment/enc_nonexistent_directory_environment.rb`. Don't know if I'll have batter to finish the suite here, but I think that's enough to validate that this is generally working correctly (that and it's the same solution to what is working in non-upgrade tests).

Magisus:
Agreed, going to just merge. With how long the tests take to run, I'm wondering what it would take to get PR beaker testing working for this repo, similar to what we do for Puppet. That would allow us to run things without it having to sit on someone's laptop. I'll file a ticket.

PR #2101, opened 2019-07-08
(maint) Don't override leiningen's default repositories by default

underscorgan:
With the current repository setup in `project.clj`, `:repositories` and
`:deploy-repositories` are set to point to internal-to-puppet
infrastructure. While builds still succeed with these settings on
external networks, there is a noticible slowdown while waiting for
network timeouts.

This commit also adds an `internal` profile that can be used for builds
running on Puppet's internal network to access local repos/mirrors. To
use this new profile, run `lein with-profile ezbake,internal ezbake
build` in instances where you are currently running `lein with-profile
ezbake ezbake build`.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -97,9 +97,6 @@
   :test-paths ["test/unit" "test/integration"]
   :resource-paths ["resources" "src/ruby"]
 
-  :repositories [["releases" "https://artifactory.delivery.puppetlabs.net/artifactory/clojure-releases__local/"]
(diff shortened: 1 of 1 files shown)


Magisus:
This seems to still be failing in Jenkins with
```
Failed to deploy artifacts/metadata: No connector available to access repository snapshots (snapshots) of type default using the available factories FileRepositoryConnectorFactory, WagonRepositoryConnectorFactory
```

Going to revert it till we figure out what's going on, since there's a release about to happen.

PR #2105, opened 2019-07-08
(MAINT) Prevent Azure double builds

nwolfe:


justinstoller:
Looks like yaml! But travis is angry?

PR #2109, opened 2019-07-09
(maint) sed zypper list when installing puppet-agent 1.x

justinstoller:
When moving to our archives the puppet-agent 1.x series (which contains
Puppet 4.x) had its zypper repo list corrupted. The invalid repo list
contains an extra "=" in it. See
https://tickets.puppetlabs.com/browse/RE-12490 for more info.

This works around it by using sed to remove the extraneous "="s.

### DIFF
--- a/acceptance/suites/pre_suite/upgrade/10_install_puppet.rb
+++ b/acceptance/suites/pre_suite/upgrade/10_install_puppet.rb
@@ -23,6 +23,8 @@ def install_pc1_repo(host)
   end
 
   if variant =~ /sles/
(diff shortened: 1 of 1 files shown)


Magisus:
Sad day. Hopefully we can revert this when the ticket gets fixed, but it shouldn't cause harm if we forget...

Can you throw a comment with the ticket number by that line?

justinstoller:
sure, also running the tests locally, right now

justinstoller:
updated. tests pass locally (running just the upgrade_tests - not the full suite).

PR #2121, opened 2019-07-17
(MAINT) Support custom entrypoint scripts

nwolfe:
Give users a directory they can fill with .sh scripts to be ran as part
of the entrypoint of the Docker image.

This alleviates the user from having to build a custom image that has
their configuration in it.

This should handle the cases when:
- Directory doesn't exist
- Directory exists but is empty
- Directory exists but has no .sh files in it
- Directory exists and has files with various extensions


nwolfe:
@puppetlabs/pupperware Any objections to this directory name `/docker-custom-entrypoint.d/` should be made now before we copy this pattern into our other images!

docker/puppetserver-standalone/docker-entrypoint.sh
```diff
@@ -10,4 +10,13 @@ for f in /docker-entrypoint.d/*.sh; do
     "$f"
 done
 
+if [ -d /docker-custom-entrypoint.d/ ]; then
+    chmod +x /docker-custom-entrypoint.d/*.sh
+    sync
+    for f in /docker-custom-entrypoint.d/*.sh; do
+        echo "Running $f"
+        "$f"
+    done
```
underscorgan: Does this work if the directory exists but is empty? should we guard against that?

nwolfe: It **does** error when the directory exists but is empty; good catch...I assumed it iterated 0 times.

Do we know of a good way to do that?

underscorgan: Looks like when I was playing around with this a while back I did 
```
if [[ -n "$(ls -A /docker-custom-entrypoint.d)" ]]; then
...
fi
```

nwolfe: Ok, there's a few ways to do this I see.

For reference here's what postgres does: https://github.com/docker-library/postgres/blob/87b15b6c65ba985ac958e7b35ba787422113066e/9.6/docker-entrypoint.sh#L145-L163

I like generalizing the pattern-match and then adding a `case` statement, because it means we don't need to check if the directory exists beforehand, we don't need subshells, and it's clear/easy to extend support for other file types in the future (and it doesn't require `bash`, which is a nice bonus).

So this:
```
for f in /docker-custom-entrypoint.d/; do
  case "$f" in
    *.sh) echo "Running $f"; "$f" ;;
    *) ;;
  esac
done
```

nwolfe: Updated with a `find`-based solution because there's a surprising number of ways this can fail.

I also wanted to avoid doing a `sync` in a loop, but I opted to put the `chmod` in a loop so we can gracefully handle non-`.sh` files in the directory without having to silence errors from `chmod +x /docker-custom-entrypoint.d/*.sh`.

docker/puppetserver-standalone/docker-entrypoint.sh
```diff
@@ -10,4 +10,12 @@ for f in /docker-entrypoint.d/*.sh; do
     "$f"
 done
 
+if [ -d /docker-custom-entrypoint.d/ ]; then
+    find /docker-custom-entrypoint.d/ -type f -name "*.sh" \
+        -exec chmod +x {} \;
```
nwolfe: Also tried just `chmod +x /docker-custom-entrypoint.d/*` but that too fails when the directory is empty.

Iristyle:
FWIW - Postgres does https://github.com/docker-library/postgres/blob/master/docker-entrypoint.sh#L145-L163

```shell
for f in /docker-entrypoint-initdb.d/*; do
  case "$f" in
    *.sh)
      # https://github.com/docker-library/postgres/issues/450#issuecomment-393167936
      # https://github.com/docker-library/postgres/pull/452
      if [ -x "$f" ]; then
        echo "$0: running $f"
        "$f"
      else
        echo "$0: sourcing $f"
        . "$f"
      fi
      ;;
    *.sql)    echo "$0: running $f"; "${psql[@]}" -f "$f"; echo ;;
    *.sql.gz) echo "$0: running $f"; gunzip -c "$f" | "${psql[@]}"; echo ;;
    *)        echo "$0: ignoring $f" ;;
  esac
  echo
done
```

nwolfe:
@Iristyle Yep: https://github.com/puppetlabs/puppetserver/pull/2121#discussion_r305063109

I'd like to follow the same pattern, but we have a call to `sync` thrown in the middle of all this, and we don't want to do that in loop.  Also postgres assumes the files are `chmod`'d or not already.

PR #2122, opened 2019-07-17
(SERVER-2555) Make test suite run in FIPS mode

Magisus:
Tests that involve installing puppet modules will not work in FIPS mode. This restriction also precludes installing PuppetDB. This commit skips tests that rely on modules or PDB being installed.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,7 +16,7 @@ gem 'jira-ruby', :group => :development
 
 group :test do
   gem 'rspec'
-  gem 'beaker', *location_for(ENV['BEAKER_VERSION'] || '~> 4.1')
+  gem 'beaker', *location_for(ENV['BEAKER_VERSION'] || '~> 4.11')
   gem "beaker-hostgenerator", *location_for(ENV['BEAKER_HOSTGENERATOR_VERSION'] || "~> 1.1")
   gem "beaker-abs", *location_for(ENV['BEAKER_ABS_VERSION'] || "~> 0.1")
   gem "beaker-vmpooler", *location_for(ENV['BEAKER_VMPOOLER_VERSION'] || "~> 1.3")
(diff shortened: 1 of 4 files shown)


Magisus:
I also have this PR against Puppet, which fixes up one test there that uses a module: https://github.com/puppetlabs/puppet/pull/7616

pcarlisle:
If we know this needs to be fixed for fips compliance where are we tracking to make sure we don't just leave these forever?

Magisus:
This also requires https://github.com/puppetlabs/beaker/pull/1592

Magisus:
I'll file something to remove these and block it on the ticket about getting `puppet module` working.

pcarlisle:
second commit says 'directoy'

pcarlisle:
This should be do not merge until the beaker pr gets into a release then?

Magisus:
Correct. I'll see what I can do to move that along.

Magisus:
Filed https://tickets.puppetlabs.com/browse/SERVER-2596 for re-enabling the `puppet module` tests.

Magisus:
The new version of beaker has been released: https://github.com/puppetlabs/beaker/releases/tag/4.11.0

Magisus:
A [PR](https://github.com/puppetlabs/puppet/pull/7624) just got merged to fix the PMT in FIPS mode, I want to retest this once that goes in, and then maybe we don't even need it.

Magisus:
Updated this to now only skip installing PuppetDB, which is still broken. I've filed a ticket for that [here](https://tickets.puppetlabs.com/browse/PDB-4457). The Puppet Module Tool works again on FIPS mode, so the other test that relied on installing a module now passes.

justinstoller:
Does this mean we should bump the minimum version of Beaker?

Magisus:
Probably not a bad idea.

PR #2125, opened 2019-07-19
SERVER-2597: Set HTTP basic auth on GET in HttpClient

ekohl:
By moving the credential code handling from `post()` to `create_common_request_options()` the HttpClient will handle HTTP basic auth on `get()` as well.

Any code that already uses `HttpClient` (such as Hiera lookup functions) and manually sets Authorization as a workaround, would start to fail. To handle this, the header values are compared and an exception is only raised when the values differ.

### DIFF
--- a/src/ruby/puppetserver-lib/puppet/server/http_client.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/http_client.rb
@@ -54,19 +54,6 @@ class Puppet::Server::HttpClient
   end
 
   def post(url, body, headers = {}, options = {})
-    # If credentials were supplied for HTTP basic auth, add them into the headers.
-    # This is based on the code in lib/puppet/reports/http.rb.
-    credentials = options[:basic_auth]
-    if credentials
-      if headers["Authorization"]
-        raise "Existing 'Authorization' header conflicts with supplied HTTP basic auth credentials."
-      end
-
-      # http://en.wikipedia.org/wiki/Basic_access_authentication#Client_side
-      encoded = Base64.strict_encode64("#{credentials[:user]}:#{credentials[:password]}")
-      headers["Authorization"] = "Basic #{encoded}"
-    end
-
     request_options = create_common_request_options(url, headers, options)
     request_options.set_body(body)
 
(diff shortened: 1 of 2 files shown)


ekohl:
Realized I used the wrong ticket number. Updated to use https://tickets.puppetlabs.com/browse/SERVER-2597

justinstoller:
Thanks, Ewoud, and sorry about the silence last week. We'll get this reviewed today.

ekohl:
FYI: how I found this is https://github.com/theforeman/puppet-foreman/blob/df7682be74087d251c018106f7185e401123d06a/lib/puppet/functions/foreman/enc.rb#L34-L54

src/ruby/puppetserver-lib/puppet/server/http_client.rb
```diff
@@ -92,6 +79,21 @@ def get(url, headers={}, options={})
   end
 
   def create_common_request_options(url, headers, options)
+    # If credentials were supplied for HTTP basic auth, add them into the headers.
+    # This is based on the code in lib/puppet/reports/http.rb.
```
justinstoller: It looks like this comment was never very correct and is very incorrect at this point. Can you remove it while you're here?

src/ruby/puppetserver-lib/puppet/server/http_client.rb
```diff
@@ -92,6 +79,21 @@ def get(url, headers={}, options={})
...
+    credentials = options[:basic_auth]
+    if credentials
+      # http://en.wikipedia.org/wiki/Basic_access_authentication#Client_side
+      encoded = Base64.strict_encode64("#{credentials[:user]}:#{credentials[:password]}")
+      authorization = "Basic #{encoded}"
+
+      if headers["Authorization"] && headers["Authorization"] != authorization
```
justinstoller: Seems more correct than what we were doing before

justinstoller:
This looks good!

Are you up for adding some a test or two at `spec/puppet-server-lib/puppet/jvm/http_client_spec.rb`? It'd be nice to have one for the change, but I can probably back fill that if you're not up for it.

justinstoller:
If anyone else is reviewing this, this is the pertinent commit that added this functionality in Puppet: https://github.com/puppetlabs/puppet/commit/747076f00ce7c6b3b86acf2823be9c6536a821d4

ekohl:
This week will be very busy so probably won't have time for it. Feel free to take finish it or leave it for me for next week.

justinstoller:
I didn't get to this this week (and I'm out Friday), I'd give it a 50/50 chance that I'll have time next week to get to it. I don't care which of us do it.

ekohl:
I've pushed some tests, but since I don't have a local dev env with clojure I'm relying on CI to test this.

PR #2129, opened 2019-07-23
(docs) add rel note, known issue, about cipher warnings

jbondpdx:


documentation/release_notes.markdown
```diff
@@ -16,6 +16,20 @@ Released 22 July 2019
 
 ### New features
 
+- This release includes an upgrade to the latest release of Jetty's 9.4 series. With this update, you may see "weak cipher" warnings about ciphers that were previously enabled by default. Puppet Server now defaults to stronger FIPS-compliant ciphers, but you must first remove the weak ciphers.
+
+The ciphers previously enabled by default have not been changed, but are considered weak by the updated standards. Remove the weak ciphers by removing the `cipher-suite` configuration section from the `webserver.conf`. After you do this, Puppet Server uses the FIPS-compliant ciphers instead. This release includes the weak ciphers for backward compatibility only.
```
justinstoller: I think this and the following paragraphs need to be indented to be included in the bullet point format.

documentation/release_notes.markdown
```diff
@@ -16,6 +16,20 @@ Released 22 July 2019
...
+
+The FIPS-compliant cipher suites, which are not considered weak, will be the default in a future version of Puppet. To maintain backwards compatibility, Puppet Server explicitly enables all cipher suites that were available as of Puppet Server 6.0. When you upgrade to Puppet Server 6.5.0, this affects you in in two ways:
```
justinstoller: > this affects you in in two ways

duplicate "in"

documentation/release_notes.markdown
justinstoller: > After you do this, Puppet Server uses....

Should it be

> After doing this, Puppet Server uses

Or

> After you do this, Puppet Server will use

?

jbondpdx: I think it's correct as it is. "After doing this..." implies that Puppet Server is removing the weak ciphers, and the future tense isn't necessary in the second possibility.

After the user does this thing, Puppet Server will use, and will continue to use on a regular basis, the new ciphers.

All that said, "after you do this" is a little lazy/rushed writing, so I'm making it more explicit.

justinstoller:
No travis testing required for docs PRs.

PR #2130, opened 2019-07-25
(maint) Clean up project.clj

justinstoller:
The commit b8e8ade that removes the hook may be a bad idea, or at least need a cjc change, I'm not sure.

I'm also not sure if the profile twiddling is better than the top level nrepl dependency or not.

I'd like to wait until I can get a lein-parent release of 0.3.7 to fix some warnings. See https://github.com/achin/lein-parent/pull/20 .

I didn't update the versions of voom, but I also don't _think_ it's used, but couldn't find the CI configs to confirm one way or the other.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -149,8 +149,7 @@
              :uberjar {:aot [puppetlabs.trapperkeeper.main]
                        :dependencies [[puppetlabs/trapperkeeper-webserver-jetty9]]}
              :ci {:plugins [[lein-pprint "1.1.1"]
-                            [lein-exec "0.3.7"]]}
-             :voom {:plugins [[lein-voom "0.1.0-20150115_230705-gd96d771" :exclusions [org.clojure/clojure]]]}}
+                            [lein-exec "0.3.7"]]}}
 
   :test-selectors {:integration :integration
                    :unit (complement :integration)}
--- a/project.clj
+++ b/project.clj
@@ -25,7 +25,7 @@
 (defproject puppetlabs/puppetserver ps-version
   :description "Puppet Server"
 
-  :min-lein-version "2.7.1"
+  :min-lein-version "2.8.1"
 
   :parent-project {:coords [puppetlabs/clj-parent "3.0.0"]
                    :inherit [:managed-dependencies]}
@@ -113,8 +113,8 @@
 
   :profiles {:dev {:source-paths  ["dev"]
                    :dependencies  [[org.clojure/tools.namespace]
-                                   [puppetlabs/trapperkeeper-webserver-jetty9 nil]
-                                   [puppetlabs/trapperkeeper-webserver-jetty9 nil :classifier "test"]
+                                   [puppetlabs/trapperkeeper-webserver-jetty9]
+                                   [puppetlabs/trapperkeeper-webserver-jetty9 :classifier "test"]
                                    [puppetlabs/trapperkeeper nil :classifier "test" :scope "test"]
                                    [puppetlabs/trapperkeeper-metrics :classifier "test" :scope "test"]
                                    [puppetlabs/kitchensink nil :classifier "test" :scope "test"]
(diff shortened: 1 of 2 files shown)


Magisus:
I'm pretty sure this is the only project that still does use voom. https://github.com/puppetlabs/ci-job-configs/blob/master/resources/scripts/packaging-clj-lein-voom-deploy.sh is used when we promote puppetserver into pe-puppet-server-extensions, see https://github.com/puppetlabs/ci-job-configs/blob/master/resources/job-groups/puppetserver.yaml#L88.

justinstoller:
Do you know if that profile is also defined in a user profiles.clj? Or where to look at least?

Magisus:
Ah yeah, found it: https://github.com/puppetlabs/puppetlabs-modules/blob/dff6c54101df746a2454b3d061bfc47689a5837e/site/profile/templates/jenkins/slave/cljtools/profiles.clj.erb#L12 which is pulled in by the [cljtools](profile) which in turn is used when using a [`worker` node](https://github.com/puppetlabs/puppetlabs-modules/blob/839a0db934ed61e98a8d57b1e181e400a3640c52/site/role/manifests/jenkins/slave/worker.pp#L21) (which we do for our voom deploy) in Jenkins.

Sorry, didn't even see that the profile was named differently. Since we use this instead of the per-project definition, you're right, I think it's safe to remove it from this `project.clj`.

project.clj
```diff
@@ -22,10 +22,6 @@
     heap-size-from-profile-clj
     default-heap-size))
 
-(def figwheel-version "0.3.7")
-(def cljsbuild-version "1.1.7")
-(def clojurescript-version "1.10.238")
```
Magisus: Oh wow, I'm impressed this is still in here. I feel like we've done several rounds of cleaning up the vestiges of that :P

project.clj
```diff
@@ -82,7 +82,7 @@
                  ["snapshots" "https://artifactory.delivery.puppetlabs.net/artifactory/clojure-snapshots__local/"]]
 
   :plugins [[lein-parent "0.3.1"]
-            [puppetlabs/i18n "0.8.0"]]
+            [puppetlabs/i18n "0.8.0" :hooks false]]
```
Magisus: Love how the way to disable this isn't well documented anywhere obvious...

Magisus:
Oh boy, something terrible is going on in Travis... Did something change in the logging maybe?

justinstoller:
Removed the voom stuff.

The i18n stuff doesn't seem to be a blocker to normal development workflows so I'm pretty happy dropping it and its warnings and if/when we add we start translating server we can look into doing it in a non-deprecated way.

I'm torn on the nrepl bit, the override is more explicit, though I think the way I'm doing it in this PR is less work long term. Love additional views from @tvpartytonight or @pcarlisle (especially since I'm not an emacs user).

Didn't look into the travis output, but I don't see anything weird in my tests locally (though JRuby seems to be logging at debug for dev-setup?). Will look into travis when it finishes.

justinstoller:
Ugh, the regression in travis is because JRuby during the `dev-setup` script is being killed by travis for producing too much output.

The reason is because of the base profile override... It's not just overriding the dependencies but all of the defaults that lein sets in that profile. The pertinent one is adding `dev-resources` to the classpath, w/o that the dev logback configuration isn't loaded and JRuby logging isn't configured correctly.

In my mind that kinda decides it between the explicit override and the base override options....


However, I am inclined to update our min lein version to 2.9.1 and remove the nrepl line for now, with 2.9.1 we don't need it and when we do need it again it will be on `nrepl/nrepl` rather than `tools.nrepl`... thoughts?

Magisus:
Using lein 2.9.1 and removing the nrepl line is what we've done in the later branches of pe-pse, and it seems to be working well. I don't think we officially updated the min lein version there, but I think we probably should, both here and there.

justinstoller:
Okay, updated and passed travis!

PR #2131, opened 2019-07-25
(maint) Update healthcheck to use simple status endpoint

underscorgan:


docker/puppetserver-standalone/healthcheck.sh
```diff
@@ -13,6 +13,6 @@ curl --fail \
 --cert   $hostcert \
 --key    $hostprivkey \
 --cacert $localcacert \
-"https://${hostname}:${PUPPET_MASTERPORT}/${PUPPET_HEALTHCHECK_ENVIRONMENT}/status/test" \
-|  grep -q '"is_alive":true' \
+"https://${hostname}:${PUPPET_MASTERPORT}/status/v1/simple" \
+|  grep -q 'running' \
```
justinstoller: I think the http code is enough to disambiguate between up and down, fwiw (see how Puppet does it here: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/configurer.rb#L407)

I don't think it matters here but for PE at least you might want to specify the master service like the agent does. fwiw, during HA fail over some PE services besides the master may be degraded and w/o specifying a specific service if any services in the tk app is bad then the resopnse is bad.

nwolfe: @underscorgan We should remove the `PUPPET_HEALTHCHECK_ENVIRONMENT` variable from the Dockerfile(s) and READMEs too?

nwolfe: Ah I see what you mean Justin...

Good to know, since we have a different healthcheck.sh for the pe-puppetserver container, so we may want to do what you're suggesting there.

underscorgan: I'm going to kick this a few more times to see if I can be confident this is resolving the network issue, then will remove the unneeded var :)

Iristyle: Making curl *only* return a 0 exit code on a HTTP 200 OK is actually quite difficult.

So the basic approach we have here is the right one.

Iristyle:
I created a new PR for reworking the healthcheck at https://github.com/puppetlabs/puppetserver/pull/2137 before realizing we already had this PR... but still, I think we can close this one!

PR #2132, opened 2019-07-29
(WIP) Pr/2087

Iristyle:
Some tests based on #2087


Iristyle:
I think we determined the network flakiness issues were resolved by updating to a newer docker nightly for Windows built on Aug 1st 2019.

So this can be closed given #2087 was merged.

PR #2134, opened 2019-07-31
(SERVER-1914) Allow running Puppet Server dev environment in FIPS compliant mode

justinstoller:


project.clj
```diff
@@ -89,10 +89,10 @@
                  [puppetlabs/trapperkeeper-scheduler]
                  [puppetlabs/trapperkeeper-status]
                  [puppetlabs/kitchensink]
-                 [puppetlabs/ssl-utils]
+                 [puppetlabs/ssl-utils "2.0.2-SNAPSHOT"]
                  [puppetlabs/ring-middleware]
                  [puppetlabs/dujour-version-check]
-                 [puppetlabs/http-client]
+                 [puppetlabs/http-client "1.1.0-SNAPSHOT"]
```
justinstoller: these need to be built locally for now

project.clj
```diff
@@ -184,12 +185,19 @@
...
+                                               ;; FOSS we always want to include the FOSS Bouncy Castle
+                                               [org.bouncycastle/bcpkix-jdk15on]
                                                [puppetlabs/puppetserver ~ps-version]
                                                [puppetlabs/trapperkeeper-webserver-jetty9 ~jetty9-version]
                                                [org.clojure/tools.nrepl nil]]
                       :plugins [[puppetlabs/lein-ezbake "2.0.4"]]
                       :name "puppetserver"}
-             :uberjar {:aot [puppetlabs.trapperkeeper.main
+             :uberjar {;; When running the uberjar task we want to build with FOSS Bouncy Castle,
+                       ;; unclear if this renders the value in the :ezbake profile redundant.
```
justinstoller: Need to do some ezbaking to double check

src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
```diff
@@ -42,6 +42,14 @@ def self.initialize_puppet(puppet_config)
 
     Puppet::ApplicationSupport.push_application_context(master_run_mode)
 
+    # We should never end up using the ssl_context that Puppet creates in
+    # JRuby, however Puppet will load one before it calls to
+    # `Puppet::Network::HttpPool.connection` (which is what then delegates to
+    # http client). This may cause warnings if only an http (vs https)
+    # connection is requested (though we never request an http connection in
+    # JRuby from our code).
+    Puppet.push_context({ssl_context: true})
```
justinstoller: This may be incorrect if users have disabled SSL routing/using the http client for HTTP based integrations. See this line: https://github.com/puppetlabs/puppet/blob/6.7.2/lib/puppet/network/http_pool.rb#L87  However, correcting it would mean altering the behavior the HttpPool in Puppet.

justinstoller:
A different approach to putting the bc dependency in our dev, uberjar, and ezbake profiles would be to put it as a top level dependency and have projects that pull puppetserver in as a library (pepse) explicitly exclude the bc dependency.

src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
justinstoller: @joshcooper do you know if there's a valid use case for consumers of the `HttpPool` to request HTTP (`use_ssl: false`) connections in a running server, afaict, if so we'll need to refactor the context handling linked above, or live with a warning being emitted.

joshcooper: I can imagine a report processor might want to. I'm fine removing the warning in puppet if that simplifies things.

src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
```diff
@@ -42,6 +42,15 @@ def self.initialize_puppet(puppet_config)
...
     Puppet::ApplicationSupport.push_application_context(master_run_mode)
 
+    # Puppet's https machinery expects to find an object at
+    # `Puppet.lookup(:ssl_context)` which it will then use as an input
+    # to the Verifier, which is passed to the client. We ignore these
+    # values when passed to our https client and manage our SSLContext
+    # in a completely different way.
+    #
+    # See `Puppet::Network::HttpPool.connection`
+    Puppet.push_context({ssl_context: :unused})
```
justinstoller: @pcarlisle this is the puppet-server-using-push_context thing I mentioned at the soiree.

justinstoller:
This requires https://github.com/puppetlabs/puppet/pull/7642 for our own tests to pass in fips mode. I would love to get this in and when the above PR get promoted into server updating the .travis.yml to test in fips regularly.

Magisus:
Looks like the docker test failure might be something related to this:
```
Syntax error macroexpanding at (core.clj:1:1).

Execution error (ClassNotFoundException) at java.net.URLClassLoader/findClass (URLClassLoader.java:382).

org.bouncycastle.asn1.x500.X500Name
```
Are we concerned about that?

Magisus:
> A different approach to putting the bc dependency in our dev, uberjar, and ezbake profiles would be to put it as a top level dependency and have projects that pull puppetserver in as a library (pepse) explicitly exclude the bc dependency.

I think it depends on whether we intend FIPS mode to be usable in FOSS, even if we don't support it (or ship packages for it). If we don't, it might make more sense to move the decision up to the consuming project. But I don't fully understand all the tradeoffs here.

justinstoller:
wrt the profile setup. this is technically a breaking change for those who use puppetserver.jar as a library (like we do in PE) but not for those who develop, test from source or consume the packages. 

The alternative would be to keep it in our dependencies, which would maintain the explicit dependency of our jar on bouncycastle foss and negate having to add the dependency in the dev and uberjar profiles. However it would require us to exclude the bouncycastle dependency when we pull puppetserver in as a dependency in PE.

This approach moves some of the complexity into puppetserver to make our PE build process simpler. It also is the same approach that we took with ssl-utils and http-client libraries so there's consistency between projects wrt fips support.

justinstoller:
>> Syntax error macroexpanding at (core.clj:1:1).

> Are we concerned about that?

It looks like docker is calling `lein ezbake local-build` which... I've never used. I validated this running `lein with-profile ezbake ezbake build`, I thought all invocations of ezbake eventually called `lein uberjar` and unpacked that. I'll look more into this tomorrow and may have to add back the bc dependency in the ezbake profile....

Magisus:
Local ezbake build is failing for me with this:
```
Compiling puppetlabs.trapperkeeper.services.status.status-service
Syntax error macroexpanding at (core.clj:1:1).
Execution error (ClassNotFoundException) at java.net.URLClassLoader/findClass (URLClassLoader.java:381).
org.bouncycastle.asn1.x500.X500Name
```

I used this to invoke it:
```
lein clean && lein install && EZBAKE_ALLOW_UNREPRODUCIBLE_BUILDS=true EZBAKE_NODEPLOY=true JENKINS_USER_AUTH=<creds> lein with-profile ezbake,provided ezbake build
```

underscorgan:
@justinstoller it's definitely possible that `ezbake local-build` isn't doing quite the right thing, the changes to do that in docker are relatively recent and I think that's the first thing to really use `local-build`

PR #2135, opened 2019-07-31
(DOCUMENT-1039) Update 6.5.0 cipher-suites release note

justinstoller:


qralston:
I think these edits are even better than the ones I suggested on the upstream JIRA ticket, as they give appropriate background (e.g. the reference to FIPS mode) without getting in the way of the core explanation (why the changes were made and what Puppet administrators should do in response). Some of my suggested wording was also cleaned up (and appropriately so).

justinstoller:
Thank you, James.

PR #2138, opened 2019-08-05
[wip] (maint) test new jruby

justinstoller:
See how this takes in travis vs what's currently there.


justinstoller:
The timings between this and other PRs that aren't using a newer JRuby seem pretty similar. We're going ahead with the Jruby bump.

PR #2140, opened 2019-08-06
(maint) Azure YML standardization

Iristyle:


Iristyle:
Travis fail here is completely unrelated, so merging

PR #2141, opened 2019-08-08
Merge 6.3.x -> master

Magisus:


project.clj
```diff
@@ -162,28 +142,10 @@
...
-                                               [org.clojure/clojure nil]
-                                               ;; I honestly don't know why we should need this
-                                               ;; But building with ezbake is consistently failing and
-                                               ;; pulling in an old build of clojurescript without it.
-                                               [org.clojure/clojurescript ~clojurescript-version]
+                                               ;; brings in its own version.
+                                               [org.clojure/clojure]
                                                [puppetlabs/puppetserver ~ps-version]
                                                [puppetlabs/trapperkeeper-webserver-jetty9 ~jetty9-version]
                                                [org.clojure/tools.nrepl nil]]
```
justinstoller: I don't think this should be here.

Magisus: Ah yeah I see it, fixed.

PR #2142, opened 2019-08-09
(wip) What it takes to use the new clj-parent

justinstoller:
don't merge, just for demonstration, I need to incorporate this with the don't-use-bouncy-castle-in-regular-dependencies that's happening in the fips PR. In fact, I'll probably role this into that. But I wanted to publicly sanity check myself that we weren't off the rails with all these clj-parent changes.


justinstoller:
Going to fold this into the work for FIPS since we need the clj-parent bump there.

PR #2143, opened 2019-08-09
(SERVER-2529) Allow request handling with one shared JRuby instance

Magisus:
This commit allows you to specify "jruby-puppet.multithreaded=true" to
switch into a mode where all requests requiring a jruby instance are
passed through a single JRuby, designed to be running a thread-safe
puppet implementation.

It does not address other uses of JRuby instances (like settings
loading), it does not respect `max-requests-per-instance`, and it does
not work with pool locking (the instance is never returned so a request
to lock will block forever).

### DIFF
--- a/dev/puppetserver.conf
+++ b/dev/puppetserver.conf
@@ -87,6 +87,10 @@ jruby-puppet: {
     # (optional) Whether or not to track lookups during compilation; turning
     # this on will send that information to puppetdb
     # track-lookups: true
+
+    # For testing running requests through a single JRuby instance. DO NOT ENABLE unless
+    # explicitly testing this functionality.
+    # multithreaded: true
 }
 
 # Settings related to HTTP client requests made by Puppet Server.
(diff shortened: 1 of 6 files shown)


Magisus:
This is one idea for an approach. The next one I am looking into involves updating the pool implementation itself to hand out references to a single backing JRuby instance, rather than new instances themselves. This would potentially allow us to switch modes by simply pulling in a different JRubyPoolManager service in puppetserver, which creates the new version of the pool and uses all of the existing APIs.

justinstoller:
Seems like an approach that we could potentially go with, though I'm not sure how we'd do the locking with this.

Sharpie:
JRuby also offers a few different context models. Right now, we're hard-coded to `LocalContextScope/SINGLETHREAD` in jruby-utils, but there are options:

https://github.com/jruby/jruby/wiki/RedBridge#context-instance-type

`LocalContextScope/SINGLETON` looks pretty interesting. It seems like that would allow us to keep a lot of the pool logic unchanged: we create multiple scripting containers and hand those out. `max-active-instances` is still used to throttle request concurrency. What changes is that the containers are all sharing one Ruby runtime and variable map behind the scenes.

Sharpie:
On the other hand, implementing `max-requests-per-instance` would be interesting with `LocalContextScope/SINGLETON`.

Magisus:
Let's see if we can get this into a state where we can merge it for testing our threadsafe work with. Pinged some reviewers.

justinstoller:
Merging this for testing purposes (it allows us to more easily test other, related work). This will either come out, or be reworked to handle max-requests & locking, later.

PR #2144, opened 2019-08-12
(SERVER-2543) Add endpoints for plans list and plan metadata

branan:
This adds the endpoints to expose the plan information service
from Puppet as an HTTP service.

Primarily this code is copy&paste from the similar glue for tasks,
with bits removed as needed (mostly around metadata and file lists,
since currently plans have neither of those). It seemed better to
duplicate rather than try to share code paths, since task and plan
metadata formats are more likely to diverge than converge.

The test for this currently includes an assertion that plan metadata
is empty, which will be a nice reminder for us to add tests for
metadata when it becomes available :sunglasses:

### DIFF
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -167,6 +167,20 @@
                   :params {:environment schema/Str
                            (schema/optional-key :code_id) schema/Str}}}]})
 
+(def PlanData
+  "Response from puppet's PlanInformationService for data on a single
+  plan, *after* it has been converted to a Clojure map."
+  {:metadata (schema/maybe {schema/Keyword schema/Any})
+   :files [{:name schema/Str :path schema/Str}]
+   (schema/optional-key :error) {:msg schema/Str
+                                 :kind schema/Str
+                                 :details {schema/Keyword schema/Any}}})
+
+(def PlanDetails
+  "A filled-in map of information about a plan."
+  {:metadata {schema/Keyword schema/Any}
+   :name schema/Str})
+
 (defn obj-or-ruby-symbol-as-string
   "If the supplied object is of type RubySymbol, returns a string
   representation of the RubySymbol.  Otherwise, just returns the original
@@ -386,6 +400,11 @@
   [task :- schema/Str]
   (rr/not-found (i18n/tru "Could not find task ''{0}''" task)))
 
+(schema/defn plan-not-found :- ringutils/RingResponse
+  "Ring handler to provide a standard error when a plan is not found."
+  [plan :- schema/Str]
+  (rr/not-found (i18n/tru "Could not find plan ''{0}''" plan)))
+
 (schema/defn ^:always-validate
   module-info-from-jruby->module-info-for-json  :- EnvironmentModulesInfo
   "Creates a new map with a top level key `name` that corresponds to the
@@ -505,6 +524,30 @@
    (:kind err)
    (str/starts-with? (:kind err) "puppet.task")))
 
+(defn handle-plan-details-jruby-exception
+  "Given a JRuby RaiseException arising from a call to plan-details, constructs
+  a 4xx error response if appropriate, otherwise re-throws."
+  [jruby-exception environment module plan]
+  (cond
+    (exception-matches? jruby-exception #"^\(EnvironmentNotFound\)")
+    (environment-not-found environment)
+
+    (exception-matches? jruby-exception #"^\(MissingModule\)")
+    (module-not-found module)
+
+    (exception-matches? jruby-exception #"^\(PlanNotFound\)")
+    (plan-not-found plan)
+
+    :else
+    (throw jruby-exception)))
+
+(defn is-plan-error?
+  [err]
+  (and
+   (map? err)
+   (:kind err)
+   (str/starts-with? (:kind err) "puppet.plan")))
+
 (schema/defn ^:always-validate
   task-details-fn :- IFn
   "Middleware function for constructing a Ring response from an incoming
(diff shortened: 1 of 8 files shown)


src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -675,6 +810,24 @@
       jruby-request/wrap-with-environment-validation
       jruby-request/wrap-with-error-handling))
 
+(schema/defn ^:always-validate
+  all-plans-handler :- IFn
+  "Handler for processing an incoming all_plans Ring request"
+  [jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)]
+  (-> (all-plans-fn jruby-service)
+      (jruby-request/wrap-with-jruby-instance jruby-service)
```
McdonaldSeanp: So, if we get the jruby instance here:

Does this mean all processing - including data transformation outside of the ruby calls - is done with a jruby instance lock? I don't know how much this matters, but if that is the case that would mean we are getting a lock for slightly longer than we need, since all we really care about is the one call to `.GetPlans` (or `.GetPlanData` for `plan-details-handler`)

I don't know how _much_ that would effect performance: It's not a huge amount of processing that we do outside the call to Jruby. However, since we are adding code to share jruby instances with compilation, maybe every small amount of time counts?

I can see an argument that the small amount of processing done here doesn't really matter, I just want to be pedantic and careful about resource sharing with the jruby instances.

test/integration/puppetlabs/services/master/plans_int_test.clj
```diff
@@ -0,0 +1,143 @@
...
+      (do
+        (testutils/write-plans-files "apache" "announce" "return 'Hi!'")
+        (testutils/write-plans-files "graphite" "install" "return 'Wheeeee'")
+        (let [expected-response '({"name" "apache::announce"
+                                   "environment" [{"name" "production"
+                                                   "code_id" nil}]}
+                                  {"name" "graphite::install"
+                                   "environment" [{"name" "production"
+                                                   "code_id" nil}]})
+              response (get-all-plans "production")]
```
McdonaldSeanp: Should we test for correctly returned `environment-not-found` output when the environment doesn't exist?

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -533,6 +576,98 @@
...
+      (if-let [plan-info-for-env
+               (sort-nested-info-maps
+                 (jruby-protocol/get-plans jruby-service
+                                           (:jruby-instance
+                                             request)
+                                           environment))]
+        (all-plans-response! plan-info-for-env
+                             environment
+                             jruby-service)
+        (environment-not-found environment)))))
```
McdonaldSeanp: What happens when an environment exists but has no plans? what is returned? should it be something other than 200 if there are no plans ?

src/clj/puppetlabs/services/master/master_core.clj
branan: This is how all of puppetserver does it - jruby checkout is an HTTP middleware

src/clj/puppetlabs/services/master/master_core.clj
branan: Nope. This is consumed by code, so saying "We got a thing with no items" is much easier for other services to handle than something like a 404

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -533,6 +576,98 @@
...
+   environment :- schema/Str
+   jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)]
+  (let [format-plan (fn [plan-object]
+                      {:name (:name plan-object)
+                       :environment [{:name environment
+                                      :code_id nil}]})]
+    (->> (map format-plan info-from-jruby)
+         (middleware-utils/json-response 200))))
+
+(schema/defn ^:alwaysa-validate
```
branan: typo

src/clj/puppetlabs/services/master/master_core.clj
McdonaldSeanp: Okay, returning 200 makes sense, do we still need to capture the 'environment exists but no plans exist' situation? or do we just return an empty response?

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -505,6 +524,30 @@
    (:kind err)
    (str/starts-with? (:kind err) "puppet.task")))
 
+(defn handle-plan-details-jruby-exception
+  "Given a JRuby RaiseException arising from a call to plan-details, constructs
```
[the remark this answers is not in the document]
branan: I don't think so right now, so we don't actually load them (we just ask the environment for a list of them). In the future when we try to parse out metadata we might

test/integration/puppetlabs/services/master/plans_int_test.clj
```diff
@@ -0,0 +1,143 @@
...
+      (ks/dissoc-in [:jruby-puppet
+                     :environment-class-cache-enabled])))
+
+(deftest ^:integration all-plans-with-env
+  (testing "full stack plans listing smoke test"
+    (bootstrap/with-puppetserver-running-with-config
+      app
+      puppet-config
+      (do
+        (testutils/write-plans-files "apache" "announce" "return 'Hi!'")
```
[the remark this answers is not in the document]
branan: Sure, should be easy enough

test/integration/puppetlabs/services/master/plans_int_test.clj
branan: Given that's handled by common helpers and we do so below, I don't think it's necessary

src/clj/puppetlabs/services/master/master_core.clj
branan: We just get an empty list from the info-service and will do all our transformations on each element of that list, so it'll all just work :)

PR #2145, opened 2019-08-12
(SERVER-2585) hiera lookup config setting

tvpartytonight:
This PR introduces a new lookup_key_recorder to track hiera lookups, as well as a new setting in the jruby-puppet configuration to turn aforementioned key_recorder on or off.

### DIFF
--- a/acceptance/suites/pre_suite/foss/95_install_pdb.rb
+++ b/acceptance/suites/pre_suite/foss/95_install_pdb.rb
@@ -10,9 +10,15 @@ teardown do
   on(master, "rm -f #{sitepp}")
 end
 
-step 'Install Puppet Release Repo' do
-  install_puppetlabs_release_repo_on(master, 'puppet5')
-end
+# Install the SNAPSHOT with the necessary modifications to the
+# puppetdb terminus; once PuppetDB has tagged and released, revert
+# this commit and update the `install_puppetlabs_release_repo_on`
+# to use 'puppet6'
+
+# step 'Install Puppet Release Repo' do
+#   install_puppetlabs_release_repo_on(master, 'puppet5')
+# end
+install_puppetlabs_dev_repo master, 'puppetdb', "6.4.1.SNAPSHOT.2019.08.13T0822"
 
 step 'Install PuppetDB module' do
   on(master, puppet('module install puppetlabs-puppetdb'))
--- a/acceptance/suites/tests/00_smoke/puppetdb_integration.rb
+++ b/acceptance/suites/tests/00_smoke/puppetdb_integration.rb
@@ -29,12 +29,32 @@ run_timestamp = nil
 master_fqdn = on(master, '/opt/puppetlabs/bin/facter fqdn').stdout.chomp
 random_string = SecureRandom.urlsafe_base64.freeze
 
+puppetserver_conf_path = '/etc/puppetlabs/puppetserver/conf.d/puppetserver.conf'
+old_puppetserver_conf = on(master, "cat #{puppetserver_conf_path}").stdout
+
+common_yaml_path = '/etc/puppetlabs/code/environments/production/data/common.yaml'
+
+create_remote_file(master, common_yaml_path, <<EOM)
+---
+some-test-key: test-key-value
+EOM
+
+on(master, "chmod 644 #{common_yaml_path}")
+
+step 'Configure puppetserver to track hiera lookups' do
+  conf = read_tk_config_string(old_puppetserver_conf)
+  conf['jruby-puppet']['track-hiera-lookups'] = true
+  modify_tk_config(master, puppetserver_conf_path, conf, replace=true)
+end
+
 step 'Configure site.pp for PuppetDB' do
   sitepp = '/etc/puppetlabs/code/environments/production/manifests/site.pp'
   create_remote_file(master, sitepp, <<EOM)
 node 'resource-exporter.test' {
   @@notify{'#{random_string}': }
 }
+
+lookup('some-test-key')
 
 node '#{master_fqdn}' {
   Notify<<| title == '#{random_string}' |>>
@@ -46,7 +66,8 @@ EOM
   on(master, "chmod 644 #{sitepp}")
 
   teardown do
-    on(master, "rm -f #{sitepp}")
+    modify_tk_config(master, puppetserver_conf_path, read_tk_config_string(old_puppetserver_conf), replace=true)
+    on(master, "rm -f #{sitepp} #{common_yaml_path}")
   end
 end
 
@@ -67,6 +88,7 @@ EOM
 
     teardown do
       on(master, puppet('node', 'deactivate', 'resource-exporter.test'))
+      on(master, "puppet ssl clean --certname resource-exporter.test")
       on(master, 'puppetserver ca clean --certname=resource-exporter.test')
     end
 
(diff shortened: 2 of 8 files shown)


dev/puppetserver.conf
```diff
@@ -83,6 +83,10 @@ jruby-puppet: {
     # (optional) whether to use the environment class cache. If unspecified
     # defaults to false
     environment-class-cache-enabled: true
+
+    # (optional) Whether or not to implement a lookup recorder during compilation to track
+    # what keys are looked up and how often
```
pcarlisle: should probably use less implementation specific language and say something about what the end result is

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -59,5 +59,6 @@
    :http-client-idle-timeout-milliseconds schema/Int
    :http-client-metrics-enabled schema/Bool
    :max-requests-per-instance schema/Int
-   :use-legacy-auth-conf schema/Bool})
+   :track-hiera-lookups schema/Bool
+:use-legacy-auth-conf schema/Bool})
```
pcarlisle: something happened here with indentation?

tvpartytonight:
I wasn't sure where the default jruby-puppet configuration comes from...I looked in the ezbake section, but didn't find it there. @justinstoller any ideas?

src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -66,13 +72,19 @@ def handleRequest(request)
...
+    Puppet.override(lookup_key_recorder: lookup_key_recorder) do
+      @catalog_compiler.compile(convert_java_args_to_ruby(request_data))
+    end
   end
 
   def compileAST(compile_options)
     Puppet::Server::ASTCompiler.compile(convert_java_args_to_ruby(compile_options))
   end
 
+  def lookup_key_recorder
```
pcarlisle: should put a verb in this name so that it's clear that it's making a new one each time instead of just some saved instance variable

justinstoller:
> where the default jruby-puppet configuration comes from

It's here: https://github.com/puppetlabs/puppetserver/blob/master/resources/ext/config/conf.d/puppetserver.conf#L2

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -59,5 +59,6 @@
    :http-client-idle-timeout-milliseconds schema/Int
    :http-client-metrics-enabled schema/Bool
    :max-requests-per-instance schema/Int
-   :use-legacy-auth-conf schema/Bool})
+   :track-hiera-lookups schema/Bool
```
pcarlisle: looks like this should be an optional key?

tvpartytonight:
@Iristyle and @nwolfe it looks like the test failures here from azure and the docker container tests in travis are some sort of transient, and not a real cause for concern here?

nwolfe:
@tvpartytonight In this case I see some real-looking errors in the logs:
- https://travis-ci.org/puppetlabs/puppetserver/jobs/571095790#L4398
- https://travis-ci.org/puppetlabs/puppetserver/jobs/571095790#L4433

The only transient error that we expect is the "unknown host"-type, usually from `getaddrinfo`:
- https://travis-ci.org/puppetlabs/puppetserver/jobs/571095790#L3668

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -87,7 +87,9 @@
                            [:master-run-dir "rundir"]
                            [:master-log-dir "logdir"]]]
       (if-let [value (get config setting)]
-        (.put puppet-config dir (ks/absolute-path value))))
+        (.put puppet-config dir (ks/absolute-path value)))
+      (if-let [value (get config :track-hiera-lookups)]
```
justinstoller: Either the indentation is wrong or the nesting is. We don't want to do this in the `doseq` I assume?

resources/ext/config/conf.d/puppetserver.conf
```diff
@@ -52,6 +52,10 @@ jruby-puppet: {
     # specified in the Puppet Server HOCON-formatted auth.conf (if false or not
     # specified).
     #use-legacy-auth-conf: true
+
+    # (optional) Whether or not to track hiera lookups during compilation; turning
+    # this on will send that information to puppetdb
+    # track-hiera-lookups: true
```
justinstoller: piggybacking on Patrick's previous comment, should we mention hiera here? (I know the function moved from `heira()` to `lookup()`. I would assume `track-lookups` would be sufficient?

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
tvpartytonight: good catch here, I think this is why my test setup isn't working.

resources/ext/config/conf.d/puppetserver.conf
tvpartytonight: Commit has been added to change to remove hiera language.

tvpartytonight:
@nwolfe looks like those errors are from the fact that this work needs code from an unreleased version of puppet.

[This](https://travis-ci.org/puppetlabs/puppetserver/jobs/571095790#L4154) error is looking for the key_recorder added in this [PR](https://github.com/puppetlabs/puppet/pull/7523/files), but that won't be released until the next agent.

In case the link goes away, the error is:
```
LoadError: no such file to load -- puppet/pops/lookup/key_recorder
```

src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -66,13 +73,19 @@ def handleRequest(request)
   end
 
   def compileCatalog(request_data)
-    @catalog_compiler.compile(convert_java_args_to_ruby(request_data))
+    Puppet.override(lookup_key_recorder: create_recorder) do
```
Magisus: I was trying to think if triggering collection on this endpoint when it's used with impact analysis would get weird (e.g. overwrite inputs from real catalogs), but currently we're not saving any of the catalogs compiled that way in PDB, so I think this is fine. It will collect for ACE nodes, but that's probably a good thing.

acceptance/suites/pre_suite/foss/95_install_pdb.rb
```diff
@@ -10,9 +10,15 @@
   on(master, "rm -f #{sitepp}")
 end
 
-step 'Install Puppet Release Repo' do
-  install_puppetlabs_release_repo_on(master, 'puppet5')
-end
+# Install the SNAPSHOT with the necessary modifications to the
+# puppetdb terminus; once PuppetDB has tagged and released, revert
```
justinstoller: This is going out Tuesday, right?

acceptance/suites/tests/00_smoke/puppetdb_integration.rb
```diff
@@ -145,3 +167,33 @@
...
 end
+
+step 'Validate PuppetDB stored lookup data sent from puppetserver' do
+  query = "curl -X POST http://localhost:8080/pdb/query/v4/catalog-input-contents" <<
+    " -H 'Content-Type:application/json'" <<
+    " -d '{\"query\": \"\" }'"
+
+  retries = 3
+
+  (1..(retries + 1)).each do |i|
```
justinstoller: If we don't mind 0 indexing we could just do `retries.times do |i|` here

src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -66,13 +73,19 @@ def handleRequest(request)
...
+      @catalog_compiler.compile(convert_java_args_to_ruby(request_data))
+    end
   end
 
   def compileAST(compile_options)
     Puppet::Server::ASTCompiler.compile(convert_java_args_to_ruby(compile_options))
   end
 
+  def create_recorder
+    @track_lookups ? Puppet::Server::KeyRecorder.new : Puppet::Pops::Lookup::KeyRecorder.singleton
```
justinstoller: Why not `nil`? The Puppet code that calls this uses a default value right? Should we remove the default value in Puppet?

PR #2148, opened 2019-08-15
(SERVER-2388) Upgrade JRuby and bring in clj-parent 4.0.3

justinstoller:


justinstoller:
This is basically, https://github.com/puppetlabs/puppetserver/pull/2142 but my assumption https://github.com/puppetlabs/puppetserver/pull/2134 would get merged and bring in these changes with it first was wrong. Might as well get these changes in now.

project.clj
```diff
@@ -53,22 +51,18 @@
                  ;; send their logs to logstash, so we include it in the jar.
                  [net.logstash.logback/logstash-logback-encoder]
 
-                 [puppetlabs/jruby-utils "2.1.0"]
-                 [puppetlabs/jruby-deps "9.2.0.0-1"]
-
-                 ;; Update to flatland usable in jdk11
-                 [org.flatland/useful "0.11.6"]
-                 [org.flatland/ordered "1.5.7"]
```
Magisus: Did we update the thing that was pulling in old versions of these?

justinstoller: yeah, older flatlands were pulled in by circleci/clj-yaml

PR #2150, opened 2019-08-16
(maint) Remove jira-ruby gem

Magisus:
The jira-ruby gem used to be used for generating tickets via a rake
task, but the only tickets we autogenerate now come from a centralized
repo of tasks, so we don't need this dep anymore. This commit removes
it, following an update to it that started limiting the Ruby versions
with which it can be installed.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,7 +12,6 @@ end
 
 gem 'packaging', *location_for(ENV['PACKAGING_LOCATION'] || '~> 0.99')
 gem 'rake', :group => [:development, :test]
-gem 'jira-ruby', :group => :development
 
 group :test do
   gem 'rspec'


Magisus:
For context https://github.com/puppetlabs/puppetserver/commit/5dcf3203a27a5aa6199283b66c5667acb54834b1

justinstoller:
fwiw, the `pl:tickets` task still exists with this description:
```
rake pl:tickets                                                                       # The ticket generation tasks have moved
```

Magisus:
Yeah I saw that and figured it's probably coming in with the packaging gem, but has been replaced by Winston.

Magisus:
Verified that the `pl:tickets` task has been replaced by a stub and no longer depends on `jira-ruby`.

PR #2152, opened 2019-08-19
(maint) Set locale during container builds

underscorgan:


Iristyle:
Merged as part of #2153

PR #2153, opened 2019-08-20
(maint) use nightlies for containers

underscorgan:
Use nightly repos for resolving the puppet-agent runtime dependency in the docker containers.


Iristyle:
Even if we come to a different solution in the future, this looks necessary for now.

Thanks @underscorgan++ merging!

PR #2155, opened 2019-08-21
(PUP-9970) Move permission corrections to the end of the entrypoints

underscorgan:


Iristyle:
Ok - took a look at pe-puppetserver... it copies in a `puppetdb.conf` during the build.

https://github.com/puppetlabs/pe-puppetserver/blob/kearney/docker/pe-puppetserver/Dockerfile#L69

That feels like a better way to go here - comments @nwolfe / @cthorn42 ?

underscorgan:
Yeah, it looks like puppetserver (instead of puppetserver-standalone) copies in a dummy puppetdb.conf, which would exist before the entrypoints ran. I'm going to ask the person who submitted the issue for more information...

underscorgan:
I was able to confirm that in the puppetserver container, the puppetdb.conf file was ending up owned by root:root. Swapping the order of the use-templates and set-permissions scripts corrected this.

PR #2157, opened 2019-08-26
(maint) Use static addressing in Docker specs

Iristyle:
- We're still seeing failures with the DNS resolver in Docker related
   specs.

   Assign static addresses to the puppet infra, so that when agents
   connect they can have the values set directly in their /etc/hosts
   instead of relying on the flaky Docker DNS resolver.

   Failures are generally `getaddrinfo` failures in resolving
   `puppet.test` from the Ruby code in Puppet.

 - This requires an updated version of the shared spec helpers that
   allows passing an `extra_hosts` value to `run_agent`

### DIFF
--- a/docker/Gemfile
+++ b/docker/Gemfile
@@ -3,6 +3,6 @@ source "https://rubygems.org"
 gem 'rspec'
 gem 'rspec_junit_formatter'
 gem 'pupperware',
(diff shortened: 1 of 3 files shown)


Iristyle:
Not happening... at least not now. Closing!

PR #2164, opened 2019-08-29
(MAINT) Remove --fail-fast

nwolfe:
We already fail fast if there's an error in :before blocks, and we'd
prefer to run all tests if it gets out of the :before block, so we
actually don't want --fail-fast.


nwolfe:
The `getaddrinfo` is really doing its thing today 😡

PR #2165, opened 2019-09-05
(PUP-9933) Use global indirection settings to configure Puppet

tvpartytonight:
This change sets Indirection object settings globally so all threads have
the same settings. It requires that the `set_global_setting` definition
is available in the Indirection class; otherwise, it defaults to the
older, non-threadsafe way of using `setting=` to define various
indirection settings.

### DIFF
--- a/src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
@@ -53,11 +53,17 @@ class Puppet::Server::PuppetConfig
 
     Puppet.settings.use :main, :master, :ssl, :metrics
 
-    Puppet::FileServing::Content.indirection.terminus_class = :file_server
-    Puppet::FileServing::Metadata.indirection.terminus_class = :file_server
-    Puppet::FileBucket::File.indirection.terminus_class = :file
-
-    Puppet::Node.indirection.cache_class = Puppet[:node_cache_terminus]
+    if Puppet::Indirector::Indirection.method_defined?(:set_global_setting)
+      Puppet::FileServing::Content.indirection.set_global_setting(:terminus_class, :file_server)
+      Puppet::FileServing::Metadata.indirection.set_global_setting(:terminus_class, :file_server)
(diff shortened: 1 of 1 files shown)


Magisus:
The travis failure is potentially related to the Context work. It started showing up when the puppet submodule got bumped to a SHA with the first set of those changes, and has persisted despite updating to the fix we merged.

src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
```diff
@@ -53,12 +53,6 @@ def self.initialize_puppet(puppet_config)
 
     Puppet.settings.use :main, :master, :ssl, :metrics
 
-    Puppet::FileServing::Content.indirection.terminus_class = :file_server
-    Puppet::FileServing::Metadata.indirection.terminus_class = :file_server
-    Puppet::FileBucket::File.indirection.terminus_class = :file
-
-    Puppet::Node.indirection.cache_class = Puppet[:node_cache_terminus]
-
```
Magisus: Are we confident that these are totally unused prior to the first request coming in? And/or that the defaults are valid?

tvpartytonight: I'm confident that this fixes some of the test failures seen in the match puppet [PR](https://github.com/puppetlabs/puppet/pull/7685), but I can't be sure there's not some case where these values must be set before any requests start coming in. I'd rather they not be duplicated though...maybe, after we fix the pipeline to be more stable, we can merge this and then see if there is anything else caught in testing?

Magisus: Works for me

justinstoller:
I'm concerned that we'd need to do this for some unknown subset (maybe all?) of the methods in the Master [interface](https://github.com/puppetlabs/puppetserver/blob/master/src/java/com/puppetlabs/puppetserver/JRubyPuppet.java). At which point I think it would be a lower maintenance burden to be able to set this values one at initialization rather than set them during the request. Do you have an idea of how many of those method would need this vs how hard it would be to set these once when we initialize the server?

You have thoughts on this @pcarlisle ?

PR #2166, opened 2019-09-06
(SERVER-2423) Add facter.jar to system classpath from cli

justinstoller:
When we moved to supporting native Facter we began optionally
adding the facter.jar in clojure at runtime if it was available.

This is because Java only allows a native library to be loaded once per
JVM and attempting to load the library from different classloaders will
cause an error.

This approach was possible because the application classloader was a
global, mutable object.

See SERVER-718 & FACT-965 for the original work and rationale.

In Java 9+, the global classloaders are no longer mutable so the
classloader that the clojure code adds the facter.jar to is never seen
by any JRuby instance. This causes starting Puppet Server with more than
one JRuby instance, or reloading Puppet Server with only one JRuby
instance, or whenever Puppet Server tries to refresh the one JRuby
instance because of max-requests-per-instance, to fail with an
UnsatisfiedLinkError because the second JRuby instance will have a
different classloader than the first one.

Now that Puppet Server requires versions of the AIO agent that only
bring in native Facter and its jar, we no longer need the optional
loading of native Facter's jar.

Since we can now rely on facter.jar being available we can simply add
it to the classpath when invoking Java, which is the only way to put it
in the global classloader in Java 9+.

Notes:
  - In dev environments we still load Ruby Facter from a submodule and
    do not use native Facter or the service scripts we change in this
    patch.
  - Some Java implementations may warn if there's a classpath entry
    that does not exist - if we move the facter.jar w/o updating our
    classpath invocation. However, OpenJDK versions 8 & 11 (those
    available on my test machine of RHEL8) do not warn on non-existent
    classpath entries.
  - Because we no longer search the ruby load path for the facter.jar if
    it moves and our classpath invocation isn't changed loading Facter
    will fail, and Puppet Server will no longer start. _Unless_ the jar
    is moved to `/opt/puppetlabs/server/data/puppetserver/jars/` where
    we encourage users to put their own additional jars, or it is
    vendored w/in the uberjar `puppet-server-release.jar`.

### DIFF
--- a/resources/ext/cli_defaults/cli-defaults.sh.erb
+++ b/resources/ext/cli_defaults/cli-defaults.sh.erb
@@ -4,4 +4,4 @@ if [ -n "$JRUBY_JAR" ]; then
   echo "Warning: the JRUBY_JAR setting is no longer needed and will be ignored." 1>&2
 fi
 
-CLASSPATH="${CLASSPATH}:/opt/puppetlabs/server/data/<%= EZBake::Config[:real_name] %>/jars/*"
+CLASSPATH="${CLASSPATH}:/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/facter.jar:/opt/puppetlabs/server/data/<%= EZBake::Config[:real_name] %>/jars/*"
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
@@ -2,7 +2,6 @@
   (:require [clojure.tools.logging :as log]
             [me.raynes.fs :as fs]
             [schema.core :as schema]
-            [puppetlabs.kitchensink.classpath :as ks-classpath]
             [puppetlabs.kitchensink.core :as ks]
             [puppetlabs.services.jruby.jruby-puppet-schemas :as jruby-puppet-schemas]
             [puppetlabs.services.jruby-pool-manager.jruby-schemas :as jruby-schemas]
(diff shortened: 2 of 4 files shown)


resources/ext/cli_defaults/cli-defaults.sh.erb
```diff
@@ -4,4 +4,4 @@ if [ -n "$JRUBY_JAR" ]; then
   echo "Warning: the JRUBY_JAR setting is no longer needed and will be ignored." 1>&2
 fi
 
-CLASSPATH="${CLASSPATH}:/opt/puppetlabs/server/data/<%= EZBake::Config[:real_name] %>/jars/*"
+CLASSPATH="${CLASSPATH}:/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/facter.jar:/opt/puppetlabs/server/data/<%= EZBake::Config[:real_name] %>/jars/*"
```
justinstoller: @puppetlabs/release-engineering is there an ezbake variable that I should be interpolating here to get Facter's jar path?

justinstoller:
Note: I think removes the necessity of adding the facter.jar to the classpath from w/in Clojure code in any version of the JVM and allows us to remove our dependency on kitchensink.classpath and dynapath (which is producing an ugly warning on Java 11). I'm currently investigating that.

resources/ext/cli_defaults/cli-defaults.sh.erb
mwaggett: Not that I know of? I think this is all the ezbake variables: https://github.com/puppetlabs/ezbake/blob/master/resources/puppetlabs/lein-ezbake/staging-templates/ezbake.rb.mustache ?

justinstoller: danke

Magisus:
I kinda want to see these tests pass. The fix needed in Puppet has been merged and should promote soon.

Magisus:
Actually, these tests don't use `facter.jar`, since they load Facter from the submodule. Merging.

PR #2167, opened 2019-09-09
[dnm] (SERVER-2616) Do not force closed client connections

justinstoller:
When Server's http client wrapper was initially written Puppet's client
forced all of its connections closed, however, it was very shortly later
re-written to stop doing that w/o issue (see
puppet@789e00bb8255548dbad71ba1462a16afec624cc8).

### DIFF
--- a/src/ruby/puppetserver-lib/puppet/server/http_client.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/http_client.rb
@@ -94,8 +94,6 @@ class Puppet::Server::HttpClient
       headers["Authorization"] = authorization
     end
 
-    # Ensure multiple requests are not made on the same connection
-    headers["Connection"] = "close"
     request_options = RequestOptions.new(build_url(url))
     if options[:metric_id]
       request_options.set_metric_id(options[:metric_id])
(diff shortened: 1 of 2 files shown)


justinstoller:
Pinged @joshcooper for review, not sure I got the geneology correct on this, but the referenced commit seems to be where our pool comes from.

joshcooper:
I realized after we talked about this that it's not as simple as I was thinking. Does puppetserver's apache httpclient actually reuse the connection? If not, then I don't think removing the `Connection: close` header will improve anything.

If apache httpclient does reuse the connection, is there some guarantee that the response body is always drained prior to being reused? For example, if puppetserver submits a request, but say it gets back an HTTP 500 from puppetdb. If puppetserver does not drain the response body, and later reuses the connection, then it will read data from the previous request. Usually the httpclient library provides a way to drain/ignore the rest of the response to avoid that problem, but I'm not sure if apache httpclient does that, or puppetserver needs to?

justinstoller:
Threw a "do not merge" label on this and put the ticket into blocked until we get some test cases around Josh's concerns. This will _not_ be going into the Server 6.6.0 release.

Magisus:

PR #2168, opened 2019-09-09
(maint) Preserve vendored gems at container build time

underscorgan:
We've realized most of the data in
/opt/puppetlabs/server/data/puppetserver is runtime data that should be
preserved as a volume during container runs. However,
`vendored-jruby-gems` only ever contains files installed at package
build time. Ideally these will move out of the data directory, but in
the interim work around it using the existing /var/tmp + entrypoint
pattern we have.

### DIFF
--- a/docker/puppetserver-standalone/docker-entrypoint.d/20-use-templates-initially.sh
+++ b/docker/puppetserver-standalone/docker-entrypoint.d/20-use-templates-initially.sh
@@ -14,13 +14,13 @@ cd /
 
 # if the data dir is empty, copy over everything from build time
 if [ ! "$(ls -A /opt/puppetlabs/server/data/puppetserver)" ]; then
-  cp -a /var/tmp/puppetserver/* /opt/puppetlabs/server/data/puppetserver/
+  cp -R /var/tmp/puppetserver/* /opt/puppetlabs/server/data/puppetserver/
   # remove the tmp dir so we only run this on first runs of new containers
   rm -rf /var/tmp/puppetserver/vendored-jruby-gems
 elif [ -d /var/tmp/puppetserver/vendored-jruby-gems ]; then
   # clean up existing vendored gems
   rm -rf /opt/puppetlabs/server/data/puppetserver/vendored-jruby-gems
-  cp -a /var/tmp/puppetserver/vendored-jruby-gems /opt/puppetlabs/server/data/puppetserver/
+  cp -R /var/tmp/puppetserver/vendored-jruby-gems /opt/puppetlabs/server/data/puppetserver/
   # remove the tmp dir so we only run this on first runs of new containers
   rm -rf /var/tmp/puppetserver/vendored-jruby-gems
 fi
--- a/docker/puppetserver-standalone/docker-entrypoint.d/20-use-templates-initially.sh
+++ b/docker/puppetserver-standalone/docker-entrypoint.d/20-use-templates-initially.sh
@@ -14,10 +14,13 @@ cd /
 
 # if the data dir is empty, copy over everything from build time
 if [ ! "$(ls -A /opt/puppetlabs/server/data/puppetserver)" ]; then
+  # https://github.com/moby/moby/issues/39892
+  echo "Loading new empty VOLUME with default /opt/puppetlabs/server/data/puppetserver content (LCOW moby 39892 workaround)"
   cp -R /var/tmp/puppetserver/* /opt/puppetlabs/server/data/puppetserver/
   # remove the tmp dir so we only run this on first runs of new containers
   rm -rf /var/tmp/puppetserver/vendored-jruby-gems
 elif [ -d /var/tmp/puppetserver/vendored-jruby-gems ]; then
+  echo "Upgrading /opt/puppetlabs/server/data/puppetserver/vendored-jruby-gems"
   # clean up existing vendored gems
   rm -rf /opt/puppetlabs/server/data/puppetserver/vendored-jruby-gems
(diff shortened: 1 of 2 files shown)


docker/puppetserver-standalone/docker-entrypoint.d/20-use-templates-initially.sh
```diff
@@ -11,3 +11,11 @@ do
     test -f "$TEMPLATES/$f" && cp -np "$TEMPLATES/$f" .
 done
 cd /
+
+if [ -d /var/tmp/puppetserver/vendored-jruby-gems ]; then
```
Iristyle: I think we might want to copy the entire `/opt/puppetlabs/server/data/puppetserver` directory as that seems more future proof?

Also seems like the best generic way to correctly handle the VOLUME on LCOW given https://github.com/moby/moby/issues/39892

Thoughts?

Iristyle: Addendum - as the code is written, that approach won't work for upgrades.

But maybe we could check if the `/opt/puppetlabs/server/data/puppetserver/` is empty and copy it all over to handle the LCOW scenario?

underscorgan: I'll take a look and see if I can come up with something to gracefully handle additional contents in that directory. Right now the only things in `data/puppetserver` other than `vendored-jruby-gems` are empty directories, but we would likely want to create those if they don't exist at runtime to work around the windows issue.

Mostly want to make sure we aren't cleaning up anything other than `vendored-jruby-gems` since that's the only directory there I wouldn't expect to persist on container upgrades (since we'd always want the gems to match what was packaged).

underscorgan: Just saw the addendum, might make sense to copy everything over if `data/puppetserver` is empty, and otherwise cleanup and copy over `vendored-jruby-gems` ? That doesn't feel terrible

Iristyle: FYI - this is what I tentatively have over in `pe-puppetserver`: https://github.com/puppetlabs/pe-puppetserver/pull/381/commits/bbdbbef3996c95e20af0194ff99485d2f696d2e3

underscorgan: I had something very similar

docker/puppetserver-standalone/docker-entrypoint.d/20-use-templates-initially.sh
```diff
@@ -11,3 +11,16 @@ do
     test -f "$TEMPLATES/$f" && cp -np "$TEMPLATES/$f" .
 done
 cd /
+
+# if the data dir is empty, copy over everything from build time
+if [ ! "$(ls -A /opt/puppetlabs/server/data/puppetserver)" ]; then
+  cp -a /var/tmp/puppetserver/* /opt/puppetlabs/server/data/puppetserver/
```
Iristyle: FWIW, `cp -a` doesn't seem to work under LCOW in the PR I put up. I had to change to `cp -R` only

docker/puppetserver-standalone/docker-entrypoint.d/20-use-templates-initially.sh
```diff
@@ -11,3 +11,16 @@ do
     test -f "$TEMPLATES/$f" && cp -np "$TEMPLATES/$f" .
 done
 cd /
+
+# if the data dir is empty, copy over everything from build time
+if [ ! "$(ls -A /opt/puppetlabs/server/data/puppetserver)" ]; then
+  cp -R /var/tmp/puppetserver/* /opt/puppetlabs/server/data/puppetserver/
```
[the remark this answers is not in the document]
Iristyle: Also, I don't think the condition here is quite correct either:

```shell
#!/bin/sh

set -e

if [ ! "$(ls -A ./exists)" ]; then
  echo "1"
fi

if [ "$(ls -A ./exists)" ]; then
  echo "2"
fi

if [ ! "$(ls -A ./notexists)" ]; then
  echo "3"
fi

if [ "$(ls -A ./notexists)" ]; then
  echo "4"
fi
```

```
> ls -rtaFl
total 8
-rwxr-xr-x@  1 Iristyle  staff  220 Sep 11 11:50 test.sh*
drwxr-xr-x+ 16 Iristyle  staff  512 Sep 11 11:54 ../
drwxr-xr-x+  2 Iristyle  staff   64 Sep 11 11:54 exists/
drwxr-xr-x+  4 Iristyle  staff  128 Sep 11 11:54 ./

> ./test.sh
1
ls: ./notexists: No such file or directory
3
ls: ./notexists: No such file or directory
```

underscorgan:
@Iristyle added the echos!

docker/puppetserver-standalone/docker-entrypoint.d/20-use-templates-initially.sh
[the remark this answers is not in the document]
underscorgan: I don't think we will ever have a situation where the directory doesn't exist, just whether or not there's anything in the directory so I think that should work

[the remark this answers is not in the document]
Iristyle: Yeah, sorry... had updated my comment there b/c I misinterpreted.

We match on:

* 1 (dir exists but empty)
* 3 (dir doesn't exist) 

But not:

* 5 (dir exists but has contents)


Which is exactly what we want.

PR #2154, opened 2019-08-20
(SERVER-2193) Add thread-dump endpoint to admin API

Sharpie:
This commit adds a `jruby-pool/thread-dump` endpoint to the admin API.
This endpoint can be used to retrieve a backtrace from each JRuby
interpreter registered to the pool when `-Djruby.management.enabled=true`
has been set in JAVA_ARGS.

PR #2169, opened 2019-09-10
(maint) Bring in updated clj-parent

justinstoller:


justinstoller:
Should I also manually update the puppet submodule?

Sharpie:
A submodule update would be a good idea as it should unblock testing on other open PRs.

Magisus:
The agent is almost done promoting, PEZ seems gummed up, give me a second to look into it. It'd be easier to just run the submodule update job once that passes.

justinstoller:
I updated the submodule, but it only. See the commit message about preferring the automated bump for acceptance tests. Happy to remove the commit if the agent will promote some time soon.

Magisus:
Expecting a promotion in the next 20 minutes. We'll see who gets there first, Travis or Jenkins :)

justinstoller:
Mmm... neither travis nor that agent promotion seemed to have much haste about them. It looks like there's some new failures with the clj-parent bump wrt client http metrics w/ FIPS enabled. Are those expected?

Magisus:
Not expected by me 🤷‍♀

Magisus:
Got the submodule and agent ref updated via the Jenkins job.

Magisus:
I'm seeing the failures locally without the clj-parent update but with the submodule bump.

justinstoller:
Which ever goes in first between this and https://github.com/puppetlabs/puppetserver/pull/2154 is going to cause the other to have to rebase and I won't be around if its me. I'm just going to close this and maybe we can update to 4.2.0 in the other PR?

Magisus:
Sounds good, I'll push that if Charlie doesn't get there first.

PR #2171, opened 2019-09-11
(maint) Allow using Puppet code prior to 0f370f0fb

justinstoller:
It isn't recommended and few do it, but we should generally allow
running agents and masters of similar versions, as it at least makes
upgrade scenarios easier for our users and git bisects and code
promotions easier on us.

### DIFF
--- a/src/ruby/puppetserver-lib/puppet/server/key_recorder.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/key_recorder.rb
@@ -1,5 +1,27 @@
 require 'puppet/server'
-require 'puppet/pops/lookup/key_recorder'
+
+# Puppet::Pops::Lookup::KeyRecorder was added in Puppet 6.8 with
+# this "null" behavior. If running with agent code < 6.8 define our
+# own null recorder.
+begin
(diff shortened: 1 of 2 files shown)


Magisus:
Are we planning to do anything similar around `concurrent`? That's also blowing up similarly right now if you get a package that doesn't include the gem.

justinstoller:
I did a bisect today of Puppet code were I wanted to keep the Server code the same and ended up doing this.... so I figured I'd clean it up and PR it to see if folks thought it was valuable.

justinstoller:
re: concurrent. I don't know, I would be opposed to making it backwards compatible if we can figure out how.

src/ruby/puppetserver-lib/puppet/server/key_recorder.rb
```diff
@@ -1,5 +1,27 @@
 require 'puppet/server'
-require 'puppet/pops/lookup/key_recorder'
+
+# Puppet::Pops::Lookup::KeyRecorder was added in Puppet 6.8 with
+# this "null" behavior. If running with agent code > 6.8 define our
```
Magisus: I think this should be `< 6.8`.

PR #2172, opened 2019-09-11
(maint) Bump ezbake to 2.1.3 for Buster support

justinstoller:


Magisus:
This will also need a ci-job-configs update to actually start building it. Do we only want to build on Deb 10 for 6.9+, or should it work for all branches?

justinstoller:
Yes, we'll need to turn it on in CI as well. There may be additional work that falls out of this because Buster ships with Java 11, though I think we'll still try to pull in java 8 and I don't know if Buster even ships with a Java 8.

I'd only target Server 6.6 and later, regardless of whether or not we require a java 8 package to be available for installation, since that's when Java 11 support should ship.

justinstoller:
RE might know more about whether or not we can (or do) just depend on a java virtual package.

Magisus:
I tried installing a package built with this, and confirmed that java 8 is not installable by default (at least on our pooler machines). It seems like so long as we know Java 11 works, we should make it so that the package will install with that version, at least on this OS. Reaching out to RE to find out our options there.

justinstoller:
This isn't happening til October.

PR #2179, opened 2019-09-19
(maint) Allow using older Puppet's Ruby code

justinstoller:


Magisus:
Are expecting FIPS to be failing again? Or is Justin's branch out of date? Those look like the failures that got fixed a week ago or so.

Magisus:
I verified that this solves the problem.

PR #2181, opened 2019-09-23
[Docker container] Add 2 missing directories to fix permissions for

jay7x:
Following directories are owned by `puppet` user in container. If UID/GID is changed, those directories permissions must be updated too.

### DIFF
--- a/docker/puppetserver-standalone/docker-entrypoint.d/30-set-permissions.sh
+++ b/docker/puppetserver-standalone/docker-entrypoint.d/30-set-permissions.sh
@@ -2,3 +2,5 @@
 
 chown -R puppet:puppet /etc/puppetlabs/puppet/
 chown -R puppet:puppet /opt/puppetlabs/server/data/puppetserver/
+chown -R puppet:puppet /etc/puppetlabs/puppetserver/
+chown -R puppet:puppet /var/log/puppetlabs/puppetserver/


Iristyle:
I think we probably want to do the same in https://github.com/puppetlabs/pe-puppetserver/blob/kearney/docker/pe-puppetserver/docker-entrypoint.d/30-set-permissions.sh#L3 @nwolfe / @cthorn42

underscorgan (approved):

Iristyle:
@jay7x have you been configuring a different user to match one on the host system?

I'd be interested to hear some more details about your setup if you're able to share. Thanks!

jay7x:
@Iristyle yes, I was trying to setup dedicated puppet CA container with predictable UID/GID on host system. So I crafted passwd & group files and mapped them to the container's /etc. Actually just after that I realised "to be predictable" is good but not really required :-D
And then I hit into CPR-718..

Iristyle:
I'm working on making some of the internal paths consistent with pe-puppetdb and therefore some of these locations may change. I'll unblock this PR when they're merged in.

Iristyle:
@jay7x we can merge this for now if you rebase / update your PR. We still want to do the path changes I mentioned, but that won't be happening for a bit since we don't have the bandwidth, and this is still a useful change I think?

Magisus:

justinstoller:
We lost the thread on this. My apologies. I've created https://github.com/puppetlabs/puppetserver/pull/2615 to provide the request but rebased onto main and done by someone who has signed the CLA.

PR #2183, opened 2019-09-24
(CPR-718) Fix IP determination when running in docker with consul

underscorgan:


docker/puppetserver-standalone/Dockerfile
```diff
@@ -47,6 +47,9 @@ ENV PUPPETSERVER_MAX_ACTIVE_INSTANCES=1
 ENV PUPPETSERVER_MAX_REQUESTS_PER_INSTANCE=0
 ENV CA_ENABLED=true
 ENV CA_ALLOW_SUBJECT_ALT_NAMES=false
+ENV CONSUL_ENABLED=false
```
Iristyle: Should we do this for the other projects too?

underscorgan: yeah, I'll take a look at that

underscorgan:
see also https://github.com/puppetlabs/puppetdb/pull/3079

PR #2184, opened 2019-09-24
(PUP-9933) Configure storeconfigs directly in puppetserver initialization

tvpartytonight:


src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
```diff
@@ -63,6 +63,17 @@ def self.initialize_puppet(puppet_config)
...
       Puppet::Node.indirection.set_global_setting(:cache_class, Puppet[:node_cache_terminus])
+
+      # Instead of relying on Puppet defaults to use a hook to configure the
+      # necessary storeconfigs settings, set them directly here.
+      if Puppet[:storeconfigs]
+        Puppet::Resource::Catalog.indirection.set_global_setting(:cache_class, :store_configs)
+        Puppet.settings.override_default(:catalog_cache_terminus, :store_configs)
+        Puppet::Node::Facts.indirection.set_global_setting(:cache_class, :store_configs)
+        Puppet::Resource.indirection.set_global_setting(:terminus_class, :store_configs)
+        # Set the facts indirection terminus globally
```
Magisus: This comment seems maybe unnecessary, aren't these all being set globally?

src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
```diff
@@ -63,6 +63,17 @@ def self.initialize_puppet(puppet_config)
       Puppet::FileServing::Metadata.indirection.set_global_setting(:terminus_class, :file_server)
       Puppet::FileBucket::File.indirection.set_global_setting(:terminus_class, :file)
       Puppet::Node.indirection.set_global_setting(:cache_class, Puppet[:node_cache_terminus])
+
+      # Instead of relying on Puppet defaults to use a hook to configure the
+      # necessary storeconfigs settings, set them directly here.
+      if Puppet[:storeconfigs]
```
justinstoller: These aren't Puppet Server specific and they need to be set on the agent or puppet subcommand as well. But they will be, because those non-Puppet Server environments run in a single thread, is that right?

I think previously, I've been thinking of this as Puppet being threadsafe itself and Puppet Server using it's thread-aware constructs when it needs to set something.

But I think this moves Puppet to be not-so-thread-aware and Puppet Server configuring Puppet to be threadsafe when running within Puppet Server. Is that accurate?

If so, I think this should probably be in Puppet proper, either [here](https://github.com/puppetlabs/puppet/blob/master/lib/puppet.rb#L147), [here](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/settings.rb#L370), or [here](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/settings.rb#L298) depending on whether it's Puppet as a whole or the settings subsystem's responsibility for this, and if it's the settings subsystem's responsibility where in its lifecycle those hooks are triggered.

tvpartytonight: For reference, the accompanying puppet revert is [here](https://github.com/puppetlabs/puppet/pull/7735). To answer your first question, the hook is currently using the same `set_global_setting` used here; the revert in the puppet PR would go back to the old behavior of setting it only for a single thread.
I made the decision to move this logic to puppetserver because I didn't see how to make this work without embedding some `if master` specific logic inside puppet itself. My previous attempt was to add a hook for the `storeconfigs_backed`, which would look something like this:
```
proc do |value|
  if Puppet.settings.preferred_run_mode == :master
    Puppet::Node::Facts.indirection.set_global_setting(:terminus_class, Puppet[:storeconfigs_backend])
  end
end
```
Even if we moved this out of the hook and elsewhere, it still seems like we are changing behavior based on the `preferred_run_mode`. It seemed messier to keep that in Puppet itself, and that's why I moved the fix out of Puppet and into server's initialization.

justinstoller:
I think this was superseded by https://github.com/puppetlabs/puppet/pull/7738

PR #2185, opened 2019-09-24
(SERVER-2628) Debian 10 support

Magisus:


Magisus:
Currently, puppetdb tests will be skipped on Deb10. Once PDB gets packages up, we can enable those [here](https://github.com/puppetlabs/puppetserver/blob/master/acceptance/lib/helper.rb#L63-L72).

justinstoller:
Travis doesn't know about packaging changes.

PR #2188, opened 2019-09-27
Updates for puppetserver container

underscorgan:
This PR includes some fairly substantial changes to the puppetserver containers. 

# tl;dr
* puppetserver-standalone container is going away
* the versions are changing

build type | current tag | new tag
---------- | ----------- | --------
from source | puppet/puppetserver:6.7.0 | ~puppet/puppetserver:6.7~ n/a
from source (latest) | puppet/puppetserver:latest | puppet/puppetserver:edge
from package | n/a | puppet/puppetserver:6.7.0
from package (latest) | n/a | puppet/puppetserver:latest

# Updates to container versioning

For a while now, our puppetserver containers have included a puppetserver package built from source and versioned based on the most recent tag to the repo. While we still think building from source provides value to our users, it's become clear that they also need a way to pin to a specific, released version of puppetserver and count on that container not being updated. To address this, we're changing the versioning scheme for our container builds. 

~When we build images from source, those images will be versioned with X.Y versions based on the latest tag  on master. So, for example, the current image versioned `puppet/puppetserver:6.6.0` would move to `puppet/puppetserver:6.6`. This tag will continue to have rolling updates until the next X or Y release. If you want to follow whatever the latest version of the image from source is, you will want to pin to `puppet/puppetserver:edge`.~

When we build images from source, those images will be tagged 'edge'. We considered having something like `puppet/puppetserver:6.6-source` for source builds, but that will likely end up as approximately the same image as `puppet/puppetserver:6.7.0` since it will continue updating until the repo is tagged. So, for our source/nightly(ish) builds we're just going to publish those as `puppet/puppetserver:edge`. This tag will be updated with each commit to the master branch.

We will also start building and shipping images when puppetserver is shipped publicly. These images will be tagged with an X.Y.Z version that will match the version of puppetserver installed on that image. This tag will not receive any updates. If you want to follow the latest released version of puppetserver, you will want to pin to `puppet/puppetserver:latest`.

# Remove the `puppetserver-standalone` image

Initially the `puppetserver-standalone` image was intended as an easy way to get puppetserver up and running without needing to set up postgresql and puppetdb. While this is still a completely valid usecase, as the puppetserver container has gone through more development and been made more configurable, it is now possible to get the `puppet/puppetserver` image running without postgresql/puppetdb by setting the `USE_PUPPETDB` environment variable in your `docker run` command:
```
docker run --env USE_PUPPETDB=false puppet/puppetserver
```

# The `puppetserver-base` image

The puppetserver-base image is not going to be a published image. It does not actually contain a puppetserver package, it is just intended as an abstraction so that we can only maintain one copy of the configuration shared between the image built from source and the image built from released packages.


nwolfe:
The variables to disable postgres/puppetdb seem a bit confusing, but I don't know any better really 😄 

We say: 
> docker run --env PUPPET_STORCONFIGS=false --env PUPPET_REPORTS=log puppet/puppetserver

But I'm wondering if we'll eventually want variables like `USE_PUPPETDB` or `USE_POSTGRES` ?

I assume these variables above control other things that users might want to adjust?

Or maybe we look for `USE_PUPPETDB`, and then set `PUPPET_STORECONFIGS` and `PUPPET_REPORTS` based on that value, either in the Dockerfile or an entrypoint script?

Also there's a typo in the `PUPPET_STORCONFIGS=false` variable in the snippet above.

nwolfe:
Regarding the `edge` tag..... that took me a minute or two to realize the key word there is "from source".  

I was confused why we were introducing our own `latest` tag.

Maybe call out that we're providing both a from-source and from-packages image?

Or maybe just a summary at the top/bottom of images we will be providing, and how they differ would cover that and more?

underscorgan:
ah, yeah, I like the idea of a `USE_PUPPETDB` env variable, will look into that. I think it might just set the other environment variables, but seems alright :)

nwolfe:
Ah perfect, that TL;DR table is great!

underscorgan:
Added the `USE_PUPPETDB` environment variable to make it easier to run the new image like the `puppet/puppetserver-standalone` image, going to send this out to the mailing lists now.

Iristyle:
Ok - finally got this merged @underscorgan!

I had a bit of a time understanding some of the changes b/c how they were split up, so I reworked them a little to make sure the Azure changes were made at the same time as Makefile changes (to keep each commit green) and filled in commit messages as I understood what was happening. Vetted them all in https://github.com/puppetlabs/puppetserver/pull/2239

PR #2190, opened 2019-09-30
(maint) Add SERVER-115 release note

Magisus:


documentation/release_notes.markdown
```diff
@@ -18,6 +18,9 @@ Released 1 October 2019
 
 - Puppet Server packages are now available for Debian 10. These packages require Java 11 to be installed, rather than Java 8. [SERVER-2613](https://tickets.puppetlabs.com/browse/SERVER-2613)
 
+### Resolved Issues
+
+- Puppet Server now synchronizes write access to the CRL, so that each revoke request must update it in serial, preventing corruption of the CRL due to competing requests. [SERVER-115](https://tickets.puppetlabs.com/browse/SERVER-115)
```
jbondpdx: When you say "update it in serial," do you mean that it updates the CRL in order?

Magisus:
No, sorry, I mean "each request updates it one after the other", as in "as opposed to concurrently". Got a rewording recommendation?

documentation/release_notes.markdown
jbondpdx: How about:

Puppet Server now synchronizes write access to the CRL, so that each revoke request updates the CRL in succession, instead of concurrently. This prevents corruption of the CRL due to competing requests.

PR #2192, opened 2019-10-01
(maint) Configure more like pe puppetserver

Iristyle:
Store things in puppetserver at the same place as pe-puppetserver so upgrades are easy peasy

There's a ton more work to do here


Iristyle:
Yup @nwolfe... this whole container needs a refactor. The no `ssl.sh` thing is why I put on the brakes and tossed this to the end of the stack for now.

Iristyle:
We need to eventually come back to this... but it's not happening in the short-term.

PR #2193, opened 2019-10-02
(FM-8336) Capture and expose attribute ordering from transport schema

DavidS:
When exposing attributes in the UI, they currently need to be alphabetised by their name, since the transmission through APIs does not preserve key order of the connection_info hash.

To facilitate showing attributes in the order defined by the module author, this change embeds the key-insertion order from the TypeDefinition into the API response.

### DIFF
--- a/src/ruby/puppetserver-lib/puppet/server/master.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/master.rb
@@ -97,7 +97,16 @@ class Puppet::Server::Master
 
   def getTransportInfoForEnvironment(env)
     require 'puppet/resource_api/transport'
-    Puppet::ResourceApi::Transport.list_all_transports(env).values.map(&:definition)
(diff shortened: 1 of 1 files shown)


justinstoller:
Just to be clear, this isn't something you're targeting for Lovejoy, right?

DavidS:
@justinstoller yes, this is for lovejoy.

Magisus:
Well technically that ship has sailed... the last FOSS release going into PE had its code freeze a week ago and shipped yesterday. Is this urgent enough that I should escalate it?

justinstoller:
Seems like all the ships are coming back into port. Maybe we can ship this still. But I think we need to figure out if the other NetDev work needs Server updates, check in on the FIPS work, and figure out what the agent's schedule is for their release before we commit to a new 6.7.z.

DavidS:
Replaced by the same fix but better in Resource API: https://github.com/puppetlabs/puppet-resource_api/pull/238

PR #2197, opened 2019-10-04
(SERVER-2654) Update for recent FIPS work

justinstoller:


Magisus:
This all makes sense. Would you mind adding a comment to the SSLv3 test removal commit that explains what actually causes the support to be removed? Just deleting tests doesn't imply any actual behavior change to me.

justinstoller:
updated

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -806,22 +807,23 @@
...
+          settings (testutils/master-settings tmp-confdir)
+          ca-settings (testutils/ca-settings (str tmp-confdir "/ssl/ca"))]
 
-    (retrieve-ca-cert! (:cacert ca-settings) (:localcacert settings))
+      (retrieve-ca-cert! (:cacert ca-settings) (:localcacert settings))
 
-    (testing "should throw an error message with too short keylength"
-      (is (thrown?
-           InvalidParameterException
-           (initialize-master-ssl! (assoc settings :keylength 128) "master" ca-settings))))
```
robdaemon: we probably still want a guard in `initialize-master-ssl!` to not allow keys < 1024 bits in FIPS mode, because if it ever happens, the process will effectively be dead, but it won't actually exit because it'll trip the error handler in bc-fips.

justinstoller: I pulled that out thinking that to do it well we actually should add the keylength verification to a different part of our code (and in our puppetsever-ca cli tool which is recommended over the `initialize-master-ssl` method), that those changes aren't something we should necessarily do in a z, and while the UX is bad it is pretty unlikely (someone would have to install FIPS and then manually configure the keylength to be less than 1024?

My thinking was that we could clean that up in the next 'Y' release: https://tickets.puppetlabs.com/browse/SERVER-2655

robdaemon: maybe we do that in jvm-ssl-utils, where the calls into the bc provider happen?

justinstoller:
@puppetlabs/pupperware Is there something we should be doing regarding the docker tests? Two out of the three container related tests are failing now, but I'm unclear if we did anything to cause that?

nwolfe:
@justinstoller Looks like a legit failure:
- https://dev.azure.com/puppetlabs/puppetserver/_build/results?buildId=4968&view=logs&jobId=12f1170f-54f2-53f3-20dd-22fc7dff55f9&taskId=6adde9b7-7a2d-5c6f-65e9-22367a927d98&lineStart=1544&lineEnd=1545&colStart=1&colEnd=1
- https://travis-ci.org/puppetlabs/puppetserver/jobs/594784215#L1946

I suppose the same changes should be made around the lein call in the Dockerfile: https://github.com/puppetlabs/puppetserver/blob/master/docker/puppetserver-standalone/Dockerfile#L24

PR #2205, opened 2019-10-16
(Docker container) put DNS_ALT_NAMES in [agent] section when CA_ENABLED != true

jtriley:
A compiler-only instance's cert generation takes place in `80-ca.sh` via a noop puppet agent run, however, `DNS_ALT_NAMES` currently gets set in the `[master]` section of the config which isn't visible to the agent. This patch sets `DNS_ALT_NAMES` within the agent section of the config for compiler-only instances (ie when `CA_ENABLED != true`) so that the initial noop agent run includes `DNS_ALT_NAMES` when requesting the cert.


underscorgan:
@jtriley awesome, thank you for this submission! I'm guessing the failure is just a transient, I'll rekick it and once that passes will get this merged in 🎉

jtriley:
@underscorgan my pleasure, thank you!!

docker/puppetserver-standalone/docker-entrypoint.d/70-set-dns-alt-names.sh
```diff
@@ -4,11 +4,17 @@
 # setting will only have an effect when the container is started without
 # an existing certificate on the /etc/puppetlabs/puppet volume
 if test -n "${DNS_ALT_NAMES}"; then
+    config_section="master"
```
Iristyle: Could we have just set this in `main` instead of branching the logic? I may need to brush up on my config file processing, but I think that should then be used by either `master` or `agent` I believe.

jtriley: Yes I originally wrote it to just always use `[main]` but wasn't sure if that would have undesired side effects. I can submit another PR if you'd prefer that instead.

PR #2208, opened 2019-10-17
Merge 6.3.x -> master

Magisus:


Magisus:
Oh looks like these certs expired too. Great.

PR #2209, opened 2019-10-18
(maint) Update expired test certs

Magisus:
This commit also adds a script for doing this regen next time.


dev-resources/regen_certs.rb
```diff
@@ -0,0 +1,119 @@
...
+    end
+
+    def self.bundle(*items)
+      items.map {|i| EXPLANATORY_TEXT + i.to_pem }.join("\n")
+    end
+  end
+end
+
+ca_key = PuppetSpec::SSL.create_private_key
+ca_cert = PuppetSpec::SSL.self_signed_ca(ca_key, "/CN=Puppet CA: swanson.hsd1.or.comcast.net")
```
tvpartytonight: swanson?
swanson

Magisus: Yeah I have no idea where that name came from xD But I left it just in case there was anything relying on it.

PR #2210, opened 2019-10-19
(maint) Update healthcheck timeouts

Iristyle:
- Set reasonable upper bounds for various stages in Azure to abort
  jobs quickly when Docker misbehaves

- Based on timings of trhe open source stack under LCOW, this container
  typically starts in about 200s. Give the healthcheck an additional
  100s buffer, allowing it to take up to 5m to be healthy, rather than 6m


Iristyle:
Merging - failures are unrelated.

PR #2212, opened 2019-10-29
(PUP-10014) Add deep_merge to gem list

jtappa:


joshcooper:
I dunno if server team cares about PUP tickets in commits? Otherwise I'm 👍

Separately, it would be good to figure out how to introspect puppet gemspec and install the dependencies, instead of puppetserver having a duplicate list.

jtappa:
Yeah, I mostly added it so that Jira would hopefully catch it and put it in the git section of the ticket. Doesn't look like the server CI requires it like the puppet repo.

joshcooper:
👍 to adding deep_merge as a duplicate (following the other examples). Here's a list of gems from currently released puppet/puppetserver (doesn't include the deep_merge in this PR):

```
# ls -1 /opt/puppetlabs/server/data/puppetserver/vendored-jruby-gems/gems/
concurrent-ruby-1.1.5
fast_gettext-1.1.2
gettext-3.2.2
hocon-1.2.5
locale-2.1.2
semantic_puppet-1.0.2
text-1.3.1
# ls -1 /opt/puppetlabs/puppet/lib/ruby/vendor_gems/gems/
deep_merge-1.0.1
fast_gettext-1.1.2
hiera-eyaml-3.0.0
highline-1.6.21
hocon-1.2.5
locale-2.1.2
multi_json-1.13.1
optimist-3.0.0
puppet-resource_api-1.8.9
puppetserver-ca-1.4.0
semantic_puppet-1.0.2
# ls -1 /opt/puppetlabs/puppet/lib/ruby/gems/2.5.0/gems/
concurrent-ruby-1.1.5
did_you_mean-1.2.0
gettext-3.2.2
httpclient-2.8.3
minitest-5.10.3
net-ssh-4.2.0
net-telnet-0.1.1
power_assert-1.1.1
rake-12.3.0
rdoc-6.0.1.1
test-unit-3.2.7
text-1.3.1
xmlrpc-0.3.0
```

Here's a list of things that need updating (in future PRs):

* puppet-runtime should install `gettext` and `text` into the shared gem directory. Those gems are [missing the `GEM_HOME` override like `fast-gettext`](https://github.com/puppetlabs/puppet-runtime/blob/1e3101c24810664195876a5ea928d75964785072/configs/components/rubygem-fast_gettext.rb#L16) which allows the component to be installed in an alternate location specified by the project.
* I think it's ok to keep `httpclient`, `net-ssh` and `net-telnet` in the normal Ruby gem home, as that is only used by providers and agent side code
* [@Magisus said concurrent-ruby has native extensions](https://tickets.puppetlabs.com/browse/PA-2780?focusedCommentId=693830&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-693830), so we do need different versions installed in the runtime and puppetserver gem homes.
* I think all of the gems except for concurrent-ruby could be removed from `/opt/puppetlabs/server/data/puppetserver/vendored-jruby-gems/gems/`, as they would already be present in the shared gem home.

I think it's fine to make this change in any release, as we'd be adding gems to the shared directory location (so a new agent on an older server would "see" the gems, and be able to update puppetserver), and we'd be dropping duplicates.

PR #2213, opened 2019-10-30
(maint) Update clj-parent to 4.2.6

Magisus:
This update includes changes to Jetty cipher suites, an update of
jackson-xml to fix some CVEs, and an update for tk-filesystem-watcher to
improve error handling.

This commit also adds tk-filesystem-watcher as an explicit dependency,
rather than relying on its coming in transitively through tk-jetty9.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -27,7 +27,7 @@
 
   :min-lein-version "2.9.1"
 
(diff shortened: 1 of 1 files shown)


justinstoller:
Travis needs something to the effect of https://github.com/puppetlabs/puppetserver/pull/2212

Magisus:
Rebased.

PR #2215, opened 2019-10-31
(maint) Update clj-parent to 3.1.2

Magisus:
This updates jackson-databind to a more secure version, and
tk-fs-watcher to a version with better error handling.


justinstoller:
Can we do a merge up as part of this?

Magisus:
Yes, I've just added that to all three of these.

PR #2216, opened 2019-10-31
(maint) Only build a release build if it makes sense

underscorgan:
In our pipelines configuration, we always want to try to make a release build so that we don't have to kick that off manually. But we also only want to build and publish a release build if it doesn't already exists.

This adds some safeguards so we don't try to build the release build yet if it hasn't been published and updated in dujour, and also so that we only build and publish the release build if we haven't already pushed that tag.

Blocked on #2188 

TODO
- [ ] is it worthwhile to build and test the release package on azure? Thoughts @Iristyle ?

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -7,6 +7,12 @@ on:
 
 jobs:
   build-and-publish:
+    env:
+      PUPPERWARE_ANALYTICS_STREAM: production
+      IS_LATEST: true
+      DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
+      DOCKERHUB_PASSWORD: ${{ secrets.DOCKERHUB_PASSWORD }}
+
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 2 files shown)


docker/Makefile
```diff
@@ -12,6 +12,18 @@ PUPPERWARE_ANALYTICS_STREAM ?= dev
 
 ifeq ($(IS_RELEASE),true)
 	VERSION ?= $(shell echo $(git_describe) | sed 's/-.*//')
+	# to work around failures that occur between when the repo is tagged and when the package
+	# is actually shipped, see if this version exists in dujour
+	PUBLISHED_VERSION ?= $(shell curl --silent 'https://updates.puppetlabs.com/?product=puppetserver&version=$(VERSION)' | jq '."version"' | tr -d '"')
```
Iristyle: If we turn the `product` bit into a variable, it will be easier to copy / paste this to other projects I think

docker/Makefile
```diff
@@ -12,6 +12,18 @@ PUPPERWARE_ANALYTICS_STREAM ?= dev
...
+	# For our containers built from packages, we want those to be built once then never changed
+	# so check to see if that container already exists on dockerhub
+	CONTAINER_EXISTS = $(shell DOCKER_CLI_EXPERIMENTAL=enabled docker manifest inspect $(NAMESPACE)/puppetserver:$(VERSION) > /dev/null 2>&1; echo $$?)
```
Iristyle: I now understand this after reading https://stackoverflow.com/a/52077346

.github/workflows/docker.yml
```diff
@@ -16,22 +22,33 @@ jobs:
...
+      - name: Publish release container
+        env:
+          IS_RELEASE: true
+        working-directory: docker
+        run: |
+          make prep
+          if [ $? -eq 0 ]; then
+            make publish
+          else
+            echo "Skipping release container publishing"
```
Iristyle: Do we need to make an exception for the `edge` build? We always publish that over existing `edge` builds, right?

underscorgan: nope, the release container won't be tagged edge, so after a release is published we'll just have a container from source published and a container from packages published during the same gh action

docker/Makefile
underscorgan: updated

Iristyle:
I think we can safely skip Azure for now @underscorgan

PR #2222, opened 2019-11-12
(SERVER-2630) Turn off puppet i18n when using multithreaded jruby

tvpartytonight:
This change forces the puppet configuration to turn i18n off when in a
multithreaded jruby environment.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
@@ -202,7 +202,7 @@
       (update-in [:master-run-dir] #(or % default-master-run-dir))
       (update-in [:master-log-dir] #(or % default-master-log-dir))
       (update-in [:max-requests-per-instance] #(or % 0))
-      (update-in [:disable-i18n] #(or % multithreaded))
+      (assoc :disable-i18n multithreaded)
       (update-in [:use-legacy-auth-conf] #(if (some? %) % false))
       (dissoc :environment-class-cache-enabled)))
 
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
@@ -87,6 +87,11 @@
                            [:master-log-dir "logdir"]]]
       (if-let [value (get config setting)]
         (.put puppet-config dir (ks/absolute-path value))))
+    (if (:disable-i18n config)
+      ; The value for disable-i18n is stripped in Puppet::Server::PuppetConfig
+      ; so that only the key is outputted, so that '--disable_18n' is used, not
+      ; '--disable_i18n true'
+      (.put puppet-config "disable_i18n" true))
     puppet-config))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
@@ -204,7 +209,12 @@
    agent-shutdown-fn :- IFn
    profiler :- (schema/maybe PuppetProfiler)
    metrics-service]
-  (let [initialize-pool-instance-fn (get-initialize-pool-instance-fn jruby-puppet-config profiler metrics-service)
+  (let [modified-jruby-puppet-config (if (:multithreaded jruby-config)
+                                       (do
+                                         (log/info ( i18n/trs "Disabling i18n for puppet because using multithreaded jruby"))
+                                         (assoc jruby-puppet-config :disable-i18n true))
+                                       jruby-puppet-config)
+        initialize-pool-instance-fn (get-initialize-pool-instance-fn modified-jruby-puppet-config profiler metrics-service)
         lifecycle-fns {:shutdown-on-error agent-shutdown-fn
                        :initialize-pool-instance initialize-pool-instance-fn
                        :cleanup cleanup-fn}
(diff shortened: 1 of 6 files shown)


src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -204,7 +207,12 @@
    agent-shutdown-fn :- IFn
    profiler :- (schema/maybe PuppetProfiler)
    metrics-service]
-  (let [initialize-pool-instance-fn (get-initialize-pool-instance-fn jruby-puppet-config profiler metrics-service)
+  (let [modified-jruby-puppet-config (if (get-in jruby-config [:multithreaded])
```
Magisus: I think you can just use `(:multithreaded jruby-config)` here, since you're not looking for a nested key.

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -57,6 +57,7 @@
    :master-run-dir (schema/maybe schema/Str)
    :master-log-dir (schema/maybe schema/Str)
    (schema/optional-key :track-lookups) schema/Bool
+   (schema/optional-key :disable_i18n) schema/Bool
```
Magisus: Does this have to be optional to support upgrades? Or could we require it and just set its default in `initialize-puppet-config` above, like we do with all the other keys? Same question for `track-lookups`.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -115,6 +116,8 @@
             puppet-config (config->puppet-config config)
             puppetserver-config (HashMap.)
             env-registry (puppet-env/environment-registry)]
+        (when disable_i18n
+          (.put puppet-config "disable_i18n" "" ))
```
Magisus: Might be worth putting a comment here about why the value is an empty string, but not sure. The fact that this has to obey the Puppet CLI arg rules is kinda bizarre and not at all clear based on this section of the code.

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
Magisus: Can you add a description for this setting to the doc string?

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -204,7 +207,12 @@
    agent-shutdown-fn :- IFn
    profiler :- (schema/maybe PuppetProfiler)
    metrics-service]
-  (let [initialize-pool-instance-fn (get-initialize-pool-instance-fn jruby-puppet-config profiler metrics-service)
+  (let [modified-jruby-puppet-config (if (get-in jruby-config [:multithreaded])
+                                       (do
+                                         (log/info "Disabling i18n for puppet because using multithreaded jruby")
```
Magisus: I know we're not really doing i18n anymore, but this should probably still be wrapped in `i18n/trs`. We were asked to keep the machinery working, and continuing to mark strings will save us work if we ever re-prioritize it.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
justinstoller: Should this be in `config->puppet-config`?

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
tvpartytonight: I don't know about any requirements for this to be optional to support upgrades. My interpretation here was that those values set in `initialize-puppet-config` are required to successfully start puppet, but I don't know if that interpretation is correct.

tvpartytonight:
@justinstoller @Magisus I've changed this up a bit and addressed the feedback, let me know if I've missed anything.

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -28,6 +28,9 @@
     * :track-lookups - a boolean to turn on tracking hiera lookups during compilation;
         if not specified, no tracking is enabled.
 
+    * :disable_i18n - a boolean to pass to Puppet to control whether or not to translate;
+        if not specified, uses Puppet's own default.
```
Magisus: It looks like above we default to the value of "multithreaded", meaning this key will always be present and will always override Puppet's default. Or did I misunderstand what's going on in `initialize-puppet-config`? I think that behavior is probably correct, but that makes this comment misleading.

src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
```diff
@@ -15,9 +15,13 @@ def self.initialize_puppet(puppet_config)
     #
     # `config` is a map whose keys are the names of the settings that we wish
     # to override, and whose values are the desired values for the settings.
+    # Key values that are not strings are omitted, allowing for keys in the
```
Magisus: Maybe leave the word "key" out here, it make me think we're referring to the first part of the pair.

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
tvpartytonight: Oh, that should be a hyphen; if `disable-i18n` is not in the jruby-puppet config, then it is not processed in `config->puppet-config`, and thus uses whatever value Puppet defaults to for `disable_i18n`.

Magisus: But will it ever not be in there? This line https://github.com/puppetlabs/puppetserver/pull/2222/files/ffef18266775488e80039179603feea8e714b374#diff-5ad97bf28dd9f8f97bab3706b0367684R205 should set it to `multithreaded` if it is either `false` or `nil`, I think...

PR #2223, opened 2019-11-12
(maint) Install rugged gem on puppetserver container

underscorgan:
This is needed for the `rugged` r10k backend.

### DIFF
--- a/docker/puppetserver-standalone/Dockerfile
+++ b/docker/puppetserver-standalone/Dockerfile
@@ -1,6 +1,9 @@
 FROM ubuntu:18.04 as build
 
 ENV LANG="en_US.UTF-8"
+ENV PATH=/opt/puppetlabs/server/bin:/opt/puppetlabs/puppet/bin:/opt/puppetlabs/bin:$PATH
+ENV UBUNTU_CODENAME="bionic"
+
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
             language-pack-en \
@@ -10,6 +13,16 @@ RUN apt-get update && \
             ruby \
             ruby-dev \
             g++ \
+            pkg-config \
+            cmake \
+            libssh2-1-dev \
+            gcc \
+            libc-dev \
+            zlib1g-dev \
+            libpcre3-dev \
+            libssl-dev \
+            wget \
+            ca-certificates \
             make && \
     git config --global user.name "Puppet Release Team" && \
     git config --global user.email "release@puppet.com" && \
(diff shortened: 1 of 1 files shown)


docker/puppetserver-standalone/Dockerfile
```diff
@@ -18,6 +31,29 @@ RUN apt-get update && \
     /usr/local/bin/lein && \
     gem install --no-doc bundler fpm
 
+RUN wget http://nightlies.puppet.com/apt/puppet6-nightly-release-"$UBUNTU_CODENAME".deb && \
+    dpkg -i puppet6-nightly-release-"$UBUNTU_CODENAME".deb && \
+    rm puppet6-nightly-release-"$UBUNTU_CODENAME".deb && \
+    apt-get update && \
+    apt-get install --no-install-recommends puppet-agent && \
+    git clone --single-branch --branch native_fix https://github.com/underscorgan/rake-compiler && \
```
Iristyle: Do we want a note about why we're using a fork / should we move this over to `puppetlabs` org?

underscorgan: I have a note down a couple of lines. I got a comment on my PR yesterday so I'm holding out hope that we'll be able to not use a fork, but if it lingers too long can move it to the puppetlabs org

Iristyle: Ah sorry, totally missed that.

See also https://github.com/rake-compiler/rake-compiler/pull/160

docker/puppetserver-standalone/Dockerfile
```diff
@@ -18,6 +31,29 @@ RUN apt-get update && \
     /usr/local/bin/lein && \
     gem install --no-doc bundler fpm
 
+RUN wget http://nightlies.puppet.com/apt/puppet6-nightly-release-"$UBUNTU_CODENAME".deb && \
+    dpkg -i puppet6-nightly-release-"$UBUNTU_CODENAME".deb && \
+    rm puppet6-nightly-release-"$UBUNTU_CODENAME".deb && \
+    apt-get update && \
+    apt-get install --no-install-recommends puppet-agent && \
```
Iristyle: Looks like we need the `-y` here to appease lint

underscorgan:
@Iristyle updated!

docker/puppetserver-standalone/Dockerfile
Iristyle: So the latest comment at https://github.com/rake-compiler/rake-compiler/pull/160#issuecomment-560029680 says to just `gem install --no-ri --no-rdoc rugged` like you would normally... and then copy the generated gem out of `GEM_HOME` from the build container.

That way, no `rake-compiler` is ever needed.

underscorgan:
ugh, so I started working on that and found this PR was missing a build-flag so the rugged gem isn't linking against thing like curl and libgit, which I think will probably make this not actually useful (and ubuntu doesn't seem to have compatible version of libgit). I'm going to close this for now, this one's going to be rather a lot of work.

PR #2228, opened 2019-11-21
(PE-27515) Allow sign requests to specify ttl

tvpartytonight:
This change adds in the additional optional key `cert_ttl` for the
`certificate_status` endpoint that will override the `ca-ttl` CA
setting.

### DIFF
--- a/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
+++ b/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
@@ -243,9 +243,14 @@
               (malformed
                 (i18n/tru "State {0} invalid; Must specify desired state of ''signed'' or ''revoked'' for host {1}."
                   (name desired-state) subject))
-              ; this is the happy path. we have a body, it's parsable json,
-              ; and the desired_state field is one of (signed revoked)
-              [false {::json-body json-body}])
+              (let [cert-ttl (:cert_ttl json-body)]
+                (if (and cert-ttl (schema/check schema/Int cert-ttl))
+                     (malformed
+                      (i18n/tru "cert_ttl specified for host {0} must be an integer, not \"{1}\"" subject cert-ttl))
+                     ; this is the happy path. we have a body, it's parsable json,
+                     ; and the desired_state field is one of (signed revoked), and
+                     ; it potentially has a valid cert_ttl field
+                     [false {::json-body json-body}])))
             (malformed (i18n/tru "Missing required parameter \"desired_state\"")))
           (malformed (i18n/tru "Request body is not JSON.")))
         (malformed (i18n/tru "Empty request body.")))))
--- a/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
+++ b/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
@@ -102,6 +102,12 @@
   [context]
   (keyword (get-in context [::json-body :desired_state])))
 
+(defn merge-request-settings
+  [settings context]
+  (if-let [cert-ttl (get-in context [::json-body :cert_ttl])]
+    (assoc settings :ca-ttl cert-ttl)
+    settings))
+
 (defn invalid-state-requested?
   [context]
   (when (= :put (get-in context [:request :request-method]))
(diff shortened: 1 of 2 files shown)


src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -102,6 +102,12 @@
   [context]
   (keyword (get-in context [::json-body :desired_state])))
 
+(defn merge-request-settings
```
Magisus: Do we imagine in the future adding other settings that might need to be merged in? If we do, I like the generic name of this.

But if we don't think there will/should be others, or if we'd rather set ourselves up to follow a pattern where specific settings are merged in specific contexts (as opposed to all of them all the time), it might make more sense to make this more specific to just overriding the TTL.

I don't have a strong opinion either way right now though.

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
```diff
@@ -499,7 +499,7 @@
           test-app (-> (build-ring-handler settings "42.42.42")
                        (wrap-with-ssl-client-cert))]
       (testing "PUT"
-        (testing "signing a cert"
+        (testing "signing a cert with an hour ttl"
```
mwaggett: Would it make sense to have an additional test, rather than amending this one? What's the desired outcome if `cert_ttl` is unset?

tvpartytonight: In the test scenario, if no `cert_ttl` is in the payload, the code will use to what is in the `settings` for the `ca-ttl`. This condition—where no `cert_ttl` is supplied—is tested in the following test, "signing a cert with a CA that has multiple RDNs".  It didn't really seem necessary to me to test both of these explicitly.

mwaggett: ahh gotcha. I totally missed that you were merging in `cert_ttl` as `:ca_ttl`, so this makes sense to me.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
tvpartytonight: I thought about making this more similar to `get-desired-state`, but I opted to make it more general because it did seem like there would be things in the future that we might want to allow.

Magisus:
Nice catch on the validation.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -261,7 +272,8 @@
 
   :put!
   (fn [context]
-    (let [desired-state (get-desired-state context)]
+    (let [desired-state (get-desired-state context)
+          settings (merge-request-settings settings context)]
```
justinstoller: it might be better to call this `settings'` and pass it in to set-certificate-status below.

Magisus: Agreed, I actually had to go hunting to find out what `settings` was before. This would make it clearer.

tvpartytonight: Moved it out of the let and into the function itself.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -263,7 +274,10 @@
   (fn [context]
     (let [desired-state (get-desired-state context)]
       (locking crl-write-serializer
-        (ca/set-certificate-status! settings subject desired-state)))))
+       (ca/set-certificate-status!
+        (merge-request-settings settings context)
```
Magisus: Do we want to do this inside the lock? Seems probably pretty trivial, but maybe the more we do outside it the better?

tvpartytonight: @justinstoller was indifferent to this change, but he might have an opinion? It seems pretty trivial to me...

Magisus: Okay that's fine.

PR #2230, opened 2019-11-25
Merge 5.3.x -> 6.3.x

Magisus:
Important conflict in `src/clj/puppetlabs/services/ca/certificate_authority_core.clj`.

### DIFF
--- a/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
+++ b/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
@@ -130,6 +130,12 @@
   [context]
   (keyword (get-in context [::json-body :desired_state])))
 
+(defn merge-request-settings
+  [settings context]
+  (if-let [cert-ttl (get-in context [::json-body :cert_ttl])]
+    (assoc settings :ca-ttl cert-ttl)
+    settings))
+
 (defn invalid-state-requested?
   [context]
   (when (= :put (get-in context [:request :request-method]))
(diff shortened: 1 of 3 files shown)


src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -265,9 +271,14 @@
               (malformed
                 (i18n/tru "State {0} invalid; Must specify desired state of ''signed'' or ''revoked'' for host {1}."
                   (name desired-state) subject))
-              ; this is the happy path. we have a body, it's parsable json,
-              ; and the desired_state field is one of (signed revoked)
-              [false {::json-body json-body}])
+              (let [cert-ttl (:cert_ttl json-body)]
```
mwaggett: I just noticed.. here we're referencing the key `:cert_ttl` but above we add it to setting as `:ca_ttl` - was that intentional? or should this be `:ca_ttl` too?

mwaggett: oh wait no, I see, we're getting it from the json body as `:cert_ttl` above

mwaggett:
I'm not merging because I don't fully know the differences between Puppet 5 and 6, which I assume was the source of this conflict. Looks sane to me, but I'd feel better if someone else also took a look.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
Magisus: Yeah it's intentional, in the request body we want it to be `cert_ttl`, since we're only trying to affect the TTL of one cert, not the whole CA. Puppet doesn't currently have that granularity: `ca_ttl` controls everything (though in this case we're just overwriting it for one cert). Either there is a ticket to add a separate `cert_ttl` setting, or we should file one (I can't find it right now).

PR #2231, opened 2019-11-26
DOC-4171 Add known issue

clairecadman:
This commit re-adds advice about using Puppet::Util::Execution.execute to run external commands from Puppet Server. https://tickets.puppetlabs.com/browse/DOC-4171

### DIFF
--- a/documentation/gems.markdown
+++ b/documentation/gems.markdown
@@ -88,7 +88,7 @@ settings with Puppet. You can manage Puppet Server's gem dependencies with the
 package provider shipped in [`puppetlabs-puppetserver_gem`](https://forge.puppet.com/puppetlabs/puppetserver_gem)
 module.
 
-> Note: If you try to load a gem before it's been installed, the agent run will fail with a `LoadError`.
+> Note: If you try to load a gem before it's been installed, the agent run will fail with a `LoadError`. If this happens, reload the server to resolve the issue.
 
 ## Installing gems for use with development:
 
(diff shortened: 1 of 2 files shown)


documentation/gems.markdown
```diff
@@ -88,7 +88,9 @@ settings with Puppet. You can manage Puppet Server's gem dependencies with the
 package provider shipped in [`puppetlabs-puppetserver_gem`](https://forge.puppet.com/puppetlabs/puppetserver_gem)
 module.
 
-## Installing Gems for use with development:
+> Note: If you try to load a gem before it's been installed, the agent run will fail with a `LoadError`.
```
justinstoller: And you will need to reload the server to resolve the issue.

Magisus: Why do you only need to reload the server if you tried to load the gem before it was installed? Does something get cached?

justinstoller: Yeah, that was the guy's issue that triggered this. The load error gets cached and it won't try to reload it again until a new JRuby gets it.

clairecadman:
DOC-4171 is for the 6.0 docs and DOC-4226 is for latest. Is 6.3x the right branch for this?

Magisus:
Do we not want to add the gem note to 5.x? That was probably relevant then.

If you want, we can open PRs for this update, now that we have the text from you. I know the branching has been a huge hassle :(

clairecadman:
@Magisus ah yes it would be good to have the note there as well. Yes that would be great thank you! :)

Magisus:
Filed https://github.com/puppetlabs/puppetserver/pull/2233 and https://github.com/puppetlabs/puppetserver/pull/2234, closing this.

PR #2232, opened 2019-11-26
DOC-4226 Add a step about restarting server

clairecadman:
https://tickets.puppetlabs.com/browse/DOC-4226


mwaggett:
Is this the right target branch? I didn't even know `6.6.0` was a thing..?

Magisus:
It shouldn't be... this probably should be targeted at 5.3.x.

documentation/gems.markdown
```diff
@@ -88,6 +88,8 @@ settings with Puppet. You can manage Puppet Server's gem dependencies with the
 package provider shipped in [`puppetlabs-puppetserver_gem`](https://forge.puppet.com/puppetlabs/puppetserver_gem)
 module.
 
+>Note: If you try to load a gem before it's been installed, the agent run will fail with a `LoadError`.
```
Magisus: It seems like the important part of that ticket was the note that the server needs to be restarted after installing the gem. This note might also be useful, but I notice we also don't mention the actual solution to his problem anywhere in this section.

Magisus:
It might be easier to make a new PR based against that branch. The git UI doesn't handle retargeting from newer branches to older ones well...

PR #2236, opened 2019-12-02
(SERVER-2684) Take up new jruby-utils

Magisus:
This reverts the interim implementation of the `multithreaded` mode, replacing it with the new JRuby reference pool.

PR #2237, opened 2019-12-03
(SERVER-2526) Add basic testing for thread safety

pcarlisle:
Run with `lein thread-test` and then run ext/thread_test/test.rb. See
the file for setting iterations and number of processes. Individual
processes will stop on the first error and cause the main process to
exit 1 if there is a failure.

### DIFF
--- a/ext/thread_test/Gemfile
+++ b/ext/thread_test/Gemfile
@@ -1,0 +1,3 @@
+source 'https://rubygems.org'
+
+gem 'rspec-expectations'
--- a/ext/thread_test/puppetserver.conf
+++ b/ext/thread_test/puppetserver.conf
@@ -1,0 +1,200 @@
+global: {
+    logging-config: ./dev/logback-dev.xml
+}
+
+product: {
+    update-server-url: "http://some.update.server.that.does.not.exist/"
+    name: {group-id: puppetlabs.dev
+           artifact-id: puppetserver}
+}
+
+webserver: {
+    access-log-config: ./dev/request-logging-dev.xml
+    # client-auth: want
+    # ssl-host controls what networks the server will accept connections from.
+    # The default value below is '0.0.0.0', so will accept connections from
+    # any client.  For better security, you might wish to set this to 'localhost'.
+    host: localhost
+    port: 8140
+}
+
+# configure the mount points for the web apps
+web-router-service: {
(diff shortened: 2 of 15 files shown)


ext/thread_test/puppetserver.conf
```diff
@@ -0,0 +1,200 @@
...
+    master-var-dir: ${HOME}"/.puppetlabs/opt/puppet/cache"
+
+    # (optional) path to puppet run dir; if not specified, will use the puppet default
+    master-run-dir: ${HOME}"/.puppetlabs/var/run"
+
+    # (optional) path to puppet log dir; if not specified, will use the puppet default
+    master-log-dir: ${HOME}"/.puppetlabs/var/log"
+
+    # (optional) maximum number of JRuby instances to allow
+    max-active-instances: 1
```
Magisus: With the ReferencePool, this setting is used to control the number of threads that can go through the instance at once. Once we have integrated that, we probably want to set it higher to properly test the multithreaded behavior.

Related tangential question: should we double up on the usage of this setting, or create new ones specific to the multithreaded environment?

ext/thread_test/Gemfile
```diff
@@ -0,0 +1 @@
+gem 'rspec-expectations'
```
justinstoller: I needed to add a 
```rb
source 'https://rubygems.org'
```
in order fro this to work

justinstoller: Is there a reason we don't use the Gemfile at the root of the project?

justinstoller:
👍 was able to cause threading issues with this

Magisus:
I was also able successfully cause threading errors when I combined this with https://github.com/puppetlabs/puppetserver/pull/2236 and turned the `max-active-instances` up to 2. This uses the new JRuby pool architecture to achieve concurrent runs, rather than the interim implementation. Turning `max-active-instances` back down to 1 prevented the errors (as expected). In both cases the server was able to shut down gracefully following a ^C.

ext/thread_test/puppetserver.conf
pcarlisle: Honestly I'm not sure

Magisus: We can leave the setting the same for now and update it if it seems bad or confusing once we show it to other people. I think it's okay if the API is a little fluid at the start.

PR #2238, opened 2019-12-04
(maint) Migrate container builds from pipelines to GitHub Actions

underscorgan:
Pipelines is EOL Jan 31, so we need to migrate to something else. This
adds a github action to lint, build, test, and publish containers on
pushes to the master branch.

Pipelines auto-builds have been disabled and necessary secrets have been
added to the puppetserver github repo.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,38 @@
+name: Docker test and publish
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  build-and-publish:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@master
+      - name: Set up Ruby 2.6
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: 2.6.x
+      - name: Lint container
+        working-directory: docker
+        run: |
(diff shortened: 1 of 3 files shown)


underscorgan:
Tested at https://github.com/puppetlabs/puppetserver/commit/d51eb35a8b069bd5054d5e0392da6a17b2cb4a74/checks?check_suite_id=342410096 (only difference: test ran on all branches)

.github/workflows/docker.yml
```diff
@@ -0,0 +1,38 @@
...
+        working-directory: docker
+        run: |
+          make lint
+      - name: Build container
+        env:
+          PUPPERWARE_ANALYTICS_STREAM: production
+          IS_LATEST: true
+        working-directory: docker
+        run: |
+          gem install bundler
```
nwolfe: This could go with the "setup ruby" step so we have our Ruby environment complete and ready-to-go for all subsequent steps.

Not a big deal, just thinking about organization/cohesion.

underscorgan: Updated. It has to be its own step since you can't have `run` and `uses` in the same step, but moved it higher up at least

PR #2239, opened 2019-12-11
(DO NOT MERGE) CI Validating #2188 clean-up

Iristyle:
Cleaned up #2188 while reviewing.... Verifying that each step builds in CI.


Iristyle:
416bf7c passed everywhere already and it's very similar to ce9742b (cleaned up a little and vetted on dev box), so skipping that commit and moving on to 3710bc0 since Travis takes so long here!

PR #2240, opened 2019-12-12
(maint) Improve Dockerfile organization

Iristyle:
- Reduce Dockerfile build steps
- Improve ability to cache build steps
- Make sure resulting images have all metadata correctly (including `build_date` and `vcs_ref`) 
- Ensure all constituent `Dockerfile` source files are copied into image (base Dockerfile and derivatives)

### DIFF
--- a/docker/puppetserver-base/Dockerfile
+++ b/docker/puppetserver-base/Dockerfile
@@ -6,29 +6,29 @@ ARG version="6.0.0"
 # Used by entrypoint to submit metrics to Google Analytics.
 # Published images should use "production" for this build_arg.
 ARG pupperware_analytics_stream="dev"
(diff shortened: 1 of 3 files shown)


docker/puppetserver-base/Dockerfile
```diff
@@ -64,4 +64,4 @@ RUN chmod +x /docker-entrypoint.sh /healthcheck.sh && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
-COPY Dockerfile /
+COPY Dockerfile /base.Dockerfile
```
Iristyle: We might be able to get rid of the base file completely later, if we use `--target` along with `DOCKER_BUILDKIT=1`... so this may ultimately not be too important. But makes sense to do this for now IMHO.

PR #2242, opened 2019-12-16
(maint) Enable Docker builds with buildkit

Iristyle:
This is an experiment to see if we can improve Docker build times by using [buildkit](https://docs.docker.com/develop/develop-images/build_enhancements/). Buildkit will enable a few things:

* Using [multiple targets](https://docs.docker.com/develop/develop-images/multistage-build/#stop-at-a-specific-build-stage) inside of a single Dockerfile and smart evaluation (i.e. no need to have multiple Dockerfiles like base, release, etc)
* Concurrent build processes where possible, rather than strictly sequential
* Better caching (not applicable to Travis, but very applicable to Windows / LCOW)

It looks like this is generally a no-go at this point in time due to the LCOW error, which could be a permissions issue (though I tested setting `Everyone: (F)` for `C:\ProgramData\docker\tmp` *and* `C:\ProgramData\docker\windowsfilter` and didn't see anything stand out in `procmon`):

```
failed to solve with frontend dockerfile.v0: failed to read dockerfile: failed to mount C:\ProgramData\docker\tmp\buildkit-mount984162069: [{Type:bind Source:C:\ProgramData\docker\windowsfilter\8es1az6gkgzbc4u4w031q6maz Options:[rbind ro]}]: invalid windows mount type: 'bind'
```

This is something we probably want to evaluate again in the future (or help fix under LCOW) - buildkit currently says Windows is unsupported - https://github.com/moby/buildkit/issues/616

Update: It looks like it may be possible to use buildkit more directly rather than going through `docker build`. [buildctl](https://github.com/moby/buildkit/releases) has a Windows binary and it's possible to run the [buildkit daemon in a container](https://hub.docker.com/r/moby/buildkit).

That's a reasonably different workflow from what we're doing now, but seems like it may plausibly work without much modification. The convention for passing command line args would vary a bit. For instance, `--build-arg vcs_ref=$(vcs_ref)` would become `--opt build-arg:vcs_ref=$(vsc_ref)`. See https://github.com/moby/buildkit#exploring-dockerfiles for more details.


azure-pipelines.yml
```diff
@@ -14,6 +14,7 @@ pool:
   name: Default
 
 variables:
+  DOCKER_BUILDKIT: 1
```
Iristyle: Looks like this might not be possible on Windows (at least not easily):

```
docker build --build-arg version=6.7.2 --build-arg vcs_ref=3c27f646dc61db97a94a8f6f73795b00f0fcdb77 --build-arg build_date=--file docker/puppetserver-base/Dockerfile --memory 4g --build-arg namespace=puppet --tag puppet/puppetserver-base:6.7.2 --pull docker/puppetserver-base
#1 [internal] load build definition from Dockerfile
#1 transferring dockerfile: 2.62kB done
#1 DONE 0.2s

#2 [internal] load .dockerignore
#2 transferring context: 2B done
#2 DONE 0.2s
failed to solve with frontend dockerfile.v0: failed to read dockerfile: failed to mount C:\ProgramData\docker\tmp\buildkit-mount519259336: [{Type:bind Source:C:\ProgramData\docker\windowsfilter\fvor99jmpaj39d3lp5ljauvuu Options:[rbind ro]}]: invalid windows mount type: 'bind'
##[error]PowerShell exited with code '1'.
##[section]Finishing: Build puppetserver-base Container

```

docker/Makefile
```diff
@@ -1,3 +1,4 @@
+export DOCKER_BUILDKIT=1
```
Iristyle: Travis docker version is super old, and buildkit isn't supported here either.

```
docker version
Client:
 Version:           18.06.0-ce
 API version:       1.38
 Go version:        go1.10.3
 Git commit:        0ffa825
 Built:             Wed Jul 18 19:11:02 2018
 OS/Arch:           linux/amd64
 Experimental:      false
Server:
 Engine:
  Version:          18.06.0-ce
  API version:      1.38 (minimum version 1.12)
  Go version:       go1.10.3
  Git commit:       0ffa825
  Built:            Wed Jul 18 19:09:05 2018
  OS/Arch:          linux/amd64
  Experimental:     false
```

.travis.yml
```diff
@@ -1,10 +1,20 @@
 language: clojure
 lein: 2.9.1
-dist: xenial
+dist: bionic
 
 services:
   - docker
 
+before_install:
+  - curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
```
Iristyle: I think it's probably quicker to just pull the packages from https://download.docker.com/linux/ubuntu/dists/xenial/pool/stable/amd64/ and install that way rather than using the PPA. The Xenial PPA doesn't have the latest builds for some reasons anyway!

Iristyle:
Update timings... looks like the host might have just been slightly faster. A 30s savings is good, but will further refactor the build process around targets to see if that helps any further.

before_install 0.64s -> 26.36s
before_script 350.44s -> 320.47s
script - 141.40s -> 131.81s

.travis.yml
Iristyle: Nope, not true. Verified that the time is about the same - except the code to install packages is a fair amount more complicated (especially when resolving dependencies).

Better to use the PPA

.travis.yml
```diff
@@ -2,36 +2,44 @@ language: clojure
...
       env:
+        - DOCKER_BUILD_FLAGS="--progress plain"
         - DOCKER_COMPOSE_VERSION=1.25.0-rc2
       before_install: |
         OS_ARCH=$(uname --machine)
+        OS_RELEASE=$(lsb_release -cs)
+        PKG_ARCH=$(dpkg --print-architecture)
+        curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+        sudo add-apt-repository "deb [arch=${PKG_ARCH}] https://download.docker.com/linux/ubuntu $OS_RELEASE stable"
+        sudo apt-get update
```
Iristyle: Not sure if it makes sense to only upgrade the Docker PPA - i.e. something like

```
apt-get update -o Dir::Etc::sourcelist=/path/to/repo.list
```

https://askubuntu.com/a/698028

Iristyle:
Buildkit support was merged in https://github.com/puppetlabs/puppetserver/pull/2328

This seems to have been left open by accident

PR #2243, opened 2019-12-19
(maint) Restructure Travis container scripts

Iristyle:
- Segregate into:

   before_install - for setting up tools like compose
   before_script - for linting / building
   script - for testing

 - This allows for better timings of these individual sequences in
   the Travis UI (for comparison with other build strategies later)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,18 +26,14 @@ jobs:
       rvm: 2.5.5
       env:
         - DOCKER_COMPOSE_VERSION=1.25.0-rc2
-      script:
-        - |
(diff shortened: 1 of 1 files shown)


Iristyle:
Mission accomplished.

Screen Shot 2019-12-19 at 6 41 05 AM

before_install - 0.64 seconds
before_script - 350.44 seconds
script - 141.40 seconds

This now gives us a much better basis for comparison later when the build or test suites change.

Iristyle:
Merging as this only impacts Docker work!

PR #2244, opened 2019-12-19
(DOC-4397) Add step on how to reload PE

clairecadman:
Based on the following docs feedback:
https://tickets.puppetlabs.com/browse/DOC-4397

### DIFF
--- a/documentation/restarting.markdown
+++ b/documentation/restarting.markdown
@@ -39,7 +39,7 @@ OS distributions which use systemd service configurations:
 
     systemctl reload puppetserver
 
-> **Note:** If you're using Puppet Enterprise (PE), you can reload the server from the command line by running `service pe-puppetserver reload`. However if you need to change a setting, do so in console or with Heira, and then the agent will reload the server when it applies the change. For more information, see [Configuring and tuning Puppet Server](https://puppet.com/docs/pe/2019.2/config_puppetserver.html#configuring_and_tuning_puppet_server). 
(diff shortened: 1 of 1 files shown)


documentation/restarting.markdown
```diff
@@ -39,6 +39,8 @@ OS distributions which use systemd service configurations:
 
     systemctl reload puppetserver
 
+> **Note:** If you're using Puppet Enterprise (PE), you can reload the server by running `service pe-puppetserver reload`, or by using the console or Hiera. For more information, see [Configuring and tuning Puppet Server](https://puppet.com/docs/pe/2019.2/config_puppetserver.html#configuring_and_tuning_puppet_server).
```
justinstoller: >  or by using the console or Hiera.

I meant along the lines of " but if you are changing settings that require a reload of the server we recommend you manage those setting via the console or Hiera allow the puppet agent to reload the server as needed."

clairecadman: I'm not sure I understand. Something like this? "If you are using Puppet Enterprise (PE), you can reload the server by running `service pe-puppetserver reload`. However if you are changing settings that require a reload of the server, use the console or Hiera to run the Puppet agent". 

Do you have an example of the kind of settings someone might be changing?

justinstoller: So, I wasn't trying to say to use the console to reload the server. Assuming one is trying to reload the server because they changed some setting, I would recommend that instead of ad-hoc changing server settings and manually reloading the server, they should instead manage that setting in the console or with heira and let the agent reload the server when it applies the change.

justinstoller: > Do you have an example of the kind of settings someone might be changing?

Any setting changed in the puppet.conf (like custom report processor installation) or in puppetserver's conf.d (like changing max-instances) requires a reload of the server. Any changed setting in the service file (like java_args) requires a restart of the server.

clairecadman: Ah right sorry, I get what you're saying now! I've updated it, let me know what you think. Thanks!

PR #2245, opened 2019-12-19
(DOC-4395) Clean up installing from packages docs

clairecadman:
Based on user feedback: https://tickets.puppetlabs.com/browse/DOC-4395

### DIFF
--- a/documentation/install_from_packages.markdown
+++ b/documentation/install_from_packages.markdown
@@ -7,103 +7,58 @@ canonical: "/puppetserver/latest/install_from_packages.html"
 [repodocs]: https://puppet.com/docs/puppet/latest/puppet_platform.html
 [passengerguide]: https://puppet.com/docs/puppet/latest/passenger.html
 
-## System Requirements
-
 Puppet Server is configured to use 2 GB of RAM by default. If you'd like to just play around with an installation on a Virtual Machine, this much memory is not necessary. To change the memory allocation, see [Memory Allocation](#memory-allocation).
 
 > If you're also using PuppetDB, check its [requirements](https://puppet.com/docs/puppetdb/latest/index.html#system-requirements).
 
-## Platforms with Packages
-
-Puppet provides official packages that install Puppet Server 5.1 and all of its prerequisites on x86_64 architectures for the following platforms, as part of [Puppet Platform][repodocs].
-
-### Red Hat Enterprise Linux
-
--   Enterprise Linux 6
--   Enterprise Linux 7
-
-### Debian
-
--   Debian 8 (Jessie)
-
-Java 8 runtime packages do not exist in the standard repositories for Debian 8 (Jessie).  To install Puppet Server on Jessie, [configure the `jessie-backports` repository](https://backports.debian.org/Instructions/), which includes openjdk-8:
-
-```
-echo "deb http://ftp.debian.org/debian jessie-backports main" > /etc/apt/sources.list.d/jessie-backports.list
-apt-get update
-apt-get -t jessie-backports install "openjdk-8-jdk-headless"
-apt-get install puppetserver
-```
-
-If you upgraded on Debian from older versions of Puppet Server, or from Java 7 to Java 8, you must also configure your server to use Java 8 by default for Puppet Server 5.x:
-
-```
-update-alternatives --set java /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java
-```
-
-### Ubuntu
-
--   Ubuntu 18.04 (Bionic)
--   Ubuntu 16.04 (Xenial)
-
-On Ubuntu 18.04, enable the [universe repository](https://help.ubuntu.com/community/Repositories/Ubuntu), which contains packages necessary for Puppet Server.
-
-### SUSE Linux Enterprise Server
-
--   SLES 12 SP1
-
-## Quick Start
+## Install Puppet Server from packages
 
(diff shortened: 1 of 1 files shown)


documentation/install_from_packages.markdown
```diff
@@ -7,103 +7,58 @@ canonical: "/puppetserver/latest/install_from_packages.html"
...
 
-    Alternatively, if you don't need to keep the Apache service running, you can stop Apache with `service httpd stop` or `service apache2 stop`.
-
-3.  Install the Puppet Server package by running:
+2.  Install the Puppet Server package by running:
 
         yum install puppetserver
 
-    Or
+    or
```
clairecadman: @justinstoller When would you use one or the other? Would the user know which to use?

clairecadman:
@justinstoller I've tried to clean this page up a bit, based on your suggestions. Let me know what you think. Thanks!

documentation/install_from_packages.markdown
justinstoller: Yeah, the preferred package manager depends on the "family" the OS belongs to. This list `yum` & `apt` is actually incomplete (SLES uses `zypper`, iirc) however. I expect users that are doing this level of system administration to know how to install packages and grok which one (if either) is applicable.

documentation/install_from_packages.markdown
```diff
@@ -7,103 +7,58 @@ canonical: "/puppetserver/latest/install_from_packages.html"
...
         apt-get install puppetserver
 
     Note that there is no `-` in the package name.
 
-4.  Start the Puppet Server service:
+3.  Start the Puppet Server service:
+
+`systemctl start puppetserver` or `service puppetserver start`
+        
+> Note: If you're uprgading, stop any existing `puppetmaster` or `puppetserver` service by running `service <service_name> stop` or `systemctl stop <service_name>`.
```
justinstoller: This note kinda bleeds into step three when I looked at the GH rendering of the page. I don't know if that's how it will be on the site? It would be great to have a numerical reference from step 2 to this note to mark that they are related.

documentation/install_from_packages.markdown
```diff
@@ -7,103 +7,58 @@ canonical: "/puppetserver/latest/install_from_packages.html"
...
+* Red Hat Enterprise Linux 6
+* Red Hat Enterprise Linux 7
+* Debian 8 (Jessie)
+* Debian 10 (Stretch)
+* Ubuntu 18.04 (Bionic) - enable the [universe repository](https://help.ubuntu.com/community/Repositories/Ubuntu), which contains packages necessary for Puppet Server.
+* Ubuntu 16.04 (Xenial)
+* SLES 12 SP1
 
-        service puppetserver start
+> Note: Java 8 runtime packages do not exist in the standard repositories for Debian 8 (Jessie).  To install Puppet Server on Jessie, [configure the `jessie-backports` repository](https://backports.debian.org/Instructions/).
```
justinstoller: Since I made a comment how I thought the other note was a little confusing I'll write here why I _don't_ find this one confusing. In the GH render the list above is sufficiently different that the note is well differentiated and the note itself calls out where it's applicable (Jessie). 👍

documentation/install_from_packages.markdown
justinstoller: > uprgading

is also misspelled

documentation/install_from_packages.markdown
```diff
@@ -7,103 +7,58 @@ canonical: "/puppetserver/latest/install_from_packages.html"
...
+    or
 
         apt-get install puppetserver
 
     Note that there is no `-` in the package name.
 
-4.  Start the Puppet Server service:
+3.  Start the Puppet Server service:
+
+`systemctl start puppetserver` or `service puppetserver start`
```
justinstoller: In the GH render I looked at this was also rendered as a regular sentence with a line break and monospaced sections. While the above `yum install ...` or `apt-get install ...` are both on their own lines broken out. I think having those two pairs rendered/treated differently is confusing. I don't care which (inline or broken out) but they're both CLI commands that vary across OSes and I think should be treated the same.

documentation/install_from_packages.markdown
```diff
@@ -7,103 +7,70 @@ canonical: "/puppetserver/latest/install_from_packages.html"
...
-update-alternatives --set java /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java
-```
-
-### Ubuntu
-
--   Ubuntu 18.04 (Bionic)
--   Ubuntu 16.04 (Xenial)
-
-On Ubuntu 18.04, enable the [universe repository](https://help.ubuntu.com/community/Repositories/Ubuntu), which contains packages necessary for Puppet Server.
+There is no `-` in the package name.
```
clairecadman: @justinstoller Is this correct? As there is a `-` in `apt-get`?.

clairecadman:
@justinstoller Thanks! I made the changes you suggested, and also commented on one more thing. Let me know if you notice anything else.

documentation/install_from_packages.markdown
justinstoller: yeah, debian's package manager is called "apt" and it's command line tools are "apt-get" and "apt-cache", etc.

Our package used to be called "puppet-server" (which was the package name in Puppet Server 1.x) but for a long time it's just been "puppetserver".

clairecadman: Ah okay great, thanks!

PR #2247, opened 2019-12-19
(SERVER-2405,SERVER-2696) Improve max-queued-requests behavior

Sharpie:
This patchset contains two improvements for the Server-side behavior of the `max-queued-requests` setting:

  - 503 responses with `Retry-After` are not sent to HTTP clients that do not set a `X-Puppet-Version` header with `5.3.1` or newer. Puppet agents older than this version, or non-puppet agents won't understand what to do with the `Retry-After` header.

  - A `Connection: close` header is added to any 503 response that includes `Retry-After`. This will cause Puppet agents to close their TCP connection to the server before sleeping for the amount of time specified by `RetryAfter`.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -49,6 +49,7 @@
                  [commons-io]
 
                  [clj-time]
+                 [grimradical/clj-semver "0.3.0" :exclusions [org.clojure/clojure]]
                  [prismatic/schema]
                  [me.raynes/fs]
                  [liberator]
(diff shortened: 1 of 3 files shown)


Sharpie:
Unfortunately, this set of changes doesn’t keep an agent herd from perpetuating. It just mitigates two issues that occur when herds form up:

  - Requests from clients that don’t understand how to sleep and retry, like older Puppet agents or `pxp-agent` task downloads, are allowed to go into the wait queue.

  - Puppet agents that are served a 503 with a `Retry-After` are also instructed to close their connections. This keeps stale connections from piling up on the Server while the agents sleep.

I’m working on a separate set of changes for the puppet project that should avoid perpetuating the herd when the master hands out 503s.

Magisus:
Ah okay, I think I misunderstood. Nvm then :)

Sharpie:
@alexjfisher Opened https://tickets.puppetlabs.com/browse/SERVER-2720 for that and added some comments.

PR #2248, opened 2020-01-02
Merge 5.3.x -> 6.3.x

Magisus:


justinstoller:
Is that the semver library not liking the SNAPSHOT version?

Magisus:
No, it appears to be this: https://github.com/puppetlabs/puppetserver/blob/master/test/integration/puppetlabs/services/master/master_service_test.clj#L762 Looking into why.

PR #2249, opened 2020-01-03
(maint) Remove test cases that have bad behavior with mock JRubies

Magisus:
The tests for the (deprecated) whitelist access control for the admin
API were previously testing each of the various endpoints under
`puppet-admin-api`, even though the whitelist applies the same to each
of them. Moreover, two of the endpoints (to get thread dumps and flush
the JRuby pool) also do not behave correctly with mock JRubies.

Since the whitelist tests only need to show that the endpoint returns a
success code instead of 'access denied', this commit updates them to
only test one of the endpoints, the one that behaves correctly when
JRubies are mocked, as a representative case for testing the whitelist.
The other two endpoints are tested for their functionality elsewhere.


Magisus:
This is prep work for running puppetserver's tests in multithreaded mode (see https://tickets.puppetlabs.com/browse/SERVER-2680). The bad behavior in these tests was made evident when switching to that mode, where things like flushing work a little bit differently.

PR #2251, opened 2020-01-07
(SERVER-2680) Parameterize Clojure tests for multithreaded mode

Magisus:
This allows the puppetserver clojure tests to be run in multithreaded mode when the `MULTITHREADED` environment variable is set to `true`. Setting this var will update the test configs to toggle the `multithreaded` setting. The tests should then be run with `lein test :multithreaded` to ensure correct filtering, since some tests require different implementations between the two modes.

### DIFF
--- a/test/integration/puppetlabs/puppetserver/bootstrap_testutils.clj
+++ b/test/integration/puppetlabs/puppetserver/bootstrap_testutils.clj
@@ -39,6 +39,9 @@
 (def master-log-dir
   "./target/master-var/log")
 
+(def multithreaded
+  (= "true" (System/getenv "MULTITHREADED")))
+
 (defn load-dev-config-with-overrides
   [overrides]
   (let [tmp-conf (ks/temp-file "puppetserver" ".conf")]
@@ -50,6 +53,7 @@
         (assoc-in [:jruby-puppet :master-var-dir] master-var-dir)
         (assoc-in [:jruby-puppet :master-run-dir] master-run-dir)
         (assoc-in [:jruby-puppet :master-log-dir] master-log-dir)
+        (assoc-in [:jruby-puppet :multithreaded] multithreaded)
         (ks/deep-merge overrides))))
 
 (def services-from-dev-bootstrap
--- a/test/integration/puppetlabs/services/jruby/jruby_metrics_service_test.clj
+++ b/test/integration/puppetlabs/services/jruby/jruby_metrics_service_test.clj
@@ -283,7 +283,7 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Tests
 
-(deftest ^:metrics basic-metrics-test
+(deftest ^:metrics ^:single-threaded-only basic-metrics-test
   (with-metrics-test-env test-env default-test-config
     (let [{:keys [num-jrubies num-free-jrubies requested-count
                   requested-jrubies-histo borrow-count borrow-timeout-count
@@ -320,7 +320,7 @@
           (is (= expected-metrics (-> (current-jruby-status-metrics)
                                       (jruby-status-metric-counters)))))))))
 
-(deftest ^:metrics borrowed-instances-test
+(deftest ^:metrics ^:single-threaded-only borrowed-instances-test
   (with-metrics-test-env test-env default-test-config
     (let [{:keys [coordinator update-expected-values expected-metrics-values
                   current-metrics-values jruby-service]} test-env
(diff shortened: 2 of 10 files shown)


justinstoller:
fwiw, looks like a good approach

Magisus:
So I've been watching Travis run, and it seems like it's only running two cells in parallel at a time, so adding these does extend the time taken to run Travis by ~20 minutes.

test/integration/puppetlabs/puppetserver/bootstrap_testutils.clj
```diff
@@ -39,6 +39,9 @@
 (def master-log-dir
   "./target/master-var/log")
 
+(def multithreaded
+  (or (= "true" (System/getenv "MULTITHREADED")) false))
```
mwaggett: What's the purpose of wrapping this in an `or`?

Magisus: Good point, that's redundant.

test/unit/puppetlabs/services/jruby/jruby_puppet_testutils.clj
```diff
@@ -37,6 +37,9 @@
 (def run-dir "./target/master-var/run")
 (def log-dir "./target/master-var/log")
 
+(def multithreaded
+  (or (= "true" (System/getenv "MULTITHREADED")) false))
```
mwaggett: same here re: `or`

test/integration/puppetlabs/services/jruby/jruby_metrics_service_test.clj
```diff
@@ -886,7 +886,7 @@
               (is (= 1 (.getCount timer))
                   (str "Timer has a accumulated a count for borrow reason: " timer-name)))))))))
 
-(deftest ^:metrics borrow-timeout-test
+(deftest ^:metrics ^:single-threaded-only borrow-timeout-test
```
mwaggett: I'm curious how you decided which tests should be single-threaded only. 

This test in particular is one I was using to test my worker-id stuff, so seems like it could work in multithreaded?

Magisus: I flagged every test that currently fails in `multithreaded` mode. The metrics tests in particular I assumed were going to be shortly fixed by your work. You should feel free to delete those tags once the tests pass in both modes.

Some of the others need alternative versions written (e.g. the ones that test for different variables being set in different JRuby instances), but I wanted to do that as a separate ticket to keep this manageable to review.

mwaggett: gotcha! cool, so then as long as there are tickets with the necessary follow-up work, this makes sense to me.

test/integration/puppetlabs/services/jruby/tasks_test.clj
```diff
@@ -140,8 +140,8 @@
         (testing "with the init task"
           (let [res (get-task-data "env1" "apache" "init")]
             (is (nil? (schema/check mc/TaskData res)))
-            (is (= "init.rb" (-> res :files first :name))
-            (is (re-matches #".*init\.rb" (-> res :files first :path))))))
```
justinstoller: that's kinda terrifying

test/integration/puppetlabs/services/jruby/jruby_metrics_service_test.clj
justinstoller: Is there (tickets?)? Or is it just that whichever of you two has their PR merged second should change this to not have a thread label?

mwaggett: I'll remove this one on my PR, since I think this PR should go first.
But we should ticket the rest.

test/integration/puppetlabs/puppetserver/bootstrap_testutils.clj
```diff
@@ -39,6 +39,9 @@
 (def master-log-dir
   "./target/master-var/log")
 
+(def multithreaded
+  (= "true" (System/getenv "MULTITHREADED"))
```
justinstoller: Missing a paren

test/unit/puppetlabs/services/jruby/jruby_puppet_testutils.clj
```diff
@@ -37,6 +37,9 @@
 (def run-dir "./target/master-var/run")
 (def log-dir "./target/master-var/log")
 
+(def multithreaded
+  (= "true" (System/getenv "MULTITHREADED"))
```
justinstoller: ditto

test/integration/puppetlabs/puppetserver/bootstrap_testutils.clj
Magisus: Oh I think I did this when I updated for Molly's comment, whoops. I'm really hoping it didn't work :P

test/integration/puppetlabs/services/jruby/tasks_test.clj
Magisus: I guess `(is)` returns true??

Magisus: Yeah not really sure what this was doing, but it started showing up as a syntax error only when run not in my IDE...

justinstoller: The output of the second test was a parameter to the first `is` (it takes an optional string to output on failure)

Magisus: Oh I didn't know about the optional param. Joy.

PR #2252, opened 2020-01-08
(SERVER-2681) Fix `borrow-timeout-test` for multithreaded mode

mwaggett:
This commit updates the expected values in the metrics service tests to be
conditional on whether we are running with multithreaded turned on or not.
There is only ever 1 JRuby instance in multithreaded mode, so this ensures that
tests reflect that.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
@@ -2,6 +2,7 @@
   (:require [schema.core :as schema]
             [puppetlabs.metrics :as metrics]
             [puppetlabs.services.jruby-pool-manager.jruby-schemas :as jruby-schemas]
+            [puppetlabs.services.jruby.jruby-puppet-core :as jruby-puppet-core]
             [clojure.string :as str]
             [clojure.tools.logging :as log]
             [puppetlabs.trapperkeeper.services.status.status-core :as status-core]
@@ -366,7 +367,7 @@
    free-instances-fn :- IFn
    registry :- MetricRegistry]
   {:num-jrubies (metrics/register registry (metrics/host-metric-name hostname "jruby.num-jrubies")
-                  (metrics/gauge max-active-instances))
+                  (if jruby-puppet-core/multithreaded? (metrics/gauge 1) (metrics/gauge max-active-instances)))
    :requested-count (.counter registry (metrics/host-metric-name hostname "jruby.request-count"))
    ;; See the comments on `jruby-metrics-service/schedule-metrics-sampler!` for
    ;; an explanation of how we're using these histograms.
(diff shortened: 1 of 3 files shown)


mwaggett:
I'm guessing we probably don't want to actually turn on multithreaded right now? but figured I'd leave this in for now so we can see the tests run. though I guess all the other ones will fail..

mwaggett:
with multithreaded on:
```
molly.waggett@molly:~/puppet-projects/puppetserver(ticket/master/server-2681-multithreaded-metrics-test👌 ) $ lein test :only puppetlabs.services.jruby.jruby-metrics-service-test/borrow-timeout-test

lein test puppetlabs.services.jruby.jruby-metrics-service-test
2020-01-08 10:27:10,457 WARN  [qtp2097252933-29] [p.s.j.jruby-metrics-core] JRuby instance '1' borrowed, but it appears to have already been in use!
2020-01-08 10:27:11,513 WARN  [qtp2097252933-29] [p.s.j.jruby-metrics-core] JRuby instance '1' returned, but no record of when it was borrowed!

Ran 1 tests containing 4 assertions.
0 failures, 0 errors.
```

test/integration/puppetlabs/services/jruby/jruby_metrics_service_test.clj
```diff
@@ -202,6 +203,19 @@
...
+    (atom {:num-jrubies num-jrubies
+           :num-free-jrubies num-jrubies
+           :requested-count 0
+           :borrow-count 0
+           :borrow-timeout-count 0
+           :borrow-retry-count 0
+           :return-count 0
+           :current-requested-instances 0
+           :current-borrowed-instances 0})))
+
```
Magisus: Since this is only used in one spot, can we leave it in-line like the others are?

mwaggett: sure

src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
```diff
@@ -366,7 +367,7 @@
    free-instances-fn :- IFn
    registry :- MetricRegistry]
   {:num-jrubies (metrics/register registry (metrics/host-metric-name hostname "jruby.num-jrubies")
-                  (metrics/gauge max-active-instances))
+                  (if jruby-puppet-core/multithreaded? (metrics/gauge 1) (metrics/gauge max-active-instances)))
```
Magisus: It might make more sense to change the caller here to pass in the max pool size instead of `max-active-instances`, so we don't have to do a conditional. That happens here. https://github.com/puppetlabs/puppetserver/blob/master/src/clj/puppetlabs/services/jruby/jruby_metrics_service.clj#L51-L58

mwaggett: not entirely sure what you mean, but I'm guessing it doesn't matter anymore since we decided to overload. this can just go back to `max-active-instances`.

mwaggett:
closing this, since we decided to take a different approach wrt overloading the metrics

PR #2253, opened 2020-01-09
(PUP-10222) Add test case for looking up class params via hiera data

Magisus:


Magisus:
This currently causes breakages, which hopefully will ultimately be fixed by https://github.com/puppetlabs/puppet/pull/7917

dinhthang5:
#7917

ext/thread_test/environments/production/modules/threader/manifests/params_class.pp
```diff
@@ -0,0 +1,3 @@
+class threader::params_class (String $input = 'softserve') {
```
mwaggett: Why does this have a default value but the other one doesn't?

Magisus: Oh, I was testing something. But I don't mind having them be different, actually. We prefer the data in hiera in both cases, maybe I should remove the hiera data from the one with the default to test that failed lookup and falling through behavior 🤔

mwaggett: yeah, that's what I thought you were going for at first

Magisus:
The submodule update containing the code that fixes this test case just made it into this branch. So I think this can be merged when we're happy with it.

PR #2254, opened 2020-01-09
6.3.3 release notes

jbondpdx:
plus change to http cert status page

### DIFF
--- a/documentation/http_certificate_status.md
+++ b/documentation/http_certificate_status.md
@@ -42,10 +42,12 @@ Save
 
 Change the status of the specified certificate. The desired state
 is sent in the body of the PUT request as a one-item PSON hash; the two
-allowed complete hashes are `{"desired_state":"signed"}` (for signing a
-certificate signing request; similar to `puppet cert --sign`) and
-`{"desired_state":"revoked"}` (for revoking a certificate; similar to
+allowed complete hashes are:
+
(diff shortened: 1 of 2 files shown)


documentation/http_certificate_status.md
```diff
@@ -42,11 +42,13 @@ Save
 
 Change the status of the specified certificate. The desired state
 is sent in the body of the PUT request as a one-item PSON hash; the two
-allowed complete hashes are `{"desired_state":"signed"}` (for signing a
-certificate signing request; similar to `puppet cert --sign`) and
-`{"desired_state":"revoked"}` (for revoking a certificate; similar to
+allowed complete hashes are:
+
+* `{"desired_state":"signed"}` (for signing a certificate signing request, similar to `puppet cert --sign`). To set the validity period of the signed certificate, specify the `cert_ttl` key in the body of the request, with a integer value. By default, this key specifies the number of seconds, but you can specify another time unit. See [configuration](https://puppet.com/docs/puppet/latest/configuration.html#configuration-settings) for a list of Puppet's accepted time unit markers.
```
mwaggett: I don't know if we care, but it seems like we've been adhering to a max line length in this file. This line is very long and wraps a bunch, so we might want to add line breaks for consistency?

jbondpdx: Docs has always taken the position that documentation should not have hard line breaks, even when all of our docs were Markdown, likely because we don't want to render hard line breaks on the site. 

That said, these files are owned by Team Froyo, so I'm content to defer that decision to you all.

PR #2257, opened 2020-01-09
(maint) Add CODEOWNERS file

mwaggett:


mwaggett:
Looks like it doesn't want to install these java versions??

justinstoller:
Yeah, travis doesn't support JDK7 anymore and only openjdk for 8. I don't think updating this is branch is necessary.

PR #2259, opened 2020-01-10
(maint) Fix Debian Stretch version

mwaggett:


Magisus:
Looks like this page never got updated to add Debian 10 on the master branch.

mwaggett:
yeah, I can add that when I merge up to master (since the automerge will fail the same way 5.3.x -> 6.3.x has).

PR #2262, opened 2020-01-13
(maint) Update clj-parent to 4.2.11

Magisus:
This commit updates some ambiguous type specifications that were causing
problems on Java 11.


Magisus:
This needs to go in before we can tag the 6.8.0 release.

PR #2264, opened 2020-01-13
(SERVER-2679) Fix borrow and return event logging for ReferencePool

mwaggett:


src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
```diff
@@ -277,18 +276,17 @@
...
-    (let [id (:id instance)]
-      (if-let [ta (get @borrowed-instances id)]
-        (let [elapsed-time (- (System/currentTimeMillis) (:time ta))
-              per-reason-timer (timer-for-borrow-reason metrics ta)]
-          (.update borrow-timer elapsed-time (TimeUnit/MILLISECONDS))
-          (.update per-reason-timer elapsed-time (TimeUnit/MILLISECONDS))
-          (MDC/put "jruby.borrow-time" (Long/toString elapsed-time))
-          (swap! borrowed-instances dissoc id))
-        (log/warn (trs "JRuby instance ''{0}'' returned, but no record of when it was borrowed!" id))))))
+    (if-let [ta (get @borrowed-instances worker-id)]
```
Magisus: Do you have any idea what `ta` is supposed to stand for here? Is there a better name we could use?

src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
```diff
@@ -277,18 +276,17 @@
 
 (schema/defn track-return-instance!
   [{:keys [return-count borrowed-instances borrow-timer] :as metrics} :- JRubyMetrics
-   {:keys [instance reason]} :- jruby-schemas/JRubyReturnedEvent]
```
Magisus: Oh huh, weren't using that... weird.

test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
```diff
@@ -149,7 +149,7 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Tests
 
-(deftest ^:integration ^:single-threaded-only admin-api-flush-jruby-pool-test
```
Magisus: This test really shouldn't pass... if it does, that's coincidental, since the assertions don't make sense for a pool with one instance. I still think it should have a different version written for the multithreaded pool that demonstrates we're only dealing with a single instance.

src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj
mwaggett: honestly no idea..

test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
mwaggett: oh yeah. I removed this when I saw things pass for me locally earlier this week, but I don't think I actually ran in multithreaded. I can put it back.

If there isn't a ticket for the tests that need addressing, can you open one?

test/integration/puppetlabs/services/jruby/puppet_environments_int_test.clj
```diff
@@ -61,7 +61,7 @@
 ;; not ideal, but we discussed it at length and agreed that the test has
 ;; value in terms of simulating an end user's experience, which could not
 ;; be achieved w/o some such assumptions, and thus is worth keeping.
-(deftest ^:integration ^:single-threaded-only environment-flush-integration-test
```
Magisus: If this test really passes now, that's cool, I guess it sort of demonstrates that our attempts to make the environment threadsafe worked!

Magisus: Ah yeah, this still also fails in multithreaded mode, because we try to grab a catalog from specific JRubies, which isn't possible. So this tag should go back for now.

test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
Magisus: Yes.

mwaggett: Thank you!!

test/integration/puppetlabs/services/jruby/puppet_environments_int_test.clj
mwaggett: done

Magisus:
Looks good, gonna double check that the logging errors are actually gone.

test/integration/puppetlabs/services/jruby/jruby_puppet_pool_int_test.clj
Magisus: https://tickets.puppetlabs.com/browse/SERVER-2721

Magisus:
Since this now includes the new pool for our package builds, would you mind updating the puppetserver version to 6.9.0, so we don't have to do it right before the next release?

PR #2266, opened 2020-01-17
Merge 6.3.x -> master

Magisus:
Kept all release note things from master.


Magisus:
Possibly, the script lives here if you want to see what you can do: https://github.com/puppetlabs/ci-job-configs/blob/master/resources/scripts/puppetserver-mergeup.sh Currently it auto-resolves submodule conflicts and `acceptance/config/beaker/options.rb`, both by keeping what's in the target branch. Seems like release notes can probably be added to that, now that we're not interleaving all the Z releases in later streams anymore.

PR #2268, opened 2020-01-22
(SERVER-2681) Enable metrics tests in multithreaded mode

mwaggett:


Magisus:
Excited to see what fixing these metrics does for the "average borrow time" that seems messed up right now.

PR #2270, opened 2020-01-22
(SERVER-2722) Add ppAuthCertExt to custom_extensions

shrug:
Related: https://github.com/puppetlabs/puppet/pull/7922

PUP-6258 introduced the new oid ppAuthCertExt, but this new OID was
never added to the custom_extensions method. This commit adds the
missing OID and ensures that it can be accessed

### DIFF
--- a/src/ruby/puppetserver-lib/puppet/server/certificate.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/certificate.rb
@@ -45,7 +45,8 @@ class Puppet::Server::Certificate < Puppet::SSL::Certificate
     else
       valid_oids = exts.select do |ext|
         subtree_of?(get_name_from_oid('ppRegCertExt'), ext['oid']) or
-            subtree_of?(get_name_from_oid('ppPrivCertExt'), ext['oid'])
+            subtree_of?(get_name_from_oid('ppPrivCertExt'), ext['oid']) or
+            subtree_of?(get_name_from_oid('ppAuthCertExt'), ext['oid'])
       end
 
       valid_oids.collect do |ext|


src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -680,7 +680,8 @@
       (= subject-alt-names-oid oid)
       (and (utils/subtree-of? ppAuthCertExt oid) (not (= cli-auth-oid oid)))
```
mwaggett: Looks like this is already here?

shrug: oh hey look. Apparently reading is hard. So it just needs to be added to the ruby?

mwaggett: Is the ruby being actively maintained or is that stuff deprecated? I ask because it seems to have not been updated since 2016.. but I'm still new to this repo soooo ¯\_(ツ)_/¯

Magisus: Yes we still need it. It hasn't changed because it hasn't needed to.

src/ruby/puppetserver-lib/puppet/server/certificate.rb
```diff
@@ -45,7 +45,8 @@ def custom_extensions
     else
       valid_oids = exts.select do |ext|
         subtree_of?(get_name_from_oid('ppRegCertExt'), ext['oid']) or
-            subtree_of?(get_name_from_oid('ppPrivCertExt'), ext['oid'])
+            subtree_of?(get_name_from_oid('ppPrivCertExt'), ext['oid'] or
```
Magisus: Missing paren here, at the end.

shrug: Doh, thank you

Magisus:
Looks like https://github.com/puppetlabs/puppetserver/blob/master/test/integration/puppetlabs/general_puppet/general_puppet_int_test.clj#L321 is failing because we now report the CA client extension in addition.

justinstoller:
This should target 5.3.x, I think, the related Puppet PR targeted 5.5.x.

justinstoller:
I'll retarget this and fixup the tests. Thanks, Rob!

justinstoller:
Opened https://github.com/puppetlabs/puppetserver/pull/2276

PR #2271, opened 2020-01-22
(SERVER-2721) Explain single-threaded-only test

mwaggett:
This commit adds a comment to the `environment-flush-integration-test`
explaining why it doesn't make sense to run this test (or a test like it) in
multithreaded mode.


Magisus:
Travis tests won't tell us anything for this change, merging.

PR #2274, opened 2020-01-23
(maint) Exclude :multithreaded-only tests by default

Magisus:
This commit updates the default test filter to exclude tests that are
tagged with `:multithreaded-only`, since those tests require the
`MULTITHREAED` env var to be set in order to work correctly.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -215,7 +215,8 @@
              :ci {:plugins [[lein-pprint "1.1.1"]
                             [lein-exec "0.3.7"]]}}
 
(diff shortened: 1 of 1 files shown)


mwaggett:
Can you update the commit summary to be imperative? It's not clear at first glance whether we excluded in the past and are fixing now or we have now excluded, ya know?

Magisus:
Oh, typo, thanks!

Magisus:
Did I hear you were still working on doing test builds for this?

PR #2275, opened 2020-01-27
(maint) Scoping for Docker ARGs is not intuitive

underscorgan:
For multi-stage builds, setting `ARG`s before the `FROM` makes their
scope global, but you still need to have an `ARG foo` line in each build
step to use the global `ARG foo`.

For single-stage builds, it doesn't make sense to define `ARG`s before
the `FROM` except for `ARG`s used in the `FROM`, since global scope
doesn't really help with only one stage.

### DIFF
--- a/docker/puppetserver/release.Dockerfile
+++ b/docker/puppetserver/release.Dockerfile
@@ -1,8 +1,10 @@
-ARG vcs_ref
-ARG build_date
 ARG version="6.0.0"
 ARG namespace="puppet"
 FROM "$namespace"/puppetserver-base:"$version"
+ARG vcs_ref
+ARG build_date
+ARG version
+ARG namespace
 
 LABEL org.label-schema.maintainer="Puppet Release Team <release@puppet.com>" \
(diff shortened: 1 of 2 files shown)


underscorgan:
See also https://github.com/moby/moby/issues/37345

docker/puppetserver/release.Dockerfile
```diff
@@ -1,8 +1,10 @@
-ARG vcs_ref
-ARG build_date
 ARG version="6.0.0"
 ARG namespace="puppet"
 FROM "$namespace"/puppetserver-base:"$version"
+ARG vcs_ref
```
Iristyle: Since there's only one stage in this file, why not also move the `version` and `namespace` values below the `FROM`?

underscorgan: because those are used in the `FROM` line

Iristyle: Hahahaha, I suppose that's a good reason! Sheesh.

PR #2276, opened 2020-01-28
(SERVER-2722) Add ppAuthCertExt to custom_extensions

justinstoller:
PUP-6258 introduced the new oid ppAuthCertExt, but this new OID was
never added to the custom_extensions method. This commit adds the
missing OID and ensures that it can be accessed

### DIFF
--- a/dev-resources/puppetlabs/general_puppet/general_puppet_int_test/custom_trusted_oid_mapping.yaml
+++ b/dev-resources/puppetlabs/general_puppet/general_puppet_int_test/custom_trusted_oid_mapping.yaml
@@ -6,3 +6,6 @@ oid_mapping:
   1.3.6.1.4.1.34380.1.2.3:
     shortname: 'short'
     longname: 'A Very Long Name'
+  1.3.6.1.4.1.34380.1.3.39:
+    shortname: 'pp_auth_doodad'
+    longname: 'PP Auth Doodad Long Name'
--- a/src/ruby/puppetserver-lib/puppet/server/certificate.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/certificate.rb
@@ -45,7 +45,8 @@ class Puppet::Server::Certificate < Puppet::SSL::Certificate
     else
       valid_oids = exts.select do |ext|
         subtree_of?(get_name_from_oid('ppRegCertExt'), ext['oid']) or
-            subtree_of?(get_name_from_oid('ppPrivCertExt'), ext['oid'])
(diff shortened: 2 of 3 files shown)


dev-resources/puppetlabs/general_puppet/general_puppet_int_test/custom_trusted_oid_mapping.yaml
```diff
@@ -5,4 +5,7 @@ oid_mapping:
     longname: 'Shining Finger'
   1.3.6.1.4.1.34380.1.2.3:
     shortname: 'short'
-    longname: 'A Very Long Name'
\ No newline at end of file
+    longname: 'A Very Long Name'
+  1.3.6.1.4.1.34380.1.3.39:
+    shortname: 'pp_auth_doodad'
+    longname: 'PP Auth Doodad Long Name'
```
mwaggett: How is this used in the tests?

mwaggett: like, I see the check you're doing below, just curious like.. what the other info is used for and how..

justinstoller: The longname isn't used in tests but is part of the custom oid spec: https://puppet.com/docs/puppet/latest/config_file_oid_map.html

justinstoller: Does that answer your question?

mwaggett: Partly! so the `longname` just needs to be there for Reasons™ 👍 
but what about the oid? Is that something you made up or is that specific to this extension or ???

justinstoller: It already existed on the cert (these extensions have to baked into the CSR). It just wasn't being properly displayed before because of the bug this PR fixed.

justinstoller: Both long and short names for the oid I made up. We only define names for .1 and .13 https://github.com/puppetlabs/puppet/blob/master/lib/puppet/ssl/oids.rb#L69-L70

mwaggett: ahh okay, I think I understand better now, thank you!!

PR #2278, opened 2020-02-04
(SERVER-2726) Implement thread aware Puppetserver settings

tvpartytonight:
This implementation of a Puppetserver puppet settings class wraps
the original Puppet::Settings and generates new Puppet::Settings
as necessary per compilation. If, over the course of a compilation,
the settings are modified, there will be a new ThreadLocal settings
object that will be used for the duration of that compilation.
Those modified settings will not persist after the request has
been handled.

### DIFF
--- a/spec/puppet-server-lib/puppet/jvm/settings_spec.rb
+++ b/spec/puppet-server-lib/puppet/jvm/settings_spec.rb
@@ -1,0 +1,39 @@
+require 'spec_helper'
+
+require 'puppet/server/settings'
+
+describe 'Puppet::Server::Settings' do
+
+  before :each do
+    @old_settings = Puppet.settings
+    Puppet.replace_settings_object(Puppet::Settings.new)
+    Puppet.initialize_default_settings!(Puppet.settings)
+
+
+    @new_max_errors = Puppet[:max_errors] + 10
+    puppet_config = {max_errors: @new_max_errors.to_s}
+    Puppet::Server::PuppetConfig.initialize_puppet(puppet_config: puppet_config)
+
+    Puppet.replace_settings_object(Puppet::Server::Settings.new(global_settings: Puppet.settings,
+                                                                puppet_config: puppet_config))
+  end
+
+  after :each do
+    Puppet.replace_settings_object(@old_settings)
+  end
+
+  it 'settings configured prior to replacement are kept in new threads' do
+    thread = Thread.new do
+      expect(Puppet[:max_errors]).to be(@new_max_errors)
+    end
+    thread.join
+  end
+
+  it 'does not write settings to the global config when set in threads' do
+    thread = Thread.new do
+      Puppet[:max_errors] = 0
+    end
+    thread.join
+    expect(Puppet[:max_errors]).to be(@new_max_errors)
+  end
+end
--- a/src/ruby/puppetserver-lib/puppet/server/master.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/master.rb
@@ -12,7 +12,7 @@ require 'puppet/server/network/http/handler'
 require 'puppet/server/compiler'
 require 'puppet/server/ast_compiler'
 require 'puppet/server/key_recorder'
-
+require 'puppet/server/settings'
 require 'java'
 
 ##
@@ -32,8 +32,9 @@ class Puppet::Server::Master
     # There is a setting that is routed from the puppetserver.conf to
     # configure whether or not to track hiera lookups.
     @track_lookups = puppet_server_config.delete('track_lookups')
+    multithreaded = puppet_server_config.delete('multithreaded')
     Puppet::Server::Config.initialize_puppet_server(puppet_server_config)
-    Puppet::Server::PuppetConfig.initialize_puppet(puppet_config)
+    Puppet::Server::PuppetConfig.initialize_puppet(puppet_config: puppet_config)
     # Tell Puppet's network layer which routes we are willing handle - which is
     # the master routes, not the CA routes.
     master_prefix = Regexp.new("^#{Puppet::Network::HTTP::MASTER_URL_PREFIX}/")
@@ -44,6 +45,11 @@ class Puppet::Server::Master
     @env_loader = Puppet.lookup(:environments)
     @transports_loader = Puppet::Util::Autoload.new(self, "puppet/transport/schema")
     @catalog_compiler = Puppet::Server::Compiler.new
+
+    if multithreaded
+      Puppet.replace_settings_object(Puppet::Server::Settings.new(global_settings: Puppet.settings,
+                                                                  puppet_config: puppet_config))
+    end
   end
 
   def handleRequest(request)
@@ -69,6 +75,8 @@ class Puppet::Server::Master
         body_to_return,
         response[:content_type],
         response["X-Puppet-Version"])
+  ensure
+    Puppet.settings.clear_local_settings if Puppet.settings.is_a?(Puppet::Server::Settings)
   end
 
   def compileCatalog(request_data)
--- a/src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
@@ -44,7 +44,7 @@ class Puppet::Server::PuppetConfig
       Puppet[:always_retry_plugins] = false
     end
 
-    app_defaults = Puppet::Settings.app_defaults_for_run_mode(master_run_mode.call).
+    app_defaults = Puppet::Settings.app_defaults_for_run_mode(Puppet::Util::RunMode[:master]).
         merge({:name => "master",
                :facts_terminus => 'yaml'})
     Puppet.settings.initialize_app_defaults(app_defaults)
@@ -65,7 +65,7 @@ class Puppet::Server::PuppetConfig
                                require_config: true,
                                push_settings_globally: true)
 
-    Puppet::ApplicationSupport.push_application_context(master_run_mode.call)
+    Puppet::ApplicationSupport.push_application_context(Puppet::Util::RunMode[:master])
 
     # Puppet's https machinery expects to find an object at
     # `Puppet.lookup(:ssl_context)` which it will then use as an input
@@ -109,7 +109,4 @@ class Puppet::Server::PuppetConfig
     @@oid_defns
   end
 
-  def self.master_run_mode
-    proc { Puppet::Util::RunMode[:master] }
-  end
 end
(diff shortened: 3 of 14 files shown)


tvpartytonight:
This PR won't pass any testing until https://github.com/puppetlabs/puppet/pull/7960 and https://github.com/puppetlabs/puppet/pull/7959 have been merged.

src/ruby/puppetserver-lib/puppet/server/settings.rb
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  def [](key)
+    @settings[key]
+  end
+end
+
+class Puppet::Server::Settings
+  extend Forwardable
+  THREAD_LOCAL_SETTINGS = Set.new([:strict, :code, :tasks])
```
pcarlisle: Just define this is one place and have one class reference the other

src/ruby/puppetserver-lib/puppet/server/settings.rb
```diff
@@ -0,0 +1,85 @@
...
+
+  def initialize(global_settings, puppet_config)
+    @global_settings = global_settings
+    @local_settings = Puppet::ThreadLocal.new {nil}
```
pcarlisle: i don't think you need this block argument

src/ruby/puppetserver-lib/puppet/server/settings.rb
```diff
@@ -0,0 +1,85 @@
...
+
+  def [](key)
+    if THREAD_LOCAL_SETTINGS.include?(key)
+      @compiler_context.value[key]
+    else
+      current_settings[key]
+    end
+  end
+
+  def []=(key, value)
```
tvpartytonight: When @joshcooper originally came up with this idea, this method would simply raise an error `Settings are now immutable`. This PR currently makes it possible to actually set a setting, which makes this change more backwards compatible with how settings were accessible previously. I'm open to discussion if we should make this setter available or not; the outcome of that discussion determines if we really need [this](https://github.com/puppetlabs/puppet/pull/7960) or not.

src/ruby/puppetserver-lib/puppet/server/settings.rb
```diff
@@ -0,0 +1,85 @@
...
+    @local_settings = Puppet::ThreadLocal.new {nil}
+    @puppet_config = puppet_config.reduce([]) do |acc, entry|
+      acc << "--#{entry[0]}"
+      acc << entry[1] if entry[1].kind_of?(String)
+      acc
+    end
+    @compiler_context = Puppet::ThreadLocal.new(Puppet::Server::CompilerContext.new(global_settings))
+  end
+
+  def_delegators :current_settings, :preferred_run_mode, :clear_environment_settings, :values, :parse_file,
```
pcarlisle: Was there a reason for this version instead of just delegating everything not defined here? The downside is that we can break if the implementation in puppet adds anything.

src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -44,6 +44,9 @@ def initialize(puppet_config, puppet_server_config)
     @env_loader = Puppet.lookup(:environments)
     @transports_loader = Puppet::Util::Autoload.new(self, "puppet/transport/schema")
     @catalog_compiler = Puppet::Server::Compiler.new
+
+    # This should probably be some sort of method defined in Puppet?
```
pcarlisle: is it not?

pcarlisle:
I don't understand how `@local_settings` and `@compiler_context` interact. It seems like they're competing implementations of the same thing.

src/ruby/puppetserver-lib/puppet/server/settings.rb
tvpartytonight: These delegated methods are ostensibly the "public" API of Puppet::Settings that are needed for basic compilations to work. I initially thought it would be a smaller number of delegated methods, but it seems to have grown to be a bit larger than I anticipated. Any thoughts on just inheriting from the Puppet::Settings rather than delegating all these methods?

tvpartytonight: Just moved to the stand method_missing approach to delegate the to the correct settings object.

src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
```diff
@@ -87,4 +108,8 @@ def self.initialize_puppet(puppet_config)
   def self.oid_defns
     @@oid_defns
   end
+
+  def self.master_run_mode
+    proc { Puppet::Util::RunMode[:master] }
+  end
```
Magisus: Why did this need to become a proc?

tvpartytonight: It's called in both the `initialize_puppet_settings` and  `initialize_puppet`, and I didn't want to store any state.

src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -69,6 +75,8 @@ def handleRequest(request)
         body_to_return,
         response[:content_type],
         response["X-Puppet-Version"])
+  ensure
+    Puppet.settings.clear_local_settings if Puppet.settings.is_a?(Puppet::Server::Settings)
```
tvpartytonight: One thing that might make sense here is to change this to check for a `multithreaded` instance variable that we could set during initialization. I don't know think it's much of a real difference, but I could change it if people had a preference.

spec/puppet-server-lib/puppet/jvm/settings_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+
+describe 'Puppet::Server::Settings' do
+
+  before :each do
+    @old_settings = Puppet.settings
+    Puppet.replace_settings_object(Puppet::Settings.new)
+    Puppet.initialize_default_settings!(Puppet.settings)
+
+
+    @new_max_errors = (Puppet[:max_errors] + 10)
```
mwaggett: Why parentheses here?

tvpartytonight: Oh, that's leftover from when I had that interpolated elsewhere. Whoops.

src/ruby/puppetserver-lib/puppet/server/master.rb
mwaggett: hmm, checking for `multithreaded` seems more "right", but not sure if it's worth an extra variable? maybe you could just add a comment here that this is done for thread safety?

src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
```diff
@@ -22,29 +29,43 @@ def self.initialize_puppet(puppet_config)
...
+        require_config, push_settings_globally
     )
     Puppet[:trace] = true
 
     Puppet::Server::Logger.set_log_level_from_logback
 
     # (SERVER-410) Cache features in puppetserver for performance.  Avoiding
     # the cache is intended for agents to reload features mid-catalog-run.
-    # As of (PUP-5482) setting always_retry_plugins to false implies that
+    # As of (PUP-548) setting always_retry_plugins to false implies that
```
justinstoller: this was correct before

src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
justinstoller: I don't think I follow.

I also don't think it's a valid thing to change/have changed during a running server.

src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
tvpartytonight: :doh:

I blame emacs

tvpartytonight: fixed

src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
tvpartytonight: Looking at the implementation of [Puppet::Util::RunMode](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/run_mode.rb), it doesn't look like calling it would ever change the value; a UnixRunMode object just gets stored in the Runmode class in the `@run_modes` hash.

I guess you could say this is slightly over-refactored; I wasn't quite sure what kind of object was coming back, and so I stored it in that proc since I needed it in two different spots. I don't think it needs to be a proc...but I don't think it hurts either?

Magisus:
Jenkins please test this with servertests

Magisus:
Jenkins please test this with servertests

src/ruby/puppetserver-lib/puppet/server/settings.rb
```diff
@@ -0,0 +1,52 @@
...
+  def current_settings
+    local_settings? ? @local_settings.value : @global_settings
+  end
+
+  def [](key)
+      current_settings[key]
+  end
+
+  def []=(key, value)
+    if @local_settings.value.nil?
```
mwaggett: Why not use the `local_settings?` method you wrote above?

src/ruby/puppetserver-lib/puppet/server/settings.rb
```diff
@@ -0,0 +1,52 @@
...
+  def local_settings?
+    !@local_settings.value.nil?
+  end
+
+  def current_settings
+    local_settings? ? @local_settings.value : @global_settings
+  end
+
+  def [](key)
+      current_settings[key]
```
mwaggett: this indentation is weird

src/ruby/puppetserver-lib/puppet/server/settings.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  # @param global_settings[Puppet::Settings] The puppet settings that all threads will spawn from
+  # @param puppet_config[Hash]
+  def initialize(global_settings:, puppet_config:)
+    @global_settings = global_settings
+    @local_settings = Puppet::ThreadLocal.new
+    @puppet_config = puppet_config
+  end
+
+  def method_missing(method, *args, &block)
```
mwaggett: definitely a nitpick, and possible just a personal preference, so feel free to ignore, buuut I feel like this belongs at the bottom of the class? especially since it uses a method defined below, I had to scroll down and back up to see what this was doing.

tvpartytonight: This is anecdotal, but I feel like I generally see `method_missing` defined early on in a class near the `initialize`, so that it's clear it is delegating a significant portion of method calls to some other entity.

mwaggett: that's fair

PR #2279, opened 2020-02-04
(SERVER-2724) set pe_serverversion in set_server_facts

tkishel:
Set pe_serverversion in set_server_facts as per the same in puppet.
Document the other set_server_facts in puppet,
to reduce the chance of them being out of sync.

### DIFF
--- a/src/ruby/puppetserver-lib/puppet/server/compiler.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/compiler.rb
@@ -245,9 +245,7 @@ module Puppet
         # Add our server Puppet Enterprise version, if available.
         pe_version_file = '/opt/puppetlabs/server/pe_version'
         if File.readable?(pe_version_file) and !File.zero?(pe_version_file)
-          pe_version_data = File.read(pe_version_file).chomp
-          pe_serverversion = pe_version_data.match(/(\d+\.\d+\.\d+)/)
-          @server_facts['pe_serverversion'] = pe_serverversion[1] if pe_serverversion
+          @server_facts['pe_serverversion'] = File.read(pe_version_file).chomp
         end
 
         # Add our server version to the fact list
--- a/src/ruby/puppetserver-lib/puppet/server/compiler.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/compiler.rb
@@ -238,9 +238,17 @@ module Puppet
       # Initialize our server fact hash; we add these to each catalog, and they
       # won't change while we're running, so it's safe to cache the values.
       #
-      # This is lifted directly from Puppet::Indirector::Catalog::Compiler.
+      # See also set_server_facts in Puppet::Indirector::Catalog::Compiler in puppet.
       def set_server_facts
         @server_facts = {}
+
+        # Add our server Puppet Enterprise version, if available.
+        pe_version_file = '/opt/puppetlabs/server/pe_version'
+        if File.readable?(pe_version_file) and !File.zero?(pe_version_file)
+          pe_version_data = File.read(pe_version_file).chomp
+          pe_serverversion = pe_version_data.match(/(\d+\.\d+\.\d+)/)
+          @server_facts['pe_serverversion'] = pe_serverversion[1] if pe_serverversion
+        end
 
         # Add our server version to the fact list
         @server_facts['serverversion'] = Puppet.version.to_s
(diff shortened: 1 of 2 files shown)


tkishel:
This is a POC!

Magisus:
I notice the ticket also mentions `ipaddress6`, is that going to be a separate PR?

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -238,10 +238,18 @@ def get_facts_from_pdb(nodename, environment)
       # Initialize our server fact hash; we add these to each catalog, and they
       # won't change while we're running, so it's safe to cache the values.
       #
-      # This is lifted directly from Puppet::Indirector::Catalog::Compiler.
+      # See also set_server_facts in Puppet::Indirector::Catalog::Compiler in puppet.
       def set_server_facts
         @server_facts = {}
 
+        # Add our server Puppet Enterprise version, if available.
+        pe_version_file = '/opt/puppetlabs/server/pe_version'
```
mwaggett: Is this path a setting somewhere we could read?

tkishel:
I was just trying to clean up my omission for pe_serverversion.
I could look at addressing ipaddress6, but not today ...

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -238,10 +238,18 @@ def get_facts_from_pdb(nodename, environment)
...
+        if File.readable?(pe_version_file) and !File.zero?(pe_version_file)
+          pe_version_data = File.read(pe_version_file).chomp
+          pe_serverversion = pe_version_data.match(/(\d+\.\d+\.\d+)/)
```
mwaggett: What else could be in the `pe_version_file`? I'm slightly wary of making assumptions about what the version could look like, since there was some talk of different versioning schemes (e.g. 2019.4 as opposed to 2019.4.0), though I bet those assumptions are EVERYWHERE..

src/ruby/puppetserver-lib/puppet/server/compiler.rb
tkishel: It appears to be hardcoded everywhere:

https://github.com/puppetlabs/puppet-enterprise-modules/search?q=%22%2Fopt%2Fpuppetlabs%2Fserver%2Fpe_version%22&unscoped_q=%22%2Fopt%2Fpuppetlabs%2Fserver%2Fpe_version%22

src/ruby/puppetserver-lib/puppet/server/compiler.rb
tkishel: I haven't found a specification.
Certainly could loosen up the regex.

mwaggett: I'm wondering if just reading the file is enough? Not sure what else might be in there.

mwaggett: oh, I see this is literally exactly what's in puppet.. I guess we should probably keep them the same.

src/ruby/puppetserver-lib/puppet/server/compiler.rb
tkishel: It's "safe" because it's hardcoded everywhere :(

src/ruby/puppetserver-lib/puppet/server/compiler.rb
tkishel: It's a valid point. I'm open to dropping the match in both. 
I don't mind doing that work, if necessary.

joshcooper: In puppet we generally try to parse version strings either using `Gem::Version.new` or `SemanticPuppet::Version.parse`. I don't know if one is a subset of the other? I assume both are available server-side? Might be good to switch both puppet and puppetserver.

tkishel: Perhaps we should just read the file, and defer to the code that writes the file in regard to the validity of its contents, since this code doesn't use the version itself?

```
    pe_version_file = '/opt/puppetlabs/server/pe_version'
    if File.readable?(pe_version_file) and !File.zero?(pe_version_file)
      @server_facts['pe_serverversion'] = File.read(pe_version_file).chomp
    end
```


I can easy add this to the commits for both puppet and here in puppetserver.

tkishel: yolo

mwaggett:
holding off on merge since @joshcooper had some thoughts on the version parsing

mwaggett:
okay since Josh merged https://github.com/puppetlabs/puppet/pull/7965, I'm calling this good

PR #2282, opened 2020-02-14
(maint) Condition calls to new method/arity in Puppet

justinstoller:


src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -47,8 +47,14 @@ def initialize(puppet_config, puppet_server_config)
     @catalog_compiler = Puppet::Server::Compiler.new
 
     if multithreaded
```
mwaggett: Are we running multithreaded in CI?

src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
```diff
@@ -24,14 +24,19 @@ def self.initialize_puppet_settings(puppet_config:, require_config:, push_settin
...
-        end,
-        require_config, push_settings_globally
-    )
+    cli_flags = puppet_config.reduce([]) do |acc, entry|
+      acc << "--#{entry[0]}"
+      acc << entry[1] if entry[1].kind_of?(String)
+      acc
+    end
+
+    # We check parameter length because Method#arity returns -1 for varargs
```
mwaggett: Maybe good to include *why* this check is necessary at all? or is that overkill?

src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
```diff
@@ -24,14 +24,19 @@ def self.initialize_puppet_settings(puppet_config:, require_config:, push_settin
...
+    if Puppet.method(:initialize_settings).parameters.length == 3
+      Puppet.initialize_settings(cli_flags, require_config, push_settings_globally)
```
mwaggett: I suppose it's too late to make changes to puppet, but we probably should've just made these optional args?

src/ruby/puppetserver-lib/puppet/server/master.rb
tvpartytonight: yes. I think so? at least we are in travis...

mwaggett: I know we are in Travis, I guess I was wondering how we encountered this issue in the Jar Jar tests. but I guess it's probably not this part but the `initialize_settings` part, which isn't inside a `multithreaded` conditional, that caused the error.

mwaggett: so I guess I answered my own question lol

Magisus: Not yet, apart from the travis unit tests.

src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
justinstoller: Just to be pedantic, they are already optional, but the previous method takes 0-2 args and new one takes 0-3 args (that variable argument-ness is why arity return -1 for both signatures). But I know what you mean.

mwaggett: ugh that's annoying.. I feel like having a variable number of args is supposed to prevent breakages exactly like this??

mwaggett: oh wait, I see what you're saying

mwaggett: we're calling it with the 3rd one here

tvpartytonight: They are optional in [Puppet](https://github.com/puppetlabs/puppet/blob/master/lib/puppet.rb#L154), but I think the problem is when running the server with an older version of Puppet that isn't aware of the 3rd positional argument, `push_settings_globally`.

src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
justinstoller: I think it is, but I'm the one who did it this way to begin with. : D   Or at least, I'd put the rationale for a patch in the commit, and non-obvious peculiarities of the implementation in code comments.

Opinions @Magisus, @tvpartytonight ?

tvpartytonight: I think that you answered why you use length rather than arity, but you don't answer why you use length at all? Maybe something like 

> Check the method arguments because of additional arguments added in Puppet 6.13.0.

src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
mwaggett: yep 👍

src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
mwaggett: That's fair. Then maybe I think the commit message could use a little more detail. 😝

PR #2283, opened 2020-02-15
(SERVER-2727) Use Facter NG as our facter submodule

Magisus:
We have been rewriting Facter from C++ back into Ruby, so puppetserver
can finally take up a new Facter version for testing and repl workflows.
This commit updates our facter submodule to point at the new version.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,12 +1,12 @@
 [submodule "ruby/puppet"]
 	path = ruby/puppet
 	url = https://github.com/puppetlabs/puppet.git
-[submodule "ruby/facter"]
-	path = ruby/facter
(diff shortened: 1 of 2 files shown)


Magisus:
This is spitting out Facter logs at debug right now when testing, which doesn't seem desirable. I'm following up with Night's Watch about ways to fix this.

Magisus:
Updating to latest Facter 4, resolving an issue where Facter 2 was getting installed by Bundler, and fixing some test mocking seems to have cleaned up most of the issues for me. There's still some extraneous logging in the clojure tests that we might want to deal with.

Magisus:
Saw these tests go green, squashing commits. This should be ready to go.

PR #2284, opened 2020-02-15
(SERVER-2677) Make initial puppet run more resilient to CI Load

tvpartytonight:
When first initializing ssl for an agent, there are sporadic issues
with the autosigning; the output in the puppet logs indicate that
the autosigning didn't work, and the cert is still not signed.

This change tries one more time after 3 seconds; hopefully, allowing
for more time will give the server the time it needs to sign the
certificate.


Magisus:
Travis isn't going to tell us anything here, going to merge.

PR #2286, opened 2020-02-20
Use 'puppet ssl bootstrap' instead of 'puppet agent' to retrieve certificate

baurmatt:


underscorgan:
Thank you @baurmatt !

docker/puppetserver-base/docker-entrypoint.d/80-ca.sh
```diff
@@ -40,7 +39,7 @@ EOF
       sleep 1
     done
 
-    puppet agent --noop --server="${CA_HOSTNAME}" --masterport="${CA_MASTERPORT}"
+    puppet ssl bootstrap --server="${CA_HOSTNAME}" --masterport="${CA_MASTERPORT}"
```
joshcooper: if the cert hasn't been signed yet, then `puppet ssl bootstrap` will retry the client cert download every `waitfor cert` seconds. Is that ok/desired? If you want it to just try once, you can do `puppet ssl bootstrap --waitforcert 0`. There's also a `maxwaitforcert` option to prevent the agent from waiting indefinitely.

baurmatt: Isn't that the same for `puppet agent --noop ...` 🤔 From my point of view: This seems fine! :)

PR #2287, opened 2020-02-21
(FACT-2390) Update facter submodule to f81a7066c98e2c47e0f9f9f17b709c…

Magisus:
…cf34cbf52e

This update fixes an issue where recusion checking when resolving facts
was not threadsafe.


Magisus:
No, only puppet is. I think it's changed like 3 times in the last two years.

PR #2288, opened 2020-02-21
(PUP-10266) Add basic testing for environment cache invalidation 

mwaggett:
This commit adds a test to ensure that the environment cache is not invalidated
in the midst of a compile using the old version of the environment.


ext/test/cache_expiration_test.rb
```diff
@@ -0,0 +1,75 @@
...
+    @environment     = 'production'
+    @environment_uri = URI("http://localhost:8140/puppet/v3/catalog/#{@catalog}?environment=#{@environment}")
+    @cache_uri       = URI("http://localhost:8140/puppet-admin-api/v1/environment-cache")
+  end
+
+  def modify_environment
+    modules_path = File.join(__dir__, 'environments', @environment, 'modules')
+    tmp_path = File.join(__dir__, 'for-now')
+    FileUtils.mkdir_p(tmp_path)
+    FileUtils.mv(modules_path, tmp_path)
```
Magisus: Are you planning to then swap in different content?

mwaggett: possibly? first I was just trying to see if it could continue to find these modules after I'd moved them, but that might be flawed approach..

justinstoller:
Is this still a thing?

Magisus:
I don't think so... I feel like we figured out it wasn't going to work the way I thought when Molly and I talked through this.

mwaggett:
ahh yeah nah, I'll close this

PR #2294, opened 2020-03-11
(RE-13304) Switch to using beaker subcommands

Magisus:
This commit updates our test scripts to use beaker subcomands, so we can
take advantage of some setup that occurs in the beaker helpers when
using `beaker provision`.


justinstoller:
This seems like a good change though I think in CI we use https://github.com/puppetlabs/ci-job-configs/blob/master/resources/scripts/integration-beaker.sh don't we?

Magisus:
Ah yeah, dang, looks like we do, nice catch. I bet updating that will affect a lot of jobs... Might give it a shot anyway though. Seems like the right kind of migration we need to be making.

Magisus:
Here's the CJC PR: https://github.com/puppetlabs/ci-job-configs/pull/6893. Going to probably need broader review on that one.

PR #2295, opened 2020-03-11
Allow executables in Docker entrypoint; avoid chmod; fix run ordering

runejuhl:
Because the script is run with `set -e` executing `chmod` on a read-only file
will make the script exit. With these changes `chmod` is only run for files that
end with `.sh` and that are not already executable.

As `find` doesn't guarantee any particular ordering `sort` has been added to the
execute pipelines to ensure that executables are run in correct (numeric) order.

### DIFF
--- a/docker/puppetserver-base/docker-entrypoint.sh
+++ b/docker/puppetserver-base/docker-entrypoint.sh
@@ -2,20 +2,32 @@
 
 set -e
 
-chmod +x /docker-entrypoint.d/*.sh
-# sync prevents aufs from sometimes returning EBUSY if you exec right after a chmod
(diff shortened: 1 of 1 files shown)


Iristyle:
I think we're actually better off just requiring a user has put a proper `+x` on files rather than making assumptions.

I think the Postgres container does this right -- https://github.com/docker-library/postgres/blob/master/docker-entrypoint.sh#L146-L172

* if `+x`, file is executed
* otherwise, source the script

Magisus:

justinstoller:
Feels like @Iristyle shot this down and its just been hanging out in our backlog since. I'm going to close this and let you two hash it out if there's work in here that should be kept.

PR #2303, opened 2020-03-23
(SERVER-2758) Update jruby-utils to 3.1.0

Magisus:
This update also bumps JRuby to 9.2.11.0.

### DIFF
--- a/dev-setup
+++ b/dev-setup
@@ -21,7 +21,7 @@ do
 done < ./resources/ext/build-scripts/jruby-gem-list.txt
 
 echo "Installing vendored gems to '${VENDORED_GEMS_HOME}'"
-lein gem install --install-dir "${VENDORED_GEMS_HOME}" --no-ri --no-rdoc "${gem_list[@]}"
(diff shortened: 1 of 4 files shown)


Magisus:
Still trying to track down the issue where the uberjar built by ezbake does not contain the overridden version of jruby-utils, pulling in the version from clj-parent instead.

dev-setup
```diff
@@ -21,7 +21,7 @@ do
 done < ./resources/ext/build-scripts/jruby-gem-list.txt
 
 echo "Installing vendored gems to '${VENDORED_GEMS_HOME}'"
-lein gem install --install-dir "${VENDORED_GEMS_HOME}" --no-ri --no-rdoc "${gem_list[@]}"
+lein gem install --install-dir "${VENDORED_GEMS_HOME}" --no-document "${gem_list[@]}"
```
mwaggett: nitpick, but perhaps these should go in a separate commit.

Magisus:
See last commit for details on how this got around the issue when ezbaking.

Magisus:
Looks like packaging uses `no-ri` and is apparently struggling now?

justinstoller:
Travis is broken

PR #2304, opened 2020-03-26
(SERVER-2763) Pass no-document flag when installing gems

Magisus:
This is a workaround to get our pipelines unblocked. Ultimately we would
like to revert this and have the test still pass.


Magisus:
Added a commit that _might_ help with the issues Jenkins is seeing when running the ruby specs. I don't really have a good way to test this pre-merge, because it only occurs on certain platforms. But this replacement worked well for me locally.

PR #2305, opened 2020-03-27
(DOC-4476) Add basic CA and external root diagram to repo

clairecadman:


Magisus:
Ooh, excited we're getting these up. We're not using the 6.3.x branch anymore though, would you mind retargeting this at master? I tried just doing it in the GH UI, but it's bringing along some commits we don't want.

clairecadman:
@Magisus Actually apparently these aren't the final diagrams. Let me follow up with Joe first, and then I'll add to master. I'll keep you posted!

PR #2306, opened 2020-03-30
(SERVER-2660) Update jruby to 9.2.11.1

Magisus:
This update fixes a `sprintf` buffer bug.


Magisus:
Figured since we updated JRuby we should actually do a Y release next.

PR #2307, opened 2020-03-30
(SERVER-2718) Return authoriziation extensions with cert status response

Magisus:
This adds a new field to the GET response for the
`certificate_status` endpoint, listing the authorization extensions
present on the cert or CSR being queried. This will be used to help
inform users of potentially risky extensions, so they can make an
informed decision about whether or not to sign a certificate.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -11,6 +11,7 @@
   (:require [me.raynes.fs :as fs]
             [schema.core :as schema]
             [clojure.string :as str]
+            [clojure.set :as set]
             [clojure.java.io :as io]
             [clojure.tools.logging :as log]
             [clj-time.core :as time]
@@ -109,6 +110,7 @@
               :state             CertificateState
               :dns_alt_names     [schema/Str]
               :subject_alt_names [schema/Str]
+              :authorization_extensions {schema/Str schema/Str}
               :fingerprint       schema/Str
               :fingerprints      {schema/Keyword schema/Str}}]
     ;; The map should either have all of the CertificateDetails keys or none of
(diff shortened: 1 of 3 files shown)


Magisus:
I'd like to add a test that verifies you get this info for a cert too, not just a CSR.

Magisus:
Updated with cert test.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -109,6 +110,7 @@
               :state             CertificateState
               :dns_alt_names     [schema/Str]
               :subject_alt_names [schema/Str]
+              :authorization_extensions {schema/Str schema/Str}
```
mwaggett: if we're already cleaning up indentation.. might wanna realign everything with this 😅

Magisus: We decided to stop doing aligning maps like that. It causes too much whitespace churn each time a key is added.

The indentation I fixed is what was needed for my IDE to do fancy parenthesis things correctly.

mwaggett: ah okay, that's fair!

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
```diff
@@ -465,8 +464,78 @@
...
+         :certificate-authority {:allow-authorization-extensions true}}
+        (testing "Auth extensions on a CSR"
+          (let [response (http-client/get
+                           (str "https://localhost:8140"
+                                "puppet-ca/v1/certificate_status/test_cert_with_auth_ext")
+                           {:ssl-cert (str bootstrap/master-conf-dir "/ssl/ca/ca_crt.pem")
+                            :ssl-key (str bootstrap/master-conf-dir "/ssl/ca/ca_key.pem")
+                            :ssl-ca-cert (str bootstrap/master-conf-dir "/ssl/ca/ca_crt.pem")
+                            :as :text
+                            :headers {"content-type" "application/json"}})
```
mwaggett, line 511: Does it make sense to put this inside a var of some sort, since it's used in a bunch of these requests?

Magisus: I'll look into it.

Magisus: See what you think of the commit I added. If you think it's better I'll squash it.

mwaggett: I like it ¯\_(ツ)_/¯

PR #2308, opened 2020-03-31
(DOC-4476) Add CA diagrams

clairecadman:
I'm not sure about the external diagram, should this be for PE? If so I think we only have the one diagram for FOSS. All the diagrams are listed in this ticket:

https://tickets.puppetlabs.com/browse/DOC-4476


Magisus:
The docs about setting up the external CA are in the FOSS section. As stated in the ticket, both diagrams should ultimately go on https://puppet.com/docs/puppetserver/latest/intermediate_ca.html, in the relevant sections.

Magisus:
Can you rename the files to drop the `(1)` in the name?

Magisus:
I re-uploaded the FOSS external root diagram as "CA - External Root - FOSS". This PR should have that one instead of "CA - External Root"

Magisus:
Would you mind also renaming the other diagram to not have the `(1)` in the path name?

mwaggett:
Closing in favor of https://github.com/puppetlabs/puppetserver/pull/2310

PR #2309, opened 2020-03-31
(SERVER-2660) Update to clj-parent 4.4.1

Magisus:
This update brings in the new jruby-utils version, so this commit also
removes the jruby-utils pin.


Magisus:
Updated to new public clj-parent version.

PR #2313, opened 2020-03-31
(maint) Remove default `set -e ` for gh actions shell

underscorgan:
Skipped release build without failing the job at https://github.com/puppetlabs/puppetserver/runs/550391873?check_suite_focus=true

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -34,6 +34,7 @@ jobs:
         env:
           IS_RELEASE: true
         working-directory: docker
+        shell: bash {0}
         run: |
           make prep
           if [ $? -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


underscorgan:
@mwaggett that looks like it failed in java11 puppetserver tests which is not related to this fix?

mwaggett:
@underscorgan That push job that's failing looks like a docker thing? Honestly not sure what that job is or why it's running.

underscorgan:
I think that push job got tied to this somehow but is pulling from the branch I initially had pushed upstream to test out the actions stuff, it's pulling from a branch/commit that no longer exist, so I think it should be ignored

Magisus:
Does this only apply to master? I've lost track of where we do Docker stuff.

underscorgan:
yep, only master

PR #2315, opened 2020-04-02
(DOC-4476) Add CA diagrams

clairecadman:
@Magisus Let me know if the description of the diagrams is accurate and whether you would like to add / change anything else. Thanks!


Magisus:
These look awesome, Claire, so exciting!

PR #2317, opened 2020-04-04
(DOCS) Release notes for Puppet Server 6.10.0

clairecadman:
This commit contains a known issue for Puppet Server 6.10.0, as part of an emergency PE release.

### DIFF
--- a/documentation/release_notes.markdown
+++ b/documentation/release_notes.markdown
@@ -18,7 +18,7 @@ Released 14 April 2020
 
 - The `GET /certificate_status` endpoint now returns certificate or CSR's authorization extensions. [SERVER-2718](https://tickets.puppetlabs.com/browse/SERVER-2718)
 
-- Puppet's "ppRegCertExt" arc has been extended with OID "1.3.6.1.4.1.34380.1.1.26" and the short name "pp_owner". This OID is meant to help users in cloud environments, and the short name will be displayed when using the `puppetserver ca` cli tool.
+- Puppet's `ppRegCertExt` arc has been extended with OID `1.3.6.1.4.1.34380.1.1.26` and the short name `pp_owner`. This OID is meant to help users in cloud environments. The short name will be displayed when using the `puppetserver ca` cli tool.
 
 ### Resolved issues
 
@@ -26,9 +26,7 @@ Released 14 April 2020
 
 ### Known issues 
 
-- An update to JRuby 9.2.11.1 has caused a change in defaults when installing gems with the `puppetserver gem` command. It attempts to install documentation by default, but this will not work. To avoid this bug, pass `--no-document` when installing gems. [SERVER-2758](https://tickets.puppetlabs.com/browse/SERVER-2758).
-
-- Part of the `$LOAD_PATH` for Puppet Server is no longer valid when being passed to `File.expand_path`. This will cause Rubygems' `Gem.list_files()` and calls to `Dir.glob()` where the invalid portion of the $LOAD_PATH is passed as the `base:` parameter to fail. This causes features that depend on this functionality (like RDoc, the tool Rubygems uses to install documentation) to fail.
+- An update to JRuby 9.2.11.1 has caused a change in defaults when installing gems with the `puppetserver gem` command. It attempts to install documentation by default, but this will not work. To avoid this bug, pass `--no-document` when installing gems. This is caused by an inability to use the `classpath:/puppetserver-lib` portion of the `$LOAD_PATH` as a parameter to `Gem.list_files` or `Dir.glob`, which Rdoc relies on to install documentation. [SERVER-2758](https://tickets.puppetlabs.com/browse/SERVER-2758).
 
 ## Puppet Server 6.9.2
 
--- a/documentation/release_notes.markdown
+++ b/documentation/release_notes.markdown
@@ -18,7 +18,7 @@ Released 14 April 2020
 
 - The `GET /certificate_status` endpoint now returns certificate or CSR's authorization extensions. [SERVER-2718](https://tickets.puppetlabs.com/browse/SERVER-2718)
 
-- Puppet's `ppRegCertExt` arc has been extended with OID `1.3.6.1.4.1.34380.1.1.26` and the short name `pp_owner`. This OID is meant to help users in cloud environments. The short name will be displayed when using the `puppetserver ca` cli tool.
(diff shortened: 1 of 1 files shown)


documentation/release_notes.markdown
```diff
@@ -10,6 +10,14 @@ canonical: "/puppetserver/latest/release_notes.html"
 [puppetserver.conf]: ./config_file_puppetserver.markdown
 [product.conf]: ./config_file_product.markdown
 
+## Puppet Server 6.10.0
+
+Released 3 April 2020
```
mwaggett, line 15: This date will need to be updated.

clairecadman: Will do! When is this going out?

Magisus:
Should this also include the other release notes for this release https://tickets.puppetlabs.com/issues/?jql=fixVersion%20%3D%20%22SERVER%206.10.0%22%20? Or is that being done separately?

clairecadman:
I've added these other notes. Let me know when you know the date. Thanks!

documentation/release_notes.markdown
```diff
@@ -10,13 +10,31 @@ canonical: "/puppetserver/latest/release_notes.html"
...
+### Resolved issues
+
+- Using a precision number to truncate a string in Puppet's `sprintf` function no longer interpolates extra characters.[SERVER-2660](https://tickets.puppetlabs.com/browse/SERVER-2660).
+
+### Known issues 
+
+- An update to JRuby 9.2.11.1 has caused a change in defaults when installing gems with the `puppetserver gem` command. It attempts to install documentation by default, but this will not work. To avoid this bug, pass `--no-document` when installing gems. [SERVER-2758](https://tickets.puppetlabs.com/browse/SERVER-2758).
+
+
+
```
mwaggett, line 30: nitpick, but looks like we usually only have one empty line between versions

clairecadman: Fixed, thanks for spotting!

documentation/release_notes.markdown
```diff
@@ -10,13 +10,29 @@ canonical: "/puppetserver/latest/release_notes.html"
 [puppetserver.conf]: ./config_file_puppetserver.markdown
 [product.conf]: ./config_file_product.markdown
 
+## Puppet Server 6.10.0
+
+Released 9 April 2020
```
mwaggett, line 15: Looks like April 14 is the current target date

documentation/release_notes.markdown
```diff
@@ -26,6 +28,8 @@ Released 14 April 2020
 
 - An update to JRuby 9.2.11.1 has caused a change in defaults when installing gems with the `puppetserver gem` command. It attempts to install documentation by default, but this will not work. To avoid this bug, pass `--no-document` when installing gems. [SERVER-2758](https://tickets.puppetlabs.com/browse/SERVER-2758).
 
+- Part of the `$LOAD_PATH` for Puppet Server is no longer valid when being passed to `File.expand_path`. This will cause Rubygems' `Gem.list_files()` and calls to `Dir.glob()` where the invalid portion of the $LOAD_PATH is passed as the `base:` parameter to fail. This causes features that depend on this functionality (like RDoc, the tool Rubygems uses to install documentation) to fail.
+
```
Magisus, line 32: Out of curiosity (and for posterity) which part is invalid?

documentation/release_notes.markdown
```diff
@@ -10,13 +10,33 @@ canonical: "/puppetserver/latest/release_notes.html"
...
 
+## Puppet Server 6.10.0
+
+Released 14 April 2020
+
+### New features
+
+- The `GET /certificate_status` endpoint now returns certificate or CSR's authorization extensions. [SERVER-2718](https://tickets.puppetlabs.com/browse/SERVER-2718)
+
+- Puppet's "ppRegCertExt" arc has been extended with OID "1.3.6.1.4.1.34380.1.1.26" and the short name "pp_owner". This OID is meant to help users in cloud environments, and the short name will be displayed when using the `puppetserver ca` cli tool.
```
mwaggett, line 21: should this use backticks instead of quotes? not really sure if there are conventions around that.

mwaggett: also, I don't think a comma belongs in that second sentence. either no comma or make it two sentences, imo.

documentation/release_notes.markdown
justinstoller: `"classpath:/puppetserver-lib"`. I assumed that was too much detail for the ticket but I can add it if you want. I will definitely add it to the ticket.

Relatedly, should I add the same ticket reference to this blurb?

documentation/release_notes.markdown
justinstoller: I can do that. I'm not really sure what the conventions are.

documentation/release_notes.markdown
mwaggett: If there's a corresponding ticket, I say def add it..

I think the piece of the load path is more relevant than how the exact error plays out?

justinstoller: So this new blurb is the root cause of the gem doc thing. ie the thing in parens is trying to relate this with the blurb above it.

mwaggett: ahhh, that was not clear to me

justinstoller: Love suggestions on how to make it clearer  ( :

mwaggett: An update to JRuby 9.2.11.1 has caused a change in defaults when installing gems with the `puppetserver gem` command. It attempts to install documentation by default, but this will not work. To avoid this bug, pass `--no-document` when installing gems. This is caused by an inability to expand the `classpath:/puppetserver-lib` portion of the `$LOAD_PATH`, which Rdoc relies on to install documentation.

documentation/release_notes.markdown
```diff
@@ -10,13 +10,31 @@ canonical: "/puppetserver/latest/release_notes.html"
...
 
+## Puppet Server 6.10.0
+
+Released 14 April 2020
+
+### New features
+
+- The `GET /certificate_status` endpoint now returns certificate or CSR's authorization extensions. [SERVER-2718](https://tickets.puppetlabs.com/browse/SERVER-2718)
+
+- Puppet's `ppRegCertExt` arc has been extended with OID `1.3.6.1.4.1.34380.1.1.26` and the short name `pp_owner`. This OID is meant to help users in cloud environments. The short name will be displayed when using the `puppetserver ca` cli tool.
```
Magisus, line 21: cli -> CLI ?

justinstoller: Updated

PR #2320, opened 2020-04-08
(SERVER-2591) Update CA CLI gem to 1.7.0

Magisus:
This update will display a certificate or CSR's authorization extensions
when they are available from the API.


Magisus:
This can go into 5.x eventually too, but since the API update didn't land there, I wasn't going to bother for now.

Magisus:
We decided to do the 5.x update.

PR #2324, opened 2020-04-10
(test) See what it takes for deserializing bolt types

donoghuc:
DO NOT MERGE: This is just an illustration of the problem of Bolt types bount to plan vars in compiling apply blocks.


donoghuc:
Only opened this to make it easier to discuss https://gist.github.com/donoghuc/3049bf58d656e65cbdf96da3125343bc

PR #2328, opened 2020-04-23
(maint) Refactor + simplify Dockerfile build stages / enable Docker Buildkit / update README ENV vars

Iristyle:
- Updates all cells to run on Bionic instead of Xenial
 - Requires updating to a 19.03+ version of Docker in TravisCI
    This upgrade takes roughly 30 seconds, but typically still saves
    time in the end since Buildkit is faster.

    The expectation is that Travis will soon support Ubuntu 20.04 LTS
    "fossa", which should include a newer version of Docker

 - Windows already supports buildkit with WSL2
 - Docker upgrade Travis compose to 1.25.5
 - Adjust Docker hadolint linting
 - Optimize Dockerfile builds around buildkit, by moving from 3 separate Dockerfiles (including 1 base) to a single Dockerfile with stages
 - Remove Dockerfile duplication amongst stages
 - Minor Dockerfile refactor to set ENV / ARG where it makes sense
 - Refactor Dockerfile "edge" and "release" so that they mostly share the final install step
 - Dockerfile change how dumb-init downloaded / upgrade dumb-init to 1.2.2 from 1.2.1
 - Docker doesn't need wget
 - Remove TravisCI HipChat notifications since we don't use HipChat
 - Use ADD to install lein instead of curl inside a RUN
 - Refactor packages out to ARG like other containers
 - Update README to use a table for config values like other projects (for consistency)
 - Add settable ENV vars to Dockerfile where missing


Iristyle:
The Travis fail here is peculiar:

```
bundler: command not found: --gemfile
Install missing gem executables with `bundle install`
Makefile:84: recipe for target 'test' failed
```

PR #2331, opened 2020-04-28
(maint) Update Puppet Agent version for 5.x compat job

Magisus:
We recently removed some of the older versions of the puppet5 stream
from our internal repos, so this commit updates us to the latest of that
stream to test against.


Magisus:
I verified that this is the recommended path forward from RE for now.

PR #2329, opened 2020-04-27
(PE-28877) Teach `ASTCompiler` about bolt

donoghuc:
Previously when the compile endpoint was asked to compile AST with references to Bolt objects the compile would fail. This commit updates the compiler to accept serialized pcore and to deserialize it for compilation. This adds the requirement of bolt library code (and some supporting gems [logging, addressable, publis_suffix]) as well as the boltlib module. Part of making the serialized types perform as expected the compiler now creates specialized bolt inventory to back operations on bolt Target options.

PR #2332, opened 2020-04-29
(PE-28877, PE-28941) Add option to compile-ast endpoint to load bolt types

donoghuc:
Previously when the compile endpoint was asked to compile AST with references to Bolt objects the compile would fail. This commit updates the compiler to accept serialized pcore and to deserialize it for compilation. This adds the requirement of some Bolt library code (and some supporting gems [addressable, public_suffix]) as well as the boltlib module. Part of making the serialized types perform as expected the compiler now creates specialized bolt inventory to back operations on bolt Target options. This work depends on two new additions to the `jruby-puppet` configuration section for pe-puppetserver. The first is appending the path to the bolt library code in the `ruby-load-path` array. The second is a new setting called `boltlib-path` which is an array pointing to the path of any bolt modules to prepend to the modulepath for compilation. The compile request now accepts an option to load bolt types. The default behavior is to *not* attempt to load bolt types.

### DIFF
--- a/documentation/puppet-api/v3/compile.markdown
+++ b/documentation/puppet-api/v3/compile.markdown
@@ -36,7 +36,8 @@ The request body must look like:
   "transaction_uuid": "<uuid string>",
   "job_id": "<id string>",
   "options": { "capture_logs": <true/false>,
-               "log_level": <one of debug/info/warning/err>}
+               "log_level": <one of debug/info/warning/err>
+               "compile_for_plan": <true/false> }
 }
 ```
 
--- a/spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
+++ b/spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
@@ -33,32 +33,42 @@ end
 
 describe Puppet::Server::ASTCompiler do
   context 'when compiling AST' do
+    let(:boltlib_path) { nil }
+
     it 'handles basic resources' do
-      response = Puppet::Server::ASTCompiler.compile(request("notify { 'my_notify': }"))
+      response = Puppet::Server::ASTCompiler.compile(request("notify { 'my_notify': }"), boltlib_path)
       notify = find_notify(response[:catalog])
       expect(notify).not_to be_nil
       expect(notify['title']).to eq("my_notify")
     end
 
     it 'correctly interpolates supplied variables' do
-      response = Puppet::Server::ASTCompiler.compile(request('notify { "$foo": }'))
+      response = Puppet::Server::ASTCompiler.compile(request('notify { "$foo": }'), boltlib_path)
       notify = find_notify(response[:catalog])
       expect(notify).not_to be_nil
       expect(notify['title']).to eq("bar")
     end
 
     it 'correctly interpolates supplied facts' do
-      response = Puppet::Server::ASTCompiler.compile(request('notify { "$my_fact": }'))
+      response = Puppet::Server::ASTCompiler.compile(request('notify { "$my_fact": }'), boltlib_path)
       notify = find_notify(response[:catalog])
       expect(notify).not_to be_nil
       expect(notify['title']).to eq("fact_value")
     end
 
     it 'correctly interpolates supplied trusted facts' do
-      response = Puppet::Server::ASTCompiler.compile(request('notify { "${trusted[\'my_trusted\']}": }'))
+      response = Puppet::Server::ASTCompiler.compile(request('notify { "${trusted[\'my_trusted\']}": }'), boltlib_path)
       notify = find_notify(response[:catalog])
       expect(notify).not_to be_nil
       expect(notify['title']).to eq("trusted_value")
+    end
+
+    it 'fails gracefully when boltlib is not found when request requires bolt types' do
+      bolt_request = request('notify { "${trusted[\'my_trusted\']}": }')
+      bolt_request.merge!({'options' => { 'compile_for_plan' => true } })
+      expect {
+        Puppet::Server::ASTCompiler.compile(bolt_request, boltlib_path)
+      }.to raise_error(Puppet::Error, /the path to boltlib modules must be provided/)
     end
   end
 end
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
@@ -53,7 +53,11 @@
 
     * :use-legacy-auth-conf - Whether to use the legacy core Puppet auth.conf
         (true) or trapperkeeper-authorization (false) to authorize requests
-        being made to core Puppet endpoints."
+        being made to core Puppet endpoints.
+
+    * :boltlib-path - Optional array containing path(s) to bolt modules. This path
+         will be prepended to AST compilation modulepath. This is required for
+         compiling AST that contains bolt types."
   {:master-conf-dir (schema/maybe schema/Str)
    :master-code-dir (schema/maybe schema/Str)
    :master-var-dir (schema/maybe schema/Str)
@@ -67,5 +71,6 @@
    :http-client-idle-timeout-milliseconds schema/Int
    :http-client-metrics-enabled schema/Bool
    :max-requests-per-instance schema/Int
-   :use-legacy-auth-conf schema/Bool})
+   :use-legacy-auth-conf schema/Bool
+   (schema/optional-key :boltlib-path) [schema/Str]})
 
--- a/src/clj/puppetlabs/services/legacy_routes/legacy_routes_service.clj
+++ b/src/clj/puppetlabs/services/legacy_routes/legacy_routes_service.clj
@@ -45,7 +45,8 @@
                                                                    (throw (IllegalStateException.
                                                                             (i18n/trs "Versioned code not supported."))))
                                                                  (constantly nil)
-                                                                 false))
+                                                                 false
+                                                                 nil))
           master-route-handler (comidi/routes->handler master-routes)
           master-mount (master-core/get-master-mount
                         master-ns
--- a/src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
@@ -5,7 +5,7 @@ require 'puppet/server/logging'
 module Puppet
   module Server
     class ASTCompiler
-      def self.compile(compile_options)
+      def self.compile(compile_options, boltlib_path)
         options = compile_options['options'] || {}
 
         log_level = options['log_level']
@@ -13,41 +13,122 @@ module Puppet
 
         if options['capture_logs']
           catalog, logs = Logging.capture_logs(log_level) do
-            compile_ast(code, compile_options)
+            compile_ast(code, compile_options, boltlib_path)
           end
 
           { catalog: catalog, logs: logs }
         else
-          catalog = compile_ast(code, compile_options)
+          catalog = compile_ast(code, compile_options, boltlib_path)
           { catalog: catalog }
         end
       end
 
-      def self.compile_ast(code, compile_options)
-        Puppet[:node_name_value] = compile_options['certname']
+      def self.compile_ast(code, compile_options, boltlib_path)
+        # If the request requires that bolt be loaded we assume we are in a properly 
+        # configured PE environment
+        if compile_options.dig('options', 'compile_for_plan')
+          unless boltlib_path
+            msg = "When compile_for_plan is set, the path to boltlib modules " \
+            "must be provided by setting boltlib-path as a jruby-puppet setting in pe-puppet-server.conf"
+            raise(Puppet::Error, msg)
+          end
 
-        # Use the existing environment with the requested name
-        Puppet::Pal.in_environment(compile_options['environment'],
-                                   envpath: Puppet[:environmentpath],
-                                   facts: compile_options['facts']['values'],
-                                   variables: compile_options['variables']['values']) do |pal|
-          Puppet.lookup(:pal_current_node).trusted_data = compile_options['trusted_facts']['values']
+          # TODO: PE-28677 Develop entrypoint for loading bits of bolt we need here.
+          require 'bolt/apply_inventory'
+          require 'bolt/apply_target'
+          require 'bolt/pal/issues'
 
-          # This compiler has been configured with a node containing
-          # the requested environment, facts, and variables, and is used
-          # to compile a catalog in that context from the supplied AST.
-          pal.with_catalog_compiler do |compiler|
-            # We have to parse the AST inside the compiler block, because it
-            # initializes the necessary type loaders for us.
-            ast = Puppet::Pops::Serialization::FromDataConverter.convert(code)
+          Puppet[:node_name_value] = compile_options['certname']
 
-            compiler.evaluate(ast)
-            compiler.compile_additions
-            compiler.catalog_data_hash
+          env_conf = {
+            pre_modulepath: boltlib_path,
+            envpath: Puppet[:environmentpath],
+            facts: compile_options['facts']['values'],
+            variables: compile_options['variables']['values']
+          }
+
+          # Use the existing environment with the requested name
+          Puppet::Pal.in_environment(compile_options['environment'], env_conf) do |pal|
+            # TODO: Given we hide this from plan authors this current iteration has only
+            # the "required" data for now. Once we can get https://github.com/puppetlabs/bolt/pull/1770
+            # merged and promoted we can just use an empty hash.
+            fake_config = {
+              'transport' => 'redacted',
+              'transports' => {
+                'redacted' => 'redacted'
+              }
+            }
+            bolt_inv = Bolt::ApplyInventory.new(fake_config)
+            Puppet.override(bolt_inventory: bolt_inv) do
+              Puppet.lookup(:pal_current_node).trusted_data = compile_options['trusted_facts']['values']
+              # This compiler has been configured with a node containing
+              # the requested environment, facts, and variables, and is used
+              # to compile a catalog in that context from the supplied AST.
+              pal.with_catalog_compiler() do |compiler|
+                # TODO: PUP-10476 Explore setting these as default in PAL. They are the defaults in Puppet
+                Puppet[:strict] = :warning
+                Puppet[:strict_variables] = false
+
+                ast = build_program(code)
+                compiler.evaluate(ast)
+                compiler.evaluate_ast_node
+                compiler.compile_additions
+                compiler.catalog_data_hash
+              end
+            end
+          end
+        else
+          # When we do not need to load bolt we assume there are no Bolt types to resolve in
+          # AST compilation.
+          Puppet[:node_name_value] = compile_options['certname']
+
+          # Use the existing environment with the requested name
+          Puppet::Pal.in_environment(compile_options['environment'],
+                                     envpath: Puppet[:environmentpath],
+                                     facts: compile_options['facts']['values'],
+                                     variables: compile_options['variables']['values']) do |pal|
+            Puppet.lookup(:pal_current_node).trusted_data = compile_options['trusted_facts']['values']
+
+            # This compiler has been configured with a node containing
+            # the requested environment, facts, and variables, and is used
+            # to compile a catalog in that context from the supplied AST.
+            pal.with_catalog_compiler do |compiler|
+              # We have to parse the AST inside the compiler block, because it
+              # initializes the necessary type loaders for us.
+              ast = Puppet::Pops::Serialization::FromDataConverter.convert(code)
+
+              compiler.evaluate(ast)
+              compiler.compile_additions
+              compiler.catalog_data_hash
+            end
           end
         end
       end
       private_class_method :compile_ast
+
+      def self.build_program(code)
+        ast = Puppet::Pops::Serialization::FromDataConverter.convert(code)
+        # Node definitions must be at the top level of the apply block.
+        # That means the apply body either a) consists of just a
+        # NodeDefinition, b) consists of a BlockExpression which may
+        # contain NodeDefinitions, or c) doesn't contain NodeDefinitions.
+        # See https://github.com/puppetlabs/bolt/pull/1512 for more details
+        definitions = if ast.is_a?(Puppet::Pops::Model::BlockExpression)
+                        ast.statements.select { |st| st.is_a?(Puppet::Pops::Model::NodeDefinition) }
+                      elsif ast.is_a?(Puppet::Pops::Model::NodeDefinition)
+                        [ast]
+                      else
+                        []
+                      end
+        # During ordinary compilation, definitions are stored on the parser at
+        # parse time and then added to the Program node at the root of the AST
+        # before evaluation. Because the AST for an apply block has already been
+        # parsed and is not a complete tree with a Program at the root level, we
+        # need to rediscover the definitions and construct our own Program object.
+        # https://github.com/puppetlabs/bolt/commit/3a7597dda25cdb25854c7d08d37c5c58ab6a016b
+        Puppet::Pops::Model::Factory.PROGRAM(ast, definitions, ast.locator).model
+      end
+      private_class_method :build_program
     end
   end
 end
--- a/src/ruby/puppetserver-lib/puppet/server/master.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/master.rb
@@ -192,13 +192,30 @@ class Puppet::Server::Master
 
   private
 
+  # This helper is used to resolve all java objects in an array.
+  # Each array element is examined, if it is expected to be a map
+  # we call back to the convert_java_args_to_ruby method, if it
+  # is expected to be an array, we recurse otherwise we do not modify
+  # the value. 
+  def resolve_java_objects_from_list(list)
+    list.map do |value|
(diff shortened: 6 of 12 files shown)


donoghuc:
Orch changes: https://github.com/puppetlabs/orchestrator/pull/1340
Ticket @McdonaldSeanp is working on to ship the required gems and add new settings to enterprise modules https://tickets.puppetlabs.com/browse/PE-28881
Ticket i'm working on to add acceptance tests in orch. https://tickets.puppetlabs.com/browse/PE-28597

mwaggett:
Do we usually mention PE stuff in FOSS commit messages? I know the puppet repo doesn't like non-PUP tickets, but not sure if we have a similar policy here.

documentation/puppet-api/v3/compile.markdown
```diff
@@ -36,7 +36,8 @@ The request body must look like:
   "transaction_uuid": "<uuid string>",
   "job_id": "<id string>",
   "options": { "capture_logs": <true/false>,
-               "log_level": <one of debug/info/warning/err>}
+               "log_level": <one of debug/info/warning/err>
+               "bolt": <true/false> Whether or not to attempt to load bolt }
```
mwaggett, line 40: What if we called this `load_bolt` so that we can drop the explanation? Just noticing that the other options don't have that.

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -53,7 +53,11 @@
 
     * :use-legacy-auth-conf - Whether to use the legacy core Puppet auth.conf
         (true) or trapperkeeper-authorization (false) to authorize requests
-        being made to core Puppet endpoints."
+        being made to core Puppet endpoints.
+
+    * :botlib-path - Optional array containng path(s) to bolt modules. This path
```
mwaggett, line 58:
```suggestion
    * :botlib-path - Optional array containing path(s) to bolt modules. This path
```

documentation/puppet-api/v3/compile.markdown
McdonaldSeanp: I forgot to push my nitpick in the last PR: personally I think this should be something like `load_bolt_types` or similar to identify _why_ you would need to load bolt.

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -53,7 +53,11 @@
...
+         will be prepended to AST compilation modulepath. This is required for
+         compiling AST that contains bolt types"
```
mwaggett, line 60:
```suggestion
         compiling AST that contains bolt types."
```

documentation/puppet-api/v3/compile.markdown
donoghuc: Bolt https://github.com/puppetlabs/bolt/blob/master/lib/bolt/applicator.rb#L94 adds a fact for compilation indicating it is being done in a special "bolty" way. I ported that behavior to orch https://github.com/puppetlabs/orchestrator/pull/1340/files#diff-e3cb1255de6ee2d1bf0cf60785425dd0R305 . I named this accordingly. Perhaps it is a bolt centric way of thinking about it.

McdonaldSeanp: `compile_for_plan` maybe then?

donoghuc: Ok it will be trivial to update. I just has to happen in two projects. We can bike shed on it for a bit and i'll update it to whatever we come up with.

justinstoller:
fwiw, I had these two concerns about this approach that I outlined here: https://github.com/puppetlabs/puppetserver/pull/2329#discussion_r416210989 I was fine merging the current approach in since we were trying to make the 6.15.0 deadline. I'm curious, since we didn't make that deadline, can we attempt those approaches now?

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -5,45 +5,121 @@
...
-
-            compiler.evaluate(ast)
-            compiler.compile_additions
-            compiler.catalog_data_hash
+      def self.compile_ast(code, compile_options, boltlib_path)
+        # If the request requires that bolt be loaded we assume we are in a properly 
+        # configured PE environment
+        if compile_options.dig('options', 'bolt')
+          # CODEREVIEW: Where to we handle this error and what should it say?
+          raise "The path to boltib modules must be specified in puppetserver config" unless boltlib_path
```
mwaggett, line 31: I think this should specify the exact setting and file where it needs to be set.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -5,45 +5,121 @@
...
+              # This compiler has been configured with a node containing
+              # the requested environment, facts, and variables, and is used
+              # to compile a catalog in that context from the supplied AST.
+              pal.with_catalog_compiler do |compiler|
+                # TODO: PUP-10476 Explore setting these as default in PAL. They are the defaults in Puppet
+                Puppet[:strict] = :warning
+                Puppet[:strict_variables] = false
+                vars = Puppet::Pops::Serialization::FromDataConverter.convert(compile_options['variables']['values'])
+
+                compile_options['facts']['values'].keys.each {|fact_name| vars.delete(fact_name)}
```
mwaggett, line 69: Can you put back the comment that explains why you do this?

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
donoghuc: I agree the error message should be more detailed. I am wondering if there is a specific error type to throw.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -5,45 +5,121 @@
...
+                # TODO: PUP-10476 Refactor PAL api such that we do not need to use private methods
+                pal.send(:add_variables, compiler.send(:topscope), vars)
+                Puppet.warning("#{compile_options}")
```
mwaggett, line 72: Will this not get coerced to a string without the `#{}`?

donoghuc: OOps. that is debuggging logging. neeed to delete.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -5,45 +5,121 @@
...
+                # contain NodeDefinitions, or c) doesn't contain NodeDefinitions.
+                # See https://github.com/puppetlabs/bolt/pull/1512 for more details
+                definitions = if ast.is_a?(Puppet::Pops::Model::BlockExpression)
+                                ast.statements.select { |st| st.is_a?(Puppet::Pops::Model::NodeDefinition) }
+                              elsif ast.is_a?(Puppet::Pops::Model::NodeDefinition)
+                                [ast]
+                              else
+                                []
+                              end
+                ast = Puppet::Pops::Model::Factory.PROGRAM(ast, definitions, ast.locator).model
```
mwaggett, line 89: Would it be better to call this something else, instead of redefining `ast`?

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
donoghuc: It used to be `Puppet.warning("WHY IS THERE JAVA IN MY: #{compile_options}")`

mwaggett: lol

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
donoghuc: I dont think so. Its still AST. It just transformed.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
donoghuc: I now raise a `Puppet::Error` and wrote a test for it. Happy to iterate on the wording.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
donoghuc: deleted.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -5,45 +5,124 @@
...
-            compiler.compile_additions
-            compiler.catalog_data_hash
+      def self.compile_ast(code, compile_options, boltlib_path)
+        # If the request requires that bolt be loaded we assume we are in a properly 
+        # configured PE environment
+        if compile_options.dig('options', 'bolt')
+          # CODEREVIEW: Where to we handle this error and what should it say?
+          unless boltlib_path
+            msg = "The path to boltlib modules must be provided in jruby-puppet boltlib-path puppetserver setting"
+            raise Puppet::Error, _(msg)
```
mwaggett, line 33: What's the underscore for?

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
mwaggett: @donoghuc I found your original comment clearer:

> # If fact conflicts with plan var name, fact wins, so delete plan var

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
mwaggett: you could call it `transformed_ast`?

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
donoghuc: I think it has something to do with translation. I looked at https://github.com/puppetlabs/puppetserver/blob/master/src/ruby/puppetserver-lib/puppet/server/compiler.rb#L56 tho it seems in the same file we do https://github.com/puppetlabs/puppetserver/blob/59d3ff7d075b2fbb5507ec143979d5d21eb3f9ff/src/ruby/puppetserver-lib/puppet/server/compiler.rb#L166 It probably makes sense to not mark it if its not going to be translated.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
donoghuc: Actually I do think this is an improvement. (Though I don't think variable reassignment/mutations in inherently bad). In bolt we need to conditionally do the transformation because AST snippets from apply blocks do not have a full tree with Program at the root. Where manifests applied with `bolt apply` _will_ have the full tree. (https://github.com/puppetlabs/bolt/commit/3a7597dda25cdb25854c7d08d37c5c58ab6a016b) Here we know we will always need to do it. I will update this and perhaps add a comment as to why.

adreyer:
Regarding the two questions raised in the old PR

> 1. AFAICT, Bolt resource compilation is the primary, quite probably only, consumer of PAL. It looks like currently Bolt has a few hacks around the PAL API to enable what it needs. Since those changes aren't upstreamed into PAL proper those hacks need to be copy-pasted here. I would love to see Puppet's version of PAL allow setting variables like you need to, as well as compiling ast code as you need it, and having setting defaults that you can use by default.

I agree we should clean this up. I'm not sure whether that should happen with immediately on as part of Puppet 7. I'm still concerned about the time crunch with the LTS approaching but doing this immediately post-LTS will definitely increase the maintenance burden of the LTS. I also don't know how stable PAL needs to be. Should we set up a meeting to figure this out?

> 2. We are now loading all of Bolt to, afaict, have the apply_target and apply_result PCore types available. Does this mean that functions like file::readable are now available? I have a lot of reservations about loading an entire other product within Puppet Server (where will its logging go? etc). Puppet Server's already kind of a Turducken, and now we've added the Christmas Ham... I would much prefer we separate out the PCore types into a Puppet module that server can load, though I understand that disentangling the PCore types from Bolt may be involved.

We could extract the boltlib module with it's PCore types pretty easily but some of the ruby implementations are used without PAL so that would be messier.  I think the question of whether to load the bolt gem into puppet-server is a question of how we want to scale plans in general. Long term we need to horizontally scale *all* of bolt in pe not just catalog compilation. Do we want to set up a new service for scaling plan execution and ssh/winrm or put it in puppet-server to keep our footprint simple? It seems like ssh and winrm probably do belong in puppet-server so we have a single "edge server" for users to install in network partitions which will mean pulling bolt in anyway.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -5,45 +5,124 @@
...
+          end
+          require 'bolt/target'
```
donoghuc, line 35: I'm not convinced I actually need to require `bolt/target`. I think I can drop that one. I'm not exactly sure why it does not appear that I need result_set. Once I experiment with running a few more plans I should have a better idea.

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -67,5 +71,6 @@
    :http-client-idle-timeout-milliseconds schema/Int
    :http-client-metrics-enabled schema/Bool
    :max-requests-per-instance schema/Int
-   :use-legacy-auth-conf schema/Bool})
+   :use-legacy-auth-conf schema/Bool
+   (schema/optional-key :botlib-path) [schema/Str]})
```
donoghuc, line 76: Well it appears I have a pretty bad typo here 😬

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -5,45 +5,126 @@
...
+                # TODO: PUP-10476 Explore setting these as default in PAL. They are the defaults in Puppet
+                Puppet[:strict] = :warning
+                Puppet[:strict_variables] = false
+                vars = Puppet::Pops::Serialization::FromDataConverter.convert(compile_options['variables']['values'])
+
+                # Facts trump plan variables and target vars.
+                # See https://github.com/puppetlabs/bolt/issues/1725 for details
+                compile_options['facts']['values'].keys.each {|fact_name| vars.delete(fact_name)}
+                # TODO: PUP-10476 Refactor PAL api such that we do not need to use private methods
+                pal.send(:add_variables, compiler.send(:topscope), vars)
```
donoghuc, line 71: Promising potential solution for this https://github.com/puppetlabs/bolt/pull/1784

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
Magisus: I'm not honestly sure where we stand on marking strings these days. I _think_ the i18n effort is pretty much dead for now, but last I heard we were still being asked to keep infrastructure like string marking and POT file generation in place. Though that was over a year ago, so things may have changed.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -5,49 +5,132 @@
...
-            compiler.evaluate(ast)
-            compiler.compile_additions
-            compiler.catalog_data_hash
+      def self.compile_ast(code, compile_options, boltlib_path)
+        # If the request requires that bolt be loaded we assume we are in a properly 
+        # configured PE environment
+        if compile_options.dig('options', 'compile_for_plan')
+          # CODEREVIEW: Where to we handle this error and what should it say?
+          unless boltlib_path
+            msg = "The path to boltlib modules must be provided in jruby-puppet boltlib-path puppetserver setting"
```
mwaggett, line 32: I think this could be improved, but that can be a follow-up. I would want to know the exact file and setting to set and under what use case it's necessary.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -5,49 +5,132 @@
...
+            raise Puppet::Error, _(msg)
+          end
+
+          require 'bolt/apply_inventory'
+          require 'bolt/apply_target'
+          # TODO: Bolt's pal requires this so we dont have to do it in every boltib function that
+          # is explicitly forbidden in an apply block. Would it make sense to require in each function?
+          # or just do it here?
```
mwaggett, line 40: I don't understand this comment. So we don't have to do _what_ in every boltlib function? What is explicitly forbidden in an apply block?

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -5,49 +5,132 @@
...
+            end
           end
         end
       end
       private_class_method :compile_ast
+
+      def self.build_program(code)
+        ast = Puppet::Pops::Serialization::FromDataConverter.convert(code)
+        # Node definitions must be at the top level of the apply block.
+         # That means the apply body either a) consists of just a
```
mwaggett, line 114: (nit)
```suggestion
        # That means the apply body either a) consists of just a
```

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -5,49 +5,132 @@
...
+        # See https://github.com/puppetlabs/bolt/pull/1512 for more details
+        definitions = if ast.is_a?(Puppet::Pops::Model::BlockExpression)
+                        ast.statements.select { |st| st.is_a?(Puppet::Pops::Model::NodeDefinition) }
+                      elsif ast.is_a?(Puppet::Pops::Model::NodeDefinition)
+                        [ast]
+                      else
+                        []
+                      end
+        # During ordinary compilation, definitions are stored on the parser at
+         # parse time and then added to the Program node at the root of the AST
```
mwaggett, line 126: (nit)
```suggestion
        # parse time and then added to the Program node at the root of the AST
```

mwaggett:
@donoghuc If this is ready from your perspective, can you remove the DO NOT MERGE label?

This seems sane to me, but I'd like someone with more puppetserver experience to take a look.

donoghuc:
Is DO NOT MERGE lable discouraging review?  I added it because this is still blocked on getting the pse bits merged/promoted. It is also dependent on some orchestrator changes.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
donoghuc: Lets improve it. My thought process is that this code path will only be hit when requested from orchestrator, which probably means its from PE, and pe lays down the `jruby-puppet` `boltlib-path` setting in `/etc/puppetlabs/puppetserver/conf.d/pe-puppet-server.conf`.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
donoghuc: There are some (many) functions in boltlib that are explicitly forbidden in apply blocks. The implementation of forbidding them requires a file that is loaded by Bolt's PAL. There is not an easy way to get that change in to bolt without adding an equivalent `require` to every single function definition that is explicitly forbidden in boltlib.

mwaggett:
oh, I thought this was one of the ones you said was ready. if it's still blocked, you can leave the label. 😊

donoghuc:
To be clear, It is ready for review, we just need to be thoughtful on merge order.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
mwaggett: Is the `compile_for_plan` option publicly exposed? If so, how about something like:
"When `compile_for_plan` is set, the path to boltlib modules must be provided by setting `boltlib-path` as a `jruby-puppet` setting in `pe-puppet-server.conf`."

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
mwaggett: Can you update your comment to something of that^ effect? I want this to be clear for future code spelunkers.

donoghuc: I will update. It should be a link to an issue. i think this is actionable. Requiring it here is just a quick workaround.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_service.clj
```diff
@@ -45,7 +46,8 @@
                                                                    (throw (IllegalStateException.
                                                                             (i18n/trs "Versioned code not supported."))))
                                                                  (constantly nil)
-                                                                 false))
+                                                                 false
+                                                                 boltlib-path))
```
Magisus, line 50: Do we care about this config setting for the legacy routes? Can't we just pass nil instead of pulling it out of the config hash?

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -5,49 +5,132 @@
...
-            ast = Puppet::Pops::Serialization::FromDataConverter.convert(code)
-
-            compiler.evaluate(ast)
-            compiler.compile_additions
-            compiler.catalog_data_hash
+      def self.compile_ast(code, compile_options, boltlib_path)
+        # If the request requires that bolt be loaded we assume we are in a properly 
+        # configured PE environment
+        if compile_options.dig('options', 'compile_for_plan')
+          # CODEREVIEW: Where to we handle this error and what should it say?
```
Magisus, line 30: Does this bring down the server right now if this happens? We probably don't want that, we might just want to log an error and abort, rather than raising.

Magisus: I _think_ logged errors will get returned with a 500 with the request. I'll double check.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -5,49 +5,132 @@
...
+
+                ast = build_program(code)
+                compiler.evaluate(ast)
+                compiler.evaluate_ast_node
+                compiler.compile_additions
+                compiler.catalog_data_hash
+              end
+            end
+          end
+        else
```
Magisus, line 82: Do we have a use case right now for this code branch? If not (and we can't think of one) I'm inclined to just get rid of it (and also the boolean in the request). As we pointed out early you, this endpoint isn't even documented.

src/clj/puppetlabs/services/legacy_routes/legacy_routes_service.clj
donoghuc: I can update that. I didnt know if that would be needed there...

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
donoghuc: Not that I know of. I only added that code path to satisfy https://github.com/puppetlabs/puppetserver/pull/2329#discussion_r416055718

src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -192,13 +192,30 @@ def getPlanData(environment_name, module_name, plan_name)
 
   private
 
+  # This helper is used to resolve all java object in an array.
```
Magisus, line 195: "object" -> "objects"

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
Magisus: Okay let's leave it then.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
Magisus: We tested this out, and it logs an error (with a stack trace) and returns a 500 with the request, then carries on. That seems like just what we want.

spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
```diff
@@ -33,32 +33,42 @@ def find_notify(catalog)
...
       expect(notify).not_to be_nil
       expect(notify['title']).to eq("trusted_value")
     end
+
+    it 'fails gracefully when boltlib is not found when request requires bolt types' do
+      bolt_request = request('notify { "${trusted[\'my_trusted\']}": }')
+      bolt_request.merge!({'options' => { 'compile_for_plan' => true } })
+      expect {
+        Puppet::Server::ASTCompiler.compile(bolt_request, boltlib_path)
+      }.to raise_error(Puppet::Error, /The path to boltlib modules must be provided/)
```
mwaggett, line 71: this needs to be updated to match the new message

donoghuc: OMG! I thought i had it. the update still matches the words, but not the case. Will update with a lowercase `the` in my regex  :cry:

mwaggett:
@Magisus or @donoghuc do one of you want to make a ticket for making a decision about the non-Bolt code path?

otherwise, looks like this is ready to go?

Magisus:
No. If the code being there ever gets annoying we can revisit it.

PR #2334, opened 2020-04-29
(SERVER-2565) Write files atomically in CA

pcarlisle:
When writing files always use the atomic-write function from kitchensink which
ensures the file is either fully written or not at all by writing out a temp
file and moving it into place.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -4,7 +4,7 @@
            [java.io InputStream ByteArrayOutputStream ByteArrayInputStream File StringReader IOException]
            [java.nio.file Files Paths LinkOption]
            [java.nio.file.attribute FileAttribute PosixFilePermissions]
-           (java.security KeyPair)
+           (java.security PrivateKey PublicKey KeyPair)
            (org.joda.time DateTime)
            (java.security.cert CRLException)
            (sun.security.x509 X509CertImpl))
@@ -133,6 +133,9 @@
 
 (def CertificateRevocationList
   (schema/pred utils/certificate-revocation-list?))
+
+(def CertificateRequest
+  (schema/pred utils/certificate-request?))
 
 (def OutcomeInfo
   "Generic map of outcome & message for API consumers"
@@ -315,6 +318,7 @@
     {:not-before (.toDate not-before)
      :not-after  (.toDate not-after)}))
 
+
 (schema/defn settings->cadir-paths
   "Trim down the CA settings to include only paths to files and directories.
   These paths are necessary during CA initialization for determining what needs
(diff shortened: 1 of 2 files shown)


src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -423,6 +432,51 @@
...
+
+(schema/defn write-private-key
+  "Encode a key to PEM format and write it to a file atomically and with
+  appropriate permissions for a private key."
+  [key :- PrivateKey
+   path :- schema/Str]
+  (ks-file/atomic-write path (partial utils/key->pem! key) private-key-perms))
+
+(schema/defn write-cert
+  "Encode a certificate PEM format and write it to a file atomically and with
```
mwaggett, line 453:
```suggestion
  "Encode a certificate to PEM format and write it to a file atomically and with
```

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -570,29 +632,37 @@
...
+        (try
+          (let [infra-serial (-> (path-to-cert signeddir infra-node)
+                                 (utils/pem->cert)
+                                 (utils/get-serial))]
+            (.write writer (str infra-serial))
+            (.newLine writer))
+          (catch java.io.FileNotFoundException ex
+            (log/warn
+             (i18n/trs
+              (str
```
mwaggett, line 649: I don't think you need to wrap this in `str` - `trs` should let you interpolate with `{0}`.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -423,6 +432,51 @@
...
+   path :- schema/Str]
+  (ks-file/atomic-write path (partial utils/objs->pem! crls) public-key-perms))
+
+(schema/defn write-csr
+  "Encode a CSR to PEM format and write it to a file atomically and with
+  appropriate permissions."
+  [csr :- CertificateRequest
+   path :- schema/Str]
+  (ks-file/atomic-write path (partial utils/obj->pem! csr) public-key-perms))
+
```
Magisus, line 479: The `*->pem!` functions that these are replacing were all originally defined in the ssl-utils library. Do we think these should actually replace them there at some point? I don't have a strong opinion, but these seem like simply better versions of those.

Magisus:
Really like how neat this turned out!

src/clj/puppetlabs/puppetserver/certificate_authority.clj
pcarlisle: I wasn't sure about this. Maybe at some point?

Magisus: Yeah, I'm totally fine merging this as is though.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
pcarlisle: That's true, but it's fine either way and this isn't code that actually changed in this PR so I'm not going to touch it.

mwaggett: fair

PR #2337, opened 2020-05-07
(DOCS) Release notes for 6.11.1 release

Magisus:


Magisus:
Wasn't sure exactly how to describe this, we're not sure the issue is totally fixed and the ticket is not resolved, so I didn't want to say "Resolved Issue". But let me know if that (or something else) would be better.

documentation/release_notes.markdown
```diff
@@ -10,6 +10,14 @@ canonical: "/puppetserver/latest/release_notes.html"
 [puppetserver.conf]: ./config_file_puppetserver.markdown
 [product.conf]: ./config_file_product.markdown
 
+## Puppet Server 6.11.1
+
+Released May 7 2020
+
+### Improvement
+
+- JRuby has been rolled back to 9.2.8.0 to attempt to mitigate an intermittent problem where some requests that go through JRuby error repeatedly with StackOverflow exceptions. [SERVER-2793](https://tickets.puppetlabs.com/browse/SERVER-2793)
```
clairecadman, line 19: I'd either have this as a 'New feature', but remove "attempts to" —  "JRuby has been rolled back to 9.2.8.0 which mitigates an intermittent problem where some requests that go through JRuby error repeatedly with StackOverflow exceptions" (if this is confirmed). Or if not, maybe it should be a 'known issue'? And rephrased to "When doing XXX, an intermittent problem where some requests that go through JRuby error repeatedly with StackOverflow exceptions. JRuby has been rolled back to 9.2.8.0 in an attempt to mitigate this issue".

clairecadman:
@Magisus Sorry I was off on Friday. I've added a comment to the note. Did this release already? If not the date needs to be updated too. Thanks!

mwaggett:
just realized we might want to remove the `sprintf` thing from the release notes, since we've rolled back

Magisus:
Well if you're on that version, it's still fixed. I think it's pretty clear that the downgrade to JRuby would remove that fix if you upgrade to latest, but I can add a note that this also undoes that fix. I'd rather not remove the original note since it's still accurate.

mwaggett:
That's fair. but yeah, I think we should add a note then, because the original release note doesn't associate the `sprintf` fix with jruby.

Magisus:
@puppetlabs/puppetserver-maintainers thoughts on whether this should a "known issue" or a "bug fix"? How confident are we that this fixed the problem? I don't want to do new feature, because the downgrade is actually a regression from most viewpoints.

Magisus:
And yes @clairecadman this released already, as an out-of-band release.

Magisus:
Ugh, annoying. We should probably have made that connection (it's in the ticket). But yeah I'll add a note.

documentation/release_notes.markdown
```diff
@@ -10,6 +10,18 @@ canonical: "/puppetserver/latest/release_notes.html"
 [puppetserver.conf]: ./config_file_puppetserver.markdown
 [product.conf]: ./config_file_product.markdown
 
+## Puppet Server 6.11.1
+
+Released May 7 2020
```
mwaggett, line 15: other dates have the day first

documentation/release_notes.markdown
```diff
@@ -10,6 +10,18 @@ canonical: "/puppetserver/latest/release_notes.html"
...
+
+### Bug Fix
```
mwaggett, line 17: seems like we use 'Resolved Issues' rather than 'Bug Fix' elsewhere in here.. Though I realize we're not positive that it's been resolved.. ¯\_(ツ)_/¯

Magisus: I'll update.

documentation/release_notes.markdown
```diff
@@ -10,6 +10,18 @@ canonical: "/puppetserver/latest/release_notes.html"
 [puppetserver.conf]: ./config_file_puppetserver.markdown
 [product.conf]: ./config_file_product.markdown
 
+## Puppet Server 6.11.1
+
+Released 7 May 2020
+
+### Resolved Issue
```
Magisus, line 17: Think we should move both notes under "known issue"? A rollback doesn't feel like a resolution, and the ticket isn't closed.

mwaggett: works for me

PR #2339, opened 2020-05-11
(maint) Switch Travis to Ubuntu 20.04 builder

Iristyle:
- Focal fossa time!
 - Only upgrade the Docker cell as the other cells can't handle focal
   yet
 - No need to upgrade Docker for buildkit as image has 19.03 already,
   so this should remove ~30s of setup time
 - Switch to Ruby 2.6.6 for specs, which is known good and installed
 - docker-compose in the 19.03 release included here is 1.23.1, so
   continue to upgrade it on Travis to 1.25.5
 - Separate out Travis operations so that logs have timings for lint, build and test


Iristyle:
make lint: 2.76s
make build: 317.43s
make test: 175.17s

PR #2346, opened 2020-05-26
(DOCS) Add SERVER-2535 limitation

clairecadman:
Requested in DOC-4521 (https://tickets.puppetlabs.com/browse/DOC-4521).

### DIFF
--- a/documentation/intermediate_ca.md
+++ b/documentation/intermediate_ca.md
@@ -34,12 +34,11 @@ To set up Puppet as an intermediate CA with an external root:
 * The root cert
 * Any intermediate CA certs
 * The CA cert that you will use to issue a CA signing cert for your Puppet infrastructure
-2. Collect the corresponding CRLs for each of these certificates. Take note of the expiration dates of each CRL. When one expires, you need to refresh it for Puppet to continue working. For more information, see [SERVER-2535](https://tickets.puppetlabs.com/browse/SERVER-2535). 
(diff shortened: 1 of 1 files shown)


clairecadman:
Also, the last two steps are labelled as 1 in the docs: https://puppet.com/docs/puppetserver/latest/intermediate_ca.html#set-up-puppet-as-an-intermediate-ca-with-an-external-root. But they are correct in the markdown, so I'm not sure how to fix this. Let me know if you have any ideas, thanks!

Magisus:
Maybe the blank lines around the code blocks later in the list are messing with the numbered list? Could try removing them.

documentation/intermediate_ca.md
```diff
@@ -34,7 +34,7 @@ To set up Puppet as an intermediate CA with an external root:
 * The root cert
 * Any intermediate CA certs
 * The CA cert that you will use to issue a CA signing cert for your Puppet infrastructure
-2. Collect the corresponding CRLs for each of these certificates.
+2. Collect the corresponding CRLs for each of these certificates. Take note of the expiration dates of each CRL. When one expires, you need to refresh it for Puppet to continue working. For more information, see [SERVER-2535](https://tickets.puppetlabs.com/browse/SERVER-2535).
```
mwaggett, line 37: Do we often point to tickets in docs? It's not clear to me what "more information" someone is supposed to get from that ticket.

clairecadman: Sometimes when it's a limitation. But you're right if there is not much more to learn from the ticket we shouldn't have it there. Thanks!

PR #2353, opened 2020-05-28
(SERVER-2793) Update clj-parent to 4.5.2

Magisus:
This commit adds a flag to not use invoke-dynamic for yields in JRuby,
since that mode may be causing StackOverflows.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -27,7 +27,7 @@
 
   :min-lein-version "2.9.1"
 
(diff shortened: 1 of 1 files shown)


mwaggett:
Worth noting in the commit message that this also reverts the i18n bump?

Magisus:
Oh, does this need to revert the plugin version?

mwaggett:
oh yeah, good call, I would say so.

PR #2356, opened 2020-06-01
(SERVER-2804) Add release notes for 6.12.0

mwaggett:


documentation/release_notes.markdown
```diff
@@ -10,6 +10,18 @@ canonical: "/puppetserver/latest/release_notes.html"
...
 
+## Puppet Server 6.12.0
+
+Released 3 June 2020
+
+### Resolved Issue
+
+- JRuby has been bumped to 9.2.11.1 again, with `invokedynamic.yield` set to false to resolve a stackoverflow error. [SERVER-2793](https://tickets.puppetlabs.com/browse/SERVER-2793)
+
+### Deprecation
```
mwaggett, line 21: Wasn't sure which heading to use here, lmk if there's something else we've used in the past.

documentation/release_notes.markdown
```diff
@@ -10,6 +10,18 @@ canonical: "/puppetserver/latest/release_notes.html"
 [puppetserver.conf]: ./config_file_puppetserver.markdown
 [product.conf]: ./config_file_product.markdown
 
+## Puppet Server 6.12.0
+
+Released 3 June 2020
+
+### Resolved Issue
```
mwaggett, line 17: lmk if we're comfortable calling the jruby issue "resolved"..

documentation/release_notes.markdown
Magisus: "Deprecation" seems fine to me.

mwaggett:
^ the test failure is the expected docker failure that occurs between tagging and releasing.

documentation/release_notes.markdown
```diff
@@ -10,6 +10,18 @@ canonical: "/puppetserver/latest/release_notes.html"
...
+
+- JRuby has been bumped to 9.2.11.1 again, with `invokedynamic.yield` set to false to resolve a stackoverflow error. [SERVER-2793](https://tickets.puppetlabs.com/browse/SERVER-2793)
+
+### Deprecation
+
+- We have deprecated the v1 metrics endpoint, which was recently disabled by default. It will be removed entirely in a future release. Please use the v2 endpoint instead. [TK-486](https://tickets.puppetlabs.com/browse/TK-486)
```
clairecadman, line 23:
```suggestion
- The v1 metrics endpoint, which was recently disabled by default, is now deprecated. Instead, use the v2 endpoint. [TK-486](https://tickets.puppetlabs.com/browse/TK-486)
```

clairecadman: Usually we avoid saying things about future releases in the docs, and being deprecated implies that it will be removed later anyway I think.

documentation/release_notes.markdown
clairecadman, line 17: Minor, but lower case for issue, so it's consistent.

documentation/release_notes.markdown
mwaggett: Good point, thanks! updated 😊

PR #2357, opened 2020-06-02
(Docs) Incorrect webserver.conf client-auth example

clairecadman:
Based on the following customer ticket: https://tickets.puppetlabs.com/browse/DOCUMENT-1114.

I wasn't sure whether the equals sign in `access-log-config = /etc/puppetlabs/puppetserver/request-logging.xml` should also be replaced with a colon?


Magisus:
Technically both `=` and `:` are valid HOCON, but it sounds like at some point we decided to use `:`. So this file should be consistent.

PR #2360, opened 2020-06-16
(SERVER-2822) Doc updates for multithreaded mode

Magisus:


Magisus:
This is a first pass, there might be other spots that could use updating or an added note.

documentation/config_file_puppetserver.markdown
```diff
@@ -54,6 +54,8 @@ The `puppetserver.conf` file contains settings for Puppet Server software. For a
 
         JRuby flushing can be useful for working around buggy module code that would otherwise cause memory leaks, but it slightly reduces performance whenever a new JRuby instance reloads all of the Puppet Ruby code. If memory leaks from module code are not an issue in your deployment, the default value of 0 performs best.
 
+    * `multithreaded`: Optional, false by default. Configures Puppet Server to use a single JRuby instance to process requests that require a JRuby, processing a number of threads up to `max-active-instances` at a time. Reduces the memory footprint of the server by only requiring a single JRuby. When this setting is configured, Puppet's `multithreaded` setting must also be set to `true`.
```
mwaggett, line 57: Should we add a note to `max-active-instances` about what it means when `multithreaded` is on?

documentation/puppet_server_metrics.markdown
```diff
@@ -284,7 +284,7 @@ Puppet Server exports each metric in the lists below by default.
 
 #### JRuby metrics
 
-Puppet Server uses an embedded JRuby interpreter to execute Ruby code. JRuby spawns parallel instances known as JRubies to execute Ruby code, which occurs during most Puppet Server activities.
+Puppet Server uses an embedded JRuby interpreter to execute Ruby code. In normal mode, JRuby spawns parallel instances known as JRubies to execute Ruby code, which occurs during most Puppet Server activities. In multithreaded mode, a single JRuby is used instead to process a limited number of threads in parallel. For each of these metrics, they refer to JRuby instances in normal mode and JRuby threads in multithreaded mode.
```
mwaggett, line 287:
```suggestion
Puppet Server uses an embedded JRuby interpreter to execute Ruby code. By default, JRuby spawns parallel instances known as JRubies to execute Ruby code, which occurs during most Puppet Server activities. When `multithreaded` is set to `true`, a single JRuby is used instead to process a limited number of threads in parallel. For each of these metrics, they refer to JRuby instances by default and JRuby threads in multithreaded mode.
```

documentation/puppet_server_metrics_performance.markdown
```diff
@@ -50,9 +52,11 @@ If you are setting up Puppet Server for the first time, start by increasing your
 
 If you must add JRubies, remember that Puppet Server is tuned by default to use one fewer than your total number of CPUs, with a maximum of 4 CPUs, for the number of available JRubies. You can change this by setting `max-active-instances` in [`puppetserver.conf`][puppetserver.conf], under the `jruby-puppet` section.
 
+> **Note:** The following applies only to the normal mode where increasing `max-active-instances` creates whole independent JRuby instances. In multithreaded mode, this setting instead controls the number of threads that the single JRuby instance will process concurrently, and therefore has different scaling characteristics. Tuning recommendations for this mode are under development, see [SERVER-2823](https://tickets.puppetlabs.com/browse/SERVER-2823).
```
mwaggett, line 55: Why not put this note above the first paragraph of this section?

documentation/puppet_server_metrics_performance.markdown
```diff
@@ -31,6 +31,8 @@ Puppet Server uses JRuby, which rations server resources in the form of JRubies
 
 Several factors can limit your Server infrastructure's ability to produce JRubies.
 
+> **Note:** In multithreaded mode, instead of producing multiple independent JRuby instances, the server runs a single JRuby instance that can process a limited number of threads. These two concepts are analogous when considering request-handling capacity, though threads are much cheaper in terms of system resources, so you may have more vertical scalability in multithreaded mode.
```
mwaggett, line 34: This note seems a little out of place here? The first sentence seems relevant to mention here, but the second should maybe go in the `Request-handling capacity` section?

documentation/puppet_server_metrics_performance.markdown
```diff
@@ -50,9 +52,11 @@ If you are setting up Puppet Server for the first time, start by increasing your
...
 
-As a general rule, adding a JRuby requires a bare minimum of 40MB of memory under JRuby 1.7, and at least 60MB under JRuby9k if the `jruby-puppet.compile-mode` setting in [`puppetserver.conf`][puppetserver.conf] is set to `off` --- the amount of memory for the scripting container, plus Puppet's Ruby code, plus additional memory overhead --- just to compile an nearly empty catalog.
+As a general rule, adding a JRuby requires a bare minimum of 40MB of memory under JRuby 1.7, and at least 60MB under JRuby9k if the `jruby-puppet.compile-mode` setting in [`puppetserver.conf`][puppetserver.conf] is set to `off` --- the amount of memory for the scripting container, plus Puppet's Ruby code, plus additional memory overhead --- just to compile a nearly empty catalog.
```
clairecadman, line 59:
```suggestion
As a general rule, adding a JRuby requires a bare minimum of 40MB of memory under JRuby 1.7, and at least 60MB under JRuby9k if the `jruby-puppet.compile-mode` setting in [`puppetserver.conf`][puppetserver.conf] is set to `off` - the amount of memory for the scripting container, plus Puppet's Ruby code and additional memory overhead - just to compile a nearly empty catalog.
```

documentation/puppet_server_metrics_performance.markdown
clairecadman: Yes I agree. And also starting with "Adding more JRubies applies only to...etc". Instead of "The following...".

documentation/config_file_puppetserver.markdown
clairecadman, line 57:
```suggestion
    * `multithreaded`: Optional, false by default. Configures Puppet Server to use a single JRuby instance to process requests that require a JRuby, processing a number of threads up to `max-active-instances` at a time. Reduces the memory footprint of the server by only requiring a single JRuby. When you configure this setting, you must also set Puppet's `multithreaded` setting to `true`.
```

documentation/puppet_server_metrics_performance.markdown
Magisus: I can split it up.

documentation/puppet_server_metrics_performance.markdown
Magisus: If you're considering "a JRuby" to just as easily mean "a JRuby thread", the first paragraph applies to both modes equally. We use the same defaults and control the number with the same setting. But I'll rework this a bit, into that first paragraph. I still want to indicate that all the stuff in the paragraphs below about resourcing doesn't apply to threaded mode, but we don't know what to recommend instead yet.

Magisus:
Updated with suggestions, and worked some of the notes more into the main paragraphs. Let me know what you think.

documentation/puppet_server_metrics_performance.markdown
```diff
@@ -88,3 +92,5 @@ Puppet Server also requests facts as HTTP requests while handling a node request
...
 
 If you can't identify the source of a memory leak, setting the `max-requests-per-instance` setting in [`puppetserver.conf`][puppetserver.conf] to something other than the default of 0 limits the number of requests a JRuby handles during its lifetime and enables automatic JRuby flushing. Enabling this setting reduces overall performance, but if you enable it and no longer see signs of persistent memory leaks, check your module code for inefficiencies or memory-consuming bugs.
+
+> **Note:** In multithreaded mode, the `max-requests-per-instance` setting refers to the sum total number of requests processed by the single JRuby instance, across all of its threads. While that single JRuby is being flushed, all requests will suspend until the instance becomes available again.
```
mwaggett, line 96:
```suggestion
> **Note:** In `multithreaded` mode, the `max-requests-per-instance` setting refers to the sum total number of requests processed by the single JRuby instance, across all of its threads. While that single JRuby is being flushed, all requests will suspend until the instance becomes available again.
```
(nit)

documentation/config_file_puppetserver.markdown
mwaggett: I still think we should do this.

documentation/puppet_server_metrics_performance.markdown
Magisus: This is inconsistent right now, but I think we should only use the backticks when referring to the setting. When just describing the mode, I think they should be left off. I can fix the other places if that makes sense to folks?

documentation/puppet_server_metrics_performance.markdown
mwaggett: I don't love considering "a JRuby" to be an instance or a thread. I think, since we've already been referring to instances as "JRubies", we should be explicit when referring to threads.

I think it's much clearer now. 👍

documentation/config_file_puppetserver.markdown
```diff
@@ -54,6 +54,8 @@ The `puppetserver.conf` file contains settings for Puppet Server software. For a
 
         JRuby flushing can be useful for working around buggy module code that would otherwise cause memory leaks, but it slightly reduces performance whenever a new JRuby instance reloads all of the Puppet Ruby code. If memory leaks from module code are not an issue in your deployment, the default value of 0 performs best.
 
+    * `multithreaded`: Optional, false by default. Configures Puppet Server to use a single JRuby instance to process requests that require a JRuby, processing a number of threads up to `max-active-instances` at a time. Reduces the memory footprint of the server by only requiring a single JRuby. When you configure this setting, you must also set Puppet's `multithreaded` setting to `true`.
```
pcarlisle, line 57: I don't remember putting a `multithreaded` setting into Puppet. Where is that?

Magisus: Oh, maybe we didn't? Oops. I guess we're mostly just checking if we run in JRuby?

documentation/puppet_server_metrics_performance.markdown
mwaggett: That works for me, as long as we're consistent. 😊

documentation/config_file_puppetserver.markdown
pcarlisle: Yeah, everything should just use the threadsafe implementation if it's running on java

documentation/config_file_puppetserver.markdown
Magisus: Oh yeah, I think this comment got hidden, sorry.

documentation/config_file_puppetserver.markdown
```diff
@@ -48,12 +48,14 @@ The `puppetserver.conf` file contains settings for Puppet Server software. For a
 
     * `master-log-dir`: Optional. The path to the log directory. If nothing is specified, it uses the Puppet default `/var/log/puppetlabs/puppetserver`.
 
-    * `max-active-instances`: Optional. The maximum number of JRuby instances allowed. The default is 'num-cpus - 1', with a minimum value of 1 and a maximum value of 4.
+    * `max-active-instances`: Optional. The maximum number of JRuby instances allowed. The default is 'num-cpus - 1', with a minimum value of 1 and a maximum value of 4. In multithreaded mode, this control the number of threads allowed to run concurrently through the single JRuby instance.
```
mwaggett, line 51:
```suggestion
    * `max-active-instances`: Optional. The maximum number of JRuby instances allowed. The default is 'num-cpus - 1', with a minimum value of 1 and a maximum value of 4. In multithreaded mode, this controls the number of threads allowed to run concurrently through the single JRuby instance.
```

documentation/config_file_puppetserver.markdown
Magisus: Fixed. Thanks for catching that.

Magisus:
Did you want all the commits squashed down to one? No opinion.

mwaggett:
Doesn't need to be a single one, just wanted the edits squashed into the logical commits.

Magisus:
Okay fixed.

Magisus:
@clairecadman other thoughts on this? If not, feel free to merge it.

documentation/puppet_server_metrics_performance.markdown
```diff
@@ -48,11 +50,13 @@ If you are setting up Puppet Server for the first time, start by increasing your
 
 #### Adding more JRubies
 
-If you must add JRubies, remember that Puppet Server is tuned by default to use one fewer than your total number of CPUs, with a maximum of 4 CPUs, for the number of available JRubies. You can change this by setting `max-active-instances` in [`puppetserver.conf`][puppetserver.conf], under the `jruby-puppet` section.
+If you must add JRubies, remember that Puppet Server is tuned by default to use one fewer than your total number of CPUs, with a maximum of 4 CPUs, for the number of available JRubies. You can change this by setting `max-active-instances` in [`puppetserver.conf`][puppetserver.conf], under the `jruby-puppet` section. In the default mode, increasing `max-active-instances` creates whole independent JRuby instances. In multithreaded mode, this setting instead controls the number of threads that the single JRuby instance will process concurrently, and therefore has different scaling characteristics. Tuning recommendations for this mode are under development, see [SERVER-2823](https://tickets.puppetlabs.com/browse/SERVER-2823).
```
clairecadman, line 53: I think a typo? [`puppetserver.conf`][puppetserver.conf]

documentation/puppet_server_metrics_performance.markdown
```diff
@@ -48,11 +50,13 @@ If you are setting up Puppet Server for the first time, start by increasing your
...
 
-As a general rule, adding a JRuby requires a bare minimum of 40MB of memory under JRuby 1.7, and at least 60MB under JRuby9k if the `jruby-puppet.compile-mode` setting in [`puppetserver.conf`][puppetserver.conf] is set to `off` --- the amount of memory for the scripting container, plus Puppet's Ruby code, plus additional memory overhead --- just to compile an nearly empty catalog.
+As a general rule, adding a JRuby requires a bare minimum of 40MB of memory under JRuby 1.7, and at least 60MB under JRuby9k if the `jruby-puppet.compile-mode` setting in [`puppetserver.conf`][puppetserver.conf] is set to `off` - the amount of memory for the scripting container, plus Puppet's Ruby code and additional memory overhead - just to compile a nearly empty catalog.
```
clairecadman, line 59:
```suggestion
A JRuby requires a minimum of 40MB of memory under JRuby 1.7, and at least 60MB under JRuby9k if the `jruby-puppet.compile-mode` setting in [`puppetserver.conf`][puppetserver.conf] is set to `off` to compile a nearly empty catalog. This includes memory for the scripting container, Puppet's Ruby code and additional memory overhead.
```

clairecadman:
@Magisus I've made a suggestion to line 55, but otherwise looks good!

Magisus:
Updated.

Magisus:
@mwaggett would you mind glancing over this one more time and merging if you think it looks good?

documentation/puppet_server_metrics_performance.markdown
```diff
@@ -48,11 +50,13 @@ If you are setting up Puppet Server for the first time, start by increasing your
...
 
-As a general rule, adding a JRuby requires a bare minimum of 40MB of memory under JRuby 1.7, and at least 60MB under JRuby9k if the `jruby-puppet.compile-mode` setting in [`puppetserver.conf`][puppetserver.conf] is set to `off` --- the amount of memory for the scripting container, plus Puppet's Ruby code, plus additional memory overhead --- just to compile an nearly empty catalog.
+A JRuby requires a minimum of 40MB of memory under JRuby 1.7, and at least 60MB under JRuby9k if the `jruby-puppet.compile-mode` setting in [`puppetserver.conf`][puppetserver.conf] is set to `off` to compile a nearly empty catalog. This includes memory for the scripting container, Puppet's Ruby code and additional memory overhead.
```
mwaggett, line 59:
```suggestion
To compile a nearly-empty catalog, a JRuby requires a minimum of 40MB of memory under JRuby 1.7 and at least 60MB under JRuby9k if the `jruby-puppet.compile-mode` setting in [`puppetserver.conf`][puppetserver.conf] is set to `off`. This includes memory for the scripting container, Puppet's Ruby code, and additional memory overhead.
```

mwaggett: also, does the `compile-mode` bit only refer to JRuby9k? that part is a bit confusing to me.

Magisus: Looks like the default is "off" for 1.7 and "jit" for 9k, so this may indicate that the comparison was done with the setting "off" for both versions.

mwaggett: hmm okay then how about..
```suggestion
With the `jruby-puppet.compile-mode` setting in [`puppetserver.conf`][puppetserver.conf] set to `off`, a JRuby requires at least 40MB of memory under JRuby 1.7 and at least 60MB under JRuby9k in order to compile a nearly empty catalog. This includes memory for the scripting container, Puppet's Ruby code and additional memory overhead.
```

Magisus: I think that makes sense.

PR #2361, opened 2020-06-29
(Docs) Reorganise server installing instructions

clairecadman:
Do not merge yet (first iteration of changes).

I've made a few updates to start addressing the following feedback:

- The instructions jump around a lot - linking elsewhere on the same page - it really just needs a good prereq section
- The doc links to Enabling package repositories - that page should just list the available packages for each OS and the correct `wget` instead of making the user search for them (a nice touch would be adding the command for each OS to view their current OS build)
- There is an `apt-get` command, so perhaps the issue is that the commands are not clearly separated by platform. Separating the docs in this way would make them more user-friendly, especially for newer users who may not already be familiar with the command-line.

If anyone has other thoughts on how we could address this, please let me know or edit this PR. Thank you!

### DIFF
--- a/documentation/install_from_packages.markdown
+++ b/documentation/install_from_packages.markdown
@@ -12,10 +12,8 @@ Puppet provides official packages that install Puppet Server 6 and all of its pr
 
 * Red Hat Enterprise Linux 6, 7
 * Debian 8 (Jessie), 9 (Stretch), 10 (Buster)
-* Ubuntu 18.04 (Bionic), 16.04 (Xenial) — Enable the [universe repository](https://help.ubuntu.com/community/Repositories/Ubuntu), which contains packages necessary for Puppet Server.
+* Ubuntu 18.04 (Bionic), 16.04 (Xenial)
 * SLES 12 SP1
-
-> Note: Java 8 runtime packages do not exist in the standard repositories for Debian 8 (Jessie).  To install Puppet Server on Jessie, [configure the `jessie-backports` repository](https://backports.debian.org/Instructions/). 
 
 #### Java support
 
@@ -30,6 +28,8 @@ Puppet Server versions are tested against the following versions of Java:
 
 
 Some Java versions may work with other Puppet Server versions, but we do not test or support those cases. Community submitted patches for support greater than Java 11 are welcome. Both Java 8 and 11 are considered long-term support versions and are planned to be supported by upstream maintainers until 2022 or later.
+
+> Note: Java 8 runtime packages do not exist in the standard repositories for Debian 8 (Jessie) or Ubuntu 18.04 (Bionic).  To install Puppet Server on Jessie, [configure the `jessie-backports` repository](https://backports.debian.org/Instructions/). To install Puppet Sevrer on Bionic, enable the [universe repository](https://help.ubuntu.com/community/Repositories/Ubuntu).
 
 ## Installing Puppet Server
 
@@ -63,13 +63,7 @@ There is no `-` in the package name.
 systemctl start puppetserver
 ``` 
 
-or 
-
-```
-service puppetserver start
-```
-
-### Running Puppet Server on a VM
+#### Running Puppet Server on a VM
 
 By default, Puppet Server is configured to use 2GB of RAM. However, if you want to experiment with Puppet Server on a VM, you can safely allocate as little as 512MB of memory. To change the Puppet Server memory allocation, you can edit the init config file.
 
--- a/documentation/install_from_packages.markdown
+++ b/documentation/install_from_packages.markdown
@@ -1,16 +1,25 @@
 ---
 layout: default
-title: "Puppet Server: Installing From Packages"
+title: "Install Puppet Server"
 canonical: "/puppetserver/latest/install_from_packages.html"
 ---
 
 [repodocs]: https://puppet.com/docs/puppet/latest/puppet_platform.html
 
+Puppet provides official packages that install Puppet Server 6.0 and all of its prerequisites on x86_64 architectures for the following platforms, as part of [Puppet Platform][repodocs].
+
+* Red Hat Enterprise Linux 6, 7
+* Debian 8 (Jessie), 9 (Stretch), 10 (Buster)
+* Ubuntu 18.04 (Bionic) - enable the [universe repository](https://help.ubuntu.com/community/Repositories/Ubuntu), which contains packages necessary for Puppet Server, 16.04 (Xenial)
+* SLES 12 SP1
+
+> Note: Java 8 runtime packages do not exist in the standard repositories for Debian 8 (Jessie).  To install Puppet Server on Jessie, [configure the `jessie-backports` repository](https://backports.debian.org/Instructions/). 
+
+### Before you begin
+
 Puppet Server is configured to use 2 GB of RAM by default. If you're just testing an installation on a Virtual Machine, this much memory is not necessary. To change the memory allocation, see [Memory Allocation](#memory-allocation).
 
-> If you're also using PuppetDB, check its [requirements](https://puppet.com/docs/puppetdb/latest/index.html#system-requirements).
-
-## Java support
+#### Java support
 
 Puppet Server versions are tested against the following versions of Java:
 
@@ -24,17 +33,19 @@ Puppet Server versions are tested against the following versions of Java:
 
 Some Java versions may work with other Puppet Server versions, but we do not test or support those cases. Community submitted patches for support greater than Java 11 are welcome. Both Java 8 and 11 are considered long-term support versions and are planned to be supported by upstream maintainers until 2022 or later.
 
-## Install Puppet Server from packages
+## Install the Puppet Server package
 
 1.  [Enable the Puppet package repositories][repodocs], if you haven't already done so.
 
-2.  Install the Puppet Server package by running:
+2.  Install the Puppet Server package by running one of the following commands.
+
+Red Hat operating systems: 
 
 ````
 yum install puppetserver
 ````
 
-or
+Debian and Ubuntu: 
 
 ```
 apt-get install puppetserver
@@ -71,13 +82,8 @@ Puppet provides official packages that install Puppet Server 6.0 and all of its 
 
 > Note: Java 8 runtime packages do not exist in the standard repositories for Debian 8 (Jessie).  To install Puppet Server on Jessie, [configure the `jessie-backports` repository](https://backports.debian.org/Instructions/). 
 
-## Platforms without packages
 
-For platforms and architectures where no official packages are available, you can build Puppet Server from source. Such platforms are not tested, and running Puppet Server from source is not recommended for production use.
-
-For details, see [Running from Source](./dev_running_from_source.markdown).
-
-## Memory allocation
+## Running Puppet Server on a VM
 
 By default, Puppet Server is configured to use 2GB of RAM. However, if you want to experiment with Puppet Server on a VM, you can safely allocate as little as 512MB of memory. To change the Puppet Server memory allocation, you can edit the init config file.
 
@@ -94,7 +100,3 @@ By default, Puppet Server is configured to use 2GB of RAM. However, if you want 
     For more information about the recommended settings for the JVM, see [Oracle's docs on JVM tuning.](http://docs.oracle.com/cd/E15523_01/web.1111/e13814/jvm_tuning.htm)
 
 2. Restart the `puppetserver` service after making any changes to this file.
-
-## Reporting Issues
-
-Submit issues to our [bug tracker](https://tickets.puppet.com/browse/SERVER).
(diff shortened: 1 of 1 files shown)


documentation/install_from_packages.markdown
```diff
@@ -1,16 +1,25 @@
...
+* SLES 12 SP1
 
-## Java support
+> Note: Java 8 runtime packages do not exist in the standard repositories for Debian 8 (Jessie).  To install Puppet Server on Jessie, [configure the `jessie-backports` repository](https://backports.debian.org/Instructions/). 
+
+### Before you begin
+
+Puppet Server is configured to use 2 GB of RAM by default. If you're just testing an installation on a Virtual Machine, this much memory is not necessary. To change the memory allocation, see [Memory Allocation](#memory-allocation).
+
+#### Java support
```
clairecadman, line 22: Maybe this information be moved elsewhere?

documentation/install_from_packages.markdown
```diff
@@ -24,17 +33,19 @@ Puppet Server versions are tested against the following versions of Java:
...
 
 ````
 yum install puppetserver
 ````
 
-or
+Debian and Ubuntu: 
 
 ```
 apt-get install puppetserver
```
clairecadman, line 51: Is one of the following commands for Red Hat and the other for Debian and Ubuntu? 

```
systemctl start puppetserver	
``` 	
or 	
```	
service puppetserver start
```

documentation/install_from_packages.markdown
```diff
@@ -1,16 +1,25 @@
...
 layout: default
-title: "Puppet Server: Installing From Packages"
+title: "Install Puppet Server"
 canonical: "/puppetserver/latest/install_from_packages.html"
 ---
 
 [repodocs]: https://puppet.com/docs/puppet/latest/puppet_platform.html
 
-Puppet Server is configured to use 2 GB of RAM by default. If you're just testing an installation on a Virtual Machine, this much memory is not necessary. To change the memory allocation, see [Memory Allocation](#memory-allocation).
+Puppet provides official packages that install Puppet Server 6.0 and all of its prerequisites on x86_64 architectures for the following platforms, as part of [Puppet Platform][repodocs].
```
Magisus, line 9: This should probably just be "6" not "6.0", since the 6.x stream has many other Y releases that are also installed this way.

documentation/install_from_packages.markdown
Magisus: The former is the newer systemd command, but I think at this point all supported OSes support that? I'm not sure if there is a reason to prefer one over the other at this point.

documentation/install_from_packages.markdown
```diff
@@ -1,16 +1,25 @@
...
+* Debian 8 (Jessie), 9 (Stretch), 10 (Buster)
+* Ubuntu 18.04 (Bionic) - enable the [universe repository](https://help.ubuntu.com/community/Repositories/Ubuntu), which contains packages necessary for Puppet Server, 16.04 (Xenial)
+* SLES 12 SP1
 
-## Java support
+> Note: Java 8 runtime packages do not exist in the standard repositories for Debian 8 (Jessie).  To install Puppet Server on Jessie, [configure the `jessie-backports` repository](https://backports.debian.org/Instructions/). 
+
+### Before you begin
+
+Puppet Server is configured to use 2 GB of RAM by default. If you're just testing an installation on a Virtual Machine, this much memory is not necessary. To change the memory allocation, see [Memory Allocation](#memory-allocation).
```
Magisus, line 20: This link no longer works now that the section has been renamed.

documentation/install_from_packages.markdown
Magisus: I think it's useful to have here, since you need to manually install java a lot of the time along with the server package.

documentation/install_from_packages.markdown
```diff
@@ -1,16 +1,25 @@
...
 
 [repodocs]: https://puppet.com/docs/puppet/latest/puppet_platform.html
 
-Puppet Server is configured to use 2 GB of RAM by default. If you're just testing an installation on a Virtual Machine, this much memory is not necessary. To change the memory allocation, see [Memory Allocation](#memory-allocation).
+Puppet provides official packages that install Puppet Server 6.0 and all of its prerequisites on x86_64 architectures for the following platforms, as part of [Puppet Platform][repodocs].
 
-> If you're also using PuppetDB, check its [requirements](https://puppet.com/docs/puppetdb/latest/index.html#system-requirements).
+* Red Hat Enterprise Linux 6, 7
+* Debian 8 (Jessie), 9 (Stretch), 10 (Buster)
+* Ubuntu 18.04 (Bionic) - enable the [universe repository](https://help.ubuntu.com/community/Repositories/Ubuntu), which contains packages necessary for Puppet Server, 16.04 (Xenial)
```
mwaggett, line 13: not understanding this note.. We have ubuntu 16.04 packages..

documentation/install_from_packages.markdown
mwaggett: is `repodocs` meant to be a place holder for something?

documentation/install_from_packages.markdown
```diff
@@ -94,7 +100,3 @@ By default, Puppet Server is configured to use 2GB of RAM. However, if you want
     For more information about the recommended settings for the JVM, see [Oracle's docs on JVM tuning.](http://docs.oracle.com/cd/E15523_01/web.1111/e13814/jvm_tuning.htm)
 
 2. Restart the `puppetserver` service after making any changes to this file.
-
-## Reporting Issues
```
mwaggett, line 98: Why remove this?

documentation/install_from_packages.markdown
clairecadman: Do you think we could just have `systemctl start puppetserver` for the Server 6 docs then?

documentation/install_from_packages.markdown
clairecadman: I've moved 16.04 back to where this was. Is enabling the universe repository still needed for these versions?

mwaggett: @clairecadman looks like yes, that is still necessary for ubuntu 18.04 (bionic).

I think it makes sense to move this note, though. It's pretty much the same as the one here: https://github.com/puppetlabs/puppetserver/pull/2361/files#diff-bb34594aa08d2f40dd4382784111dd7aR18
The reason this additional repository is necessary is because Java 8 runtime packages (openjdk-8-jre-headless) are not available in the standard repositories on ubuntu-18.04.

documentation/install_from_packages.markdown
mwaggett: imo, yes

documentation/install_from_packages.markdown
```diff
@@ -95,6 +81,4 @@ By default, Puppet Server is configured to use 2GB of RAM. However, if you want
 
 2. Restart the `puppetserver` service after making any changes to this file.
 
-## Reporting Issues
-
-Submit issues to our [bug tracker](https://tickets.puppet.com/browse/SERVER).
+> Note: Submit issues to our [bug tracker](https://tickets.puppet.com/browse/SERVER).
```
mwaggett, line 84: I feel like putting this in a Note sort of hides this info.. It also seems like it's specific to the `Running Puppet Server on a VM`? which it's not.

I kinda like the separate `Reporting Issues` section. Is there a reason we removed it?

clairecadman: Yes you're right it does look like it belongs to that section, so I have moved it to the top now. Let me know what you think.

I initially removed it to keep this page consistent with the Puppet docs pages, and also other Server docs pages, which don't have a report an issue section.  Is it more likely that they would need to report a problem when installing? If so we can put it back, or if you think the note at the top works just as well, we can leave it there.

clairecadman:
@hestonhoffman I think this is close to being ready to merge. If you could take a look and see if it addresses the problems you ran into when following these steps. (I'm working on fixing up the enabling package repositories page separately). Thank you!

documentation/install_from_packages.markdown
mwaggett: ahh, I'm cool with removing the section in order to be consistent 👍

documentation/install_from_packages.markdown
```diff
@@ -50,34 +65,7 @@ There is no `-` in the package name.
 systemctl start puppetserver
```
hestonhoffman, line 65: Most users will probably know/figure this out, but I wonder if it's worth adding `sudo` to these commands?

hestonhoffman: I think it would be good to add a command to the end, just for users to check that it's working. Maybe just `puppetserver -v` to check the version?

clairecadman:
Thanks @hestonhoffman! I also made a few more changes @Magisus and @mwaggett , if you could please take one last look and merge if you're happy with it. Thanks!

documentation/install_from_packages.markdown
```diff
@@ -1,16 +1,27 @@
 ---
 layout: default
-title: "Puppet Server: Installing From Packages"
+title: "Install Puppet Server"
 canonical: "/puppetserver/latest/install_from_packages.html"
 ---
 
-[repodocs]: https://puppet.com/docs/puppet/latest/puppet_platform.html
+Puppet Server is a required application that runs on the Java Virtual Machine (JVM). It performs the role of the master node that controls configuration information for one or more managed agent nodes
```
mwaggett, line 7: 1. Missing a period at the end here.
2. Since we're working on moving away from the "master" terminology, I wonder if there is a better way to describe this role.

documentation/install_from_packages.markdown
```diff
@@ -1,16 +1,27 @@
...
 
-Puppet Server is configured to use 2 GB of RAM by default. If you're just testing an installation on a Virtual Machine, this much memory is not necessary. To change the memory allocation, see [Memory Allocation](#memory-allocation).
+> Note: If you have any issues with the steps below, submit these to our [bug tracker](https://tickets.puppet.com/browse/SERVER).
 
-> If you're also using PuppetDB, check its [requirements](https://puppet.com/docs/puppetdb/latest/index.html#system-requirements).
+## Before you begin
 
-## Java support
+Review the system requirements and make sure you have a supported version of Java.
```
mwaggett, line 13: What are you referring to when you say 'system requirements'? Do you just mean the supported operating systems below? Or another section we should link to? I think this could be a little clearer.

documentation/install_from_packages.markdown
```diff
@@ -1,16 +1,27 @@
...
-## Java support
+Review the system requirements and make sure you have a supported version of Java. 
+
+### Supported operating systems
+
+Puppet provides official packages that install Puppet Server 6 and all of its prerequisites on x86_64 architectures for the following platforms:
+
+* Red Hat Enterprise Linux 6, 7
+* Debian 8 (Jessie), 9 (Stretch), 10 (Buster)
+* Ubuntu 18.04 (Bionic), 16.04 (Xenial)
```
mwaggett, line 21: nitpick, but Red Hat and Debian have older versions listed first while Ubuntu lists newest first - I think we should be consistent.

documentation/install_from_packages.markdown
```diff
@@ -21,20 +32,25 @@ Puppet Server versions are tested against the following versions of Java:
 | 6.0-6.5  | 8, 11 (experimental)  |
 | 6.6 and later  | 8, 11  |
 
-
 Some Java versions may work with other Puppet Server versions, but we do not test or support those cases. Community submitted patches for support greater than Java 11 are welcome. Both Java 8 and 11 are considered long-term support versions and are planned to be supported by upstream maintainers until 2022 or later.
 
-## Install Puppet Server from packages
+> Note: Java 8 runtime packages do not exist in the standard repositories for Debian 8 (Jessie) or Ubuntu 18.04 (Bionic).  To install Puppet Server on Jessie, [configure the `jessie-backports` repository](https://backports.debian.org/Instructions/). To install Puppet Sevrer on Bionic, enable the [universe repository](https://help.ubuntu.com/community/Repositories/Ubuntu).
```
mwaggett, line 37:
```suggestion
> Note: Java 8 runtime packages do not exist in the standard repositories for Debian 8 (Jessie) or Ubuntu 18.04 (Bionic).  To install Puppet Server on Jessie, [configure the `jessie-backports` repository](https://backports.debian.org/Instructions/). To install Puppet Server on Bionic, enable the [universe repository](https://help.ubuntu.com/community/Repositories/Ubuntu).
```

clairecadman:
@mwaggett Thanks for spotting these! I've fixed the errors and made the updates. Let me know what you think if line 7 re. 'master'. We will have to go through the docs and change this term when they come up with a new name anyway, so we can always come back to this.

clairecadman:
Thanks @mwaggett! I've just had this feedback come in from a user: 

"We should enlist the ways of creating Puppet Infrastructure - e.g. we should explicitly state the possibility to deploy Puppet using the Helm chart of Puppet Server (https://github.com/puppetlabs/puppetserver-helm-chart)."

I've never heard of doing this before, is this a common thing / something you would want to advertise in the docs? 

If not please merge this when you're ready. Thanks!

mwaggett:
@clairecadman I think that's related to our pupperware (puppet in containers) effort. I would recommend addressing that in a separate ticket with the pupperware team. /cc @underscorgan

PR #2362, opened 2020-07-01
Rewrite Docker health check to avoid making assumptions

runejuhl:
The previous health check would break if more than one certificate (other than
`ca.pem`) is present in `${SSLDIR}/certs`:

~~~
# bash -x healthcheck.sh
+ set -x
+ set -e
++ ls /etc/puppetlabs/puppet/ssl/certs
++ grep --invert-match ca.pem
+ certname='puppet.company.com.pem
puppet.company.internal.pem
puppet01.company.pem
puppet02.company.pem
puppetdiff.company.pem
puppetserver.company.internal.pem'
++ basename puppet.company.com.pem puppet.company.internal.pem puppet01.company.pem puppet02.company.pem puppetdiff.company.pem puppetserver.company.internal.pem .pem
basename: extra operand 'puppet01.company.pem'
Try 'basename --help' for more information.
+ hostname=
+ exit 1
~~~

This PR makes use of `puppet config print` to read values from the config
instead of making assumption based on the name of PEM files.

To avoid unnecessarily running `puppet config print` the output is cached in
`/tmp/puppet-config-values`. If the file exists it will be used, otherwise
`puppet config print` will be called and the result placed in this file.

### DIFF
--- a/docker/puppetserver/healthcheck.sh
+++ b/docker/puppetserver/healthcheck.sh
@@ -1,18 +1,26 @@
 #!/usr/bin/env bash
 
-set -x
-set -e
+set -euo pipefail
 
-certname=$(ls "${SSLDIR}/certs" | grep --invert-match ca.pem) && \
-hostname=$(basename $certname .pem) && \
-hostprivkey="${SSLDIR}/private_keys/$certname" && \
-hostcert="${SSLDIR}/certs/$certname" && \
-localcacert="${SSLDIR}/certs/ca.pem" && \
-curl --fail \
---resolve "${hostname}:${PUPPET_MASTERPORT}:127.0.0.1" \
---cert   $hostcert \
---key    $hostprivkey \
---cacert $localcacert \
-"https://${hostname}:${PUPPET_MASTERPORT}/status/v1/simple" \
-|  grep -q '^running$' \
-|| exit 1
+declare certname hostprivkey hostcert localcacert
+
+while read -r line; do
+  IFS=' = ' read -r key value <<< "${line}"
+  printf -v "${key}" '%s' "${value}"
+done < <(cat /tmp/puppet-config-values 2>/dev/null ||
+           puppet config print certname hostprivkey hostcert localcacert |
+             tee /tmp/puppet-config-values)
+# In the above we first try to read cached values to avoid calling `puppet
+# config print` on every health check invocation. If the file doesn't exist we
+# call `puppet config print` on the values we need and then use `read` to
+# extract the key and value and then use `printf` to declare the variable using
+# these.
+
+curl --fail                                                 \
+     --verbose                                              \
+     --resolve "${certname}:${PUPPET_MASTERPORT}:127.0.0.1" \
+     --cert   "${hostcert}"                                 \
+     --key    "${hostprivkey}"                              \
+     --cacert "${localcacert}"                              \
+     "https://${certname}:${PUPPET_MASTERPORT}/status/v1/simple" |
+  grep -q '^running$' || exit 1


Iristyle:
We very intentionally moved away from using `puppet config print` for performance reasons -- see https://github.com/puppetlabs/puppetserver/commit/72eea601719c18c8e5556349f8875ec6b2293300

Why do you have so many certs there? That doesn't seem right...

runejuhl:
> We very intentionally moved away from using `puppet config print` for performance reasons -- see [72eea60](https://github.com/puppetlabs/puppetserver/commit/72eea601719c18c8e5556349f8875ec6b2293300)

I understand, which is why this check makes sure to only run `puppet config print` a single time; see the comment in the source and in the commit message.

> Why do you have so many certs there? That doesn't seem right...

That's not really the point; the issue is that the current health check makes assumptions that may not hold true. This would also be an issue with just a single extraneous certificate in the folder, as the current check is written to only ever be able to handle a single certificate.

What's worse is that the current check makes assumptions about the location of `hostprivkey`, `hostcert` and  `localcacert`, even though the Puppet server configuration explicitly allows these values to be configured using [`puppet.conf`](https://puppet.com/docs/puppetserver/latest/configuration.html).

Even if you had these certificates in `SSLDIR` and everything else worked, you still might run into issues, as the variables in the current health check are unquoted, meaning that whitespace in `SSLDIR` would break the check. `SSLDIR` is a [`Dockerfile` environment variable](https://github.com/puppetlabs/puppetserver/blob/5e6065eb7775638c4548bdcc79ff61f2c8b2373a/docker/puppetserver/Dockerfile#L26), which I take to mean that it is configurable.

These examples might seem a bit contrived, but I did manage to hit at least one of the issues today, even though the setup causing the error is completely valid.

Iristyle:
`SSLDIR` is actually not intended to (and shouldn't be) user modified. We should change that to avoid confusion, since the wrong message has clearly been sent here. All of the documented configuration values are at https://github.com/puppetlabs/puppetserver/tree/master/docker/puppetserver#configuration

The intent of the container is to be prescriptive about many things to reduce the possible configuration surface area and complexity. I think that philosophically we want to only surface the bare minimum number of possible options to end users -- at times that means hardcoding values where it makes sense (such as `SSLDIR` and `LOGDIR`) b/c of how userdata should be mapped to volumes and how upgrades from open source to enterprise containers should happen.

We've actually talked about generating an error on startup if there are extraneous certs in SSLDIR, as (at least in development) it can be caused by accidentally keeping around volumes, but changing the servers certname (which can be problematic). So I'd still like to understand how you ended up with all of those certs in there to see if your workflow is one we should consider.

runejuhl:
> `SSLDIR` is actually not intended to (and shouldn't be) user modified. We should change that to avoid confusion, since the wrong message has clearly been sent here. All of the documented configuration values are at https://github.com/puppetlabs/puppetserver/tree/master/docker/puppetserver#configuration
> 
> The intent of the container is to be prescriptive about many things to reduce the possible configuration surface area and complexity. I think that philosophically we want to only surface the bare minimum number of possible options to end users -- at times that means hardcoding values where it makes sense (such as `SSLDIR` and `LOGDIR`) b/c of how userdata should be mapped to volumes and how upgrades from open source to enterprise containers should happen.

FWIW I'm on the fence about this argument. From a sysadmin perspective I prefer having the possibility of making changes like these, even if they may seem nonsensical at first. It's not rare that I find a solution that makes use of a rarely used configuration item, or that I end up with something that works perfectly in a not-quite-standard setup, due to various external limitations, but may be achieved easily because the components were written in a transparent manner and took care to make everything work correctly, even if non-default values were used.

From a programmer perspective I absolutely see the value in reducing complexity; it cuts down on weird workarounds, simplifies testing and makes it easier to wrap your head around a system.

...but that's not even the actual issue here; the issue is that the Puppetserver Docker image makes some assumptions that may not always be true. As I wrote, it makes the assumptions that `hostprivkey`, `hostcert` and `localcacert` are all stored under `SSLDIR`, even though the documentation explicitly states that these are configurable values.

This PR handles all cases; it allows both a non-default `SSLDIR` **and** it allows `hostprivkey`, `hostcert` and `localcacert` to be configured, for whatever reason there might be.

> We've actually talked about generating an error on startup if there are extraneous certs in SSLDIR, as (at least in development) it can be caused by accidentally keeping around volumes, but changing the servers certname (which can be problematic). So I'd still like to understand how you ended up with all of those certs in there to see if your workflow is one we should consider.

That is indeed what happened in our case; the system started out as physical ages ago and has been through some moves before it ended up being run in Docker. But whatever the reason might be for having multiple certificates, I don't see why it should not be handled properly. Especially when the patch to resolve this issue is right here.

Iristyle:
Sorry this ended up getting lost in the shuffle. 

> it makes the assumptions that hostprivkey, hostcert and localcacert are all stored under SSLDIR, even though the documentation explicitly states that these are configurable values.

Not sure I agree. Where is this stated?


I'm also reluctant to add a custom caching solution to an entrypoint. I'll try and put up a PR in the next week to tackle the underlying problem for a different angle. I'll also get the containers updated so that `SSLDIR` and `LOGDIR` are not viewed as configurable env vars, given they are not.

Magisus:

justinstoller:
> I'll try and put up a PR in the next week to tackle the underlying problem from a different angle. I'll also get the containers updated so that SSLDIR and LOGDIR are not viewed as configurable env vars, given they are not.

@Iristyle is this done? I assume whatever work you're doing/did happened elsewhere?

PR #2367, opened 2020-07-07
(TK-489) Bump to clj-parent 4.5.4

mwaggett:
This commit updates clj-parent to 4.5.4, which updates trapperkeeper-metrics to
1.3.1. This includes support for trapperkeeper-authorization.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -27,7 +27,7 @@
 
   :min-lein-version "2.9.1"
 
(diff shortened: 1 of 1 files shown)


Magisus:
Do we want to wait and pick up the tk-metrics change before bumping this?

mwaggett:
oh yeah, I forgot that would also go through clj-parent. sure, we can wait.

PR #2372, opened 2020-07-21
(SERVER-2795) Fix CSR deletion and cleanup whitespace

Magisus:
This commit fixes a misplaced parenthesis that made it impossible to
delete CSRs via the API. It also cleans up some erroneous whitespace.

### DIFF
--- a/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
+++ b/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
@@ -64,36 +64,36 @@
       ;; Respond to all CSR validation failures with a 400
       (middleware-utils/plain-response 400 msg))))
 
-  (schema/defn format-http-date :- (schema/maybe DateTime)
-    "Formats an http-date into joda time.  Returns nil for malformed or nil
-     http-dates"
-    [http-date :- (schema/maybe schema/Str)]
-    (when http-date
-      (try
-        (time-format/parse
-          (time-format/formatters :rfc822)
-          (string/replace http-date #"GMT" "+0000"))
-        (catch IllegalArgumentException e
-          nil))))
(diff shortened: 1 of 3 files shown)


Magisus:
So we have [tests](https://github.com/puppetlabs/puppetserver/blob/master/test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj#L176) that call the handlers for this function directly, but nothing for the CSR HTTP API. Do we think it's worth adding those?

Magisus:
I had a test that used mock JRubies, and I just found where `autosign=true` was coming from in that context: https://github.com/puppetlabs/puppetserver/blob/master/test/unit/puppetlabs/services/jruby/jruby_puppet_testutils.clj#L232. We do not provide good mechanisms for overriding things in that hash. I think changing that would be the most straightforward solution here.

mwaggett:
I think it would be good to add some sort of test that would've caught this typo, whatever that looks like.

Magisus:
Verified that this test failed without the fix.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -341,9 +341,9 @@
         (GET [""] [subject]
           (handle-get-certificate-request subject ca-settings))
         (PUT [""] [subject :as {body :body}]
-          (handle-put-certificate-request! subject body ca-settings)))
+          (handle-put-certificate-request! subject body ca-settings))
         (DELETE [""] [subject]
-          (handle-delete-certificate-request! subject ca-settings))
+          (handle-delete-certificate-request! subject ca-settings)))
```
justinstoller, line 346: 🤦

Magisus: Pretty much.

PR #2374, opened 2020-07-22
(SERVER-2551) Make CA cert and CRL expirations available via API

Magisus:
This commit adds a `puppet-ca/v1/ca-expirations` endpoint that returns the expiration date of each CA cert and CRL in their respective bundles, keyed by common name. For the certs, this is the "not-after" date, while for CRLs it is the "next-update" date.

Interested in feedback on the format of the response and the name of the endpoint, as well as the usual things.

### DIFF
--- a/ezbake/config/conf.d/auth.conf
+++ b/ezbake/config/conf.d/auth.conf
@@ -87,6 +87,17 @@ authorization: {
             name: "puppetlabs cert statuses"
         },
         {
+            # Allow authenticated access to the CA expirations endpoint
+            match-request: {
+                path: "/puppet-ca/v1/expirations"
+                type: path
+                method: get
+            }
+            allow: "*"
+            sort-order: 500
+            name: "puppetlabs CA cert and CRL expirations"
+        },
+        {
             # Allow unauthenticated access to the status service endpoint
             match-request: {
                 path: "/status/v1/services"
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -6,7 +6,7 @@
            [java.nio.file.attribute FileAttribute PosixFilePermissions]
            (java.security PrivateKey PublicKey KeyPair)
            (org.joda.time DateTime)
-           (java.security.cert CRLException)
+           (java.security.cert CRLException X509CRL)
            (sun.security.x509 X509CertImpl))
   (:require [me.raynes.fs :as fs]
             [schema.core :as schema]
@@ -1598,3 +1598,38 @@
   [custom-oid-mapping-file :- (schema/maybe schema/Str)]
   (merge (get-custom-oid-mappings custom-oid-mapping-file)
          (clojure.set/map-invert puppet-short-names)))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; CA status info
+
+(schema/defn ca-expiration-dates
+  "Returns of a map of subject names of certs in the CA bundle
+  to their expiration dates."
+  [ca-cert-file :- schema/Str]
+  (let [ca-cert-bundle (utils/pem->certs ca-cert-file)]
+    (reduce (fn [cert-expirations cert]
+              (assoc cert-expirations
+                (-> cert
+                    (.getSubjectDN)
+                    (.getName)
+                    (utils/x500-name->CN))
+                (-> cert
+                    (.getNotAfter)
+                    (format-date-time))))
+            {}
+            ca-cert-bundle)))
+
+(schema/defn crl-expiration-dates
+  [crl-chain-file :- schema/Str]
+  (let [crl-chain (utils/pem->crls crl-chain-file)]
+    (reduce (fn [crl-expirations crl]
+              (assoc crl-expirations
+                (-> crl
+                    (.getIssuerDN)
+                    (.getName)
+                    (utils/x500-name->CN))
+                (-> crl
+                    (.getNextUpdate)
+                    (format-date-time))))
+            {}
+            crl-chain)))
--- a/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
+++ b/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
@@ -104,6 +104,14 @@
         (rr/status ((response :outcome) outcomes->codes))
         (rr/content-type "text/plain"))))
 
+(schema/defn handle-get-ca-expirations
+  [ca-settings :- ca/CaSettings]
+  (let [response {:ca-certs (ca/ca-expiration-dates (:cacert ca-settings))
+                  :crls (ca/crl-expiration-dates (:cacrl ca-settings))}]
+    (-> (rr/response (cheshire/generate-string response))
+        (rr/status 200)
+        (rr/content-type "application/json"))))
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Web app
 
@@ -345,7 +353,9 @@
         (DELETE [""] [subject]
           (handle-delete-certificate-request! subject ca-settings)))
       (GET ["/certificate_revocation_list/" :ignored-node-name] request
-        (handle-get-certificate-revocation-list request ca-settings)))
+        (handle-get-certificate-revocation-list request ca-settings))
+      (GET ["/expirations"] request
+        (handle-get-ca-expirations ca-settings)))
     (comidi/not-found "Not Found")))
 
 (schema/defn ^:always-validate
--- a/test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
+++ b/test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
@@ -1,22 +1,24 @@
 (ns puppetlabs.services.certificate-authority.certificate-authority-int-test
   (:require
-    [clojure.test :refer :all]
-    [puppetlabs.kitchensink.core :as ks]
-    [puppetlabs.ssl-utils.core :as utils]
-    [puppetlabs.puppetserver.bootstrap-testutils :as bootstrap]
-    [puppetlabs.services.jruby.jruby-puppet-testutils :as jruby-testutils]
-    [puppetlabs.puppetserver.testutils :as testutils :refer
-     [ca-cert localhost-cert localhost-key ssl-request-options http-get]]
-    [puppetlabs.trapperkeeper.testutils.logging :as logutils]
-    [schema.test :as schema-test]
-    [me.raynes.fs :as fs]
-    [cheshire.core :as json]
-    [puppetlabs.http.client.sync :as http-client]
-    [puppetlabs.ssl-utils.core :as ssl-utils]
-    [puppetlabs.puppetserver.certificate-authority :as ca]
-    [puppetlabs.services.ca.certificate-authority-core :refer :all]
-    [ring.mock.request :as mock]
-    [me.raynes.fs :as fs])
+   [clojure.test :refer :all]
+   [puppetlabs.kitchensink.core :as ks]
+   [puppetlabs.ssl-utils.core :as utils]
+   [puppetlabs.puppetserver.bootstrap-testutils :as bootstrap]
+   [puppetlabs.services.jruby.jruby-puppet-testutils :as jruby-testutils]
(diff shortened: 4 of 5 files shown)


Magisus:
Going to test this out in a build.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1598,3 +1598,39 @@
...
+(schema/defn ca-expiration-dates
+  "Returns of a map of subject names of certs in the CA bundle
+  to their expiration dates."
+  [ca-cert-file :- schema/Str]
+  (let [ca-cert-bundle (utils/pem->certs ca-cert-file)]
+    (reduce (fn [cert-expirations cert]
+              (assoc cert-expirations
+                ;; Is this what we want to key off of?
+                (-> cert
+                    (.getSubjectDN)
```
justinstoller, line 1614: I think so, this will give something more like "Puppet CA: foo.bar.baz" in actual deployments

ezbake/config/conf.d/auth.conf
```diff
@@ -86,6 +86,16 @@ authorization: {
             sort-order: 500
             name: "puppetlabs cert statuses"
         },
+        {
```
mwaggett, line 89: nit, but the other entries in here have a comment about what they do

src/clj/puppetlabs/puppetserver/certificate_authority.clj
jonathannewman: From a customer facing standpoint I think that makes a lot of sense.  Works for me.

ezbake/config/conf.d/auth.conf
```diff
@@ -86,6 +86,16 @@ authorization: {
...
+            match-request: {
+                path: "/puppet-ca/v1/ca-expirations"
+                type: path
+                method: get
+            }
+            allow: "*"
```
mwaggett, line 95: what made you choose this, as opposed to, like, what we allow for the `certificate_statuses` endpoint above?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
Magisus: Yeah that's what I saw in testing, I'd like @jonathannewman to confirm that's a useful key.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -104,6 +104,14 @@
         (rr/status ((response :outcome) outcomes->codes))
         (rr/content-type "text/plain"))))
 
+(schema/defn handle-get-ca-expirations
+  [ca-settings :- ca/CaSettings]
+  (let [response {:ca-certs (ca/ca-expiration-dates (:cacert ca-settings))
+                  :crls (ca/crl-expiration-dates (:cacrl ca-settings))}]
```
mwaggett, line 110: Would it be more or less confusing if the response keys more-closely matched the settings keys, e.g. `:cacerts` and `cacrls`?

Magisus:
Tested:
```
curl --cacert /Users/aileen/.puppetlabs/etc/puppet/ssl/ca/ca_crt.pem --cert /Users/aileen/.puppetlabs/etc/puppet/ssl/certs/localhost.pem --key /Users/aileen/.puppetlabs/etc/puppet/ssl/private_keys/localhost.pem https://localhost:8140/puppet-ca/v1/ca-expirations
{"ca-certs":
   {"Puppet CA: localhost":"2034-06-20T17:55:52UTC","Puppet Root CA: 4ba26d12fd6474":"2034-06-20T17:55:52UTC"},
 "crls":
   {"Puppet CA: localhost":"2024-12-16T23:11:22UTC","Puppet Root CA: 4ba26d12fd6474":"2034-06-20T17:55:52UTC"}}
```

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
```diff
@@ -581,3 +581,27 @@
...
+      {:ssl-cert (str bootstrap/master-conf-dir "/ssl/certs/localhost.pem")
+       :ssl-key (str bootstrap/master-conf-dir "/ssl/private_keys/localhost.pem")
+       :ssl-ca-cert (str bootstrap/master-conf-dir "/ssl/ca/ca_crt.pem")
+       :ssl-crl-path (str bootstrap/master-conf-dir "/ssl/crl.pem")}}
+     (let [options {:ssl-cert (str bootstrap/master-conf-dir "/ssl/ca/ca_crt.pem")
+                    :ssl-key (str bootstrap/master-conf-dir "/ssl/ca/ca_key.pem")
+                    :ssl-ca-cert (str bootstrap/master-conf-dir "/ssl/ca/ca_crt.pem")
+                    :as :text}
+           response (http-client/get "https://localhost:8140/puppet-ca/v1/ca-expirations"
+                                     (merge options {:headers {"Accept" "application/json"}}))]
```
mwaggett, line 603: why `merge` instead of just including `:headers` when you define `options`?

justinstoller:
I think "ca-expirations" sounds redundant, but if we provided an "expirations" endpoint I assume folks might think it provides more than the CA & CRL expirations. Also, "expirations" is a vague enough word to match both "nextUpdate" and "notAfter" though I think those are fairly different concepts. I think the usefulness of this outweighs concerns about coherence of terminology.

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -1400,3 +1400,17 @@
     (testing "and the file doesn't exist"
       (testing "the result is nil"
         (is (nil? (create-csr-attrs-exts "does/not/exist.yaml")))))))
+
+(deftest ca-expiration-dates-test
+  (testing "returns a map of names to dates"
+    (let [settings  (testutils/ca-sandbox! bundle-cadir)
```
mwaggett, line 1406:
```suggestion
    (let [settings (testutils/ca-sandbox! bundle-cadir)
```

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -1400,3 +1400,17 @@
...
+          expiration-map (ca-expiration-dates (:cacert settings))]
+      (is (= "2036-09-06T05:58:33UTC" (get expiration-map "rootca.example.org")))
+      (is (= "2036-09-06T06:09:14UTC" (get expiration-map "intermediateca.example.org"))))))
+
+(deftest crl-expiration-dates-test
+  (testing "returns a map of names to dates"
+    (let [settings  (testutils/ca-sandbox! bundle-cadir)
```
mwaggett, line 1413:
```suggestion
    (let [settings (testutils/ca-sandbox! bundle-cadir)
```

ezbake/config/conf.d/auth.conf
Magisus: That endpoint is generally used to change the state of certificates (e.g. to sign or revoke them). This endpoint is read-only, so I figured it didn't need to be as secured. The one above allows _only_ the master to query it, which isn't particularly useful here. This one requires someone to have puppet certs, but anyone with puppet certs can read it, for example, the console.

In PE, we could lock it down tighter and specifically allow the console only. But in FOSS we don't have that option, so this is a good compromise.

If someone thinks we should do something different, please let me know.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
Magisus: I didn't want to do `cacrls` because there isn't really any such thing as a " CA crl". We make the distinction in the settings because we keep the CRL two places, one for the CA and another for the client. But they have the same expiration, it's just two copies of the same file.

No strong opinion on whether to keep or get rid of the hyphen. In general, only the very oldest settings do not have hyphens. In newer code, we tend to use them (or underscores) to make things easier to read. Since this info isn't directly tied to Puppet's settings at all, I decided to add one.

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
Magisus: Copied from another test. Probably don't need the `let` at all here since there's no repetition.

ezbake/config/conf.d/auth.conf
mwaggett: That makes sense to me, thanks for explaining!

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
mwaggett: Fair. I don't have strong feelings.

Magisus:
Do we want to drop the `ca` and just do `expirations` to give us the freedom to add more? Is there any use case for also reporting the expirations of signed certs somewhere down the line, for example?

ezbake/config/conf.d/auth.conf
Magisus: Filed https://tickets.puppetlabs.com/browse/PE-29718 for adding the auth rules in PE. Appreciate thoughts on the question in there about whether to allow only the console.

Magisus:
Oh dang, forgot the stuff in `target` is generated and not actual fixtures. Need to update the endpoint tests to handle that.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -6,7 +6,7 @@
            [java.nio.file.attribute FileAttribute PosixFilePermissions]
            (java.security PrivateKey PublicKey KeyPair)
            (org.joda.time DateTime)
-           (java.security.cert CRLException)
+           (java.security.cert CRLException X509CRL)
```
mwaggett, line 9: Is this actually getting used?

Magisus: Yeah, it's where `.getNextUpdate` comes from.

mwaggett: ahh 👍

Magisus:
I think I like removing `ca` from the URL. Going to make that change.

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -1400,3 +1400,17 @@
...
+    (let [settings (testutils/ca-sandbox! bundle-cadir)
+          expiration-map (ca-expiration-dates (:cacert settings))]
+      (is (= "2036-09-06T05:58:33UTC" (get expiration-map "rootca.example.org")))
+      (is (= "2036-09-06T06:09:14UTC" (get expiration-map "intermediateca.example.org"))))))
+
+(deftest crl-expiration-dates-test
+  (testing "returns a map of names to dates"
+    (let [settings (testutils/ca-sandbox! bundle-cadir)
+          expiration-map (crl-expiration-dates (:cacrl settings))]
+      (is (= "2016-10-11T06:42:52UTC" (get expiration-map "rootca.example.org")))
```
jonathannewman, line 1415: Is the UTC timestamp suffix the norm for puppetserver?  Most of the PE apps use "Z" for the suffix.

Magisus: I used the existing `time-format` methods in this codebase, so yes, I think it is. But I can return something different if you'd prefer. If so, an example formatter from one of the apps you're thinking of would be very helpful.

jonathannewman: I think we try to align with ISO-8601 as much as possible.  https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators

jonathannewman: I'm trying to find an example in one of the other repos.

Magisus: Specifically, we do it this way for compatability with the Ruby, probably a holdover from the old CA. See https://github.com/puppetlabs/puppetserver/blob/master/src/clj/puppetlabs/puppetserver/certificate_authority.clj#L557. But I don't have an issue with returning that format from this endpoint.

jonathannewman: Likely best to stick with tradition.

jonathannewman: If that's the format used in other puppetserver APIs, I think sticking with it makes sense.

Magisus: Looks like it is what's used to return `not-after` dates [for agent certs](https://github.com/puppetlabs/puppetserver/blob/master/src/clj/puppetlabs/puppetserver/certificate_authority.clj#L1424-L1434) via the API.

PR #2375, opened 2020-07-28
Merge 5.3.x -> master

Magisus:


Magisus:
Hm, yeah maybe. Though I still feel like it usually does...

PR #2377, opened 2020-07-30
(SERVER-2233) Add query params for filtering certificate statuses

mwaggett:
This commit updates the `certificate_statuses` API endpoint to accept query
params for filtering by a given 'state'. For example, this can be used to
identify all agents with unsigned (i.e. requested) certs.

### DIFF
--- a/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
+++ b/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
@@ -318,7 +318,7 @@
         (assoc-in [:representation :media-type] "text/plain")))))
 
 (defresource certificate-statuses
-  [settings]
+  [request settings]
   :allowed-methods [:get]
 
   :available-media-types media-types
@@ -329,9 +329,12 @@
 
   :handle-ok
   (fn [context]
-    (->
-      (ca/get-certificate-statuses settings)
-      (as-json-or-pson context))))
+    (let [queried-state (get-in request [:params "state"])]
+      (->
+        (if (some #(= queried-state %) ["requested" "signed" "revoked"])
+          (ca/filter-by-certificate-state settings queried-state)
+          (ca/get-certificate-statuses settings))
+        (as-json-or-pson context)))))
 
 (schema/defn ^:always-validate web-routes :- bidi-schema/RoutePair
   [ca-settings :- ca/CaSettings]
@@ -340,8 +343,8 @@
       (ANY ["/certificate_status/" :subject] [subject]
         (certificate-status subject ca-settings))
       (comidi/context ["/certificate_statuses/"]
-        (ANY [[#"[^/]+" :ignored-but-required]] []
-          (certificate-statuses ca-settings))
+        (ANY [[#"[^/]+" :ignored-but-required]] request
+          (certificate-statuses request ca-settings))
         (ANY [""] [] (middleware-utils/plain-response 400 "Missing URL Segment")))
       (GET ["/certificate/" :subject] [subject]
         (handle-get-certificate subject ca-settings))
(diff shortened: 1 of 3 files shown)


src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -329,9 +329,19 @@
 
   :handle-ok
   (fn [context]
-    (->
-      (ca/get-certificate-statuses settings)
-      (as-json-or-pson context))))
+    (let [query-params (:params request)
```
mwaggett, line 332: `:params` doesn't include the query params when testing.
I was originally using a `:query-params` key, but that key doesn't exist at all in the `request` from testing.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -329,9 +329,18 @@
...
+          queried-state (get query-params "state")]
+      (->
+        (case queried-state
+          "requested" (ca/filter-requested-certificates settings)
+          "signed" (ca/filter-signed-certificates settings)
+          "revoked" (ca/filter-revoked-certificates settings)
```
Magisus, line 338: I've been going back and forth all day about having separate functions for these... Since we are pretty sure the set of filters won't change (there aren't any other valid cert states that I know of), it's probably alright, but this kind of does seem like a lot of boiler plate for what could instead just be something like:
```
(if (some %(= queried-state) ["requested" "signed" "revoked"])
  (ca/filter-by-certificate-state settings queried-state)
  (ca/get-certificate-statuses settings))
```
Thoughts? I don't have strong opinions, there's just not really any processing in the specific methods so I'm not sure it's worth having that layer.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -329,9 +329,18 @@
...
+          ;; This will return all statuses if someone requests a state that we
+          ;; don't know about. Is that desirable? Or should we only return all
+          ;; statuses if no state is requested?
+          (ca/get-certificate-statuses settings))
```
Magisus, line 342: Seems like opinion is split on whether to silently ignore bad query params or raise, for example, a 400 Bad Request with a useful message (perhaps including a list of valid states). I feel like the second is probably better UX, but on the other hand, we probably don't really expect people to be using this by hand much.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
mwaggett: that's fair, I like yours.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
mwaggett: yeah, I agree that a message is better UX, but figured I'd ask before investigating how to raise 😅

mwaggett: the other reason I didn't just add it is because, currently, if you supply _any_ query params, they are happily ignored.

Magisus: Hmm... wonder if this becomes a backwards-compatability question then.

mwaggett: yeah, should we ask Nick about it or..?

Magisus: No, he probably won't know. @pcarlisle or @justinstoller might have feelings though.

mwaggett:
During this work, we discovered inconsistencies with how we make use of middleware wrappers in different services. I will open a follow-up PR to improve that.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
pcarlisle: I don't really know if anyone is sending anything else right now. I guess I'd lean to not worrying about it and continue to allow whatever but possibly put something in for the next major version.

PR #2378, opened 2020-07-31
(SERVER-2848) Branch twiddling

Magisus:
This PR updates:
* `build_defaults.yaml` to point to the right build and nightly repos
* Rakefile to ensure that the puppet and puppet-agent versions point to the right branches
* project version to 7.0.0


Magisus:
Two other things that will probably need to change:
* [min required agent version](https://github.com/puppetlabs/puppetserver/blob/master/resources/ext/ezbake.conf#L10) - we don't know what this should be yet, and likely don't have any incompatibilities yet, though @joshcooper might be able to provide more insight here.
* compatibility test suite - we should once again evaluate the utility of these in their preset form and see if it's worth continuing to expand them. Currently we have tests for all the way back to Puppet 3, though in the 6.x Jenkins pipeline we only run the Puppet 5 ones.

mwaggett:
@Magisus should we open tickets for those additional things if they won't be done right now?

Magisus:
I'm asking about the first one real quick, I'll file something for the second.

PR #2380, opened 2020-08-03
Environment classes.markdown

K8med:
Fixed a broken link per https://tickets.puppetlabs.com/browse/DOC-3892

### DIFF
--- a/documentation/puppet-api/v3/environment_classes.markdown
+++ b/documentation/puppet-api/v3/environment_classes.markdown
@@ -5,11 +5,9 @@ canonical: "/puppetserver/latest/puppet-api/v3/environment_classes.html"
 ---
 
 [classes]: https://puppet.com/docs/puppet/latest/lang_classes.html
-[deprecated WEBrick Puppet master]: https://puppet.com/docs/puppet/5.5/services_master_webrick.html
 [node definitions]: https://puppet.com/docs/puppet/latest/lang_node_definitions.html
 [defined types]: https://puppet.com/docs/puppet/latest/lang_defined_types.html
 [`environment_timeout`]: https://puppet.com/docs/puppet/latest/config_file_environment.html#environmenttimeout
-[resource type API]: https://puppet.com/docs/puppet/4.10/http_api/http_resource_type.html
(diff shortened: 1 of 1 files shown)


documentation/puppet-api/v3/environment_classes.markdown
```diff
@@ -5,11 +5,10 @@ canonical: "/puppetserver/latest/puppet-api/v3/environment_classes.html"
 ---
 
 [classes]: https://puppet.com/docs/puppet/latest/lang_classes.html
-[deprecated WEBrick Puppet master]: https://puppet.com/docs/puppet/latest/services_master_webrick.html
+[deprecated WEBrick Puppet master]: https://puppet.com/docs/puppet/5.5/services_master_webrick.html
```
Magisus, line 8: I don't see this link being used anywhere in the page...

mwaggett: ooh good call. +1 to removing this altogether

justinstoller: this should be updated like this in 5.5.x and then removed per this comment when merged up, right?

mwaggett: sure, I guess everything on our 5.3.x branch should be updated to point to `5.5` instead of `latest`.. But that seems separate to this ticket.

K8med: @mwaggett: You're right that pointing 5.3.x to `5.5` and not `latest` should be done and in a separate ticket. I've asked Doc and Heston is currently working on it as part of the larger architecture, but it may take a while. In the meantime, we're making a redirect for _latest/services_master_webrick_ mentions to _5.5./services_master_webrick_. Would you still like me to remove the _services_master_webrick_ link here?

mwaggett: @K8med yeah, I think we want to remove it altogether because nothing is actually linking to it, as far as I can tell. If I'm missing something, let me know!

K8med: @mwaggett I'll remove it!

PR #2381, opened 2020-08-04
(maint) Bump to clj-parent 4.6.0 + remove `ring.middleware.params` namespace

mwaggett:


project.clj
```diff
@@ -193,7 +193,6 @@
                              puppetlabs.puppetserver.ringutils
                              puppetlabs.puppetserver.certificate-authority
                              puppetlabs.services.ca.certificate-authority-core
-                             puppetlabs.puppetserver.ring.middleware.params
```
justinstoller, line 196: Was this always unnecessary or should we now be compiling the ring-middleware namespace?

mwaggett: This was the namespace I just removed, so it was previously necessary.
We aren't aot compiling any other ring-middleware namespaces, so I assumed we didn't need to do it, but I will admit I don't really know how it's decided which namespaces get aot compiled.

justinstoller: We usually have to explicitly declare namespaces to aot that have services defined in the tk `defservice` macro - and, I don't remember why, others.

We may have just added all the namespaces in puppetserver to the aot list rather than understanding if they all required being added (adding them doesn't hurt).

I tested this by running `lein uberjar` and it compiled fine. I then ran the jar with `java -jar target/puppet-server-release.jar -m puppetlabs.trapperkeeper.main -b dev/bootstrap.cfg -c dev/puppetserver.conf` and it started fine. I then ran an agent against it and it compiled a catalog without issue.

PR #2383, opened 2020-08-07
(maint) Update facter submodule to point at facter#4.x

Magisus:
Previously, the Facter 4 code lived in the facter-ng repo. Now that it
has moved to the main facter repo, this commit updates our submodule pin
to point there, to the 4.x branch.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,4 +9,4 @@
 	url = https://github.com/puppetlabs/puppet-resource_api.git
 [submodule "ruby/facter"]
 	path = ruby/facter
-	url = https://github.com/puppetlabs/facter-ng.git
+	url = https://github.com/puppetlabs/facter.git


mwaggett:
How does it know which branch to use?

Magisus:
It's pinned to a specific SHA. In this PR, I put it to the HEAD of the 4.x branch.

Magisus:
There's a ticket to add auto updating like we have for Puppet, but that's a "manual" process of fetching the new commit you want, checking it out in the submodule, and committing that.

mwaggett:
ah right, I always forget this one doesn't auto-update, thanks!

Magisus:
Something about this update seems to have legitimately broken FIPS testing. Looking into it.

Magisus:
The EC2 fact resolver started requiring `Net::HTTP`, which in turn requires `OpenSSL`. We need to figure out a way to avoid that. In the meantime, I'm going to pin us back to 4.0.26, before that resolver went in.

PR #2385, opened 2020-08-07
(SERVER-1689) Document short name auth.conf caveat

Magisus:
This commit adds a note to the docs that OID short names cannot be used
in auth.conf extension-based rules when the CA is disabled.

### DIFF
--- a/documentation/config_file_auth.markdown
+++ b/documentation/config_file_auth.markdown
@@ -7,6 +7,7 @@ canonical: "/puppetserver/latest/config_file_auth.html"
 [`trapperkeeper-authorization`]: https://github.com/puppetlabs/trapperkeeper-authorization
 [Puppet `auth.conf`]: https://puppet.com/docs/puppet/latest/config_file_auth.html
 [deprecated]: ./deprecated_features.markdown
+[short names]: https://puppet.com/docs/puppet/6.17/ssl_attributes_extensions.html#puppet_registered_ids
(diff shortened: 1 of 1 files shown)


documentation/config_file_auth.markdown
```diff
@@ -176,6 +177,9 @@ If a request matches the rule, Puppet Server checks the request's authenticated
     -   A `certname` key equivalent to a bare string.
 
     If the request's authenticated name matches the parameter's value, Puppet Server allows it.
+
+> **Note:** If you are using Puppet Server with the CA disabled, you must use OID values for the extensions. Puppet Server will not be able to resolve [short names][] in this mode.
```
mwaggett, line 181: Would it make more sense to link from `OID values for the extensions` rather than `short names`? I expect a link to take me to a page about whatever the link says, ya know?

Magisus: I guess I wanted to link "short names" because they're kind of a puppet-specific concept, where as OID values are common to OpenSSL and certs in general (in fact our docs link you out to a wikipedia page). There isn't an actual linkable section that talks about them that I could find, however. Do you think linking to the [table](https://puppet.com/docs/puppet/6.17/ssl_attributes_extensions.html#puppet_registered_ids) is better?

mwaggett: ah, that makes sense. yeah, I think the table is clearer for this.

PR #2388, opened 2020-08-19
(SERVER-2845) Serve file_content for bolt projects

pcarlisle:
This adds support to the v3/file_content endpoint for serving files out
of projects. This is used by passing the query parameter `project`
instead of `environment`. This is implemented entirely in clojure,
unlike environment file serving support which still uses ruby.


src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -873,6 +873,127 @@
...
+       first))
+
+(defn mount->path-component
+  "Map the fileserving \"mount\" to the subdirectory of modules it will serve
+  files out of."
+  [mount]
+  (case mount
+    "modules" "files"
+    ;; default to itself
+    mount))
```
pcarlisle, line 923: I probably need to be explicit about which are allowed, i think right now you could probably request out of any subdirectory because of this

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -873,6 +873,127 @@
...
+        (rr/content-type "application/octet-stream"))
+    {:status 404
+     :headers {"Content-Type" "text/plain"}
+     :body (i18n/tru "Could not find file_content for path: {0}" requested-file-path)}))
+
+(defn handle-project-file-content
+  "Handle a file_content request for a bolt project."
+  [bolt-project-dir request]
+  (let [project (get-in request [:params "project"])
+        version (get-in request [:params "version"])
```
mwaggett, line 959: Where does `version` come from?

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -873,6 +873,127 @@
...
+  [project-dir project-name]
+  (->> project-dir
+       io/file
+       .list
+       seq
+       (filter #(is-bolt-project? (str project-dir "/" %)))
+       (some #{project-name})))
+
+(defn get-project-root
+  [project-dir project-name]
```
mwaggett, line 892: can `project-dir` be `bolt-project-dir`, to match what's used below? or maybe something like `bolt-projects-dir` might be clearer, to distinguish from what you've been calling `project-root`? but really it's the consistency I care about.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -873,6 +873,127 @@
...
+       [[#"plugins|pluginfacts" :mount-point] [#".*" :file-path]] :pluginsync}])
+
+(defn make-file-content-response
+  "Given a File, generate an appropriate ring response map. Returns a 404
+  response if passed `nil`."
+  [file requested-file-path]
+  (if file
+    (-> file
+        str
+        rr/file-response
```
mwaggett, line 949: Is this what actually reads the file content? I was poking around and it didn't totally look like it? but I can't find where else that might be happening..?

src/clj/puppetlabs/services/master/master_service.clj
```diff
@@ -100,6 +100,7 @@
          certname (get-in config [:puppetserver :certname])
          localcacert (get-in config [:puppetserver :localcacert])
          puppet-version (get-in config [:puppetserver :puppet-version])
+         bolt-project-dir (get-in config [:bolt :project-dir])
```
mwaggett, line 103: ¯\_(ツ)_/¯ I like using plurals for things that will contain many of the thing, e.g. projects. so `:projects-dir`. I don't know if now is when we have to set these in stone, but figured I'd offer my opinion 😝

test/unit/puppetlabs/services/master/master_core_test.clj
```diff
@@ -439,3 +441,51 @@
...
+      (testing "regular project"
+        (is (= (str bolt-project-dir "/local")
+               (get-project-root bolt-project-dir "local"))))
+
+      (testing "embedded project (Boltdir)"
+        (is (= (str bolt-project-dir "/embedded/Boltdir")
+               (get-project-root bolt-project-dir "embedded")))))
+
+    (testing "finding a file in a project"
+      (testing "in the modules mount"
```
mwaggett, line 475: should we test the tasks mount too?

test/integration/puppetlabs/services/master/master_service_test.clj
```diff
@@ -766,3 +766,26 @@
 
             (is (= 503 status-code))
             (is (<= 0 retry-after 1800))))))))
+
+(deftest ^:integration project-file-content
```
mwaggett, line 770: Was there a plan to do any testing with the `named` project?
There seem to be a handful of unused test files (though maybe some are just necessary for structuring a project correctly/realistically)?

documentation/puppet-api/v3/file_content.markdown
```diff
@@ -0,0 +1,32 @@
...
+
+### Response
+
+A successful request to this endpoint returns an `HTTP 200` response code and
+`application/octet-stream` Content-Type header, and the contents of the specified file
+in the response body. An unsuccessful request returns an error response
+code with a `text/plain` Content-Type header:
+
+-   400: returned when neither the environment nor the project query parameters are provided, or when both project and environment are provided. 
+-   404: returned when requesting a file that is not within a module's `files` or `tasks`
```
mwaggett, line 30: Should we note the error responses for when the `project` isn't found or the file path is missing from the request?
I'm comparing to the existing docs for this endpoint: https://puppet.com/docs/puppet/6.17/http_api/http_file_content.html

src/clj/puppetlabs/services/master/master_core.clj
mwaggett: +1

src/clj/puppetlabs/services/master/master_core.clj
pcarlisle: ring automatically handles a File being returned in the body

mwaggett: Can you explain how it does that? I'm looking at https://github.com/ring-clojure/ring/blob/master/ring-core/src/ring/util/response.clj and not seeing what actually gets the content of the file..

src/clj/puppetlabs/services/master/master_core.clj
pcarlisle: Nevermind, this is already handled through the routing table and the error is returned in `handle-project-file-content`

test/integration/puppetlabs/services/master/master_service_test.clj
pcarlisle: Originally I had a question of whether we ever needed to identify a project by it's name from the yaml file when that's different from the directory name. I think that we should just avoid this though, since we're going to write how deployment works.

documentation/puppet-api/v3/file_content.markdown
pcarlisle: they're all 404 but i added something to this line to mention that

src/clj/puppetlabs/services/master/master_core.clj
pcarlisle: https://github.com/ring-clojure/ring/blob/master/ring-core/src/ring/core/protocols.clj#L37

src/clj/puppetlabs/services/master/master_core.clj
pcarlisle: leftover line from an alternate implementation where i treated the different components of the project name independently (which is still something we might choose to do later, not hard to change)

test/unit/puppetlabs/services/master/master_core_test.clj
pcarlisle: added this in the service test

pcarlisle:
addressed all feedback up to this point, added some more docstrings, removed a function that I didn't realize was unused and its tests, expanded the service test a bit

src/clj/puppetlabs/services/master/master_core.clj
mwaggett: oh god. thanks.

test/integration/puppetlabs/services/master/master_service_test.clj
mwaggett: hmm, I think it makes sense to avoid that here. but I'm wondering where in the workflow that needs to happen.. Do we allow a user to deploy a project using the name from the yaml file (as opposed the name of the directory/repo)? If so, will the front end handle that or do we need to handle that in code-manager? 

If we're not gonna deal with it here, I think we should remove the associated test files.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -873,6 +873,121 @@
...
+  "Lookup a project by name and return the path where its project files are
+  located as a string."
+  [projects-dir project-name]
+  (let [boltdir-path (str projects-dir "/" project-name "/Boltdir")]
+    (if (fs/exists? boltdir-path)
+      boltdir-path
+      (str projects-dir "/" project-name))))
+
+(defn dirs-in-project-modulepath
+  "List all directories in a bolt project's modulepath. Returns a sequence of
```
mwaggett, line 892: Is it really a sequence?? That's the one where order matters, right? I didn't think that's what was returned here..

test/unit/puppetlabs/services/master/master_core_test.clj
```diff
@@ -439,3 +441,33 @@
...
+      (testing "embedded project (Boltdir)"
+        (is (= (str bolt-projects-dir "/embedded/Boltdir")
+               (get-project-root bolt-projects-dir "embedded")))))
+
+    (testing "finding a file in a project"
+      (testing "in the modules mount"
+        (is (= (fs/file (str bolt-projects-dir "/local/modules/helpers/files/marco.sh" ))
+               (fs/file (find-project-file bolt-projects-dir "local" "modules" "helpers" "marco.sh"))))
+        (is (= (fs/file (str bolt-projects-dir "/local/site-modules/utilities/files/etc/greeting" ))
+               (fs/file (find-project-file bolt-projects-dir "local" "modules" "utilities" "etc/greeting")))))
```
mwaggett, line 461: probably should just compare strings here, instead of wrapping both sides in `fs/file`

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -873,6 +873,121 @@
          :headers {"Content-Type" "application/octet-stream"}
          :body (get-code-content environment code-id file-path)}))))
 
+(defn is-bolt-project?
+  [dir]
+  (or (fs/exists? (str dir "/bolt-project.yaml"))
```
donoghuc, line 878: I think the *only* configuration we can support is `bolt-project.yaml` and `Boltdir/bolt-project.yaml`. We should not care about bolt.yaml and in the case we have a Boltdir, we should make sure that there is a `Boltdir/bolt-project.yaml`.

PR #2390, opened 2020-08-20
(SERVER-2677) Extend sleep for unsigned certs

tvpartytonight:
When puppet is trying to manage the puppetserver service and starts
it, the shell command seemed to return very early, 1.37 seconds for
redhat 8 as opposed to ~18 for other platforms.. Beaker does a
check to http://localhost:8140 after the command, but it is only
ostensibly successful; this change extends the timeout to 20 seconds
to see if this allows the pre-suite step to complete successfully.

If this is successful, we should probably look into why redhat 8 is
completing the service start command so quickly.


mwaggett:
What do you mean when you say that the beaker check is "only ostensibly successful"? Is there something in beaker that should be fixed?

mwaggett:
oh wow, it literally says 'ostensibly successful'. I guess I'm just thinking that that doesn't seem like a very useful check at all, if it can't be trusted..

PR #2391, opened 2020-08-20
(SERVER-2859) Batch cert clean

Magisus:


Magisus:
In order to sign a new cert for the test fixtures, I ended having to regenerate them all, because they were expired and our signing machinery doesn't like that.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1278,16 +1278,22 @@
...
-                              (.getPublicKey (utils/pem->ca-cert cacert cakey))
-                              serial)]
-    (utils/crl->pem! new-crl cacrl)
-    (log/debug (i18n/trs "Revoked {0} certificate with serial {1}" subject serial))))
+   subjects :- [schema/Str]]
+  (let [original-crl (utils/pem->crl cacrl)
+        ca-private-key (utils/pem->private-key cakey)
+        ca-public-key (.getPublicKey (utils/pem->ca-cert cacert cakey))
+        new-crl (reduce
+                 (fn [crl subject]
```
justinstoller, line 1286: this makes a lot of sense, and I'm sure is a huge perf boost, but if we're not trying to get this in before the release we might want to look at creating a new `revokeMultiple` or something similar that takes a list of serials and loops over them calling `crl.addCRLEntry` each time so we aren't creating a new CRL for each serial. see https://github.com/puppetlabs/jvm-ssl-utils/blob/master/src/java/com/puppetlabs/ssl_utils/SSLUtils.java#L360

Magisus: Ah okay.

Magisus: I've opened https://github.com/puppetlabs/jvm-ssl-utils/pull/97 for this. However, that library is two major versions past what this 5.x branch has, but hasn't been branched yet. So we should talk about what is right there.

Magisus:
Updated to show usage of `ssl-utils` change.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1274,20 +1274,24 @@
     (fs/delete csr-path)
     (log/debug (i18n/trs "Removed certificate request for {0} at ''{1}''" subject csr-path))))
 
-(schema/defn revoke-existing-cert!
-  "Revoke the subject's certificate. Note this does not destroy the certificate.
-   The certificate will remain in the signed directory despite being revoked."
+(schema/defn revoke-existing-certs!
```
Magisus, line 1277: Do y'all think it's okay to only have the "multiple" version of this, or is the extra overhead from the loop machinery (both here and in the Java code) problematic enough to make it worth it to use a single version for the single-revoke call?

ezbake/config/conf.d/auth.conf
```diff
@@ -75,6 +75,21 @@ authorization: {
             sort-order: 500
             name: "puppetlabs cert statuses"
         },
+        {
+            # Allow the CA CLI to access the certificate clean endpoint
+            match-request: {
+                path: "/puppet-ca/v1/clean"
+                type: path
+                method: [put]
```
mwaggett, line 83: (nit) Based on other entries in here, it looks like single methods don't need to be wrapped in brackets.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1274,20 +1274,24 @@
...
-                              (utils/pem->private-key cakey)
-                              (.getPublicKey (utils/pem->ca-cert cacert cakey))
-                              serial)]
-    (utils/crl->pem! new-crl cacrl)
-    (log/debug (i18n/trs "Revoked {0} certificate with serial {1}" subject serial))))
+   subjects :- [schema/Str]]
+  (let [original-crl (utils/pem->crl cacrl)
+        ca-private-key (utils/pem->private-key cakey)
+        ca-public-key (.getPublicKey (utils/pem->ca-cert cacert cakey))
+        serials (reduce (fn [serials subject]
```
tvpartytonight, line 1285: Where does `subject` come from now?

ezbake/config/conf.d/auth.conf
```diff
@@ -75,6 +75,21 @@ authorization: {
...
+            }
+            allow: {
+               extensions: {
+                   pp_cli_auth: "true"
+               }
+            }
+            sort-order: 500
+            name: "puppetlabs cert status"
```
mwaggett, line 91: probably want to update this to something like "puppetlabs cert clean"

src/clj/puppetlabs/puppetserver/certificate_authority.clj
mwaggett: What extra overhead are you worried about? Are you just worried that it will now take longer to revoke a single cert? Is that something we could test?

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -315,7 +336,9 @@
         (DELETE [""] [subject]
           (handle-delete-certificate-request! subject ca-settings)))
       (GET ["/certificate_revocation_list/" :ignored-node-name] []
-        (handle-get-certificate-revocation-list ca-settings)))
+        (handle-get-certificate-revocation-list ca-settings))
+      (PUT ["/clean"] request
```
mwaggett, line 340: Would it work/make senes to do something like `[subject :as {body :body}]` like we do above, in the `certificate_request` endpoint? Mostly I'm just not seeing this `request` pattern elsewhere, so it's a little confusing to me whether that's a totally different thing or not.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
Magisus: From the `subjects` array passed as the third arg to `reduce`. It will split off elements of that list one at a time and pass them to this function.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -76,16 +76,37 @@
...
       (log/debug e))))
 
+(schema/defn handle-cert-clean
+  [{:keys [body]}
+   ca-settings :- ca/CaSettings]
+  (let [{:keys [certnames async]} (try-to-parse body)
+        [existing-certs missing-certs] (split-with
+                                        #(ca/certificate-exists? ca-settings %)
+                                        certnames)
+        message (when (seq missing-certs)
```
mwaggett, line 93: Why does this need to be wrapped in `seq`?

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
Magisus: I was having a hard time getting this to bind right, and didn't want to use "subject" because unlike the other endpoints here, this doesn't act on a single "subject" but rather a list of them, but we can't extract them until after we parse them out of the JSON. So I handle pulling apart the request body in the handler. But maybe I've done something wrong and there is a way to get the destructuring to work? For now, I've followed the pattern we use for other (non-CA) endpoints in puppetserver (see the routes in `master_core`) that process a body withouth going through liberator like the `certificate_status` endpoint does.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
Magisus: That's apparently preferred shorthand for `(not (empty?))` . See the note on [`empty?`](https://clojuredocs.org/clojure.core/empty_q).

mwaggett: oh, interesting.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
Magisus: Yeah mostly that. I imagine it's not meaningfully slower though.

Magisus: However, it does mean I had to wrap some args in arrays unnecessarily, which could be kind of annoying clutter?

mwaggett: Will that affect consumers or is it just internal clutter?

When I first read this PR, I did find it a little strange to get rid of the "single" version, but I think that's because I'm used to the multiple version just being a loop of the single version. Since this isn't that, it doesn't really seem like we need the single one?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
Magisus: I decided to branch that lib.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1274,20 +1274,24 @@
...
+                          (conj serials (-> (path-to-cert signeddir subject)
```
justinstoller, line 1286: not a big deal, but this is effectively a map.

Magisus: Oh you're right, thanks. Trying to get more fluent with these functional transformers.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -76,16 +76,37 @@
...
+                  (format "The following certs do not exist and cannot be revoked: %s"
+                          (vec missing-certs)))]
+    (try
+      (ca/revoke-existing-certs! ca-settings existing-certs)
+      (ca/delete-certificates! ca-settings existing-certs)
+      (-> (rr/response (or message "Successfully cleaned all certs."))
+          (rr/status 200)
+          (rr/content-type "text/plain"))
```
justinstoller, line 101: I was expecting a json response, but this matches the existing api well.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -76,16 +76,37 @@
...
   [body]
   (try
     (cheshire/parse-stream (io/reader body) true)
     (catch Exception e
       (log/debug e))))
 
+(schema/defn handle-cert-clean
+  [{:keys [body]}
+   ca-settings :- ca/CaSettings]
+  (let [{:keys [certnames async]} (try-to-parse body)
```
justinstoller, line 89: .... We probably want to validate that async, if passed, is not true.

Magisus: Okay.

justinstoller: Implicit there is that I would understand if we said, w/o the option we assume sync behavior and reserve the async field for the future if time scales require this to no longer be synchronous.

Magisus: That is how I was thinking about this, yes.

Magisus:
Updated responding to feedback, and rebased to pick up new clj-parent.

Magisus:
Oh good, Travis passed.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -76,16 +76,42 @@
...
+        [existing-certs missing-certs] (split-with
+                                        #(ca/certificate-exists? ca-settings %)
+                                        certnames)
+        message (when (seq missing-certs)
+                  (format "The following certs do not exist and cannot be revoked: %s"
+                          (vec missing-certs)))]
+    ;; TODO support async mode
+    (if (true? async)
```
mwaggett, line 97: Is it worth having this check right at the beginning, so we don't bother with checking cert existence if we're just gonna fail?

Magisus: I can try that.

mwaggett:
Since @justinstoller had a lot of comments, I think it would be good for him to take another look before merge.

Magisus:
Got this into my test env, might as well validate before merge.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -76,16 +76,42 @@
...
+  [{:keys [body]}
+   ca-settings :- ca/CaSettings]
+  (let [{:keys [certnames async]} (try-to-parse body)]
+    ;; TODO support async mode
+    (if (true? async)
+      (-> (rr/response "Async mode is not currently supported.")
+          (rr/status 400)
+          (rr/content-type "text/plain"))
+      (let [[existing-certs missing-certs] (split-with
+                                            #(ca/certificate-exists? ca-settings %)
```
Magisus, line 96: Oh this does not do what I think it does. Whoops.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -76,16 +76,43 @@
...
   [body]
   (try
     (cheshire/parse-stream (io/reader body) true)
     (catch Exception e
       (log/debug e))))
 
+(schema/defn handle-cert-clean
+  [{:keys [body]}
+   ca-settings :- ca/CaSettings]
+  (let [{:keys [certnames async]} (try-to-parse body)]
```
Magisus, line 89: I think we need some better error handling here. The exception from malformed JSON was swallowed and it printed a success message >.>

Magisus:
Added validation for bad JSON and missing certnames.

Magisus:
Should probably write some docs for this. Do we want those in a separate PR, so they don't get published right away?

mwaggett:
ah yeah, a separate PR is probably good, that's what I did for that other update I did

PR #2397, opened 2020-08-25
(PE-29443) Preserve hash order for plan variables

donoghuc:
This commit updates the compile_ast endpoint to support loading plan variables from an ordered list. Deserialization requires ordered hashes. Hash order is lost when moving from ruby -> clojure -> json so the ordered data is preserved in a list and then an ordered hash is created.


donoghuc:
WIP, i meant to open this as a draft.

mwaggett:
How has this been tested?

McdonaldSeanp:
Acceptance tests for https://github.com/puppetlabs/orchestrator/pull/1439 have been run with these updates.

PR #2400, opened 2020-08-26
Merge 5.3.x -> 6.x

Magisus:
This was complicated, please take a close look.

I also noticed that we lost a bit of debug logging with the batch revoke work. That might be worth figuring out how to put back in...


src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1494,36 +1494,39 @@
...
         [our-full-crl & rest-of-full-chain] (utils/pem->crls cacrl)
-        new-full-crl (utils/revoke our-full-crl
+        new-full-crl (utils/revoke-multiple our-full-crl
                                    (utils/pem->private-key cakey)
                                    (.getPublicKey (utils/pem->ca-cert cacert cakey))
-                                   serial)
+                                   serials)
         new-full-chain (cons new-full-crl (vec rest-of-full-chain))]
     (write-crls new-full-chain cacrl)
-    (log/debug (i18n/trs "Revoked {0} certificate with serial {1}" subject serial))
```
mwaggett, line 1513: hmm yeah, would be good to get this back somehow. the options I see are:
* log here, but just log everything at once, e.g. "Revoked certificates <all subjects> with serials <all serials>"
* log at the java level, where we access the certs/serials one at a time

the latter feels more "correct", but more annoying, since we'd need a whole new set of releases

locales/puppetserver.pot
```diff
@@ -251,14 +251,6 @@ msgstr ""
...
 msgstr ""
 
-#: src/clj/puppetlabs/puppetserver/certificate_authority.clj
-msgid "Revoked {0} certificate with serial {1}"
-msgstr ""
-
-#: src/clj/puppetlabs/puppetserver/certificate_authority.clj
-msgid "Infra node certificate being revoked; publishing updated infra CRL"
-msgstr ""
-
```
mwaggett, line 261: why was this one removed? looks like that message is still there?

Magisus: Bad conflict resolution, I'll put it back. Though it may not be exactly right anymore...

src/clj/puppetlabs/puppetserver/certificate_authority.clj
Magisus: Agreed... yeah it would be a set of releases for both branches, dang. Hm.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1494,36 +1494,39 @@
...
 
     ;; Publish infra-crl if an infra node is getting revoked.
     (when (and enable-infra-crl
-               (fs/exists? infra-node-serials-path)
-               (seq-contains? (read-infra-nodes infra-node-serials-path) (str serial)))
-      (let [[our-infra-crl & rest-of-infra-chain] (utils/pem->crls infra-crl-path)
-            new-infra-crl (utils/revoke our-infra-crl
+               (fs/exists? infra-node-serials-path))
+      (let [infra-revocations (filter #(seq-contains?
```
justinstoller, line 1518: few of things.

We shouldn't put the `read-infra-nodes` - which reaches out to the file system and (should) always return the same data during this loop - in the fn that gets called for every serial to revoke. I'd pull that into its own binding above.

I would also probably rewrite this to use `set`, and `set/intersection` instead of `filter` & `contains`.

To keep with the original conditional above we should test that `infra-revocations` is not empty before continuing.

Magisus: Thanks I'll clean that up.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
justinstoller: We could keep it fairly similar with:
```clj
 (log/debug (i18n/trs "Revoked {0} certificates(s)." (count serials)))
```
The the latter that molly was saying at the trace level?

Might want to ask CS what level / info would be most helpful?

Magisus: Okay. For now I'll add that debug in here.

locales/puppetserver.pot
Magisus: This should get regenerated in Jenkins with the new messages.

Magisus:
Might be good to add a test that revokes multiple infra CRLs... I'll see if I can make that happen without regenerating everything... >.>

Magisus:
Added that test.

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
```diff
@@ -482,6 +482,70 @@
...
+        infra-inventory-content (slurp infra-inventory-path)
+        cert1-path (ca/path-to-cert (str master-conf-dir "/ssl/ca/signed") subject1)
+        cert1 (utils/pem->cert cert1-path)
+        cert2-path (ca/path-to-cert (str master-conf-dir "/ssl/ca/signed") subject2)
+        cert2 (utils/pem->cert cert2-path)]
+    ;; Add another cert to the infra inventory
+    (spit infra-inventory-path (str infra-inventory-content "agent-node"))
+    (bootstrap/with-puppetserver-running-with-mock-jrubies
+     "JRuby mocking is safe here because all of the requests are to the CA
+     endpots, which are implemented in Clojure."
```
mwaggett, line 508: lol
```suggestion
     endpoints, which are implemented in Clojure."
```

Magisus: Turns out this was all over the place due to copy-pasta.

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
```diff
@@ -482,6 +482,70 @@
...
+        infra-serials-path (str master-conf-dir "/ssl/ca/infra_serials")
+        infra-serials-content (slurp infra-serials-path)
+        infra-inventory-path (str master-conf-dir "/ssl/ca/infra_inventory.txt")
+        infra-inventory-content (slurp infra-inventory-path)
+        cert1-path (ca/path-to-cert (str master-conf-dir "/ssl/ca/signed") subject1)
+        cert1 (utils/pem->cert cert1-path)
+        cert2-path (ca/path-to-cert (str master-conf-dir "/ssl/ca/signed") subject2)
+        cert2 (utils/pem->cert cert2-path)]
+    ;; Add another cert to the infra inventory
+    (spit infra-inventory-path (str infra-inventory-content "agent-node"))
```
mwaggett, line 505:
```suggestion
    (spit infra-inventory-path (str infra-inventory-content subject2))
```

PR #2403, opened 2020-08-28
Merge 5.3.x -> 6.x

Magisus:


justinstoller:
This doesn't take the version from 5.3.x of 30144b8  which honestly I think is better than the version in 6.x. 

https://github.com/puppetlabs/puppetserver/blob/6.x/src/clj/puppetlabs/puppetserver/certificate_authority.clj#L1516

Magisus:
Oh yeah, good point. I'll change that.

justinstoller:
I was expecting 30144b8 to create a merge conflict and that the conflict could be resolved manually by taking the changes rather than discarding them vs adding a new commit on top?

Magisus:
Okay, I can do it that way. It got all merged into one block with the infra-crl changes. But I can fake it to make it look right.

justinstoller:
Sorry, I didn't mean to make you do a bunch more work, I was just expecting something else and was curious if I was expecting wasn't doable.

Magisus:
No biggie, easy fix :)

PR #2409, opened 2020-09-03
(SERVER-2875) Serve file_content from top level of projects

pcarlisle:
Bolt allows accessing the top level of a project by referring to a module of the
same name. This ticket supports that for everything we've added so far for
file_content serving from projects.


Magisus (approved):

donoghuc (changes requested):
This is not quite right. We need the _project name_ to be the module name. This implementation makes the module name the concatenation of project name and the ref. 

for example if my bolt-project.yaml has: `name: my_project` and the ref on disk is `my_project_someref` i would expect the task `tasks/foo.sh` to be served from:

```
[root@bereft-start ~]# curl --silent -v --cacert $BOLT_CACERT --cert $BOLT_CERT --key $BOLT_KEY $PUPPETSERVER_ROOT/puppet/v3/file_content/tasks/my_project/foo.sh?project=my_project_somesha 2>/dev/null
Could not find file_content for path: foo.sh
```
and not
```
[root@bereft-start ~]# curl --silent -v --cacert $BOLT_CACERT --cert $BOLT_CERT --key $BOLT_KEY $PUPPETSERVER_ROOT/puppet/v3/file_content/tasks/my_project_somesha/foo.sh?project=my_project_somesha 2>/dev/null
#!/bin/bash

echo 'hi'
```

I'm also concerned that this might not handle multifile task implementations. For example if within a task metadata file i add dependency files in top level `lib` or `files` in the project, I'm not sure i see how this would support that.

donoghuc:
So it does look like things under `files` (and presumably `lib`) are served with this implementation (though there is still the bug where we would expect the inserted module name to be the project name (not projectname_ref)
```
[root@bereft-start my_project_somesha]# tree
.
├── bolt-project.yaml
├── files
│   └── foo.data
├── Puppetfile
└── tasks
    ├── foo.json
    └── foo.rb

2 directories, 5 files
[root@bereft-start my_project_somesha]# cat files/foo.data 
hi
[root@bereft-start my_project_somesha]# cat tasks/foo.*
{ "files": ["my_project/files/foo.data"] }
#!/opt/puppetlabs/puppet/bin/ruby
puts File.read(File.join(ENV['PT__installdir'],'/my_project/files/foo.data'))
[root@bereft-start my_project_somesha]# curl --silent -v --cacert $BOLT_CACERT --cert $BOLT_CERT --key $BOLT_KEY $PUPPETSERVER_ROOT/puppet/v3/file_content/modules/my_project_somesha/foo.data?project=my_project_somesha 2>/dev/null
hi
```

donoghuc (changes requested):
How does this handle `modulepath` defined in `bolt-project.yaml`? it almost looks like it is hard coded to just serve things from `site`, `modules`, `site-modules`. I think we talked about allowing you to configure a modulepath **relative to bolt project dir**. So for:
```
[root@bereft-start my_project_somesha]# tree
.
├── bolt-project.yaml
├── files
│   └── foo.data
├── foo
│   └── my_module
│       └── tasks
│           └── foo.sh
├── Puppetfile
└── tasks
    ├── foo.json
    └── foo.rb
5 directories, 6 files
```
and 
```
[root@bereft-start my_project_somesha]# cat bolt-project.yaml 
name: my_project
tasks:
  - my_project::foo
modulepath: foo
```
The tasks that should be served are: `puppet/v3/file_content/tasks/my_project/foo.rb` and `puppet/v3/file_content/tasks/my_module/foo.rb`

It almost seems like we need to read bolt-project.yaml and extract the `name` (for magically prepending top level content with module name) and `modulepath` (for figuring out which directories modules may come from).

pcarlisle:
> How does this handle `modulepath` defined in `bolt-project.yaml`? it almost looks like it is hard coded to just serve things from `site`, `modules`, `site-modules`. I think we talked about allowing you to configure a modulepath **relative to bolt project dir**.

I took this from bolt directly. I don't see anything in bolt about defining a modulepath in the yaml file, though of course I'm happy to add that if it's something we're doing.

I am considering splitting the ref into separate query parameters just to have more flexbility anyway, which will make having the name easier. Though it's not clear to me at the moment how much we need some kind of namespace as a parameter, and if so what that's called.

Magisus:
It looks like you can [define a modulepath](https://puppet.com/docs/bolt/latest/bolt_project_reference.html#modulepath) in a bolt project config, but I'm also fine leaving that for a separate ticket/PR.

pcarlisle:
Okay, I now see how other config options are loaded from a project config. I didn't understand the first time I read it since it's outside of the project-specific code.

test/integration/puppetlabs/services/master/master_service_test.clj
```diff
@@ -777,38 +777,56 @@
...
-      (let [response (http-get "/puppet/v3/file_content/modules/utilities/etc/greeting?project=local")]
+      (let [response (http-get "/puppet/v3/file_content/modules/utilities/etc/greeting?project=local-23")]
         (is (= 200 (:status response)))
         (is (= "Good morning\n" (:body response)))
         (is (= "13" (get-in response [:headers "content-length"])))
         (is (= "application/octet-stream" (get-in response [:headers "content-type"])))))
 
     (testing "can retrieve file_content from the tasks mount from a project"
-      (let [response (http-get "/puppet/v3/file_content/tasks/utilities/blah?project=local")]
+      (let [response (http-get "/puppet/v3/file_content/tasks/utilities/blah?project=local-23")]
```
Magisus, line 787: I think we're using an underscore separator now.

test/integration/puppetlabs/services/master/master_service_test.clj
```diff
@@ -777,38 +777,56 @@
                     :master-conf-dir master-service-test-runtime-dir}}
 
     (testing "can retrieve file_content from the modules mount from a project"
-      (let [response (http-get "/puppet/v3/file_content/modules/utilities/etc/greeting?project=local")]
+      (let [response (http-get "/puppet/v3/file_content/modules/utilities/etc/greeting?project=local-23")]
```
Magisus, line 780: I think we're using an underscore separator now. Doesn't really matter since you're using fixtures, but it might be good to have it reflect how things will really look, so as not to confuse ourselves :)

Magisus (commented):
Looks like this does what was asked for so far. Did you say there was other work coming to support a `.modules` folder, or is that separate?

test/integration/puppetlabs/services/master/master_service_test.clj
pcarlisle: :disappointed:

pcarlisle:
Okay I've added the new modulepath stuff as I've understood it

test/integration/puppetlabs/services/master/master_service_test.clj
Magisus: Thanks. Sorry about that. Still think it was a little unnecessary that were asked to change that in the first place, but oh well.

Magisus (approved):

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -916,13 +926,50 @@
...
+
+(defn get-project-modulepath
+  "Given a map representing the bolt project configuration, return the
+  modulepath. If the modulepath is not defined or the configuration map passed
+  is nil then return the default modulepath."
+  [project-config]
+  (let [has-modules-key? (contains? project-config :modules)]
+    (if-let [modulepath (parse-modulepath (get project-config :modulepath))]
+      (if has-modules-key?
+        (concat modulepath [".modules"])
```
mwaggett, line 954: There's a lot I didn't understand about Alex's email, so I might be wrong, but I thought ".modules" gets appended to the specified module path, so shouldn't this be something like `(map #(str % ".modules") modulepath)`?
Or an alternative interpretation.. `.modules` is the default modulepath, so we should put that first instead of last (`(concat [".modules"] modulepath)`)?

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -916,13 +926,50 @@
...
-  (if (is-bolt-project? (str bolt-projects-dir "/" project-name))
-    (let [project-root (get-project-root bolt-projects-dir project-name)
-          module-root (find-project-module project-root module)
+  Returns the path as as string. If the module name is the same as the project
+  name then files are served from the project directly."
+  [bolt-projects-dir project-ref mount module path]
+  (if (is-bolt-project? (str bolt-projects-dir "/" project-ref))
+    (let [project-root (get-project-root bolt-projects-dir project-ref)
+          project-config (read-bolt-project-config project-root)
+          project-name (get project-config :name)
```
mwaggett, line 968: Is this the only way a project's name is defined? What if `project-config` is nil?

pcarlisle: Bolt documents that in order to load top level files as a module then you have to define this name explicitly in the project config. [https://github.com/puppetlabs/bolt/issues/1871]

If project-config is nil then project-name will be nil.

src/clj/puppetlabs/services/master/master_core.clj
pcarlisle: I think the way I did it is right, based on reading the code in bolt.

pcarlisle: puppetlabs/bolt/pull/2189

src/clj/puppetlabs/services/master/master_core.clj
mwaggett: okay, I was wondering if we needed to fall back to the name from `project-ref` or something, but if Bolt requires a `:name` to do load top-level files, then sounds like we're good. 👍🏼

mwaggett (approved):

dev-resources/puppetlabs/services/master/master_core_test/bolt_projects/local_afafaf/tasks/init.sh
```diff
@@ -0,0 +1,2 @@
+. $PT__installdir/helpers/files/marco.sh
```
donoghuc, line 1: Does this task actually run without a shebang? (not that is needs to...)

Magisus:
@donoghuc did you have any other comments on this? I'd rather not merge past a "request changes" review, so if not, would you mind updating that?

donoghuc:
@Magisus I'm currently trying to make this work. I am having issues... Still working on review. This is my top priority today.

Magisus:
Oh, great, thanks!

dev-resources/puppetlabs/services/master/master_core_test/bolt_projects/local_afafaf/tasks/init.sh
pcarlisle: no idea, i just copied these out of bolt's tests

donoghuc (commented):
This is looking good! My biggest concern now is that my implementation in bolt-server for constructing URIs for task lib code does not appear to be working with this implementation. 

For example, with the following project:
```
[root@wily-censorship my_project_someref]# tree
.
├── bolt-project.yaml
├── files
│   └── source_me.sh
├── lib
│   └── source_me.sh
├── Puppetfile
└── tasks
    ├── init.json
    └── init.sh

3 directories, 6 files
[root@wily-censorship my_project_someref]# cat tasks/init.json 
{
  "files": ["my_project/files/source_me.sh", "my_project/lib/source_me.sh"]
}
[root@wily-censorship my_project_someref]# cat tasks/init.sh 
#!/bin/bash

source "$PT__installdir/my_project/files/source_me.sh"
source "$PT__installdir/my_project/lib/source_me.sh"
```
I ask orchestrator (which in turn asks bolt) for information about the task (most notably where to download the content from puppetserver)
```
[root@wily-censorship my_project_someref]# curl -k -X GET -H X-Authentication:$(cat ~/.puppetlabs/token) "https://wily-censorship.delivery.puppetlabs.net:8143/orchestrator/v1/project_tasks/my_project/init?project_ref=my_project_someref"
{
  "id" : "https://wily-censorship.delivery.puppetlabs.net:8143/orchestrator/v1/project_tasks/my_project/init",
  "name" : "my_project",
  "permitted" : true,
  "metadata" : {
    "files" : [ "my_project/files/source_me.sh", "my_project/lib/source_me.sh" ]
  },
  "files" : [ {
    "filename" : "init.sh",
    "sha256" : "f96decaf77eb1b1c8ce86b79e008fe72e337f4c5546c2f2131d1ba7473da0abe",
    "size_bytes" : 121,
    "uri" : {
      "path" : "/puppet/v3/file_content/tasks/my_project/init.sh",
      "params" : {
        "project" : "my_project_someref"
      }
    }
  }, {
    "filename" : "my_project/files/source_me.sh",
    "sha256" : "80384ff3aedc2e9bf0985b2c5c6163807d06f7aa6295fda04b4bc0c77f76813f",
    "size_bytes" : 35,
    "uri" : {
      "path" : "/puppet/v3/file_content/modules/my_project/source_me.sh",
      "params" : {
        "project" : "my_project_someref"
      }
    }
  }, {
    "filename" : "my_project/lib/source_me.sh",
    "sha256" : "03237c703bf64377ed9fb3cf9da5ad6d82930ff999d97dc5aee6e59b4036afc9",
    "size_bytes" : 32,
    "uri" : {
      "path" : "/puppet/v3/file_content/plugins/source_me.sh",
      "params" : {
        "project" : "my_project_someref"
      }
    }
  } ],
  "project_ref" : {
    "name" : "my_project_someref"
  }
}
```
As you can see according to this algorithm https://github.com/puppetlabs/bolt/blob/72c2a659bec8294a419dd6e358fd62575887f1f7/lib/bolt_server/transport_app.rb#L249-L270 we assume that files under `lib` will go under the plugins mount (this is how it appears to work for environments). The content under `tasks` and `files` appears to be available but not the lib code. 
```
[root@wily-censorship my_project_someref]#curl --silent -v --cacert $BOLT_CACERT --cert $BOLT_CERT --key $BOLT_KEY $PUPPETSERVER_ROOT/puppet/v3/file_content/tasks/my_project/init.sh?project=my_project_someref 2>/dev/null
#!/bin/bash

source "$PT__installdir/my_project/files/source_me.sh"
source "$PT__installdir/my_project/lib/source_me.sh"
[root@wily-censorship my_project_someref]# curl --silent -v --cacert $BOLT_CACERT --cert $BOLT_CERT --key $BOLT_KEY $PUPPETSERVER_ROOT/puppet/v3/file_content/modules/my_project/source_me.sh?project=my_project_someref 2>/dev/null
#!/bin/bash

echo 'hi from source'
[root@wily-censorship my_project_someref]# curl --silent -v --cacert $BOLT_CACERT --cert $BOLT_CERT --key $BOLT_KEY $PUPPETSERVER_ROOT/puppet/v3/file_content/plugins/source_me.sh?project=my_project_someref 2>/dev/null
Unsupported mount: plugins
```
We get that `Unsupported mount: plugins` error. 

Is there somewhere else we should train bolt-server to look for task library code?

donoghuc (approved):
Lib code being served under plugins mount will likely land as part of https://tickets.puppetlabs.com/browse/SERVER-2876 which addresses my last concern.

mwaggett:
@pcarlisle any squashing you wanna do before this gets merged?

pcarlisle:
No, I think it's fine.

Magisus (approved):

PR #2410, opened 2020-09-03
(PUP-10386) Remove tk-auth rule for puppet status endpoint

Magisus:
The `/puppet/v3/status` endpoint and associated indirection were
recently removed. This commit removes the associated tk-auth rule and
test for the endpoint.

### DIFF
--- a/acceptance/suites/tests/authorization/default_rules.rb
+++ b/acceptance/suites/tests/authorization/default_rules.rb
@@ -137,11 +137,6 @@ with_puppet_running_on(master, {}) do
     assert_denied(/\/puppet\/v3\/file_bucket_file\/md5\/123 \(method :get\)/)
   end
 
-  step 'status endpoint' do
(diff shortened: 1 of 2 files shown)


joshcooper (approved):

justinstoller:
Is this comment incorrect then? https://github.com/puppetlabs/puppet/pull/8303#issuecomment-685330087 eg the api endpoint is no longer functional without the indirection and rest termini?

Magisus:
I guess it must be. The tests failed with `{"message":"Not Found: Could not find indirection 'status'","issue_kind":"HANDLER_NOT_FOUND"}`.

joshcooper:
I'll comment on the other PR, but the indirected based status endpoint went away so we should update the API docs. He may have confused it with the simple status endpoint?

mwaggett (approved):

PR #2412, opened 2020-09-04
(maint) Test for 403 on locked down catalog endpoint

Magisus:
This commit updates the test that verifies whether extensions can be
used to prevent acces via tk-auth to use the correct assertions. The
master cert does not have the extension and so cannot use the catalog
endpoint, which requires it.


Magisus:
This makes the test pass for me locally. Curious to see if it'll fail in Jenkins, given that is passes right now.

justinstoller (approved):

PR #2416, opened 2020-09-10
(SERVER-2778) Remove references to legacy auth classes

joshcooper:
The `use-legacy-auth-conf` setting was removed in f2b6a1b57001, but puppetserver
still required and accessed puppet's deprecated auth classes. In order to remove
those classes from puppet, we first have to delete these references.

### DIFF
--- a/src/ruby/puppetserver-lib/puppet/server/config.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/config.rb
@@ -6,8 +6,6 @@ require 'puppet/environments'
 require 'puppet/server/logger'
 require 'puppet/server/jvm_profiler'
 require 'puppet/server/http_client'
-require 'puppet/server/auth_config_loader'
-require 'puppet/server/auth_provider'
 require 'puppet/server/execution'
 require 'puppet/server/environments/cached'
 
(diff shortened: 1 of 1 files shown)


mwaggett (approved):
looks fine to me, but I don't have much context on this, so might be good to get another set of eyes

src/ruby/puppetserver-lib/puppet/server/config.rb
```diff
@@ -6,8 +6,6 @@
 require 'puppet/server/logger'
 require 'puppet/server/jvm_profiler'
 require 'puppet/server/http_client'
-require 'puppet/server/auth_config_loader'
-require 'puppet/server/auth_provider'
```
joshcooper, line 10: oh jeez, I need more coffee or less wildfires/smoke... I removed the wrong require. Let me fix that.

joshcooper:
There's a bit of a catch-22 here. PR https://github.com/puppetlabs/puppet/pull/8320 deletes `Puppet::Network::Authorization` and `Puppet::Network::AuthConfig`. However, the last promoted puppetserver will try to register its auth provider:

https://github.com/puppetlabs/puppetserver/blob/1bb5001b02d09d21e084e4789d22989d1a1c60d7/src/ruby/puppetserver-lib/puppet/server/config.rb#L32-L35

And fail since those methods no longer exist. And I can't delete those lines from puppetserver, because the last promoted puppet-agent try to use legacy auth as it normally does.

So I'm going to update the puppet PR to remove its legacy auth implementation, but preserve the setters. Once that passes CI, then I'll PR puppetserver to remove the calls to the setters, and delete puppetserver's `AuthConfigLoader` and `AuthProvider` classes.

That may take a bit, so going to close this.

PR #2417, opened 2020-09-10
(maint) Don't set `use-legacy-auth-conf` to true

mwaggett:
This commit removes the teardown step to set `use-legacy-auth-conf` back to
true. We no longer use the legacy auth, but this setting was bleeding into other
tests. This caused test failures because legacy auth doesn't have a way to
restrict access based on cert extensions.


joshcooper (approved):

joshcooper:
Is there a way to PR test this? Or just want to merge?

mwaggett:
I'm not aware of any PR testing for the acceptance tests. I could run all the acceptance tests locally, but I think it's easier to just merge.

PR #2419, opened 2020-09-23
(SERVER-2780) New HTTP client API

Magisus:
This is the work to update the Ruby wrapper for our apache HTTP client to match the new client API in Puppet.


Magisus:
So one thing this doesn't have is a [`create_session`](https://github.com/puppetlabs/puppet/blob/main/lib/puppet/http/client.rb#L51-L53) function. None of the tests so far have actually needed it, but maybe they will?

src/ruby/puppetserver-lib/puppet/server/http_response.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  # Yield each header name and value. Returns an enumerator if no block is given.
+  def each_header
+    if block_given?
+      @java_response.get_headers.each do |name, value|
+        yield(name, value)
+      end
+    else
+      @java_response.get_headers.each
+    end
```
Magisus, line 44: Is there a more concise way to do this? Little rusty on my Ruby block handling.

justinstoller: I think the `&block` param is always considered optional and you can then pass it to `each` and it will do what you want, ie:
```rb
def foo(&block)
   {a: 1, b: 2, c: 3}.each(&block)
end

foo
=> #<Enumerator: {:a=>1, :b=>2, :c=>3}:each>

foo {|k,v| puts v }
1
2
3
```

src/ruby/puppetserver-lib/puppet/server/http_client.rb
```diff
@@ -96,7 +90,11 @@ def create_common_request_options(url, headers, options)
 
     # Ensure multiple requests are not made on the same connection
     headers["Connection"] = "close"
-    request_options = RequestOptions.new(build_url(url))
+
+    # Java can't handle the Ruby URI type
+    url = url.to_s if url.is_a?(URI)
+    # TODO Add params to this URL
```
justinstoller, line 96: ??

Magisus: I'm not actually sure what we're supposed to do with the query params now, I need to look into it a little more. The old API didn't have a separate arg for them, so I assumed they used to be just part of the URL if you supplied them. However, I also suspect that nothing that uses this client actually does supply them, so it might coincidentally not matter right now...

src/ruby/puppetserver-lib/puppet/server/http_response.rb
Magisus: Oh great, thanks!

Magisus:
The md5 failures were there before this PR, I think due to a recent change in Puppet. I'll put up something else to deal with them.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
justinstoller: It would be nice to clarify that.

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -176,59 +173,66 @@ jruby-config :- jruby-schemas/JRubyConfig
...
+
+          (testing "invalid compression format"
+            (let [compress "{ :compress => :bunk }"]
+              (is (= "Unsupported compression specified for request: bunk"
+                     (.runScriptlet
+                      sc
+                      (str "begin;"
+                           "  $response = $c.post('"
+                           url
+                           "', 'foo', options: "
```
mwaggett, line 229: why put this on three separate lines now?

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -262,59 +266,64 @@ jruby-config :- jruby-schemas/JRubyConfig
...
-          (with-http-client sc 10080 {:use-ssl true}
-            (.runScriptlet sc "$response = $c.get('/', {})")
-            (is (= "200" (.runScriptlet sc "$response.code")))
-            (is (= "hi" (.runScriptlet sc "$response.body"))))))))
+    (let [url (str "https://localhost:10080")]
+      (testing "Can connect via TLSv1 by default"
+        (with-webserver-with-protocols ["TLSv1"] ["TLS_RSA_WITH_AES_128_CBC_SHA"]
+          (with-scripting-container sc
+            (with-http-client sc {}
+              (let [url (str "https://localhost:10080")]
```
mwaggett, line 287: looks like you've defined `url` both here and above

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -262,59 +266,64 @@ jruby-config :- jruby-schemas/JRubyConfig
...
     (logutils/with-test-logging
       (jetty9/with-test-webserver ring-app port
         (with-scripting-container sc
-          (with-http-client sc port {:use-ssl false}
+          (with-http-client sc {}
             (let [client1 (.runScriptlet sc "$c.class.client")
-                  client2 (.runScriptlet sc (str (format "$c2 = Puppet::Server::HttpClient.new('localhost', %d, {:use_ssl => %s});"
-                                                         port false)
+                  client2 (.runScriptlet sc (str (format "$c2 = Puppet::Server::HttpClient.new;"
+                                                         false)
```
mwaggett, line 326: I don't think this `false` is needed anymore. Looks like it was the `:use_ssl` value.

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -323,60 +332,63 @@ jruby-config :- jruby-schemas/JRubyConfig
...
-                                    "$response = $c.get('/', {})"))
-              (is (= "200" (.runScriptlet sc "$response.code")))
-              (is (= "bye" (.runScriptlet sc "$response.body")))))))))
+            (with-http-client sc {}
+              (let [url (str "https://localhost:10080")]
+                (.runScriptlet sc (format "$response = $c.get('%s')" url))
+                (is (= 200 (.runScriptlet sc "$response.code")))
+                (is (= "hi" (.runScriptlet sc "$response.body")))
+                (.runScriptlet sc (str (format "$c = Puppet::Server::HttpClient.new;"
+                                               false)
```
mwaggett, line 376: same re: not needing `false`

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -176,59 +173,66 @@ jruby-config :- jruby-schemas/JRubyConfig
 (deftest test-ruby-http-client
   (jetty9/with-test-webserver ring-app port
```
mwaggett, line 174: Where doest the value for the `port` come from?

mwaggett (commented):
I noticed a lot of removals around `use_ssl`. How does this get handled now?

src/ruby/puppetserver-lib/puppet/server/http_response.rb
```diff
@@ -0,0 +1,52 @@
...
+    @java_response.get_reason_phrase
+  end
+
+  def body
+    @java_response.get_body
+  end
+
+  def read_body
+    yield body
+  end
```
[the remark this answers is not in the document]
Magisus: It is not used right now, but is a placeholder for an API for streaming the response body. In a real implementation, it would return chunks of the body one by one, ferrying them from the server to the client. But I talked to Josh and we decided that implementation is out of scope right now, given that the [clojure client](https://github.com/puppetlabs/clj-http-client) we're wrapping doesn't currently implement streaming. It's possible, the underlying java library supports it, just more work than we need right now, given nothing in Puppet Server uses it at the moment.

src/ruby/puppetserver-lib/puppet/server/http_response.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  # The HTTP standard counts all codes 200-299 as successes
+  def success?
+    code > 199 && code < 300
+  end
+
+  def [](name)
```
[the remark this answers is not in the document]
Magisus: This is required to match the API of Puppet's [response object](https://github.com/puppetlabs/puppet/blob/main/lib/puppet/http/response.rb), which I think in turn mirrors the API of the underlying Net::HTTP response implementation.

src/ruby/puppetserver-lib/puppet/server/http_response.rb
```diff
@@ -0,0 +1,52 @@
...
+      @java_response.get_headers.each do |name, value|
+        yield(name, value)
+      end
+    else
+      @java_response.get_headers.each
+    end
+  end
+
+  # Drain the response body
+  def drain
```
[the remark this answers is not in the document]
Magisus: It appears to be used when handling redirects, see this [commit message](https://github.com/puppetlabs/puppet/commit/deafcdec7463c900a26de1b322660635007c1cf5).

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
Magisus: It gets filled and then yielded to the body by that [macro](https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/master/test/clj/puppetlabs/trapperkeeper/testutils/webserver.clj#L15-L16).

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
Magisus: Before `url` wasn't interpolated. This can probably be refactored into a `format` call like the others.

Magisus:
So previously, the value of `use_ssl` was only used to determine whether the URL -- which got constructed from the server and port stored by the class -- should start with `http` or `https`. Now that you have to provide the whole URL, including the protocol, for each request, it was needed as an option on the client itself.

src/ruby/puppetserver-lib/puppet/server/http_response.rb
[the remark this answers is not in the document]
mwaggett: so if this were to be called right now, what would it do? Is a block required or optional (the [puppet implementation](https://github.com/puppetlabs/puppet/blob/main/lib/puppet/http/response.rb#L75) does a check)? I guess I never really got my head wrapped around `yield`..

src/ruby/puppetserver-lib/puppet/server/http_response.rb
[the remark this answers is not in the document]
mwaggett: I guess I still just don't know what is meant by `drain`.. And I don't really understand what it does, other than return true??

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
mwaggett: Is [this](https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/master/test/clj/puppetlabs/trapperkeeper/testutils/webserver.clj#L34) what gets the actual value? Where is it getting the value from? I still find macros extremely confusing..

src/ruby/puppetserver-lib/puppet/server/http_response.rb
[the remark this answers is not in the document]
Magisus: Yeah a block is required. Ruby will error if you call this without it now, when it tries to yield. I can give it a more explicit error like we do in Puppet if we think that's better.

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
Magisus: Yes.

src/ruby/puppetserver-lib/puppet/server/http_response.rb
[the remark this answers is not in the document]
Magisus: I'd recommend asking @joshcooper for details. But speculation: from what I can gather doing some cursory research is that sometimes, if you don't explicitly read the body from a response, it doesn't get pulled out of the connection state (this _might_ only be an issue if you asked for the body to be streamed, but I'm not sure). So if you then reuse the connection (e.g. during a redirect where you immediately visit another URL before actually returning), the body from the previous connection is still sitting there, in front of the one you actually want. So this function asks for the whole body, to clear it out and free up the space for the body of the next response.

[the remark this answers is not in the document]
mwaggett: so calling `body` has the effect of clearing it out of somewhere? interesting.....

src/ruby/puppetserver-lib/puppet/server/http_response.rb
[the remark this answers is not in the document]
mwaggett: Might be nice for consistency's sake, but I don't care too much if Ruby's error is clear.

src/ruby/puppetserver-lib/puppet/server/http_response.rb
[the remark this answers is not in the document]
justinstoller: I don't think it does, but the response isn't a shared object. IIUC, this is more for ensuring the underlying connection has drained all of this response's body before being re-used by another request. I think this is important when streaming but I believe the apache library handles this for us in non-streaming mode (ie, I think the connection is drained when creating the java response object).

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
Magisus: Looks like we initialize the webserver right above that with a config that has `:port 0` in it. Then we ask the server for its post on that line.

spec/puppet-server-lib/puppet/jvm/http_client_spec.rb
```diff
@@ -5,11 +5,17 @@
...
+    Puppet::Server::HttpClient.new
   end
 
+  let(:url) { 'http://localhost:0/' }
+
   context "when settings are initialized with specific values" do
     before :all do
+      Puppet[:hostcert] = "spec/fixtures/localhost-cert.pem"
+      Puppet[:hostprivkey] = "spec/fixtures/localhost-privkey.pem"
+      Puppet[:localcacert] = "spec/fixtures/ca-cert.pem"
```
Magisus, line 17: There was an order-dependent spec failure, where the [`config_spec`](https://github.com/puppetlabs/puppetserver/blob/6.x/spec/puppet-server-lib/puppet/jvm/config_spec.rb) sets these up and causes the `ssl_context` to be generated, which was then [memoized](https://github.com/puppetlabs/puppetserver/blob/6.x/src/ruby/puppetserver-lib/puppet/server/config.rb#L50-L56) and returned if this file ran after that one. But then this file couldn't be run in isolation.

Magisus:
Also added a commit to fix up the md5 issue. I can break it out if we think there are other blockers to merging this PR and want to get that in to clear up CI in general.

Magisus:
I'm running those acceptance tests now.

src/ruby/puppetserver-lib/puppet/server/http_response.rb
Magisus: Oh still need to fix this.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
```diff
@@ -207,4 +178,25 @@ def self.client_get(request_options)
...
+    raise NotImplementedError
+  end
+
+  def put(url, headers: {}, params: {}, options: {})
+    raise NotImplementedError
+  end
+
+  def delete(url, headers: {}, params: {}, options: {})
+    raise NotImplementedError
+  end
```
mwaggett, line 201: Why are all these needed?

Magisus: I ended up making this class inherit from `Puppet::HTTP::Client` in order to get access to its [URL encoding functions](https://github.com/puppetlabs/puppet/blob/main/lib/puppet/http/client.rb#L279-L285). That class has implementations for these functions that aren't valid inside Puppet Server. 

I'm open to alternatives if people don't like that solution, though I don't really want to duplicate the code. But maybe there's a refactor in Puppet that we want to do to make those more shareable?

mwaggett: ahh, I didn't realize in my re-review that this wasn't the response class that was already inheriting, gotcha. it's really easy to skim PRs I've looked at already.. 🙃

Magisus: Sorry, I should have made a separate commit!

Magisus:
Did get a number of failures in my acceptance run, something is definitely still calling the old 3-arg constructor. I'll be looking into those today.

Magisus:
I'm breaking out the commit to fix the md5 specs in light of these other failures, see https://github.com/puppetlabs/puppetserver/pull/2420. I'll drop that commit from here.

Magisus:
I think that PDB PR will fix most of the acceptance failures I'm seeing.

src/ruby/puppetserver-lib/puppet/server/http_response.rb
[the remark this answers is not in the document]
Magisus: I put in the same error Puppet has.

justinstoller:
Is this a thing where we should provide side by side implementations in 6 and then rip out the old one in 7 rather than updating the existing one?

Magisus:
Yeah that might be good. I mostly did it as a clean cutover here so I could see where all the dragons are.

So, the stages of that would be:
1) Add this implementation in 6.x under a different class name and register it in the runtime. That way anything already pulling the client out of the runtime gets the new API.
2) Update the termini to pull the client out of the context instead of instantiating HttpPool directly.
3) On main, remove HttpPool from Puppet and the old client from Puppetserver.

Sound right?

justinstoller:
That sounds right, though by all means we don't have to do it that way.

I assumed earlier that the higher level indirection that Josh provided in 6 was sufficient and what clients could use to move to the new world in 6 before being forced in 7.

So... what you outlined seemed correct, though I'm not sure, if the PDB PR is the only thing blocked, if it's worth it.

Magisus:
Looking over conversations with Josh and the [shims](https://github.com/puppetlabs/puppet/blob/main/lib/puppet/runtime.rb#L17) we currently have in place, I don't think we gain much by putting this client in 6.x, because I don't think there is any desire to update the `http_pool` to handle the new client API. Instead, we want callers to cut over to using `runtime[:http]` and using the new API from there, which they can do in 6.x with no changes from us because of https://github.com/puppetlabs/puppet/blob/main/lib/puppet/runtime.rb#L17. When we register our own new implementation with the runtime, it will override the use of that shim and seamlessly switch people over to the new version, leaving us free to delete the shim.

src/ruby/puppetserver-lib/puppet/server/http_response.rb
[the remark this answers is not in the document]
joshcooper: Ah the `drain` method is needed in cases where the underlying http client supports persistent HTTP connections and we need to make sure the response body is fully read from the socket. If we don't do that, and we put the connection back in the pool, then the next HTTP request using the same connection/socket will read the (partial) response body from the previous request. One way this can occur is during an HTTP redirect, as it's not super clear what the HTTP server is supposed to send in the body for all of the different redirect codes. https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections.

Although puppetserver's http client doesn't support persistent connections, you can implement the `drain` method by calling `body`, which will ensure the response body is fully read, and then thrown away.

By making `drain` part of the API, then we can easily add persistent connection handling to puppetserver's httpclient in the future without breaking consumers of the API, such as puppetdb terminus, report processor, etc.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
joshcooper: The old API accepted query parameters in the [`path` parameter](https://github.com/puppetlabs/puppet/blob/f638777d6ece022dac8f8c2f6313870d494fc4aa/lib/puppet/network/http/connection.rb#L92) which could either be `http://.../a?b=c` or `/a?b=c`. It just "kind of worked" because `Net::HTTP` accepts either form, yay ruby! The new API makes it possible to specify the URL and params separately as that's the common use case in puppet, e.g. we have a [`Puppet[:reporturl]` and want to pass along some params](https://github.com/puppetlabs/puppet/blob/f638777d6ece022dac8f8c2f6313870d494fc4aa/lib/puppet/reports/http.rb#L17-L26). If you include explicit query parameters then they [overwrite the query in the URL](https://github.com/puppetlabs/puppet/blob/f638777d6ece022dac8f8c2f6313870d494fc4aa/lib/puppet/http/client.rb#L283). I thought about making that an error or warning if they were in both places, but didn't get around to doing that.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
joshcooper: This is good because it makes it explicit that those methods aren't supported. We were just getting lucky before, because if anyone did call `head`, etc then they'd get NoMethodError... but it was hard to tell just by looking at the source.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
```diff
@@ -96,7 +90,14 @@ def create_common_request_options(url, headers, options)
 
     # Ensure multiple requests are not made on the same connection
     headers["Connection"] = "close"
-    request_options = RequestOptions.new(build_url(url))
+
+    if url.is_a?(String)
+      url = URI(url)
```
joshcooper, line 95: I believe the `url` needs to be URL encoded before being passed to the `URI(..)` parser. A good test is to pass in a URL with a space like `http://a.com/b c`. If the space is not encoded, then it will fail to parse:

```
irb(main):023:0> URI('http://a.com/b c')
Traceback (most recent call last):
        6: from /usr/local/opt/rbenv/versions/2.5.7/bin/irb:11:in `<main>'
        5: from (irb):23
        4: from /usr/local/Cellar/rbenv/1.1.2/versions/2.5.7/lib/ruby/2.5.0/uri/common.rb:743:in `URI'
        3: from /usr/local/Cellar/rbenv/1.1.2/versions/2.5.7/lib/ruby/2.5.0/uri/common.rb:237:in `parse'
        2: from /usr/local/Cellar/rbenv/1.1.2/versions/2.5.7/lib/ruby/2.5.0/uri/rfc3986_parser.rb:73:in `parse'
        1: from /usr/local/Cellar/rbenv/1.1.2/versions/2.5.7/lib/ruby/2.5.0/uri/rfc3986_parser.rb:67:in `split'
URI::InvalidURIError (bad URI(is not URI?): http://a.com/b c)
```

That said I'm curious why puppetserver would receive a String and not a URI? I wanted to define the API in terms of a URI object so that the implementation could be certain that encoding had already been performed (to avoid confusion like the string being double URL encoded)

src/ruby/puppetserver-lib/puppet/server/http_client.rb
```diff
@@ -96,7 +90,14 @@ def create_common_request_options(url, headers, options)
...
+    end
+    url = encode_query(url, params)
+
+    # Java will reparse the string into its own URI object
+    request_options = RequestOptions.new(url.to_s)
```
joshcooper, line 100: I naively expected JRuby to be able to use the MRI URI directly, so that we didn't have two different URL parser implementations... If `url` has an escaped space, then `to_s` will yield:

```
irb(main):024:0> URI('http://www.example.com/a%20b').to_s
=> "http://www.example.com/a%20b"
```

Does that need to be unencoded before passing it to `RequestOptions.new`? Or does `RequestOptions` assume it's being passed a URL encoded string?

Magisus: The problem that I saw is that `RequestOptions` is not a JRuby class, it's a Java class. So it does accept a URI, but the Java version, not the Ruby version. Alternatively, it takes a string and converts it to the Java URI. Doing some researach, it looks the  constructor does not do any escaping, i.e. it expects things to be properly esacped/encoded already.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
Magisus: Most of our testing is passing strings right now. I'm not sure if there is anywhere in real code that passes strings, but I can look into it if we want to avoid that.

Magisus: @joshcooper so is it true that the puppet-agent implementation does not worth with a bare string, and I should explicitly disallow that here (that is, error if you pass one instead of coercing)? I can go through and update all of our tests (notably [these](https://github.com/puppetlabs/puppetserver/blob/6.x/test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj)) to wrap their strings in `URI()` if that is the expectation.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
Magisus: The alternative would seem to be to create the Java version here in the Ruby code, using the Java URI's piecemeal constructor, which will do encoding on the various bits, but that seems worse?

src/ruby/puppetserver-lib/puppet/server/http_client.rb
Magisus: I did add a test just now that passes an unescaped string, and it fails as you say. But if we update this API to not accept strings, seems like we don't really need that test.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
```diff
@@ -44,17 +44,18 @@ def self.settings
...
+  end
 
-    @server = server
-    @port = port
-    @use_ssl = options[:use_ssl]
-    @protocol = @use_ssl ? "https" : "http"
+  def get(url, headers: {}, params: {}, options: {}, &block)
+    request_options = create_common_request_options(url, headers, params, options)
+    response = self.class.client_get(request_options)
+    Puppet::Server::HttpResponse.new(response, url)
```
Magisus, line 54: Looks like the return here doesn't match the API, it needs to either return just the body or yield the response object.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
joshcooper: > is it true that the puppet-agent implementation does not work with a bare string, and I should explicitly disallow that here

That's correct, it requires a URI object so that it's clear that URL encoding needs to be handled by the caller.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
joshcooper: I think the client docs are incorrect (and based on an earlier code). The client's `get`, etc methods should return the response when a block is not provided https://github.com/puppetlabs/puppet/blob/c3a37a92d5d08d097160a8bebc8b96d9c1b89402/spec/unit/http/client_spec.rb#L156 or if it is provided https://github.com/puppetlabs/puppet/blob/c3a37a92d5d08d097160a8bebc8b96d9c1b89402/lib/puppet/http/service/file_server.rb#L148.

Magisus: Ah okay, I missed what `execute_streaming` actually does. Cool that makes it easier.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
joshcooper: As long as the `to_s` contains the URL encoded path/query, then passing it through as a string should be ok I think. For example to handle the case where the request path contains a `?` character and there is also a query string, then calling `URI#to_s` needs to return `http://www.example.com/a%3Fb?c=d` instead of `http://www.example.com/a?b?c=d` so that we don't "lose" the request path `/a?b` with query `c=d`.

src/ruby/puppetserver-lib/puppet/server/http_response.rb
```diff
@@ -0,0 +1,48 @@
+require 'puppet/http/response'
+
+class Puppet::Server::HttpResponse < Puppet::HTTP::Response
+  attr_reader :url
+
+  def initialize(java_response, url)
+    @java_response = java_response
+    @url = url
```
Magisus, line 8: Looks like this should also be a URI object that includes the encoded query params, correct?

src/ruby/puppetserver-lib/puppet/server/http_client.rb
```diff
@@ -68,17 +69,10 @@ def post(url, body, headers = {}, options = {})
...
     response = self.class.client_post(request_options)
-    ruby_response(response)
-  end
-
-  def get(url, headers={}, options={})
-
-    request_options = create_common_request_options(url, headers, options)
-    response = self.class.client_get(request_options)
-    ruby_response(response)
+    Puppet::Server::HttpResponse.new(response, url)
```
Magisus, line 72: This url here is wrong because it doesn't include the query params yet. Do you think it makes sense to do the param processing outside of creating the `RequestOptions`, and passing that result here, or would it be more correct to round-trip though Java, and make sure that the URL we put in the response is version as Java understood it?

Something like `Puppet::Server::HttpResponse.new(response, URI(request_options.uri.to_s))` seems to do the right thing if we want to round-trip, but not sure if that's overkill, or even a bad idea.

Magisus: I pushed up a commit with what I suggested here, but still not sure if it's necessary to do something that convoluted.

src/ruby/puppetserver-lib/puppet/server/http_response.rb
joshcooper: Yep. Since https://github.com/puppetlabs/puppetserver/pull/2419/files#diff-5805ba6eb4857e236651f3c3fcdffeb0R72 passes in a URL encoded string, I think you could do `@url = URI(url)` here? Or maybe clearer to have the caller do that and pass the `URI` object.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
```diff
@@ -20,10 +21,9 @@ def initialize(message, cause = nil)
   end
 end
 
-class Puppet::Server::HttpClient
+class Puppet::Server::HttpClient < Puppet::HTTP::Client
 
   OPTION_DEFAULTS = {
-      :use_ssl => true,
       :verify => nil,
       :redirect_limit => 10,
```
joshcooper, line 28: The new client accepts keyword arguments for `:pool`, `:ssl_context`, `:system_ssl_context`, :redirect_limit` and `:retry_limit`. For now I think only the latter two make sense for puppetserver. The `:redirect_limit` can also be overridden on a per-request basis. So 👍  to dropping `:use_ssl` and `:verify`.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
```diff
@@ -44,17 +44,18 @@ def self.settings
     @settings ||= {}
   end
 
-  def initialize(server, port, options = {})
+  def initialize(options = {})
     options = OPTION_DEFAULTS.merge(options)
```
joshcooper, line 48: The new client defaults the `X-Puppet-Version` and `User-Agent` headers based on `Puppet.version` and `Puppet[:http_user_agent]`. Does puppetserver/clj-http-client also do that (or something similar)?

src/ruby/puppetserver-lib/puppet/server/http_client.rb
Magisus: I think what I've done here will be same as that.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
Magisus: Okay I'll update this to make sure that works here.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
Magisus: No, I don't think so. But we can add that here if we want them to be the same.

Magisus: Just realizing that I'm pretty sure this merge does nothing. `options` is a local var that we throw away. I'll need to look into how to properly pass that other configuration.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
Magisus: Currently those options are not honored or passed to the Java implementation when this class is created. The client itself is created lazily with an entirely different set of methods, and the `ClientOptions` class from the clojure HTTP client currently does not support either of those options. I'll have to dig some more.

Magisus: Notes for myself: The retry limit can be configured by setting a `RetryHandler` on the client obejct ([example](https://memorynotfound.com/apache-httpclient-httprequestretryhandler-example/)).

The redirect limit is apparently configurable at least on the client, though I haven't been able to wade through the layers of Java abstraction to figure out exactly how yet.

Magisus: I looked into how to support configuring `redirect_limit` in our Clojure client, and the APIs we use to build the Java client class are the wrong level of abstraction to set it. Looks like the default redirect_limit is 100 (which is much higher than Puppet’s default, but at least it is limited). Considering it wasn’t configurable before, we are going to leave it unconfigurable now.

It also appears that the version of the HttpAsyncClient we’re using doesn’t support retries at all. So I think configuring that isn’t going to happen. They added it in the next major version, which has a totally different API.

Magisus:
I think I've responded to all the feedback, it would be great to get another review of this so that when PDB's change goes in this can be ready to go.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
```diff
@@ -96,7 +94,15 @@ def create_common_request_options(url, headers, options)
 
     # Ensure multiple requests are not made on the same connection
     headers["Connection"] = "close"
-    request_options = RequestOptions.new(build_url(url))
+
+    # These defaults match what the agent attaches to its HTTP requests
+    headers["X-Puppet-Version"] ||= Puppet.version
```
Magisus, line 99: Do we want this to be the Puppet Server version instead? Not sure if that's readily available from here, but we _might_ be able to make that happen.

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -176,59 +173,72 @@ jruby-config :- jruby-schemas/JRubyConfig
...
+      (with-http-client sc {}
+        (let [url (str "http://localhost:" port)]
+          (testing "no credentials"
+            (.runScriptlet sc (format "$response = $c.post(URI('%s'), 'foo')" url))
+            (is (= 401 (.runScriptlet sc "$response.code")))
+            (is (= "Unauthorized" (.runScriptlet sc "$response.reason")))
+            (is (= "access denied" (.runScriptlet sc "$response.body"))))
+
+          (testing "valid credentials"
+            (let [auth "{ :basic_auth => { :user => 'foo', :password => 'bar' }}"]
```
mwaggett, line 205: Are these credentials set up somewhere? I'm not noticing a difference between this test and the "invalid credentials" test, other than the password value. Just curious how this works.

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -323,60 +336,62 @@ jruby-config :- jruby-schemas/JRubyConfig
...
+            (with-http-client sc {}
+              (.runScriptlet sc (format "$response = $c.get(URI('http://localhost:%s'))" port))
+              (is (= 200 (.runScriptlet sc "$response.code")))
               (is (= "bye" (.runScriptlet sc "$response.body")))
-              (.runScriptlet sc (str (format "$c = Puppet::Server::HttpClient.new('localhost', %d, {:use_ssl => %s});"
-                                             10080 true)
-                                    "$response = $c.get('/', {})"))
-              (is (= "200" (.runScriptlet sc "$response.code")))
+              (.runScriptlet sc (str (format "$c = Puppet::Server::HttpClient.new;"
+                                             true)
```
mwaggett, line 394: I don't think this `true` is needed. Looks leftover from the `use_ssl` option.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
```diff
@@ -44,17 +39,23 @@ def self.settings
     @settings ||= {}
   end
 
-  def initialize(server, port, options = {})
-    options = OPTION_DEFAULTS.merge(options)
+  def initialize(options = {})
```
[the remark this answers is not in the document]
Magisus: `Puppet::HTTP::Client` has a [constructor](https://github.com/puppetlabs/puppet/blob/main/lib/puppet/http/client.rb#L31) that specifies arguments we don't need or want and does things we don't need or want, so this overrides it.

If we ever decide we need to support those things described in the comment, they would go in here, but as it is, all of the actual construction of the client (the Clojure and Java classes) happens lazily on the first request.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
```diff
@@ -68,17 +69,14 @@ def post(url, body, headers = {}, options = {})
     end
 
     response = self.class.client_post(request_options)
-    ruby_response(response)
+    Puppet::Server::HttpResponse.new(response, request_options.get_uri.to_s)
```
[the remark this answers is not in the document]
Magisus: Yes.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
```diff
@@ -44,17 +39,23 @@ def self.settings
...
+    #
+    # Similarly, `redirect_limit` is not easily configurable, so we just use the default
+    # of 100. The param will be ignored if passed.
+  end
 
-    @server = server
-    @port = port
-    @use_ssl = options[:use_ssl]
-    @protocol = @use_ssl ? "https" : "http"
+  def get(url, headers: {}, params: {}, options: {}, &block)
```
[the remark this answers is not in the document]
Magisus: I think I lost something here (that isn't used because the old API didn't support it). This should yield the response object like Puppet does if a block is given, in addition to always returning the response object at the end.

spec/puppet-server-lib/puppet/jvm/http_client_spec.rb
Magisus: This fixes it.

src/ruby/puppetserver-lib/puppet/server/http_response.rb
Magisus: I'd rather have it be called "URL" because we really only expect this to ever be an HTTP url, since this is explicitly an HTTP response class. But I'll add a doc comment.

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -176,59 +173,72 @@ jruby-config :- jruby-schemas/JRubyConfig
...
+          (testing "HTTP POST"
+            (is (= "hi" (.runScriptlet sc (format "$c.post(URI('%s'), 'foo').body" url))))))))))
+
+(deftest http-escaped-urls-test
+  (jetty9/with-test-webserver ring-app port
+    (with-scripting-container sc
+      (with-http-client sc {}
+        (let [url (str "http://localhost:" port)]
+          (testing "HTTP GET"
+            (.runScriptlet sc (str "$response = $c.get(URI('" url "/a%20b%3Fc'), params: { foo: 'bar' })"))
```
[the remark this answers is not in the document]
Magisus: I didn't because `%` is used heavily in format strings and I thought it was even worse and more confusing with the escaping it would have needed.

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
Magisus: We set up an app with authenticated middleware [here](https://github.com/puppetlabs/puppetserver/blob/be1af42078c344d5f070ff36df7a70be33c1fb25/test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj#L53-L59) and pass it as the app the macro that sets up the webserver at the top of this test.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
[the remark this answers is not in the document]
Magisus: Though I'm actually a little confused what https://github.com/puppetlabs/puppet/blob/main/lib/puppet/http/client.rb#L133 accomplishes in Puppet's implementation. @joshcooper what is that line for? Do we need it here or does it have something to do with the underlying client implementation? Or am I just spacing on something odd Ruby is doing here?

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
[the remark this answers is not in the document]
mwaggett: ahh right, that makes sense.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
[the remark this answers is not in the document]
joshcooper: Sorry I misspoke, it's really due to persistent connections. When we make a `get` etc request, we borrow a connection from the pool and return it before the `get` method returns. We need to know when the connection is returned whether it can be reused later or should be discarded. If the `get` method returned a response object, but the body wasn't read until sometime later, then it would be reading from the socket associated with a connection that was back in the pool, or worse handed out to someone else, leading to streams getting crossed. So we need to make sure that the response body is fully consumed before the connection is handed back to the pool, which needs to happen before the `get` method returns.

I think it would be good for puppetserver to keep the same semantics as I believe that's how Apache httpclient behaves also. Note sure I have the apache httpclient version correct, but what I've been calling `drain` is what they call `consume` in the description example: https://hc.apache.org/httpcomponents-client-4.5.x/httpclient/apidocs/org/apache/http/impl/client/package-summary.html#package_description

```
      HttpGet httpGet = new HttpGet("http://targethost/homepage");
      CloseableHttpResponse response = httpclient.execute(httpGet);
      try {
          System.out.println(response.getStatusLine());
          HttpEntity entity = response.getEntity();
          // do something useful with the response body
          // and ensure it is fully consumed
          EntityUtils.consume(entity);
      } finally {
          response.close();
      }
```

If puppetserver kept the same semantics it would make it easier to support persistent connections between puppetserver and puppetdb in the future.

mwaggett (approved):
I'd like someone with more familiarity to also approve.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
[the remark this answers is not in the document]
joshcooper: I'm also working on a PR to updated the http client yardocs as it's definitely missing details like this.

Magisus:
Also still need to wait for the PDB changes to actually merge this. Thanks for the review though!

src/ruby/puppetserver-lib/puppet/server/http_client.rb
```diff
@@ -44,17 +39,29 @@ def self.settings
...
+    request_options = create_common_request_options(url, headers, params, options)
+    java_response = self.class.client_get(request_options)
+    ruby_response = Puppet::Server::HttpResponse.new(java_response,
+                                                     URI(request_options.uri.to_s))
```
joshcooper, line 55: Since [`create_common_request_options` now raises if `url` is not a `URI`](https://github.com/puppetlabs/puppetserver/pull/2419/files#diff-37e461ce55c05f31a9b64b7eadb239a8ae8261574f3236d3928fd20e8e07485aR89) can we pass `url` directly to the `HttpResponse` constructor?

src/ruby/puppetserver-lib/puppet/server/http_client.rb
joshcooper: IMO `X-*-Version` is mostly useful to include in HTTP responses from the server so that the HTTP client can alter it's behavior, such as downgrade to an older protocol. It's less useful for HTTP clients to send in requests, because generally the client can negotiate by sending `Accept` header and query parameters. In puppet we first added it for the HTTP server case, but then also added it for the HTTP client case. I'd be fine leaving this out. But if you wanted to add it, I think adding `X-PuppetServer-Version` with the puppetserver version would be better?

src/ruby/puppetserver-lib/puppet/server/http_client.rb
```diff
@@ -96,7 +106,15 @@ def create_common_request_options(url, headers, options)
 
     # Ensure multiple requests are not made on the same connection
     headers["Connection"] = "close"
-    request_options = RequestOptions.new(build_url(url))
+
+    # These defaults match what the agent attaches to its HTTP requests
+    headers["X-Puppet-Version"] ||= Puppet.version
+    headers["User-Agent"] ||= Puppet[:http_user_agent]
```
joshcooper, line 112: The main thing with the user agent string is that it's logged in HTTP server access logs and having unique user agents can be helpful to understand who made the request. If puppet apply and puppetserver both connected to puppetdb, then it might be useful to have different user agent strings. Puppet sends a user agent string of the form:

```
User-Agent: Puppet/7.0.0 Ruby/2.5.7-p206 (x86_64-darwin18)
```

So something like this might be useful (using  puppetserver's version, `JRUBY_VERSION` and `PLATFORM`):

```
User-Agent: Puppetserver/7.0.0 JRuby/9.2.6.0 (jruby)
```

But this is super lower priority. Also t appears `Puppet[:http_user_agent]` is already "different" due to the `java` platform:

```
 "Puppet/7.0.0 Ruby/2.5.3-p0 (java)"
```

src/ruby/puppetserver-lib/puppet/server/http_client.rb
Magisus: Okay. I think I'll leave it out for now. We can always add it later if we need it for something.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
Magisus: Yeah we can. This was useful in the middle to make sure the query params were making it to Java correctly, but I think it's excessive to leave in.

joshcooper:
LGTM, I verified puppetserver could post reports to my local httpserver using config:

```ini
[server]
reports=http
reporturl=http://localhost:8080
```

And it looks good:

```
Path: /
Headers:
Content-Type: application/x-yaml; charset=UTF-8
Connection: close
X-Puppet-Version: 7.0.0
Accept-Encoding: gzip, deflate
User-Agent: Puppet/7.0.0 Ruby/2.5.7-p0 (java)
Content-Length: 6576
Host: localhost:8080

--- !ruby/object:Puppet::Transaction::Report
host: localhost
time: '2020-10-16T15:32:10.289718000-07:00'
...
```

src/ruby/puppetserver-lib/puppet/server/http_client.rb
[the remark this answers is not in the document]
Magisus: Currently getting a bunch of test failures when I try to add `EntityUtils.consume`. I sort of suspect the fact that we are using the async Apache client under the hood, even for sync clients, might be interfering with it. Need to look into it more to see if what we're already doing is sufficient.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
Magisus: Given that getting Puppet Server's version into the Ruby code is hard (I looked into it a bit), I think leaving this as-is, with the `(java)` identifier, is okay for now. Still an improvement over sending nothing before. I think I'll file a follow-up ticket so we can look into plumbing that through and cleaning this up.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
Magisus: Updated.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
Magisus: Removed.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
Magisus: Added a comment about this to the code too.

Magisus:
I'm running this through acceptance tests again.

joshcooper (approved):

src/ruby/puppetserver-lib/puppet/server/http_client.rb
Magisus: Oh right, we can't because we haven't attached the query params yet... I assume we want those params included in the URL saved in the response?
```
FAIL in (http-escaped-urls-test) (http_client_test.clj:191)
HTTP GET
expected: (= (str url "/a%20b%3Fc?foo=bar") (.runScriptlet sc "$response.url.to_s"))
  actual: (not (= "http://localhost:40935/a%20b%3Fc?foo=bar" "http://localhost:40935/a%20b%3Fc"))
```

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -176,59 +173,72 @@ jruby-config :- jruby-schemas/JRubyConfig
...
+            (is (= 200 (.runScriptlet sc "$response.code")))
+            (is (= (str url "/a%20b%3Fc?foo=bar") (.runScriptlet sc "$response.url.to_s")))))))))
```
joshcooper, line 191: I think this is one area where puppet/ruby behaves differently. The response always contains the URL that the caller specified, not the URL that results when query parameters are "merged".

src/ruby/puppetserver-lib/puppet/server/http_client.rb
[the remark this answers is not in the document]
justinstoller: So we're definitely not calling `EntityUtils.consume`, but my understanding was that, if the response is set as TEXT (which we force [here](https://github.com/puppetlabs/puppetserver/blob/6.x/src/ruby/puppetserver-lib/puppet/server/http_client.rb#L104)) then we read the response body, attempt to coerce it to the encoding, and close the underlying resource ([here](https://github.com/puppetlabs/clj-http-client/blob/ec78805007db944b0ef7b4e4212d4969d7d8d0a4/src/java/com/puppetlabs/http/client/impl/JavaClient.java#L666)). We do similar for streaming response [here](https://github.com/puppetlabs/clj-http-client/blob/ec78805007db944b0ef7b4e4212d4969d7d8d0a4/src/java/com/puppetlabs/http/client/impl/StreamingAsyncResponseConsumer.java#L49) in our client wrapper. I don't see a safe way to use the non-streaming byte responses in our jvm client library, but we don't expose that in the Ruby client. I'm also not sure if the TEXT processing wouldn't fail to close the resources if there was an encoding error (but I've been unable to verify that).

test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
Magisus: Okay, then I think we want to also. I'll put this back so that that happens.

src/ruby/puppetserver-lib/puppet/server/http_client.rb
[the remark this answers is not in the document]
Magisus: We could move that `if body then close` bit to a finally block maybe to be safe? Not sure if there's a reason it's not like that already...

acceptance/scripts/generic/testrun.sh
```diff
@@ -16,7 +16,7 @@ do_init()
   BEAKER_INIT="$BEAKER_INIT --helper $BEAKER_HELPER"
   BEAKER_INIT="$BEAKER_INIT --options-file $BEAKER_OPTIONS"
   BEAKER_INIT="$BEAKER_INIT --post-suite $BEAKER_POSTSUITE"
-  BEAKER_INIT="$BEAKER_INIT --load-path acceptance/lib"
+  BEAKER_INIT="$BEAKER_INIT --load-path $BEAKER_LOADPATH"
```
mwaggett, line 19: Why did this change? Would we still want to default to `acceptance/lib` if `$BEAKER_LOADPATH` isn't set?

Magisus: We do, the specification for `BEAKER_LOADPATH` below in the main script already supplies the default. We were just ignoring the env var entirely when it came to actually building the beaker invocation. I think I broke this when I refactored this file a while back.

mwaggett (approved):

PR #2421, opened 2020-09-25
(SERVER-2809) Add rake task for updating facter submodule

Magisus:
This commit adds a new rake task that can be used to update the Facter
submodule commit based on the last passing version published by the
puppet-agent Jenkins pipeline.

It introduces a new variable to track which branch to pull the Facter
SHA from, because while the 6.x agent is still using Facter 3, in our
submodule we use Facter 4, because we need it to be a Ruby lib rather
than C++.

This also renames some tasks to be more descriptive of the new state, so
any automation around these tasks will need to be updated accordingly.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -13,10 +13,15 @@ FACTER_LIB = File.join(PROJECT_ROOT, 'ruby', 'facter', 'lib')
 PUPPET_SERVER_RUBY_SRC = File.join(PROJECT_ROOT, 'src', 'ruby', 'puppetserver-lib')
 PUPPET_SERVER_RUBY_SPEC = File.join(PROJECT_ROOT, 'spec')
 PUPPET_SUBMODULE_PATH = File.join('ruby','puppet')
+FACTER_SUBMODULE_PATH = File.join('ruby','facter')
 # Branch of puppetserver for which to update submodule pins
 PUPPETSERVER_BRANCH = ENV['PUPPETSERVER_BRANCH'] || 'master'
 # Branch of puppet-agent to track for passing puppet SHA
 PUPPET_AGENT_BRANCH = ENV['PUPPET_AGENT_BRANCH'] || 'master'
+# Branch of puppet-agent to track for Facter's passing SHA.
+# This needs to be separate because for 6.x, we want to use facter#main
+# instead of Facter 3.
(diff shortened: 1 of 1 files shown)


Magisus:
Related CJC PR: https://github.com/puppetlabs/ci-job-configs/pull/7294

That should be merged soon after this goes in, or else next time it runs it won't be able to find the rake task.

Rakefile
```diff
@@ -125,30 +137,39 @@ def replace_puppet_pins(passing_puppetagent_sha)
...
     system(git_checkout_command)
     # replace puppet-agent sha pin in beaker options file
     replace_puppet_pins(lookup_passing_puppetagent_sha(my_jenkins_passing_json))
   end
+  desc 'update facter submodule commit'
+  task :update_facter_version do
+    my_jenkins_passing_json = jenkins_passing_json_parsed(FACTER_BRANCH)
+    git_checkout_command = "cd #{FACTER_SUBMODULE_PATH} && git fetch origin && " \
+      "git checkout #{lookup_passing_facter_sha(my_jenkins_passing_json)}"
+    puts("checking out known passing puppet version in submodule: `#{git_checkout_command}`")
```
mwaggett, line 157:
```suggestion
    puts("checking out known passing facter version in submodule: `#{git_checkout_command}`")
```

Rakefile
```diff
@@ -125,30 +137,39 @@ def replace_puppet_pins(passing_puppetagent_sha)
...
+    my_jenkins_passing_json = jenkins_passing_json_parsed(PUPPET_AGENT_BRANCH)
     git_checkout_command = "cd #{PUPPET_SUBMODULE_PATH} && git fetch origin && " \
       "git checkout #{lookup_passing_puppet_sha(my_jenkins_passing_json)}"
     puts("checking out known passing puppet version in submodule: `#{git_checkout_command}`")
     system(git_checkout_command)
     # replace puppet-agent sha pin in beaker options file
     replace_puppet_pins(lookup_passing_puppetagent_sha(my_jenkins_passing_json))
   end
+  desc 'update facter submodule commit'
+  task :update_facter_version do
```
mwaggett, line 153: might be worth putting this logic into a separate function, since it's nearly identical to the task for puppet?

mwaggett (approved):
Have you tested this at all?

Magisus:
Yes. Seems fine.

PR #2425, opened 2020-09-30
(SERVER-2763) Use `uri:classloader` to avoid globbing issue

Magisus:
The upgrade to Rubygems 3 broken some of JRuby's globbing code. They
fixed the issue around `uri:classloader`, but not issues with the
`classpath` protocol, which we were using to load our Ruby code inside
puppetserver. However, it appears we can switch to using
`uri:classloader` instead, so this PR changes our loadpath addition to
use that protocol. This means that globbing against our loadpath (e.g.
when installing gem documentation) now works.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
@@ -30,9 +30,9 @@
   directory is placed into the root of the jar structure which is on the
   classpath.
 
-  See also:  http://jruby.org/apidocs/org/jruby/runtime/load/LoadService.html"
+  See also:  https://www.javadoc.io/doc/org.jruby/jruby-core/latest/org/jruby/runtime/load/LoadService.html"
 
-  "classpath:/puppetserver-lib")
+  "uri:classloader:/puppetserver-lib")
 
 (def default-http-connect-timeout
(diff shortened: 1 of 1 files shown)


Magisus:
I was not able to find a clear definition of the difference between these two protocols. The docs are very sparse, though I've updated the link in the code to a working link anyway, even though I didn't find it very helpful.

mwaggett:
I assume you manually played around with this?

Magisus:
Yeah, first I did the repro from the ticket using `lein irb` to `glob` in isolation.
Then I started puppetserver in repl to make sure it could function with the protocol changed.
Then I built a jar and installed it on a machine and used `puppetserver gem install` successfully without the `--no-document` flag.

I've asked in the ticket if there's an easy test we can do to verify the issue with Impact Analysis in our Ops environment.

mwaggett (approved):

Magisus:
JRuby issue https://github.com/jruby/jruby/issues/6421

Magisus:
We're going to hold off on merging this for a couple days to see if we can get any input from the JRuby folks on the issue filed.

Magisus:
No word from JRuby yet, so I think this should just go in for now, so we don't accidentally miss this upcoming release.

PR #2428, opened 2020-10-09
(PUP-10673) add simple server endpoint

gimmyxd:
This commit adds `/status/v1/simple/server` endpoint
similar to `/status/v1/simple/master`

It also updates the `/status/v1/services` endpoint
with a new field `server` containing the same
information as `master`.

LE: this might do more than i thought.

### DIFF
--- a/src/clj/puppetlabs/services/master/master_service.clj
+++ b/src/clj/puppetlabs/services/master/master_service.clj
@@ -184,6 +184,11 @@
       (status-core/get-artifact-version "puppetlabs" "puppetserver")
       master-service-status-version
       (partial core/v1-status http-metrics http-client-metric-ids-for-status registry))
+     (register-status
+      "server"
+      (status-core/get-artifact-version "puppetlabs" "puppetserver")
+      master-service-status-version
+      (partial core/v1-status http-metrics http-client-metric-ids-for-status registry))
      (-> context
          (assoc :http-metrics http-metrics)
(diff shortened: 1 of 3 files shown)


gimmyxd:
`/status/v1/services/server`
```
{
  "service_version": "6.14.0-SNAPSHOT",
  "service_status_version": 1,
  "detail_level": "info",
  "state": "running",
  "status": {},
  "active_alerts": [],
  "service_name": "server"
}
```

`/status/v1/simple/server`

```
running
```

Magisus:
This looks right to me! Are we thinking we would remove the `master` status in 7, or do we need to keep it around one more cycle?

joshcooper:
Since versions of 6.x that still call the `master` status endpoint, I think we'll need to acccept `master` status requests for a long time.

joshcooper:
```
expected: (= 8140 (get-puppet-config-value jruby-puppet :masterport))
  actual: (not (= 8140 "8140"))
```

It looks like the puppet submodule in puppetserver#6.x doesn't contain the [revert](https://github.com/puppetlabs/puppet/commit/69c20ebe66f73be9b672b5d77d00ef5a8f6ada5d) that resolves the issue above. But it looks like the submodule bump only happens if puppetserver clojure unit tests pass, so catch-22. Can we manually bump the puppet submodule?

Magisus:
The bump actually runs periodically (daily at 2am), not as part of the main pipeline. But I did just kick it. This might need to be rebased to pick that up.

Magisus (approved):

PR #2433, opened 2020-10-21
(PE-30068) Log certnames that are revoked

reidmv:
Previously, when a cert was revoked the certname was not logged. This
caused problems keeping track of what certnames were revoked. Prior to
the bulk API existing, customers had been largely able to review
revocation activity based on puppetserver-access.log. This new logging
effectively restores prior log functionality.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -1339,9 +1339,10 @@
         new-crl (utils/revoke-multiple original-crl ca-private-key
                                        ca-public-key serials)]
     (write-crl new-crl cacrl)
-    (log/debug (if (= 1 serial-count)
-                 (i18n/trs "Revoked 1 certificate.")
-                 (i18n/trs "Revoked {0} certificates." serial-count)))))
+    (log/info (i18n/trsn "Revoked 1 certificate: {1}"
+                         "Revoked {0} certificates: {1}"
+                         serial-count
+                         (str/join ", " subjects)))))
 
 (schema/defn ^:always-validate set-certificate-status!
   "Sign or revoke the certificate for the given subject."


src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1339,9 +1339,10 @@
         new-crl (utils/revoke-multiple original-crl ca-private-key
                                        ca-public-key serials)]
     (write-crl new-crl cacrl)
-    (log/debug (if (= 1 serial-count)
-                 (i18n/trs "Revoked 1 certificate.")
-                 (i18n/trs "Revoked {0} certificates." serial-count)))))
+    (log/debug (i18n/trsn "Revoked 1 certificate: {1}"
```
justinstoller, line 1342: Doesn't this still need the conditional?
Also, we originally didn't do this because we were wondering about the usefulness of printing out what may be thousands of certnames in a log line. I assume that isn't a concern?

reidmv: My understanding of `i18n/trsn` is that it is specifically designed to address this sigularity/plurality thing, right?

>trsn/trun are similar to trs/tru except that they support pluralization of strings. The first argument is the singular version of the string, the second argument must be the plural form of the string. The third argument is the count value to determine the level of pluralization. Any additional arguments will be used for additional formatting

If I'm reading that right, it should mean we don't need the `(if)` here. https://github.com/puppetlabs/clj-i18n#developer-usage

Re: certname printing, there could be many, yes, but based on the customer reception it seems fairly critical that when certname(s) are revoked, a record is made somewhere of that activity occurring. In the previous iteration using the old endpoint, this was incidentally logged in puppetserver-access.log. So, we shouldn't actually be logging much more than we were.

Would line-length present any problems logging, in the theoretical event of thousands of certnames? Is it alright to emit one log entry for all, or do we need to batch/individualize events?

reidmv: Oh, and I guess heads up: the customer is going to be formally pushing for this logging as a hotfix Real Soon Now™.

Magisus: Given that need, is it okay that this is only logged at debug? Or does it need to switch to info?

reidmv: Oh shoot, yeah, it needs to be info. Glad you caught that!!

reidmv: Oh shoot, yeah, it needs to be info. Glad you caught that!! Fixed

Magisus:
I tested this with 1500 certs, it looks fine, and no obvious truncation is going on (nor could I find evidence in the docs that that happens).
```
2020-10-23 15:47:55,652 INFO  [qtp452338781-22] [p.p.certificate-authority] Revoked 1,497 certificates: node4, node5, node6, node7, node8, node9, node10, node11, node12, node13, node14, node15, node16, node17, node18, node19, node20, node21, node22, node23, node24, node25, node26,.....
```

Magisus (approved):

mwaggett (approved):

PR #2437, opened 2020-10-27
(SERVER-2791) Remove legacy routes service from bootstrap config

Magisus:
This commit removes the legacy routes service, so we will no longer
handle requests to the legacy routes. It also removes the associated
tests for the service, but leaves the code in place, to be removed
later.


test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
```diff
@@ -217,8 +208,8 @@
         app
         {:jruby-puppet {:master-conf-dir master-conf-dir}}
         (let [response (http-client/put
-                        (str "https://localhost:8140"
-                            "puppet-ca/v1/certificate_status/test_cert_ca_true")
+                        (str "https://localhost:8140/"
```
Magisus, line 211: These have always been wrong, but apparently weren't failing before. Maybe the legacy routes service used to do some munging that avoided this?

Magisus:
Beaker smoke tests pass with this, the regular puppet tests tend to take too long when I run them locally (the VMs die), but I don't see any obvious ones that exercise these routes. @joshcooper do you know of any?

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
justinstoller: That's so weird

joshcooper: Maybe due to https://github.com/puppetlabs/puppetserver/blob/f296fdf660535c48e80cf74f2e823a7b6bd848f3/src/ruby/puppetserver-lib/puppet/server/http_client.rb#L151-L153?

Magisus: Except that isn't in the new client implementation... which is where these were failing.

Magisus: That is removed from the new client implementation, but the tests still pass before this PR.

Magisus:
We don't think there are any beaker tests in Puppet that exercise this. I'd be okay waiting for CI to tell us for sure.

Magisus:
The PE side of this is ticketed as https://tickets.puppetlabs.com/browse/PE-30415 and can wait till after Platform 7 stop ship.

joshcooper:
yay 🔥 !

justinstoller (approved):

PR #2444, opened 2020-11-04
Mergeup, docs removal

Magisus:


documentation/README.md
```diff
@@ -0,0 +1,5 @@
+## Docs have moved!
+
+Puppetserver's markdown for documentation has moved to https://github.com/puppetlabs/osp-docs. Please file a ticket in the SERVER JIRA project to request changes.
```
mwaggett, line 3: SERVER? Not DOCUMENT?

mwaggett (commented):
We should up the README's references to docs.

documentation/README.md
Magisus: We do most of our own docs edits, and SERVER will show up on our radar more reliably than DOCUMENT. But I could be convinced otherwise.

mwaggett: no that's fine. Just wasn't really sure if that was something we'd jump on or whether the docs team would do it. But if we'll be doing the edits, SERVER project makes sense.

Magisus:
Superseded by https://github.com/puppetlabs/puppetserver/pull/2451

PR #2446, opened 2020-11-05
(PUP-10720, SERVER-2870) Puppet 7 updates

justinstoller:


test/integration/puppetlabs/puppetserver/testutils.clj
```diff
@@ -375,6 +375,17 @@
   ([puppet-conf-file dest-dir]
    (with-puppet-conf-fixture {"puppet.conf" puppet-conf-file} dest-dir)))
 
+(defmacro with-stub-puppet-conf
+  "Provides similar functionality of `with-puppet-conf` but uses a default
```
mwaggett, line 379: I was a little confused by your comments on the other PR - should this reference `with-puppet-conf-files` instead?

Magisus (approved):
Glad to see us getting in the changes to avoid deprecation warnings!

test/integration/puppetlabs/puppetserver/testutils.clj
justinstoller: No, there's two ways in clojure to do the equivalent of Ruby's passing in a block. Define a function that takes another function, or define a macro that takes forms.

Clojure's test library has equivalents to Ruby RSpec's before, after, and around hooks, (these are the `use-fixtures` mentioned below) and they use the former approach, they take functions which are then passed functions (this is what `with-puppet-conf-fixture` and transitively `with-puppet-conf` are geared towards).

Inside of actual deftests this project tends to go with the latter, macro approach to "block passing".

So this macro does something most similar to what `with-puppet-conf` does, but is implemented in a way to be used outside of the "rspec around-style hook" that `with-puppet-conf` is geared towards.

(Note you can actually mix and match those approaches (iirc, code-manager uses function passing for everything), but I was just trying to follow _what seemed idiomatic for this project_)

mwaggett: gotcha, thank you for explaining!

mwaggett (approved):

mwaggett:
Looks like a schema somewhere needs to be updated.

mwaggett:
looks like this needs the master -> server update: https://github.com/puppetlabs/puppetserver/blob/main/spec/puppet-server-lib/puppet/jvm/master_spec.rb

mwaggett (approved):

mwaggett:
not sure what's up with Travis at this point..

justinstoller:
container tests

justinstoller:
I looked at the build log and the last several failures on Azure come from failure to find remote dependencies, though in the two different builds I saw it failed to find _different_ dependencies. At least once for the history of this PR it also passed (successfully built a docker container). Based on that I'm going to merge past the Azure container pipeline failure because it looks like they're just having networking troubles.

PR #2447, opened 2020-11-06
(SERVER-2876) Support pluginsync for bolt projects

pcarlisle:
Add support for retrieving file metadata and content for plugins (lib and facts.d) for bolt projects.

### DIFF
--- a/src/clj/puppetlabs/services/master/file_serving.clj
+++ b/src/clj/puppetlabs/services/master/file_serving.clj
@@ -1,10 +1,97 @@
 (ns puppetlabs.services.master.file-serving
   (:require [bidi.bidi :as bidi]
             [clj-yaml.core :as yaml]
+            [clojure.java.io :as io]
             [clojure.string :as str]
+            [digest :as digest]
             [me.raynes.fs :as fs]
             [puppetlabs.i18n.core :as i18n]
-            [ring.util.response :as rr]))
+            [puppetlabs.ring-middleware.utils :as middleware-utils]
+            [ring.util.response :as rr])
+  (:import com.sun.security.auth.module.UnixSystem
+           [java.nio.file Files FileSystems FileVisitOption FileVisitResult LinkOption Paths SimpleFileVisitor]))
+
+(def follow-links
+  (into-array LinkOption []))
+
+(defn as-path
+  [string & more]
+  (Paths/get string (into-array String more)))
+
+(defn path-exists?
+  [path]
+  (Files/exists path follow-links))
+
+(defn get-checksum
+  [file algorithm]
+  (case algorithm
+    "md5" (str "{md5}" (digest/md5 (io/file file)))
+    "sha256" (str "{sha256}" (digest/sha-256 (io/file file)))
+    :default (throw (Exception. "Unsupported digest"))))
+
+(defn read-attributes
+  [path checksum-type ignore-source-permissions]
+  (let [attributes (Files/readAttributes path "unix:*" follow-links)]
+    {:owner (if ignore-source-permissions
+              (.getUid (UnixSystem.))
+              (get attributes "uid"))
+     :group (if ignore-source-permissions
+              (.getGid (UnixSystem.))
+              (get attributes "gid"))
+     :mode (if ignore-source-permissions
+             0644  ;; Yes, Puppet even sends 644 for directories in this case
+             (bit-and (get attributes "mode") 07777))
+     :type (if (get attributes "isDirectory")
+             "directory"
+             "file")
+     :checksum (if (get attributes "isDirectory")
+                 {:type "ctime"
+                  :value (str "{ctime}" (get attributes "ctime"))}
+                 {:type checksum-type
+                  :value (get-checksum path checksum-type)})}))
+
+(defn get-file-metadata
+  [checksum-type ignore-source-permissions [relative-path root]]
+  (merge (read-attributes (.resolve root relative-path) checksum-type ignore-source-permissions)
+         {:path (.toString root)
+          :relative_path (.toString relative-path)
+          :links "follow"
+          :destination nil}))
+
+(defn relativize
+  "A version of java's Path#relativize that matches puppet's behavior"
+  [root path]
+  (let [relative-path (.relativize root path)]
+    (if (= (.toString relative-path) "")
+      (as-path ".")
+      relative-path)))
+
+(defn make-visitor
+  [tree root ignores]
+  (let [filesystem (FileSystems/getDefault)
+        matchers (map #(.getPathMatcher filesystem (str "glob:" %)) ignores)
+        ignore? (fn [path]
+                  (let [file-name (.getFileName path)]
+                   (some #(.matches % file-name) matchers)))]
+    (proxy [SimpleFileVisitor] []
+     (preVisitDirectory [path attributes]
+       (if (ignore? path)
+         FileVisitResult/SKIP_SUBTREE
+         (do
+           (swap! tree assoc (relativize root path) root)
+           FileVisitResult/CONTINUE)))
+     (visitFile [path attributes]
+       (when-not (ignore? path)
+         (swap! tree assoc (relativize root path) root))
+       FileVisitResult/CONTINUE))))
+
+(defn walk-directory
+  [path ignores]
+  (let [results (atom {})
+        absolute-path (.toAbsolutePath path)
+        visitor (make-visitor results absolute-path ignores)]
+    (Files/walkFileTree absolute-path #{FileVisitOption/FOLLOW_LINKS} Integer/MAX_VALUE visitor)
+    @results))
 
 (defn is-bolt-project?
   [dir]
@@ -15,11 +102,11 @@
 (defn get-project-root
   "Lookup a project by name and return the path where its project files are
   located as a string."
-  [projects-dir project-name]
-  (let [boltdir-path (str projects-dir "/" project-name "/Boltdir")]
+  [projects-dir project-name-and-ref]
+  (let [boltdir-path (str projects-dir "/" project-name-and-ref "/Boltdir")]
     (if (fs/exists? boltdir-path)
       boltdir-path
-      (str projects-dir "/" project-name))))
+      (str projects-dir "/" project-name-and-ref))))
 
 (defn list-dirs-in-paths
   "Given a list of directories, return a list of all the subdirectories
@@ -28,7 +115,7 @@
   [paths]
   (->> paths
        (filter fs/exists?)
-       (mapcat fs/list-dir)
+       (mapcat (comp sort fs/list-dir))
        (filter fs/directory?)))
 
 (defn dirs-in-project-modulepath
@@ -61,7 +148,7 @@
 (defn read-bolt-project-config
   [project-dir]
   (let [config-path (str project-dir "/bolt-project.yaml")]
-    (if (fs/file? config-path)
+    (when (fs/file? config-path)
      (yaml/parse-string (slurp config-path)))))
 
 (def default-project-modulepath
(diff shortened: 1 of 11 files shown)


pcarlisle:
Accidentally merged my first refactoring commit with the main feature commit the first time, this push fixes that to make review easier, no actual code changes

src/clj/puppetlabs/services/master/file_serving.clj
```diff
@@ -103,16 +190,64 @@
                         project-root
                         (find-project-module project-root module modulepath))
           file-path (str module-root "/" (mount->path-component mount) "/" path)]
-      (if (fs/exists? file-path)
+      (when (fs/exists? file-path)
         file-path))))
 
+(defn mount-dirs-in-modulepath
+  [mount modulepath project-root project-as-module?]
```
Magisus, line 197: The two places this is called, the last arg you're passing is `project-name`, which it looks like we extract from the project config. Do we expect that to be nil in some scenarios and that's why it's being treated as a boolean here? I feel like maybe it shouldn't ever be nil, but I may be missing some conflating of ideas that's going on here.

src/clj/puppetlabs/services/master/file_serving.clj
```diff
@@ -1,10 +1,97 @@
...
+
+(defn as-path
+  [string & more]
+  (Paths/get string (into-array String more)))
+
+(defn path-exists?
+  [path]
+  (Files/exists path (into-array LinkOption [])))
+
+(as-path "/home/patrick" "data")
```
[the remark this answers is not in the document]
pcarlisle: oops, yes this is how i use the repl

src/clj/puppetlabs/services/master/file_serving.clj
pcarlisle: Yes, the existence of a project name is how people opt in to treating the project itself as a module. I can try to make this more clear.

src/clj/puppetlabs/services/master/file_serving.clj
```diff
@@ -166,3 +303,62 @@
...
+  "Handle a file_metadatas request for a bolt project."
+  [bolt-projects-dir request]
+  (let [project (get-in request [:params "project"])
+        path (get-in request [:params :rest])
+        match (bidi/match-route project-routes path)]
+    (when match
+      (let [mount-point (get-in match [:route-params :mount-point])
+            mount-type (get-in match [:handler])
+            ignore (get-in request [:params "ignore"] [])
+            checksum-type (get-in request [:params "checksum_type"] "md5")]
```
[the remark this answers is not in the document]
pcarlisle: I was wondering about this...

I think in practice we always get the checksum type specified by the client, but also that puppet is still using md5 here.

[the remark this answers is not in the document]
Magisus: Okay. I think that changed in Puppet 7. @joshcooper can you confirm?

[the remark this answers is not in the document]
joshcooper: Yeah, when the [server creates the metadata object](https://github.com/puppetlabs/puppet/blob/f664d6a21656f80b54b396529c84ee7e24108921/lib/puppet/file_serving/terminus_helper.rb#L9) it defaults to the [server's `Puppet[:digest_algorithm]` value](https://github.com/puppetlabs/puppet/blob/f664d6a21656f80b54b396529c84ee7e24108921/lib/puppet/file_serving/metadata.rb#L135). But if the client requests a specific checksum, then [it takes precedence](https://github.com/puppetlabs/puppet/blob/f664d6a21656f80b54b396529c84ee7e24108921/lib/puppet/file_serving/terminus_helper.rb#L12), so that the call to `FileMetadata#collect` uses that.

Puppet 7 does default to `sha256`, but it could be switched to `md5` I think the only reason to do that would be if you're using remote file buckets and you have pre-7 agents.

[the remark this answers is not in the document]
pcarlisle: This is all new stuff, I'll make it default to sha256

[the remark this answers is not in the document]
Magisus: However, I do think we want this targeted at 6.x, right? Everything else has landed in that stream so far.

src/clj/puppetlabs/services/master/file_serving.clj
```diff
@@ -0,0 +1,364 @@
...
+  dir in a module, return the path of the file if it exists in the lib dir."
+  [relative-path lib-root]
+  (let [file-path (.resolve lib-root relative-path)]
+    (when (path-exists? file-path)
+      (.toString file-path))))
+
+(defn find-project-plugin-file
+  "Given a relative path as received by the plugins mount, search all lib dirs
+  and return the path to the file if it's found in any of them."
+  [bolt-projects-dir project-ref mount relative-path]
```
mwaggett, line 221: Is `project-ref` what you're using to refer to name+ref? I think [this email thread](https://groups.google.com/a/puppet.com/d/msgid/team-skeletor/CAOPg%3DQck4EPde9w%2B8gYakbQWdsAAmHtevbLmO_C0L0aVwevqBQ%40mail.gmail.com?utm_medium=email&utm_source=footer) is suggesting we just use `project` for that? `project-ref` makes me think it's _just_ the ref.

src/clj/puppetlabs/services/master/file_serving.clj
```diff
@@ -0,0 +1,364 @@
...
+(defn is-bolt-project?
+  [dir]
+  (or (fs/exists? (str dir "/bolt-project.yaml"))
+      (fs/exists? (str dir "/bolt.yaml"))
+      (fs/exists? (str dir "/Boltdir"))))
+
+(defn get-project-root
+  "Lookup a project by name and return the path where its project files are
+  located as a string."
+  [projects-dir project-name]
```
mwaggett, line 105: `project-name` needs to include a ref, right? Should we use `project` instead?

src/clj/puppetlabs/services/master/file_serving.clj
```diff
@@ -0,0 +1,364 @@
...
+  modules and returns a list of files smashed together."
+  [bolt-projects-dir project-ref mount checksum-type ignores ignore-source-permissions]
+  (when (is-bolt-project? (str bolt-projects-dir "/" project-ref))
+    (let [project-root (get-project-root bolt-projects-dir project-ref)
+          project-config (read-bolt-project-config project-root)
+          project-name (get project-config :name)
+          modulepath (get-project-modulepath project-config)
+          files (->> (mount-dirs-in-modulepath mount modulepath project-root project-name)
+                     (map #(walk-directory % ignores))
+                     reverse
```
mwaggett, line 240: Why reverse?

mwaggett:
seconding Maggie that this should probably target 6.x, since that's the stream where the rest of the related work has been landing.

src/clj/puppetlabs/services/master/file_serving.clj
```diff
@@ -0,0 +1,364 @@
...
+(defn read-attributes
+  [path checksum-type ignore-source-permissions]
+  (let [attributes (Files/readAttributes path "unix:*" follow-links)]
+    {:owner (if ignore-source-permissions
+              (.getUid (UnixSystem.))
+              (get attributes "uid"))
+     :group (if ignore-source-permissions
+              (.getGid (UnixSystem.))
+              (get attributes "gid"))
+     :mode (bit-and (get attributes "mode") 0777)
```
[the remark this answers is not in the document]
pcarlisle: It's removing data that isn't just the user/group/other permissions by masking it out

barriserloth (approved):
LGTM from the skeletor side of things

src/clj/puppetlabs/services/master/file_serving.clj
pcarlisle: Yeah, I wrote the initial code before that thead and didn't revisit that issue. I think plain `project` is kind of iffy though so i'm not sure what's best.

mwaggett: `project-id-and-ref`? lol I'm not sure either. but that thread was sort of implying to me that `project` means id+ref.

Magisus: I don't think using just `project` makes sense here. Regardless, this is internal and so not exactly subject to the requirements being discussed in that thread. In the context of this code, do we ever deal with name and ref being separate? If not, I'm totally fine leaving this as `project-ref`, because a ref by itself is meaningless (we don't use it, so it's not confusing), but `project-ref` indicates you're passing more than just the name of the project.

Magisus: If we're going to use `project-ref` to indicate both things in one place though, we should use it everywhere to make it clear that's _always_ what it refers to.

mwaggett: ¯\_(ツ)_/¯ when I see `ref` I always think just git ref, but if y'all wanna leave it, that's fine. I just think we should be as consistent as possible, even if things are internal-only.

src/clj/puppetlabs/services/master/file_serving.clj
pcarlisle: so that the merging priority is correct

src/clj/puppetlabs/services/master/file_serving.clj
```diff
@@ -0,0 +1,364 @@
...
+
+(defn get-checksum
+  [file algorithm]
+  (case algorithm
+    "md5" (str "{md5}" (digest/md5 (io/file file)))
+    "sha256" (str "{sha256}" (digest/sha-256 (io/file file)))
+    :default (throw (Exception. "Unsupported digest"))))
+
+(defn read-attributes
+  [path checksum-type ignore-source-permissions]
```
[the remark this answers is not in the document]
pcarlisle: `source_permissions` is an option on file resources that indicates whether the client wants to receive the actual on disk permissions of the files or whether it gets a default set of permissions and uid/gid of the server process. It is deprecated, but still used for pluginsync (where it's generally set to `ignore` which gets converted to `true` in this representation). I have no idea why anyone ever wanted to use it.

src/clj/puppetlabs/services/master/file_serving.clj
```diff
@@ -0,0 +1,364 @@
...
+     :checksum (if (get attributes "isDirectory")
+                 {:type "ctime"
+                  :value (str "{ctime}" (get attributes "ctime"))}
+                 {:type checksum-type
+                  :value (get-checksum path checksum-type)})}))
+
+(defn get-file-metadata
+  [checksum-type ignore-source-permissions [relative-path root]]
+  (merge (read-attributes (.resolve root relative-path) checksum-type ignore-source-permissions)
+         {:path (.toString root)
```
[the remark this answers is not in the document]
pcarlisle: I don't think it matters. I'm explicitly using the java filesystem apis here so that's why this felt natural at the time.

src/clj/puppetlabs/services/master/file_serving.clj
pcarlisle: I abstracted the check into its own function.

pcarlisle:
should be ready, i will change the checksum algorithm for the 7 branch after merge up

Magisus (approved):

src/clj/puppetlabs/services/master/file_serving.clj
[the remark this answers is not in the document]
mwaggett: so we default to ignoring..? which uses the default set of permissions? That seems really weird to me.. Not saying you did something wrong, just trying to understand.

mwaggett (approved):

src/clj/puppetlabs/services/master/file_serving.clj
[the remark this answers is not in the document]
joshcooper: People surprisingly like being able to copy the file mode when managing a recursive directory resource (https://tickets.puppetlabs.com/browse/PUP-10253) Copying the owner and group is nonsensical but currently supported, and hopefully will be removed at some point. Puppet agents use `source_permissions => use` for the pluginfacts mount, because we need to preserve the executable-ness of external executable facts.

PR #2451, opened 2020-11-09
(maint) Remove docs, add note about migration

Magisus:
This also updates the docs links in the readme to point to published
docs.


README.md
```diff
@@ -94,10 +94,9 @@ issues](./documentation/known_issues.markdown) that you should be aware of.
 
 If want to play with our code, these documents should prove useful:
 
--   [Running Puppet Server from
-    source](./documentation/dev_running_from_source.markdown)
--   [Debugging](./documentation/dev_debugging.markdown)
--   [Puppet Server subcommands](./documentation/subcommands.markdown)
+-   [Running Puppet Server from source](https://puppet.com/docs/puppet/6.19/server/dev_running_from_source.html)
+-   [Debugging](https://puppet.com/docs/puppet/6.19/server/dev_debugging.html)
+-   [Puppet Server subcommands](https://puppet.com/docs/puppet/6.19/server/subcommands.html)
```
mwaggett, line 99: hmm I was gonna suggest we use `latest` in these urls instead of a hard-coded version, but I guess that should only apply to main. Just thinking that these could get out-of-date quickly.

mwaggett (commented):
There are several other places in the README that reference the `./documentation` directory.

README.md
Magisus: Ugh, I was going to use latest but it updated automatically with a redirect. I'll fix it.

justinstoller, line 99: should we use "latest" rather than direct versions?

Magisus: Yes.

Magisus:
Okay, see what you think of that. Did some other edits at the same time, to remove some of the out-of-date language.

mwaggett (approved):

PR #2452, opened 2020-11-10
(SERVER-2913) Use Jetty default ciphers

justinstoller:
In Server 6.5, Jetty began warning when using what it considered weak
cipher suites. Since then, we've also audited the cipher suites for
FIPS compliance to only support modern cipher suites.

In FOSS however, we've shipped with the older cipher suites that were
enabled at the start of the 6.x series.

With Puppet 7, remove the older ciphers and bring FOSS up to PE
standards.


justinstoller:
I should note: in PE the cipher suites can be managed by the puppet_enterprise module, but I believe we use the tk-jetty9 defaults as the PE defaults as of at least 2019.8.

Magisus:
So just to be clear, this change moves us to using Jetty's defaults, which are correct and match what we're doing in PE?

mwaggett (approved):

Magisus (approved):

PR #2457, opened 2020-11-12
(SERVER-2914) Change query param from `project` to `versioned_project`

justinstoller:
Note, the first commit isn't needed and I'm not sure it is an improvement. Happy to hear folks' thoughts on it.


Magisus (approved):

Magisus:
I like the first change, I don't think any of us _liked_ that variable name, we just didn't have a better idea.

PR #2460, opened 2020-12-02
(PE-30410) Shadow fact/variable collisions in the PAL for plans

ekinanp:
This takes advantage of the recently merged Puppet 7 PR (https://github.com/puppetlabs/puppet/pull/8412)

### DIFF
--- a/spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
+++ b/spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
@@ -25,9 +25,10 @@ def request(code)
    "variables" => {"values" => {"foo" => "bar"}}}
 end
 
-def find_notify(catalog)
+def find_notify(catalog, title: nil)
   catalog['resources'].find do |item|
-    item['type'] == 'Notify'
+    next item['type'] == 'Notify' unless title
+    item['type'] == 'Notify' && item['title'] == title
   end
 end
 
@@ -69,6 +70,70 @@ describe Puppet::Server::ASTCompiler do
       expect {
         Puppet::Server::ASTCompiler.compile(bolt_request, boltlib_path)
       }.to raise_error(Puppet::Error, /the path to boltlib modules must be provided/)
+    end
+
+    context 'when compiling for a plan with fact and plan/target variable collisions' do
+      let(:boltlib_path) { [] }
+
+      it 'properly shadows the relevant variables' do
+        # TODO: This is a temporary stub of the Bolt requires to get this
+        # test to pass. We should figure out how to properly load Bolt code
+        # later for the tests.
+        allow(Puppet::Server::ASTCompiler).to receive(:load_bolt)
+        mock_apply_inventory_class = double('Bolt::ApplyInventory')
+        allow(mock_apply_inventory_class).to receive(:new).with(anything).and_return(double('mock'))
+        stub_const("Bolt::ApplyInventory", mock_apply_inventory_class)
+
+        facts = {
+          "values" => {
+            "plan_var_fact"   => "fact_value",
+            "target_var_fact" => "fact_value",
+          }
+        }
+        variables = {
+          "values" => [
+            # should be shadowed by plan_var_fact fact
+            {"plan_var_fact"  => "plan_value"},
+            {"plan_var_plain" => "plan_value"},
+          ]
+        }
+        target_variables = {
+          "values" => [
+            # should be shadowed by target_var_fact fact
+            {"target_var_fact" => "target_value"},
+            # should be shadowed by plan_var_plain plan variable
+            {"plan_var_plain"  => "target_value"},
+          ]
+        }
+
+        code = <<CODE
+notify { "plan_var_fact": message => $plan_var_fact }
+notify { "target_var_fact": message => $target_var_fact }
+notify { "plan_var_plain":  message => $plan_var_plain }
+CODE
+        bolt_request = request(code)
+        bolt_request.merge!(
+          'facts'            => facts,
+          'variables'        => variables,
+          'target_variables' => target_variables,
+          'options'          => { 'compile_for_plan' => true },
+        )
+
+        response = Puppet::Server::ASTCompiler.compile(bolt_request, boltlib_path)
+        catalog = response[:catalog]
+
+        # assert the shadowed values
+        expected_collisions = {
+          'plan_var_fact'   => 'fact_value',
+          'target_var_fact' => 'fact_value',
+          'plan_var_plain'  => 'plan_value',
+        }
+        expected_collisions.each do |var, expected_value|
+          notify = find_notify(catalog, title: var)
+          expect(notify).not_to be_nil
+          expect(notify['parameters']['message']).to eql(expected_value)
+        end
+      end
     end
   end
 end
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -894,6 +894,7 @@
    (schema/required-key "trusted_facts") {(schema/required-key "values") {schema/Str schema/Any}}
    (schema/required-key "facts") {(schema/required-key "values") {schema/Str schema/Any}}
    (schema/required-key "variables") {(schema/required-key "values") (schema/either [{schema/Str schema/Any}] {schema/Str schema/Any})}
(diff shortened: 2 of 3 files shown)


src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -894,6 +894,7 @@
    (schema/required-key "trusted_facts") {(schema/required-key "values") {schema/Str schema/Any}}
    (schema/required-key "facts") {(schema/required-key "values") {schema/Str schema/Any}}
    (schema/required-key "variables") {(schema/required-key "values") (schema/either [{schema/Str schema/Any}] {schema/Str schema/Any})}
+   (schema/optional-key "target_variables") {(schema/required-key "values") [{schema/Str schema/Any}]}
```
ekinanp, line 897: Made this optional for backwards compatibility. Also chose the `Array[Hash]` syntax since that seems to be the preferred way of specifying target var values.

Note that the Orch-side changes are basically to pass-in the `target_variables` to the request. I _think_ it consists of eliminating https://github.com/puppetlabs/orchestrator/blob/main/src/puppetlabs/orchestrator/bolt/plan_actions.clj#L335 and letting the compile endpoint handle the shadowing (and it will only be done on the `main` branch).

ekinanp: I've tested that `target_variables` works as expected manually via `curl`.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -51,11 +51,23 @@ def self.compile_ast(code, compile_options, boltlib_path)
...
 
+          target_variables = if compile_options.key?('target_variables')
+                               compile_options['target_variables']['values'].each_with_object({}) do |param_hash, acc|
+                                 acc[param_hash.keys.first] = param_hash.values.first
+                               end
+                             else
+                               {}
+                             end
+
+          variables = {
```
ekinanp, line 62: Inspired by https://github.com/puppetlabs/bolt/pull/2277

spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
```diff
@@ -70,5 +75,70 @@ def find_notify(catalog)
...
+        target_variables = {
+          "values" => [
+            # should be shadowed by target_var_fact fact
+            {"target_var_fact" => "target_value"},
+            # should be shadowed by plan_var_plain plan variable
+            {"plan_var_plain"  => "target_value"},
+          ]
+        }
+
+        code = <<CODE
```
ekinanp, line 114: Based on https://github.com/puppetlabs/puppet/blob/main/lib/puppet/pal/pal_impl.rb#L516-L519, I think these cases should test everything.

src/clj/puppetlabs/services/master/master_core.clj
ekinanp: Where would I make the relevant doc updates for this?

spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
```diff
@@ -70,5 +75,70 @@ def find_notify(catalog)
...
     end
+
+    context 'when compiling for a plan with fact and plan/target variable collisions' do
+      let(:boltlib_path) { [] }
+
+      it 'properly shadows the relevant variables' do
+        # TODO: This is a temporary stub of the Bolt requires to get this
+        # test to pass. We should figure out how to properly load Bolt code
+        # later for the tests.
+        allow(Puppet::Server::ASTCompiler).to receive(:require).and_call_original
```
ekinanp, line 86: I know this sucks, but unless there's a _very_ good reason for us not to mock `require` in the tests, I'd rather not spend anymore time figuring out how to handle the Bolt requires in the AST compiler.

donoghuc:
I really hope (and it looks like it is the case) that this landed in puppet6 based on https://github.com/puppetlabs/puppet/pull/8412#pullrequestreview-515880018 

We will need this in puppet6 to take advantage of this in both our stream.

src/clj/puppetlabs/services/master/master_core.clj
donoghuc: https://github.com/puppetlabs/osp-docs/blob/b0af18f7c554097a54e36f22d0403c592fe48272/server/puppet-api/v3/compile.md

donoghuc:
Cool. so this did land in the 6.x stream it is just not in a released version yet. So this is fine.

src/clj/puppetlabs/services/master/master_core.clj
donoghuc: IMO the `plan_compile` specific options do not need to maintain backwards compatibility.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
donoghuc: This is fine. We could nitpick on the names but this is what bolt has done.

spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
donoghuc: I briefly looked at this. When i did found that it was actually a pretty small subset of all the bolt dependency gems that we actually would need. I think so far we have relied heavily on testing this in orchestrator acceptance tests.

spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
```diff
@@ -25,12 +25,17 @@ def request(code)
...
-def find_notify(catalog)
+def find_notify(catalog, title: nil)
   catalog['resources'].find do |item|
-    item['type'] == 'Notify'
+    next item['type'] == 'Notify' unless title
+    item['type'] == 'Notify' && item['title'] == title
   end
 end
 
+def collision_warning(msg_regex)
```
donoghuc, line 35: I dont see where this is used?

spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
ekinanp: Yeah, I'm going to add Orch acceptance tests for the shadowing if none currently exist. These tests at least are an easy way for us to detect a regression there if it happens so still useful to have even if the mocking currently sucks.

spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
ekinanp: This was from a previous version of the unit tests when I was looking through the logs. Removed.

mwaggett:
@ekinanp I think this should target 6.x, since the associated work is landing in the 2019.8.x stream.

spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
nicklewis: How about we pull out the `require`s into a single `require_bolt` method we can stub normally?

donoghuc: works for me. I'll make that update. good idea as usual @nicklewis :100:

spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
McdonaldSeanp: Would there be any reason to cover the case that a var is shadowed more than once (i.e. set in all three of target var/plan var/fact)

McdonaldSeanp: Would there be any reason to cover the case that a var is shadowed more than once (i.e. set in all three of target var/plan var/fact)

McdonaldSeanp (approved):
I do have one question, but it's a narrow test case so I still approve.

donoghuc (approved):

spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
donoghuc: No. we have that covered in orch and in bolt.

PR #2468, opened 2020-12-10
(SERVER-2917) Ensure old cadir symlink has same perms as actual cadir

mwaggett:


mwaggett:
okay this works

Magisus (approved):

PR #2472, opened 2020-12-11
(SERVER-2934) Emit warning when cadir is still in puppet agent's ssl dir

tvpartytonight:


src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -57,6 +57,7 @@ def initialize(puppet_config, puppet_server_config)
           "supports threadsafe settings. Please upgrade to Puppet 6.13.0 or greater.")
       end
     end
+    check_cadir_for_deprecation_warning
```
Magisus, line 60: So doing this here, this will run on every JRuby recycle. Is that what we want, as opposed to logging this in Clojure, say when the certificate authority starts?

Magisus:
If we do want this per-JRuby init, this looks good to me. I know we talked about the tradeoffs there before, but I don't remember what we decided.

spec/puppet-server-lib/puppet/jvm/master_spec.rb
```diff
@@ -56,4 +56,41 @@ def shutdown()
...
+  context "#check_ca_dir_for_deprecation_warning" do
+    before do
+      @master = Puppet::Server::Master.new({}, {})
+      @ssldir = Puppet[:ssldir]
+      @cadir = File.join(@ssldir, 'ca')
+    end
+
+    it "warns when there is a directory in puppet's ssl dir" do
+      expect(File).to receive(:exist?).with(@cadir).and_return(true)
+      expect(File).to receive(:symlink?).with(@cadir).and_return(false)
```
mwaggett, line 69: Should these be `allow`s instead of `expect`s? I'm rusty on my rspec, I think both might work, but when I see `expect`, I think "this is the thing being tested", as opposed to `allow`, which is used for mocking? right? 
Otherwise, how are these `expect`s different from the one we *actually* want to test?

mwaggett (commented):
I don't really have an opinion on where we do this logging. I guess I _thought_ we were just gonna do it on service start, but I'm not opposed to logging more often, to encourage migration..

spec/puppet-server-lib/puppet/jvm/master_spec.rb
tvpartytonight: The difference between `allow` and `expect` is that tests fail if the expectation is not met, while tests will pass if the allowance does not occur. I think it's clearer what the logic should be with `expect`, so that you aren't possibly introducing any side effects you didn't anticipate.

src/ruby/puppetserver-lib/puppet/server/master.rb
tvpartytonight: I feel like since the setting being deprecated is a Puppet setting, it made more sense to run the message from Puppet itself. I don't have a strong opinion either way, but it seemed weirder to me to warn in clojure about a setting that Puppet itself is removing.

tvpartytonight:
I don't believe there is any urgency to merging this for the release.

spec/puppet-server-lib/puppet/jvm/master_spec.rb
mwaggett: hmm. I guess then I'm confused about what is testing vs. what is just setting up the scenario we're trying to test, if they're all `expect`s. Like, if the above^ `expect`s are just mocking a directory, then how is the below `expect` not just also mocking? And if none of them are mocking, then how are we setting up this specific scenario to test?
Am I making any sense?

Magisus: The `and_return` is the part that does the mocking. Without it, it just tests that the thing is called and lets it return whatever it really returns. I'm fine with these being `expect`.

Magisus (approved):

PR #2476, opened 2020-12-18
(CPR-768) Update github workflow to build on pushes to 'main' branch

underscorgan:
I think the differences in the workflow between 6.x (the default branch)
and main (the branch we build from) was the source of the issues.

I was able to confirm on my fork that this change allowed builds to happen on push to main.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -3,7 +3,7 @@ name: Docker test and publish
 on:
   push:
     branches:
-      - master
+      - main
 
 jobs:
   build-and-publish:


.github/workflows/docker.yml
```diff
@@ -3,7 +3,7 @@ name: Docker test and publish
 on:
   push:
     branches:
-      - master
+      - main
```
mwaggett, line 6: Why not 6.x?

underscorgan: We only want to be building containers from the newest branch (main), but ghactions will use the action as defined in the repo's default branch (6.x)

mwaggett: ahh, gotcha

mwaggett (approved):

PR #2480, opened 2021-01-20
(PE-30968) Include the builtin bolt content's pluginsync

ekinanp:


ekinanp:
I have no idea why this PR's failing the Azure build. Seems like something's up with the Docker daemon there?

ekinanp:
Also, I don't think this needs any more testing than what's here since the integration tests provide good coverage. So this means, for example, that we probably don't need any new Orch-side acceptance tests.

ekinanp:
I'm also not sure what's causing the other, Docker container CI failure.

McdonaldSeanp (approved):

McdonaldSeanp:
Will merge pending tests

ekinanp:
Cool, looks like this is good to go now.

PR #2481, opened 2021-01-21
(maint) Bump jruby hocon to 1.3.1

Iristyle:
- Includes feature from 1.2.7 to load environment variables into lists
   like:

    FOO.1
    FOO.2

    See https://github.com/puppetlabs/ruby-hocon/pull/117

### DIFF
--- a/resources/ext/build-scripts/jruby-gem-list.txt
+++ b/resources/ext/build-scripts/jruby-gem-list.txt
@@ -1,5 +1,5 @@
 semantic_puppet 1.0.2
-hocon 1.2.5
+hocon 1.3.1
 text 1.3.1
 locale 2.1.2
 gettext 3.2.2
--- a/test/unit/puppetlabs/services/request_handler/request_handler_core_test.clj
+++ b/test/unit/puppetlabs/services/request_handler/request_handler_core_test.clj
@@ -266,8 +266,8 @@
   (testing "Improperly URL encoded content"
     (is (thrown+? [:kind :bad-request
                    :msg (str "Unable to URL decode the x-client-cert header: "
-                             "For input string: \"1%\"")]
-                  (jruby-request-with-client-cert-header "%1%2"))))
+                             "For input string: \"1Q\"")]
+                  (jruby-request-with-client-cert-header "%1Q%2"))))
   (testing "Bad certificate content"
     (is (thrown+? [:kind :bad-request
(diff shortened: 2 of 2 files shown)


MikaelSmith:
Oh, we do use ruby hocon here. Probably need both.

Iristyle:
I'm not sure all the places things will need to be updated, nor do I know yet where `puppetlabs/typesafe-config` is referenced

project.clj
```diff
@@ -27,7 +27,7 @@
 
   :min-lein-version "2.9.1"
 
-  :parent-project {:coords [puppetlabs/clj-parent "4.6.10"]
+  :parent-project {:coords [puppetlabs/clj-parent "4.6.11"]
```
Magisus, line 30: This update will be automatic when clj-parent is released. However, we will need to do the bump by hand in pe-pse.

Magisus:
This can be rebased now, the clj-parent bump is in 6.x.

project.clj
Iristyle: 👍 

PR is at https://github.com/puppetlabs/pe-puppet-server-extensions/pull/1262 

I'll update this PR once clj-parent ships and after this repo is auto-updated

mwaggett (approved):

Iristyle:
Looks like we have some failing specs

```
FAIL in (cert-decoding-failures) (request_handler_core_test.clj:267)
Improperly URL encoded content
expected: (thrown+? [:kind :bad-request :msg (str "Unable to URL decode the x-client-cert header: " "For input string: \"1%\"")] (jruby-request-with-client-cert-header "%1%2"))
  actual: "thrown+?: {:kind :bad-request, :msg \"No certs found in PEM read from x-client-cert\"} did not match [:kind :bad-request :msg (str \"Unable to URL decode the x-client-cert header: \" \"For input string: \\\"1%\\\"\")]"
```

Iristyle:
Yeah, so the clj-parent 4.6.10 to 4.6.11 bump didn't pass at https://github.com/puppetlabs/puppetserver/runs/1743672108 - 41f70ff addresses that

Magisus:
For reference, the commit that broke us is https://github.com/ring-clojure/ring-codec/commit/780088590b57a0caf2d64f378f8ab852547a698e

mwaggett (approved):

PR #2483, opened 2021-01-26
(maint) Switch to using `server` instead of `master` puppet.conf section

Magisus:
This commit updates the settings section we load from `master` to
`server` to avoid deprecation warnings. Puppet will make sure both
sections are always loaded, but specifying `server` instead of `master`
avoids the warning.

### DIFF
--- a/src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
@@ -86,7 +86,7 @@ class Puppet::Server::PuppetConfig
       Puppet.push_context(dummy_ssl_context)
     end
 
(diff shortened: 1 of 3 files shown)


src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
```diff
@@ -86,7 +86,7 @@ def self.initialize_puppet(puppet_config:)
       Puppet.push_context(dummy_ssl_context)
     end
 
-    Puppet.settings.use :main, :master, :ssl, :metrics
+    Puppet.settings.use :main, :server, :ssl, :metrics
```
Magisus, line 89: The warning was specifically coming from here, but I updated the two fixture files as well for good measure.

mwaggett (approved):

PR #2484, opened 2021-01-27
(maint) Remove unused metrics helpers

Magisus:
These helpers are leftover from before all the metrics testing moved to
the PE suite. They are duplicated in the helpers there. However, due to
Ruby loadpath shenanigans, they were still getting used sometimes
instead of the versions in PE. This is misleading and unnecessary, so
this commit removes the unused FOSS versions.


Magisus:
I tested that we don't need them in for the PE tests (removed them from the submodule), and a grep doesn't reveal any uses of them in the FOSS suite anymore.

justinstoller (approved):

PR #2488, opened 2021-01-28
(SERVER-2948) Treat ASCII-8BIT body as bytes

pcarlisle:
Puppet returns binary data for some endpoints as a string encoded with
Ruby's ASCII-8BIT encoding. This encoding is unique to ruby and is
essentially meant to be treated as raw bytes. As of version 1.6.0,
ring-servlet attempts to encode the response body when it's a string
using the charset in the header, falling back to UTF-8. When we
encounter this encoding now we will convert it to a java byte array and
ring won't attempt to encode it.

### DIFF
--- a/src/ruby/puppetserver-lib/puppet/server/master.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/master.rb
@@ -69,10 +69,16 @@ class Puppet::Server::Master
 
     body = response[:body]
     body_to_return =
-        if body.is_a? String or body.nil?
-          body
+        if body.is_a? String
+          if body.encoding == Encoding::ASCII_8BIT
+            body.to_java_bytes
+          else
+            body
+          end
         elsif body.is_a? IO
           body.to_inputstream
+        elsif body.nil?
+          body
         else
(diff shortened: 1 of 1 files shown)


pcarlisle:
note that this doesn't put the new version of ring back

Magisus (approved):

mwaggett (approved):
Tested this with clj-parent 4.6.14 (and a host config update I'll make a separate PR for) by running the following:
```
bundle install && PACKAGE_BUILD_VERSION=6.15.1.SNAPSHOT.2021.01.28T1533 bundle exec beaker --xml --debug --repo-proxy --hosts acceptance/config/beaker/jenkins/redhat7-64m-64a.cfg --type aio --keyfile ~/.ssh/id_rsa-acceptance --tests acceptance/suites/compat_tests,acceptance/suites/puppet5_tests --helper acceptance/lib/helper.rb --options-file acceptance/config/beaker/options.rb --post-suite acceptance/suites/post_suite --pre-suite acceptance/suites/pre_suite/compat_common,acceptance/suites/pre_suite/puppet5_compat --tag server --load-path acceptance/lib
```

PR #2489, opened 2021-01-29
(maint) Update acceptance testing docs

mwaggett:
This commit updates the acceptance README to be prescriptive about how to run
beaker tests locally. This also removes old documentation and host configs that
are no longer necessary when using the handy shell script to test.


acceptance/config/beaker/jenkins/el-6-x86_64-mdca-a.cfg
```diff
@@ -1,25 +1,21 @@
 HOSTS:
-  master:
+  primary:
```
justinstoller, line 2: nice

mwaggett: I assumed that these were arbitrary names and thus changeable, as opposed to the actual roles we set below.

tvpartytonight: Do we actually use these anywhere? I assumed everything is a BHG string for our ci...

justinstoller:
fwiw, I use `acceptance/scripts/generic/testrun.sh` when I run locally and that use beaker hostgenerator. We may want to consolidate how we run these tests locally.

Magisus:
I also use the `testrun` script. I recently(ish) updated it to use beaker subcommands.

mwaggett:
I don't think Jenkins does, since they were still using vcloud. I was using these when I was reproducing the ring failures.
If there are other ways to run tests, we should really update our developer docs. I also struggled to figure out how to test things via the repl.

Magisus:
There is a really long doc about acceptance testing. The script Justin and I are talking about it is described here https://github.com/puppetlabs/puppetserver/tree/6.x/acceptance#using-the-handy-shell-script

mwaggett:
oh right, I did read that, but I wanted to run specific tests. I didn't actually look at the test script to see that you could override the suites. I thought it just ran 'em all or something.

Magisus:
No it's really customizable.

mwaggett:
fwiw, I don't find "really long" docs useful. I think we should remove everything about other ways to run the tests if the test script is what's recommended.

I guess I can update this PR and remove these configs altogether if we think that's the right direction.

Magisus:
It looks like Jenkins uses a [generic script](https://github.com/puppetlabs/ci-job-configs/blob/master/resources/scripts/integration-beaker.sh) that leverages many of these same env vars and uses beaker subcommands.

mwaggett:
okay let me know what y'all think of this update

Magisus (approved):
I think this looks good! Thanks for cleaning this up.

Magisus (approved):

PR #2491, opened 2021-02-04
(SERVER-2338) Always sign agent certs with CN as SAN

tvpartytonight:


src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1187,8 +1205,10 @@
                          #{:key-encipherment
                            :digital-signature} true)
                        (utils/subject-key-identifier
-                         subj-pub-key false)]]
-    (concat base-ext-list csr-ext-list)))
+                        subj-pub-key false)]
+        subject (get-csr-subject csr)
+        modded-csr-list (ensure-ext-list-has-cn-san subject csr-ext-list)]
```
tvpartytonight, line 1210: I would ideally like to pass `(concat csr-ext-list base-ext-list)` to `ensure-ext-list-has-cn-san`, but for reasons I don't quite understand, those two lists are `clojure.lang.PersistentVector`, but when `concat`'d, it becomes a `clojure.lang.LazySeq` and is incompatible with the function, generating the error:


```
Internal Server Error: java.lang.ClassCastException: clojure.lang.LazySeq cannot be cast to clojure.lang.Associative
```

justinstoller: I think this is from our usage of `update-in` in `ensure-ext-list-has-cn-san` which isn't defined on a list, which is what concat returns.

I think alternative ways of doing it would be to call `vec` on the list, either before or calling or inside of `ensure-ext-list-has-cn-san`, call `(apply conj ...)` instead of `concat`, or reworking `ensure-ext-list-has-cn-san` to not use `update-in`.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1168,6 +1168,31 @@
...
+  [extension]
+  (= utils/subject-alt-name-oid (:oid extension)))
+
+(schema/defn ensure-ext-list-has-cn-san
+  "Given a list of extensions to be signed onto a certificate, ensure that a CN is provided
+   as a subject alternative name; if no subject alternative name is found, generate a new
+   extension and add it to the list with the CN supplied"
+  [cn :- schema/Str
+   extensions :- (schema/pred utils/extension-list?)]
+  (let [[san] (filter is-san? extensions)
```
mwaggett, line 1189: Why is this wrapped in an extra set of brackets?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1187,8 +1212,10 @@
                          #{:key-encipherment
                            :digital-signature} true)
                        (utils/subject-key-identifier
-                         subj-pub-key false)]]
-    (concat base-ext-list csr-ext-list)))
+                        subj-pub-key false)]
+        subject (get-csr-subject csr)
+        combined-list (vec (concat base-ext-list csr-ext-list))]
```
mwaggett, line 1217: Why does this need to be wrapped in `vec` now?

mwaggett (commented):
Should these new functions have unit tests?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
mwaggett: oh, I see your convo with Justin now

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1168,6 +1168,31 @@
          (whitelist-matches? autosign subject))
        false))))
 
+(schema/defn ensure-cn-as-san :- utils/SSLExtension
```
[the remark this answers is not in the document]
Magisus: Yes. I feel like they are more common than the spelled-out versions in the cert world actually, especially given how long "subject alternative name" is 😛

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1168,6 +1168,31 @@
...
+  "Given an SSLExtension for alt names and a cn, ensure that the cn is listed as a DNS alt name."
```
Magisus, line 1172: However, I might write out "common name" here in the comment, or at least capitalize the acronym.

mwaggett: see, I didn't even know it was "common name". I was gonna guess "certificate name". 😝

src/clj/puppetlabs/puppetserver/certificate_authority.clj
tvpartytonight: This is the equivalent of this: ```(let [san (first (filter is-san? extensions))]```

The ```[san]``` just tells it to return the first element of that sequence returned by the filter.

mwaggett: oooh interesting! Thanks for explaining!

tvpartytonight:
I've added additional unit tests for when a csr has the cn in the SAN and for when CSR's SAN doesn't contain the cn in the dns-list. I think I've addressed all feedback, let me know if I missed something.

Magisus (approved):

Magisus:
Looks like some of the new tests are failing, but I think the changeset looks right.

mwaggett (approved):

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
```diff
@@ -203,9 +203,7 @@
           subjectDN (ssl-utils/cn "test_cert_ca_true")
           serial 1
           public-key (ssl-utils/get-public-key key-pair)
-          ca-ext (ca/create-ca-extensions subjectDN
-                                          serial
-                                          public-key)
+          ca-ext [(ssl-utils/basic-constraints-for-ca)]
```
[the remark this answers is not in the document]
tvpartytonight: Because `create-ca-extensions` now has a SAN extension attached, and that check with `allow-subject-alt-names` get's triggered before the check for the oid—2.5.29.19—for certificate signing.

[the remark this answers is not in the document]
mwaggett: so you're saying that without this change, this test would fail because we specifically _don't_ allow subject alt names here? Still trying to make sense of what exactly this is testing.. 😅 

also thanks for letting me use your PRs to learn more about the various things we're responsible for 🙃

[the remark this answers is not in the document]
tvpartytonight: yeah, remember how I made a mention in our team chat about how
> TIL the dev/puppetserver.conf affects `lein test`

There is a CA setting that is false for `allow-subject-alt-names` in that test conf file. I had turned it one to allow for additional testing in my local env, so I didn't see the errors that CI caught.

Because the test was using the `create-ca-extensions`, the way the list is formed is modified because we now ensure the SAN extension is there; that's what caused it fail with the same 409 status response code but a differnt body. The test always should have been more targeted to use a single extension, to avoid this exact situation.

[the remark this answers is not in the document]
mwaggett: gotcha, thanks for explaining! Is there a reason we don't just set `allow-subject-alt-names: true` in the test conf? I don't have a sense of when someone would or would not want to allow SANs, but if we've added this functionality, it seems like most people must have `allow-subject-alt-names: true`?

PR #2493, opened 2021-02-12
(PE-31161) Update get-project-modulepath with Bolt 3 conventions

ekinanp:
In Bolt 3, the project's modulepath is calculated as
  modulepath = (config['modulepath'] || 'modules') + '.modules'

This commit updates Puppetserver to also do that calculation so that
file content and plugins can still work with Bolt 3.


McdonaldSeanp (approved):

src/clj/puppetlabs/services/master/file_serving.clj
```diff
@@ -168,14 +165,13 @@
...
-      (if has-modules-key?
-        ["modules" ".modules"]
-        default-project-modulepath))))
+  ;; Note that the :modulepath config key's useful when the project
+  ;; organizes its modules inside relative directories. It doesn't
+  ;; make sense for absolute paths since those may not exist
+  ;; on the Puppetserver host.
+  (-> (get project-config :modulepath "modules")
+      (parse-modulepath)
+      (concat [".modules"])))
```
donoghuc, line 174: Does it matter that we no longer return a vector?
```
user=> (concat (clojure.string/split "foo" #":") ["foo"])
("foo" "foo")
user=> (reduce conj (clojure.string/split "foo" #":") ["foo"])
["foo" "foo"]
```

ekinanp: I don't think so, because Line 174 in the previous code reduced to `(concat (parse-modulepath (get project-config :modulepath)) [".modules"])`, which is symmetric with the code here. So, the previous version had the possibility of returning both a seq and a vector, meaning that the distinction between seq and vector probably did not matter in the first place (as long as the return value was a "Clojure array of strings", which it should be with the new code).

ekinanp: Actually, a more simpler summary of ^ is that the new code is Line 174 in the old code, so the return value's still fine.

donoghuc: Makes sense.

donoghuc (approved):

Magisus (approved):
So the implication of this is that `site` and `site-modules` are no longer supported by default?

donoghuc:
correct

PR #2495, opened 2021-02-22
(SERVER-2558) Use puppet's copy of CRL as jetty CRL

Magisus:
This commit changes the default CRL path supplied to Jetty from the CA's
to the local server's copy. This latter path, in the ssldir rather than
the CA, is present even if the CA service is disabled, because the
puppet agent will write the CRL it fetches to that location, while the
CA's copy is not always available.

This mirrors what we configure in PE, and makes setting up compilers in
FOSS easier.

### DIFF
--- a/src/clj/puppetlabs/services/config/puppet_server_config_core.clj
+++ b/src/clj/puppetlabs/services/config/puppet_server_config_core.clj
@@ -109,13 +109,13 @@
      (assoc config :puppet-version (.puppetVersion jruby-puppet)))))
 
 (defn init-webserver!
-  "Initialize Jetty with paths to the master's SSL certs."
+  "Initialize Jetty with paths to puppetserver's SSL certs."
   [override-webserver-settings! webserver-settings puppet-config]
-  (let [{:keys [hostcert localcacert cacrl hostprivkey]} puppet-config
+  (let [{:keys [hostcert localcacert hostcrl hostprivkey]} puppet-config
         overrides {:ssl-cert     hostcert
                    :ssl-key      hostprivkey
                    :ssl-ca-cert  localcacert
-                   :ssl-crl-path cacrl}]
+                   :ssl-crl-path hostcrl}]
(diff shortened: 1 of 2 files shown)


mwaggett (approved):

mwaggett:
this wouldn't be a breaking change, right?

Magisus:
I don't think so, seeing as this file has existed more consistently even than the other one. It also did not move in the CA migration. But the upgrade scenario would also be my main concern about it.

Magisus:
@justinstoller @Sharpie does this seem safe/sane to you?

justinstoller (approved):
This seems _sane_, but iiuc could break if a user runs the CA but not an agent... Could that happen in a containerized environment? If so we should reach out to the dockerizing folks and definitely make sure to highlight this in release notes.

Magisus:
Maybe I should do this as a fallback after all just to be safe.

Magisus:
So I think you're right that this config would be invalid in the situation where you're running the CA service without the Master service (which [initializes](https://github.com/puppetlabs/puppetserver/blob/710645789f9d23c99a7e2eb12ece3b77d894ee4c/src/clj/puppetlabs/services/master/master_service.clj#L155-L157) the master SSL files). However, that's also not currently a use case we have. All of the existing containers just load up the whole package/uberjar. I would rather optimize for making it easier to FOSS users to spin up compilers, which is a documented workflow, than try to account for future containerized setups, where we already have a lot of special config. It seems reasonable in that case to have the containers configure this explicitly to point at the CA CRL. Does that seem reasonable?

Side note, I don't think it's really valid to do a fallback, because we initialize this service before any of the files are guaranteed to be created, so we can't check for existence to choose one.

Magisus:
I also realized that the other files in this overrides block would also not exist in the situation we were thinking of. So this whole thing will be invalid in that setup, and need special thought. This change brings the CRL in line with the paths of the other artifacts.

PR #2496, opened 2021-02-22
(SERVER-2321) Refer to ca.conf in allow-* CA error messages

Magisus:
Previously, the error messages issued when auth extensions or SANs were
disallowed referenced the `puppetserver.conf` file. But these settings
live in the `ca.conf` file in both PE and FOSS now. This commit updates
the messages to refer to that file instead.


justinstoller (approved):

justinstoller:
needs a test update

mwaggett (approved):

PR #2497, opened 2021-02-23
(SERVER-2965) Better environment handling for v4 catalogs

justinstoller:
Previously, the v4 catalong endpoint had these issues with environment
handling:
1. Would not use the environment in the body of the request for code_id
generation.
2. Did not validate environment names.

The issue was needed for feature parity with Puppet while the
first issue caused spurious warnings to be issued.

### DIFF
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -25,14 +25,14 @@
             [ring.util.response :as rr]
             [schema.core :as schema]
             [slingshot.slingshot :refer [throw+ try+]])
-  (:import clojure.lang.IFn
-           [com.codahale.metrics Gauge MetricRegistry]
-           com.fasterxml.jackson.core.JsonParseException
-           java.io.FileInputStream
-           java.lang.management.ManagementFactory
-           [java.util List Map Map$Entry]
-           org.jruby.exceptions.RaiseException
-           org.jruby.RubySymbol))
+  (:import (clojure.lang IFn)
+           (com.codahale.metrics Gauge MetricRegistry)
+           (com.fasterxml.jackson.core JsonParseException)
+           (java.io FileInputStream)
+           (java.lang.management ManagementFactory)
+           (java.util List Map Map$Entry)
+           (org.jruby.exceptions RaiseException)
+           (org.jruby RubySymbol)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Constants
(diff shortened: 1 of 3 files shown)


Magisus (approved):
Typo in commit message though, missing the word "second" in the last sentence. I'm fine ignoring this though in the interest of not having to wait more on Travis >.>

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -25,14 +25,14 @@
...
             [schema.core :as schema]
             [slingshot.slingshot :refer [throw+ try+]])
-  (:import clojure.lang.IFn
-           [com.codahale.metrics Gauge MetricRegistry]
-           com.fasterxml.jackson.core.JsonParseException
-           java.io.FileInputStream
-           java.lang.management.ManagementFactory
-           [java.util List Map Map$Entry]
-           org.jruby.exceptions.RaiseException
-           org.jruby.RubySymbol))
```
Magisus, line 35: Was this old-style or something? Surprised this works both ways.

Magisus (approved):

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -874,11 +874,15 @@
          :headers {"Content-Type" "application/octet-stream"}
          :body (get-code-content environment code-id file-path)}))))
 
+(defn valid-env-name?
+  [string]
+  (re-matches #"\w+" string))
```
joshcooper, line 879: I don't know if this matters, but the ruby version of this uses `\A` and `\Z` to prevent against multiline inputs: https://github.com/puppetlabs/puppet/blob/c6412ea369aa1613ca166b10fb819fcf6130814b/lib/puppet/node/environment.rb#L127 Does this version do the same?

justinstoller: My understanding is that Java regexes are anchored from start of string to end of string, newlines inclusive, by default, lemme double check in my repl.

annotated test in repl:
```
dev-tools=> (def wlines "foo\nfoo")
#'dev-tools/wlines

;; prn, and most of clojure string processing doesn't do anything special with these characters
dev-tools=> (prn wlines)
"foo\nfoo"
nil

;; but split-lines recognizes them
dev-tools=> (clojure.string/split-lines wlines)
["foo" "foo"]

;; regexes are processed across the whole string irrespective of special characters
dev-tools=> (re-matches #"\w+" "fo\nfo")
nil

dev-tools=> (re-matches #"\w+" "fofo")
"fofo"
```

src/clj/puppetlabs/services/master/master_core.clj
pcarlisle: https://github.com/bbatsov/clojure-style-guide#comprehensive-ns-form

It's not explicit, but all the examples here show it the way it was. Is this change editor managed or is it intentional? If I tell my editor to manage the ns declaration for me it'll put it back to the way it was.

pcarlisle: the ns form accepts a million different things and i don't think this is actually important, just curious about it

justinstoller: It was actually intentional. I originally had an implementation that required a new import and when adding the new import made the section look like https://stuartsierra.com/2016/clojure-how-to-ns.html which is the style I thought we were doing for nses. I meant to at least split this out into its own commit but got lazy. fwiw, I'm happy to put it back.

pcarlisle: I have no opinion on what style is better or if it even matters to have a consistent one here, I just do the other because it's what emacs does when i tell it to automatically import the things i'm using or to clean up the namespace.

PR #2499, opened 2021-03-01
(SERVER-1061) Remove pre-suite step to redirect dujour ping

Magisus:
Beaker has had a built-in option to configure `/etc/hosts` to redirect
update.puppetlabs.com to localhost for a long time, and it is enabled by
default. This commit removes our duplication of this configuration.


Magisus:
I ran the pre-suite without this step and verified that `/etc/hosts` is still getting updated to do this redirect.

mwaggett (approved):

PR #2500, opened 2021-03-04
(SERVER-2870) Update Puppet APIs that use master

justinstoller:
This work was to update Puppet Server to ensure all the modern
non-harmful terminology APIs were being used. However, some terminology
in Puppet around routing was missed in the 7.0 release. This changes as
much terminology as possible while commenting as to why the rest hasn't
changed.


mwaggett (approved):

mwaggett:
Travis is failing because of the facter issue. Would it be worth manually updating to a good version of facter, even though it might get put back to the bad one over night?

justinstoller:
I don't think so, it's not a pressing ticket.

Magisus:
The facter pin was updated, you should be able to rebase this and get the correct SHA.

PR #2501, opened 2021-03-05
(SERVER-2867) Make server-* configs preferred to master-* variants

justinstoller:
This creates new server-* variants of the existing master-* config
names (master-conf-dir, master-code-dir, etc). When resolving values
we will prefer the new setting, the old setting, then the default.

This updates the default config files shipped with Puppet Server to use
the new setting names. It also updates the variable names containing the
default values in code. This updates the tests to use the new names,
changes the name of the variables that hold their test values, and in
select cases changes the names of the test directories from `master` to
`server` (only for the autogenerated test directiories, not the
fixtures).

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
@@ -10,19 +10,19 @@
 
   The keys should have the following values:
 
-    * :master-conf-dir - file path to puppetmaster's conf dir;
+    * :server-conf-dir - file path to Puppet Server's conf dir;
         if not specified, will use the puppet default.
 
-    * :master-code-dir - file path to puppetmaster's code dir;
+    * :server-code-dir - file path to Puppet Server's code dir;
         if not specified, will use the puppet default.
 
-    * :master-var-dir - path to the puppetmaster's var dir;
+    * :server-var-dir - path to the Puppet Server's var dir;
         if not specified, will use the puppet default.
 
-    * :master-run-dir - path to the puppetmaster's run dir;
+    * :server-run-dir - path to the Puppet Server's run dir;
         if not specified, will use the puppet default.
 
-    * :master-log-dir - path to the puppetmaster's log dir;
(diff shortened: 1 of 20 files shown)


src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -67,5 +67,12 @@
...
    :max-requests-per-instance schema/Int
-   (schema/optional-key :boltlib-path) [schema/Str]})
+   (schema/optional-key :boltlib-path) [schema/Str]
+   ;; Deprecated in favor of their `server-*` counterparts.
+   ;; to be removed in Puppet 8.
+   (schema/optional-key :master-conf-dir) (schema/maybe schema/Str)
+   (schema/optional-key :master-code-dir) (schema/maybe schema/Str)
+   (schema/optional-key :master-var-dir) (schema/maybe schema/Str)
+   (schema/optional-key :master-run-dir) (schema/maybe schema/Str)
+   (schema/optional-key :master-log-dir) (schema/maybe schema/Str)})
```
Magisus, line 77: I guess we lose a little bit here, in that it seems that now we'd allow neither variant to be specified, but I'm not really worried about it.

Magisus (approved):
Looks great!

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
mwaggett: What would happen if neither were specified?

Magisus: I don't think it's technically possible unless we messed something up in code, since we fill in defaults for everything.

mwaggett: ahkay, so a user doesn't have to worry about making sure one of these is specified? then yeah, I'm not really worried about it either.

mwaggett (approved):

PR #2505, opened 2021-03-11
(maint) Handle new CA path

justinstoller:


justinstoller:
This lets the docker entry-point understand the difference between the new and old CA dirs. I believe by exposing the CADIR as an env var it will also be configurable(??) I think in addition to this pupperware should be updated to have a puppet-certs volume label and the docs should be updated to call out the Puppet 7 change and importance of mounting place to keep your certs.

However, this is my first change to the repo and I don't use it much. I'm not sure the best way to test this change or it in concert with any commensurate pupperware changes. Hoping to get some advice from @Iristyle on that.

docker/puppetserver/docker-entrypoint.d/80-ca.sh
```diff
@@ -43,13 +43,15 @@ else
...
-  # Generate the same SSL directory that the PE installer creates.
-  #
-  # The steps in this file correspond to the steps in the PE installer.
-  #
-  # See https://github.com/puppetlabs/puppet-enterprise-modules/blob/kearney/modules/pe_install/manifests/prepare/certificates.pp
+  [ -d "$CADIR" ] && [ -f "$CADIR/ca_crt.pem" ]
+  new_ca_exists=$?
+
+  [ -d "$SSLDIR" ] && [ -f "$SSLDIR/ca_crt.pem" ]
+  old_ca_exists=$?
```
Magisus, line 50: These variable names don't match what you have below, `new_ca_exists` vs `new_ca_dir`

justinstoller: :lol:

justinstoller: Okay, I had some pretty basic mistakes where I started the variables updates in this script and then iterated in my shell with something similar and then came back to this script but things didn't line up. I think I've addressed them. Though it does beg the question of how best to test docker images.

docker/puppetserver/docker-entrypoint.d/80-ca.sh
```diff
@@ -60,11 +62,15 @@ else
...
+      ca_name="Puppet CA generated on ${HOSTNAME} at $timestamp"
 
       # See puppet.conf file for relevant settings
       puppetserver ca setup \
           --ca-name "$ca_name" \
           --config /etc/puppetlabs/puppet/puppet.conf
+
+  elif [ $new_ca_exists -eq 1 ] && [ $old_ca_exists -eq 0 ]; then
+      # Legacy CA upgrade
+      puppetserver ca migrate
```
mwaggett, line 74: I know we recently talked about how you don't need the `--config` flag to run this, but I noticed that we do pass the `--config` flag for the `setup` command. Is the `setup` command just being super explicit, or are we doing some custom thing that means we should pass it here too?

justinstoller: The `setup` command is, afaict, being super explicit. However, that branch also writes to that config file so it seemed fine to leave it explicit. We don't touch the config as part of the migration and it seemed fine to leave it implicit there. But I'm happy either removing the setup's `--config` flag usage or adding it here.

justinstoller:
Love to get a review from @Iristyle

docker/puppetserver/Dockerfile
```diff
@@ -24,6 +24,7 @@ ENV PUPPERWARE_ANALYTICS_TRACKING_ID="UA-132486246-4" \
     PUPPETSERVER_JAVA_ARGS="-Xms512m -Xmx512m" \
     PATH=/opt/puppetlabs/server/bin:/opt/puppetlabs/puppet/bin:/opt/puppetlabs/bin:$PATH \
     SSLDIR=/etc/puppetlabs/puppet/ssl \
+    CADIR=/etc/puppetlabs/puppetserver/ca \
```
Iristyle, line 27: Do we expect users to change this here?

If not, we've been trying to move `ENV` vars out of the `Dockerfile` because it signals something that can / should be touched by end users.  Instead, better to just embed it in the script where it's used (in this case maybe do a `puppet config print` to retrieve it?)

Puppetserver is the one container that's a bit out of date with the "best practices" we've been trying to put in everywhere -- partially because its the first container, and partially b/c there's a lot of practices to unwind that people may be relying on b/c we have a lot of active users.

docker/puppetserver/docker-entrypoint.d/80-ca.sh
Iristyle: To test just do `make lint build test`

docker/puppetserver/docker-entrypoint.d/80-ca.sh
```diff
@@ -43,13 +43,15 @@ else
   hocon -f /etc/puppetlabs/puppetserver/conf.d/ca.conf \
     set certificate-authority.allow-subject-alt-names "${CA_ALLOW_SUBJECT_ALT_NAMES}"
 
-  # Generate the same SSL directory that the PE installer creates.
-  #
-  # The steps in this file correspond to the steps in the PE installer.
-  #
-  # See https://github.com/puppetlabs/puppet-enterprise-modules/blob/kearney/modules/pe_install/manifests/prepare/certificates.pp
+  [ -d "$CADIR" ] && [ -f "$CADIR/ca_crt.pem" ]
```
Iristyle, line 46: I don't think the file test will ever pass if the directory isn't there, so do we need the `-d`?

Iristyle: Also, I think these variables have slightly confusing names since they're named opposite of what I would expect. Maybe do something like this to flip the exit code:

```shell
new_ca_exists=$([ -f "$CADIR/ca_crt.pem" ] && echo 1)
old_ca_exists=$([ -f "$SSLDIR/ca/ca_crt.pem" ] && echo 1)
```

So that then you can use the the `-n` (not null) and `-z` (null) tests in shell code like:

```shell
# no existing CA
if [ -z "$new_ca_exists" ] && [ -z "$old_ca_exists" ]; then
# legacy CA upgrade
elif [ -z "$new_ca_exists" ] && [ -n "$old_ca_exists" ]; then
fi
```

docker/puppetserver/docker-entrypoint.d/80-ca.sh
justinstoller: Is this expected:
```
latest: Pulling from hadolint/hadolint
Digest: sha256:d6ee991347e4e0bd38e12b6bf915ac01c757e92e7da52c7a51a39e723ee7f68d
Status: Downloaded newer image for ghcr.io/hadolint/hadolint:latest
ghcr.io/hadolint/hadolint:latest
hadolint: Dockerfile: openBinaryFile: inappropriate type (is a directory)
make: *** [Makefile:51: lint] Error 1
```

justinstoller: Weird:
```
justin@sellout:~/backups/Backup/code/work/server/puppetserver/docker$ sudo docker run --rm -v $(pwd)/puppetserver/Dockerfile:/Dockerfile -i ghcr.io/hadolint/hadolint:latest hadolint Dockerfile
justin@sellout:~/backups/Backup/code/work/server/puppetserver/docker$ echo $?
0
justin@sellout:~/backups/Backup/code/work/server/puppetserver/docker$ sudo make lint
latest: Pulling from hadolint/hadolint
Digest: sha256:d6ee991347e4e0bd38e12b6bf915ac01c757e92e7da52c7a51a39e723ee7f68d
Status: Image is up to date for ghcr.io/hadolint/hadolint:latest
ghcr.io/hadolint/hadolint:latest
hadolint: Dockerfile: openBinaryFile: inappropriate type (is a directory)
make: *** [Makefile:51: lint] Error 1
```

It looks like it gets past lint in Travis though and then builds the container, spams with a bunch of docker help output and then fails the test portion of that command.

justinstoller:
Sorry for the premature squash! I adopted all of the recommendations that came from review.

This PR will require this Pupperware PR first: https://github.com/puppetlabs/pupperware/pull/272

One can test this by pulling down Pupperware and applying the above PR.
* Export the `PUPPETSERVER_IMAGE` as something from 6.x. I used: `export PUPPETSERVER_IMAGE=puppet/puppetserver:6.14.1`.
* Export `PUPPET_DNS_ALT_NAMES=` your host name.
* Then spin up a Pupperware cluster using docker-compose.
* Have a local agent check into the Puppetware cluster (from the puppet source repo): `bundle exec puppet agent -t --server <your host name>` (you may need to delete `~/.puppetlabs` for this to work).
* Validate that your agent's cert is in `~/.local/share/docker/volumes/compose-services_puppetserver-config/_data/ssl/ca/signed/` and that `~/.local/share/docker/volumes/compose-services_puppetserver-config2/_data/ca` does not exist.
* Leave your cluster up, and build this PR and give it an unambiguous version.
* Set the `PUPPETSERVER_IMAGE` variable to the new image you've created and run `docker-compose up -d`. This should trigger the new image to migrate the CA.
* Check your local agent back into the server container and validate that it does not re-request a cert.
* Validate that `~/.local/share/docker/volumes/compose-services_puppetserver-config/_data/ssl/ca/` is now a symlink and that `~/.local/share/docker/volumes/compose-services_puppetserver-config2/_data/ca` contains the contents that were in the old ca dir.

Magisus (approved):

PR #2510, opened 2021-03-24
(maint) Update CA CLI gem to 1.9.3

Magisus:
This update ensures that ca-client certs will also be signed with the CN
as a SAN.

### DIFF
--- a/resources/ext/build-scripts/mri-gem-list-no-dependencies.txt
+++ b/resources/ext/build-scripts/mri-gem-list-no-dependencies.txt
@@ -1,1 +1,1 @@
-puppetserver-ca 1.9.1
+puppetserver-ca 1.9.3


mwaggett:
Do you want to do another 1.x release with your memoization fix?

Magisus:
Yeah good idea.

Magisus:
updated

mwaggett (approved):
Is the docker failure a known thing?

Magisus:
I'm not sure... do you know how to get more info from the output there?

mwaggett:
I do not, other than maybe by running it locally?

But we can leave that for a separate effort. I'll merge this.

Iristyle:
#2513 has the fix for the Docker build fail in Travis... there's something buggy in their version of Docker which causes those intermittent failures.

Iristyle:
Actually decided to land it separately in #2514 b/c I think the `Dockerfile` needs a little rework to support being built without buildkit.

PR #2513, opened 2021-03-29
(maint) Docker consistency updates

Iristyle:
- Bump docker-compose to 1.28.6 (necessary for moving away from deprecated -no-ansi flag)
- Pull hadolint from gchr.io to avoid rate-limiting on Docker Hub
- Set +x on entrypoints at build time to be a better non-root citizen in k8s (more non-root work required on this container)
- Add note about why the entrypoint should always use `/bin/bash` and not `/bin/sh`
- Prune containers on the builder to prevent disk related build /publish failures

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -24,7 +24,10 @@ jobs:
       - run: gem install bundler
       - name: Build container
         working-directory: docker
-        run: make lint build test
+        run: |
+          docker system prune --all --force --volumes
+          docker builder prune --force --keep-storage=10GB
+          make lint build test
       - name: Publish container
         working-directory: docker
         run: |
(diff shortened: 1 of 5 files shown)


Iristyle:
Well, unfortunately the `Dockerfile` doesn't seem to work with buildkit disabled - investigating.

Iristyle:
Yanked the TravisCI buildkit commit out to put in a net new PR. There are going to need to be a few changes to the Dockerfile it looks like, so keeping those changes separate.

jpartlow (approved):

Iristyle:
So the `Dockerfile` depends quite a bit on buildkit features, so I'm looking into upgrading docker to a newer version instead to work around the Travis specific problem. See #2515

PR #2514, opened 2021-03-29
(maint) Disable BUILDKIT for Travis containers

Iristyle:
- Travis has been failing frequently with an error like:

   failed to solve with frontend dockerfile.v0: failed to build LLB:
   failed commit on ref "layer-sha256:XXX":
   unexpected commit size 0, expected 162: failed precondition

   This appears to be related to the version of Docker / Buildkit in
   use in Travis as this problem only shows up there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ jobs:
       language: ruby
       rvm: 2.6.5
       env:
+        - DOCKER_BUILDKIT=0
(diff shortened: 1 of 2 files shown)


Iristyle:
Disabling is probably going to be a bit of a headache with this multi-stage `Dockerfile`, so going to try and go the #2515 upgrade route instead.

PR #2516, opened 2021-03-29
(maint) Filter scanned certs in container startup

Iristyle:
- With cert preloading added to test suites, there is a server.crt
   present alongside puppet.pem. This entrypoint script errors and
   prevents the container from starting and the healthcheck fails
   because of the existence of server.crt, so update the filter to be
   more sensible

   This fix is not visible to this test suite as this suite will
   always generate certificates - but it will be visible elsewhere in
   suites like puppetdb, pupperware, etc

 - Log the certname as well
 - Allow `DOCKER_BUILDKIT` to be overridden (even though we need buildkit)
 - Refactors container healthcheck to be similar to PE variant
     - curl --no-progress-meter cleans up livenessprobe logs
     - Readiness probes in Kubernetes that use exec ignore timeout and
       require the command itself to enforce a timeout. Use `curl --max-time`
       to timeout if the request takes too long. Make it configurable so we
       can tune it as necessary.

       Also increases the healthcheck interval in Docker. Retry interval
       should be longer than the timeout.

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -9,7 +9,7 @@ hadolint_container := ghcr.io/hadolint/hadolint:latest
 export BUNDLE_PATH = $(PWD)/.bundle/gems
 export BUNDLE_BIN = $(PWD)/.bundle/bin
 export GEMFILE = $(PWD)/Gemfile
-export DOCKER_BUILDKIT = 1
+export DOCKER_BUILDKIT ?= 1
 PUPPERWARE_ANALYTICS_STREAM ?= dev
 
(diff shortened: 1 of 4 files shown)


Iristyle:
One more commit to push... but I built this container locally, and in conjunction with the work in https://github.com/puppetlabs/pupperware/pull/241, I can now successfully run

> PRELOAD_CERTS=1 bundle exec rspec spec --fail-fast

Also, note that the behavior between Puppet 7 and Puppet 6 differs a little, so once `edge` containers start coming off of `main` branch, the pupperware repo will need to change a bit.

PR #2517, opened 2021-03-30
(wip) accept crl updates

mwaggett:


src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1381,6 +1381,20 @@
...
        (time-coerce/from-long last-modified-milliseconds)))
 
+(schema/defn ^:always-validate update-crl :- CertificateRevocationList
+  [upstream-crl-pem :- InputStream
+   crl-path :- schema/Str]
+  (log/info (i18n/trs "Updating CRL at {0}" crl-path))
+  (let [upstream-crls (utils/pem->crls upstream-crl-pem)
+        our-crl (first (utils/pem->crls crl-path))
+        new-crl-chain (cons our-crl upstream-crls)]
+    ;; what verification do we need?
```
justinstoller, line 1391: I think in Puppet's Ruby code we used to create an SSLContext like object and validate that it is well formed. We might be able to do something similar here, load the ca certs and validate that the existing CA cert is still valid given the new CRLs

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1381,6 +1381,20 @@
   (let [last-modified-milliseconds (.lastModified (io/file cacrl))]
        (time-coerce/from-long last-modified-milliseconds)))
 
+(schema/defn ^:always-validate update-crl :- CertificateRevocationList
+  [upstream-crl-pem :- InputStream
+   crl-path :- schema/Str]
+  (log/info (i18n/trs "Updating CRL at {0}" crl-path))
+  (let [upstream-crls (utils/pem->crls upstream-crl-pem)
+        our-crl (first (utils/pem->crls crl-path))
```
Magisus, line 1389: `our-crl` here will already contain the upstream ones. We need to remove those and swap in the new ones.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1381,6 +1381,20 @@
...
+        new-crl-chain (cons our-crl upstream-crls)]
+    ;; what verification do we need?
+    ;; do we need to confirm that a custom intermediate CA is being used?
+    ;; do we need to verify everything against the cert chain? i.e. confirm the order?
```
Magisus, line 1393: I think doing this will answer all of the others, i.e. the verification will fail if any of those other things are true.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1381,6 +1381,20 @@
...
+    ;; do we need to check that our ca cert was signed by the subject of first crl in upstream-crls?
```
justinstoller, line 1394: iirc, the ordering of the certs and crls is simply a convention. One we rely on to load our cert, but I think in theory the rest of the CRLs could be in any order or contain other unrelated CRLs. So I'm not super into doing the order or quantity dependent checks you mention.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -95,6 +95,16 @@
           (rr/status 304)
           (rr/content-type "text/plain")))))
 
+(schema/defn handle-put-certificate-revocation-list!
+  [upstream-crl-pem :- InputStream
+   {:keys [cacrl infra-crl-path enable-infra-crl]} :- ca/CaSettings]
+  (ca/update-crl upstream-crl-pem cacrl)
```
justinstoller, line 101: I think we want to put wrap this in a lock

Magisus (commented):
We'll want to make sure to make use of this [write lock](https://github.com/puppetlabs/puppetserver/commit/8bcdc284dcbb52667672f1bb5097e3ecb20b7b56#diff-caf7efdcbdce3e60f52b15dd0489e7f16cad9c1dcfe2993ef1099ec843a50db8) when doing these updates.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
Magisus: I was hoping that whatever verification does would warn us if anything is actually awry, and call that good enough. But I haven't verified that that's actually true.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
mwaggett: yeah, haven't gotten to adding the locking yet, but it's on my list

src/clj/puppetlabs/puppetserver/certificate_authority.clj
Magisus: Oh I missed the `first`, I'm sorry.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
Magisus: Yes. Verifying an SSL context object like Justin suggests.

mwaggett: kk, thanks

mwaggett:
ended up on another branch, so closing this in favor of https://github.com/puppetlabs/puppetserver/pull/2530

PR #2523, opened 2021-04-15
(maint) Update Puppet Server for gzip suffix changes

justinstoller:


mwaggett (commented):
changes look good, but I think this should target 6.x and the commit message should reference the relevant ticket and include a little more detail about why we're making this change.

Magisus:
The ticket is a PE ticket, so not sure how I feel about referencing it. Agree that we could say this changed with the jetty update though.

mwaggett:
ugh. well we did decide we were okay with referencing internal tickets for r10k.. ¯\_(ツ)_/¯

justinstoller:
Updated

mwaggett (approved):

PR #2524, opened 2021-04-19
Merge 6.x -> main

Magisus:


mwaggett:
@Magisus I was gonna hold off on this until we've tagged, since those version bumps always make the job fail anyway. Does that seem reasonable? Or did you want this stuff to run through `main` before we tag anything?

Magisus:
I was trying to move things along for main, yeah, but you're running the releases, and if you're not worried about it, then I don't mind if it waits.

PR #2525, opened 2021-04-21
Merge 6.x -> main

mwaggett:
Conflicts were resolved by maintaining the state of the current branch.
Conflicts:
      acceptance/config/beaker/options.rb
      project.clj
      resources/ext/build-scripts/mri-gem-list-no-dependencies.txt

PR #2526, opened 2021-04-21
(maint) Use Docker 19.03 in Travis again

Iristyle:
- Instead of incurring the upgrade cost to Travis 20.10, use the
   existing 19.03 version, which includes the buildx plugin.

   To enable buildx requires enabling the experimental CLI via an
   environment variable

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,17 +27,17 @@ jobs:
       rvm: 2.6.5
       env:
         - DOCKER_COMPOSE_VERSION=1.28.6
+        - DOCKER_BUILDX_VERSION=0.5.1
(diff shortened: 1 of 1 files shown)


Iristyle:
@Magisus / @mwaggett - opening this as an experiment in response to fails noticed in #2525

Iristyle:
Well, my editor plugin somehow merged / pushed the `6.x` branch instead of the PR branch. I'll revert if the job at https://travis-ci.com/github/puppetlabs/puppetserver/builds/223797810 doesn't go green.

Apologies for the inadvertent commit, but glad the blast radius was limited to just this change!

Iristyle:
Annnnnd it went green - I'll take it!

PR #2529, opened 2021-04-28
(docs) Just a little note about server persistence across restarts

awrenn:
IMO, the easiest way to get started with puppetserver is using docker. I wasn't familiar with what would be required from to have a puppetserver that could survive across a container restart, so I had to look at some other sources. 

This just a little note that would help a person trying to get started with puppetserver by using the docker deployment. Please feel free to change any thing or reject it - this is just an idea that would have helped me get started slightly easier.

### DIFF
--- a/docker/puppetserver/README.md
+++ b/docker/puppetserver/README.md
@@ -50,6 +50,10 @@ The following environment variables are supported:
 If you would like to do additional initialization, add a directory called `/docker-custom-entrypoint.d/` and fill it with `.sh` scripts.
 These scripts will be executed at the end of the entrypoint script, before the service is ran.
 
+## Persistance 
+
+If you plan to use the in-server CA, restarting the container can cause the server's keys and certificates to change, causing agents and the server to stop trusting each other. To prevent this, you can persist the default ssldir, `/etc/puppetlabs/puppet/ssl`. For example, `docker run -v $PWD/ca-ssl:/etc/puppetlabs/puppet/ssl puppetlabs/puppetserver:latest`.
+
 ## Analytics Data Collection
 
 The puppetserver container collects usage data. This is disabled by default. You can enable it by passing `--env PUPPERWARE_ANALYTICS_ENABLED=true`
--- a/docker/puppetserver/README.md
+++ b/docker/puppetserver/README.md
@@ -52,7 +52,7 @@ These scripts will be executed at the end of the entrypoint script, before the s
 
 ## Persistance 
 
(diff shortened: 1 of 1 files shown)


Iristyle:
This container / build practice is the one inconsistent amongst all our containers unfortunately.

We should really be using a `VOLUME` declaration in the `Dockerfile` for state that should always persist across restarts like we do elsewhere (i.e. puppetdb):

https://github.com/puppetlabs/puppetdb/blob/6.x/docker/puppetdb/Dockerfile#L73-L76

Then at least an anonymous volume gets created rather than requiring named volumes are around. It's also a useful signal to users about what data is important. I think the ones we should add (for now, until we can sync up the layout with PE puppetserver) are:
https://github.com/puppetlabs/pupperware/blob/main/gem/lib/pupperware/compose-services/puppet.yml#L25-L27

awrenn:
Yeah - I think putting `VOLUME` declarations in the `Dockerfile` is a good idea, but mostly for documentation purposes. I haven't played with other OCI runtimes in a while, but in the case of deploying on kube, a consumer would need to be made aware of these volumes, and not just so that Docker specifically could handle it transparently. 

If there was what you are describing, plus maybe a single line in the README with 'Check the Dockerfile for notable volume mount points', that would have solved my issues I had getting started.

docker/puppetserver/README.md
```diff
@@ -50,6 +50,10 @@ The following environment variables are supported:
 If you would like to do additional initialization, add a directory called `/docker-custom-entrypoint.d/` and fill it with `.sh` scripts.
 These scripts will be executed at the end of the entrypoint script, before the service is ran.
 
+## Persistance 
+
+If you plan to use the in-server CA, restarting the container can cause the server's keys and certificates to change, causing agents and the server to stop trusting each other. To prevent this, you can persist the default ssldir, `/etc/puppetlabs/puppet/ssl`. For example, `docker run -v $PWD/ca-ssl:/etc/puppetlabs/puppet/ssl puppetlabs/puppetserver:latest`.
```
justinstoller, line 55:
```suggestion
If you plan to use the in-server CA, restarting the container can cause the server's keys and certificates to change, causing agents and the server to stop trusting each other. To prevent this, you can persist the default cadir, `/etc/puppetlabs/puppetserver/ca`. For example, `docker run -v $PWD/ca-ssl:/etc/puppetlabs/puppetserver/ca puppetlabs/puppetserver:latest`.
```

justinstoller (approved):
The default location of the CA changed from 6.x to 7.x so I updated that. With that change I'm happy to merge.

PR #2530, opened 2021-05-07
(SERVER-2550) Add CA API endpoint for updating CRLs

mwaggett:


Magisus:
So it looks like if someone submits a CRL that is older (has a lower number) than what we currently have, we will silently reject the update. That _seems_ like the correct thing to do probably, to ensure we're not dropping revocations.

But I was wondering about the use case @Sharpie mentioned, where someone might use this to make it easier to put in a truncated CRL for the Puppet CA. I think he said that currently people "truncate" them by creating a brand new CRL, which I assume would have a much lower CRL number. We would reject the update in that case in favor of the one we already have, because they have the same issuer, right? Is that actually a problem, or are we saying instead that we just don't want to support that kind of thing? If it is a problem, what would be the solution? It seems like there are more good reasons to reject lower-number CRLs than bad ones like this example...

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1381,6 +1381,63 @@
...
+  [incoming-crl-pem :- InputStream
+   crl-path :- schema/Str
+   cert-chain-path :- schema/Str]
+  (log/info (i18n/trs "Updating CRL at {0}" crl-path))
+  (let [incoming-crls (utils/pem->crls incoming-crl-pem)
+        current-crls (utils/pem->crls crl-path)
+        cert-chain (utils/pem->certs cert-chain-path)
+        crls-by-issuer (->> incoming-crls
+                            (concat current-crls)
+                            set
```
justinstoller, line 1432: What does set do here?

mwaggett:
yeah @Magisus a truncated CRL should be considered a "new" CRL and therefore should be created with a higher CRL Number. The CRL Number is not the number of revocations, if that's what you were thinking.

Magisus:
Okay, I wasn't sure if that was possible. If it is, that makes sense 👍

src/clj/puppetlabs/puppetserver/certificate_authority.clj
mwaggett: Making sure there aren't duplicates. Like if someone submits their whole CRL chain, but not all the individual CRLs have changed. The way I currently have it implemented, if there were duplicates, the update would fail because it would see two CRLs from the same issuer with the same CRL Number.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1381,6 +1381,63 @@
...
+                            (group-by .getIssuerX500Principal))
+        new-crl-by-issuer (into {} (for [[issuer crls] crls-by-issuer]
+                                     [issuer (get-newest-crl crls)]))
+        new-crl-chain (map ((.getSubjectX500Principal ^X509Certificate %) new-crl-by-issuer)
+                           cert-chain)]
```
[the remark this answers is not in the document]
mwaggett: uhh no idea. but I can't think of a reason?

mwaggett:
@Magisus Wasn't sure what was possible?
ooh, are you thinking that people would be hand-editing the CRLs?? No idea what would happen to the CRL Number if you did that, but I'm guessing nothing and I'm thinking that's fine because we probably shouldn't accept hand-edited CRLs??

Magisus:
No, more just thinking that if someone creates a new CRL with a key with openssl, I wasn't sure how hard it would be to make sure the CRL number was later than the old CRL's. I assume by default it would be just 0 (that's what I see on a baseline CRL from our infra that has no certs revoked yet).

mwaggett:
It probably is 0 by default, but I'm sure that can be configured somehow (I know you can do it in Java). Based on https://tools.ietf.org/html/rfc5280#section-5.2.3, I don't think any other behavior than requiring increasing CRL Numbers makes sense.

so yes, anything that mentions truncating CRLs should mention the CRL Number. We could definitely handle it with a tool, but that is out of scope for this ticket.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
justinstoller: Wow. I would have assumed that pem->certs would create different objects from those different streams. That's neat.

Magisus:
In that case, does it make sense to log that we're rejecting a submitted CRL because we had one that was newer than it? Otherwise someone who'd missed the memo about needing to use a higher CRL number might be confused why their update isn't taking effect...

src/clj/puppetlabs/puppetserver/certificate_authority.clj
Magisus: Does it not? That's also surprising to me.

mwaggett: It might, I didn't test it, but calling `set` seemed to be a good idea just in case? ¯\_(ツ)_/¯

justinstoller: It does work like Molly mentioned, I tested it before I replied.

mwaggett:
I think logging a warning that older CRLs were ignored makes sense. I'm assuming we'll tell people to submit a single, updated CRL for any and all issuers whose CRLs they want to update, even though we allow lots of junk? I'll have to adjust some things to get the right logging in, since right now I just smush the incoming CRLs together with the existing ones and don't really distinguish between them, but it's worth it if we think that will be a helpful message and not just drowned out by junk.

Magisus:
I think it would be useful if it was specific to the case where we kept the one already on disk and ignored all submissions for that issuer. Not for other cases (like choosing the latest among submitted things, or ignore extra CRLs). Because in that case the submission would look like a no-op, which might seem weird.

mwaggett:
ahh gotcha, that makes sense. 

so I have no idea how performant the stuff I've written is, but performance is something I've been somewhat trying to think about while writing this. My only concern is that keeping the current CRLs separate in order to log might be less performant? but I don't even know haha. so if y'all have thoughts/opinions/knowledge on performance and/or how idiomatic my code looks, I'm all ears. but in the meantime, I'll try refactoring this to log that scenario. and eventually I'll actually test it haha

Magisus:
I think there are few enough objects here that it shouldn't matter. The slowest parts are probably reading and writing from disk, and those won't be affected by the logging.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1427,14 +1450,13 @@
...
-                            set
-                            (group-by .getIssuerX500Principal))
-        new-crl-by-issuer (into {} (for [[issuer crls] crls-by-issuer]
-                                     [issuer (get-newest-crl crls)]))
-        new-crl-chain (map ((.getSubjectX500Principal ^X509Certificate %) new-crl-by-issuer)
-                           cert-chain)]
+        incoming-crls-by-issuer (->> incoming-crls
+                                     ;; even without combining with the current-crls,
+                                     ;; it doesn't seem like a bad idea to call `set`
+                                     set
```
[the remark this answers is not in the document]
mwaggett: oh, I thought @justinstoller said he tried it and got identical objects? but I'm happy to remove this if it's not necessary.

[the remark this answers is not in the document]
mwaggett: Tried this out:
```
dev-tools=> (count (set [(utils/pem->crl crl-path) (utils/pem->crl crl-path)]))
1
```
so I think it makes sense to keep `set`, just in case we get sent multiples of the same CRL (since it sounds like we're trying to be as permissive as is reasonably safe here).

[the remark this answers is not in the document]
Magisus: Thanks for trying this out! Since it works like you expected, it makes sense to keep it here. Sorry for the distraction :)

Sharpie:
> I wasn't sure how hard it would be to make sure the CRL number was later than the old CRL's.

The current CRL number can be retrieved with `openssl crl -crlnumber`, so it should be easy for folks to determine what the next number should be. Requiring a new number would be beneficial for cases where people aren't completely emptying the CRL but are trying to reduce the size  by removing duplicate entries or entries for private keys that are known to be destroyed (e.g. VM is long gone). Requiring a new number would let folks know if they happened to submit an update based on an outdated data.

E.g. sending in a modified CRL as "number 8" but having it fail because revocations came in while you were working on the update and the the next number in the sequence is now 9 or 10 is a good thing.

The one downside I can think of is the case where you are trying to re-initialize an empty CRL because the existing one was damaged or corrupted for some reason. But, this case should probably be handled as a separate operation that is executed when the service is shut down.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -95,6 +95,20 @@
...
           (rr/content-type "text/plain")))))
 
+(schema/defn handle-put-certificate-revocation-list!
+  [incoming-crl-pem :- InputStream
+   {:keys [cacrl cacert infra-crl-path enable-infra-crl]} :- ca/CaSettings]
+  (locking crl-write-serializer
+    (try
+      (ca/update-crls incoming-crl-pem cacrl cacert)
+      (when enable-infra-crl
+        (ca/update-crls incoming-crl-pem infra-crl-path cacert))
```
Magisus, line 105: If this endpoint is used to update our internal CRL, it will do the wrong thing with this infra CRL (i.e. set it to be the same as the full CRL). Do we want to try to avoid that somehow?

Magisus:
I put up https://github.com/puppetlabs/puppetserver/pull/2534, which carries this work forward. Closing this.

PR #2531, opened 2021-05-11
(SERVER-3014) Stop putting double slashes in test urls

Magisus:
As of version 4.5.8 of the apache httpclient, URLs with a double slash
in them are processed differently. Our tests were accidentally adding an
extra slash between the server and the URL path, which ran afoul of
this. This commit updates some callers of the test helper to stop adding
an additional slash.

### DIFF
--- a/test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
+++ b/test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
@@ -215,7 +215,7 @@
         app
         {:jruby-puppet {:master-conf-dir master-conf-dir}}
         (let [response (http-client/put
-                        (str "https://localhost:8140"
+                        (str "https://localhost:8140/"
                             "puppet-ca/v1/certificate_status/test_cert_ca_true")
                         {:ssl-cert (str master-conf-dir "/ssl/ca/ca_crt.pem")
                          :ssl-key (str master-conf-dir "/ssl/ca/ca_key.pem")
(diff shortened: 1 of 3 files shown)


mwaggett (approved):

Magisus:
There are more failures that still need addressing. Working on it.

test/integration/puppetlabs/services/certificate_authority/certificate_authority_int_test.clj
```diff
@@ -215,7 +215,7 @@
         app
         {:jruby-puppet {:master-conf-dir master-conf-dir}}
         (let [response (http-client/put
-                        (str "https://localhost:8140"
+                        (str "https://localhost:8140/"
```
Magisus, line 218: ...how did this work before??

justinstoller:
This seems like a pretty big change, I know it is incorrect to have more than one slash but I think a lot of scripts will.

Magisus:
This dependency was updated to deal with a CVE. This behavior change happened about 5 versions before the one we had to move to. It's also unclear to me right now (and we should probably test it) whether it affects the async client that we actually use within the product.

mwaggett:
Would it help if we used `fs` to construct these paths instead of `str`? I believe that should handle putting slashes in all and only the right places.

Magisus:
No, I think that's misleading, since they're not directories. And I think `fs` will return file objects rather than strings. It's fine with either internally, but `http/get` probably is not.

Magisus:
Okay, I confirmed that this does also affect the async client that we use internally, by modifying some tests in https://github.com/puppetlabs/puppetserver/blob/6.x/test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj.

So, do we think that's a deal breaker? Are we relying on letting double forward slashes through for any of the requests that might be made by the server? Maybe if it is, or even if we're worried about it, we should [opt out of URI normalization](https://issues.apache.org/jira/browse/HTTPCLIENT-1968?focusedCommentId=16813865&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16813865)? That would require plumbing some stuff through our clj-http-client.

mwaggett (approved):

PR #2534, opened 2021-05-18
(SERVER-2550) Add API endpoint for updating CRLs 

Magisus:
This adds a PUT mode to the `certificate_revocation_list`
endpoint that allows the user to submit CRLs, which will replace their
matching counterparts on disk if they are newer. The resulting CRL chain
is validated against the existing certs.

It does not allow replacing the signing cert of our CA.

Supersedes https://github.com/puppetlabs/puppetserver/pull/2530. Requires https://github.com/puppetlabs/clj-parent/pull/288.


src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1381,6 +1382,94 @@
...
+
+(schema/defn ^:always-validate maybe-replace-crl :- CertificateRevocationList
+  "Given a CRL and a map of key identifiers to CRLs, determine the
+  newest CRL with the key-id of the given CRL. Warn if the newest CRL
+  is the given CRL."
+  [crl :- CertificateRevocationList
+   key-crl-map :- {schema/Any [CertificateRevocationList]}
+   ca-name :- schema/Str]
+  (if (= ca-name (utils/get-cn-from-x500-principal
+                  (.getIssuerX500Principal crl)))
```
justinstoller, line 1431: Should we be checking the keyi d here?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1381,6 +1382,94 @@
...
+              ^String (i18n/trs "Could not determine newest CRL."))))))
+
+(schema/defn ^:always-validate maybe-replace-crl :- CertificateRevocationList
+  "Given a CRL and a map of key identifiers to CRLs, determine the
+  newest CRL with the key-id of the given CRL. Warn if the newest CRL
+  is the given CRL."
+  [crl :- CertificateRevocationList
+   key-crl-map :- {schema/Any [CertificateRevocationList]}
+   ca-name :- schema/Str]
+  (if (= ca-name (utils/get-cn-from-x500-principal
```
justinstoller, line 1430: this check feels outside of the scope of this function, but I don't feel too strongly about it.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1381,6 +1382,94 @@
...
+      (first newest-crl)
+      (throw (IllegalArgumentException.
+              ^String (i18n/trs "Could not determine newest CRL."))))))
+
+(schema/defn ^:always-validate maybe-replace-crl :- CertificateRevocationList
+  "Given a CRL and a map of key identifiers to CRLs, determine the
+  newest CRL with the key-id of the given CRL. Warn if the newest CRL
+  is the given CRL."
+  [crl :- CertificateRevocationList
+   key-crl-map :- {schema/Any [CertificateRevocationList]}
```
justinstoller, line 1428: Tightening up this schema would be nice for documentation purposes.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
Magisus: Yeah... I put it here after the original place I had it caused issues, but I think I understand those issues better now and could put it outside while still avoiding them.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
Magisus: Yes, probably. Need to figure out a good way to find the right cert to extract it from.

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -510,6 +510,97 @@
...
+       (let [incoming-crl-stream (testutils/pem-to-stream new-crls-for-multiple-certs-path)]
+         (testutils/with-backed-up-crl crl-path crl-backup-path
+           (update-crls incoming-crl-stream crl-path cert-chain-path "foo")
+           (let [old-crls (utils/pem->crls crl-backup-path)
+                 new-crls (utils/pem->crls crl-path)]
+             (doseq [new-crl new-crls
+                     old-crl old-crls]
+               (is (> (utils/get-crl-number new-crl)
+                      (utils/get-crl-number old-crl))))))))
+     (testing "unrelated CRLs are ignored while newer relevant CRLs are used"
```
justinstoller, line 558: Is the point of this that incoming-crl-stream if filtered and/or sorted before being written?

Magisus: Yeah, I think the key is the filtering, plus using the key-id rather than the name. The fixture in question has four CRLs in it, two sets with the same names but different key-ids. It tests that we use the matching set and ignore the others.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
Magisus: So the alternative to having it here is to slice off the leaf and put it back after the call to `map`. Do you think that's better? Felt a little more brittle/complicated to me.

justinstoller: Yeah, nbd.

justinstoller (approved):
Looks good once the tests are passing

src/clj/puppetlabs/puppetserver/certificate_authority.clj
Magisus: I'm working on this. The key is a a strange object.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
Magisus: Still need to do this.

Magisus:
The last commit here needs https://github.com/puppetlabs/jvm-ssl-utils/pull/113 and some new testutils for one of the test cases, that I'm working on generating.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1381,6 +1386,96 @@
...
+  "Given a CRL and a map of key identifiers to CRLs, determine the
+  newest CRL with the key-id of the given CRL. Warn if the newest CRL
+  is the given CRL. Never replaces the CRL corresponding to the Puppet
+  CA signing cert."
+  [crl :- CertificateRevocationList
+   key-crl-map :- {KeyIdExtension [CertificateRevocationList]}
+   ca-cert-key-id :- [Byte]]
+  (let [key-id-ext (utils/get-extension-value crl utils/authority-key-identifier-oid)
+        maybe-new-crls (get key-crl-map key-id-ext)]
+    (if (= (:key-identifier key-id-ext) ca-cert-key-id)
```
justinstoller, line 1437: Would it be easier to use https://github.com/puppetlabs/jvm-ssl-utils/blob/main/src/clojure/puppetlabs/ssl_utils/core.clj#L616 ?

Magisus: Easier, yes, but I think that method is also wrong for the same reason that what I was doing here before was wrong, right? As in, it's going by name and not key ID. Are you suggesting we update that method to be better?

tvpartytonight (approved):

PR #2535, opened 2021-05-20
(SERVER-2550) Update infra CRL via API

Magisus:
This commit ensures that the new PUT endpoint for the
`certificate_revocation_list` API also updates the infra CRL.

### DIFF
--- a/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
+++ b/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
@@ -18,12 +18,12 @@
             [schema.core :as schema]
             [cheshire.core :as cheshire]
             [liberator.core :refer [defresource]]
-            ;[liberator.dev :as liberator-dev]
             [liberator.representation :as representation]
             [ring.util.request :as request]
             [ring.util.response :as rr]
             [puppetlabs.trapperkeeper.services.status.status-core :as status-core]
-            [puppetlabs.i18n.core :as i18n]))
+            [puppetlabs.i18n.core :as i18n]
+            [puppetlabs.ssl-utils.core :as utils]))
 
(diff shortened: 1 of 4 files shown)


src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -97,10 +97,13 @@
 
 (schema/defn handle-put-certificate-revocation-list!
   [incoming-crl-pem :- InputStream
-   {:keys [cacrl cacert ca-name]} :- ca/CaSettings]
+   {:keys [cacrl cacert enable-infra-crl infra-crl-path]} :- ca/CaSettings]
   (locking crl-write-serializer
     (try
-      (ca/update-crls incoming-crl-pem cacrl cacert)
+      (let [incoming-crls (utils/pem->crls incoming-crl-pem)]
```
Magisus, line 103: You can only read input streams once, so we read it here and pass the contents on to the `update-crls` calls.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -18,12 +18,12 @@
             [schema.core :as schema]
             [cheshire.core :as cheshire]
             [liberator.core :refer [defresource]]
-            ;[liberator.dev :as liberator-dev]
```
justinstoller, line 21: ?

Magisus: My IDEA had done something weird with the indentation here, so I noticed it, and thought it was leftover, and deleted it. If you think it's in here for a good reason I'll put it back.

tvpartytonight (approved):
LGTM, the only thing that I am not sure of is if there is a use case to only update one or the other CRLs; if that's the case, we can probably just do that as a separate enhancement down the line.

Magisus:
I think it's pretty important that the two stay in sync. So if this was used because either one of them had an issue, it makes sense to me to always update both of them, just to be safe.

PR #2536, opened 2021-05-21
(SERVER-2550) Miscellaneous fixes

Magisus:


src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1463,17 +1463,24 @@
         cert-chain (utils/pem->certs cert-chain-path)
         ca-cert-key (utils/get-extension-value (first cert-chain)
                                                utils/subject-key-identifier-oid)
+        external-crl-chain (remove #(= ca-cert-key
+                                       (:key-identifier (utils/get-extension-value % utils/authority-key-identifier-oid)))
+                                   current-crls)
+        ca-crl (first (filter
+                       #(= ca-cert-key
+                           (:key-identifier (utils/get-extension-value % utils/authority-key-identifier-oid)))
+                       current-crls))
```
Magisus, line 1472: I'm still not sure this mess is better than doing the check in the loop. The key to this commit is dropping the logging.

Magisus: We could cheat by just dropping the first CRL (which would make this cleaner), but I feel like we're trying not to perpetuate relying on the ordering? If that's true, should I find the cert by searching for the one with our "ca-name", or...?

Magisus:
Doing manual validation with this.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
justinstoller: fwiw, extracting the anonymous function might make it more readable?

Magisus:
Travis is being slow and lame.

tvpartytonight (approved):

PR #2542, opened 2021-06-05
(SERVER-3021) Add `lein-nvd` security scanner

mwaggett:
This commit adds the `lein-nvd` security scanner and adds a vulnerability check
to the set of checks that Travis runs on each PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,9 @@ jobs:
     - jdk: openjdk11
       env: MULTITHREADED=true
       name: "Java 11 w/ multithreaded"
+    - stage: puppetserver vulnerability scan
+      script:
+        - lein nvd check
     - stage: puppetserver container tests
       dist: focal
       language: ruby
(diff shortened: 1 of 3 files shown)


mwaggett:
My local run uncovered some vulnerabilities, so not sure if we want to merge this before resolving those, but I wanted to get this up to test that Travis will actually run the check.

Magisus:
I heard some other teams talking about adding some exceptions too, so if we wanted to get this in and add those for the ones we either know we can't fix (like JRuby's BouncyCastle) or are planning to fix later, I think that's an option too?

mwaggett:
yep, I just want us to make those decisions consciously.

mwaggett:
woooo, it works in Travis lol

Magisus:
Looks like this is causing some issues for ezbake... probably because it has that separate deps profile?

mwaggett:
woo finally green!

Magisus (approved):

PR #2545, opened 2021-06-16
(SERVER-3033) Fix auth rule for CRL update endpoint

mwaggett:
Prior to this change, all PUT requests to the `certificate_revocation_list`
endpoint would get denied, since the request path wouldn't exactly match the
string in the auth rule. Using type `regex` ensures that we can access this
endpoint appropriately.


Magisus (approved):

Magisus:
Travis doesn't test this.

PR #2546, opened 2021-06-16
(PE-32070) Update Java 11 security file for new BCFIPS

Magisus:
There is an issue with the latest bc-fips version involving circular
provider loading, that can be fixed by using the full class name for the
BCJSSE FIPS provider, rather than relying on an alias. This commit
updates our java-security file for Java 11 to use the full name.

It also updates the key manager factory algorithm following a
recommendation from the BC maintainers.


mwaggett (approved):

Magisus:
Updated this to also change the files for Java 8.

mwaggett (approved):

mwaggett:
This looks like the docker failure they've been dealing with for a while, so gonna merge past that.

PR #2548, opened 2021-06-21
(PE-32070) Use BCFKS key factory in FIPS mode

Magisus:
The PKIX factory is correct for non-FIPS BC, but in FIPS mode we want to
use BCFKS.


mwaggett (approved):

Magisus:
The docker failure is a known issue, unrelated to this change.

PR #2550, opened 2021-06-30
(SERVER-3032) Always allow SAN that matches CSR subject

Magisus:
Previously, when `allow-subject-alt-names` was set to false, we rejected
any and all SANs on CSRs. However, now that we are adding the CSR's
subject as an alt name when generating CSRs, per RFC 2818, we need to
make an exception, so as to not block all internally generated CSRs.
This affects, for example, the `puppetserver ca generate` command. This
commit updates the validation function to allow a single alt name
matching the CSR's subject, even when `allow-subject-alt-names` is
false.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -1314,18 +1314,27 @@
 (schema/defn ensure-subject-alt-names-allowed!
   "Throws an exception if the CSR contains subject-alt-names AND the user has
    chosen to disallow subject-alt-names. Subject alt names can be allowed by
-   setting allow-subject-alt-names to true in the ca.conf file."
+   setting allow-subject-alt-names to true in the ca.conf file. Always allows
+   a single subject alt name that matches the CSR subject, which may be
+   present to comply with RFC 2818 (see SERVER-2338)."
   [csr :- CertificateRequest
    allow-subject-alt-names :- schema/Bool]
   (when-let [subject-alt-names (not-empty (subject-alt-names csr))]
     (if (false? allow-subject-alt-names)
-      (let [subject (get-csr-subject csr)]
-        (sling/throw+
-          {:kind :disallowed-extension
-           :msg (format "%s %s %s"
-                        (i18n/tru "CSR ''{0}'' contains subject alternative names ({1}), which are disallowed." subject (str/join ", " subject-alt-names))
-                        (i18n/tru "To allow subject alternative names, set allow-subject-alt-names to true in your ca.conf file.")
-                        (i18n/tru "Then restart the puppetserver and try signing this certificate again."))})))))
+      (let [subject (get-csr-subject csr)
+            cn-alt-name (str "DNS:" subject)]
+        (if (and (= 1 (count subject-alt-names))
+                 (= (first subject-alt-names) cn-alt-name))
+          (log/debug "Allowing subject alt name that matches CSR subject.")
+          (let [disallowed-alt-names (filter #(not (= cn-alt-name %))
+                                             subject-alt-names)]
+            (sling/throw+
+             {:kind :disallowed-extension
+              :msg (format "%s %s %s"
+                           (i18n/tru "CSR ''{0}'' contains extra subject alternative names ({1}), which are disallowed."
+                                     subject (str/join ", " disallowed-alt-names))
+                           (i18n/tru "To allow subject alternative names, set allow-subject-alt-names to true in your ca.conf file.")
+                           (i18n/tru "Then restart the puppetserver and try signing this certificate again."))})))))))
 
 (schema/defn ^:always-validate process-csr-submission!
   "Given a CSR for a subject (typically from the HTTP endpoint),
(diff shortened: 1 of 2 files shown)


Magisus:
Tested this successfully with `puppetserver ca generate`.

Magisus:
I'm actually not sure this is the right approach, need to look into something...

Magisus:
Never mind, we're going to stick with this approach.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1314,18 +1314,23 @@
...
-                        (i18n/tru "CSR ''{0}'' contains subject alternative names ({1}), which are disallowed." subject (str/join ", " subject-alt-names))
-                        (i18n/tru "To allow subject alternative names, set allow-subject-alt-names to true in your ca.conf file.")
-                        (i18n/tru "Then restart the puppetserver and try signing this certificate again."))})))))
+        (if (and (= 1 (count subject-alt-names))
+                 (= (first subject-alt-names) (str "DNS:" subject)))
+          (log/debug "Allowing subject alt name that matches CSR subject.")
+          (sling/throw+
+           {:kind :disallowed-extension
+            :msg (format "%s %s %s"
+                         (i18n/tru "CSR ''{0}'' contains subject alternative names ({1}), which are disallowed." subject (str/join ", " subject-alt-names))
```
mwaggett, line 1331: We probably don't want to include the certname SAN in this error message.

justinstoller: why?

mwaggett: As a user reading this message, I would interpret this to mean that all the SANs listed were disallowed.

Magisus: I see what you're saying, but I'm not sure how I feel about it. We should maybe discuss?

justinstoller: Ah. So, either we list all the SANs for completeness, but make a comment about all but the certname one are allowed, or we filter the certname one from the list of prohibited SANs.

Magisus: Yeah. Not sure which makes more sense. The error is already pretty wordy...

Magisus: Updated with the filtering and a slight rewording, lemme know what you think 🙂

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1314,18 +1314,26 @@
...
+        (if (and (= 1 (count subject-alt-names))
+                 (= (first subject-alt-names) (str "DNS:" subject)))
+          (log/debug "Allowing subject alt name that matches CSR subject.")
+          (let [disallowed-alt-names (filter #(not (= (str "DNS:" subject) %))
+                                             subject-alt-names)]
+            (sling/throw+
+             {:kind :disallowed-extension
+              :msg (format "%s %s %s"
+                           (i18n/tru "CSR ''{0}'' contains extra subject alternative names ({1}), which are disallowed."
+                                     subject (str/join ", " disallowed-alt-names))
```
mwaggett, line 1334: I think I like this!

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1314,18 +1314,26 @@
...
     (if (false? allow-subject-alt-names)
       (let [subject (get-csr-subject csr)]
-        (sling/throw+
-          {:kind :disallowed-extension
-           :msg (format "%s %s %s"
-                        (i18n/tru "CSR ''{0}'' contains subject alternative names ({1}), which are disallowed." subject (str/join ", " subject-alt-names))
-                        (i18n/tru "To allow subject alternative names, set allow-subject-alt-names to true in your ca.conf file.")
-                        (i18n/tru "Then restart the puppetserver and try signing this certificate again."))})))))
+        (if (and (= 1 (count subject-alt-names))
+                 (= (first subject-alt-names) (str "DNS:" subject)))
```
mwaggett, line 1326: might be good to pull the allowed SAN into a binding, since we use it a couple times now.

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -1462,6 +1462,42 @@
...
+               (ensure-subject-alt-names-allowed! csr false))))
+        (let [alt-name-ext {:oid utils/subject-alt-name-oid
+                              :value {:dns-name ["bad-name" subject]}
+                              :critical false}
+                csr (utils/generate-certificate-request subject-keys subject-dn [alt-name-ext])]
+          (is (thrown+-with-msg?
+               [:kind :disallowed-extension]
+               #".*new-cert.*subject alternative names.*bad-name.*"
+               (ensure-subject-alt-names-allowed! csr false)))))
+      (testing "allows a single alt name matching the subject"
```
mwaggett, line 1487: would it be worth testing that the debug message is logged when `allow-subject-alt-names` is false and not when it's true? or nah?

Magisus:
Updated.

mwaggett (approved):

PR #2555, opened 2021-07-19
(PE-32484) Update ast_compiler to work for projects

McdonaldSeanp:
This commit adds logic to the compile endpoint to support compiling AST for a
bolt project rather than an environment. Mostly project compiles are the same
as environment compiles (after all, we are still just compiling puppet code)
with a few important differences:
* We pass some additional data about the project to the compile operation
  (this is mostly data about the project name and where it's located on disk)
* The context of project compiles requires us to use `in_tmp_environment`
  instead of `in_environment` so that we can 1. ensure that we are not loading
  things from environment.conf and 2. that we can control the entirety of the
  modulepath for the compile (projects do not make use of
  basemodulepath/vendormodulepath and we need to ensure we have only exactly
  the modulepath passed in by the project)

### DIFF
--- a/spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
+++ b/spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
@@ -15,12 +15,26 @@ def generate_ast(code)
   end
 end
 
-def request(code)
+def environment_request(code)
   {"certname" => "localhost",
    "facts" => {"values" => {"my_fact" => "fact_value"}},
    "trusted_facts" => {"values" => {"my_trusted" => "trusted_value"}},
    "options" => {"capture_logs" => false},
    "environment" => "production",
+   "code_ast" => generate_ast(code).to_json,
+   "variables" => {"values" => {"foo" => "bar"}}}
+end
+
+def project_request(code)
+  {"certname" => "localhost",
+   "facts" => {"values" => {"my_fact" => "fact_value"}},
+   "trusted_facts" => {"values" => {"my_trusted" => "trusted_value"}},
+   "options" => {"capture_logs" => false, "compile_for_plan" => true},
+   "versioned_project" => "fakeproject_GITSHA",
+   "project_root" => "/etc/puppetlabs/puppetserver/projects",
+   "modulepath" => ['/etc/puppetlabs/puppetserver/projects/fakeproject_GITSHA', '/etc/puppetlabs/puppetserver/projects/fakeproject_GITSHA/.modules'],
+   "hiera_config" => ['/etc/puppetlabs/puppetserver/projects/fakeproject_GITSHA/hiera.yaml'],
+   "project_name" => 'fakeproject',
    "code_ast" => generate_ast(code).to_json,
    "variables" => {"values" => {"foo" => "bar"}}}
 end
@@ -37,42 +51,42 @@ describe Puppet::Server::ASTCompiler do
     let(:boltlib_path) { nil }
 
     it 'handles basic resources' do
-      response = Puppet::Server::ASTCompiler.compile(request("notify { 'my_notify': }"), boltlib_path)
+      response = Puppet::Server::ASTCompiler.compile(environment_request("notify { 'my_notify': }"), boltlib_path)
       notify = find_notify(response[:catalog])
       expect(notify).not_to be_nil
       expect(notify['title']).to eq("my_notify")
     end
 
     it 'correctly interpolates supplied variables' do
-      response = Puppet::Server::ASTCompiler.compile(request('notify { "$foo": }'), boltlib_path)
+      response = Puppet::Server::ASTCompiler.compile(environment_request('notify { "$foo": }'), boltlib_path)
       notify = find_notify(response[:catalog])
       expect(notify).not_to be_nil
       expect(notify['title']).to eq("bar")
     end
 
     it 'correctly interpolates supplied facts' do
-      response = Puppet::Server::ASTCompiler.compile(request('notify { "$my_fact": }'), boltlib_path)
+      response = Puppet::Server::ASTCompiler.compile(environment_request('notify { "$my_fact": }'), boltlib_path)
       notify = find_notify(response[:catalog])
       expect(notify).not_to be_nil
       expect(notify['title']).to eq("fact_value")
     end
 
     it 'correctly interpolates supplied trusted facts' do
-      response = Puppet::Server::ASTCompiler.compile(request('notify { "${trusted[\'my_trusted\']}": }'), boltlib_path)
+      response = Puppet::Server::ASTCompiler.compile(environment_request('notify { "${trusted[\'my_trusted\']}": }'), boltlib_path)
       notify = find_notify(response[:catalog])
       expect(notify).not_to be_nil
       expect(notify['title']).to eq("trusted_value")
     end
 
     it 'fails gracefully when boltlib is not found when request requires bolt types' do
-      bolt_request = request('notify { "${trusted[\'my_trusted\']}": }')
+      bolt_request = environment_request('notify { "${trusted[\'my_trusted\']}": }')
       bolt_request.merge!({'options' => { 'compile_for_plan' => true } })
       expect {
         Puppet::Server::ASTCompiler.compile(bolt_request, boltlib_path)
       }.to raise_error(Puppet::Error, /the path to boltlib modules must be provided/)
     end
 
-    context 'when compiling for a plan with fact and plan/target variable collisions' do
+    context 'when compiling for an environment plan with fact and plan/target variable collisions' do
       let(:boltlib_path) { [] }
 
       it 'properly shadows the relevant variables' do
@@ -111,7 +125,7 @@ notify { "plan_var_fact": message => $plan_var_fact }
 notify { "target_var_fact": message => $target_var_fact }
 notify { "plan_var_plain":  message => $plan_var_plain }
 CODE
-        bolt_request = request(code)
+        bolt_request = environment_request(code)
         bolt_request.merge!(
           'facts'            => facts,
           'variables'        => variables,
@@ -132,6 +146,42 @@ CODE
           notify = find_notify(catalog, title: var)
           expect(notify).not_to be_nil
           expect(notify['parameters']['message']).to eql(expected_value)
+        end
+      end
+    end
+
+    context 'when running a project plan' do
+      let(:boltlib_path) { [] }
+      context 'single threaded' do
+        it 'returns puppet settings back to normal' do
+          # TODO: This is a temporary stub of the Bolt requires to get this
+          # test to pass. We should figure out how to properly load Bolt code
+          # later for the tests.
+          allow(Puppet::Server::ASTCompiler).to receive(:load_bolt)
+          mock_apply_inventory_class = double('Bolt::ApplyInventory')
+          allow(mock_apply_inventory_class).to receive(:new).with(anything).and_return(double('mock'))
+          stub_const("Bolt::ApplyInventory", mock_apply_inventory_class)
+
+          original_settings = {
+            'vardir' => Puppet[:vardir],
+            'confdir' => Puppet[:confdir],
+            'logdir' => Puppet[:logdir],
+            'codedir' => Puppet[:codedir],
+            'basemodulepath' => Puppet[:basemodulepath],
+            'vendormoduledir' => Puppet[:vendormoduledir],
+            'hiera_config' => Puppet[:hiera_config]
+          }
+
+          # Use a simple resource, since we aren't testing the compilation itself
+          # but rather ensuring the compilation does not interfere with global state
+          response = Puppet::Server::ASTCompiler.compile(project_request("notify { 'my_notify': }"), boltlib_path)
+          notify = find_notify(response[:catalog])
+          expect(notify).not_to be_nil
+          expect(notify['title']).to eq("my_notify")
+
+          original_settings.each do |setting_name, original_setting|
+            expect(Puppet[setting_name]).to eq(original_setting)
+          end
         end
       end
     end
--- a/spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
+++ b/spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
@@ -152,37 +152,21 @@ CODE
 
     context 'when running a project plan' do
       let(:boltlib_path) { [] }
-      context 'single threaded' do
-        it 'returns puppet settings back to normal' do
-          # TODO: This is a temporary stub of the Bolt requires to get this
-          # test to pass. We should figure out how to properly load Bolt code
-          # later for the tests.
-          allow(Puppet::Server::ASTCompiler).to receive(:load_bolt)
-          mock_apply_inventory_class = double('Bolt::ApplyInventory')
-          allow(mock_apply_inventory_class).to receive(:new).with(anything).and_return(double('mock'))
-          stub_const("Bolt::ApplyInventory", mock_apply_inventory_class)
+      it 'returns puppet settings back to normal' do
+        # TODO: This is a temporary stub of the Bolt requires to get this
+        # test to pass. We should figure out how to properly load Bolt code
+        # later for the tests.
+        allow(Puppet::Server::ASTCompiler).to receive(:load_bolt)
+        mock_apply_inventory_class = double('Bolt::ApplyInventory')
+        allow(mock_apply_inventory_class).to receive(:new).with(anything).and_return(double('mock'))
+        stub_const("Bolt::ApplyInventory", mock_apply_inventory_class)
 
-          original_settings = {
-            'vardir' => Puppet[:vardir],
-            'confdir' => Puppet[:confdir],
-            'logdir' => Puppet[:logdir],
-            'codedir' => Puppet[:codedir],
-            'basemodulepath' => Puppet[:basemodulepath],
-            'vendormoduledir' => Puppet[:vendormoduledir],
-            'hiera_config' => Puppet[:hiera_config]
-          }
-
-          # Use a simple resource, since we aren't testing the compilation itself
-          # but rather ensuring the compilation does not interfere with global state
-          response = Puppet::Server::ASTCompiler.compile(project_request("notify { 'my_notify': }"), boltlib_path)
-          notify = find_notify(response[:catalog])
-          expect(notify).not_to be_nil
-          expect(notify['title']).to eq("my_notify")
-
-          original_settings.each do |setting_name, original_setting|
-            expect(Puppet[setting_name]).to eq(original_setting)
-          end
-        end
+        # Use a simple resource, since we aren't testing the compilation itself
+        # but rather ensuring the compilation does not interfere with global state
+        response = Puppet::Server::ASTCompiler.compile(project_request("notify { 'my_notify': }"), boltlib_path)
+        notify = find_notify(response[:catalog])
+        expect(notify).not_to be_nil
+        expect(notify['title']).to eq("my_notify")
       end
     end
   end
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -892,11 +892,14 @@
 
 (def CompileRequest
   {(schema/required-key "certname") schema/Str
-   (schema/required-key "environment") schema/Str
    (schema/required-key "code_ast") schema/Str
    (schema/required-key "trusted_facts") {(schema/required-key "values") {schema/Str schema/Any}}
    (schema/required-key "facts") {(schema/required-key "values") {schema/Str schema/Any}}
    (schema/required-key "variables") {(schema/required-key "values") (schema/either [{schema/Str schema/Any}] {schema/Str schema/Any})}
+  ;;  Both environment and versioned project are technically listed in the schema as
+  ;;  "optional" but we will check later that exactly one of them is set.
+   (schema/optional-key "environment") schema/Str
+   (schema/optional-key "versioned_project") schema/Str
    (schema/optional-key "target_variables") {(schema/required-key "values") {schema/Str schema/Any}}
    (schema/optional-key "job_id") schema/Str
    (schema/optional-key "transaction_uuid") schema/Str
@@ -937,26 +940,62 @@
                                                      "code_id"
                                                      (current-code-id-fn (get request-options "environment")))))})))
 
+(defn parse-project-compile-data
+  "Parse data required to compile a catalog inside a project. Data required includes
+   * Root path to project
+   * modulepath
+   * hiera config
+   * project_name"
+  [request-options
+   versioned-project
+   bolt-projects-dir]
+  (let [project-root (file-serving/get-project-root bolt-projects-dir versioned-project)
+        project-config (file-serving/read-bolt-project-config project-root)]
+    (assoc request-options "project_root" project-root
+                           "modulepath" (map #(str project-root "/" %) (file-serving/get-project-modulepath project-config))
+                           "hiera_config" (str project-root "/" (get project-config :hiera-config "hiera.yaml"))
+                           "project_name" (:name project-config))))
+
 (schema/defn ^:always-validate compile-fn :- IFn
   [jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)
    current-code-id-fn :- IFn
-   boltlib-path :- (schema/maybe [schema/Str])]
+   boltlib-path :- (schema/maybe [schema/Str])
+   bolt-projects-dir :- (schema/maybe schema/Str)]
   (fn [request]
-    (let [env (jruby-request/get-environment-from-request request)
-          request-options (-> request
+    (let [request-options (-> request
                               :body
                               slurp
                               (validated-body CompileRequest))
-          compile-options (assoc request-options
-                                 "code_id"
-                                 (current-code-id-fn env))]
-      {:status 200
-       :headers {"Content-Type" "application/json"}
-       :body (json/encode
-               (jruby-protocol/compile-ast jruby-service
-                                           (:jruby-instance request)
-                                           compile-options
-                                           boltlib-path))})))
+          versioned-project (get request-options "versioned_project")
+          environment (get request-options "environment")]
+      ;; Check to ensure environment/versioned_project are mutually exlusive and
+      ;; at least one of them is set.
+      (cond
+        (and versioned-project environment)
+        {:status 400
+         :headers {"Content-Type" "text/plain"}
+         :body (i18n/tru "A compile request cannot specify both `environment` and `versioned_project` parameters.")}
+        
+        (and (nil? versioned-project) (nil? environment))
+        {:status 400
+         :headers {"Content-Type" "text/plain"}
+         :body (i18n/tru "A compile request must include an `environment` or `versioned_project` parameter.")}
+        
+        :else
+        (let [compile-options (if versioned-project
+                              ;; we need to parse some data from the project config for project compiles
+                                (parse-project-compile-data request-options versioned-project bolt-projects-dir)
+                              ;; environment compiles only need to set the code ID
+                                (assoc request-options
+                                       "code_id"
+                                       (current-code-id-fn environment)))]
+          {:status 200
+           :headers {"Content-Type" "application/json"}
+           :body (json/encode
+                  (jruby-protocol/compile-ast jruby-service
+                                              (:jruby-instance request)
+                                              compile-options
+                                              boltlib-path))})))))
 
 (schema/defn ^:always-validate
   v4-catalog-handler :- IFn
@@ -973,8 +1012,9 @@
   [jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)
    wrap-with-jruby-queue-limit :- IFn
    current-code-id-fn :- IFn
-   boltlib-path :- (schema/maybe [schema/Str])]
-  (-> (compile-fn jruby-service current-code-id-fn boltlib-path)
+   boltlib-path :- (schema/maybe [schema/Str])
+   bolt-projects-dir :- (schema/maybe schema/Str)]
+  (-> (compile-fn jruby-service current-code-id-fn boltlib-path bolt-projects-dir)
       (jruby-request/wrap-with-jruby-instance jruby-service)
       wrap-with-jruby-queue-limit
       jruby-request/wrap-with-error-handling))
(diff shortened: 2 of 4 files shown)


src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -34,91 +34,197 @@ def self.compile_ast(code, compile_options, boltlib_path)
...
+            compiler.evaluate(ast)
+            compiler.compile_additions
+            compiler.catalog_data_hash
+          end
+        end
+      end
+      private_class_method :compile_for_environment
+
+
+      def self.with_puppet_settings(overrides = {})
```
McdonaldSeanp, line 207: This function seems like the most dangerous part of the functionality for compiling for projects. In it we basically attempt to clear all settings for puppet so that the compile in the project context isn't polluted by settings fetched automatically by puppet.

This was taken almost wholesale from what we do in Bolt, but we need to discuss what impact this has on the JRuby environment and if we can even use this function or something like it.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -893,6 +893,7 @@
 (def CompileRequest
   {(schema/required-key "certname") schema/Str
    (schema/required-key "environment") schema/Str
+   (schema/required-key "versioned_project") schema/Str
```
donoghuc, line 896: are we stuck with a required "environment" argument? Could we make them both optional and then ensure there is only one? It seems to me like right now you have to have environment and versioned project, but versioned_project alone is not allowed?

McdonaldSeanp: Whoops! This is just a typo, they should both be optional

McdonaldSeanp: updated

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
Magisus: Unlike in Bolt, where something runs once and then is done, this loaded Puppet instance will continue to be used after the project compile is done, and in that case, we need all of the original settings.

Maybe there is a way you can leverage the Settings object we define here in Puppetserver, which has a concept of "local" settings overrides: https://github.com/puppetlabs/puppetserver/blob/2f9122bb2198f2b1e7b0ee638c33da8766f26a8b/src/ruby/puppetserver-lib/puppet/server/settings.rb#L3 It's designed for multithreaded, but might be usable here too with some extension, e.g. a way to create the "local" settings with more than one new value at once. And then I think you can just call `clear_local_settings` when you're done to go back to our main global ones. Worth looking into anyway, I didn't think about it too hard, so hopefully this isn't super off-base.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -34,52 +34,72 @@ def self.compile_ast(code, compile_options, boltlib_path)
...
+        # TODO: Given we hide this from plan authors this current iteration has only
+        # the "required" data for now. Once we can get https://github.com/puppetlabs/bolt/pull/1770
+        # merged and promoted we can just use an empty hash.
+        fake_config = {
+          'transport' => 'redacted',
+          'transports' => {
+            'redacted' => 'redacted'
+          }
+        }
+        bolt_inv = Bolt::ApplyInventory.new({fake_config})
```
donoghuc, line 63:
```suggestion
        bolt_inv = Bolt::ApplyInventory.new(fake_config)
```
Same in the other construction of a bolt_inv.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -34,52 +34,72 @@ def self.compile_ast(code, compile_options, boltlib_path)
...
-              'transport' => 'redacted',
-              'transports' => {
-                'redacted' => 'redacted'
-              }
-            }
-            bolt_inv = Bolt::ApplyInventory.new(fake_config)
-            Puppet.override(bolt_inventory: bolt_inv) do
+        env_conf = {
+          pre_modulepath: boltlib_path,
+          modulepath: compile_options['modulepath']
```
donoghuc, line 89:
```suggestion
          modulepath: compile_options['modulepath'],
```

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -34,52 +34,72 @@ def self.compile_ast(code, compile_options, boltlib_path)
...
+        # merged and promoted we can just use an empty hash.
+        fake_config = {
+          'transport' => 'redacted',
+          'transports' => {
+            'redacted' => 'redacted'
+          }
+        }
+        bolt_inv = Bolt::ApplyInventory.new({fake_config})
+        puppet_overrides = {
+          bolt_inventory: inv,
```
donoghuc, line 65:
```suggestion
          bolt_inventory: bolt_inv,
```

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -937,26 +940,63 @@
...
+        (assoc "modulepath" (file-serving/get-project-modulepath project-config))
+        (assoc "hiera_config" (:hiera-config project-config))
+        (assoc "project_name" (:name project-config)))))
+
 (schema/defn ^:always-validate compile-fn :- IFn
   [jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)
    current-code-id-fn :- IFn
-   boltlib-path :- (schema/maybe [schema/Str])]
+   boltlib-path :- (schema/maybe [schema/Str])
+   bolt-projects-dir :- (schema/maybe [schema/Str])]
```
donoghuc, line 964: This must be a copy paste error? We dont expect this to be an array of paths right? Ultimately, the versioned_project we send along in the request would resolve to a single file path to a deployed bolt project?

McdonaldSeanp: Oh, yep! good catch

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -937,26 +940,63 @@
...
-                                 (current-code-id-fn env))]
-      {:status 200
-       :headers {"Content-Type" "application/json"}
-       :body (json/encode
-               (jruby-protocol/compile-ast jruby-service
-                                           (:jruby-instance request)
-                                           compile-options
-                                           boltlib-path))})))
+          versioned-project (get request-options :versioned_project)
+          environment (get request-options :environment)]
```
donoghuc, line 971: I'm not convinced this is how you would get environment... I have not dug in yet, but it appears this is always null based on EVERY compile request (from environments) failng with this message.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -34,52 +34,72 @@ def self.compile_ast(code, compile_options, boltlib_path)
...
-            bolt_inv = Bolt::ApplyInventory.new(fake_config)
-            Puppet.override(bolt_inventory: bolt_inv) do
+        env_conf = {
+          pre_modulepath: boltlib_path,
+          modulepath: compile_options['modulepath'],
+          facts: compile_options['facts']['values'],
+        }
+
+        puppet_settings = {
+          node_name_value: target['name'],
```
donoghuc, line 94: This looks like a copy paste issue from bolt (`target` is undefined in this context). It looks like we set this (for env plans) like: `Puppet[:node_name_value] = compile_options['certname']`. Maybe we should update environments to do the override like we do here for hiera config. I'm not sure what the implications of doing puppet settings are like that. 

https://github.com/puppetlabs/puppetserver/pull/2555/files#diff-6f7f4ebbd198c182ba9b89fe43839b117bc1899b5bb3f7ffeddcd1cc64cd79abR119

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -34,52 +34,72 @@ def self.compile_ast(code, compile_options, boltlib_path)
...
+          hiera_config:    request['hiera_config']
```
donoghuc, line 95:
```suggestion
          hiera_config:    request['compile_options']
```

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -91,34 +111,120 @@ def self.compile_ast(code, compile_options, boltlib_path)
...
+          # Override module locations, Bolt includes vendored modules in its internal modulepath.
+          Puppet.settings.override_default(:basemodulepath, '')
+          Puppet.settings.override_default(:vendormoduledir, '')
+
+          Puppet.initialize_settings(cli)
+          overrides.each do |setting, value|
+            Puppet.settings[setting] = value
+          end
+
+          Puppet.settings[:log_level] = @log_level
```
donoghuc, line 223: `@log_level` wont be set here. Not going to bother making this dynamic for now because we may or may not even go this route. I would suggest either leaving it unset (and hopefully picking up some default) or hard coding it to something reasonable.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -91,34 +111,120 @@ def self.compile_ast(code, compile_options, boltlib_path)
...
 
+        target_variables = compile_options.dig('target_variables', 'values') || {}
+
+        variables = {
+          variables: plan_variables,
+          target_variables: target_variables,
+        }
+
+        env_conf = {
+          pre_modulepath: boltlib_path,
```
donoghuc, line 140: This is not a kwarg for in_tmp_enviornment like it is for in_environment. You maybe want something like `Array(boltlib_path) + Array(compile_options['modulepath'])`. Not sure exactly though

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
McdonaldSeanp: There shouldn't be any implications for changing things to do it the bolt way.

I just realized we _can't_ use the `Puppet[:node_name_value] = ` directly because we completely wipe the settings between assigning that setting and performing the compile. For project compiles we _have_ to pass the value through the puppet_settings hash

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -937,26 +940,63 @@
...
+   * hiera config
+   * project_name"
+  [request-options
+   bolt-projects-dir]
+  (let [versioned-project (get request-options "versioned_project")
+        project-root (file-serving/get-project-root bolt-projects-dir versioned-project)
+        project-config (file-serving/read-bolt-project-config project-root)]
+    (-> request-options
+        (assoc "project_root" project-root)
+        (assoc "modulepath" (map #(str project-root "/" %) (file-serving/get-project-modulepath project-config))
```
donoghuc, line 956: Missing a closing paren ;)

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -91,34 +108,128 @@ def self.compile_ast(code, compile_options, boltlib_path)
...
+        original_settings = Puppet.settings
+        original_basemodulepath = Puppet[:basemodulepath]
+        original_vendormoduledir = Puppet[:vendormoduledir]
+
+        Dir.mktmpdir('bolt_compile') do |dir|
+          puppet_config = {}
+          Puppet::Settings::REQUIRED_APP_SETTINGS.each do |setting|
+            puppet_config[setting] = dir
+          end
+          Puppet::Server::PuppetConfig.initialize_puppet(puppet_config: puppet_config)
```
donoghuc, line 214: I know you are working on a new way of safely setting puppet settings. I think in bolt we use that clear_all_settings_for_test method. Anyway, for this context this blows up:
```
2021-07-22T18:04:27.336Z ERROR [qtp1366641407-62] [p.r.core] Internal Server Error: org.jruby.exceptions.RuntimeError: (DevError) Attempting to initialize global default settings more than once!
        at RUBY.initialize_global_settings(/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/settings.rb:301)
        at RUBY.do_initialize_settings_for_run_mode(/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet.rb:184)
        at RUBY.initialize_settings(/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet.rb:159)
        at RUBY.initialize_puppet_settings(uri:classloader:/puppetserver-lib/puppet/server/puppet_config.rb:35)
        at RUBY.initialize_puppet(uri:classloader:/puppetserver-lib/puppet/server/puppet_config.rb:69)
        at RUBY.with_puppet_settings(uri:classloader:/puppetserver-lib/puppet/server/ast_compiler.rb:214)
        at RUBY.mktmpdir(uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/tmpdir.rb:93)
        at RUBY.with_puppet_settings(uri:classloader:/puppetserver-lib/puppet/server/ast_compiler.rb:209)
        at RUBY.compile_for_project_plan(uri:classloader:/puppetserver-lib/puppet/server/ast_compiler.rb:95)
        at RUBY.compile_ast(uri:classloader:/puppetserver-lib/puppet/server/ast_compiler.rb:38)
        at RUBY.compile(uri:classloader:/puppetserver-lib/puppet/server/ast_compiler.rb:21)
        at RUBY.compileAST(uri:classloader:/puppetserver-lib/puppet/server/master.rb:104)
```

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -91,34 +108,130 @@ def self.compile_ast(code, compile_options, boltlib_path)
...
+          Puppet.settings.override_default(:vendormoduledir, '')
+
+          overrides.each do |setting, value|
+            Puppet.settings[setting] = value
+          end
+
+          yield
+
+          # Return settings to their original form
+          Puppet.replace_settings_object(original_settings)
```
donoghuc, line 231: So it appears that now this results in a settings object being returned (instead of the compiled catalog) https://github.com/puppetlabs/puppetserver/pull/2555/commits/3f8239405c58a397435cb254a95f513bd45beae2#diff-6f7f4ebbd198c182ba9b89fe43839b117bc1899b5bb3f7ffeddcd1cc64cd79abR95 

I think instead of doing work after the `yield` we just want to use `ensure`. This will allow the catalog to still be bubbled up from this block and ensure that we always redo the settings.

McdonaldSeanp: Oh, yeah heh

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -91,34 +106,153 @@ def self.compile_ast(code, compile_options, boltlib_path)
...
+          original_settings = Puppet.settings
+        end
+
+        Dir.mktmpdir('bolt_compile') do |dir|
+          # This config sets all required settings to an empty dir
+          # where puppet won't load things like puppet.conf
+          puppet_config = {}
+          Puppet::Settings::REQUIRED_APP_SETTINGS.each do |setting|
+            puppet_config[setting] = dir
+          end
```
donoghuc, line 219:
```suggestion
puppet_config = Puppet::Settings::REQUIRED_APP_SETTINGS.each_with_object({}) {|setting, acc| acc[setting] = dir}
```

McdonaldSeanp: I'm fine being in the minority on this, but personally I find this pattern less readable than defining the var beforehand and not using each_with_object. I get why we would use this, and again I'm cool if everyone disagrees, I just don't think it reads as well.

McdonaldSeanp:
Travis failure appears to be sec related

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
Magisus: I agree with Sean tbh.

Magisus:
If you rebase this, the security scanner flag should go away now.

Magisus:
To be replaced by a different one, yayy

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
donoghuc: I dont care too much. It is a method that does exactly a very common thing. But i also like list comprehensions in python. So take my opinion with a grain of salt.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -937,26 +940,66 @@
...
+   * modulepath
+   * hiera config
+   * project_name"
+  [request-options
+   versioned-project
+   bolt-projects-dir
+   multithreaded]
+  (let [project-root (file-serving/get-project-root bolt-projects-dir versioned-project)
+        project-config (file-serving/read-bolt-project-config project-root)]
+    (-> request-options
```
donoghuc, line 955: Given there is no logic around whether or not to actually add a KV to the map, why not just do a single assoc? I feel like multiple with the threading macro is useful when you conditionally want to add a pair, but otherwise is just kind of redundant.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -91,34 +106,153 @@ def self.compile_ast(code, compile_options, boltlib_path)
...
+            acc[param_hash.keys.first] = param_hash.values.first
+          end
+        else
+          compile_options['variables']['values']
+        end
+      end
+      private_class_method :ordered_plan_vars
+
+      # Create a puppet settings context that can compile for projects
+      def self.with_puppet_settings_for_projects(overrides = {}, multithreaded)
```
tvpartytonight, line 207: Instead of passing `multithreaded` all the way from the clojure config, we could we infer it from the class type of the `Puppet.settings` object instead. If `multithreaded` is on, then it will be a `Puppet::Server::Settings` class; otherwise, it will be a `Puppet::Settings` class.

You have already done the hard work to thread it through and it looks correct/working, but it might be simpler to just deduce it within the Ruby worker itself.

tvpartytonight: Alternatively, you could just change this to a instance variable so it is available for AST compilation:
https://github.com/puppetlabs/puppetserver/blob/02ffd0ff877f5e83398dac9bb6c4d1c6827b1aa6/src/ruby/puppetserver-lib/puppet/server/master.rb#L37

src/clj/puppetlabs/services/master/master_core.clj
McdonaldSeanp: Good point, I had gone straight to threading but you're right that it makes more sense to just use the one assoc.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
McdonaldSeanp: I thought about inferring multithreaded from the type, but I felt like that coupled together the multithreaded setting with the class definition and figured it would be at least a little better to keep those things independent where we can.

I do like the instance variable option though, I hadn't looked through the master class enough to realize I could just pass it in there.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -91,34 +106,153 @@ def self.compile_ast(code, compile_options, boltlib_path)
...
+          # Actually create the settings context
+          if multithreaded
+            # run generate_standalone_local_settings to pass in the empty
+            # puppet config to the thread local settings object
+            Puppet.settings.generate_standalone_local_settings(puppet_config)
+          else
+            # Replace the existing puppet settings object with one that has 'empty' configuration
+            # (i.e. configuration pointed at an empty dir where puppet won't load things like
+            # puppet.conf)
+            Puppet.replace_settings_object(Puppet::Server::Settings.new(global_settings: [],
```
tvpartytonight, line 230: I'm not sure that using this is a viable strategy here, but I don't know with 100% certainty.

When it is used in the `Puppet::Server::Master` class, it is only used in initialization, before we serve any catalogs. In that case, we know it is safe to change out the settings object because it is not accessed by anything else at that point.

However, that is not the case for usage with the compile endpoint; concurrent catalog requests could be coming in. I believe that the changing the settings object will affect other jruby-workers from the pool, leaking the config used for a particular project to other requests.

fwiw, here is the note I left in puppet repo:
https://github.com/puppetlabs/puppet/blob/2cca41ff4911b8822227f64a341be5c6ac327191/lib/puppet.rb#L67-L73

McdonaldSeanp: My understanding was that in single threaded mode each jruby instance's environment (including the settings for puppet) are completely isolated from other jruby instances. That's why I figured I could use this here: since in single threaded mode the AST compile should be using a jruby instance isolated from other jruby instances we can temporarily replace the settings object because nothing else should be able to use the instance while the compile has control of it's jruby.

McdonaldSeanp: For multithreaded it uses different logic that leverages `generate_standalone_local_settings` to put in the settings we want in the same way catalog requests do in multithreaded mode.

src/ruby/puppetserver-lib/puppet/server/settings.rb
```diff
@@ -30,6 +30,22 @@ def generate_new_local_settings
...
 
+  # This method does the identical operation to generate_new_local_settings with
+  # the exception that it uses config passed in as a parameter rather than the value
+  # of @puppet_config.
+  #
+  # This function is used to create a compile context in which we can compile
+  # things for a bolt project. In that scenario we need to set all puppet settings
+  # such that puppet won't read things like puppet.conf, so we pass in a config
+  # seperate from the one used for normal puppet compilation.
+  def generate_standalone_local_settings(standalone_config)
```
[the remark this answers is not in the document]
McdonaldSeanp: I think my thought process was mostly about being risk averse about these changes, so I created a new function that would only be used in the AST compiler right now and doesn't have any potential to create issues elsewhere. It's probably not particularly dangerous just to change the original function though, since I don't think anything in the catalog compiles actually calls this directly: it looks like in multithreaded catalog compiles would just rely on the behavior of `[ ]` to create the thread local settings object

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
tvpartytonight: Yeah, I think you are right; as long as the jruby instance gets put back the same way it was received, I think we will be good. However, as written now, exceptions bubbling up through the `yield` will block the way the settings are put back at the end of this method; I think it should be wrapped in an ensure block?

McdonaldSeanp: Yeah, there's an ensure there already: https://github.com/puppetlabs/puppetserver/pull/2555/files#diff-6f7f4ebbd198c182ba9b89fe43839b117bc1899b5bb3f7ffeddcd1cc64cd79abR244

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -937,26 +940,62 @@
...
+   * hiera config
+   * project_name"
+  [request-options
+   versioned-project
+   bolt-projects-dir]
+  (let [project-root (file-serving/get-project-root bolt-projects-dir versioned-project)
+        project-config (file-serving/read-bolt-project-config project-root)]
+    (assoc request-options "project_root" project-root
+                           "modulepath" (map #(str project-root "/" %) (file-serving/get-project-modulepath project-config))
+                           "hiera_config" (str project-root "/" (:hiera-config project-config))
```
donoghuc, line 956: I really think we need to default to `hiera.yaml` here. I think think bolt defaults are going to be very difficult to keep track of, but one like this seems very hard to keep track of. Also, we should probably forbid absolute paths or trickery like `~,..` etc around this stuff.

donoghuc: for now, i would almost reccomend we *only* accept a hiera.yaml at a project root.

donoghuc (changes requested):
In addition to handling the default for hiera config files in a project we will need to figure out how to use the pdb client.

Currently the compile for a plan like:
```
plan run_plan_apply::pdb_query(TargetSpec $targets) {
  return apply($targets) {
    $certs = puppetdb_query('nodes[certname] {}')
    notify { "found ${certs}": }
  }
}


```
blows up with with something like:
```
2021-07-28T23:14:25.692Z ERROR [qtp1225671935-62] [puppetserver] Puppet Evaluation Error: Error while evaluating a Function Call, /opt/puppetlabs/server/apps/p
uppetserver/tmp/bolt_compile20210728-15590-m4zjnp/ssl/certs/thorny-dusting.delivery.puppetlabs.net.pem (No such file or directory) (file: /opt/puppetlabs/serve
r/data/orchestration-services/projects/my_project_238ab968ac65f330adbbfc0cc6a0ccf60231e305/modules/run_plan_apply/plans/pdb_query.pp, line: 3, column: 14) on n
ode waxy-lavender.delivery.puppetlabs.net_1
```

Could it be that somehow the settings we are clearing out makes us not be able to use puppet's pdb client anymore in the context of a project compile? Perhaps we need to look in to how to not clobber that or at least carry it over?

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -34,91 +34,174 @@ def self.compile_ast(code, compile_options, boltlib_path)
...
+      def self.compile_for_project_plan(code, compile_options, boltlib_path, multithreaded)
+        # We do not set the certname puppet setting directly in project plans because
+        # these compiles leverage with_puppet_settings, which temporarily moves away
+        # all currently exisitng settings from the puppet context and replaces them
+        # with empty ones. Anything set directly with the Puppet[:setting_name] would
+        # be overwritten when with_puppet_settings is called.
+        #
+        # Instead, project plan compiles set the node_name_value below by passing
+        # that setting with the puppet_settings hash to with_puppet_settings.
+        puppet_settings = {
```
donoghuc, line 60: Uhh, i think we forgot to actually use this...

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -34,91 +34,164 @@ def self.compile_ast(code, compile_options, boltlib_path)
...
+        }
+        bolt_inv = Bolt::ApplyInventory.new(fake_config)
+        bolt_project = Struct.new(:name, :path, :load_as_module?).new(compile_options['project_name'],
+                                                                      compile_options['project_root'],
+                                                                      true)
+        puppet_overrides = {
+          bolt_inventory:  bolt_inv,
+          bolt_project:    bolt_project,
+          node_name_value: compile_options['certname'],
+          hiera_config:    compile_options['hiera_config']
```
donoghuc, line 83: For some reason it almost looks like `Puppet.override` does not adequately set this. We may just have to do `Puppet.settings[setting]` for node_name_value and hiera_config.

src/ruby/puppetserver-lib/puppet/server/ast_compiler.rb
```diff
@@ -34,91 +34,167 @@ def self.compile_ast(code, compile_options, boltlib_path)
...
-
-          env_conf = {
-            pre_modulepath: boltlib_path,
-            envpath: Puppet[:environmentpath],
-            facts: compile_options['facts']['values'],
+      def self.compile_for_project_plan(code, compile_options, boltlib_path, multithreaded)
+
+        original_hiera_config = Puppet[:hiera_config]
+
+        Puppet[:node_name_value] = compile_options['certname']
```
donoghuc, line 52: Do we not need to reset this back to original?

donoghuc: A comment couldnt hurt around why we have to modify these settings directly instead of with Puppet.override :grimacing:

donoghuc (commented):
Do we still need to thread through keeping track of `multithreaded`?

donoghuc (approved):
I have run a suite of acceptance tests in https://github.com/puppetlabs/orchestrator/pull/1778 (including pdb and hiera tests) against this for both environment and project based plans. :+1: very happy this ended up being so simple.

spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
```diff
@@ -135,5 +149,41 @@ def find_notify(catalog, title: nil)
         end
       end
     end
+
+    context 'when running a project plan' do
+      let(:boltlib_path) { [] }
+      context 'single threaded' do
```
tvpartytonight, line 155: nit: is this single-threaded context here still helpful?

Magisus, line 155: I notice this PR no longer has any special handling for multithreaded. If we're not going to test that mode, but we still sort of expect this to work there, I think we should drop this context wrapper. Fwiw I don't think there's a super good way to actually test it in that mode, so I'm fine not having it, but having this context makes it look like there's something special about single threaded.

If, however, we _don't_ think this works in multithreaded, we should perhaps be more aggressive about preventing people from using it in that mode.

Magisus (approved):
So long as this works for your usecase, I think it looks alright in the wider context.

spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
```diff
@@ -135,5 +149,41 @@ def find_notify(catalog, title: nil)
...
+        it 'returns puppet settings back to normal' do
```
tvpartytonight, line 156: Is this test still necessary? It seems like not...I think we are not munging the settings any more?

spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
donoghuc: Deleted the context and the redundant settings expectations. This is currently blocking a bunch of other work. Can we ticket out investigating the multithreaded bit so we can get this merged and start really testing it?

spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
donoghuc: deleted.

spec/puppet-server-lib/puppet/jvm/ast_compiler_spec.rb
Magisus: Oh yeah, I don't think it's very important right this minute, I just wanted to understand what the current expectation is.

Magisus (approved):

PR #2558, opened 2021-07-22
(SERVER-2740) Update GEM version to 1.11.0

Tu2607:


Magisus (approved):

mwaggett:
Would you mind updating the commit message to include what is being added with this gem bump? Something like 'Update puppetserver-ca gem, which includes a new `prune` action' would be sufficient, imo.

mwaggett (approved):

PR #2561, opened 2021-07-23
(SERVER-2509) Prevent duplicate entries in Puppet's CA CRL

Tu2607:
This commit include changes made to the function 'revoke-existing-certs' that will now check the list of serials to be added to the revoked lists for possible duplicate entries.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -1628,28 +1628,45 @@
     (fs/delete csr-path)
     (log/debug (i18n/trs "Removed certificate request for {0} at ''{1}''" subject csr-path))))
 
+(schema/defn filter-already-revoked-serials :- [schema/Int]
+  "Given a list of serials and Puppet's CA CRL, returns vector of serials with
+   any already-revoked serials removed."
+  [serials :- [schema/Int]
+   crl :- X509CRL]
+  (let [crl-revoked-list (.getRevokedCertificates crl)
+        existed-serials (set (map #(.getSerialNumber %) crl-revoked-list))
+        duplicate-serials (set/intersection (set serials) existed-serials)]
+        (when (> (count duplicate-serials) 0)
+          (doseq [serial duplicate-serials]
+            (log/debug (i18n/trs "Certificate with serial {0} is already revoked." serial))))
+        (vec (set/difference (set serials) existed-serials))))
+
 (schema/defn revoke-existing-certs!
   "Revoke the subjects' certificates. Note this does not destroy the certificates.
    The certificates will remain in the signed directory despite being revoked."
   [{:keys [signeddir cacert cacrl cakey infra-crl-path
            infra-node-serials-path enable-infra-crl]} :- CaSettings
    subjects :- [schema/Str]]
-  (let [serials (map #(-> (path-to-cert signeddir %)
-                       (utils/pem->cert)
-                       (utils/get-serial))
-                  subjects)
-        serial-count (count serials)
-        [our-full-crl & rest-of-full-chain] (utils/pem->crls cacrl)
-        new-full-crl (utils/revoke-multiple our-full-crl
-                                            (utils/pem->private-key cakey)
-                                            (.getPublicKey (utils/pem->ca-cert cacert cakey))
-                                            serials)
-        new-full-chain (cons new-full-crl (vec rest-of-full-chain))]
-    (write-crls new-full-chain cacrl)
-    (log/info (i18n/trsn "Revoked 1 certificate: {1}"
-                         "Revoked {0} certificates: {1}"
-                         serial-count
-                         (str/join ", " subjects)))
+  (let [[our-full-crl & rest-of-full-chain] (utils/pem->crls cacrl)
+        serials (filter-already-revoked-serials (map #(-> (path-to-cert signeddir %)
+                                                          (utils/pem->cert)
+                                                          (utils/get-serial))
+                                                          subjects)
+                                                our-full-crl)
+        serial-count (count serials)]
+    (if (= 0 serial-count)
+      (log/info (i18n/trs "No revoke action needed. The certs are already in the CRL."))
+      (let [new-full-crl (utils/revoke-multiple our-full-crl
+                                                (utils/pem->private-key cakey)
+                                                (.getPublicKey (utils/pem->ca-cert cacert cakey))
+                                                serials)
+            new-full-chain (cons new-full-crl (vec rest-of-full-chain))]
+        (write-crls new-full-chain cacrl)
+        (log/info (i18n/trsn "Revoked 1 certificate: {1}"
+                             "Revoked {0} certificates: {1}"
+                             serial-count
+                             (str/join ", " subjects)))))
+
 
     ;; Publish infra-crl if an infra node is getting revoked.
     (when (and enable-infra-crl (fs/exists? infra-node-serials-path))
@@ -1657,13 +1674,16 @@
             infra-revocations (vec (set/intersection infra-nodes (set serials)))]
         (when (seq infra-revocations)
           (let [[our-infra-crl & rest-of-infra-chain] (utils/pem->crls infra-crl-path)
-                new-infra-crl (utils/revoke-multiple our-infra-crl
-                                                     (utils/pem->private-key cakey)
-                                                     (.getPublicKey (utils/pem->ca-cert cacert cakey))
-                                                     infra-revocations)
-                full-infra-chain (cons new-infra-crl (vec rest-of-infra-chain))]
-            (write-crls full-infra-chain infra-crl-path)
-            (log/info (i18n/trs "Infra node certificate(s) being revoked; publishing updated infra CRL"))))))))
+                new-infra-revocations (filter-already-revoked-serials infra-revocations our-infra-crl)]
+            (if (= 0 new-infra-revocations)
+              (log/info (i18n/trs "No revoke action needed. The infra certs are already in the infra CRL"))
+              (let [new-infra-crl (utils/revoke-multiple our-infra-crl
+                                                         (utils/pem->private-key cakey)
+                                                         (.getPublicKey (utils/pem->ca-cert cacert cakey))
+                                                         new-infra-revocations)
+                    full-infra-chain (cons new-infra-crl (vec rest-of-infra-chain))]
+                (write-crls full-infra-chain infra-crl-path)
+                (log/info (i18n/trs "Infra node certificate(s) being revoked; publishing updated infra CRL"))))))))))
 
 (schema/defn ^:always-validate set-certificate-status!
   "Sign or revoke the certificate for the given subject."
--- a/test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
+++ b/test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
@@ -503,6 +503,25 @@
       (is (true? (revoked? cert1)))
       (is (true? (revoked? cert2))))))
 
+(deftest filter-already-revoked-serials-test
+  (let [crl (-> (get-certificate-revocation-list cacrl)
+                  StringReader.
+                  utils/pem->crl)]
+  (testing "Return an empty vector when all supplied serials are already in CRL"
+    (let [test-serial (vec [4])
+          filtered-serial (filter-already-revoked-serials test-serial crl)]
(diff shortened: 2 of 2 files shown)


Tu2607:
Haven't written/patch the tests from my changes yet.  Will do once I can establish a clear approach.

mwaggett (commented):
This is looking like a good approach to me!

In `revoke-existing-certs!`, you can see that we also operate on what we call the infra CRL. This is a smaller CRL that only contains revocations for "infra nodes", which would be things like compilers, as opposed to also including agents (right @Magisus?). 

While I think it would _probably_ work to use the same deduplicated set of serials when compared with the full CRL, I think it would be more correct to do a separate deduplication using the infra CRL.

I'll save my style comments for later, unless you want them now. 😜

Tu2607:
> This is looking like a good approach to me!
> 
> In `revoke-existing-certs!`, you can see that we also operate on what we call the infra CRL. This is a smaller CRL that only contains revocations for "infra nodes", which would be things like compilers, as opposed to also including agents (right @Magisus?).
> 
> While I think it would _probably_ work to use the same deduplicated set of serials when compared with the full CRL, I think it would be more correct to do a separate deduplication using the infra CRL.
> 
> I'll save my style comments for later, unless you want them now. 😜

Would you say I should create a separate function for preventing duplication with infra-CRL (over in utils file)? Personally, I don't think there is a need to do so since they're using pretty the same code and logic.

Magisus:
I agree that it should probably be its own deduplication, with its own serials. But you should be able to use the same function for both, with a different input for which CRL it acts on.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1628,42 +1628,60 @@
...
                                             (.getPublicKey (utils/pem->ca-cert cacert cakey))
-                                            (utils/prevent-duplicate serials our-full-crl))
+                                            serials)
         new-full-chain (cons new-full-crl (vec rest-of-full-chain))]
     (write-crls new-full-chain cacrl)
-    (log/info (i18n/trsn "Revoked 1 certificate: {1}"
-                         "Revoked {0} certificates: {1}"
-                         serial-count
-                         (str/join ", " subjects)))
+    (if (= 0 serial-count)
```
justinstoller, line 1662: IMO, this conditional should wrap the revoking and writing of the chain as well.

justinstoller: There's no point in writing the same content to disk again (just creates more load/disk contention), and then you don't need to change jvm-ssl-utils at all.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1628,42 +1628,60 @@
     (fs/delete csr-path)
     (log/debug (i18n/trs "Removed certificate request for {0} at ''{1}''" subject csr-path))))
 
+(schema/defn prevent-duplicate :- [schema/Int]
```
mwaggett, line 1631: I would call this something a little more specific, maybe something like `filter-already-revoked-serials`?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1628,42 +1628,60 @@
...
+  "Given a list of serials and Puppet's CA CRL, returns vector of serials with
+   any already-revoked serials removed."
+  [serials :- [schema/Int]
+   crl :- X509CRL]
+  (let [crl-revoked-list (.getRevokedCertificates crl)
+        existed-serials (set (map #(.getSerialNumber %) crl-revoked-list))
+        duplicate-serials (set/intersection (set serials) existed-serials)] ; For info output
+        (when (> (count duplicate-serials) 0)
+          (doseq [x duplicate-serials]
+            (log/info (str "Serial " x " existed in CRL.")))) ; Feedback on where to place the info output.
```
mwaggett, line 1641: We might want to put this just at the debug level? Not sure if users will want/need this info all the time.

I would also maybe say something like "Certificate with serial x has already been revoked." ?

Magisus: I actually kind of like this being at info. It'll help surface how often we try to do duplicate revocations, which is kind of an unknown right now.

Agreed about the rephrasing.

Tu2607:
Sample output of running `puppetserver-ca revoke --certname yin,yang`
```
2021-07-30 11:42:42,357 DEBUG [qtp1013911640-79] [p.p.certificate-authority] Certificate with serial 3 is already revoked.
2021-07-30 11:42:42,357 INFO  [qtp1013911640-79] [p.p.certificate-authority] No revoke action needed.  The certs are already in revoked list.
2021-07-30 11:42:42,383 DEBUG [qtp1013911640-79] [p.p.certificate-authority] Certificate with serial 4 is already revoked.
2021-07-30 11:42:42,383 INFO  [qtp1013911640-79] [p.p.certificate-authority] No revoke action needed.  The certs are already in revoked list.
```
I'm looking into why the `No revoke action needed...` message is repeated twice.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1628,42 +1628,61 @@
...
-                       (utils/get-serial))
-                  subjects)
-        serial-count (count serials)
-        [our-full-crl & rest-of-full-chain] (utils/pem->crls cacrl)
-        new-full-crl (utils/revoke-multiple our-full-crl
+  (let [[our-full-crl & rest-of-full-chain] (utils/pem->crls cacrl)
+        serials (filter-already-revoked-serials (map #(-> (path-to-cert signeddir %)
+                                                          (utils/pem->cert)
+                                                          (utils/get-serial))
+                                                          subjects) our-full-crl)
```
mwaggett, line 1654: I would line up `our-full-crl` with `(map ...)` so it's clear that it's an arg to `filter-already-revoked-serials`.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1628,42 +1628,61 @@
...
+        serial-count (count serials)]
+    (if (= 0 serial-count)
+      (log/info "No revoke action needed.  The certs are already in revoked list.") ; Info output here?
```
mwaggett, line 1657:
```suggestion
      (log/info "No revoke action needed. The certs are already in the CRL.")
```

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1628,42 +1628,61 @@
...
     ;; Publish infra-crl if an infra node is getting revoked.
     (when (and enable-infra-crl (fs/exists? infra-node-serials-path))
       (let [infra-nodes (set (map biginteger (read-infra-nodes infra-node-serials-path)))
             infra-revocations (vec (set/intersection infra-nodes (set serials)))]
         (when (seq infra-revocations)
           (let [[our-infra-crl & rest-of-infra-chain] (utils/pem->crls infra-crl-path)
-                new-infra-crl (utils/revoke-multiple our-infra-crl
+                new-infra-revocations (filter-already-revoked-serials infra-revocations our-infra-crl)]
+            (if (= 0 new-infra-revocations)
+              (log/info "No revoke action needed. The infra certs are already in the revoked list of infra-CRL")
```
mwaggett, line 1678:
```suggestion
              (log/info "No revoke action needed. The infra certs are already in the infra CRL.")
```

src/clj/puppetlabs/puppetserver/certificate_authority.clj
mwaggett, line 1657: I guess we should probably also wrap all these log messages with `i18n/trs`, since we do that in the existing log message.
(This is for potential translation. `trsn` is used when there are different forms of the string depending on some count. Most of these messages don't need that, so `trs` is sufficient.)

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1628,42 +1628,61 @@
...
+                                                          (utils/pem->cert)
+                                                          (utils/get-serial))
+                                                          subjects) our-full-crl)
+        serial-count (count serials)]
+    (if (= 0 serial-count)
+      (log/info "No revoke action needed.  The certs are already in revoked list.") ; Info output here?
+      (let [new-full-crl (utils/revoke-multiple our-full-crl
                                             (utils/pem->private-key cakey)
                                             (.getPublicKey (utils/pem->ca-cert cacert cakey))
                                             serials)
```
mwaggett, line 1661: these should align with `our-full-crl`

mwaggett (commented):
This is looking good! Mostly just nitpicks from me at this point, and then we'll need tests. 😊

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -503,6 +503,22 @@
...
+(deftest filtering-serials-to-be-add-to-revoked-list
+  (testing "The filtering function return an empty set"
+    (let [crl (-> (get-certificate-revocation-list cacrl)
+                  StringReader.
+                  utils/pem->crl)
+          test-serial (vec [4])]
+      (is (empty? (filter-already-revoked-serials test-serial crl)))))
+
+  (testing "The filtering function return a set of serials not yet in CRL"
+    (let [crl (-> (get-certificate-revocation-list cacrl)
```
mwaggett, line 515: since you're using the same CRL for both of these tests, you could put this in a `let` just inside the `deftest`, before each of the `testing` blocks.

mwaggett (commented):
This looks really good! Are you planning any other tests?

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
Tu2607: Definitely! I'm looking into whether I should add tests related to revoking but it seems that the existing tests already cover most of the scenario.

mwaggett (approved):
I see this is still in Draft mode - did you want to mark it ready or are there still changes you're planning to make?

Also would probably be good to squash your commits when you're ready.

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -503,6 +503,24 @@
       (is (true? (revoked? cert1)))
       (is (true? (revoked? cert2))))))
 
+(deftest filtering-serials-to-be-add-to-revoked-list
```
mwaggett, line 506: We generally try to match the test names with the functions they're testing:
```suggestion
(deftest filter-already-revoked-serials-test
```

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -503,6 +503,24 @@
...
+  (let [crl (-> (get-certificate-revocation-list cacrl)
+                  StringReader.
+                  utils/pem->crl)]
+  (testing "The filtering function return an empty vector"
```
mwaggett, line 510:
```suggestion
  (testing "returns an empty vector when all supplied serials are already in CRL"
```

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -503,6 +503,24 @@
...
+    (let [test-serial (vec [4])]
+      (is (empty? (filter-already-revoked-serials test-serial crl)))))
+
+  (testing "The filtering function return a vector of serials not yet in CRL"
```
mwaggett, line 514:
```suggestion
  (testing "returns a vector of serials not yet in CRL"
```

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -503,6 +503,24 @@
...
+    (let [test-serial (vec [1 2 3 4])
+          filtered-serial (filter-already-revoked-serials test-serial crl)]
+      (is (true? (= (sort filtered-serial) [1 2 3])))))
+
+  (testing "Deduplicate the vector of serials to be add in CRL"
```
mwaggett, line 519:
```suggestion
  (testing "Deduplicates the vector of serials to be revoked"
```

mwaggett (approved):
This looks good to me! My comments are really nitpicks at this point.

Not sure if anyone else wanted to take a look.

mwaggett (approved):

PR #2566, opened 2021-08-04
[IAC-1685] - package dropsonde module and gem up with puppet server

adrianiurca:


resources/ext/build-scripts/install-dropsonde.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -x
+set -e
+
+FORGE_API="forgeapi-cdn.puppet.com"
```
mwaggett, line 6: I think this should just be `forgeapi.puppet.com`? Not sure where the `-cdn` bit comes from.

resources/ext/build-scripts/install-dropsonde.sh
```diff
@@ -0,0 +1,15 @@
...
+MODULE_SLUG="puppetlabs-dropsonde"
+MODULE_NAME="dropsonde"
+MODULE_VERSION="0.0.5"
+
+curl -O "https://${FORGE_API}/v3/files/${MODULE_SLUG}-${MODULE_VERSION}.tar.gz"
+mkdir "${DESTDIR}/opt/puppetlabs/server/data/puppetserver/puppetserver_modules"
```
mwaggett, line 12: Where does `$DESTDIR` come from?

mwaggett: also might be good to put `${DESTDIR}/opt/puppetlabs/server/data/puppetserver/puppetserver_modules` into a variable for re-use.

resources/ext/build-scripts/install-dropsonde.sh
```diff
@@ -0,0 +1,15 @@
...
+tar -xvf "${MODULE_SLUG}-${MODULE_VERSION}.tar.gz" -C "${DESTDIR}/opt/puppetlabs/server/data/puppetserver/puppetserver_modules"
+mv "${DESTDIR}/opt/puppetlabs/server/data/puppetserver/puppetserver_modules/${MODULE_SLUG}-${MODULE_VERSION}" "${DESTDIR}/opt/puppetlabs/server/data/puppetserver/puppetserver_modules/${MODULE_NAME}"
```
mwaggett, line 14: What will happen here on upgrades? I think we might need to remove the unversioned `$MODULE_NAME` directory before we can move the newly-installed version over.

resources/ext/build-scripts/install-dropsonde.sh
adrianiurca: All things related to the Forge API I take from here https://forgeapi.puppet.com/#section/Features.
`forgeapi-cdn.puppet.com` supports the IPv6 domain.

justinstoller:
Does this mean you now need an internet connection to successfully install Puppet Server?

resources/ext/build-scripts/install-dropsonde.sh
adrianiurca: My inspiration was from https://github.com/puppetlabs/puppetserver/blob/5f51e1a301312327b18bb2ffa565b1642ff7c4c6/resources/ext/build-scripts/install-vendored-gems.sh#L17 the script used to install vendored gems.
`DESTDIR` i think is an environment variable.
That is a good idea to use a variable for re-use.

resources/ext/build-scripts/install-dropsonde.sh
adrianiurca: that step is required when a module is installed.
This script reproduces `puppet module install puppetlabs-dropsonde --version 0.0.5`
How does the upgrade look like? It's a script?
If it is a script that runs over an installed puppetserver the command 
`/opt/puppetlabs/puppet/bin/puppet module install puppetlabs-dropsonde --modulepath /opt/puppetlabs/server/data/puppetserver/puppetserver_modules` should work.

Magisus:
Ugh yes 👎 This isn't going to fly.

Also been talking to @adrianiurca about not using cron, which we had discussed before, and using tk-scheduler instead. But that's moot if we still don't have a good way to make the install happen.

resources/ext/build-scripts/install-dropsonde.sh
Magisus: I think this comes down to how the package would handle differences in the module's files, right? Do you know how that works, Molly?

Magisus:
We're looking into failing gracefully if the gem can't be downloaded.

resources/ext/build-scripts/install-dropsonde.sh
mwaggett: @Magisus what do you mean by 'how the package would handle differences in the module's files'?

My concern was around the `mv` command and how it behaves when the destination exists already. I'm pretty sure this will put the versioned module _into_ `$DESTDIR/opt/puppetlabs/server/data/puppetserver/puppetserver_modules/$MODULE_NAME`, rather than renaming it, if that directory exists already and is non-empty.

mwaggett:
superseded by https://github.com/puppetlabs/puppetserver/pull/2591

PR #2568, opened 2021-08-05
(SERVER-3020) Improve certificate_status endpoint speed when querying CSR

Tu2607:
Improving the speed of querying for CSR statuses by dividing up the process.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -1567,16 +1567,12 @@
    :serial_number (-> cert
                     (.getSerialNumber))})
 
-(schema/defn get-certificate-status*
-  [signeddir :- schema/Str
-   csrdir :- schema/Str
-   crl :- CertificateRevocationList
-   subject :- schema/Str]
-  (let [is-cert?            (fs/exists? (path-to-cert signeddir subject))
-        cert-or-csr         (if is-cert?
-                              (utils/pem->cert (path-to-cert signeddir subject))
-                              (utils/pem->csr (path-to-cert-request csrdir subject)))
-        default-fingerprint (fingerprint cert-or-csr "SHA-256")]
+(schema/defn get-cert-or-csr-status*
+  [crl :- CertificateRevocationList
+   is-cert? :- schema/Bool
+   subject :- schema/Str
+   cert-or-csr :- (schema/either Certificate CertificateRequest)]
+  (let [default-fingerprint (fingerprint cert-or-csr "SHA-256")]
     (merge
       {:name              subject
        :state             (certificate-state cert-or-csr crl)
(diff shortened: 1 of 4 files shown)


Tu2607:
I'm not too fond of the `get-queried-state-statuses` function.  I feel like there could be a nice way to modify the `get-certificate-statuses` function to filter for `requested` state.  Anyways, thanks for reviewing! 😄

mwaggett:
so yeah, I agree that what you have is a little funky / confusing to read..

I'm wondering if it would be better put the conditional for whether the `state` is "requested" into the `filter-by-certificate-state` function and then have two separate functions for getting CSR statuses vs. (signed or revoked) certificate statuses that we call depending on the passed `state`. 

Does that make sense?

justinstoller:
Maybe, we could change `get-certificate-status*` to take the path to the "cert" to read as well as the value of `is-cert?` since the caller knows both of those. Then `get-certificate-status*` doesn't need to know both the signeddir and csrdir.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1598,26 +1594,52 @@
...
+(schema/defn ^:always-validate get-CSRs-statuses :- [CertificateStatusResult]
+  "Get the status of only certificate requests by only going
+   through csrdir."
+  [csrdir :- schema/Str
+   crl :- CertificateRevocationList]
+  (let [pem-pattern   #"^.+\.pem$"
+        all-subjects  (map #(fs/base-name % ".pem") (fs/find-files csrdir pem-pattern))
+        all-CSRs      (map #(utils/pem->csr (path-to-cert-request csrdir %)) all-subjects)
+        is-cert?      false]
+    (map (partial get-certificate-status* crl is-cert?) all-subjects all-CSRs)))
```
mwaggett, line 1614: you don't need to put `is-cert?` into the `let` here, you can just use `false` directly.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1598,26 +1594,52 @@
...
+(schema/defn ^:always-validate get-signed-or-revoked-statuses :- [CertificateStatusResult]
+  "Get the status of only signed or revoked certs by only going
+   through signeddir."
+  [signeddir :- schema/Str
+   crl :- CertificateRevocationList]
+  (let [pem-pattern   #"^.+\.pem$"
+        all-subjects  (map #(fs/base-name % ".pem") (fs/find-files signeddir pem-pattern))
+        all-certs     (map #(utils/pem->cert (path-to-cert signeddir %)) all-subjects)
+        is-cert?      true]
+    (map (partial get-certificate-status* crl is-cert?) all-subjects all-certs)))
```
mwaggett, line 1625: you don't need to put `is-cert?` into the `let` here, you can just use `true` directly.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1598,26 +1594,52 @@
...
+        (get-certificate-status* crl is-cert? subject csr)))))
+
+(schema/defn ^:always-validate get-CSRs-statuses :- [CertificateStatusResult]
+  "Get the status of only certificate requests by only going
+   through csrdir."
+  [csrdir :- schema/Str
+   crl :- CertificateRevocationList]
+  (let [pem-pattern   #"^.+\.pem$"
+        all-subjects  (map #(fs/base-name % ".pem") (fs/find-files csrdir pem-pattern))
+        all-CSRs      (map #(utils/pem->csr (path-to-cert-request csrdir %)) all-subjects)
```
mwaggett, line 1612: (nit)
```suggestion
        all-csrs      (map #(utils/pem->csr (path-to-cert-request csrdir %)) all-subjects)
```

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1598,26 +1594,52 @@
...
+        is-cert?      false]
+    (map (partial get-certificate-status* crl is-cert?) all-subjects all-CSRs)))
+
+(schema/defn ^:always-validate get-signed-or-revoked-statuses :- [CertificateStatusResult]
```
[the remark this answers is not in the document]
mwaggett: I guess that would also imply that we should change the names of `get-certificate-status*` and `get-certificate-status` as well. Open to other opinions on this, but just thinking it might be clearer to not use 'certificate' to mean 'cert or CSR'.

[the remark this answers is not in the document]
mwaggett: Anyway, I think there's some simplification to be done around the `get-CSRs-statuses` and `get-signed-or-revoked-statuses` functions. They're nearly identical, other than one using the `csrdir` and one using the `signeddir`. Wondering if it makes sense for that directory to just be an argument to something.. Trying not to just do it for you haha so let me know if you want to chat more about what I had in mind.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
Tu2607: yeah. I was worried that the code would be hard to interpret with so I created a binding for clarity.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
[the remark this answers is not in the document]
Tu2607: Yeah I split them up into two separate function mostly because I wanted to be clear. The other approach is just having one function basically take either `csrdir` or `signeddir` and a boolean that indicate whether we're dealing with a cert of csr and make the function call to `get-certificate-status*` from there. I do see the repetitiveness but thought that I wanted to be clear.

justinstoller:
:+1: from me, I found the diff a bit hard to read, but looking at the code outside of a diff it looked good.

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -391,7 +391,7 @@
       (->
         (if (some #(= queried-state %) ["requested" "signed" "revoked"])
```
Tu2607, line 392: Found a very interesting bug (maybe?) here.  Apparently `queried-state` is `nil` when I ran the command `puppetserver-ca list` so the function `filter-by-certificate-state` is not getting called.  I checkout a fresh 6.x branch without any of my changes and ran the same command again and `filter-by-certificate-state` is not getting called as well.  I tried to log the value of `queried-state` before the conditional and it is always `nil`, with or without the `--all` flag.  So at of right now, any of my changes are not effective because of this conditional always being false.

Magisus (approved):

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1592,32 +1588,48 @@
...
   "Get the status of all certificates and certificate requests."
   [{:keys [csrdir signeddir cacert cacrl cakey]} :- CaSettings]
-  (let [crl           (utils/pem->ca-crl cacrl (utils/pem->ca-cert cacert cakey))
-        pem-pattern   #"^.+\.pem$"
-        all-subjects  (map #(fs/base-name % ".pem")
-                           (concat (fs/find-files csrdir pem-pattern)
-                                   (fs/find-files signeddir pem-pattern)))]
-    (map (partial get-certificate-status* signeddir csrdir crl) all-subjects)))
+  (let [crl       (utils/pem->ca-crl cacrl (utils/pem->ca-cert cacert cakey))
+        all-csr   (get-cert-or-csr-statuses csrdir crl false)
```
mwaggett, line 1620: (nit)
```suggestion
  (let [crl (utils/pem->ca-crl cacrl (utils/pem->ca-cert cacert cakey))
        all-csrs (get-cert-or-csr-statuses csrdir crl false)
```

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1592,32 +1588,48 @@
...
+    (get-cert-or-csr-status* crl is-cert? subject cert-or-csr)))
+
+(schema/defn ^:always-validate get-cert-or-csr-statuses :- [CertificateStatusResult]
+  "Get the statuses of either all the CSR or all the certificate."
+  [dir :- schema/Str
+   crl :- CertificateRevocationList
+   fetch-cert? :- schema/Bool]
+  (let [pem-pattern      #"^.+\.pem$"
+        all-subjects     (map #(fs/base-name % ".pem") (fs/find-files dir pem-pattern))
+        all-certs-or-csr (if fetch-cert?
```
mwaggett, line 1611: (nit)
```suggestion
  (let [pem-pattern #"^.+\.pem$"
        all-subjects (map #(fs/base-name % ".pem") (fs/find-files dir pem-pattern))
        all-certs-or-csrs (if fetch-cert?
```

mwaggett (commented):
This looks good! You're still manually testing, right?

Magisus:
As a general note, we make a point of not aligning things in `let` blocks, because it makes diffs very hard to read if you add a new param that's long, and all the other things have to move to stay aligned.

mwaggett (approved):

mwaggett:
Is this ready to go or are you still doing manual tests?

Tu2607:
we're good to go!

PR #2571, opened 2021-08-05
(maint) Provide a reasonable default for bolt-projects dir

donoghuc:
For some testing it is desireable for puppetserver to serve some project content from some pre-defined project content. By setting a bolt-projects dir we can enable puppetserver to serve project content assuming it is in a proper structure even if it is not actually managed with code manager/file sync.

### DIFF
--- a/docker/puppetserver/puppetserver.conf
+++ b/docker/puppetserver/puppetserver.conf
@@ -72,6 +72,12 @@ http-client: {
     #metrics-enabled: true
 }
 
+# Provide a reasonable default here. Use case is mounting a valid projects volume to have PS serve files from there even if 
+# CM did not actually put them there.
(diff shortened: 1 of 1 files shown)


donoghuc:
Here is a use case https://github.com/puppetlabs/bolt/pull/2962/files#r683840461

lucywyman:
This would be super helpful for the Bolt team too when we pick up https://github.com/puppetlabs/bolt/issues/2920 next week!

lucywyman (approved):
Not sure my approval means much 😅 but this is where I'd expect the default to be.

Magisus (approved):
This is okay with me, don't really see a downside. Going to ask around just a bit before merging.

PR #2576, opened 2021-08-18
allow to install Ruby gems before puppetserver starts

skoef:
I'm running puppetserver from kubernetes and require the vault gem (and dependencies) to be installed before puppetserver starts. This PR allows me to hook into the start process and install those gems.


justinstoller:
This makes sense. Would you want to specify a version for each gem?

We'd maybe want to expose the gemdir as a volume as installing variable things on start up seems less docker-y than ideal. But I don't think we care to ever get this data _out_ again, so I think this is probably the right way to go. @Iristyle has expressed a preference for not adding flags in the past, curious to know his thoughts on this.

skoef:
It would be ideal if puppetserver would honor settings in a Gemfile but other than that: I currently just want the latest version installed.

Iristyle:
I totally understand why you'd want to do this... I'd be a bit concerned about running a `gem install` on every container restart though.

Iristyle:
Also @skoef ... if you're running in k8s, I would just use an `initContainer` for this personally

jonathannewman:
docker configuration is no longer supported in puppetserver.  Closing.

PR #2578, opened 2021-08-19
(main) Update puppetserver-ca GEM version to 2.3.3

Tu2607:
This GEM bump contains:
- Update to the `prune` action so that it only prune a single Puppet's CRL.
- Update logging information of the `prune` action to log the total number of
  certificates in Puppet's CRL.
- Fix a bug when not checking `:query` parameters to be `nil` when create a URL,
  causing an error.


Tu2607:
Will do a merge-up once [6.x branch](https://github.com/puppetlabs/puppetserver/pull/2577) is merged.

mwaggett (approved):

PR #2585, opened 2021-09-28
(SERVER-3050) Allow retrieving facts from any terminus

Magisus:
Previously, the v4 catalog endpoint would only fetch facts (as opposed
to requiring them in the request) if the PuppetDB facts terminus was
configured. This caused issues for people wanting to use extensions of
the PDB terminus, like satellite, because the endpoint would error
claiming that PDB wasn't configured.

This commit removes the restriction on the fact retrieval logic. Now, if
facts are not supplied in the request body, puppetserver will attempt to
fetch them from whatever facts terminus is configured, whether that is
yaml/json (default), PuppetDB, or some custom terminus. So long as the
terminus implements `find`, this will succeed.

### DIFF
--- a/spec/puppet-server-lib/puppet/jvm/compiler_spec.rb
+++ b/spec/puppet-server-lib/puppet/jvm/compiler_spec.rb
@@ -93,11 +93,11 @@ describe Puppet::Server::Compiler do
 
         it 'requests facts from pdb after classifying and attempts to classify again' do
           allow(Puppet::Node::Facts.indirection.terminus).to receive(:name).and_return(:puppetdb)
-          expect(compiler).to receive(:get_facts_from_pdb)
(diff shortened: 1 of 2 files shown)


Magisus:
I think this should do the right thing. I tested it with the default (yaml) terminus, will maybe try this with PDB tomorrow.

justinstoller (approved):
This looks right. Let me know if you decide to try it with PDB or a custom terminus or not. If not I might give it a whirl.

Magisus:
I'm spinning up PE and building a dev package right now to test PDB, in an env like CD4PE would use.

Magisus:
This looked good with the `puppetdb` terminus too.

tvpartytonight (approved):

PR #2586, opened 2021-10-01
(SERVER-3050) Bypass the fact cache in the v4 catalog endpoint

Magisus:
When we stopped explicitly querying the PuppetDB terminus for facts, we
inadventently started using the yaml cache terminus instead of PuppetDB,
even when PuppetDB was configured. However, trusted facts are not stored
in the yaml cache, and the primary use case for this endpoint requires
stored trusted facts to be retrieved from the system. So this commit
updates the fact retrieval to bypass the cache terminus for facts, going
instead directly to the primary terminus.

When PuppetDB is configured, this means we will query PDB for facts,
rather than the default yaml cache, and therefore retrieve its stored
trusted facts.

### DIFF
--- a/src/ruby/puppetserver-lib/puppet/server/compiler.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/compiler.rb
@@ -208,8 +208,10 @@ module Puppet
       end
 
       def get_facts_from_terminus(nodename, environment)
+        # Ignore the cache terminus, which is not guaranteed to store trusted facts
         facts = Puppet::Node::Facts.indirection.find(nodename,
-                                                     {environment: environment})
+                                                     {environment: environment,
+                                                      ignore_cache: true})
 
         # If no facts have been stored for the node, the terminus will return nil
         if facts.nil?
--- a/acceptance/scripts/generic/testrun.sh
+++ b/acceptance/scripts/generic/testrun.sh
@@ -8,7 +8,7 @@ do_init()
     exit -1;
   fi
 
-  bundle exec beaker-hostgenerator $GENCONFIG_LAYOUT > $BEAKER_CONFIG
+  bundle exec beaker-hostgenerator --hypervisor abs $GENCONFIG_LAYOUT > $BEAKER_CONFIG
 
   BEAKER_INIT="bundle exec beaker init --debug"
   BEAKER_INIT="$BEAKER_INIT --type aio"


Magisus:
Does anyone think there's a different way we should approach this? I assume it's intentional that trusted facts are not stored in yaml?

src/ruby/puppetserver-lib/puppet/server/compiler.rb
```diff
@@ -208,8 +208,14 @@ def extract_trusted_facts(request_data, facts)
...
-        facts = Puppet::Node::Facts.indirection.find(nodename,
-                                                     {environment: environment})
+        # Bypass the cache terminus, which is not guaranteed to store trusted facts
+        terminus = Puppet::Node::Facts.indirection.terminus
+        request = Puppet::Indirector::Request.new(terminus.class.name,
+                                                  :find,
+                                                  nodename,
+                                                  nil,
+                                                  environment: environment)
+        facts = terminus.find(request)
```
joshcooper, line 218: I think there’s an option to find to ignore the cache so you don’t need to construct the request

Magisus: Oh okay I wondered about that but wasn't sure. I can look for it.

Magisus: https://github.com/puppetlabs/puppet/blob/ba1e8af0717c6116f77d8665db115887d30c2ff1/lib/puppet/indirector/indirection.rb#L265

Magisus: Updated to use that, tested that it still works 👍

joshcooper:
About loading trusted facts from yaml/etc, I think the concern was we didn't want to accidentally load trusted facts from the wrong place... as in, it's more due to the confusing ways that trusted information can be loaded, as opposed to an explicit design decision to exclude trusted info from yaml/etc.

The issue is avoided when using the v3 catalog endpoint running in puppetserver, because the trusted info is always extracted from the context (which is pushed onto the context in the request handler layer):

https://github.com/puppetlabs/puppet/blob/ba1e8af0717c6116f77d8665db115887d30c2ff1/lib/puppet/indirector/catalog/compiler.rb#L52-L53

And if the trusted info is not present (such as when running `puppet apply`), then it falls back to the block passed to `Puppet.lookup`, which loads trusted info from the local node object.

But for the v4 catalog endpoint, do you want to prefer trusted info that was sent with the request:

https://github.com/puppetlabs/puppetserver/blob/01acfc93749ecaa8214ffc87d4b86381ff738ec5/src/ruby/puppetserver-lib/puppet/server/compiler.rb#L38

And fallback to the terminus if trusted info is missing? Or always use the facts from the terminus?

Magisus:
If trusted fact data is sent with the catalog request, we use that. If not, we try to get it from the terminus. 

Previously we [checked](https://github.com/puppetlabs/puppetserver/blob/4d55cf32cd13cbdd1c7f4e345b68fb39d28627b2/src/ruby/puppetserver-lib/puppet/server/compiler.rb#L178-L182) if PDB was configured and then [used that terminus explicitly](https://github.com/puppetlabs/puppetserver/blob/4d55cf32cd13cbdd1c7f4e345b68fb39d28627b2/src/ruby/puppetserver-lib/puppet/server/compiler.rb#L213-L219). But some users want to use custom termini (e.g. `satellite`), that are wrappers around the PDB terminus, so our explicitness around `puppetdb` made the endpoint unusable for them. We decided to just let the indirector do its thing, to support these extension points.

Magisus:
Also fwiw that process you described (checking the context and then falling back to the local node) is what the `puppetdb` terminus does too. It's why we override the trusted_information in the context when [saving facts](https://github.com/puppetlabs/puppetserver/blob/4d55cf32cd13cbdd1c7f4e345b68fb39d28627b2/src/ruby/puppetserver-lib/puppet/server/compiler.rb#L129-L131) in this endpoint.

justinstoller (approved):

PR #2587, opened 2021-10-04
(maint) Update ca-certificates

Magisus:
Recently a major root certificate expired and had to be replaced. This
caused postgres to fail to install, on platforms where we install PDB.
This commit adds a step to our presuite that fetches new CA certificates,
to ensure that next time a cert is rotated, we will fetch it automatically.


Magisus:
Tried this on a few canary platforms, and they were able to proceed with installing postgres. This doesn't deal with sles because we do not install PDB on that platform (and therefore don't need to install postgres).

justinstoller (approved):

PR #2589, opened 2021-10-06
(maint) Fix ca-certificates update for el6; re-enable PDB

Magisus:
This commit edits the yum repos on RHEL6 to allow us to install a newer
`ca-certificates` package that avoids the expired Let's Encrypt root
cert. This allows us to successfully install postgres, which in turn
allows this commit to also re-enable testing with PDB on EL6.

### DIFF
--- a/acceptance/lib/helper.rb
+++ b/acceptance/lib/helper.rb
@@ -64,6 +64,7 @@ module PuppetServerExtensions
     [
       /debian-8/,
       /debian-9/,
+      /el-6/,
       /el-7/,
       /ubuntu-16.04/,
       /ubuntu-18.04/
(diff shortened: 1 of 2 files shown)


mwaggett (approved):

Magisus:
Still broken, at least on `main`. The DNF module support, which fixes it for 6.x, doesn't work on 7.x because they don't support Postgres 11 (just 10 and 12). Austin said he thought we were supposed to be doing something different: disabling postgres's repos in favor of RHEL's, which do have 11, but that doesn't appear to be working like he thought. See https://github.com/puppetlabs/puppetlabs-puppetdb/pull/337 for details. We're not sure if this requires changes to the `postgresql` module, or just different configuration of it in the PDB module.

PR #2597, opened 2021-11-17
(SERVER-3079) Use updated terminology for puppet settings

Magisus:
In Puppet Server 7 we renamed several settings from `master-*` to
`server-*`. This commit updates our reading of these settings to use the
existing default resolution logic, which check both values before
falling back to the hard-coded default.


Magisus:
Going to make sure this works.

mwaggett (approved):

Magisus:
This works now.

mwaggett (approved):

PR #2598, opened 2021-11-17
(SERVER-3079) Add dropsonde namespace to AOT list

Magisus:


Magisus:
I think we need this, right?

mwaggett (approved):
No idea if we need this or not, but seems like it can't hurt.

Magisus:
I think we actually don't need this, there's a theory we only need things that define TK services or other macros, which this doesn't.

PR #2600, opened 2021-11-29
(SERVER-3079) add Dropsonde wrapper script

binford2k:
This allows users to invoke Dropsonde as `puppetserver dropsonde` and
access the local functionality such as generating reports for their own
usage or to validate what data will be submitted.

### DIFF
--- a/resources/ext/cli/dropsonde.erb
+++ b/resources/ext/cli/dropsonde.erb
@@ -9,7 +9,7 @@ export GEM_PATH=${DROPSONDE_DIR}
 
 if [[ -f "$DROPSONDE_BIN" && -x "$DROPSONDE_BIN" ]]
 then
-    $DROPSONDE_BIN "$@"
+    $PUPPET_AGENT_RUBY $DROPSONDE_BIN "$@"
 else
     echo "Dropsonde does not appear to be installed properly."
 fi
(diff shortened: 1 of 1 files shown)


resources/ext/cli/dropsonde.erb
```diff
@@ -0,0 +1,16 @@
...
+PUPPET_AGENT_RUBY="/opt/puppetlabs/puppet/bin/ruby"
+DROPSONDE_DIR="/opt/puppetlabs/server/data/puppetserver/dropsonde"
+DROPSONDE_BIN="${DROPSONDE_DIR}/bin/dropsonde"
+
+export GEM_HOME=${DROPSONDE_DIR}
+export GEM_PATH=${DROPSONDE_DIR}
+
+if [[ -f "$DROPSONDE_BIN" && -x "$DROPSONDE_BIN" ]]
+then
+    $DROPSONDE_BIN "$@"
```
Magisus, line 12: I think this might need to have `PUPPET_AGENT_RUBY` in front of it. Otherwise it will try to run under JRuby due to a quirk of how it was installed.

binford2k: good catch, that's what I meant to do

Magisus:
I think this is probably right. I can try it out in a bit.

Magisus (approved):
I tried this out in a package build, seems to work as expected.

PR #2603, opened 2022-01-05
(SERVER-3114) Fix authority key ID and alt names of self-signed CA cert

Magisus:
There are two fixes in this PR:

Firstly, for consistency, the auto CA generation code in puppetserver now creates its self-signed CA cert with a `keyid` authority key identifier. This is consistent with what the newer (recommended) chained CA generation code in the CLI tool does.

Secondly, it removes the subject alt names from the self-signed cert, since they're unnecessary, and the fact that its common name is not a valid DNS name caused issues with LibreSSL.

### DIFF
--- a/test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
+++ b/test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
@@ -667,7 +667,12 @@
         (testutils/assert-issuer cert "CN=test ca")
         (testing "has at least one expected extension - key usage"
           (let [key-usage (utils/get-extension-value cert "2.5.29.15")]
-            (is (= #{:key-cert-sign :crl-sign} key-usage))))))
+            (is (= #{:key-cert-sign :crl-sign} key-usage))))
+        (testing "authority key identifier is SHA of public key"
+          (let [ca-pub-key (-> settings :capub utils/pem->public-key)
+                pub-key-sha (pubkey-sha1 ca-pub-key)]
+            (is (= pub-key-sha
+                   (:key-identifier (utils/get-extension-value cert utils/authority-key-identifier-oid))))))))
 
     (testing "cakey"
       (let [key (-> settings :cakey utils/pem->private-key)]
@@ -1424,9 +1429,8 @@
                                                :dns-alt-names dns-alt-names))))))
 
     (testing "basic extensions are created for a CA"
-      (let [serial        42
-            exts          (create-ca-extensions subject-dn
-                                                serial
+      (let [exts          (create-ca-extensions subject-dn
+                                                subject-pub
                                                 subject-pub)
             exts-expected [{:oid      "2.16.840.1.113730.1.13"
                             :critical false
@@ -1436,9 +1440,9 @@
                             :value    {:dns-name [subject]}}
                            {:oid      "2.5.29.35"
                             :critical false
-                            :value    {:issuer-dn     (str "CN=" subject)
-                                       :public-key    nil
-                                       :serial-number (biginteger serial)}}
+                            :value    {:issuer-dn     nil
+                                       :public-key    subject-pub
+                                       :serial-number nil}}
                            {:oid      "2.5.29.19"
                             :critical true
                             :value    {:is-ca true}}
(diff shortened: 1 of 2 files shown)


justinstoller:
For some reason I thought that the CN was deprecated and we needed some kind of alternative name. But I may be conflating the CN's deprecation as a valid DNS name. I may also be wrong, but I thought that @joshcooper had, at some point, brought up the correct replacement alternative name for the CN that was not a DNS altname (might also be misremembering that).

justinstoller:
Also, (assuming my concern about needing to have some kind of alternative name is incorrect - wouldn't surprise me if it was) the SAN change could probably use a test (I'm surprised there isn't one that requires changing). Otherwise, looks good!

Magisus:
I was basing this change off of Josh's comment in the ticket:
> My understanding is that our CA shouldn't add a Subject Alternative Name extension to CA certs (root or intermediate).

Magisus:
There was a test, [here](https://github.com/puppetlabs/puppetserver/blob/3937d10907e3212cac38f5f16e778bae5ef4c4b8/test/unit/puppetlabs/puppetserver/certificate_authority_test.clj#L1417-L1419) that I had to delete. Do you want something explicitly verifying that the resulting cert has no SANs?

Magisus:
I went ahead and added that 🙂

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
```diff
@@ -1407,21 +1408,16 @@
...
-                           {:oid       "2.5.29.17"
-                            :critical false
-                            :value    {:dns-name [subject]}}
                            {:oid      "2.5.29.35"
                             :critical false
-                            :value    {:issuer-dn     (str "CN=" subject)
-                                       :public-key    nil
-                                       :serial-number (biginteger serial)}}
+                            :value    {:issuer-dn     nil
+                                       :public-key    subject-pub
```
joshcooper, line 1419: Does the value of the AuthorityKeyIdentifier extension (with oid `2.5.29.35`) contain the raw public key or is it a value derived from the public key as described in https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.1:

> The value of the keyIdentifier field SHOULD be derived from the public key used to verify the certificate's signature or a method that generates unique values, e.g. Two common methods for generating key identifiers from the public key are described in Section 4.2.1.2.

Where https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.2 describes type 1 and 2:

> Two common methods for generating key identifiers from the public key are:
> 
>       (1) The keyIdentifier is composed of the 160-bit SHA-1 hash of the
>            value of the BIT STRING subjectPublicKey (excluding the tag,
>            length, and number of unused bits).
> 
>        (2) The keyIdentifier is composed of a four-bit type field with
>            the value 0100 followed by the least significant 60 bits of
>            the SHA-1 hash of the value of the BIT STRING
>            subjectPublicKey (excluding the tag, length, and number of
>            unused bits).
>

Magisus: From what I can tell, the extension has the raw key, which is then processed into the SHA1 hash. See [this test](https://github.com/puppetlabs/jvm-ssl-utils/blob/4c47e1bb680c07aa54e90ad987e7bdc753ec2d73/test/puppetlabs/ssl_utils/core_test.clj#L312-L323). I can add a test in this repo to verify that if we want.

Magisus: As indicated in https://tickets.puppetlabs.com/browse/SERVER-2662, we use the first method to generate these IDs.

joshcooper: ah cool, makes sense!

joshcooper (approved):

test/unit/puppetlabs/puppetserver/certificate_authority_test.clj
Magisus: I added a test for this.

Magisus: And this surprised me too when I first noticed it. I had written the first test originally checking for that SHA1, and it failed of course.

PR #2604, opened 2022-01-06
(SERVER-3137) Update ezbake to 2.3.0

Magisus:
This update brings in support for building Debian 11 Bullseye packages.


justinstoller (approved):

Magisus:
They retagged the same ezbake as a Y, I switched this to pulling in that tag instead.

mwaggett (approved):

PR #2607, opened 2022-01-27
(SERVER-3157) Add Snyk GH Action

Magisus:


Magisus:
This doesn't seem to be working right yet.

Magisus:
For reasons that are entirely unclear to me, using the `pull_request_target` trigger does not appear to work. `pull_request` does.

Magisus:
Update: apparently you have to merge actions using the `pull_request_target` trigger before they will run. So I'm going to put this to `pull_request` for now so we can see it go, then flip to `pull_request_target` right before we merge.

rileynewton (changes requested):
I found out from Jeremy that there's a specific puppet-foss token (there's a token for every org now I realize 🤦 ), you'll likely want to switch to that.

Magisus:
I asked him about that several times and he told me there was only one for anything that goes into PE...

.travis.yml
```diff
@@ -21,9 +21,6 @@ jobs:
     - jdk: openjdk11
       env: MULTITHREADED=true
       name: "Java 11 w/ multithreaded"
-    - stage: puppetserver vulnerability scan
-      script:
-        - lein nvd check
```
mwaggett, line 26: Should we also remove the plugin from project.clj and remove the suppression file?

mwaggett (approved):

Magisus:
Flipping this to `pull_request_target` now that Snyk has passed

mwaggett (approved):

PR #2613, opened 2022-02-18
Log config fails on multiple certificates

nod0n:
This PR prevents this script from failing, when there are multiple certificates found in `$SSLDIR`. It just picks the first one and ignores any other certificates.

### DIFF
--- a/docker/puppetserver/docker-entrypoint.d/90-log-config.sh
+++ b/docker/puppetserver/docker-entrypoint.d/90-log-config.sh
@@ -10,7 +10,11 @@ if test -n "${PUPPETSERVER_HOSTNAME}"; then
   certname=${PUPPETSERVER_HOSTNAME}.pem
 else
   echo "* PUPPETSERVER_HOSTNAME: unset"
-  certname=$(cd "${SSLDIR}/certs" && ls *.pem | grep --invert-match ca.pem)
+  # choose first certificate found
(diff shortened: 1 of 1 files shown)


justinstoller:
Thanks, @nod0n !

Yeah, that Travis failure doesn't look related to your change.

I'm wondering why this is failing for you? And if we should go this route or calling something like `puppet config print --section server certname` and printing that (maybe also printing whether or not a cert with that name exists in the certs dir.

nod0n:
I discussed this in slack: https://puppetcommunity.slack.com/archives/C0W298S9G/p1645024274499229 and following
I did generate a certificate to use with puppetboard, but what I use it for isn't really important. Maybe something was wrong with how I did it:
```
./pupperware/bin/puppetserver ca generate --certname puppetboard
```
The generation worked well and I could use the certificates in puppetboard. I just did not delete the files from the pupperware container, or to be more precise from the volume attached to the pupperware puppet container.

nod0n:
> 
> `puppet config print --section server certname`

If that is possible, it might be a good idea. parsing the output of `ls` is generally a bad idea anyways. ;)

bastelfreak:
I don't think selecting the first certificate is a sane approach. That will break depending on the name of other certs. I think something like using the suggest `puppet config print --section server certname` to get the correct name is a better approach.

jonathannewman:
Docker configuration is no longer relevant.  Closing.

PR #2614, opened 2022-02-18
(maint) Add on-push Snyk action, tweak PR action

Magisus:


.github/workflows/snyk_push.yml
```diff
@@ -0,0 +1,27 @@
+---
+name: Snyk Clojure Merge
```
mwaggett, line 2: (nit) to match the PR workflow name
```suggestion
name: Clojure Synk Push Check
```

.github/workflows/snyk_push.yml
```diff
@@ -0,0 +1,27 @@
...
+
+on: push
```
mwaggett, line 4: Why does this need a separate action rather than configuring multiple triggers?
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#using-activity-types-and-filters-with-multiple-events

.github/workflows/snyk_push.yml
```diff
@@ -0,0 +1,27 @@
...
+    - name: Run Clojure Snyk Scan
+      id: scan
+      uses: puppetlabs/security-snyk-clojure-action@v1
+      with:
+        snykToken: ${{ secrets.SNYK_PE_TOKEN }}
+        rproxyKey: ${{ secrets.SEC_PUBLIC_RPROXY_KEY }}
+        rproxyUser: ${{ secrets.SEC_PUBLIC_RPROXY_USER }}
+        snykOrg: 'puppet-enterprise'
+        snykProject: 'puppetserver'
+        snykPolicy: '.snyk'
```
mwaggett, line 24: Should the PR workflow have this policy too? I noticed some of Steve's PRs had this commented out, so wasn't sure what that was about.

.github/workflows/snyk_push.yml
Magisus: I kept them separate because the `push` variety doesn't need the label check, since you can only push branches to the repo if you have write access, so the label isn't necessary. This follows the templates that other teams have been using.

.github/workflows/snyk_push.yml
Magisus: Jon's have left it in, even with no file present. So I copied that.

.github/workflows/snyk_push.yml
mwaggett: ^ that doc suggests you could do
```
on:
  push:
  pull_request_target:
    types: [opened, reopened, synchronize, labeled]
```
and the types would only apply to the PR trigger

Magisus: Right, I'm talking about this conditional: https://github.com/puppetlabs/puppetserver/blob/70250ea6e25157b63894415cbf050506427dcede/.github/workflows/snyk.yml#L11. I think I saw another example somewhere (maybe the vanagon examples?) that had both in one file, but still needed two separate workflows because we want to gate it running more specifically than just on the triggers, in a way that only applies to PRs. Maybe I'm still missing something? But I almost think this is clearer regardless 🤷‍♀️

mwaggett: ahh, okay. ¯\_(ツ)_/¯

mwaggett (approved):

PR #2615, opened 2022-03-01
(maint) Ensure needed container directories are owned by puppet

justinstoller:
The container is created with the puppet user owning these directories
and if a user causes the UID of the puppet user to change than they will
not be readable.

This forces those directories to be owned by the puppet user regardless
of UID on start up.

h/t to jay7x who originally proposed this.

### DIFF
--- a/docker/puppetserver/docker-entrypoint.d/30-set-permissions.sh
+++ b/docker/puppetserver/docker-entrypoint.d/30-set-permissions.sh
@@ -2,3 +2,5 @@
 
 chown -R puppet:puppet /etc/puppetlabs/puppet/
 chown -R puppet:puppet /opt/puppetlabs/server/data/puppetserver/
+chown -R puppet:puppet /etc/puppetlabs/puppetserver/
+chown -R puppet:puppet /var/log/puppetlabs/puppetserver/


docker/puppetserver/docker-entrypoint.d/30-set-permissions.sh
```diff
@@ -2,3 +2,5 @@
 
 chown -R puppet:puppet /etc/puppetlabs/puppet/
 chown -R puppet:puppet /opt/puppetlabs/server/data/puppetserver/
+chown -R puppet:puppet /etc/puppetlabs/puppetserver/
+chown -R puppet:puppet /var/log/puppetlabs/puppetserver/
```
Iristyle, line 6: I think this makes sense... also FYI @jpartlow

Iristyle: We may not have had to do this on the pe-puppetserver side because we were much more intentional about how to manage state there. We never came back to the puppetserver container to make it consistent with everything else, so it's a bit of the wild west here.

Magisus (approved):

docker/puppetserver/docker-entrypoint.d/30-set-permissions.sh
jpartlow: Yeah, looks like in the pe-puppetserver Dockerfile we explicitly chown /etc/puppetlabs and /opt/puppetlabs/server/data (which is where the logs end up in that image).

PR #2617, opened 2022-03-02
Release docker image for arm64

odidev:
Modified **Dockerfile**, **Makefile** and **.github/workflows/docker.yml** file to release the docker image for both the platforms using buildx.
Signed-off-by: odidev [odidev@puresoftware.com](mailto:odidev@puresoftware.com)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -22,6 +22,12 @@ jobs:
         with:
           ruby-version: 2.6.x
       - run: gem install bundler
+      - uses: actions/checkout@v2
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
       - name: Build container
         working-directory: docker
         run: |
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -77,9 +77,9 @@ test: prep
 		rspec --options puppetserver/.rspec spec
 
 push-image: prep
-	@docker push puppet/puppetserver:$(VERSION)
+	 @docker buildx build --platform linux/arm64,linux/amd64 -t $(NAMESPACE)/puppetserver:$(VERSION) -f puppetserver/Dockerfile --push $(PWD)/..
 ifeq ($(IS_LATEST),true)
-	@docker push puppet/puppetserver:$(LATEST_VERSION)
+	 @docker buildx build --platform linux/arm64,linux/amd64 -t $(NAMESPACE)/puppetserver:$(LATEST_VERSION) -f puppetserver/Dockerfile --push $(PWD)/..
 endif
 
 push-readme:
--- a/docker/puppetserver/Dockerfile
+++ b/docker/puppetserver/Dockerfile
@@ -8,7 +8,8 @@ ARG UBUNTU_CODENAME=bionic
 FROM ubuntu:18.04 as base
 
 ARG PACKAGES=ca-certificates\ git
-ARG DUMB_INIT_VERSION="1.2.2"
+ARG DUMB_INIT_VERSION="1.2.5"
+ARG TARGETARCH
 
 LABEL org.label-schema.maintainer="Puppet Release Team <release@puppet.com>" \
       org.label-schema.vendor="Puppet" \
@@ -47,7 +48,7 @@ EXPOSE 8140
 ENTRYPOINT ["dumb-init", "/docker-entrypoint.sh"]
 CMD ["foreground"]
 
-ADD https://github.com/Yelp/dumb-init/releases/download/v"$DUMB_INIT_VERSION"/dumb-init_"$DUMB_INIT_VERSION"_amd64.deb /
+ADD https://github.com/Yelp/dumb-init/releases/download/v"$DUMB_INIT_VERSION"/dumb-init_"$DUMB_INIT_VERSION"_${TARGETARCH}.deb /
 
 COPY docker/puppetserver/docker-entrypoint.sh \
      docker/puppetserver/healthcheck.sh \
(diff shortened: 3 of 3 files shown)


docker/Makefile
```diff
@@ -77,9 +77,9 @@ test: prep
 		rspec --options puppetserver/.rspec spec
 
 push-image: prep
-	@docker push puppet/puppetserver:$(VERSION)
```
justinstoller, line 80: similar to my r10k comment, what is the rationale for doing this in the `push-image` target vs the `build` target?

justinstoller:
Looks great! I have a question about the rationale for why the logic lives in image push vs build (and whatever the rationale is maybe it would be valuable in a comment and the commit message for us in the future?)

Otherwise, we're happy to merge!

docker/Makefile
odidev: [@justinstoller](https://github.com/justinstoller)
Thanks for your response,
Here, we are using **buildx build** for building multiarchitecture images, where we need to use `--push` flag with buildx in makefile. 
If we want to build for single platform at a time, we can use `--load` flag to build and then push later.

.github/workflows/docker.yml
```diff
@@ -22,6 +22,12 @@ jobs:
         with:
           ruby-version: 2.6.x
       - run: gem install bundler
+      - uses: actions/checkout@v2
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      - name: Set up Docker Buildx
```
Iristyle, line 28: This is unnecessary - we already install / use buildx

docker/Makefile
Iristyle: The Makefile already calls `docker buildx build` - new targets should be added there ... this make target is specifically for pushing.

Iristyle (changes requested):

docker/Makefile
odidev: @Iristyle
```The Makefile already calls docker buildx build - new targets should be added there ... this make target is specifically for pushing.```

We need to use `–push` flag here instead of `--load`, as it needs to push multi-architecture images at a time, and we can use `--load` when we need to build an image for a specific platform, then we can push image separately for that specific architecture.

File: .github/workflows/docker.yml (hunk 2 above)
Comment (odidev) [reply .github/workflows/docker.yml:28]: @Iristyle 
```This is unnecessary - we already install / use buildx```

We need to setup buildx with qemu to build and push for multiple architecture. If we do not set up buildx before building with `--push` flag, then it fails with below error: 

image

File: docker/puppetserver/Dockerfile [hunk 3]
```diff
@@ -100,16 +101,16 @@ ENV RUBYOPT=-r/puppetserver/docker/ruby-docker-copy-patch
 
 RUN lein clean && \
     lein install && \
-    EZBAKE_ALLOW_UNREPRODUCIBLE_BUILDS=true EZBAKE_NODEPLOY=true COW=base-bionic-amd64.cow MOCK='' GEM_SOURCE=https://rubygems.org lein with-profile ezbake ezbake local-build && \
+    EZBAKE_ALLOW_UNREPRODUCIBLE_BUILDS=true EZBAKE_NODEPLOY=true COW=base-bionic-${TARGETARCH}.cow MOCK='' GEM_SOURCE=https://rubygems.org lein with-profile ezbake ezbake local-build && \
```
Iristyle, line 104: Have you verified that this works? I don't think EZBake supports ARM

https://github.com/puppetlabs/packaging/blob/374e704ae9683712fdbca71f50699b33c8d494fa/lib/packaging/platforms.rb#L196

docker/puppetserver/Dockerfile
```diff
@@ -100,16 +101,16 @@ ENV RUBYOPT=-r/puppetserver/docker/ruby-docker-copy-patch
...
     mv /puppetserver/output/deb/bionic/*/*.deb /puppetserver.deb
 
 FROM base as edge
 
 COPY --from=build /puppetserver.deb /puppetserver.deb
 
 ARG UBUNTU_CODENAME
-ARG install_path=/puppetserver.deb
```
Iristyle, line 111: These should not be changed

docker/puppetserver/Dockerfile
```diff
@@ -8,7 +8,8 @@ ARG UBUNTU_CODENAME=bionic
 FROM ubuntu:18.04 as base
 
 ARG PACKAGES=ca-certificates\ git
-ARG DUMB_INIT_VERSION="1.2.2"
+ARG DUMB_INIT_VERSION="1.2.5"
+ARG TARGETARCH
```
Iristyle, line 12: A comment here that this is set by `buildx` would be useful... it looks like magic because it's not passed in during the build

docker/Makefile
Iristyle: My point is that we still want completely separate `build` and `push` targets in the Makefile. Since the `build` target wasn't modified to produce multiple architecture outputs, then this `push-image` target is also building, which is against the design of the Makefile

Iristyle:
I don't think this is going to work because I don't think EZBake is capable of producing ARM binaries... someone should see if that's even possible before this goes any further.

docker/puppetserver/Dockerfile
Iristyle: I think the only arm support we have is for M1 macs - https://github.com/puppetlabs/packaging/blob/374e704ae9683712fdbca71f50699b33c8d494fa/lib/packaging/platforms.rb#L112

Iristyle:
I'm going to close this as our internal EZBake tooling doesn't support producing ARM artifacts on Linux platforms.

Thanks for the PR @odidev, but we don't have the resources to make internal tooling changes for ARM Java builds at this point.

/cc @e-gris

docker/puppetserver/Dockerfile
hittingray: > Have you verified that this works? I don't think EZBake supports ARM
> 
> https://github.com/puppetlabs/packaging/blob/374e704ae9683712fdbca71f50699b33c8d494fa/lib/packaging/platforms.rb#L196

@Iristyle On that line you linked, I can see that `aarch64` is listed (which is just another name for `arm64`). Is there something else that would prevent EZBake from working?

PR #2619, opened 2022-03-08
Error in the documentation for puppetserver

gdelhumeau:
The 'Persistance' paragraph was wrong.

The documentation said the CA certificates are stored in `/etc/puppetlabs/puppetserver/ca`, meanwhile it seems to be `/etc/puppetlabs/puppet/ca`, as I have discovered later.

I hope it helps,

Guillaume

### DIFF
--- a/docker/puppetserver/README.md
+++ b/docker/puppetserver/README.md
@@ -52,7 +52,7 @@ These scripts will be executed at the end of the entrypoint script, before the s
 
 ## Persistance 
 
-If you plan to use the in-server CA, restarting the container can cause the server's keys and certificates to change, causing agents and the server to stop trusting each other. To prevent this, you can persist the default cadir, `/etc/puppetlabs/puppetserver/ca`. For example, `docker run -v $PWD/ca-ssl:/etc/puppetlabs/puppetserver/ca puppetlabs/puppetserver:latest`.
+If you plan to use the in-server CA, restarting the container can cause the server's keys and certificates to change, causing agents and the server to stop trusting each other. To prevent this, you can persist the default cadir, `/etc/puppetlabs/puppet/ca`. For example, `docker run -v $PWD/ca-ssl:/etc/puppetlabs/puppet/ca puppetlabs/puppetserver:latest`.
 
 ## Analytics Data Collection
 


Enygma2002:
The `SSLDIR=/etc/puppetlabs/puppet/ssl` defined in https://github.com/puppetlabs/puppetserver/blob/a7008e7/docker/puppetserver/Dockerfile#L26 clearly does not match the one mentioned in the documentation.

Enygma2002:
@justinstoller This PR just fixes the documentation of the currently released docker image for puppet 7.x (the one we tried) that made it work for us.

The current docker image defines a default for the ENV variable `SSLDIR` to `/etc/puppetlabs/puppet/ssl`. That var is then used in the docker container entry point to initialize certificates if they don't exist: https://github.com/puppetlabs/puppetserver/blob/a7008e7c8c662ed5c15a596140c12fd5a80b74d0/docker/puppetserver/docker-entrypoint.d/80-ca.sh#L46-L70

When trying to use the current docker image, if we go by the documentation and map a volume with existing certificates (from a previous installation) to `/etc/puppetlabs/puppetserver/ca`, puppet would still try to initialize fresh CA certificates but fail and complain that some already eixst. When we have tried to mount `/etc/puppetlabs/puppetserver/ca` with an empty content, puppet initialized fresh certificates in it successfully, but after a restart (when the folder now had certificates in it), puppet would again refuse to initialize, complaining the certificates already existed.

So we were unable to use the docker image unless we mounted `/etc/puppetlabs/puppet/ssl` with our existing certificates.

I'm not really up to date with Puppet's developments or plans for changing locations, we're just currently documenting what we have experienced with the current image. If the plan is to change the default location (i.e. default value of `SSLDIR`), then sure, but until then, IMO, the documentation of the image should be synchronized with its behavior.

Perhaps, another approach would be to consider that the entry point script that causes the CA generation (https://github.com/puppetlabs/puppetserver/blob/a7008e7c8c662ed5c15a596140c12fd5a80b74d0/docker/puppetserver/docker-entrypoint.d/80-ca.sh#L46-L70) should maybe use a secondary (e.g.) `SSLDIRNEW` ENV variable that would check the new location (introduced in puppet 6.x and 7.x, as you say) so that both old and new SSL locations are supported and the "certificates already exist" error is avoided at a restart time by not trying to generate new CAs when certificates are already present inside `/etc/puppetlabs/puppetserver/ca`.

jonathannewman:
The docker configuration is no longer relevant.  Closing.

PR #2620, opened 2022-03-08
(SERVER-1381) Create puppet user with uid/gid 52

alexjfisher:
Support for fixed UIDs/GIDs was added to ezbake way back in
EZ-128 (puppetlabs/ezbake#512)

This commit uses this feature to use the reserved UID of 52 when the
`puppet` user is created when the puppetserver RPM is first installed.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -90,6 +90,7 @@
   :uberjar-name "puppet-server-release.jar"
   :lein-ezbake {:vars {:user "puppet"
                        :group "puppet"
+                       :numeric-uid-gid 52
                        :build-type "foss"
                        :java-args ~(str "-Xms2g -Xmx2g "
                                      "-Djruby.logger.class=com.puppetlabs.jruby_utils.jruby.Slf4jLogger")


alexjfisher:
Note: At the moment, I've not tested this.  I don't have access to the internal puppet Jenkins systems needed to drive the whole ezbake process, (and am doubtful I could replicate it locally).  It should be a trivial change though and https://github.com/puppetlabs/ezbake/pull/512 was reviewed and tested at the time.

mwaggett:
I was able to build packages with this change and confirm that uid/gid 52 is used:
```
[root@huge-crackling ~]# ls -l /etc/puppetlabs/
total 0
drwxr-xr-x 4 root   root   39 Apr 14 00:47 code
drwxr-xr-x 3 root   root   67 Apr 14 00:48 puppet
drwxr-x--- 4 puppet puppet 80 Apr 14 00:48 puppetserver
drwxr-xr-x 3 root   root   20 Apr 14 00:47 pxp-agent
[root@huge-crackling ~]# ls -ln /etc/puppetlabs/
total 0
drwxr-xr-x 4  0  0 39 Apr 14 00:47 code
drwxr-xr-x 3  0  0 67 Apr 14 00:48 puppet
drwxr-x--- 4 52 52 80 Apr 14 00:48 puppetserver
drwxr-xr-x 3  0  0 20 Apr 14 00:47 pxp-agent
```

mwaggett (approved):

mwaggett:
Thanks for this, @alexjfisher !

PR #2621, opened 2022-03-09
(SERVER-2662) Use existing cacert's subject key id when signing

tvpartytonight:
With the release of jvm-ssl-utils 3.4.0, the authority key identifer
can be copied from an existing certificate and used when signing
CSRs, instead of always computing a type 1 key from the CA's public
key. This commit changes puppetserver to use the CA's certificate
during signing so that puppetserver signs CSRs with the existing CA's
subject key id as the authority key id for agent requests. This will
allow puppetserver to use externally created CA certificates that
use a type 2 key.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -855,14 +855,13 @@
   "Create a list of extensions to be added to the master certificate."
   [master-certname :- schema/Str
    master-public-key :- (schema/pred utils/public-key?)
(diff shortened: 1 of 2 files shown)


justinstoller:
We probably want to add this to the `create-master-extensions` too?

tvpartytonight:
@justinstoller I was thinking that I wanted to leave all other workflows untouched, but I think you are right that there is a scenario where we have only the imported CA key and cert, and so we want to master extensions to have this set to using the cert as well. I've amended the PR to include that.

justinstoller (approved):

PR #2624, opened 2022-03-22
(SERVER-3156) Enable sles15 builds

Magisus:


mwaggett:
Since redhatfips-8 was added in ezbake 2.3.1, I would also include that in this commit message.

Magisus:
So I'm actually confused about this. We apparently already have Focal enabled in CI: https://github.com/puppetlabs/ci-job-configs/blob/e57c99cbfa1c634c6698618db3dd257541084254/jenkii/platform/projects/puppetserver.yaml#L65 despite the fact it looks like it wasn't in ezbake until this version: https://github.com/puppetlabs/ezbake/blob/main/CHANGELOG.md#232---2022-02-15. Is the thing we update in ezbake just a default that can be overridden completely independently in CI?

Magisus:
@mwaggett I deliberately left it out since we do not build it for FOSS.

mwaggett:
oooh right, good point 😅

mwaggett (approved):

Magisus:
I also dropped the reference to Ubuntu 20.04, since we actually did that one a while ago in FOSS. The update in ezbake for this bump was PE-only.

Magisus:
Is that a new jackson thing or have we just not resolved that yet?

mwaggett:
Looks like the same one, but it should be fixed already..

PR #2626, opened 2022-03-30
(SERVER-3170) make dropsonde opt-out by default

jonathannewman:
This changes dropsonde to be opt-out by default if the configuration value is
not specified per this announcement:

https://puppet.com/blog/the-next-step-in-puppet-content-telemetry/


Magisus (approved):
I'll test this real quick, but I think it's good.

mwaggett (approved):

PR #2629, opened 2022-04-14
(SERVER-3133) Remove deprecated JRuby function

Magisus:
JRuby deprecated `java_kind_of?` many years ago and finally removed it
in 9.3. This commit switches to the general replacement `kind_of?` in
preparation for upgrade to JRuby 9.3.


mwaggett:
Should we make this change in 6.x, even though we're not planning to bump JRuby there? Or do you think there's enough risk that it's not worth it?

Magisus:
I don't think there's risk, I guess I just wanted to make it really clear that we're not planning to upgrade JRuby there.

Magisus:
Switch to 6.x to make backporting JRuby 9.3 easier if we ever need to do that.

mwaggett (approved):

PR #2630, opened 2022-04-18
(maint) Update acceptance test for git CVE-2022-24765

justinstoller:
Following CVE-2022-24765 (see
https://github.blog/2022-04-12-git-security-vulnerability-announced/)
git will no longer trust repositories owned by another user without
being explicitly allowed. This fix is in Git 2.35.2.

The code_scripts.rb test has the git user own a test repositiory so
it may be served by a local git server. Once the repository is owned
by the git user the root user may not commit to it.

This commit marks the shared local repository as safe as soon as its
ownership is given to the git user.

### DIFF
--- a/acceptance/suites/tests/code_commands/code_scripts.rb
+++ b/acceptance/suites/tests/code_commands/code_scripts.rb
@@ -108,6 +108,7 @@ end
 
 step 'SETUP: Initialize the local git repository' do
   on master, "chown git #{git_local_repo}"
+  on master, "git config --global --add safe.directory #{git_local_repo}"
   on master, "cd #{git_local_repo} && git config --global user.name 'TestUser'"
   on master, "cd #{git_local_repo} && git config --global user.email 'you@example.com'"
   on master, "cd #{git_local_repo} && git init"


justinstoller:
So... I've opened this against the main branch because this is failing in our newer Ubuntu distros and they are only in our main branch. However, maybe I should target this at the 6.x branch since this is distro specific and at some point we may update our existing images?

justinstoller:
retargeted to 6.x

mwaggett (approved):
Looks good to me! assuming you tested that this fixes the issue?

justinstoller:
I tested it on `ubuntu2004-64ma-64a` running just the test in question before and after the change and it resolved the issue. Though that was before I rebased from main -> 6.x. I can retest in a few if you'd like.

mwaggett:
meh, seems fine if you tested on `main`.

`AppVeyor was unable to build non-mergeable pull request` - is this because Travis was disabled? Is there a path forward with that?

justinstoller:
I think that comes from me retargeting the PR, I'll close and reopen to try and rekick.

PR #2632, opened 2022-04-22
(PE-33240) Properly serve task files from a module called "tasks"

nicklewis:
Previously, the URLs generated for certain task files were incorrect and
led to those tasks not being runnable. Task files can be loaded from the
`tasks`, `scripts`, `files` or `lib` directory of a module. If the
module itself had any of those names, the URL generated was incorrect.

The trouble was that we were trying to determine the module and
subdirectory of the file purely based on its absolute path on disk,
without knowing the modulepath or the module name. Without that
information, it's impossible to accurately interpret the file path in
all cases.

We now use the additional "file name" field which has the information we
need in a much easier to parse form. Specifically, the file name is
supplied in one of two forms:

* the basename for a task implementation file, in which case we know
  that it must be at the top-level of the `tasks/` directory of its
  module (because tasks can't be nested), and we can use that
  information to find the module name from the absolute path
* <module>/<module subdir>/<path to the file>, which are simply the
  exact three fields we need

Given the module name, the file "type" (tasks/, scripts/, etc) and the
path within that subdirectory, we can build the appropriate file serving
URL.

The one wrinkle is for static file serving, where we instead use the
verbatim path on disk relative to the root of the environment. Since we
don't have the root of the environment available, we use the fields we
previously computed to deduce it from the absolute path.

### DIFF
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -332,16 +332,30 @@
     (->> (map format-task info-from-jruby)
          (middleware-utils/json-response 200))))
 
+(defn task-file-uri-components
+  "The 'id' portion for a task implementation is a single component like
+  'foo.sh' and can never be nested in subdirectories. In that case we know the
+  'file' must be structured <environment>/<module root>/<module>/tasks/<filename>.
+  Other task files are the path relative to the module root, in the form
+  <module>/<mount>/<path> where <path> may have subdirectories. For those, we
+  just slice that off the end of the file path and take the last component that
+  remains as the module root."
+  [file-id file]
+  (if (str/includes? file-id "/")
+    (let [[module mount subpath] (str/split file-id #"/" 3)
+          module-root (fs/base-name (subs file 0 (str/last-index-of file file-id)))]
+      [module-root module mount subpath])
+    (take-last 4 (str/split file #"/"))))
+
 (defn describe-task-file
   [get-code-content env-name code-id file-data]
   (let [file (:path file-data)
-        subpath (:name file-data)
-        basename (fs/base-name subpath)
+        file-id (:name file-data)
         size (fs/size file)
         sha256 (ks/file->sha256 (io/file file))
-        ;; we trust the file path from Puppet, so extract the subpath from file
-        static-path (re-find (re-pattern (str #"[^/]+/[a-z][a-z0-9_]*/(?:tasks|files|scripts|lib)/.*" basename)) file)
-        [_ module-name mount rest] (str/split static-path #"/" 4)
+        ;; we trust the file path from Puppet, so extract the relative path info from the file
+        [module-root module-name mount relative-path] (task-file-uri-components file-id file)
+        static-path (str/join "/" [module-root module-name mount relative-path])
         uri (try
               ;; if code content can be retrieved, then use static_file_content
               (when (not= sha256 (ks/stream->sha256 (get-code-content env-name code-id static-path)))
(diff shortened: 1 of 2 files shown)


src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -332,16 +332,30 @@
...
+  'foo.sh' and can never be nested in subdirectories. In that case we know the
+  'file' must be structured <environment>/<module root>/<module>/tasks/<filename>.
+  Other task files are the path relative to the module root, in the form
+  <module>/<mount>/<path> where <path> may have subdirectories. For those, we
+  just slice that off the end of the file path and take the last component that
+  remains as the module root."
+  [file-name file]
+  (if (str/includes? file-name "/")
+    (let [[module mount subpath] (str/split file-name #"/" 3)
+          module-root (fs/base-name (subs file 0 (- (count file) (count file-name))))]
```
donoghuc, line 346: Curious why you would do "math" and string indexing instead of a regex (maybe something like `(clojure.string/replace file (re-pattern (str file-name "$")) "")` ?

donoghuc (approved):
Took this for a spin in a PE install and It is solid. Nice work, this was a hairy one.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -332,16 +332,30 @@
...
 
+(defn task-file-uri-components
+  "The 'name' portion for a task implementation is a single component like
+  'foo.sh' and can never be nested in subdirectories. In that case we know the
+  'file' must be structured <environment>/<module root>/<module>/tasks/<filename>.
+  Other task files are the path relative to the module root, in the form
+  <module>/<mount>/<path> where <path> may have subdirectories. For those, we
+  just slice that off the end of the file path and take the last component that
+  remains as the module root."
+  [file-name file]
```
mwaggett, line 343: (nit) I think these parameter names could be a little clearer. When I see `file-name`, I imagine a single component, like you describe in the docstring. I think `subpath`, as is used in `describe-task-file` below, would be clearer. I would also probably call `file` something like `full-path`, but I think I'd rather you maintain the consistency with `describe-task-file` than just change this one.

src/clj/puppetlabs/services/master/master_core.clj
nicklewis: I think either way is probably fine. Since the original bug was caused by an imprecise use of regex, I was feeling somewhat disinclined to use that approach. It's actually somewhat tricky to formulate a regex that does _precisely_ 
what is intended here. For instance, your suggested code has a bug in that it doesn't _escape_ the string used to create the pattern, meaning special characters will retain their special meaning. _Unlikely_ to cause an issue, but certainly not impossible. Taking a substring sidesteps any of that and does _exactly_ what is meant: remove this 
substring from the end of the string.

A better way might be `(fs/base-name (subs file 0 (str/last-index-of file file-name)))` to avoid the finnicky math

mwaggett: I think using `last-index-of` makes the intention clearer. 
I agree that avoiding a regex is ideal.

src/clj/puppetlabs/services/master/master_core.clj
nicklewis: I updated this to go with `file-id` in both places, as that reflects what that field actually means. It's supposed to specify _which_ file should be used, separate from _where_ the file is located. Bolt calls it `name` but I agree that `file-name` is ambiguous, so `id` seems like a decent compromise.

I _don't_ like `subpath` because it's not actually a "sub" of the absolute path. It has its own semantics.

src/clj/puppetlabs/services/master/master_core.clj
nicklewis: Updated to use `last-index-of`. I think I couldn't wrap my brain around the logic on friday :D

mwaggett (approved):

nicklewis:
The one test failure doesn't look related to my change.

src/clj/puppetlabs/services/master/master_core.clj
donoghuc: TIL `last-index-of`. Clojure is so cool.

donoghuc (approved):

mwaggett:
Re-ran the tests, everything looks good!

PR #2633, opened 2022-05-05
(maint) Add custom merge driver script

mwaggett:
This commit adds a script that can be used as a merge driver for project.clj
and for the beaker options file. It will resolve common conflicts around version
bumps, but any unfamiliar conflicts will need manual intervention. This script
was copied from pe-puppet-server-extensions and modified slightly to handle the
beaker options file. Previously, puppetserver mergeup conflicts were handled by
a custom merge driver created as part of the `puppetserver-mergeup.sh` script in
ci-job-configs, but this did not handle conflicts in project.clj. This strategy
will allow us to resolve fewer conflicts by hand and will allow us to simplify
the mergeup script in jenkins.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,7 @@
+project.clj merge=version-mergeup
+acceptance/config/beaker/options.rb merge=version-mergeup
+documentation/release_notes.markdown merge=ours
+resources/ext/build-scripts/mri-gem-list-no-deps.txt merge=ours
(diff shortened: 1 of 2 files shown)


mwaggett:
This will require updates to ci-job-configs to use, but there's no harm in merging this before those changes go in.

Magisus (approved):

PR #2635, opened 2022-05-20
(PE-33991) remove use of slurp to reduce memory footprint

jonathannewman:
This removes the use of slurp in the clojure code to instead
present the input stream to ruby to process. While ruby will
still read the string when the body is requested, this change
will prevent the clj layer from having a referenced string during
the lifetime of the request. This should help reduce memory overhead.

### DIFF
--- a/src/ruby/puppetserver-lib/puppet/server/network/http/handler.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/network/http/handler.rb
@@ -36,7 +36,25 @@ module Puppet::Server::Network::HTTP::Handler
   def body(request)
     body = request["body"]
     if body.kind_of?(InputStream)
-      body.to_io.read()
+      io = body.to_io
+      content_type = ''
+      has_headers = request.key?('headers')
+      if has_headers && request['headers'].key?('content-type')
+         content_type = request['headers']['content-type']
+      end
+
+      case content_type
+      when nil, "", "application/octet-stream", "application/x-msgpack"
+        io.binmode
+      else
+        encoding = 'UTF-8'
+        if has_headers
+            encoding = request['headers'].fetch('content-encoding', 'UTF-8')
+        end
+        io.set_encoding(encoding)
+      end
+
+      io.read
     else
       body
     end
--- a/src/clj/puppetlabs/services/request_handler/request_handler_core.clj
+++ b/src/clj/puppetlabs/services/request_handler/request_handler_core.clj
@@ -45,48 +45,32 @@
   "Converts a JRubyPuppetResponse instance to a map."
   [response]
   { :pre [(instance? JRubyPuppetResponse response)]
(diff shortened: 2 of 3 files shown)


src/ruby/puppetserver-lib/puppet/server/network/http/handler.rb
```diff
@@ -36,7 +36,17 @@ def path(request)
   def body(request)
     body = request["body"]
     if body.kind_of?(InputStream)
-      body.to_io.read()
```
justinstoller, line 39: We may want to move this to where we turn the Java response into a Ruby response object https://github.com/puppetlabs/puppetserver/blob/6.x/src/ruby/puppetserver-lib/puppet/server/http_client.rb#L169 as I'm not 100% sure if the Http::Handler code wraps all Http responses in Puppet.

jonathannewman: I *think* they are handled by this?  I thought the bottleneck was here:  https://github.com/puppetlabs/puppetserver/blob/750958954896578ad877d43b8bb61eec66620c70/src/ruby/puppetserver-lib/puppet/server/master.rb#L62-L68  and that includes the above addition.  Is there an alternative path?

jonathannewman:
I need to special case `application/x-www-form-urlencoded` as that is read to extract the params, which leaves the stream closed.

src/ruby/puppetserver-lib/puppet/server/network/http/handler.rb
jonathannewman: @justinstoller I looked at your comment again, and I'm a little confused.  I think what you linked deals with the response, and this is dealing with the request?  I certainly could be wrong about how this flow works.

justinstoller: Oh, you're right. I was just in the HttpClient and was thinking of responses to Ruby requests like [classification](https://github.com/puppetlabs/classifier/blob/main/puppet/lib/puppet/indirector/node/classifier.rb#L125), but those strings are created in the Java http-client library not through clojure processing in server itself.

jonathannewman:
My only remaining question is, does this have any value?  Given the form-encoding still has to be slurped, are there meaningful paths where avoiding that slurp makes sense?

Thoughts @Sharpie ?

jonathannewman:
I think this has merit for handling report processing to reduce some of the memory footprint there.

mwaggett:
@jonathannewman Do we still want to land this? If so, I assume we'd land it in 7.x.

jonathannewman:
I think so.  Given the risk it should likely only land in main.  I'll move it there.

jonathannewman:
Closing in favor of https://github.com/puppetlabs/puppetserver/pull/2661

PR #2637, opened 2022-06-02
Update 90-log-config.sh

ldaneliukas:
Bringing up a new Puppetserver 7.8.0 container that has a CA present, but not a signed cert will see the container exit due to `openssl` errors when attempting to read a non-existent file. This happens as the cert is not yet signed when bringing up the server, but will be signed by the CA once it is up and running.

Log certificate information only if cert is actually present.


jonathannewman:
The docker configuration is no longer relevant for this project.  Closing.

PR #2641, opened 2022-06-27
(PUP-11540) Update all-tasks-response! to read private/description

mcdonaldseanp:
Updates the master_core service to return "private" and "description"
keys from tasks when hitting the /v3/tasks API. Depends on
https://github.com/puppetlabs/puppet/pull/8921

### DIFF
--- a/src/clj/puppetlabs/services/master/master_core.clj
+++ b/src/clj/puppetlabs/services/master/master_core.clj
@@ -327,6 +327,8 @@
    jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)]
   (let [format-task (fn [task-object]
                       {:name (:name task-object)
+                       :private (get-in task-object [:metadata :private] false)
+                       :description (get-in task-object [:metadata :description] "")
                        :environment [{:name environment
                                       :code_id nil}]})]
     (->> (map format-task info-from-jruby)
--- a/test/unit/puppetlabs/services/master/master_core_test.clj
+++ b/test/unit/puppetlabs/services/master/master_core_test.clj
@@ -28,7 +28,9 @@
   [task-name]
   (let [[module _] (split task-name #"::")]
     {:module {:name module}
-     :name task-name}))
+     :name task-name
+     :private false
+     :description "test description"}))
 
 (defn build-ring-handler
   [request-handler puppet-version jruby-service]
@@ -378,7 +380,9 @@
             response-format (fn [task-name]
                                  {:name task-name
                                   :environment [{:name "production"
-                                                 :code_id nil}]})
+                                                 :code_id nil}]
+                                  :private false
+                                  :description "test description"})
             expected-response (fn [task-names]
                                 (map response-format task-names))
             task-names ["apache" "apache::configure" "mongodb::uninstall"]]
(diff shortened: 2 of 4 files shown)


mcdonaldseanp:
These specs are going to fail until the puppet PR is promoted to puppetserver

donoghuc:
Looks good. I hate the default output for these tests:
```
FAIL in (all-tasks-response-test) (master_core_test.clj:406)
all-tasks query formats response body properly
expected: (= (set (expected-response task-names)) (-> (map construct-info-from-task-name task-names) response :body (json/decode true) set))
  actual: (not (= #{{:name "apache::configure", :environment [{:name "production", :code_id nil}], :private false, :description "test description"} {:name "mongodb::uninstall", :environment [{:name "production", :code_id nil}], :private false, :description "test description"} {:name "apache", :environment [{:name "production", :code_id nil}], :private false, :description "test description"}} #{{:name "apache::configure", :private false, :description "", :environment [{:name "production", :code_id nil}]} {:name "mongodb::uninstall", :private false, :description "", :environment [{:name "production", :code_id nil}]} {:name "apache", :private false, :description "", :environment [{:name "production", :code_id nil}]}}))
```
I wonder if people have tried/suggested integrating humane test output in this repo... Might PR that real quick and see if it takes.

donoghuc (approved):
Looks good when puppet promotes.

src/clj/puppetlabs/services/master/master_core.clj
```diff
@@ -327,6 +327,8 @@
    jruby-service :- (schema/protocol jruby-protocol/JRubyPuppetService)]
   (let [format-task (fn [task-object]
                       {:name (:name task-object)
+                       :private (get (:metadata task-object) :private false)
+                       :description (get (:metadata task-object) :description "")
```
mwaggett, line 331: Would probably be cleaner to use `get-in` here, e.g. `(get-in task-object [:metadata :private] false)`

mcdonaldseanp: Ah, yeah that makes more sense.

mcdonaldseanp:
@mwaggett Yeah, we should try to get it out by the platform release.

mwaggett (approved):

test/unit/puppetlabs/services/master/master_core_test.clj
```diff
@@ -28,7 +28,9 @@
   [task-name]
   (let [[module _] (split task-name #"::")]
     {:module {:name module}
-     :name task-name}))
+     :name task-name
+     :private false
```
donoghuc, line 32: I think these mocks are wrong. I believe `private` and `description` should be nested under `metadata`.  https://github.com/puppetlabs/puppet/pull/8921/files

donoghuc (changes requested):
I think the test failures are legit. One looks like incorrect mocking, the others may just need expectations updated to reflect the new changes.

mwaggett (commented):
Looks like these need to be update as well: https://github.com/puppetlabs/puppetserver/blob/fb9ef3de1a1c87a487dd70181b075e463f51a480/test/integration/puppetlabs/services/jruby/tasks_test.clj#L112-L123

puppetserver#main is failing now, since the puppet changes were promoted, so hoping we can get this in ASAP.
I'll be out the whole week of July 4, but feel free to merge without me, or if you want, you can get a review from someone else on the team.

donoghuc:
Cleaned up the tests, this should be g2g.

donoghuc (approved):

steveax (approved):
👍

beechtom (approved):

PR #2642, opened 2022-06-27
(maint) Specify cipher suites when testing TLS 1.3

mwaggett:
This commit ensures that the test http client uses the appropriate cipher suites
when testing TLSv1.3. Previously, this test would fail with
`javax.net.ssl.SSLHandshakeException: Received fatal alert: protocol_version` on
newer Java 8 versions.

### DIFF
--- a/test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
+++ b/test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
@@ -287,7 +287,7 @@
   (testing "Can connect via TLSv1.3 by default"
     (with-webserver-with-protocols ["TLSv1.3"] nil
       (with-scripting-container sc
-        (with-http-client sc {}
+        (with-http-client sc {:cipher-suites ["TLS_AES_256_GCM_SHA384" "TLS_AES_128_GCM_SHA256"]}
           (let [url (str "https://localhost:10080")]
             (.runScriptlet sc (format "$response = $c.get(URI('%s'))" url))
             (is (= 200 (.runScriptlet sc "$response.code")))


test/unit/puppetlabs/puppetserver/ruby/http_client_test.clj
```diff
@@ -287,7 +287,7 @@
   (testing "Can connect via TLSv1.3 by default"
     (with-webserver-with-protocols ["TLSv1.3"] nil
       (with-scripting-container sc
-        (with-http-client sc {}
+        (with-http-client sc {:cipher-suites ["TLS_AES_256_GCM_SHA384" "TLS_AES_128_GCM_SHA256"]}
```
steveax, line 290: Looks like `TLS_CHACHA20_POLY1305_SHA256` is [3rd in order of preference](https://confluence.puppetlabs.com/pages/viewpage.action?spaceKey=SRE&title=TLS+Usage+Standard#TLSUsageStandard-CipherSuitePreferences) should we add that as well?

mwaggett: According to https://github.com/puppetlabs/puppet-enterprise-modules/commit/8f53836b4dde706284902ca3fbbcc3126ff6cd83, `TLS_CHACHA20_POLY1305_SHA256` is not supported on FIPS, so is not included in the [default TLS 1.3 list](https://github.com/puppetlabs/puppet-enterprise-modules/blob/71dc19d91ac5b1377c4b21aeae1ac817293fbbcd/modules/puppet_enterprise/functions/ciphers.pp#L50-L53). Since the other two ciphers are preferred anyway and it works without the CHACHA cipher, I don't think we'd gain anything from conditionally adding for non-FIPS tests.

steveax (approved):
👍

PR #2643, opened 2022-06-28
(maint) Humane test output for clojure tests

donoghuc:
Use the pjstadig/humane-test-output library to pretty print test results as well as diffs between actual vs expected in assertion failures. This will help debugging failing tests from just the Travis logs instead of having to get a repl going locally just to understand what the difference between actual and expected is.


mwaggett:
If you have it handy, would you mind pasting an example of what the humane output looks like compared to without it?

mwaggett (approved):

PR #2644, opened 2022-06-30
(maint) Remove trailing whitespace

mwaggett:
This causes automatic mergeups to fail, since it thinks that conflict markers
were not cleaned up appropriately.


steveax (approved):
👍 we should think about adding an [editorconfig file in the repo](https://editorconfig.org/)

mwaggett:
@steveax Feel free to open a ticket if you think that's something we should do! or add a retro card if you think we should discuss as a team first.

PR #2645, opened 2022-07-02
(maint) Skip external cert chain test on FIPS

mwaggett:

This should hopefully unblock pe-puppetserver CI.

### DIFF
--- a/acceptance/suites/tests/https/client_may_use_external_cert_chains.rb
+++ b/acceptance/suites/tests/https/client_may_use_external_cert_chains.rb
@@ -73,30 +73,24 @@ EOF
   report_content = on(master, "cat #{directory_to_serve}/#{master}").stdout.chomp
   assert_match /(un)?changed/, report_content
 
+  system_ssl_tmpdir = master.tmpdir('system_ssl_store')
+  script_location = system_ssl_tmpdir + '/connection_test.rb'
 
-  # TODO: Because of PA-4440 puppet-cacerts may not be available on all systems
-  if 0 == on(master, 'test -f /opt/puppetlabs/puppet/ssl/puppet-cacerts', accept_all_exit_codes: true).exit_code
-    system_ssl_tmpdir = master.tmpdir('system_ssl_store')
-    script_location = system_ssl_tmpdir + '/connection_test.rb'
+  test_script = <<EOF
+    require 'puppet'
+    require 'puppet/server/puppet_config'
+    require 'puppet/server/http_client'
+    require 'uri'
 
-    test_script = <<EOF
-      require 'puppet'
-      require 'puppet/server/puppet_config'
-      require 'puppet/server/http_client'
-      require 'uri'
+    Puppet::Server::PuppetConfig.initialize_puppet(puppet_config: {})
+    client = Puppet.runtime[:http]
 
-      Puppet::Server::PuppetConfig.initialize_puppet(puppet_config: {})
-      client = Puppet.runtime[:http]
+    response = client.get(URI('https://github.com/index.html'), options: {include_system_store: true})
 
-      response = client.get(URI('https://github.com/index.html'), options: {include_system_store: true})
-
-      puts response.code
+    puts response.code
 EOF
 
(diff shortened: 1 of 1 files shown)


mwaggett:
The test failures are unrelated to this. They should be fixed by changes to https://github.com/puppetlabs/puppetserver/pull/2641.

acceptance/suites/tests/https/client_may_use_external_cert_chains.rb
```diff
@@ -1,5 +1,8 @@
 test_name "Ensure Puppet Server's HTTP client may use external cert chains" do
 
+  ## This test currently relies on WEBrick/OpenSSL, which is not supported on FIPS
```
justinstoller, line 3: That's weird, we run the webrick server using the agent's Ruby which I thought has OpenSSL available to it. Was there anything in the `webrick.log`?

joshcooper: Ah sorry @mwaggett I think I gave you bad advice earlier. I didn't realize the first part of the test used webrick and the second used puppetserver's ruby. To @justinstoller point, the first part should work on FIPS hosts since we're using the agent's ruby to load net/http and openssl.

jonathannewman (approved):

acceptance/suites/tests/https/client_may_use_external_cert_chains.rb
mwaggett: ahh @justinstoller @joshcooper so you're saying this *should* work on FIPS? Should we revert this commit then? https://github.com/puppetlabs/puppetserver/pull/2645/commits/6e281516fdffb430ad91c9a4a886d8bff2d24231

justinstoller: I think @jonathannewman merged it to unblock CI and let folks investigate the FIPS failures at a lower priority.

mwaggett: @jonathannewman Is there a ticket to investigate the FIPS failures and revert https://github.com/puppetlabs/puppetserver/commit/6e281516fdffb430ad91c9a4a886d8bff2d24231 once we get it figured out?

jonathannewman: I have not written one.  Feel free to.  As I understood it, these tests weren't expected to pass in this environment.

mwaggett: That's what I had originally thought, but it sounds like @justinstoller and @joshcooper are saying it actually _should_ work on FIPS, since we use the agent's Ruby. Am I understanding that right?

justinstoller: There's issues in general with HTTPS connections in Server on FIPS, so I don't know how important it is that we can run this test on FIPS right now.

However, the stated reason for the failure in this comment shouldn't be an issue on FIPS because we're not loading Webrick in JRuby.

mwaggett: @justinstoller I'm trying to determine if there's a need to open a followup ticket to re-enable this test on FIPS, since it sounds like this _should_ work. I don't know about these "issues in general with HTTPS connections in Server on FIPS" - is that something we expect to resolve at some point? is there a specific ticket (or tickets) for fixing this? or should we never plan to run this test on FIPS..?

justinstoller: I think you should create a ticket. I was referring to the fact Net::HTTPS / OpenSSL library doesn't work in JRuby on FIPS.

mwaggett: okay, opened https://tickets.puppetlabs.com/browse/PE-34102 to resolve the failure and https://github.com/puppetlabs/puppetserver/pull/2647 to fix the comment.

PR #2652, opened 2022-08-30
(maint) fix packaging gem installation

steveax:
v5.x of the public_suffix gem no longer supports ruby 2.5 which
causes installation failures during the bundle install. This commit
pins public_suffix to v4.0.7, which is compatible. Note this is not
needed for main branch which is running jruby 9.3.x (compatible with
Ruby 2.6.x).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,6 +10,9 @@ def location_for(place, fake_version = nil)
   end
 end
 
+# The public_suffix gem used in the dependencies for the packaging gem no longer supports Ruby 2.5
+# and bundler is not smart enough to realize that an older version would work in the mix.
+gem 'public_suffix', '= 4.0.7'
 gem 'packaging', *location_for(ENV['PACKAGING_LOCATION'] || '~> 0.99')
 gem 'rake', :group => [:development, :test]
 


nmburgan (approved):

Gemfile
```diff
@@ -26,6 +26,10 @@ group :test do
 
   # docker-api 1.32.0 requires ruby 2.0.0
   gem 'docker-api', '1.31.0'
+
+  # The public_suffix gem used in the dependencies for the packaging gem no longer supports Ruby 2.5
+  # and bundler is not smart enough to realize that an older version would work in the mix.
+  gem 'public_suffix', '= 4.0.7'
```
nmburgan, line 32: This might need to go outside the `:test` block since I think it's `packaging` that pulls in `public_suffix`

nmburgan (changes requested):
This might need to go outside the `:test` group since I think it's `packaging` that pulls in public_suffix

steveax:
Updated

Gemfile
```diff
@@ -10,6 +10,10 @@ def location_for(place, fake_version = nil)
   end
 end
 
+# The public_suffix gem used in the dependencies for the packaging gem no longer supports Ruby 2.5
+# and bundler is not smart enough to realize that an older version would work in the mix.
+gem 'public_suffix', '= 4.0.7'
+
```
mwaggett, line 16: (nit) can this go under the 'packaging' entry to emphasize that this is a dependency of the packaging gem and not something that we need otherwise? I know the comment mentions it, but the grouping would help my brain. ¯\_(ツ)_/¯

mwaggett (approved):

mwaggett:
Looks like `fpm` also depends on `public_suffix`, which we install as part of the pupperware testing: https://github.com/puppetlabs/puppetserver/blob/6.x/docker/puppetserver/Dockerfile#L92 

I'm not sure where the Ruby version is coming from in that context, but we might just need to add `gem install public_suffix -v 4.0.7` to the Dockerfile, before we attempt to install `fpm`.

Gemfile
steveax: Doesn't the install need to be _before_ the packaging gem in order for the packaging install not to fail?

jonathannewman:
The dockerfile isn't something we are maintaining anymore, so it isn't a concern.

Gemfile
mwaggett: oh hmm maybe. I guess I haven't looked into how exactly Gemfiles are evaluated. In that case, can we just remove the extra newline that separates this from the 'packaging' entry?

steveax: Sure thing, done.

mwaggett (approved):

nmburgan (approved):

PR #2655, opened 2022-09-12
(PE-34508) Unpin bundler in Rakefile

justinstoller:
We previously pinned bundler to a version of 1.x because when 2.x was released it required a Ruby version compatible with 2.3 or greater but we were stuck support JRuby 1.7.

Those constraints are long gone and we need newer bundler versions to better handle dependency restrictions, particularly based on Ruby versions.


nmburgan (approved):

steveax (approved):
👍

PR #2656, opened 2022-09-13
(SERVER-3219) Exclude JRuby stdlib dependencies from uberjar

justinstoller:
We provide our own versions of these libraries and it's best to only have one of each on the classpath.


nmburgan (approved):
Assuming this tests out okay, LGTM

PR #2658, opened 2022-09-15
(PE-34096) Certificate authority regexp

ragingra:
Updating certificate authority validation regexp to restrict hostname format

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -806,15 +806,15 @@
       {:kind :invalid-subject-name
        :msg  (i18n/tru "Certificate names must be lower case.")}))
 
-  (when-not (re-matches #"\A[ -.0-~]+\Z" subject)
-    (sling/throw+
-      {:kind :invalid-subject-name
-       :msg  (i18n/tru "Subject contains unprintable or non-ASCII characters")}))
-
   (when (.contains subject "*")
     (sling/throw+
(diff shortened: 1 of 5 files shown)


jonathannewman:
It is important to have a good commit description so future archeologists understand the why and what of the changes that are made.  Can you update the commit description to describe why the change was made and what was changed?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -806,15 +806,15 @@
...
-  (when-not (re-matches #"\A[ -.0-~]+\Z" subject)
+  (when (.contains subject "*")
     (sling/throw+
       {:kind :invalid-subject-name
-       :msg  (i18n/tru "Subject contains unprintable or non-ASCII characters")}))
-
-  (when (.contains subject "*")
+       :msg  (i18n/tru "Subject contains a wildcard, which is not allowed: {0}" subject)}))
+  
+  (when-not (re-matches #"^([a-z0-9](?:(?:[a-z0-9-_]*|(?<!-)\.(?![-.]))*[a-z0-9]+)?)$" subject)
```
jonathannewman, line 814: I believe the dashes in the ranges that are supposed to be dashes need to be escaped:
`^([a-z0-9](?:(?:[a-z0-9\-_]*|(?<!-)\.(?![\-.]))*[a-z0-9]+)?)$`

Can you add some test cases to make sure those are accepted correctly?   I know we have a test around capital letters. I think more testing would be a great idea.

PR #2660, opened 2022-09-20
(PE-34096) Certificate authority regexp

ragingra:
Updating certificate authority validation regexp to restrict hostname format. Originally our regex validation was simply checking that the hostname did not contain non ascii characters. As this allowed invalid hostnames and the potential for command injection, we have now updated the regex to reduce the allowed characters as well as follow  a hostname or domain format.


jonathannewman (approved):
LGTM! Regexps for the win? : allowed?

PR #2661, opened 2022-09-20
(PE-33991) remove use of slurp to reduce memory footprint

jonathannewman:
This removes the use of slurp in the clojure code to instead present the input stream to ruby to process. While ruby will still read the string when the body is requested, this change will prevent the clj layer from having a referenced string during the lifetime of the request. This should help reduce memory overhead.

### DIFF
--- a/src/clj/puppetlabs/services/request_handler/request_handler_core.clj
+++ b/src/clj/puppetlabs/services/request_handler/request_handler_core.clj
@@ -45,48 +45,31 @@
   "Converts a JRubyPuppetResponse instance to a map."
   [response]
   { :pre [(instance? JRubyPuppetResponse response)]
-    :post [(map? %)] }
-    { :status  (.getStatus response)
-      :body    (.getBody response)
-      :headers {"Content-Type"     (.getContentType response)
-                "X-Puppet-Version" (.getPuppetVersion response)}})
+    :post [(map? %)]}
+  { :status  (.getStatus response)
+    :body    (.getBody response)
+    :headers {"Content-Type"     (.getContentType response)
+              "X-Puppet-Version" (.getPuppetVersion response)}})
 
-(defn body-for-jruby
-  "Converts the body from a request into a String if it is a non-binary
-   content type.  Otherwise, just returns back the same body InputStream.
-   Non-binary request bodies are coerced per the appropriate encoding at
-   the Clojure layer.  Binary request bodies, however, need to be preserved
-   in the originating InputStream so that they can be converted at the Ruby
-   layer, where the raw bytes within the stream can be converted losslessly
-   to a Ruby ASCII-8BIT encoded String.  Java has no equivalent to ASCII-8BIT
-   for its Strings."
+
+(defn update-body-for-jruby
+  "Converts the body from a request into a String if it is a form encoding.
+   Otherwise, just returns back the same body InputStream."
   [request]
-  (let [body         (:body request)
-        content-type (if-let [raw-type (:content-type request)]
+  (let [content-type (if-let [raw-type (:content-type request)]
                        (string/lower-case raw-type))]
-    (case content-type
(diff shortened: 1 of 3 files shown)


jonathannewman:
@mwaggett I opened a new one against main.

mwaggett:
I fixed the travis issue, so this should get rebased.

jonathannewman:
@mwaggett when you have some free time, this should be GTG.  I rebased and the tests are green now.

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
```diff
@@ -45,48 +45,32 @@
...
-      ; some ambiguities.  For "text/*" Content-Types, for example,
-      ; different RFCs specified that either US-ASCII or ISO-8859-1 could
-      ; be applied - see https://tools.ietf.org/html/rfc6657.  Ideally, this
-      ; should be filled in with a broader list of the different Content-Types
-      ; that Puppet recognizes and the default encodings to use when typical
-      ; Puppet requests do not specify a corresponding charset.
-      (slurp body :encoding (or (:character-encoding request)
-                                "UTF-8")))))
-
+    (if (=  content-type "application/x-www-form-urlencoded")
```
mwaggett, line 62: (nit)
```suggestion
    (if (= content-type "application/x-www-form-urlencoded")
```

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
```diff
@@ -45,48 +45,32 @@
...
+      (assoc request :body (slurp body :encoding (or (:character-encoding request)
+                                                     "UTF-8")))
```
mwaggett, line 64: I think it would be more idiomatic to do this with `update` instead of `assoc`
```suggestion
      (update request :body slurp :encoding (or (:character-encoding request) "UTF-8"))
```

src/ruby/puppetserver-lib/puppet/server/network/http/handler.rb
```diff
@@ -36,7 +36,25 @@ def path(request)
   def body(request)
     body = request["body"]
     if body.kind_of?(InputStream)
-      body.to_io.read()
+      io = body.to_io
+      content_type = ''
+      has_headers = request.key?('headers')
+      if has_headers && request['headers'].key?('content-type')
+         content_type = request['headers']['content-type']
```
mwaggett, line 43: (nit) this is indented too much
```suggestion
        content_type = request['headers']['content-type']
```

src/ruby/puppetserver-lib/puppet/server/network/http/handler.rb
```diff
@@ -36,7 +36,25 @@ def path(request)
...
+      end
+
+      case content_type
+      when nil, "", "application/octet-stream", "application/x-msgpack"
+        io.binmode
+      else
+        encoding = 'UTF-8'
+        if has_headers
+            encoding = request['headers'].fetch('content-encoding', 'UTF-8')
```
mwaggett, line 52: (nit) indentation
```suggestion
          encoding = request['headers'].fetch('content-encoding', 'UTF-8')
```

src/clj/puppetlabs/services/request_handler/request_handler_core.clj
jonathannewman: Good call.

jonathannewman:
Updated.  Note I specifically targeted main as this is new behavior and may have some risk associated with it.  Seemed inappropriate for an LTS.

mwaggett:
oh right, too many branches 🙃

mwaggett (approved):

mwaggett:
Snyk failure is the known jackson-databind CVE that doesn't have a fix released yet, so gonna merge.

PR #2662, opened 2022-09-21
(maint) Conditionally open Java IO only for Java 17

justinstoller:
In a previous commit we began passing flags to the JVM that allows JRuby to access protected members of the Java IO subsystem, as is currently required by JRuby 9.3.8. However, this flag does not exist in (at least) Java 8 and so we must conditionally include these flags based on Java versions.

The conditional code is based on the existing conditional JVM flags passed in FIPS mode.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -237,15 +237,17 @@
             "irb" ["trampoline" "run" "-m" "puppetlabs.puppetserver.cli.irb" "--config" "./dev/puppetserver.conf" "--"]
             "thread-test" ["trampoline" "run" "-b" "ext/thread_test/bootstrap.cfg" "--config" "./ext/thread_test/puppetserver.conf"]}
 
-  :jvm-opts ["-Djruby.logger.class=com.puppetlabs.jruby_utils.jruby.Slf4jLogger"
-               "-XX:+UseG1GC"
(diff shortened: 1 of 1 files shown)


project.clj
```diff
@@ -237,15 +237,17 @@
...
-               "java.base/java.io=ALL-UNNAMED"]
+  :jvm-opts ~(let [version (System/getProperty "java.specification.version")
+                   [major minor _] (clojure.string/split version #"\.")]
+               (concat
+                 ["-Djruby.logger.class=com.puppetlabs.jruby_utils.jruby.Slf4jLogger"
+                  "-XX:+UseG1GC"
+                  (str "-Xms" (heap-size "1G"))
+                  (str "-Xmx" (heap-size "2G"))
+                  "-XX:+IgnoreUnrecognizedVMOptions"]
+                 (if (= 17 (java.lang.Integer/parseInt major))
```
nmburgan, line 248: I think it's available in 11 as well?

jonathannewman (approved):

justinstoller:
Travis is currently attempting to install from xenial repositories despite running ubuntu jammy. There is no java 17 in xenial (16.04) which is causing the build to fail. I don't think those failures are related to the PR in anyway.

PR #2665, opened 2022-09-29
(PE-34653) update bouncycastle lib dependency to `18on` version

jonathannewman:
Recently bouncycastle changed the naming convention of their libs to use `18on` as they now only support jdk8 and above.  This required changes to the project.clj to correspond to the new versions.


shaigy (approved):

steveax (approved):
👍

PR #2668, opened 2022-10-03
Mergeup 7.x -> main

nmburgan:


mwaggett (approved):

nmburgan:
Seems like Snyk is looking at the dependency versions from the old clj-parent for whatever reason.

mwaggett:
Snyk is configured to run against the target branch, not the actual PR 🙃

PR #2669, opened 2022-10-03
(maint) Define `clj-parent-version`

mwaggett:
This commit adds a `clj-parent-version` definition to be used instead of declaring the clj-parent version directly in the dependency. This will facilitate automatic mergeups and matches what some other projects do and what is done for some other dependency versions.


mwaggett:
Automergeups have been a contentious topic lately, so closing this for now.

PR #2674, opened 2022-12-07
(PUP-11683) Remove invalid fixtures

justinstoller:
Previously we tested our task data integration using invalid fixtures and with PUP-11683 these fixtures will no longer be returned.

### DIFF
--- a/test/integration/puppetlabs/services/jruby/tasks_test.clj
+++ b/test/integration/puppetlabs/services/jruby/tasks_test.clj
@@ -181,17 +181,9 @@
 
 (deftest ^:integration all-tasks-test
   (testing "requesting all tasks"
-    (let [tasks [{:name "install"
+    (let [tasks [{:name "init"
                   :module-name "apache"
-                  :metadata (simple-impl-metadata "apache")
-                  :number-of-files 2}
-                 {:name "init"
-                  :module-name "apache"
-                  :number-of-files 1}
-                 {:name "configure"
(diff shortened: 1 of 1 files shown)


test/integration/puppetlabs/services/jruby/tasks_test.clj
```diff
@@ -181,17 +181,9 @@
 
 (deftest ^:integration all-tasks-test
   (testing "requesting all tasks"
-    (let [tasks [{:name "install"
-                  :module-name "apache"
-                  :metadata (simple-impl-metadata "apache")
-                  :number-of-files 2}
-                 {:name "init"
+    (let [tasks [{:name "init"
```
mwaggett, line 184: Would it still be useful to test with multiple tasks?

mwaggett (commented):
I think it would be helpful for the commit message to specify what exactly was invalid about these fixtures.

test/integration/puppetlabs/services/jruby/tasks_test.clj
justinstoller: I don't think it adds anything. (I was actually torn about removing the test) Really we're just making sure the JRuby API exists and returns valid data, the functionality of listing valid tasks is well tested in Puppet (what JRuby is calling into).

justinstoller:
Updated the commit message

mwaggett (approved):

PR #2678, opened 2023-01-16
(maint) Pin bundler to '< 2.4' for the 6.x branch

justinstoller:
Bunlder released a Christmas surprise in 2.4 which required Ruby 3.0.1 or greater. Later 2.4.x releases relaxed that to >= 2.6 but regardless the 2.3.x series of Bundler is the last to support Ruby < 2.6.

So for the 6.x series we need to pin Bundler to less than 2.4.


steveax (approved):
👍

PR #2679, opened 2023-01-16
(maint) Pin puppet_forge gem for dropsonde/ruby 2.5 support

justinstoller:


justinstoller:
re-running travis but it should be noted that the failure didn't involve this change (it was a race condition in a multithreaded test)

donoghuc (approved):

PR #2680, opened 2023-01-17
(SERVER-3244) Use puppetserver-ca gem that support Ruby 3.2

justinstoller:


justinstoller:
To unblock work for Puppet 8's CI we need to have Puppet Server CA CLI support Ruby 3.2. @joshcooper did this work and now we need to bump it in our main branch. I've chosen to not branch Puppet Server CA CLI and to put this in 7.x because we will be following up with work that we want in both 7.x & main and the changes are backwards compatible.

We will need to merge up though to resolve the agent's CI.

donoghuc (approved):

PR #2682, opened 2023-01-18
(maint) Do not use Ruby 2.5 workarounds outside of 6.x branch

justinstoller:


justinstoller:
This is something that could be pushed directly, but I've put the PR up for documentation purposes.

PR #2691, opened 2023-02-15
(PE-35325) Update Puppet Server to include the requester's info

ragingra:
Added an optional dependency for the ActivityReportingService, as well as new functionality for attempting to extract CA signee data from requests to sign and revoke certificates. With preference for RBAC user details, but will fall back to the name linked to the certificate used to sign, and defaulting to the “CA” signee if no other information is available. When a certificate event occurs, log messages will now include information on who signed the certificate. Additionally, if the activity service is available, the information will be submitted there as well.

Added tests checking updated log messages and also mocked activity service to check logged activities.

Manually tested signing, plus revoking individual and multiple.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -1217,13 +1217,47 @@
         combined-list (vec (concat base-ext-list csr-ext-list))]
     (ensure-ext-list-has-cn-san subject combined-list)))
 
+(defn report-cert-event
+  "Log message and report to the activity service if available about cert activties, ie signing and revoking."
+  [report-activity message subject certnames ip-address activity-type]
+  (let [commit {:service {:id "puppet-ca"}
+                :subject {:id subject
+                          :name subject
+                          :type "users"}
+                :objects (mapv (fn [cert] {:type "node" :id cert :name cert}) certnames)
+                :events [{:type (str activity-type "-certificate")
+                          :what "node"
+                          :description (str "certificate_successfully_" activity-type)
+                          :message message}]
+                :ip_address ip-address}]
+    (log/info message)
+    (report-activity {:commit commit})))
+
+(defn generate-cert-message-from-request
+  "Extract params from request and create successful cert signing message.
+  Returns message, subject, certname and ip address"
+  [request subjects activity-type]
+  (let [auth-name (get-in request [:authorization :name])
+        rbac-user (get-in request [:rbac-subject :login])
+        ip-address (:remote-addr request)
+        signee (first (remove clojure.string/blank? [rbac-user auth-name "CA"]))]
+
+    [(i18n/trsn "Entity {1} {2} 1 certificate: {3}." 
+                "Entity {1} {2} {0} certificates: {3}." 
+                (count subjects) signee activity-type (str/join ", " subjects))
+     signee
+     subjects
+     ip-address]))
+
 (schema/defn ^:always-validate
   autosign-certificate-request!
   "Given a subject name, their certificate request, and the CA settings
   from Puppet, auto-sign the request and write the certificate to disk."
   [subject :- schema/Str
    csr :- CertificateRequest
-   {:keys [cacert cakey signeddir ca-ttl serial cert-inventory]} :- CaSettings]
+   {:keys [cacert cakey signeddir ca-ttl serial cert-inventory]} :- CaSettings
+   report-activity
+   request]
   (let [validity    (cert-validity-dates ca-ttl)
         ;; if part of a CA bundle, the intermediate CA will be first in the chain
         cacert      (utils/pem->ca-cert cacert cakey)
@@ -1237,8 +1271,9 @@
                                             (utils/get-public-key csr)
                                             (create-agent-extensions
                                              csr
-                                             cacert))]
-    (log/info (i18n/trs "Signed certificate request for {0}" subject))
+                                             cacert))
+        [msg signee certnames ip] (generate-cert-message-from-request request [subject] "signed")]
+    (report-cert-event report-activity msg signee certnames ip "signed")
     (write-cert-to-inventory! signed-cert cert-inventory)
     (write-cert signed-cert (path-to-cert signeddir subject))))
 
@@ -1335,7 +1370,9 @@
    Throws a slingshot exception if the CSR is invalid."
   [subject :- schema/Str
    certificate-request :- InputStream
-   {:keys [autosign csrdir ruby-load-path gem-path allow-subject-alt-names allow-authorization-extensions] :as settings} :- CaSettings]
+   {:keys [autosign csrdir ruby-load-path gem-path allow-subject-alt-names allow-authorization-extensions] :as settings} :- CaSettings
+   report-activity
+   request]
   (with-open [byte-stream (-> certificate-request
                               input-stream->byte-array
                               ByteArrayInputStream.)]
@@ -1349,7 +1386,7 @@
         (ensure-no-authorization-extensions! csr allow-authorization-extensions)
         (validate-extensions! (utils/get-extensions csr))
         (validate-csr-signature! csr)
-        (autosign-certificate-request! subject csr settings)
+        (autosign-certificate-request! subject csr settings report-activity request)
         (fs/delete (path-to-cert-request csrdir subject))))))
 
 (schema/defn ^:always-validate delete-certificate-request! :- OutcomeInfo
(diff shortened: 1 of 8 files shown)


src/clj/puppetlabs/services/ca/certificate_authority_service.clj
```diff
@@ -9,15 +9,19 @@
             [puppetlabs.services.protocols.ca :refer [CaService]]
             [puppetlabs.comidi :as comidi]
             [puppetlabs.i18n.core :as i18n]
-            [puppetlabs.trapperkeeper.services.status.status-core :as status-core]))
+            [puppetlabs.trapperkeeper.services.status.status-core :as status-core]
+            [puppetlabs.rbac-client.protocols.activity :refer [ActivityReportingService]]
+            [puppetlabs.trapperkeeper.services :refer [maybe-get-service]]
+            [puppetlabs.rbac-client.services.activity :refer [report-activity]]))
```
justinstoller, line 15: Here you want to pull in https://github.com/puppetlabs/clj-rbac-client/blob/master/src/puppetlabs/rbac_client/protoc/activity.clj#L3 instead and call the protocol function `report-activity!` instead of the helper function in the service file `report-activity`.

justinstoller: Here's an example of requiring a protocol: https://github.com/puppetlabs/puppetserver/blob/main/src/clj/puppetlabs/services/master/master_core.clj#L22 and then using it: https://github.com/puppetlabs/puppetserver/blob/main/src/clj/puppetlabs/services/master/master_core.clj#L462-L464

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1243,15 +1243,46 @@
         combined-list (vec (concat base-ext-list csr-ext-list))]
     (ensure-ext-list-has-cn-san subject combined-list)))
 
+(defn report-successful-cert-signing
+  "Log message and report to the activity service if available about successful cert signing."
+  [report-activity-or-nil message subject certname ip-address]
+  (let [commit {:service {:id "puppetserver"}
```
justinstoller, line 1249: I think the service id should probably be "puppet-ca" which is what we name the CA's trapperkeeper service.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1243,15 +1243,46 @@
...
+(defn generate-cert-message-from-request
+  "Generate cert signing message from request parameters"
+  [request certname report-activity-or-nil]
+  (let [allow-unauthenticated (:allow-unauthenticated request)
+        auth-name (get-in request [:authorization :name])
+        rbac-subject (:rbac-subject request)
+        ip-address (:remote-addr request)
+        report-cert (partial report-successful-cert-signing report-activity-or-nil)]
+    (cond
+      (some? allow-unauthenticated) (report-cert (i18n/trs "Certificate autosigned for host {0}" certname) "CA" certname ip-address)
```
justinstoller, line 1272: I don't think `allow-unauthenticated` necessarily means it was autosigned (though it would be very dangerous if they disabled auth on the CA routes).

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1243,15 +1243,46 @@
...
+      (some? rbac-subject) (report-cert (i18n/trs "User {0} signed cert for host {1}" rbac-subject certname) rbac-subject certname ip-address)
+      :else (report-cert (i18n/trs "Host {0} signed cert for host {1}" auth-name certname) auth-name certname ip-address))))
```
justinstoller, line 1274: We could probably reduce this messaging to "Entity '{0}' signed cert for host '{1}'." where the first param is one of rbac-subject, auth-name, "CA", or "unknown".

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1243,15 +1243,46 @@
...
+                :events [{:type "sign-certificate"
+                          :what "node"
+                          :description (str "certificate_successfully_signed")
+                          :message message}]
+                :ip_address ip-address}]
+    (log/info message)
+    (if report-activity-or-nil
+      (report-activity-or-nil {:commit commit}))))
+
+(defn generate-cert-message-from-request
```
justinstoller, line 1263: This function name is a little misleading since it doesn't _just_ generate the message but also does the reporting. We should probably either just have this return message. Maybe have it return the message, ip, subject, and certname and rename it? Or just rename the function as it is to something else.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1264,7 +1294,7 @@
                                             (create-agent-extensions
                                              csr
                                              cacert))]
-    (log/info (i18n/trs "Signed certificate request for {0}" subject))
+    (apply report-successful-cert-signing report-activity-or-nil (generate-cert-message-from-request request subject))
```
justinstoller, line 1297: I think apply isn't super idiomatic in our code bases and moving the `generate-cert-message-from-request` into the let block above and pulling out the returned values, eg:
```clj
(let [...snip...
      [msg signee certname ip] (generate-cert-message-from-request)]
  (report-successful-cert-signing report-activit-or-nil msg signee certname ip)
```
but both approaches are correct and this might just be a me preference.

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
```diff
@@ -9,15 +9,19 @@
...
-            [puppetlabs.trapperkeeper.services.status.status-core :as status-core]))
+            [puppetlabs.trapperkeeper.services.status.status-core :as status-core]
+            [puppetlabs.rbac-client.protocols.activity :refer [ActivityReportingService]]
+            [puppetlabs.trapperkeeper.services :refer [maybe-get-service]]
+            [puppetlabs.rbac-client.protocols.activity :refer [report-activity!]]))
 
 (tk/defservice certificate-authority-service
   CaService
-  [[:PuppetServerConfigService get-config get-in-config]
+   {:required [[:PuppetServerConfigService get-config get-in-config]
```
justinstoller, line 19: Here's how this should be formatted:
```clj
dev-tools=> (clojure.pprint/pprint {:required [[:PuppetServerConfigService 'get-config 'get-in-config] [:WebroutingService 'add-ring-handler 'get-route]] :optional [[:Thing 'stuff]]})
{:required
 [[:PuppetServerConfigService get-config get-in-config]
  [:WebroutingService add-ring-handler get-route]],
 :optional [[:Thing stuff]]}
```

src/clj/puppetlabs/services/ca/certificate_authority_service.clj
```diff
@@ -31,14 +34,18 @@
           host-crl-file (.getCanonicalPath (fs/file
                                             (get-in-config [:puppetserver :hostcrl])))
           infra-nodes-file (.getCanonicalPath (fs/file (str (fs/parent ca-crl-file) "/infra_inventory.txt")))
-          watcher (create-watcher {:recursive false})]
+          watcher (create-watcher {:recursive false})
+          report-activity-or-nil (if-let [activity-reporting-service (maybe-get-service this :ActivityReportingService)]
+                                   (partial activity-proto/report-activity! activity-reporting-service)
+                                   nil)
```
jonathannewman, line 40: Just a thought, but you could use `(constantly nil)` to provide a no-op function in place of the `nil` here, an then the "report-activity" could always be a function.  That might reduce the cognitive burden of tracking it a bit?  Would also simplify `report-cert-event` slightly.  

https://clojuredocs.org/clojure.core/constantly

jonathannewman:
Don't forget to update the commit message with a description of what was changed and why.  This is a public repo and the internal Jira tickets aren't visible to everyone.  It also helps future archeologists with their burdens. :)

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
```diff
@@ -522,9 +532,51 @@
...
+                msg-matcher (re-pattern (str "Entity " signee " signed 1 certificate: " certname))
+                expected-path (ca/path-to-cert (:signeddir settings) certname)]
+             (try
+               (let [response (handle-put-certificate-request! settings nil request)]
+                (is (= 200 (:status response))) 
+                (is (logged? msg-matcher :info)))
+             (finally
+               (fs/delete expected-path)))))
+
+     (testing "when no signee info provided"
```
justinstoller, line 568: You should wrap each of these testing blocks in a `with-test-logging` block, or ideally in their own deftest, to ensure state doesn't get polluted between them.

dev/bootstrap.cfg
```diff
@@ -17,6 +17,9 @@ puppetlabs.services.jruby.jruby-metrics-service/jruby-metrics-service
 puppetlabs.services.analytics.analytics-service/analytics-service
 puppetlabs.trapperkeeper.services.watcher.filesystem-watch-service/filesystem-watch-service
 
+puppetlabs.rbac-client.services.activity/remote-activity-reporter
```
justinstoller, line 20: I don't think we want this in here

justinstoller (approved):

PR #2694, opened 2023-03-14
(SERVER-3246) Add linting to GH PRs

justinstoller:


justinstoller:
TODO: Still need add Rubocop!

jonathannewman:
Is there an objection to converting the `(not (empty? cn))` to `(seq cn)`?  They should be functionally equivalent, and the second is faster.

justinstoller:
> Is there an objection to converting the (not (empty? cn)) to (seq cn)? They should be functionally equivalent, and the second is faster.

Yeah, the `(not (empty? cn))` doesn't happen in the conditional of an `if` but using the presence of the `cn` to determine whether or not the request was authenticated. So the output is `{:authenticated (not (empty? cn))}`, semantically should be a boolean and is passed to other non-local consumers.

jonathannewman:
> > Is there an objection to converting the (not (empty? cn)) to (seq cn)? They should be functionally equivalent, and the second is faster.
> 
> Yeah, the `(not (empty? cn))` doesn't happen in the conditional of an `if` but the block is using the non-emptiness of the `cn` to determine whether or not the request was authenticated. So the output is `{:authenticated (not (empty? cn))}`, semantically should be a boolean and (though it isn't locally enforced with a schema) is passed to other non-local consumers so I'm very hesitant to change the functions API.

Makes total sense.

donoghuc:
Did you target main to reduce risk in the 7.y series? I could see keeping them  (7.y and main) as similar as possible being desirable.

justinstoller:
It didn't specify a fix version so I assumed it was supposed to target to main. I can rebase to 7.x (there will need to be a custom merge up, iirc). I'll do that.

justinstoller:
Rebased onto 7.x. I _may_ have royally screwed this one up, we'll see.

.github/workflows/clojure-linting.yaml
```diff
@@ -0,0 +1,33 @@
+name: Clojure Linting
+
+on:
+  push:
+    branches: [main]
```
donoghuc, line 5: I think that running this again only on `main` when merged/force-pushed is a bit unnecessary and gets in to the mess of trying to keep the branches straight here (we are already behind on https://github.com/puppetlabs/orchestrator/blob/e830895efecd06b4eb24fb9b376a8fd37b610d34/.github/workflows/clojure-linting.yaml#L5 ). Specifically i think we decided you have to list all the branches you want in all the files across branches that define a workflow with a common name. 

I think it makes sense to only run this on pull request, if you want to run in every time code lands in a branch we will have to include the list of all active branches under `on.push.branches` i think.

justinstoller: Good point. And I missed adding 7.x to it! I'll update

donoghuc (approved):
image

PR #2700, opened 2023-04-03
(PE-35786) allow trailing dot in subject validation

jonathannewman:
This updates the validation to allow trailing dots, and the subjects to start with an underscore, and have segments that start or end in underscores.

Tests are updated to reflect the updated behaviors.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -765,8 +765,8 @@
 
 
 (def pattern-match-dot #"\.")
-(def pattern-starts-with-alphanumeric #"^\p{Alnum}.*")
-(def pattern-matches-alphanumeric-with-symbols-string #"^[\p{Alnum}.\-_]+\p{Alnum}$")
+(def pattern-starts-with-alphanumeric-or-underscore #"^[\p{Alnum}_].*")
+(def pattern-matches-alphanumeric-with-symbols-string #"^[\p{Alnum}\-_]*[\p{Alnum}_]$")
 
 (schema/defn validate-subject!
   "Validate the CSR or certificate's subject name.  The subject name must:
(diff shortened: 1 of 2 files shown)


src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -795,15 +795,14 @@
...
-            (str/ends-with? subject "-"))
+  (when (str/ends-with? subject "-")
     (log/info (i18n/tru "Rejecting subject \"{0}\" as it ends with an invalid character" subject))
     (sling/throw+
      {:kind :invalid-subject-name
       :msg  (i18n/tru "Subject hostname format is invalid")}))
 
   (let [segments (str/split subject pattern-match-dot)]
-    (when-not (re-matches pattern-starts-with-alphanumeric (first segments))
+    (when-not (re-matches pattern-starts-with-alphanumeric-or-underscore (first segments))
```
joshcooper, line 805: Is this code path used to validate the subject for the puppet root CA or intermediate CA, e.g. 
```
        Subject: CN=Puppet CA: anaerobic-dram.delivery.puppetlabs.net
```

Sharpie: Any validation of the CA subjects should be over in the `puppetserver ca setup` or `puppetserver ca import` actions:

https://github.com/puppetlabs/puppetserver-ca-cli/tree/main/lib/puppetserver/ca/action

And is done while Puppet Server is offline because CA needs to be created, and a Puppet client certificate issued, in order for Puppet Server to start and serve the CA API.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -795,15 +795,14 @@
       {:kind :invalid-subject-name
        :msg  (i18n/tru "Subject contains a wildcard, which is not allowed: {0}" subject)}))
 
-  (when (or (str/ends-with? subject ".")
-            (str/ends-with? subject "-"))
+  (when (str/ends-with? subject "-")
```
donoghuc, line 798: Is this just for logging purposes? I think line 805 would handle this case?

donoghuc, line 798: IS this important for logging? I think the check on line 811 would catch this https://github.com/puppetlabs/puppetserver/pull/2700/files#diff-ac2e29a707d8bb0a3cca1bdb78e91e56a4a9a9213fb1fde7839b628837f0cab1R811

src/clj/puppetlabs/puppetserver/certificate_authority.clj
jonathannewman: This path is only for agent certs, and not CAs.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
jonathannewman: I think from a support standpoint it is useful output.  The other message is more generic and doesn't describe WHY it is invalid.  I don't feel strongly about it, and I do think we need to have much better logging diagnostics in our applications in general.

donoghuc (approved):

PR #2704, opened 2023-04-12
(SERVER-3277) Update Puppet Server CI for Puppet Platform 8

justinstoller:


justinstoller:
It's probably best to just ignore the contents of the fixtures dir and/or commit [3d27546](https://github.com/puppetlabs/puppetserver/pull/2704/commits/3d275466ac20bf991f6a129dd5ef4e261811c00a)

justinstoller:
I've tested this with redhat 8 (and an earlier version of redhat 7), so I know at least one support platform passes. I'm interested instead of continuing to run them locally against all the platforms instead putting this into Jenkins and seeing if there's any other Jenkins-specific environmental issues I missed as well.

donoghuc (approved):
Looks good enough to let Jenkins show us what we may have missed :yolo:

PR #2706, opened 2023-04-13
(maint) Target version 8.0.0

justinstoller:


justinstoller:
Version bumps are things we usually automate so I don't think this requires a review. We missed bumping this earlier and I'm trying to make sure everything is sorted well before the FC next week.

PR #2708, opened 2023-04-14
(maint) Fix bionic pdb install

justinstoller:


justinstoller:
This is the error we're encountering in CI: https://bugs.launchpad.net/ubuntu/+source/ssl-cert/+bug/1945774 and what I implement in here is the suggested fix, afaict.

PR #2710, opened 2023-04-18
(maint) fix i18n issues

jonathannewman:
In a previous commit https://github.com/puppetlabs/puppetserver/commit/a029e7762c07d1d47947edcef6485e5fcd59c2c2 some new strings were added to improve supportability.  While bench testing those issues, it was observed that some strings were not expanded correctly with multiple arguments. The issue turned out to be an escape issue with a single quote.  This resolves that issue, and replaces the use of `format` from that previous commit with the more appropriate i18n translation functions.


steveax (approved):
👍

PR #2711, opened 2023-04-18
(maint) Actually ship to puppet8 related repos

justinstoller:
This is the other half of https://github.com/puppetlabs/ci-job-configs/pull/8813 that was missed.


donoghuc (approved):

justinstoller:
Travis failed because our pin of Puppet now will not load in Ruby <3 and our CA tests shell out to the system ruby which on Focal is 2.7.

I'll add a commit updating us to use Jammy as it uses Ruby 3+ by default.

PR #2716, opened 2023-04-26
(PE-35638) make crl file manipulation routines thread safe

jonathannewman:
This updates the ca-settings to have two new locks, one for the crl, and one for the infra-crl along with appropriate timeouts for each.  The locks are used in the routines that read and write the crl files to ensure they are only read when they aren't being written. 

Tests are updated as appropriate.  See individual commits comments for more details.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -108,9 +108,7 @@
    :serial-lock                      ReentrantReadWriteLock
    :serial-lock-timeout-seconds      PosInt
    :crl-lock                         ReentrantReadWriteLock
-   :crl-lock-timeout-seconds         PosInt
-   :infra-crl-lock                   ReentrantReadWriteLock
-   :infra-crl-lock-timeout-seconds   PosInt})
+   :crl-lock-timeout-seconds         PosInt})
 
 (def DesiredCertificateState
   "The pair of states that may be submitted to the certificate
@@ -183,10 +181,6 @@
   ;; for large crls, and a slow disk a longer timeout is needed
   60)
 
-(def default-infra-crl-lock-timeout-seconds
-  ;; for large crls, and a slow disk a longer timeout is needed
-  60)
-
 (schema/defn ^:always-validate initialize-ca-config
   "Adds in default ca config keys/values, which may be overwritten if a value for
   any of those keys already exists in the ca-data"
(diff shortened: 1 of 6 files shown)


src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -174,6 +178,12 @@
 (def default-serial-lock-timeout-seconds
   5)
 
+(def default-crl-lock-timeout-seconds
+  5)
```
jonathannewman, line 182: Unclear if this is sufficient for customers with very large CRLs.

test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
```diff
@@ -273,7 +277,7 @@
   (let [settings (assoc (testutils/ca-sandbox! cadir)
                         :allow-duplicate-certs true
                         :autosign false)
-        request {:authorization {:name "authname"} :remote-addr "1.1.1.1" }]
+        request {:authorization {:name "authname"} :remote-addr "1.1.1.1"}]
```
jonathannewman, line 280: Cursive made this, and the paren changes below, because the indentation didn't match the parens.

shaigy (approved):

justinstoller:
Writing to the CRLs is already serialized with this https://github.com/puppetlabs/puppetserver/pull/2716/files#diff-caf7efdcbdce3e60f52b15dd0489e7f16cad9c1dcfe2993ef1099ec843a50db8R121 . It seems like upgrading this lock to use the ReadWrite locks with timeouts would be sufficient. I'm also leery of acquiring and releasing two locks in a single function should we remove the overall lock later on.

jonathannewman:
> Writing to the CRLs is already serialized with this https://github.com/puppetlabs/puppetserver/pull/2716/files#diff-caf7efdcbdce3e60f52b15dd0489e7f16cad9c1dcfe2993ef1099ec843a50db8R121 . It seems like upgrading this lock to use the ReadWrite locks with timeouts would be sufficient. I'm also leery of acquiring and releasing two locks in a single function should we remove the overall lock later on.

Yeah, I missed that (it was mentioned in the ticket), I meant to remove it.  Will do.

jonathannewman:
Thanks for catching that!

jonathannewman:
I pushed a separate commit to remove the other lock and apply locking in the parts that were exposed to potential races.  I think it is easier to review with whitespace changes hidden.

justinstoller:
I think, despite them being two different files and us reading them separately, conceptually we never act on only the infra or full crl. If we update one we always update the other (if it exists). So I think putting writing to them behind a single lock makes sense. ie, managing two locks for them adds a level of specificity (and complexity) that I don't think matters at the functional level.
If we did that I think we could also remove passing locks into `write-crls` and make that method private`, allowing the lock to be handled by `write-crls` callers in `update-crls` and `revoke-existing-certs`.

jonathannewman:
That seems reasonable.

jonathannewman:
I pushed a commit to address simplifying the crl locking.  Happy to squash the commits, but thought it would be helpful to see them separately.

shaigy:
LGTM

PR #2721, opened 2023-04-28
(PE-35641) promote inventory content safety

jonathannewman:
This adds a ReentrantLock that is used to protect multiple threads from writing to the inventory file at the same time.  This is important because the inventory update routine now has to read in the contents prior to appending a new entry. If two separate threads are updating the file at the same time, there is the possiblity that one of the entries would be lost.

Tests were updated as needed.


steveax (approved):
👍

PR #2722, opened 2023-05-04
(PE-35007) Stop overriding jruby from clj-parent

donoghuc:
We want to go back to using the jruby from clj-parent in the main branch. This commit updates the puppetserver project to use the version contained in clj-parent.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -51,8 +51,7 @@
                  ;; send their logs to logstash, so we include it in the jar.
                  [net.logstash.logback/logstash-logback-encoder]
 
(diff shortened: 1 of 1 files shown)


justinstoller (approved):

justinstoller:
fwiw, once this is promoted into pepse we should remove this nonsense: https://github.com/puppetlabs/pe-puppet-server-extensions/blob/main/project.clj#L50-L53 (and a similar line in the test profile).

donoghuc:
Ah, thanks. Updated https://github.com/puppetlabs/pe-puppet-server-extensions/pull/1457

PR #2723, opened 2023-05-05
(maint) Use jruby-deps from jruby-utils in the ezbake profile

justinstoller:


justinstoller:
Sorry, @donoghuc , I forgot about this half of 096a85d36 when I was telling you what was needed.

donoghuc (approved):

PR #2724, opened 2023-05-10
(PE-35699) Create duration string parsing function

rileynewton:
Adds duration string parsing function, I made it robust, able to take in invalid strings, but perhaps it should reject conversion entirely if there are any invalid strings?

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -1987,6 +1987,7 @@
                     (format-date-time))))
             {}
             crl-chain)))
+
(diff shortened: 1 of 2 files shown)


jonathannewman:
Given this is intended to be used of interpret user input from the config, I think anything malformed should be ignored.  In other words, we shouldn't try to match things unless the whole string is a valid duration.

rileynewton:
> Given this is intended to be used of interpret user input from the config, I think anything malformed should be ignored. In other words, we shouldn't try to match things unless the whole string is a valid duration.

Okay, I assume whitespace is irrelevant in the matching?

jonathannewman (approved):
Looks good!

PR #2725, opened 2023-05-10
(PE-35706) add certificate validity function

jonathannewman:
This adds a new function `cert-authority-id-match-ca-subject-id?` that validates that for a given cert and a ca-cert, that the cert was issued by the ca-cert.

Tests were added to demonstrate the behavior.


steveax (approved):
👍

PR #2727, opened 2023-05-11
(PE-35709) add in renew-certificate function

jonathannewman:
In the first commit:

This simplifies the mechanism for activity reporting to centralize the message creation and reporting into a separate function that can be called with the subjects and type of reporting without having to pass the request around. It simplifies the testing required, and creates reuse.

In the second commit:
Given an existing certificate, which is assumed to be from the
same issuer (caller to validate), generate a new signed cert
with the same extensions, and a new expiration period determined
by the auto-renewal-cert-ttl time.

Note that the issuer and subject identifier extensions need to be
regenerated for signing to occur properly.

Tests are added to demonstrate the behavior of the function.


steveax (approved):
👍

PR #2728, opened 2023-05-12
PE-35698 Add new certificate-authority parameters

shaigy:
Add three new parameters to certificate-authority. allow-auto-renewal - to toggle auto renewal of certs. auto-renewal-cert-ttl - determines the lifetime of cert. ca-ttl - default expiration time. If ca-ttl is set in certificate-authority, it will override the ca-ttl value in puppet conf.

### DIFF
--- a/ezbake/config/conf.d/ca.conf
+++ b/ezbake/config/conf.d/ca.conf
@@ -7,4 +7,10 @@ certificate-authority: {
 
     # enable the separate CRL for Puppet infrastructure nodes
     # enable-infra-crl: false
+    # Disable auto renewal of certs by default.
+    allow-auto-renewal: false
+    # This value determines the lifetime of the cert if auto-renewal is enabled
+    auto-renewal-cert-ttl: "60d"
+    # Default cert expiration time. If the value is set here, it will take precedence over ca-ttl setting in puppet.conf
+    #ca-ttl: "60d"
 }
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -31,6 +31,54 @@
             [puppetlabs.i18n.core :as i18n]))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Public utilities
+
+;; Pattern is one or more digit followed by time unit
+(def digits-with-unit-pattern #"(\d+)(y|d|h|m|s)")
+(def repeated-digits-with-unit-pattern #"((\d+)(y|d|h|m|s))+")
+
+(defn duration-string?
+  "Returns true if string is formatted with duration string pairs only, otherwise returns nil.
+   Ignores whitespace."
+  [maybe-duration-string]
+  (when (string? maybe-duration-string)
+    (let [no-whitespace-string (clojure.string/replace maybe-duration-string #" " "")]
+      (some? (re-matches repeated-digits-with-unit-pattern no-whitespace-string)))))
+
+(defn duration-str->sec
+  "Converts a string containing any combination of duration string pairs in the format '<num>y' '<num>d' '<num>m' '<num>h' '<num>s'
+   to a total number of seconds.
+   nil is returned if the input is not a string or not a string containing any valid duration string pairs."
+  [string-input]
+  (when (duration-string? string-input)
+    (let [pattern-matcher (re-matcher digits-with-unit-pattern string-input)
+          first-match (re-find pattern-matcher)]
+      (loop [[_match-str digits unit] first-match
+             running-total 0]
+        (let [unit-in-seconds (case unit
+                                "y" 31536000 ;; 365 day year, not a real year
+                                "d" 86400
+                                "h" 3600
+                                "m" 60
+                                "s" 1)
+              total-seconds (+ running-total (* (Integer/parseInt digits) unit-in-seconds))
+              next-match (re-find pattern-matcher)]
+          (if (some? next-match)
+            (recur next-match total-seconds)
+            total-seconds))))))
+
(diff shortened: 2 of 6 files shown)


ezbake/config/conf.d/ca.conf
```diff
@@ -7,4 +7,7 @@ certificate-authority: {
 
     # enable the separate CRL for Puppet infrastructure nodes
     # enable-infra-crl: false
+    #allow-auto-renewal: false
```
jonathannewman, line 10: This file gets installed, can you add some explanation comments above each entry so users will understand what they are for?  I can help with the wording if that would be helpful.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -78,14 +129,16 @@
...
+   :auto-renewal-cert-ttl            TTLDuration
    :autosign                         AutoSignInput
    :cacert                           schema/Str
    :cadir                            schema/Str
    :cacrl                            schema/Str
    :cakey                            schema/Str
    :capub                            schema/Str
    :ca-name                          schema/Str
-   :ca-ttl                           schema/Int
+   :ca-ttl                           TTLDuration
```
jonathannewman, line 141: I think this should remain an Int throughout, and we should convert it from a string to an int when we construct the settings.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -186,6 +239,9 @@
 (def default-inventory-lock-timeout-seconds
   60)
 
+(def default-auto-renewal-cert-ttl
+  "60d")
```
jonathannewman, line 243: We should do the math here, and call out what it is, `default-auto-renewal-cert-ttl-seconds`.  I have something similar in my open PR, I believe.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -199,7 +255,9 @@
                   :allow-authorization-extensions default-allow-auth-extensions
                   :serial-lock-timeout-seconds default-serial-lock-timeout-seconds
                   :crl-lock-timeout-seconds default-crl-lock-timeout-seconds
-                  :inventory-lock-timeout-seconds default-inventory-lock-timeout-seconds}]
+                  :inventory-lock-timeout-seconds default-inventory-lock-timeout-seconds
+                  :allow-auto-renewal false
+                  :auto-renewal-cert-ttl default-auto-renewal-cert-ttl}]
```
jonathannewman, line 260: If we convert the default to seconds and a number, we should be good here.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -29,6 +29,54 @@
...
+            total-seconds))))))
+
+(defn get-ca-ttl
+  "Returns ca-ttl value as an integer. If a value is set in certificate-authority that value is returned.
+   Otherwise puppet config setting is returned"
+  [puppetserver certificate-authority]
+  (when (not-empty (select-keys certificate-authority [:ca-ttl]))
+    (update certificate-authority :ca-ttl duration-str->sec)
+    (when (not-empty (select-keys puppetserver [:ca-ttl]))
+      (log/warn (i18n/trs) "Detected ca-ttl setting in puppetserver.conf which will take precedence over puppet.conf setting")))
```
jonathannewman, line 76:
```suggestion
      (log/warn (i18n/trs) "Detected ca-ttl setting in server configuration, which will take precedence over puppet.conf setting")))
```

It isn't clear which configuration file will have this value, so I think we should be more generic.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -29,6 +29,54 @@
...
+          (if (some? next-match)
+            (recur next-match total-seconds)
+            total-seconds))))))
+
+(defn get-ca-ttl
+  "Returns ca-ttl value as an integer. If a value is set in certificate-authority that value is returned.
+   Otherwise puppet config setting is returned"
+  [puppetserver certificate-authority]
+  (when (not-empty (select-keys certificate-authority [:ca-ttl]))
+    (update certificate-authority :ca-ttl duration-str->sec)
```
jonathannewman, line 74: This statement by itself won't change the value in `certificate-authority`, so you will need to restructure this.  Perhaps something like:
```
(let [ca-config-value (duration-str->sec (:ca-ttl certificate-authority))
        puppet-config-value (:ca-ttl puppetserver)]
        (when (and ca-config-value puppet-config-value)
            (log/warn (i18n/trs) "Detected ca-ttl setting in CA config which will take precedence over puppet.conf setting"))
        (or ca-config-value puppet-config-value))
```

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1230,12 +1290,14 @@
       (merge (select-keys certificate-authority (keys CaSettings)))
       (initialize-ca-config)
       (assoc :ruby-load-path (:ruby-load-path jruby-puppet)
+             :auto-renewal-cert-ttl (duration-str->sec (:auto-renewal-cert-ttl certificate-authority))
```
jonathannewman, line 1293: We need `allow-auto-renewal` to be extracted from the config and added to the structure too.

shaigy: allow-auto-renewal will be added in the initialize-ca-config here https://github.com/puppetlabs/puppetserver/pull/2728/files#diff-ac2e29a707d8bb0a3cca1bdb78e91e56a4a9a9213fb1fde7839b628837f0cab1R259 right?

jonathannewman: Yes, the default is set there.  This is the point where we override the defaults with the values from the configuration file.

jonathannewman (approved):
LGTM! 
authority?

PR #2732, opened 2023-05-13
Fix: Integration test - last one

david4ilmor:
I'm sending all the technical details using your reporting address


david4ilmor:
An email was sent to the address mentioned here: https://www.puppet.com/security

Sorry in advance for the multiply notification you may received due to this test.

PR #2735, opened 2023-05-16
(maint) enable pr testing through github actions

jonathannewman:
This enables running the clojure tests via github actions on the matrix of versions of java, multithreaded support, and with fips.

### DIFF
--- a/.github/workflows/pr-testing.yaml
+++ b/.github/workflows/pr-testing.yaml
@@ -1,0 +1,44 @@
+name: PR Testing
+
+on:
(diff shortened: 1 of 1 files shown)


jonathannewman:
This won't be enabled on PR until it gets merged.  Previous tests that were allowed to run on push can be seen here:  https://github.com/puppetlabs/puppetserver/actions/runs/4994314876

justinstoller (approved):
Looks sane. Lets merge and if we need to iterate we can do so once it's running.

PR #2738, opened 2023-05-17
(PE-35705) extract certificate from renewal request

steveax:
This commit updates the handle-cert-renewal function to extract certificates present in the request. Certificates can be included either as a ssl-client-cert property in the request, or as an x-client-cert field in the request header.

Imports for certificate-authority-core and certificate-authority-int-test were alphabetized as well.


jonathannewman (approved):

jonathannewman:
LGTM!

PR #2742, opened 2023-05-17
(PE-35710) add in certificate renewal behavior

jonathannewman:
This adds in the certificate renewal behavior, assuming that the incoming certificate has been properly validated.  Tests are updated to demonstrate that newly generated certificates can be correctly renewed.


steveax:
closing in favor of #2748

PR #2744, opened 2023-05-18
(maint) differentiate test names

jonathannewman:
The single-threaded and multi-threaded versions of `admin-api-flush-jruby-pool-test` were named nearly the same
thing, which was confusing. This changes the naming to make them different, and forces the multithreaded
test to specify the multithreaded flag to guarantee the test runs in multithreaded mode regardless of the environment variable setting.

### DIFF
--- a/test/integration/puppetlabs/puppetserver/auth_conf_test.clj
+++ b/test/integration/puppetlabs/puppetserver/auth_conf_test.clj
@@ -31,8 +31,7 @@
   (testutils/with-puppet-conf (fs/file test-resources-dir "puppet.conf")))
 
 (deftest ^:integration request-with-ssl-cert-handled-via-tk-auth
-  (testing (str "Request with SSL certificate via trapperkeeper-authorization "
-                "handled")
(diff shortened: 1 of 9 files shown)


justinstoller (approved):

justinstoller:
The commit isn't true, one has "ref" in the name and one does not. However, they obviously could be better differentiated

jonathannewman:
Good catch on the naming.  Not sure I understand the comment about the ref-test.  I added the multithreaded flag, and it does seem to make it into the config as best I can tell.

justinstoller:
Sorry, I meant that I'm not sure the ref test was testing what it said it was _before_ your change since it didn't previously have the multithreaded config set.

jonathannewman:
Updated.

jonathannewman:
I have no idea why the linter has decided to complain about those things.  I think it might be because they are using `str` in the testing name?

jonathannewman:
That seems to be it.  I'll update them all.

jonathannewman:
Ok.  There is an extra commit that fixes all those.

PR #2745, opened 2023-05-18
(PE-36083) add if-modified-since handling for get-certificate endpoint

jonathannewman:
This modifies the `/v1/certificate/<hostname>` endpoint to honor the if-modified-since header to prevent sending content if the file hasn't been modified since the date specified. This will reduce the amount of traffic from the CA, and reduce the amount of handling needed on the client.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -1170,10 +1170,10 @@
   [{:keys [puppetserver]}]
   (select-keys puppetserver (keys MasterSettings)))
 
-(schema/defn ^:always-validate
-  get-certificate :- (schema/maybe schema/Str)
-  "Given a subject name and paths to the certificate directory and the CA
-  certificate, return the subject's certificate as a string, or nil if not found.
+(schema/defn ^:always-validate get-certificate-path :- (schema/maybe schema/Str)
+  "Given a subject name and paths to the CA certificate and path
+  to the certificate directory return the path to the subject's
+  certificate as a string, or nil if not found.
   If the subject is 'ca', then use the `cacert` path instead."
   [subject :- schema/Str
    cacert :- schema/Str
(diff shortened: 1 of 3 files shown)


src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1170,10 +1170,9 @@
   [{:keys [puppetserver]}]
   (select-keys puppetserver (keys MasterSettings)))
 
-(schema/defn ^:always-validate
-  get-certificate :- (schema/maybe schema/Str)
-  "Given a subject name and paths to the certificate directory and the CA
-  certificate, return the subject's certificate as a string, or nil if not found.
+(schema/defn ^:always-validate get-certificate-path :- (schema/maybe schema/Str)
+  "Given a subject name and paths to Given a subject name and paths to the certificate directory and the CA
```
shaigy, line 1174: May be "Given a subject name and path to the Given subject name..." ?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -1406,9 +1416,8 @@
   (slurp cacrl))
 
 (schema/defn ^:always-validate
-  get-crl-last-modified :- DateTime
-  "Given the value of the 'cacrl' setting from Puppet, return
-  a Joda DateTime instance of when the CRL file was last modified."
+  get-file-last-modified :- DateTime
+  "Given the a path to a file return a Joda DateTime instance of when the CRL file was last modified."
```
shaigy, line 1420: Also may be? "Given path to a file, return a Joda DateTime instance of when the file was last modified."

shaigy:
Two trivial text updates, otherwise LGTM.

jonathannewman:
Good call, @shaigy  I updated the language, is that better?

shaigy (approved):

PR #2749, opened 2023-06-07
(PE-36178) Add config block for certificate renewal endpoint

artlawson:
This adds a configuration block that allows authenticated clients to access the certificate renewal endpoint.

### DIFF
--- a/ezbake/config/conf.d/auth.conf
+++ b/ezbake/config/conf.d/auth.conf
@@ -57,6 +57,18 @@ authorization: {
             name: "puppetlabs csr"
         },
         {
+            # Allow nodes to renew their certificate
+            match-request: {
+                path: "/puppet-ca/v1/certificate_renewal"
+                type: path
+                method: post
+            }
+            # this endpoint should never be unauthenticated, as it requires the cert to be provided.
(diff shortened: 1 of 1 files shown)


ezbake/config/conf.d/auth.conf
```diff
@@ -56,6 +56,18 @@ authorization: {
...
         },
+        {
+            # Allow nodes to renew their certificate
+            match-request: {
+                path: "/puppet-ca/v1/certificate_renewal"
+                type: path
+                method: post
+            }
+            # this endpoint should never be unauthenticated, as it requires the cert to be provided.
+            allow-unauthenticated: false
```
jonathannewman, line 67: By default `allow-unauthenticated` is false, it turns out that we need to allow anything that provides a cert so: 

```suggestion
            # this endpoint should never be unauthenticated, as it requires the cert to be provided.
            "allow": "*",
```

jonathannewman: See the "puppetlabs CA cert and CRL expiration" section below as an example of what I mean.

jonathannewman (approved):
LGTM! 
Block?

PR #2752, opened 2023-06-15
(PE-36202) add indication that agent supports certificate renewal

jonathannewman:
This adds the concept of the CA determining if a given agent CSR can support renewal or not. As a stop-gap, this introspects the x-puppet-version header to check the version of the agent. If it is present, and >= 8.2.0, then it is assumed that the agent supports the feature. Also as a stop-gap, a file is written alongside of the csr to indicate that the agent supports the feature. Future work will be to use this file to choose a short ttl (assuming renewal is enabled), and clean up the temporary file as appropriate.

Additional follow-on work will include leveraging attributes/extensions present in the CSR instead of the version to indicate that the agent supports renewal.

Also as part of this work, the puppet function `versioncmp` was ported into clojure and included. Similar tests to what is found in puppet were added. See https://github.com/puppetlabs/puppet/blob/6b400f35193e5d6871fd679d6ad49776d1cb6072/lib/puppet/util/package.rb#L3

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -442,6 +442,12 @@
   [csrdir subject]
   (str csrdir "/" subject ".pem"))
 
+(defn path-to-autosign-indication
+  "Return a path to the `subject`s certificate request file under the `csrdir`."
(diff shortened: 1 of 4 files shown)


steveax (approved):
👍

PR #2753, opened 2023-06-16
(PE-36204) ensure indicator file is deleted

jonathannewman:
This moves all the CSR deletion requests to a singular function for handling the CSR deletion, which centralizes logging and provides a bottleneck for all CSR removal behaviors.

Since all calls to the "autosign" routine were always followed by calls to remove the CSR, the delete routine was moved into the autosign routine.

In the updated delete routine, a new behavior is added to look for and remove the indicator file if it is present.  The delete routine needed to be moved up in the file because it is now used by some earlier functions.

Tests are updated to reflect the new behavior and ensure the file is properly deleted.


steveax (approved):
👍

PR #2754, opened 2023-06-16
(maint) use version compare function from kitchensink

jonathannewman:
This removes the version compare function that was added as a stop-gap until the kitchensink version was available.  Now that it is, this removes the function, the utility routines and the tests from this repo.


rileynewton (approved):
👍

PR #2755, opened 2023-06-20
(PE-36205) enable short ttl if agent is capable

jonathannewman:
If the agent verision is 8.2.0 or later, and the auto-renewal feature is enabled in the CA, then initial signings of certificates should use the configured (or default if it isn't configured) short ttl.

An issue with the version comparision logic was fixed, and an appropriate test added.


jonathannewman:
I had to make a change to a subsequent test, which had incorrect indentation.  This caused a lot of changes that are whitespace only.  Much easier to read with the whitespace changes hidden.

steveax (approved):
👍

PR #2758, opened 2023-06-28
(PE-36386) remove use of clj-yaml in favor of snakeyaml 2.0

jonathannewman:
This removes the use of clj-yaml for yaml parsing because internally it uses snakeyaml 1.33. Instead, snakeyaml 2.0 is used directly and a routine for mapping from java structures to clojure structures is added.


steveax (approved):
👍

PR #2759, opened 2023-07-05
(PE-32607) change default auto-renewal ttl to 90 days

jonathannewman:
This alters the default initial ttl for agents with auto-renewal capability to 90 days from 60 days, and changes the period of validity on renewal from 60 days to 90 days.


steveax (approved):
👍

PR #2760, opened 2023-07-11
(PE-36456) add a routine to extract attributes from a CSR

jonathannewman:
This adds a new function to puppetserver that extracts attributes from CSRs, which will be needed to determine if an agent is capable of auto-renewal or not, rather than the current method of looking at the agent version.

This adds tests to demonstrate the behavior


steveax (approved):
👍

PR #2762, opened 2023-07-12
(PE-36457) use attribute in csr for auto-reneal indication

jonathannewman:
This changes the mechanism used to determine if an agent should have the shorter "auto-renewal" ttl for Certificate Signing Requests, or if it should use the default ca-ttl.

The agent will include an attribute that indicates with a "true" value that it supports auto-renewal. If the attribute is missing, or is a "false" value, the indication is that it does not support auto-renewal.

Tests are updated as needed to demonstrate the new behaviors.

The code related to creating, interrogating and deleting the indicator file have been removed.


steveax (approved):
👍

PR #2763, opened 2023-07-18
(PE-36479) remove use of clj-yaml

jonathannewman:
This removes the use of clj-yaml, and replaces it with a use of snake-yaml as implemented in trapperkeeper. This is to help resolve https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-1471

which can't be fully resolved due to a dependency on snakeyaml 1.3.3 by jRuby

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -33,7 +33,6 @@
   :dependencies [[org.clojure/clojure]
 
                  [slingshot]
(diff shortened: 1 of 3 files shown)


steveax (approved):
👍

PR #2767, opened 2023-07-24
(PE-36185) update ezbake

jonathannewman:
This updates ezbake to 2.5.2, which brings in support for el9 builds, and addresses an issue on the el platforms where tzdata-java was missing from some dependencies because of changes in the openjdk.


steveax (approved):
👍

PR #2769, opened 2023-07-31
(PE-36076) Remove content type from certificate reject

artlawson:
Previously, when a `DELETE` request was  sent to `puppet-ca/v1/certificate_status/<nodename>` a `204` response was sent with a malformed content-type header. This commit removes the header.

### DIFF
--- a/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
+++ b/src/clj/puppetlabs/services/ca/certificate_authority_core.clj
@@ -132,9 +132,12 @@
    ca-settings :- ca/CaSettings]
   (let [response (ca/delete-certificate-request! ca-settings subject)
         outcomes->codes {:success 204 :not-found 404 :error 500}]
-    (-> (rr/response (:message response))
-        (rr/status ((response :outcome) outcomes->codes))
-        (rr/content-type "text/plain"))))
+    (if (not= (response :outcome) :success) 
+      (-> (rr/response (:message response)) 
+          (rr/status ((response :outcome) outcomes->codes))
+          (rr/content-type "text/plain"))
+      (-> (rr/response (:message response))
+          (rr/status ((response :outcome) outcomes->codes))))))
 
 (schema/defn handle-get-ca-expirations
   [ca-settings :- ca/CaSettings]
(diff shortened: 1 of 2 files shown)


test/unit/puppetlabs/services/ca/certificate_authority_core_test.clj
```diff
@@ -350,7 +348,6 @@
                    msg-matcher (re-pattern (str "Path " expected-path " exists but.*"))]
                (is (= 500 (:status response)))
                (is (re-matches msg-matcher (:body response)))
-               (is (= "text/plain" (get-in response [:headers "Content-Type"])))
```
steveax, line 353: This seems odd. The 500 should return some sort of error and a Content-Type along with it. It may be that you'll need to add the content type specifically to the 500 error (which has a body).

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
```diff
@@ -132,9 +132,12 @@
    ca-settings :- ca/CaSettings]
   (let [response (ca/delete-certificate-request! ca-settings subject)
         outcomes->codes {:success 204 :not-found 404 :error 500}]
-    (-> (rr/response (:message response))
-        (rr/status ((response :outcome) outcomes->codes))
-        (rr/content-type "text/plain"))))
+    (if (= (response :outcome) :error)
```
steveax, line 135: Looks good, but we need a Content-Type (text/plain) for both the 404 and the 500 cases as the 404 also has a body (ex: `"No certificate request for not-found-agent at expected path /var/folders/dh/ccmxd9mn1kggz9stfrm04vyh0000gn/T/null1690930482830-3832794012/ca/requests/not-found-agent.pem"`)

steveax (changes requested):

src/clj/puppetlabs/services/ca/certificate_authority_core.clj
steveax: To expand, Content-Type should only be omitted if the body is empty (content-length of 0).

steveax (approved):
👍

PR #2775, opened 2023-08-10
(maint) regenerate the fixtures for CA tests to use sha256 message digests

jonathannewman:
The fixtures, prior to this change were using a sha1 message digest, which has resulted in failures appearing in el9 based testing. This updates the fixture generation script and replaces the certificate using content generated from running the script.


justinstoller:
This looks appropriate based on similar updates in https://github.com/puppetlabs/puppetserver/pull/2704
Should I run this against rh9 and/or a very old platform we test against in 7.x?

steveax (approved):
👍

jonathannewman:
@justinstoller I've run against rh9, I can also run it against an older platform as well.

PR #2778, opened 2023-08-31
(PE-36768) add authenticated route for proxied requests

jonathannewman:
This adds a new authenticated path for certificate renewal in the case that there is puppet managed proxy between the agent and the CA.

In that case, the proxy will inject the x-client-cert header into the request. This will be allowed if the requester is listed in the infra-nodes inventory list. Otherwise the request is denied.


jonathannewman:
A little easier to review with whitespace changes hidden.

justinstoller (approved):

PR #2780, opened 2023-10-03
(maint) remove ssl-utils pin

jonathannewman:
This removes the pinned version of ssl-utils from puppetserver as the main branch is now using a newer clj-parent that contains the pinned version. The previous branch that was merged in pinned the version to avoid an infinite recursion bug.


rileynewton (approved):
👍

PR #2781, opened 2023-10-05
(PE-36952) ensure infra-serial is up to date

jonathannewman:
Please see individual commit comments.

### DIFF
--- a/src/clj/puppetlabs/puppetserver/certificate_authority.clj
+++ b/src/clj/puppetlabs/puppetserver/certificate_authority.clj
@@ -673,6 +673,47 @@
 
 (def buffer-copy-size (* 64 1024))
 
+(schema/defn read-infra-nodes
+  "Returns a list of infra nodes or infra node serials from the specified file organized as one item per line."
+  [infra-file-reader :- Reader]
+  (line-seq infra-file-reader))
+
+(schema/defn maybe-write-to-infra-serial!
+  "Determine if the host in question is an infra host, and if it is, add the provided serial number to the
+  infra-serials file"
+  [serial :- BigInteger
+   certname :- schema/Str
+   {:keys [infra-nodes-path infra-node-serials-path]} :- CaSettings]
+  (when (fs/exists? infra-nodes-path)
+    (with-open [infra-nodes-reader (io/reader infra-nodes-path)]
+      (let [infra-nodes (read-infra-nodes infra-nodes-reader)]
+        (when (seq (filter #(= certname %) infra-nodes))
+          (log/debug (i18n/trs "Appending serial number {0} for {1} " serial certname))
+          (let [stream-content-fn
+                (fn [^BufferedWriter writer]
+                  (log/trace (i18n/trs "Begin append to serial file."))
+                  (let [copy-buffer (CharBuffer/allocate buffer-copy-size)]
+                    (try
+                      (with-open [^BufferedReader reader (io/reader infra-node-serials-path)]
+                        ;; copy all the existing content
+                        (loop [read-length (.read reader copy-buffer)]
+                          ;; theoretically read can return 0, which means try again
+                          (when (<= 0 read-length)
+                            (when (pos? read-length)
+                              (.write writer (.array copy-buffer) 0 read-length))
+                            (.clear copy-buffer)
+                            (recur (.read reader copy-buffer)))))
+                      (catch FileNotFoundException _e
+                        (log/trace (i18n/trs "Inventory serial file not found.  Assume empty.")))
+                      (catch Throwable e
+                        (log/error e (i18n/trs "Error while appending to serial file."))
+                        (throw e))))
+                  (.write writer (str serial))
+                  (.newLine writer)
+                  (.flush writer)
+                  (log/trace (i18n/trs "Finish append to serial file. ")))]
+            (ks-file/atomic-write infra-node-serials-path stream-content-fn)))))))
+
 (schema/defn ^:always-validate
   write-cert-to-inventory!
   "Writes an entry into Puppet's inventory file for a given certificate.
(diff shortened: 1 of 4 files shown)


src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -523,8 +523,6 @@
             {}
             auth-exts)))
 
-(defn seq-contains? [coll target] (some #(= target %) coll))
```
jonathannewman, line 526: This wasn't used, so I removed it.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -658,15 +656,64 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Inventory File
 
+(def inventory-date-formatter
```
jonathannewman, line 659: Making the formatter once and reusing it saves some time and memory churn.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -658,15 +656,64 @@
...
     (time-coerce/from-date date-time)))
 
+(schema/defn parse-date-time :- DateTime
+  "parses a date-time string into a DateTime instance"
+  [date-time :- schema/Str]
+  (time-format/parse inventory-date-formatter date-time))
+
 (def buffer-copy-size (* 64 1024))
 
+(schema/defn read-infra-nodes
```
jonathannewman, line 676: This was moved up from below.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -818,41 +885,49 @@
...
+
+(schema/defn write-infra-serials-to-writer
+  [writer :- BufferedWriter
+   settings :- CaSettings]
+  (let [infra-serials (extract-active-infra-serials settings)]
+    (doseq [infra-serial infra-serials]
+      (.write writer (str infra-serial))
+      (.newLine writer))))
+
+(schema/defn generate-infra-serials!
```
jonathannewman, line 924: Renamed this by adding the `!` suffix to indicate that it makes changes.

test/unit/puppetlabs/services/ca/ca_testutils.clj
```diff
@@ -73,7 +73,7 @@
    :serial                           (str cadir "/serial")
    :ruby-load-path                   jruby-testutils/ruby-load-path
    :gem-path                         jruby-testutils/gem-path
-   :infra-nodes-path                 (str cadir "/ca/infra_inventory.txt")
+   :infra-nodes-path                 (str cadir "/infra_inventory.txt")
```
jonathannewman, line 76: This was a typo.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -682,31 +729,33 @@
...
                                 (with-open [^BufferedReader reader (io/reader cert-inventory)]
                                   ;; copy all the existing content
                                   (loop [read-length (.read reader copy-buffer)]
-                                    (when (< 0 read-length)
-                                        (when (pos? read-length)
-                                          (.write writer (.array copy-buffer) 0 read-length))
-                                        (.clear copy-buffer)
-                                        (recur (.read reader copy-buffer)))))
+                                    ;; theoretically read can return 0, which means try again
+                                    (when (<= 0 read-length)
```
jonathannewman, line 754: This was incorrectly handling the `0` result case.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -726,6 +776,13 @@
    (if (some? row-subject)
      (= (subs row-subject 1) cn-subject)
      false))
+(schema/defn is-not-expired? :- schema/Bool
```
justinstoller, line 779: Needs a newline separating this function from the preceding one.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
```diff
@@ -717,7 +766,8 @@
                             (log/trace (i18n/trs "Finish append to inventory file. ")))]
     (common/with-safe-write-lock inventory-lock inventory-lock-descriptor inventory-lock-timeout-seconds
       (log/debug (i18n/trs "Append \"{1}\" to inventory file {0}" cert-inventory entry))
-      (ks-file/atomic-write cert-inventory stream-content-fn))))
+      (ks-file/atomic-write cert-inventory stream-content-fn))
+    (maybe-write-to-infra-serial! serial-number cert-name settings)))
```
justinstoller, line 770: Shouldn't we put this in the inventory-lock above?

src/clj/puppetlabs/puppetserver/certificate_authority.clj
justinstoller: Just a sanity check here:
When a cert is renewed the new serial will be appended to the infra-serials file, but when the CA restarts, reloads, or the infra node certname file is updated the whole infra-serials file will be rewritten with the latest serial for each infra node. That's the intended behavior?

justinstoller (commented):
Looks good overall.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
jonathannewman: The serial isn't locked under any other circumstances, but it certainly wouldn't hurt.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
jonathannewman: In the startup case, with these changes to `generate-infra-serials!` we are now looking in the inventory as well for matching serial numbers and adding them.  So at startup there should be technically no work to do, unless someone manually changed the files.  The idea is that the infra-serial should always be more correct now

jonathannewman: So to answer your question, the intent is that we append the infra-serial on demand when a new one is generated, and when we restart, we update the state based on the current state on the filesystem, including all known valid serial numbers for the infra-nodes, not just the most recent one.  This helps at revocation time to ensure the infra-crl is updated with all possible certs for the host being revoked.

jonathannewman:
Updated to address the two issues mentioned.

src/clj/puppetlabs/puppetserver/certificate_authority.clj
justinstoller: Thanks for that.

justinstoller (approved):

PR #2785, opened 2023-10-23
(PUP-11973) Add certname as header for help debugging

tvpartytonight:


tvpartytonight:
This will work in my use case of using `puppet agent ...` to get the `X-Puppet-Compiler-Name` info, mainly just looking for feedback on if this seems generally the right approach or not.

tvpartytonight:
This PR is up to address the ask in PUP-11900; instead of sending the server name as part of the http body, this sends it as header for v3 catalog requests only. I have tested it out locally with the following results shown with `http_debug` on:
```
% bundle exec puppet agent -t --server localhost --http_debug
opening connection to localhost:8140...
opened
starting SSL for localhost:8140...
SSL established, protocol: TLSv1.3, cipher: TLS_AES_128_GCM_SHA256
<- "GET /puppet/v3/file_metadatas/plugins?...
-> "HTTP/1.1 200 OK\r\n"
-> "Date: Wed, 25 Oct 2023 15:49:00 GMT\r\n"
-> "Content-Type: application/json;charset=utf-8\r\n"
-> "X-Puppet-Version: 8.4.0\r\n"
-> "Vary: Accept-Encoding, User-Agent\r\n"
-> "Content-Encoding: gzip\r\n"
-> "Content-Length: 205\r\n"
-> "\r\n"
reading 205 bytes...
[truncated]
<- "POST /puppet/v3/catalog/localhost?environment=production...
-> "HTTP/1.1 200 OK\r\n"
-> "Date: Wed, 25 Oct 2023 15:49:06 GMT\r\n"
-> "Content-Type: application/vnd.puppet.rich+json; charset=utf-8\r\n"
-> "X-Puppet-Version: 8.4.0\r\n"
-> "X-Puppet-Compiler-Name: localhost\r\n"
-> "Vary: Accept-Encoding, User-Agent\r\n"
-> "Content-Encoding: gzip\r\n"
-> "Content-Length: 311\r\n"
-> "\r\n"
```
The above snipped output of a puppet agent run shows that the `X-Puppet-Compiler-Name` is only sent for the catalog request and not other requests. @Sharpie is this change in puppetserver sufficient for your ask in PUP-11900, or would you like some additional changes on the puppet side to emit a notice if the agent receives that header?

Additionally, is it ok to send the header with the compiler name? I am not sure if it adding this unnecessarily increases vectors for attacks.

tvpartytonight:
Are there other endpoints that this would be good for? I was thinking that we may want to do this for routes that end up calling the `compile-fn`, but I wasn't sure if that was a use case for this or not. If not, makes sense to plumb it down to the specific function it is for.

justinstoller (approved):

PR #2790, opened 2023-11-15
(PUP-11973) Only send compiler header for v3 catalog requests

tvpartytonight:


justinstoller:
I guess we'll have Tony's WIP commit with us in both branches now....

justinstoller (approved):

PR #2792, opened 2023-11-29
(PE-35512) update deps and bootstrap to tkw-j10

steveax:
This commit updates the dependencies and service declarations for trapperkeeper-webserver to the jetty 10 version.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -54,10 +54,10 @@
                  [puppetlabs/jruby-utils]
                  [puppetlabs/clj-shell-utils]
                  [puppetlabs/trapperkeeper]
-                 [puppetlabs/trapperkeeper-webserver-jetty9]
+                 [com.puppetlabs/trapperkeeper-webserver-jetty10]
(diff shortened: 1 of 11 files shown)


project.clj
```diff
@@ -54,7 +54,7 @@
                  [puppetlabs/jruby-utils]
                  [puppetlabs/clj-shell-utils]
                  [puppetlabs/trapperkeeper]
-                 [puppetlabs/trapperkeeper-webserver-jetty9]
+                 [com.puppetlabs/trapperkeeper-webserver-jetty10]
                  [puppetlabs/trapperkeeper-authorization]
                  [puppetlabs/trapperkeeper-comidi-metrics]
                  [puppetlabs/trapperkeeper-metrics]
```
jonathannewman, line 60: This will need to be pinned to "2.0.0" until we get it updated in clj-parent.

steveax: updated

jonathannewman (approved):

PR #2793, opened 2023-11-29
(maint) pin metrics in test space

jonathannewman:
metrics 2.0.0 is needed for the dev/test space as well as the default dependencies, so this pins the version to that.


steveax (approved):
👍

PR #2794, opened 2023-12-06
(main) pin metrics to 2.0.1

jonathannewman:
This adds a pin for trapper keeper metrics to 2.0.1 as clj-parent has an old version that uses jetty9.


steveax (approved):
👍

PR #2800, opened 2023-12-19
(maint) update dependencies to use vectors and maps exclusively

jonathannewman:
Per https://codeberg.org/leiningen/leiningen/src/branch/main/doc/PROFILES.md#user-content-composite-profiles best practice is to not mix vector and map dependencies in profiles, so this modifies the existing project to use exclusively one or the other for profiles.


steveax (approved):
👍

PR #2802, opened 2023-12-21
(PE-37350) ensure bulk signing endpoints enforce authentication gates

jonathannewman:
This adds the /v1/sign and the /v1/sign/all endpoints to the auth.conf so that they will use the authentication gating for `pp_cli_auth`.


steveax (approved):
👍

PR #2804, opened 2023-12-27
(SERVER-3263) correct dropsonde puppet confdir

h0tw1r3:


jonathannewman:
For reference:  https://github.com/puppetlabs/dropsonde/blob/2f9abfcccd7982c17e8baf3e248c944f71f0d562/lib/dropsonde.rb#L20

jonathannewman (changes requested):
Thanks for the fix @h0tw1r3 .  Could you open this against 7.x instead of main, and it will get merged up from there.  Also could you please add a commit message that references the corresponding line in the dropsonde gem (I added it in a previous comment) to help the future archeologists?  Thanks.

jonathannewman (approved):

PR #2805, opened 2024-01-02
(PE-34953) add sign/all endpoint

jonathannewman:
This adds the CA "sign/all" endpoint that signs any CSRs that are found in the directory, and comply with the rules for CSR signing.

It leverages the `sign-multiple-certificate-signing-requests!` which has heavy testing in itself, so a basic smoke test is provided.


steveax (approved):
👍

steveax:
Note: this is ticketed as [PE-37437](https://perforce.atlassian.net/browse/PE-37437)

PR #2808, opened 2024-01-04
(SERVER-3264) Add stdlib gems that JRuby now installs as part of their build process

justinstoller:


justinstoller:
I made this list of gems and their versions by re-installing jruby 9.4.3.0 (the version we ship in latest) and those gems shipped in 2023.5.

Before this is merged I need to manually validate it with a FOSS build (though the gems between FOSS and PE should be consistent in this regard)

We will need to audit this list whenever we upgrade JRuby. These gems were missed because JRuby changed when and how they vendor their included gems and more may be moved around their build process in minor versions.

justinstoller:
I validated this against list against the gems installed in Puppet Server 8.3.0.

justinstoller:
I'm going to leave the DNM label on as we've decided to wait until the next release (we're already passed stop ship for the upcoming FOSS release). This should be merged in a week.

donoghuc (approved):

PR #2816, opened 2024-01-22
(maint) add license to project.clj

jonathannewman:
As part of a recent change in clojars, it is necessary for the license to be part of the exported pom file. Puppetserver has been under the apache-2 license for ~12 years, so this isn't anything new. This adds the declaration to the project.clj so that leiningen will correctly add the license to the exported pom file.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -25,6 +25,9 @@
 (defproject puppetlabs/puppetserver ps-version
   :description "Puppet Server"
 
+  :license {:name "Apache License, Version 2.0"
+              :url "http://www.apache.org/licenses/LICENSE-2.0.html"}
+
   :min-lein-version "2.9.1"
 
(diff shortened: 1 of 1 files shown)


project.clj
```diff
@@ -126,13 +129,13 @@
              :dev-deps  {:dependencies [[org.bouncycastle/bcpkix-jdk18on]]}
              :dev [:defaults :dev-deps]
              :fips-deps {:dependencies [[org.bouncycastle/bcpkix-fips]
-                                       [org.bouncycastle/bc-fips]
-                                       [org.bouncycastle/bctls-fips]]
+                                        [org.bouncycastle/bc-fips]
+                                        [org.bouncycastle/bctls-fips]]
```
jonathannewman, line 133: These, and the ones below were out of alignment according to the editor, so I fixed it.

steveax (approved):
👍

PR #2821, opened 2024-02-09
(PE-37634) ensure sign-all with empty list succeeds

jonathannewman:
Prior to this change, if the sign-all behavior was invoked for the CA service, an exception would be thrown because the function returns a result that wasn't expected. This adds a test that demonstrates the failure, and adds a fix to get the test to pass.


steveax (approved):
👍

steveax:
Eastwood linting error is the lein v2.11 bug, under 2.9.10:
== Linting done in 52452 ms ==
== Warnings: 0. Exceptions thrown: 0

PR #2824, opened 2024-02-16
(PE-37145) ensure schema failures are verbosely logged

jonathannewman:
Prior to this change, a failure in schema checks would provide minimal useful context in the logs around what the schema error was. This adds specific handling for schema failures to log them more verbosely as well as forming a json error payload that contains the schema failures.

Prior to this change, the default handler for jetty was generating a generic error page with the schema failure.  Given that the consumers of this endpoint are not often webbrowsers, this form of content was deemed less useful than a json response with the details.


steveax (approved):
👍

PR #2826, opened 2024-02-20
(maint) remove metrics dependency in self-referential inclusion

jonathannewman:
The metrics version being included in the `ezbake->dependencies` was
pulling in a older version of metrics. The version of metrics is now
managed at the clj-parent level, so this removes the pin.


justinstoller (approved):

steveax (commented):
👍

PR #2827, opened 2024-02-22
(SERVER-3343) Prepare 7.16.0 release

justinstoller:


justinstoller:
NB: HEAD of 7.x has passed all our testing required for release, I just realized that we weren't using the agreed version number.

PR #2829, opened 2024-02-23
Remove facter.jar from classpath

joshcooper:
Facter 4 was introduced in Puppet 7, so the facter.jar from Facter 3 is no longer needed.


joshcooper:
Looks like this is needed in a few other repos too: https://github.com/search?q=org%3Apuppetlabs%20facter.jar&type=code

PR #2830, opened 2024-02-28
(PE-37583) add action tracking routine

jonathannewman:
This adds a routine, that defaults to a function that does nothing that will be used for action tracking within the CA.


steveax (approved):
👍

PR #2832, opened 2024-02-29
(PE-37585) add action tracking to certificate signing behaviors

jonathannewman:
This adds action tracking to the certificate signing behaviors for both the "single sign" behavior and "bulk signing".  Renewals are intentionally ignored.

Tests are added to demonstrate the behaviors.


steveax (approved):
👍

PR #2833, opened 2024-02-29
(PE-37586) add revocation action tracking to revocation routine 

jonathannewman:
This adds action tracking to the certificate revocation routine.
The list of names comes in from input. The routine will throw
and exception if the certnames aren't found in the inventory
file or on disk, so it is reasonable to report removal in
any regard. This could mean that a host that is already removed
will be reported removed again, but this isn't a big issue.

Tests were added to demonstrate the behaviors.


steveax (approved):
👍

PR #2836, opened 2024-03-11
(PE-37184) record catalog compilation actions for hosts

jonathannewman:
This adds catalog compilation action recording for the v3 and v4 endpoints through the `common/record-action` function.

A test was added to demonstrate the behavior of the v3 endpoint.  There are no existing integration tests for the v4 endpoint, and adding them was beyond the scope of this work.


steveax (changes requested):
kondo is carping about `duplicate require of puppetlabs.puppetserver.common`, otherwise, LGTM

PR #2839, opened 2024-03-16
(maint) drop absolute symlinks pointing to tmpdirs

jcharaoui:
These symlinks were most likely mistakenly added as part of 3d275466ac20bf991f6a129dd5ef4e261811c00a, and serve no purpose that I could find.

They were flagged by Debian's [Lintian](https://wiki.debian.org/Lintian) tool in the distro's own [puppetserver](https://tracker.debian.org/pkg/puppetserver) package.


jonathannewman (approved):
Yes, I'm sure those were unintentional additions.  Thank you!

PR #2846, opened 2024-04-04
(maint) Cleanup tests

justinstoller:
Replace assorted broken windows


justinstoller:
Okay, I believe with these changes, this clj-parent change https://github.com/puppetlabs/clj-parent/pull/466 (which should wait until the FOSS releases go out), and the new JRuby https://github.com/puppetlabs/jruby-deps/pull/52 we eliminate all the spurious noise during test runs that we can.

donoghuc (approved):
Verrry nice. Thanks for doing this!

PR #2854, opened 2024-05-23
(SERVER-3051) Pass cwd through to ShellUtils

joshcooper:
If a `cwd` option is passed in Puppet::Util::Execution.execute, then pass it through to ShellUtils.executeCommand.


joshcooper:
@jonathannewman could you review?

jonathannewman (approved):

PR #2855, opened 2024-05-29
(maint) Allow not applying settings catalog on startup

justinstoller:
On initialization of a JRuby instance, Puppet will compile and
apply a catalog containing resources for the configured directories
specified in the appropriate settings sections. The goal of this is
to ensure any directories Puppet requires exist. In practice this
should be managed by our packages. Attempting to apply a catalog every
time a JRuby instance is unnecessary, slow, and causes race conditions
when attempting to instantiate JRuby instances in parallel.

Puppet has a setting "settings_catalog" that controls this behavior.
This PR addresses issues that arise when we set "settings_catalog" to
`false`:
 1. In code the process of compiling a settings catalog brings most of
    Puppet's ruby code into scope. When not compiling a settings catalog
    we must actually declare our dependencies.
 2. The $yamldir/facts directory is not created by packages
    despite being used by default.


src/ruby/puppetserver-lib/puppet/server/puppet_config.rb
```diff
@@ -88,6 +95,8 @@ def self.initialize_puppet(puppet_config:)
       Puppet.push_context(dummy_ssl_context)
     end
 
+    # We have no added support for setting "settings_catalog" to false in the puppet.conf.
```
justinstoller, line 98:
```suggestion
    # We have now added support for setting "settings_catalog" to false in the puppet.conf.
```

justinstoller:
Built a package based on main with this PR in `8.6.1.SNAPSHOT.2024.06.04T0845` and successfully ran the Puppet Server smoke tests and full acceptance (based on Puppet's acceptance tests that contain the `server` tag).

jonathannewman (approved):

mcdonaldseanp (approved):

PR #2867, opened 2024-07-16
(PE-34416) Load the FIPS compatible, Puppet provided system cacert keystore if in FIPS

justinstoller:


justinstoller:
I still need to do a manual run smoke test with the agent installed on a FIPS host.

PR #2868, opened 2024-07-16
(PE-38293) Fix test for modified error message with new Bouncy Castle

nmburgan:


justinstoller:
closing and re-opening to try and kick this against the recent clj-parent bump

justinstoller:
@nmburgan , looks like we need to make the final dashes in the exception message optional. Doesn't look like the FIPS jars have been updated with the improved error message yet.

justinstoller (changes requested):
Message change hasn't happened in FIPS yet, assertion needs to handle both, or do separate assertions for FIPS vs non-FIPS.

justinstoller (approved):

PR #2875, opened 2024-09-03
(PE-38948) Respect rich_data in compileCatalog

justinstoller:
Previously, when using the compileCatalog method of
Puppet::Server::Master compilation would fail to serialize rich data.
This occurred because during serialization we introspect the
Puppet.context for the rich_data value (the setting itself defaults to
true & and strict now defaults to error, throwing an error when we
encounter an object that requires rich_data to serialize).

Normal catalog compilation works because the value of rich_data in the
Puppet.context is overridden by the caller of the serialization code for
indirected routes, using whatever is set in the environment and if unset
in the environment then falling back to the actually rich_data setting.

In the compileCatalog method however we do not use the indirected routes
for serialization, instead calling `to_data_hash` on the catalog and
serializing that ourselves (we explicitly only support JSON via that
endpoint). This means the value in the Puppet.context is the default
value which at the present is `false` regardless of what the user has
configured (or our own settings defaults).

This commit attempts to fix this by overriding the Puppet.context's
rich_data value with the Puppet setting for rich_data. Which now
defaults to true and will honor whatever the user puts in their
puppet.conf. This however does not support rich_data per environment. I
believe the optimal point to override rich_data is at the overall
compileCatalog level but we do not have the "current" environment at
that point, neccessitating a much larger rewrite of the compileCatalog
function to work. The fact that you can even set rich_data in the
environment.conf at this time is an undocumented feature, which leads me
to believe a larger rewrite of the compileCatalog function is not
a good investment.

### DIFF
--- a/src/ruby/puppetserver-lib/puppet/server/master.rb
+++ b/src/ruby/puppetserver-lib/puppet/server/master.rb
@@ -97,7 +97,19 @@ class Puppet::Server::Master
   end
 
   def compileCatalog(request_data)
-    Puppet.override(lookup_key_recorder: create_recorder) do
+    Puppet.override(
+      lookup_key_recorder: create_recorder,
+      # rich_data was moved to the context because determining if it should be true, in
+      # some code paths, needs to inspect the global settings, per-environment settings,
+      # and the request's content-type. For the `handleRequest` endpoint this logic is
+      # handled by the indirector. We need to provide some value, as the default in
+      # Puppet's context is, as of Puppet 8.9.0, hardcoded to be false, which will cause
+      # any rich_data in a catalog to fail via this endpoint (but likely not via the
+      # indirected endpoint). So here we set it to the global setting value, as the
+      # per-environment settings logic is fairly complicated to implement and have never
+      # gotten user facing documentation anyways.
+      rich_data: Puppet[:rich_data]
+    ) do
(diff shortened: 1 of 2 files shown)


justinstoller:
Just for show right now.

justinstoller:
The test failures seem to be a product of a newer JVM 17 verses the main branch. Here are the same failures without the code changes https://github.com/puppetlabs/puppetserver/pull/2876, which I'm looking into but shouldn't' block this.

src/ruby/puppetserver-lib/puppet/server/master.rb
```diff
@@ -97,7 +97,10 @@ def handleRequest(request)
   end
 
   def compileCatalog(request_data)
-    Puppet.override(lookup_key_recorder: create_recorder) do
+    Puppet.override(
+      lookup_key_recorder: create_recorder,
+      rich_data: Puppet[:rich_data]
```
mcdonaldseanp, line 102: Should probably leave a comment here explaining why this works in puppet but here we have to explicitly set it here. If I looked at just the code it would boggle my mind as to why we are calling the override function and giving it Puppet's value.

justinstoller: updated

mcdonaldseanp (approved):

PR #2878, opened 2024-09-13
compiler.rb: Switch to modern facts

bastelfreak:
All of those are available at least since Facter 3.


ekohl (commented):
I'd also take the chance to look at a new structured fact called `puppetserver` and add all the values in there.

You can probably be smart and align the facts to networking facts:

```ruby
# Copy networking facts if available
['fqdn', 'ip', 'ip6'].each do |var|
  if (value = Facter.value("networking.#{var}"))
    @server_facts[var] = value
  end
end
```

justinstoller (approved):

justinstoller:
Closing in favor of https://github.com/puppetlabs/puppetserver/pull/2894

PR #2880, opened 2024-09-23
(PE-38408) Remove expensive Regexes from puppet profiler Java impl

justinstoller:
We have triggered a nasty performance regression in the JVM with some of our Regex usage in the MetricsPuppetProfiler. Regardless of the performance regression, our Regex usage is unnecessary and unneedlessly expensive.

### DIFF
--- a/src/java/com/puppetlabs/puppetserver/MetricsPuppetProfiler.java
+++ b/src/java/com/puppetlabs/puppetserver/MetricsPuppetProfiler.java
@@ -5,6 +5,7 @@ import com.codahale.metrics.Timer;
 import org.apache.commons.lang.StringUtils;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -21,17 +22,21 @@ public class MetricsPuppetProfiler implements PuppetProfiler {
     private final MetricRegistry registry;
     private final Set<String> metric_ids;
 
-    private static final Pattern FUNCTION_PATTERN = Pattern.compile(".*\\.functions\\.([\\w\\d_]+)$");
-    private static final Pattern RESOURCE_PATTERN = Pattern.compile(".*\\.compiler\\.evaluate_resource\\.([\\w\\d_]+\\[([\\w\\d_]+::)*[\\w\\d_]+\\])$");
-    private static final Pattern CATALOG_PATTERN = Pattern.compile(".*\\.compiler\\.(static_compile_postprocessing|static_compile|compile|find_node)$");
-    private static final Pattern INLINING_PATTERN = Pattern.compile(".*\\.compiler\\.static_compile_inlining\\.(.*)$");
-    private static final Pattern PUPPETDB_PATTERN = Pattern.compile(".*\\.puppetdb\\.(resource\\.search|facts\\.encode|command\\.submit\\.replace facts|catalog\\.munge|command\\.submit\\.replace catalog|report\\.convert_to_wire_format_hash|command\\.submit\\.store report|query)$");
-
+    private final Map<String, Timer> function_timers;
+    private final Map<String, Timer> resource_timers;
+    private final Map<String, Timer> catalog_timers;
+    private final Map<String, Timer> inlining_timers;
+    private final Map<String, Timer> puppetdb_timers;
 
     public MetricsPuppetProfiler(String hostname, MetricRegistry registry) {
         this.hostname = hostname;
         this.registry = registry;
         this.metric_ids = Collections.newSetFromMap(new ConcurrentHashMap<String, Boolean>());
+        this.function_timers = new ConcurrentHashMap<String, Timer>();
+        this.resource_timers = new ConcurrentHashMap<String, Timer>();
+        this.catalog_timers = new ConcurrentHashMap<String, Timer>();
+        this.inlining_timers = new ConcurrentHashMap<String, Timer>();
+        this.puppetdb_timers = new ConcurrentHashMap<String, Timer>();
     }
 
     @Override
@@ -43,9 +48,12 @@ public class MetricsPuppetProfiler implements PuppetProfiler {
     public void finish(Object context, String message, String[] metric_id) {
         if (shouldTime(metric_id)) {
           Long elapsed = System.currentTimeMillis() - (Long)context;
-            for (Timer t : getTimers(metric_id)) {
+	    Map<String, Timer> metricsByID = getOrCreateTimersByIDs(metric_id);
+            for (Timer t : metricsByID.values()) {
                 t.update(elapsed, TimeUnit.MILLISECONDS);
             }
+
+	    updateMetricsTrackers(metric_id, metricsByID);
         }
     }
 
@@ -54,27 +62,103 @@ public class MetricsPuppetProfiler implements PuppetProfiler {
     }
 
     public Map<String, Timer> getFunctionTimers() {
-        return getTimers(FUNCTION_PATTERN);
+        return this.function_timers;
     }
 
     public Map<String, Timer> getResourceTimers() {
-        return getTimers(RESOURCE_PATTERN);
+        return this.resource_timers;
     }
 
     public Map<String, Timer> getCatalogTimers() {
-        return getTimers(CATALOG_PATTERN);
+        return this.catalog_timers;
     }
 
     public Map<String, Timer> getInliningTimers() {
-        return getTimers(INLINING_PATTERN);
+        return this.inlining_timers;
     }
 
     public Map<String, Timer> getPuppetDBTimers() {
-        return getTimers(PUPPETDB_PATTERN);
+        return this.puppetdb_timers;
     }
 
     @Override
     public void shutdown() {
+    }
+
+    private List<String> sliceOfArrayToList(String[] idSegments, int lengthOfID) {
+        // Callers expect a mutable List returned, but Arrays.asList() returns a
+	// fix length array, which is why we have to create a List and then add to it.
+        List<String> idList = new ArrayList<String>();
+        idList.addAll(Arrays.asList(Arrays.copyOf(idSegments, lengthOfID)));
+
+	return idList;
+    }
+
+    private String safeGet(String[] collection, int i) {
+        try {
+            return collection[i];
+	}  catch (IndexOutOfBoundsException _ex) {
+            return "";
+	}
+    }
+
+    private void updateMetricsTrackers(String[] metricId, Map<String, Timer> metricsByID) {
+	String firstElement = safeGet(metricId, 0);
+	String secondElement = safeGet(metricId, 1);
+
+        if ("functions".equals(firstElement)) {
+            Timer metric = metricsByID.get(getMetricName(sliceOfArrayToList(metricId, 2)));
+            this.function_timers.put(secondElement, metric);
+
+        } else if ("compiler".equals(firstElement)) {
+            String thirdElemet = safeGet(metricId, 2);
+
+            if ("evaluate_resource".equals(secondElement)) {
+                Timer metric = metricsByID.get(getMetricName(sliceOfArrayToList(metricId, 3)));
+                this.resource_timers.put(thirdElemet, metric);
+
+            } else if ("static_compile_inlining".equals(secondElement)) {
+                Timer metric = metricsByID.get(getMetricName(sliceOfArrayToList(metricId, 3)));
+                this.inlining_timers.put(thirdElemet, metric);
+
+            } else {
+                Timer metric = metricsByID.get(getMetricName(sliceOfArrayToList(metricId, 2)));
+                this.catalog_timers.put(secondElement, metric);
+            }
+
+        } else if ("puppetdb".equals(firstElement)) {
+            if ("query".equals(secondElement)) {
+                Timer metric = metricsByID.get(getMetricName(sliceOfArrayToList(metricId, 2)));
+                this.puppetdb_timers.put(secondElement, metric);
+
+	    } else {
+		String thirdElemet = safeGet(metricId, 2);
+
+                if (
+                    ("resource".equals(secondElement) && "search".equals(thirdElemet)) ||
+		    ("facts".equals(secondElement) && "encode".equals(thirdElemet)) ||
+		    ("catalog".equals(secondElement) && "munge".equals(thirdElemet)) ||
+		    ("report".equals(secondElement) && "convert_to_wire_format_hash".equals(thirdElemet))
+		) {
+                    String key = String.join(".", secondElement, thirdElemet);
+                    Timer metric = metricsByID.get(getMetricName(sliceOfArrayToList(metricId, 3)));
+                    this.puppetdb_timers.put(key, metric);
+
+                } else if ("command".equals(secondElement) && "submit".equals(thirdElemet)) {
+                    String fourthElement = safeGet(metricId, 3);
+
+                    if (
+                        "store report".equals(fourthElement) ||
+			"replace facts".equals(fourthElement) ||
+			"replace catalog".equals(fourthElement)
+                    ) {
+                        String key = String.join(".", secondElement, thirdElemet, fourthElement);
+                        Timer metric = metricsByID.get(getMetricName(sliceOfArrayToList(metricId, 4)));
+                        this.puppetdb_timers.put(key, metric);
+		    }
+		}
+	    }
+        }
     }
 
     private boolean shouldTime(String[] metric_id) {
@@ -90,8 +174,8 @@ public class MetricsPuppetProfiler implements PuppetProfiler {
         return true;
     }
 
-    private List<Timer> getTimers(String[] metric_id) {
-        List<Timer> timers = new ArrayList<Timer>();
+    private Map<String, Timer> getOrCreateTimersByIDs(String[] metric_id) {
+        Map<String, Timer> timers = new HashMap<String, Timer>();
         // If this is turns out to be a performance hit, we could cache these in a
         // map or something.
         for (int i = 0; i < metric_id.length; i++) {
@@ -101,7 +185,7 @@ public class MetricsPuppetProfiler implements PuppetProfiler {
             }
             String metric_name = getMetricName(current_id);
             registerMetricName(metric_name);
-            timers.add(registry.timer(metric_name));
+            timers.put(metric_name, registry.timer(metric_name));
         }
         return timers;
     }
@@ -113,16 +197,5 @@ public class MetricsPuppetProfiler implements PuppetProfiler {
 
     private void registerMetricName(String metric_name) {
         this.metric_ids.add(metric_name);
-    }
-
-    private Map<String, Timer> getTimers(Pattern pattern) {
-        Map<String, Timer> rv = new HashMap<>();
-        for (String metric_id : this.metric_ids) {
-            Matcher matcher = pattern.matcher(metric_id);
-            if (matcher.matches()) {
-                rv.put(matcher.group(1), registry.timer(metric_id));
-            }
-        }
-        return rv;
     }
 }
--- a/src/java/com/puppetlabs/puppetserver/MetricsPuppetProfiler.java
+++ b/src/java/com/puppetlabs/puppetserver/MetricsPuppetProfiler.java
@@ -136,9 +136,11 @@ public class MetricsPuppetProfiler implements PuppetProfiler {
 
                 if (
                     ("resource".equals(secondElement) && "search".equals(thirdElemet)) ||
-		    ("facts".equals(secondElement) && "encode".equals(thirdElemet)) ||
-		    ("catalog".equals(secondElement) && "munge".equals(thirdElemet)) ||
-		    ("report".equals(secondElement) && "convert_to_wire_format_hash".equals(thirdElemet))
+		    ("payload".equals(secondElement) && "format".equals(thirdElemet)) ||
+		    // Set.of would be preferrable but 7.x still support Java 8, which does not have Set.of
+		    ("facts".equals(secondElement) && Arrays.asList("save", "find", "search", "encode").contains(thirdElemet)) ||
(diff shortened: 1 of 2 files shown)


src/java/com/puppetlabs/puppetserver/MetricsPuppetProfiler.java
```diff
@@ -54,29 +62,59 @@ public Set<String> getAllMetricIds() {
...
+                Timer metric = metricsByID.get(getMetricName(sliceOfArrayToList(metricId, 3)));
+                this.resource_timers.put(metricId[2], metric);
+            } else if ("static_compile_inlining".equals(metricId[1])) {
+                Timer metric = metricsByID.get(getMetricName(sliceOfArrayToList(metricId, 3)));
+                this.inlining_timers.put(metricId[2], metric);
+            } else {
+                Timer metric = metricsByID.get(getMetricName(sliceOfArrayToList(metricId, 2)));
+                this.catalog_timers.put(metricId[1], metric);
+            }
+        } else if ("puppetdb".equals(metricId[0])) {
```
justinstoller, line 112: This puppetdb clause does not contain all the necessary complexity of the former PUPPETDB_PATTERN

src/java/com/puppetlabs/puppetserver/MetricsPuppetProfiler.java
```diff
@@ -54,29 +62,59 @@ public Set<String> getAllMetricIds() {
...
+        // Callers expect a mutable List returned, but Arrays.asList() returns a
+	// fix length array, which is why we have to create a List and then add to it.
+        List<String> idList = new ArrayList<String>();
+        idList.addAll(Arrays.asList(Arrays.copyOf(idSegments, lengthOfID)));
+
+	return idList;
+    }
+
+    private void updateMetricsTrackers(String[] metricId, Map<String, Timer> metricsByID) {
+        if ("functions".equals(metricId[0])) {
```
donoghuc, line 98: Just a note on how to start reviewing this because it was hard to identify for me... We need to look for areas like https://github.com/puppetlabs/puppet/blob/1fba2505e6fd5c7ed17257c07667da98d6ed1dfb/lib/puppet/pops/functions/function.rb#L115 and evaluate if all the cases are handled correctly. Using array indices for each of these fields requires that we have a good sense of the possible arrays that are coming in. 

I think there will be some in puppet and puppetdb. Not sure if there are any in puppetserver as well? 

Overall, i like the idea of doing the handling when an event comes in such that there is no cost for querying :).

src/java/com/puppetlabs/puppetserver/MetricsPuppetProfiler.java
justinstoller: These are the root metrics we report on based the PUPPETDB_PATTERN:

```
puppetdb.catalog.munge
puppetdb.command.submit.replace facts
puppetdb.command.submit.replace catalog
puppetdb.command.submit.store report
puppetdb.facts.encode
puppetdb.query
puppetdb.resource.search
puppetdb.report.convert_to_wire_format_hash
```

And, going through the PuppetDB Ruby code (at this commit https://github.com/puppetlabs/puppetdb/commit/8a1e2efe2c608c11453fae689a45150c0154aea1) I see these metrics created:
```
[:puppetdb, :aliases, :map_to_title]
[:puppetdb, :catalog, :munge]
[:puppetdb, :catalog, :save, request.key]
[:puppetdb, :command, :submit]
[:puppetdb, :command, :submit, catalog_command_name, catalog_version]
[:puppetdb, :command, :submit, command_name, version]
[:puppetdb, :command, :submit, inputs_command_name, inputs_version]
[:puppetdb, :edges, :munge]
[:puppetdb, :edges, :synthesize]
[:puppetdb, :edges, :synthesize, :make_unique]
[:puppetdb, :edges, :synthesize, :primary_synthesis]
[:puppetdb, :edges, :synthesize, :resource_table, :build]
[:puppetdb, :events_list, :build]
[:puppetdb, :facts, :encode]
[:puppetdb, :facts, :find, :parse_response, request.key]
[:puppetdb, :facts, :find, :query_nodes, request.key]
[:puppetdb, :facts, :find, request.key]
[:puppetdb, :facts, :save, request.key]
[:puppetdb, :facts, :search, :parse_query_response, request.key,]
[:puppetdb, :facts, :search, :query_request, request.key]
[:puppetdb, :facts, :search, request.key]
[:puppetdb, :keys, :filter_extraneous]
[:puppetdb, :logs_list, :build]
[:puppetdb, :metaparams, :sort]
[:puppetdb, :metrics_list, :build]
[:puppetdb, :namevar_aliases, :add]
[:puppetdb, :parameters, :add_missing]
[:puppetdb, :payload, :format]
[:puppetdb, :query, query]
[:puppetdb, :report, :convert_to_wire_format_hash]
[:puppetdb, :report, :process]
[:puppetdb, :resource, :search, :build_up_collected_objects, request.key]
[:puppetdb, :resource, :search, :parse_query_response, request.key]
[:puppetdb, :resource, :search, :query, request.key]
[:puppetdb, :resource, :search, request.key]
[:puppetdb, :resources_list, :build]
[:puppetdb, :titles, :stringify]
```

Are the current metrics correct? We could add a few while we're here. Some easier ones to add would be:
```
puppetdb.catalog.save
puppetdb.edges.munge
puppetdb.edges.synthesize
puppetdb.facts.find
puppetdb.facts.save
puppetdb.facts.search
puppetdb.payload.format
puppetdb.report.process
```
But I don't know if they duplicate some existing reported metrics, or aren't very valuable.

Thoughts @austb, @rbrw , @Sharpie ?

src/java/com/puppetlabs/puppetserver/MetricsPuppetProfiler.java
```diff
@@ -54,29 +62,59 @@ public Set<String> getAllMetricIds() {
...
+            Timer metric = metricsByID.get(getMetricName(sliceOfArrayToList(metricId, 2)));
+            this.function_timers.put(metricId[1], metric);
+        } else if ("compiler".equals(metricId[0])) {
```
justinstoller, line 101: Should add null checking to metricId comparisons

src/java/com/puppetlabs/puppetserver/MetricsPuppetProfiler.java
```diff
@@ -54,29 +62,59 @@ public Set<String> getAllMetricIds() {
...
+            if ("evaluate_resource".equals(metricId[1])) {
+                Timer metric = metricsByID.get(getMetricName(sliceOfArrayToList(metricId, 3)));
+                this.resource_timers.put(metricId[2], metric);
+            } else if ("static_compile_inlining".equals(metricId[1])) {
+                Timer metric = metricsByID.get(getMetricName(sliceOfArrayToList(metricId, 3)));
+                this.inlining_timers.put(metricId[2], metric);
+            } else {
```
justinstoller, line 108: The previous regex based selection explicitly grabbed these submetrics: static_compile_postprocessing, static_compile, compile, and find_node. This else will grab all metrics that aren't evaluate_resource or static_compile_inlining, don't know if that is a good thing or a bad thing. Thoughts @Sharpie ?

justinstoller:
I _think_ I've addressed the logic issues with the existing puppetdb metrics. Looking forward to hear from Charlie, et al if others would also be valuable. I need to add some tests, it appears that we test the function, catalog, and resource metrics from the profiler interface, but test the puppetdb metrics only via the graphite exporter (unaffected by this PR).

src/java/com/puppetlabs/puppetserver/MetricsPuppetProfiler.java
justinstoller: @Sharpie , how does adding those new puppetdb metrics sound?

justinstoller:
The unit/integration tests for puppetdb metrics I was thinking of are testing allow listing, not actual creation of those metrics (which is hard since we don't have a PDB during our unit/integration tests.

However there is an acceptance test to validate these metrics, and that's been helpful locally to make sure these are formatted correctly. The acceptance test is passing with a package built prior a small refactor (puppetserver 7.17.3.SNAPSHOT.2024.09.26T1533 on builds).

I'll add those metrics Austin mentioned (unless I hear otherwise from Charlie) and then build a final package to acceptance test and that others can validate if they wish.

src/java/com/puppetlabs/puppetserver/MetricsPuppetProfiler.java
justinstoller: Here's the updated metrics:
```
      "puppetdb-metrics": [
        {                                         
          "metric": "catalog_save",
          "count": 2,
          "mean": 243,   
          "aggregate": 486
        },
        {                            
          "metric": "resource_search",
          "count": 8,
          "mean": 46,    
          "aggregate": 368
        },
        {                                                                                                
          "metric": "query",
          "count": 1,
          "mean": 355,  
          "aggregate": 355
        },
        {                          
          "metric": "facts_find",
          "count": 2,
          "mean": 117,  
          "aggregate": 234
        },
        {                 
          "metric": "catalog_munge",
          "count": 2,
          "mean": 94, 
          "aggregate": 188         
        },
        {
          "metric": "facts_save",                                                                        
          "count": 2,
          "mean": 76,
          "aggregate": 152
        },
        {
          "metric": "command_submit_replace_catalog",
          "count": 2,
          "mean": 54,
          "aggregate": 108
        },
        {
          "metric": "command_submit_replace_facts",
          "count": 2,
          "mean": 49,
          "aggregate": 98
        },
        {
          "metric": "report_process",
          "count": 2,
          "mean": 41,
          "aggregate": 82
        },
        {
          "metric": "command_submit_store_report",
          "count": 2,
          "mean": 30,
          "aggregate": 60
        },
        {
          "metric": "payload_format",
          "count": 8,
          "mean": 2,
          "aggregate": 16
        },
        {
          "metric": "report_convert_to_wire_format_hash",
          "count": 2,
          "mean": 4,
          "aggregate": 8
        },
        {
          "metric": "facts_encode",
          "count": 2,
          "mean": 2,
          "aggregate": 4
        }
      ],
      "inline-metrics": []
    }
  },
  "active_alerts": [],
  "service_name": "puppet-profiler"
}
```

justinstoller:
Okay. I think this is ready to go. There's a package internally available with 7.17.3.SNAPSHOT.2024.09.27T1145 that works for me.

I had to update our acceptance tests - it looks like our PDB integration isn't correct. I have a feeling that mean our PDB tests aren't really doing anything in CI, but first we need to get CI working and then I take up updating the PDB integration acceptance tests after that.

src/java/com/puppetlabs/puppetserver/MetricsPuppetProfiler.java
```diff
@@ -54,29 +62,107 @@ public Set<String> getAllMetricIds() {
...
     public Map<String, Timer> getPuppetDBTimers() {
-        return getTimers(PUPPETDB_PATTERN);
+        return this.puppetdb_timers;
     }
 
     @Override
     public void shutdown() {
     }
 
+    private List<String> sliceOfArrayToList(String[] idSegments, int lengthOfID) {
```
donoghuc, line 88: I cant see how `subList` does not work for this? 

Seems like for example the `getMetricName` could be called with the result of a `metricId.subList(0,2)` as the result of `subList` I think would be mutable?

src/java/com/puppetlabs/puppetserver/MetricsPuppetProfiler.java
donoghuc: Does this mean these more specific metrics will no longer be counted in catalog timers? Does that matter?

src/java/com/puppetlabs/puppetserver/MetricsPuppetProfiler.java
```diff
@@ -54,29 +62,107 @@ public Set<String> getAllMetricIds() {
...
+
+        } else if ("puppetdb".equals(firstElement)) {
+            if ("query".equals(secondElement)) {
+                Timer metric = metricsByID.get(getMetricName(sliceOfArrayToList(metricId, 2)));
+                this.puppetdb_timers.put(secondElement, metric);
+
+	    } else {
+		String thirdElemet = safeGet(metricId, 2);
+
+                if (
```
donoghuc, line 137: Its probably overkill, but i wonder if it would be worth comparing metrics from two reference workflows. Matching the regex to this logic is hurting my brain. It may not matter too much though... These are hard to reason about 🫠

src/java/com/puppetlabs/puppetserver/MetricsPuppetProfiler.java
```diff
@@ -90,8 +176,8 @@ private boolean shouldTime(String[] metric_id) {
         return true;
     }
 
-    private List<Timer> getTimers(String[] metric_id) {
-        List<Timer> timers = new ArrayList<Timer>();
+    private Map<String, Timer> getOrCreateTimersByIDs(String[] metric_id) {
```
donoghuc, line 179: This name is not used anywhere else right? Its not important for https://github.com/puppetlabs/puppetserver/blob/ebc41ebdb39690c597147e6a9bbbc66b8283bd97/src/clj/puppetlabs/services/jruby/jruby_metrics_core.clj#L212 ?

donoghuc: I'm not sure i'm following the "or" in the name here. It seems like this just creates a new map and returns it unconditionally. Am I missing something?

donoghuc:
Do you feel strongly about this going in 7.x? Would it make sense just to target main?

src/java/com/puppetlabs/puppetserver/MetricsPuppetProfiler.java
justinstoller: We start with a `String[]` which doesn't have the `subList()` method defined, but we can turn an array into a List with `Arrays.asList()`.

I originally tried `Arrays.asList(ids).subList(0,2)`. But the problem is `asList()` returns a fixed sized List and `subList()` simply returns a view of the original fixed size list. So we end up erring when `getMetricName` attempts to prepend the hostname to the list here: https://github.com/puppetlabs/puppetserver/pull/2880/files#diff-300731d432bcbd4678957c962c2aa20d068698fff7a4441434c3d1ac5e55fdfdR196

I should probably clean up `getMetricName()` but I didn't want to lead with a larger refactor than was strictly necessary (at least after I had changed `getTimers`)

src/java/com/puppetlabs/puppetserver/MetricsPuppetProfiler.java
justinstoller: They will still be there, but we'll report a lot more now. I don't know if that will be helpful or just noise. Here are the ones I can easily find with grep:
```
init_server_facts
find_facts
iterate_on_generators
evaluate_definitions
evaluate_collections
validate_final
finish_catalog
validate_pre_finish
evaluate_generators
evaluate_node_classes
evaluate_ast_node
evaluate_main
create_settings_scope
set_node_params
```

src/java/com/puppetlabs/puppetserver/MetricsPuppetProfiler.java
justinstoller: The regex matching definitely hurt my brain as well! I got it wrong a couple of times. I can run the acceptance test and collect what metrics are reported before and after this change. I've ran it after to make sure they look right (see this comment https://github.com/puppetlabs/puppetserver/pull/2880#discussion_r1778993676) , but haven't done a full comparison. I'll do that.

src/java/com/puppetlabs/puppetserver/MetricsPuppetProfiler.java
justinstoller: > This name is not used anywhere else right? Its not important for 

It's not, the method is private and external callers are calling this method on the MetricRegistry https://www.javadoc.io/doc/io.dropwizard.metrics/metrics-core/4.1.0-rc2/com/codahale/metrics/MetricRegistry.html#getTimers--

I think the two cases of "getTimers()" here where inspired by that method. But, the call to MetricRegister.timer() in this method is actually the place in this class were we _create_ timers (if they don't exist). See https://www.javadoc.io/doc/io.dropwizard.metrics/metrics-core/4.1.0-rc2/com/codahale/metrics/MetricRegistry.html#timer-java.lang.String-

The `getOrCreateTimers` is in reference to that. A more accurate name would probably be `getOrCreateTimersAndReturnThemById` , maybe?

src/java/com/puppetlabs/puppetserver/MetricsPuppetProfiler.java
justinstoller: Here are the previous metrics from 7.17.2:
```
{"service_version":"7.17.2","service_status_version":1,"detail_level":"debug","state":"running","status":{"experimental":{"function-metrics":[{"function":"puppetdb_query","count":1,"mean":369,"aggregate":369},{"function":"lookup","count":2,"mean":45,"aggregate":90}],"resource-metrics":[{"resource":"Class[main]","count":2,"mean":193,"aggregate":386}],"catalog-metrics":[{"metric":"compile","count":2,"mean":886,"aggregate":1772},{"metric":"find_node","count":2,"mean":11,"aggregate":22}],"puppetdb-metrics":[{"metric":"query","count":1,"mean":366,"aggregate":366},{"metric":"resource_search","count":8,"mean":31,"aggregate":248},{"metric":"command_submit_replace_catalog","count":2,"mean":68,"aggregate":136},{"metric":"command_submit_store_report","count":2,"mean":51,"aggregate":102},{"metric":"command_submit_replace_facts","count":2,"mean":43,"aggregate":86},{"metric":"catalog_munge","count":2,"mean":9,"aggregate":18},{"metric":"report_convert_to_wire_format_hash","count":2,"mean":5,"aggregate":10},{"metric":"facts_encode","count":2,"mean":2,"aggregate":4}],"inline-metrics":[]}},"active_alerts":[],"service_name":"puppet-profiler"}
```
and here are the metrics returned after this pr:
```
{"service_version":"7.17.3-SNAPSHOT","service_status_version":1,"detail_level":"debug","state":"running","status":{"experimental":{"function-metrics":[{"function":"puppetdb_query","count":1,"mean":411,"aggregate":411},{"function":"lookup","count":2,"mean":46,"aggregate":92}],"resource-metrics":[{"resource":"Node[weakest-byword.delivery.puppetlabs.net]","count":1,"mean":421,"aggregate":421},{"resource":"Class[main]","count":2,"mean":184,"aggregate":368},{"resource":"Node[resource-exporter.test]","count":1,"mean":208,"aggregate":208}],"catalog-metrics":[{"metric":"compile","count":2,"mean":912,"aggregate":1824},{"metric":"evaluate_ast_node","count":2,"mean":321,"aggregate":642},{"metric":"evaluate_main","count":2,"mean":188,"aggregate":376},{"metric":"evaluate_generators","count":2,"mean":115,"aggregate":230},{"metric":"iterate_on_generators","count":3,"mean":75,"aggregate":225},{"metric":"evaluate_collections","count":2,"mean":108,"aggregate":216},{"metric":"create_settings_scope","count":2,"mean":54,"aggregate":108},{"metric":"set_node_params","count":2,"mean":14,"aggregate":28},{"metric":"find_facts","count":2,"mean":7,"aggregate":14},{"metric":"find_node","count":2,"mean":6,"aggregate":12},{"metric":"finish_catalog","count":2,"mean":1,"aggregate":2},{"metric":"init_server_facts","count":1,"mean":0,"aggregate":0},{"metric":"validate_pre_finish","count":2,"mean":0,"aggregate":0},{"metric":"evaluate_node_classes","count":2,"mean":0,"aggregate":0},{"metric":"validate_final","count":2,"mean":0,"aggregate":0},{"metric":"evaluate_definitions","count":3,"mean":0,"aggregate":0}],"puppetdb-metrics":[{"metric":"query","count":1,"mean":409,"aggregate":409},{"metric":"resource_search","count":8,"mean":36,"aggregate":288},{"metric":"facts_find","count":2,"mean":139,"aggregate":278},{"metric":"catalog_save","count":2,"mean":131,"aggregate":262},{"metric":"facts_save","count":2,"mean":91,"aggregate":182},{"metric":"command_submit_replace_catalog","count":2,"mean":67,"aggregate":134},{"metric":"command_submit_replace_facts","count":2,"mean":56,"aggregate":112},{"metric":"report_process","count":2,"mean":50,"aggregate":100},{"metric":"command_submit_store_report","count":2,"mean":39,"aggregate":78},{"metric":"payload_format","count":8,"mean":3,"aggregate":24},{"metric":"facts_encode","count":2,"mean":10,"aggregate":20},{"metric":"catalog_munge","count":2,"mean":10,"aggregate":20},{"metric":"report_convert_to_wire_format_hash","count":2,"mean":4,"aggregate":8}],"inline-metrics":[]}},"active_alerts":[],"service_name":"puppet-profiler"}
```

These aren't pretty printed, not as an f-u, but because I think putting them pretty printed in a github comment would make the content very hard to compare. I put them into files, piped them through jq, and compared two terminals side by side. I think they look correct.

justinstoller:
I don't have strong feelings. The escalation that started this research was using 7.x and so I started there.

src/java/com/puppetlabs/puppetserver/MetricsPuppetProfiler.java
donoghuc: Ah, i see, i misread that parameter type.

src/java/com/puppetlabs/puppetserver/MetricsPuppetProfiler.java
donoghuc: Ah, thanks for the reference. I think leaving it as is will be fine!

donoghuc:
I think given the main benefit will be in 7.x (even though its close to the end) i think leaving it at that stream is best. Thanks!

donoghuc (approved):

src/java/com/puppetlabs/puppetserver/MetricsPuppetProfiler.java
Sharpie: The updated set of metrics looks good to me 👍

PR #2882, opened 2024-09-25
(PE-39250) Remove project APIs and configuration

mcdonaldseanp:
Removes access to file_metadata and compile options which allowed fetching data and compiling for bolt projects. This commit does not remove all project based code, it is restricted to covering enough ground to remove access to any functionality, but there may still be some internal code left over.


justinstoller:
I assume this can also go: https://github.com/puppetlabs/puppetserver/pull/2882/files#diff-b0f611aa58814be93f011d5e22ed06752df5baf8fd4f2e960d882171a3fb4d13R966 ?

I believe if that goes then this file can go https://github.com/puppetlabs/puppetserver/blob/main/src/clj/puppetlabs/services/master/file_serving.clj

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
```diff
@@ -84,8 +84,8 @@
    (.compileCatalog jruby-instance request-options))
 
   (compile-ast
-    [this jruby-instance compile-options boltlib-path]
-    (.compileAST jruby-instance compile-options boltlib-path))
+    [this jruby-instance compile-options]
+    (.compileAST jruby-instance compile-options))
```
justinstoller, line 88: It looks like this was added as part of apply work, regardless of projects. https://github.com/puppetlabs/puppetserver/commit/e05b75e3635ae7fcbe63d295651952e40a979c3d Is that right?

justinstoller:
It looks like there was some spec updates made as part of https://github.com/puppetlabs/puppetserver/commit/9cb17ad771039baf91e4fc411a1cac12fb7e0269 that still reference projects. I know this PR wasn't meant to get rid of everything, but I'm surprised the specs are passing with this change.

src/clj/puppetlabs/services/jruby/jruby_puppet_service.clj
mcdonaldseanp: UGH, DAMNIT. 

Yeah, you're correct

mcdonaldseanp:
@justinstoller are you referrring to the ruby specs? if so there weren't any significant changes to the ruby code for compiling projects, so the tests will still pass.

Also I have no idea why linting is failing on a file I've not made any changes to

steveax:
latest kondo gets downloaded during the action run, so if there's a new version it'll sometimes carp about something that hasn't been touched

donoghuc:
Those kondo fixes were fixed in 7.x and will be merged up https://github.com/puppetlabs/puppetserver/pull/2880/files

mcdonaldseanp:
Yeah @justinstoller I don't expect that to fail. The ruby AST compiler can still compile in a project, I just removed access to that code through the API

justinstoller:
These are the code deletions I was hoping for. But it looks like there's some additional tests that need to go with it.

justinstoller (approved):

PR #2884, opened 2024-10-01
(maint) Fix pdb integration acceptance test

justinstoller:
This was triggered by some recent work into metrics, specifically to see those related to PDB. Unfortunately our PDB integration tests were so old they weren't running on many systems, and running them on newer systems required updating beaker, which required updating the tests, and the next thing I knew the mouse was drawing a picture.


Gemfile
```diff
@@ -18,11 +18,11 @@ gem 'rake', :group => [:development, :test]
 
 group :test do
   gem 'rspec'
-  gem 'beaker', *location_for(ENV['BEAKER_VERSION'] || '~> 4.11')
-  gem "beaker-hostgenerator", *location_for(ENV['BEAKER_HOSTGENERATOR_VERSION'] || "~> 1.1")
-  gem "beaker-abs", *location_for(ENV['BEAKER_ABS_VERSION'] || "~> 0.1")
+  gem 'beaker', *location_for(ENV['BEAKER_VERSION'] || '~> 5.4')
```
donoghuc, line 21: Latest and greatest is 6 now. I think 5 was short lived. Any reason we need 5?

donoghuc:
Thank you so much for doing this Justin!

Gemfile
justinstoller: It's the earliest release that supports Debian 12. I was hoping I was going to get away without having to update everything at the time. Since that was foolish naivety I can go back and try latest.

donoghuc: Ah, I think 5 -> 6 was painless compared to 4 -> 5. If it turns out to be painful we can do it later.

donoghuc (approved):

PR #2885, opened 2024-10-02
(maint) Do not install docs and pin to 3.x when installing r10k in acceptance

justinstoller:


justinstoller:
This doesn't have any changes that our PR testing will catch.

I've ran this locally against a redhat9-64mdca layout (a layout that is failing in CI currently) and the full suite passed.

PR #2894, opened 2024-10-11
compiler.rb: Switch to modern facts

justinstoller:
All of those are available at least since Facter 3.

-----

This is just https://github.com/puppetlabs/puppetserver/pull/2878 but with a fixed up spec.


justinstoller:
The work was originally done by someone else and approved, I just cleaned up a test.

PR #2896, opened 2024-10-31
test

justinstoller:
testing


donoghuc:
Just in case this is a helpful trick you were not aware of (and apologies if it is something you already know and are intentially not doing). We have a `workflow_dispatch` trigger configured for pr testing https://github.com/puppetlabs/puppetserver/blob/8c0154d45742ecde1ea3d4879b739cdd35e628be/.github/workflows/pr-testing.yaml#L4 . So if you can stand the GUI (Go to the "Actions" tab, then select the action you want to run (on the left nav) then there should be a dropdown to "Run Workflow") you can just trigger a run against the head of any branch you want without having to open a trivial PR. :)

PR #2899, opened 2025-01-25
(maint) Drop `csr.version = 2` setting

jcharaoui:
OpenSSL versions 3.4.0 and later no longer support setting a CSR version other than 1, because there are no other valid versions. Rather, it is recommended to stop specificying a CSR version altogether.

https://github.com/openssl/openssl/commit/397051a40db2d68433b842e7505e8cf3c9effb36


jcharaoui:
Superseded by #2902

